PR #11236, opened 2023-05-16
use partial decoder implementation for ptblank3 [Peter Wilhelmsen, Samuel Neves]

DavidHaywood:
rather than loading a decrypted ROM over the NAND, thus creating a 'bootleg' style set, this uses a partial decrypter with the properly calculated linear part, and only the non-linear coming from a look-up ROM.

the non-linear is still tied to the game, but this is still a step closer to the real solution.

so far the proper calculation for the non-linear is still proving illusive.

the downside of this is that the code is a bit less 'clean'
the upside is that the decryption research is in a better state for somebody to pick up later
furthermore nvram files created when running MAME using this code will be valid, rather than incorrectly containing pre-decrypted code

PR #11244, opened 2023-05-18
namco/namcos10: Add partial decrypter for Point Blank 3 [Peter Wilhelmsen, Samuel Neves]

987123879113:
Redo of https://github.com/mamedev/mame/pull/11236 + fix a small save state-related explosion caused by the SPU if you save state in any PS1-based game at the wrong time (when a "valid" reverb preset isn't in the registers). 

`spu_reverb_cfg` should also probably be set to null when a valid reverb preset isn't found, but that'd have noticeable changes to the audio for potentially all PS1 games and I don't have the energy to test that.

Edit: To give more extra info. The save state issue happens with `mrdrilr2`. Save state on the title screen or similar and load the save state again from command line using something like `mame mrdrilr2 -state 1`. It immediately crashes MAME due to the null pointer without the patch here. The reverb settings in the registers don't match any known presets so `spu_reverb_cfg` is null when the process function is called.

### DIFF
--- a/src/mame/namco/ns10crypt.cpp
+++ b/src/mame/namco/ns10crypt.cpp
@@ -124,13 +124,23 @@ really exist.
 DEFINE_DEVICE_TYPE(MRDRILR2_DECRYPTER, mrdrilr2_decrypter_device, "mrdrilr2_decrypter", "Mr Driller 2 decrypter")
 
 DEFINE_DEVICE_TYPE(NS10_TYPE2_DECRYPTER, ns10_type2_decrypter_device, "ns10_type2_decrypter", "Namco System 10 Type 2 decrypter")
+DEFINE_DEVICE_TYPE(NS10_TYPE2_DECRYPTER_NONLINEAR, ns10_type2_decrypter_nonlinear_device, "ns10_type2_decrypter_nonlinear", "Namco System 10 Type 2 decrypter (non-linear bit lookup table)")
 
 // base class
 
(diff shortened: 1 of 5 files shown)


src/mame/namco/ns10crypt.cpp
```diff
@@ -257,8 +273,106 @@ void ns10_type2_decrypter_device::init(int iv)
...
+	if (nonlinear_table == nullptr)
+		fatalerror("ns10_type2_decrypter_nonlinear_device: Non-linear bit lookup table memory region not found");
+
+	save_item(NAME(m_mask));
+	save_item(NAME(m_previous_cipherwords));
+	save_item(NAME(m_previous_plainwords));
+	save_item(NAME(m_nonlinear_count));
+
+	m_nonlinear_region_base = nonlinear_table->base();
+	m_nonlinear_region_size = nonlinear_table->bytes();
```
cuavas, line 367: Can you turn `m_nonlinear_region_base` and `m_nonlinear_region_size` into `required_memory_region m_nonlinear_region`?  Then you can get the base pointer as `m_nonlinear_region->base()` and the size as `m_nonlinear_region->bytes()` and you don’t need the glue here to get a fatal error if the region isn’t present.

987123879113: Sounds good to me, gave it test and it does everything I was going for so I changed it.

Issue #11107, opened 2023-04-15
BGFX hlsl option has duplicate scanline variation settings


old-gamer:
### MAME version

0.253 (mame0253)

### System information

System information
Windows 10 Home, 64-bit, version 22H2, USA, US English
Installed RAM 8.00 GB
AMD A6-5200 APU with Radeon(TM) HD Graphics 2.00 GHz

### INI configuration details

```INI
n/a
```

### Incorrect behaviour

This is less a bug and more of an oversight, but the actual code for hlsl in bgfx has 2 scanline variation settings, both do the same thing, but only one "sticks" while the other's setting resets every time you run a game with it.

### Expected behaviour

Should only be one scanline variation setting, like d3d version of hlsl.

### Additional details

here it is on the screen circled, the second one is the bad one that just resets to default value every start up:
![yourcodingsuxbro](https://user-images.githubusercontent.com/127580438/232174394-10056d72-e7ba-4593-80bb-9cbf29c6e6ee.png)

in the json file there is this:
		{ "type": "float", "name": "scanline_variation",     "text": "Scanline Variation",         "default": 1.00, "max": 4.00, "min": 0.0, "step": 0.01, "format": "%1.2f", "screen": "crt" },
		{ "type": "float", "name": "scanline_bright_scale",  "text": "Scanline Brightness Scale",  "default": 2.00, "max": 4.00, "min": 0.0, "step": 0.01, "format": "%1.2f", "screen": "crt" },
		{ "type": "float", "name": "scanline_bright_offset", "text": "Scanline Brightness Offset", "default": 1.50, "max": 4.00, "min": 0.0, "step": 0.01, "format": "%1.2f", "screen": "crt" },
		{ "type": "float", "name": "scanline_jitter_amount", "text": "Scanline Jitter Amount",     "default": 0.00, "max": 4.00, "min": 0.0, "step": 0.01, "format": "%1.2f", "screen": "crt" },
		{ "type": "float", "name": "scanline_variation",     "text": "Scanline Variation",         "default": 1.00, "max": 4.00, "min": 0.0, "step": 0.01, "format": "%1.2f", "screen": "crt" },

and below only one is referenced:
			"uniforms": [
				{ "uniform": "u_scanline_alpha",         "slider": "scanline_alpha" },
				{ "uniform": "u_scanline_scale",         "slider": "scanline_scale" },
				{ "uniform": "u_scanline_bright_scale",  "slider": "scanline_bright_scale" },
[...]

Issue #11106, opened 2023-04-15
D3D hlsl: pausing a game while in windowed mode causes the screen to flash grey or white


old-gamer:
### MAME version

0.253 (mame0253)

### System information

System information
Windows 10 Home, 64-bit, version 22H2, USA, US English
Installed RAM 8.00 GB
AMD A6-5200 APU with Radeon(TM) HD Graphics 2.00 GHz
d3d video output

### INI configuration details

```INI
default mame config (mame -cc)
```

### Incorrect behaviour

When running a game in windowed mode, if you pause the game (keycode P) the screen will flash grey (default pause brightness 0.65) or white (full pause brightness 1.00) and then pause.  Doesn't happen in fullscreen mode.

### Expected behaviour

Game should just pause without any funny business like flashing screen.

### Steps to reproduce

1. start mame in windowed mode
2. run any game with Mame.exe -video d3d -hlsl -window
3. pause it.

### Additional details

Here is with default ini path settings in mame.ini (inipath  .;ini;ini/presets), the screen stays grey until you go fullscreen, and is okay going back to windowed again, but will do this again if you pause again:
![Untitled](https://user-images.githubusercontent.com/127580438/232180770-4b37547a-2e91-487e-9861-6a5e7765a029.png)

here's a pic with the pause brightness all the way up while pausing (pause brightness 1) and the presets path removed (inipath  .;ini), was really hard to printscreen when this happens as it is kind of fast:
![whitescreenpausedd3d](https://user-images.githubusercontent.com/127580438/232173809-3894fedb-6cf7-4d22-bc40-cec0542b023b.png)

This also does not happen on mame 0.251 and prior than I am aware of.

Edited to be a bit more informative as I went back to double check some things, glad I did as I didn't notice right away the different behavior when using the presets vs not.  The only thing that jumps out at me when comparing mame.ini vs raster.ini is this line in raster.ini:
shadow_mask_alpha         0.25

That is zeroed in the mame.ini.  So maybe something to do with that additional layer for the rendering is my guess on why one only flashes one frame vs the other sticking until unpaused.

Last edit:  I copied the mame.ini settings from the DIRECT3D POST-PROCESSING OPTIONS section to raster and it flashes once when paused just like disabling presets does.

Issue #11104, opened 2023-04-15
D3D hlsl: shadow_mask does not work unless bloom scale is a non-zero value


old-gamer:
### MAME version

0.253 (mame0253)

### System information

Windows 10 Home, 64-bit, version 22H2, USA, US English
Installed RAM 8.00 GB
AMD A6-5200 APU with Radeon(TM) HD Graphics 2.00 GHz
d3d video output

### INI configuration details

```INI
#
# CORE CONFIGURATION OPTIONS
#
readconfig                1
writeconfig               0

#
# CORE SEARCH PATH OPTIONS
#
homepath                  .
rompath                   roms
hashpath                  hash
samplepath                samples
artpath                   artwork
ctrlrpath                 ctrlr
inipath                   .;ini
fontpath                  .
cheatpath                 cheat
crosshairpath             crosshair
pluginspath               plugins
languagepath              language
swpath                    software

#
# CORE OUTPUT DIRECTORY OPTIONS
#
cfg_directory             cfg
nvram_directory           nvram
input_directory           inp
state_directory           sta
snapshot_directory        snap
diff_directory            diff
comment_directory         comments
share_directory           share

#
# CORE STATE/PLAYBACK OPTIONS
#
state                     
autosave                  0
rewind                    0
rewind_capacity           100
playback                  
record                    
exit_after_playback       0
mngwrite                  
aviwrite                  
wavwrite                  
snapname                  %g/%i
snapsize                  auto
snapview                  auto
snapbilinear              1
statename                 %g
burnin                    0

#
# CORE PERFORMANCE OPTIONS
#
autoframeskip             0
frameskip                 0
seconds_to_run            0
throttle                  1
sleep                     1
speed                     1.0
refreshspeed              0
lowlatency                0

#
# CORE RENDER OPTIONS
#
keepaspect                1
unevenstretch             1
unevenstretchx            0
unevenstretchy            0
autostretchxy             0
intoverscan               0
intscalex                 0
intscaley                 0

#
# CORE ROTATION OPTIONS
#
rotate                    1
ror                       0
rol                       0
autoror                   0
autorol                   0
flipx                     0
flipy                     0

#
# CORE ARTWORK OPTIONS
#
artwork_crop              0
fallback_artwork          
override_artwork          

#
# CORE SCREEN OPTIONS
#
brightness                1.0
contrast                  1.0
gamma                     1.0
pause_brightness          1.00
effect                    none

#
# CORE VECTOR OPTIONS
#
[...]

PR #11249, opened 2023-05-19
BGFX and D3D9 renderer fixes for issues #11104, #11106, and #11107

MooglyGuy:
Fix for #11107: Simply removed the erroneously copy/pasted line.

Fix for #11106: Avoiding redundant state-set calls was going awry. We now ensure that the first `set_texture` call of a scene always succeeds by clearing the cached texture pointer at frame start.

Fix for #11104: Calling `post_pass` and `downsample_pass` to generate bloom targets when bloom is disabled was causing issues when applying the post/shadowmask pass. Avoid calling them entirely if bloom is disabled, since the downsampled targets are not used anyway.


MooglyGuy:
Thanks RB - be sure to close out the associated issues citing this PR. :)

PR #11251, opened 2023-05-20
Update compilingmame.rst

bswieser:
And an undocumented and obscure preprocessor flag to allow SDL to properly set up X11 input events when no window manager is present.


cuavas:
Adding that to the make command won’t do anything.  The `SDLMAME_X11` preprocessor macro is defined automatically by the build scripts on Linux and other targets that use X11 displays if you don’t set `NO_X11`.

See here for where the macro is defined: https://github.com/mamedev/mame/blob/mame0253/scripts/src/osd/sdl_cfg.lua#L38

bswieser:
That's odd, as I didn't send that in via cmd or change in the makefile and that pragma didn't get set.  I'll need to check, and or see if I can recreate - I suspect make is broken (I know it doesn't clean out archives properly, maybe it picked up a  previously built archive).

PR #11224, opened 2023-05-13
Hash updates to softlist chds for dreamcast, psx, and sega/mega cd

ldolse:
Updates to hashes on a number of titles across a few systems, attempting to address some of #2517.  The hashes are all based on the sources referenced in the comments for those entries.

Some minor fixes:
 - macross psx part 1 cdrom1 was incorrectly tagged as 'cdrom', so changed this to 'cdrom1' to match the standard being used for multi-disc entries
 - dreamcast alienfront source comment had spaces in it's 'size' field, fixed that

The xml parser I used updated these two entries, but the change seemed appropriate to leave:
- tkm2evs2 and tkm2evs3 in psx - the chd part tag in each title was spanning two lines, merged to one line

Issue #2517, opened 2017-07-25
chdman creates different metadata since 0.176, mismatches the XML hashes


Pixelnarium:
While trying to add the unfinished CDi game [Voyeur II](http://www.gametronik.com/site/news-20304-EXCLU_Gametronik__Voyeur_II_sur_CDi_enfin_releasA_.html) to my library I had some trouble to get the SHA1 value MAME lists in its cdi.xml hash. It lists:

`<disk name="voyeur ii (v1.04)(11-06-96)(testing-beta)(access code 3333) (disc1)"                                       sha1="9dfffd6736dbb85fa5317000af9f42f98aad4a20"/>`

for the first disc. If I tried to convert the bin/cue files using chdman I get:

```
chdman.exe createcd -i Voyeur\ II/Voyeur\ II\ \(v1.04\)\(11-06-96\)\(Testing-Beta\)\(access\ code\ 3333\)\ \(disc1\).cue -o 187_v2.chd                              
chdman - MAME Compressed Hunks of Data (CHD) manager 0.187 (mame0187)                                                          
Output CHD:   187_v2.chd       
Input file:   Voyeur II/Voyeur II (v1.04)(11-06-96)(Testing-Beta)(access code 3333) (disc1).cue                                
Input tracks: 1                
Input length: 56:50:02         
Compression:  cdlz (CD LZMA), cdzl (CD Deflate), cdfl (CD FLAC)                                                                
Logical size: 626,080,896      
Compression complete ... final ratio = 80.6%                   
```

the result is:

```
chdman.exe info -i 187_v2.chd
chdman - MAME Compressed Hunks of Data (CHD) manager 0.187 (mame0187)
Input file:   187_v2.chd
File Version: 5
Logical size: 626,080,896 bytes
Hunk Size:    19,584 bytes
Total Hunks:  31,969
Unit Size:    2,448 bytes
Total Units:  255,752
Compression:  cdlz (CD LZMA), cdzl (CD Deflate), cdfl (CD FLAC)
CHD size:     504,908,597 bytes
Ratio:        80.6%
SHA1:         1890d51eed522e6a9947b387f4543ceb38ca56d6
Data SHA1:    454d9381ed6319f6657ed7201ef98020f3525bca
Metadata:     Tag='CHT2'  Index=0  Length=93 bytes
              TRACK:1 TYPE:MODE2_RAW SUBTYPE:NONE FRAMES:255752 PREGAP:0 P
```
[...]

PR #11252, opened 2023-05-20
updates for PSX CHD hashes

ldolse:
Resubmission of PSX hash updates as requested in PR #11224.

These are bulk converted using a script I've been [developing here](https://github.com/ldolse/slupdate).  The intention is that this can be used for the validation activities mentioned in the last PR discussion as well, but I'll take that up on #2517

A possible warning/error condition was mentioned for psx in the last PR:
> psx.xml may have .toc SW items lingering inside, and if you bulk converted without paying attention to the warning that chdman gives then you're introducing data loss;

I looked at chdman.cpp to try to find warning messages and I could only find this:
`printf("Warning: Track %d has subcode data.  bin/cue and gdi formats cannot contain subcode data and it will be omitted.\n", tracknum+1);`

This doesn't seem to align to the quote above so I'm not sure if I'm looking in the right place for warnings to monitor.  I do see many more error messages in chdman.cpp, but I'm not clear on whether any of those errors will still output a CHD file, and they also generally differ from the description above.

I've reviewed the logs that the chdman and the script generated while building these chds, and I don't see any warnings or errors from chdman, but I've attached the logs for reference.  
Logs: [PSX-CHD-buildlog.txt](https://github.com/mamedev/mame/files/11522058/PSX-CHD-buildlog.txt)

Note from the logs, the tool found 377 ROM archives which matched the source info in the Softlist, 5 of them already matched the hashes that were in the software list after creating chds - my assumption is those 5 entries were created post 0.175, the other 372 were updated hashes which make up this commit.

If there are specific things that would be good to log, or specific error/warning conditions to monitor (outside of the one mentioned above) to increase the confidence in this script let me know.

### DIFF
--- a/hash/psx.xml
+++ b/hash/psx.xml
@@ -43,7 +43,7 @@ For an overview of US PS1 discs that have not been dumped and added to redump's 
 		<sharedfeat name="compatibility" value="NTSC-U"/>
 		<part name="cdrom" interface="psx_cdrom">
 			<diskarea name="cdrom">
-				<disk name="007 racing (usa)" sha1="b0d5e1641367d45c6f187c5c6249371c95636e5a"/>
+				<disk name="007 racing (usa)" sha1="c0fffd6939c403a0b7a179b472ae768c06c05c26"/>
 			</diskarea>
 		</part>
 	</software>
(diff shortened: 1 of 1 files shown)


ldolse:
added 805 new hashes with the latest commit - working to get everything in redump updated to simplify the validation piece, about 2/3 of the psx list is sourced there.
chdman logs for reference:
[psx_build_log2.txt](https://github.com/mamedev/mame/files/11555712/psx_build_log2.txt)

ldolse:
last commits completes all redump sourced entries, along with a few from the no-intro 'non-redump' DATs, total 2032 hash updates, out of a total of 2131 dat matches across the two source groups (of the remaining 99, 85 would be entries created post 0.1706 with matching hashes).  14 non-redump entries are using old clonecd sources, I wasn't able to get chdman to process these at all, it sits and hangs indefinitely with a single thread at 99% utilisation, I'll open a separate ticket for that.

chdman build log:
[psx_build_log-3.txt](https://github.com/mamedev/mame/files/11651858/psx_build_log-3.txt)

PR #11264, opened 2023-05-22
apple2gs: rom1 ram size logic

ksherlock:
The ROM 1, the ram size calculation is a little off. For the base case of 256K-1.125MB, the e0/e1 memory needs to be subtracted as well. (Currently, a 256K apple2gsr1 has 384k).  Once you start getting into 2MB and 4MB, etc memory cards, that memory shouldn't be subtracted - if you install a 4MB card, you have 4megs on top of 0/1/e0/e1, for a total of 4.25MB.


rb6502:
Thanks!  Also someone relayed your modulate.po.  It works in one emulator now, LOL.

PR #11267, opened 2023-05-22
gameboy.xml: Software list addition

ICEknigh7:
New partly working entry:
- Debagame Tester: Second Impact (Ver 10.24) [gameboycamera.com, LuigiBlood]
image

Also fixed entry for `Bart Simpson's Escape from Camp Deadly (Japan) [bartescj]`, which apparently didn't load due to not having a slot type assigned.


cuavas (changes requested):
Please just transliterate the Japanese title.

Also, does this use any more M64282FP features than the retail Game Boy Camera cartridge?

ICEknigh7:
So... "DebuCameTester", then? I was a bit reluctant to use the full title but that's what I was told it was called. Currently waiting to know where exactly it was called that.

Will ask about any extra features.

987123879113:
"Debagame Tester SECOND IMPACT Ver10.24"

The title's a pun. Also the 3rd character is a ガ not a カ so trying to translate it to "camera" isn't right.

ICEknigh7:
Regarding this:
> Also, does this use any more M64282FP features than the retail Game Boy Camera cartridge?

The answer is:
> There’s the ability to adjust some edge controls (50% to 500%), but that adjusts the v-ref along with it. That’s it, as far as I can tell

cuavas:
The web editor breaks badly on large files.  This no longer applies properly.

ICEknigh7:
Fixed, should I add anything else or is it good to go?

PR #11273, opened 2023-05-24
dragon_cart.xml: Entry replaced with good dump, added additional info from PCB, updated softlist to new standards

ICEknigh7:
Modified software list formatting with the new standards.

Also added this entry:
- Cyrus Chess [Clawgrip, ICEknight]

PCB has two EPROMs, previous entry had been dumped as one ROM. The new dumps come from reading the EPROMs directly.

And finally, added more info to entry for "Test de conducción", based on PCB pics.

### DIFF
--- a/hash/dragon_cart.xml
+++ b/hash/dragon_cart.xml
@@ -8,7 +8,7 @@ license:CC0-1.0
 	ROMs dumped from RAM and marked as "bad dumps":
 		- Cumana DOS v2.0
 
-	TO DO: Most cartridges with two EPROMs have been dumped as one single ROM, such entries should be marked as "bad dump" until proper split dumps exist.
+	TO DO: Most cartridges with two EPROMs have been dumped as one single file, such entries should be marked as "bad dump" until proper split dumps exist.
 
 -->
 <softwarelist name="dragon_cart" description="Dragon cartridges">
--- a/hash/dragon_cart.xml
+++ b/hash/dragon_cart.xml
@@ -17,10 +17,10 @@ license:CC0-1.0
 		<description>Astroblast</description>
 		<year>1982</year>
 		<publisher>Dragon Data</publisher>
-		<info name="developer" value="Mark Data Products" />
+		<info name="developer" value="Mark Data Products"/>
 		<part name="cart" interface="coco_cart">
-			<dataarea name="rom" size="8192">
-				<rom name="astroblast (1982) (dragon data).rom" size="8192" crc="61143386" sha1="64e103c787d551d9504a6e2c5d70fb2c676c92e1"/>
+			<dataarea name="rom" size="0x2000">
+				<rom name="astroblast (1982) (dragon data).rom" size="0x2000" crc="61143386" sha1="64e103c787d551d9504a6e2c5d70fb2c676c92e1"/>
 			</dataarea>
 		</part>
 	</software>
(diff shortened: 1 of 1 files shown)


ICEknigh7:
Pics:
image
image
image
image

ICEknigh7:
Should be the labels for what's in ic3 be included, too? Might be the same for every cartridge.
image

cuavas (commented):
* Was the old dump actually bad data, or does this just split it correctly?
* It’s a bit weird to be using hex offsets with decimal sizes.  Stick with one radix or the other within an entry.
* It’s a lot easier to see how loading works if you keep the size and offset together.

cuavas:
> Should the labels for what's in ic3 be included, too? Might be the same for every cartridge.

I don’t think there’s any need to mention the AND gates there.  PCB types are often noted when known (RE-075 for the second cartridge).

ICEknigh7:
This redump both splits it correctly and verifies the old dump.

About the hex and decimal, I just followed the style of the softlist.

Not sure what you mean with the third point. Again, I followed the formatting in the rest of the softlist.

Maybe this could do with an automated reformatting of those elements as well as some mass-splitting for the ROMs (and mark them as "not verified"), but that surpasses the purpose of this PR.

cuavas:
The current formatting of the software list is bad, and the rot should be stopped:
* Do not mix hexadecimal and decimal sizes/offsets in the same software item.  Choose one or the other and stick with it.
* Keep the size and offset attributes next to each other.  Don’t have them separated by multiple other attributes – that just makes it hard to read.

ICEknigh7:
Was bored enough today to adapt the whole softlist to this new formatting, but these things are better done via script by people who do scripts.

Also, maybe clrmamepro should be notified about these changes.

hash/dragon_cart.xml
```diff
@@ -153,25 +153,31 @@ Compiled by K1W1 and Cowering (from GoodCoCo)
...
 		<part name="cart" interface="coco_cart">
-			<dataarea name="rom" size="8192">
-				<rom name="cyrus chess (1982)(a0108)(dragon data).rom" size="8192" crc="d4c36a96" sha1="953f35586b0fc7c5b1f5ccb9fd20e0e6e1041d9c"/>
+			<feature name="pcb" value="21-83"/>
+			<feature name="ic1" value="TMS 2532JL-45 MEP8332"/>
+			<feature name="ic2" value="TMS 2532JL-45 MEP8332"/>
+			<dataarea name="rom" size="0x2000">
+				<rom name="0.ic1" size="0x1000" offset="0x0000" crc="6B452AC3" sha1="38f5c6804aa7c8c5ff791bfcfc01318449800fcd"/>
+				<rom name="1.ic2" size="0x1000" offset="0x1000" crc="9A0D3B20" sha1="104a0d27ec35f25af33e784e302b079bf05f4f45"/>
 			</dataarea>
```
cuavas, line 181: Please use lowercase hex digits in CRCs.

hash/dragon_cart.xml
```diff
@@ -101,9 +101,9 @@ Compiled by K1W1 and Cowering (from GoodCoCo)
 		<year>198?</year>
 		<publisher>Cumana</publisher>
 		<part name="cart" interface="coco_cart">
-			<feature name="slot" value="dragon_fdc" />
-			<dataarea name="rom" size="16128">
-				<rom name="cumana dos v2.0 (198x)(cumana).rom" size="16128" crc="32910d47" sha1="9fd99441d32344b1a64c8c0a61c61714761a88f9" status="baddump"/>
+			<feature name="slot" value="dragon_fdc"/>
+			<dataarea name="rom" size="0x3F00">
+				<rom name="cumana dos v2.0 (198x)(cumana).rom" size="0x3F00" crc="32910d47" sha1="9fd99441d32344b1a64c8c0a61c61714761a88f9" status="baddump"/>
 			</dataarea>
```
cuavas, line 107: Please use lowercase hex digits in sizes.

hash/dragon_cart.xml
```diff
@@ -3,24 +3,24 @@
...
 license:CC0-1.0
 
-Compiled by K1W1 and Cowering (from GoodCoCo)
+	Nearly all of the cartridges on this Dragon softwarelist will also run on the CoCo 2/3, CP400.
+	
+	ROMs dumped from RAM and marked as "bad dumps":
+		- Cumana DOS v2.0
 
-  * Nearly all of the cartridges on this Dragon softwarelist will also run on the CoCo 2/3, CP400.
+	TO DO: Most cartridges with two EPROMs have been dumped as one single file, such entries should be marked as "bad dump" until proper split dumps exist.
```
cuavas, line 11: If you write “TODO” in all caps with no space, editors and analytics tools pick it up.

The initial tabs inside the comment here will be replaced with spaces by srcclean because the opening delimiter of the comment is at the left margin.  You should probably expand them yourself to avoid churn.

cuavas (commented):
If you’re going to reformat all the self-closing tags, I’d prefer if you *added* the space before the slash to the ones that lack it, rather than removing it from the ones that have it (i.e. `<foo x="y" />` with the space between the attribute value and the slash of the self-closing tag).

ICEknigh7:
> If you’re going to reformat all the self-closing tags, I’d prefer if you added the space before the slash to the ones that lack it, rather than removing it from the ones that have it

Yeah, had already thought of doing that but wasn't too sure of what was preferable after seeing other softlists.

Should be good to go now.

PR #11274, opened 2023-05-24
spg format + [tsconf_betadisc_flop.xml] updated games/demos [holub, tslabs]

holub:
* add spg file format
* [tsconf_betadisc_flop.xml] soflist updates

spg format currently has no alternatives, because all existing don't support pagination with >128K RAM.

### DIFF
--- a/src/mame/sinclair/spec_snqk.cpp
+++ b/src/mame/sinclair/spec_snqk.cpp
@@ -130,7 +130,8 @@ SNAPSHOT_LOAD_MEMBER(spectrum_state::snapshot_cb)
 	}
 	else if (image.is_filetype("sp"))
 	{
-		if ((snapshot_data[0] != 'S' && snapshot_data[1] != 'P') && (snapshot_size != SP_NEW_SIZE_16K && snapshot_size != SP_NEW_SIZE_48K))
+		if ((snapshot_data[0] != 'S' || snapshot_data[1] != 'P')
+			|| (snapshot_size != SP_NEW_SIZE_16K && snapshot_size != SP_NEW_SIZE_48K))
 		{
 			if (snapshot_size != SP_OLD_SIZE)
 				return std::make_pair(image_error::INVALIDLENGTH, "Invalid .SP file size");
@@ -160,9 +161,9 @@ SNAPSHOT_LOAD_MEMBER(spectrum_state::snapshot_cb)
 	}
 	else if (image.is_filetype("sem"))
 	{
-		if (snapshot_data[0] != 0x05 && snapshot_data[1] != 'S' &&
-			snapshot_data[2] != 'P' && snapshot_data[3] != 'E' &&
-			snapshot_data[4] != 'C' && snapshot_data[5] != '1')
+		if (snapshot_data[0] != 0x05 || snapshot_data[1] != 'S'
+			|| snapshot_data[2] != 'P' || snapshot_data[3] != 'E'
+			|| snapshot_data[4] != 'C' || snapshot_data[5] != '1')
 		{
 			if (snapshot_size != SEM_SIZE)
 				return std::make_pair(image_error::INVALIDLENGTH, "Invalid .SEM file size");
@@ -192,8 +193,8 @@ SNAPSHOT_LOAD_MEMBER(spectrum_state::snapshot_cb)
 	}
 	else if (image.is_filetype("snx"))
 	{
-		if (snapshot_data[0] != 'X' && snapshot_data[1] != 'S' &&
-			snapshot_data[2] != 'N' && snapshot_data[3] != 'A')
+		if (snapshot_data[0] != 'X' || snapshot_data[1] != 'S'
+			|| snapshot_data[2] != 'N' || snapshot_data[3] != 'A')
 			return std::make_pair(image_error::INVALIDIMAGE, "Invalid .SNX file header");
 
 		setup_snx(&snapshot_data[0], snapshot_size);
@@ -204,6 +205,15 @@ SNAPSHOT_LOAD_MEMBER(spectrum_state::snapshot_cb)
 			return std::make_pair(image_error::INVALIDLENGTH, "Invalid .FRZ file size");
 
 		setup_frz(&snapshot_data[0], snapshot_size);
+	}
+	else if (image.is_filetype("spg"))
+	{
+		if (snapshot_data[32] != 'S' || snapshot_data[33] != 'p' || snapshot_data[34] != 'e' || snapshot_data[35] != 'c'
+			|| snapshot_data[36] != 't' || snapshot_data[37] != 'r' || snapshot_data[38] != 'u' || snapshot_data[39] != 'm'
+			|| snapshot_data[40] != 'P' || snapshot_data[41] != 'r' ||snapshot_data[42] != 'o' || snapshot_data[43] != 'g')
+			return std::make_pair(image_error::INVALIDIMAGE, "Invalid .SPG file header.");
+
+		setup_spg(&snapshot_data[0], snapshot_size);
 	}
 	else
 	{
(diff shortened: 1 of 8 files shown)


holub:
Diff looks strange with others commits leaked into this PR after merge.
I can reopen if required.

cuavas:
This seems to have got mixed up a bit with release branch commits from last month showing up as part of it.  Could you merge or rebase to make GitHub less confused?

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -205,6 +205,15 @@ SNAPSHOT_LOAD_MEMBER(spectrum_state::snapshot_cb)
 
 		setup_frz(&snapshot_data[0], snapshot_size);
 	}
+	else if (image.is_filetype("spg"))
+	{
+		if (snapshot_data[32] != 'S' && snapshot_data[33] != 'p' && snapshot_data[34] != 'e' && snapshot_data[35] != 'c'
+			&& snapshot_data[36] != 't' && snapshot_data[37] != 'r' && snapshot_data[38] != 'u' && snapshot_data[39] != 'm'
+			&& snapshot_data[40] != 'P' && snapshot_data[41] != 'r' &&snapshot_data[42] != 'o' && snapshot_data[43] != 'g' )
+			return std::make_pair(image_error::INVALIDIMAGE, "Invalid .SPG file header.\n");
```
cuavas, line 213: If you’re checking for a literal string, you should be using `||`, when the condition is negative, not `&&`.  This will check for *one* valid character, not the whole string.

Remember De Morgan’s theorem:
* !(X && Y) → !X || !Y
* !((X == a) && (Y ==b)) → !(X == a) || !(Y == b) → (X != a) || (Y != b)

Also, there should be no newline on the end of the error string here.

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -1766,6 +1775,358 @@ void spectrum_state::snx_decompress_block(address_space &space, uint8_t *source,
 	}
 }
 
+static u16 hrust_decompress_block(uint8_t *dest, uint8_t *source, uint16_t size)
+{
```
cuavas, line 1779: You should at least make the `source` data `const` to avoid accidents.  Also, is the size here the available source data, or the space at the destination?  You really need both for safety.

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -1766,6 +1775,358 @@ void spectrum_state::snx_decompress_block(address_space &space, uint8_t *source,
...
+			len = block_size;
+			idx = 0;
+			eof = false;
+
+			bits = get_byte();
+			bits += 256 * get_byte();
+		}
+
+		u8 get_byte(void)
+		{
```
cuavas, line 1804: You don’t need the `void` here – this is C++.

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -1766,6 +1775,358 @@ void spectrum_state::snx_decompress_block(address_space &space, uint8_t *source,
...
+				return 0;
+			}
+			return *p++;
+		}
+
+		u8 get_bit()
+		{
+			u16 mask[] = {0x8000, 0x4000, 0x2000, 0x1000, 0x800, 0x400, 0x200, 0x100, 0x80, 0x40, 0x20, 0x10, 0x8, 0x4, 0x2, 0x1};
+
+			u8 bit = (bits & mask[idx]) ? 1 : 0;
```
cuavas, line 1817: `mask` should be `constexpr`, but do you actually need it?  Can’t you just express this as `u8 bit = BIT(bits, 15 - idx);` anyway?

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -2397,6 +2758,124 @@ void spectrum_state::setup_z80(uint8_t *snapdata, uint32_t snapsize)
...
 
+/*
+	Load a .SPG (Spectrum Prog) file.
+
+	v1.1: https://raw.githubusercontent.com/tslabs/zx-evo/master/pentevo/docs/Formats/SPGv1_1.txt
+	v1.0: https://raw.githubusercontent.com/tslabs/zx-evo/master/pentevo/docs/Formats/SPGv1_0.txt
+	v0.2  https://raw.githubusercontent.com/tslabs/zx-evo/master/pentevo/docs/Formats/SPGv0_2.txt
+*/
+void spectrum_state::setup_spg(u8 *snapdata, u32 snapsize)
+{
```
cuavas, line 2769: You should make the input `const`.

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -2397,6 +2758,124 @@ void spectrum_state::setup_z80(uint8_t *snapdata, uint32_t snapsize)
...
+
+				case 0x02:
+					hrust_decompress_block(m_ram->pointer() + (page << 14) + offs, data, size);
+					break;
+
+				default:
+					//
+					logerror("Unsupported compression\n");
+					return;
+			}
```
cuavas, line 2871: You should check the RAM size or you could write off the end.

cuavas (commented):
Was the decompression code cribbed from somewhere?  It seems to be somewhat lacking in error checking as well as having some odd characteristics.

tslabs:
> Was the decompression code cribbed from somewhere? It seems to be somewhat lacking in error checking as well as having some odd characteristics.

The decompression code has been copied from https://github.com/tslabs/zx-evo/blame/master/pentevo/unreal/Unreal/depack.cpp with my permission to any modifications

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -2397,6 +2758,124 @@ void spectrum_state::setup_z80(uint8_t *snapdata, uint32_t snapsize)
...
+	m_maincpu->set_state_int(Z80_IFF1, spg->cpu.ei);
+
+	bank3_set_page(spg->win3_pg);
+
+	u8 *data = &spg->data;
+	for (u8 i = 0; i < spg->blocks_num; i++)
+	{
+			const u16 size = (spg->blocks[i].size512 + 1) * 512;
+			const u8 page = spg->blocks[i].page;
+			const u16 offs = (spg->blocks[i].address512) * 512;
```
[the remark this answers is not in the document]
holub: It's possible to be trunketed. 256 is max allowed but snapshot can contain any less numbers if blocks.

src/mame/sinclair/spec_snqk.cpp
```diff
@@ -2397,6 +2758,124 @@ void spectrum_state::setup_z80(uint8_t *snapdata, uint32_t snapsize)
...
+				u8 size512 : 5; // 0-512b, 1f-16kB
+				u8 reserved_2 : 1;
+				u8 compression : 2; // 0-NONE, 1-MLZ, 2-HRUST
+			};
+			u8 page; // #00-#DF
+		} blocks[256];
+		u8 data;
+	};
+
+	spg_v10* spg = (spg_v10*)snapdata;
```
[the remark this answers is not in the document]
holub: It's not possible because declared blocks array will always oversize real data. If that can cause problems I can separate blocks from struct and process them iteratively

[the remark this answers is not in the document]
holub: @cuavas sorry to bother... what is the preference here?

cuavas:
This PR still has multiple things in it that might be better separated:
* General cleanup and bug fixes in the Spectrum snapshot formats handling.  This stuff should be easy to get merged if it’s isolated.
* The SPG format handling stuff itself.  This is a bit more complicated because it makes licensing messy if it gets pasted into the middle of a MAME source file.  It would be easier to discuss this part on its own.
* A bunch of SPG snapshots in a software list.  They seem to have been added to a BetaDisc floppy image software list.  Should they have a software list of their own, since they appear to be memory snapshots that don’t correspond to physical media?  If the software list addition was isolated, it may be possible to get that merged with `supported="no"` to get the media images in circulation before the SPG format support is ready.

holub:
Oh, I'm sorry for bothering. Tried to rebase it in order to revisit. Will close the PR and reopen when ready.

holub:
#13122

PR #11276, opened 2023-05-26
heathkit/h89.cpp add support for z37 soft-sector controller

mgarlanger:
Adding the z37 soft-sectored floppy disk controller for the Heath H89.

Note: I think I connected everything properly, but this is not currently working. I'm putting up a PR to potentially get some help. From debugging, it appears that it is getting to the point of waiting for DRQ signal for reading in the sector, but instead of getting a DRQ, it is getting an INTRQ..


ajrhacker:
I'm getting a bad feeling about the interrupt control device here. I don't know how the Z37's custom interrupt control really works, but ideally the emulation should *not* be relying on `HOLD_LINE` and `set_input_line_and_vector`.

mgarlanger:
Could you expand on why HOLD_LINE and set_input_line_and_vector is bad? The interrupt controller logic in the H89 and by extension the modification that the Z37 provided, raises the interrupt line to the CPU and places either an RSTx or EI instruction on the data bus. Similar logic was already in the h89.cpp code for handling the timer interrupt, this change just moves it to the interrupt controller and generalizes it for any of the interrupt priority levels.

mgarlanger:
Making progress. I updated code to turn on the floppy motor, and fix the interrupt block logic.  It will now partially load the disk, execute the loaded program, but then reports a disk read failure.
Screen Shot 2023-05-27 at 12 01 01 AM

rb6502:
HOLD_LINE is something that hardware doesn't do - it asserts the line and automatically deasserts it after some fixed amount of time.  It's preferred that the interrupt controller explicitly does ASSERT_LINE to raise and CLEAR_LINE to lower at the appropriate time (usually on a register write to whatever device is causing the interrupt).

I'm not sure what's wrong about set_input_line_and_vector() though, @ajrhacker can you expand on that?

ajrhacker:
`set_input_line_and_vector` is every bit the hacky abstraction `HOLD_LINE` is; not surprisingly, the two are often used together. It stores the "current" vector in the execute interface instead of doing the priority mixing more typical of hardware. (On hardware, the vector normally can change until the interrupt acknowledge signal is asserted.)

rb6502:
Ok, what does the right way look like, for OP's benefit?  A live example in MAME would be good.

mgarlanger:
Yes, I would like to see a better approach. My luck, I would find one to copy that is a worse pattern than this. 🙈

mgarlanger:
Looks like `set_irq_acknowledge_callback` might be the right approach, but I'll wait to hear back, before making any changes.

ajrhacker:
Yes, `set_irq_acknowledge_callback` is much closer to the ideal. (I've been thinking of the ideal solution for Z80 for a long time, but that entails dealing with such difficult matters as daisy-chained vectoring, RETI signaling and the Z80's brain-damaged handling of CALL vectors in IM 0.)

mgarlanger:
@ajrhacker The callback approach is missing functionality that the `HOLD_LINE` had. After changing my code to use the CB approach, it was not hanging when I tried to boot. I tracked down the issue to the Z80 code: it does not accept arbitrary instructions for the call back. In my case, I put in the one instruction I needed `EI`, in a very hacky way by adding an if check in the default block: 
```				default:        /* rst (or other opcodes?) */
					if (irq_vector == 0xfb)
					{
						// EI
						ei();
					}
					else
					{
						/* RST $xx cycles */
						CC(op, 0xff);
						T(m_icount_executing - MTM * 2);
						wm16_sp(m_pc);
						PCD = irq_vector & 0x0038;
					}
					break;
```
With this change to z80.cpp, I was able to get my code performing in the same way it was before I switched from the HOLD_LINE to the cb approach. It still failing, but is loading data, executing it, and has the same error as before. 

Someone more familiar with the Z80 code, should be adding this in a proper way. Would you have time for that?

mgarlanger:
Updated to use `set_irq_acknowledge_callback`, but still debugging why it is only partially booting.

mgarlanger:
Successful booting of HDOS:
Screen Shot 2023-06-05 at 10 58 50 PM

Successful booting of H37 Diag disk
Screen Shot 2023-06-05 at 12 19 45 AM

mgarlanger:
Note system is now working and properly boots HDOS, but there is still some problems with CP/M.

mgarlanger:
Fix the issue with CP/M booting by swapping the order of handling drq/intrq from the wd floppy controller.  This PR is a ready for review. 

Screen Shot 2023-06-08 at 11 24 50 PM

mgarlanger:
Actually, it looks like reading is good. For writing, both CP/M and HDOS says the disk is write-protected, even when I select "write to another image".  The other option "write to diff" said it was unimplemented.  It's possible I'm doing something wrong in the UI.

mgarlanger:
Looks like writing is fine too. My first issue was that I used 40-track disks in an 80-track drive. Both CP/M and HDOS properly detected it and set the disk into a R/O mode. Second issue seems to be related to existing IMD -> flux converter. Even with 40-track drives it was erroring. But by creating a new disk (in MFI format), formatting (CP/M) or init (HDOS) and then SYSGENing them. I was able to boot from them, and read/write without issue. 

This PR is ready to be reviewed/merged.

mgarlanger:
@ajrhacker @rb6502 would either of you have a chance to review this PR?

PR #11281, opened 2023-05-27
factory reset ptblank3 + gunbalina to defaults and provide functional default gun calibration

DavidHaywood:
I've split this from the other ns10 PR as maybe default calibration isn't wanted


cuavas:
Will it do the “initial boot” sequence on start with this?  Personally, I prefer machines to be completely clean on first start.  I’m not a fan of them starting with some arbitrary setup in NVRAM if we know how to get them to think the NVRAM needs initialisation.

DavidHaywood:
the game doesn't show the 'first power on' regardless in this case

cuavas:
> the game doesn't show the 'first power on' regardless in this case

But do we know how to set sector 1 so it will?  I’d prefer to have it in that state if possible.

DavidHaywood:
it doesn't show that, even if it resets the sector

rb6502:
I like the First Boot state too, but for games with analog calibration it's important to ship them in a default-playable state in much the same way that we've moved away from making people do the EEPROM/Flash installs on Konami games.

cuavas:
I’ll merge it for now, just to annoy the PokéROMs.

PR #11285, opened 2023-05-28
Fixed Taito F3 / ES5510 performance regression

MooglyGuy:
-es5510.cpp: Avoid calling DESCRIBE_INSTR when not verbose-logging. [Ryan Holtz]


MooglyGuy:
This should be cherry-picked into the release branch if at all possible.

cuavas:
See 00c0b3501859a6f046b855a9589df3eda1832b32

PR #11294, opened 2023-05-31
Minor change to Markham.cpp documentation

ToastmanJack:
Markham.cpp and Markham.h drivers in their comment headings didn't actually list all the games supported so I thought it'd be best to clarify exactly what games use them in the heading. It's not a particularly large edit but as this is my first ever GitHub contribution I'm marking it as draft for now in case there's any pointers I need on formatting because I would like to comb through the rest of MAME's drivers doing similar work.

### DIFF
--- a/src/mame/sunelectronics/markham.cpp
+++ b/src/mame/sunelectronics/markham.cpp
@@ -2,7 +2,9 @@
 // copyright-holders:Uki
 /*****************************************************************************
 
-    Driver by Uki
+    Sun Electronics Markham hardware
+	
+	Driver by Uki
 
     TODO:
     - needs merging with ikki.cpp
--- a/src/mame/sunelectronics/markham.cpp
+++ b/src/mame/sunelectronics/markham.cpp
@@ -9,7 +9,7 @@
     TODO:
     - needs merging with ikki.cpp
     - look up schematics for all games
-   	- hook up actual SUN 8212 and figure out ROM mode communications
+    - hook up actual SUN 8212 and figure out ROM mode communications
 
     Notes:
     Banbam has a Fujitsu MB8841 4-Bit MCU for protection labeled SUN 8212.
--- a/src/mame/sunelectronics/markham.cpp
+++ b/src/mame/sunelectronics/markham.cpp
@@ -2,21 +2,12 @@
 // copyright-holders:Uki
 /*****************************************************************************
 
-    
-	Games supported:
-	
-	Markham (c) 1983 Sun Electronics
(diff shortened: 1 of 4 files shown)


ToastmanJack:
Does that look better?

src/mame/sunelectronics/markham.cpp
```diff
@@ -2,25 +2,22 @@
...
-    Markham (c) 1983 Sun Electronics
-    Strength & Skill (c) 1984 Sun Electronics
-
     Driver by Uki
 
     TODO:
     - needs merging with ikki.cpp
     - look up schematics for all games
-    - hook up actual SUN 8212 and figure out ROM mode communications
+   	- hook up actual SUN 8212 and figure out ROM mode communications
```
cuavas, line 11: The content of the change looks fine, but what happened to this line here?  It looks like it’s supposed to be the same, but was a space replaced with a tab or something?  Indents inside comments have indent beyond the level of the opening delimiter replaced with spaces by the srcclean tool for normalisation – if a space was replaced with a tab here, it will get changed back later causing churn.

happppp:
No, put something like "Sun Electronics Markham hardware"

src/mame/sunelectronics/markham.cpp
ToastmanJack: Not sure what happened here, I don't think I reformatted the TODO, but I'll fix it up

ToastmanJack:
Ready for merging, I think if this goes through ill work through the whole Sun Electronics driver folder next.

happppp:
Nothing personal, but IMO it was better the way it was.
When I open a driver source file, it's nice to see the first comment briefly describing what hardware is emulated. In this case (browsing through the code). There's Markham, Strength & Skill (different memorymap). And Strength & Skill + protection MCU.

ToastmanJack:
I honestly agree with you there, but if Vas says we don't need lists like that in the header then I'll trust their judgment.

happppp:
See for example this driver:
https://github.com/mamedev/mame/blob/master/src/mame/sunelectronics/arabian.cpp

The 1st comment "Sun Electronics Arabian hardware" is useful.
The game list in the comment section is as useless as it gets. Though I wouldn't go over all MAME's source code and remove them. Also in part, if the driver author wanted a game list in the driver, then so be it. There is no hard rule against it.

ToastmanJack:
OK then, I think you've got a point there.

ToastmanJack:
I changed the branch name but I forgot that closes the Pull, im a klutz

PR #11227, opened 2023-05-13
video/epic12.cpp: Fix clipping of CV1K games

buffis:
Change clipping for CV1K games to draw 32 pixels surrounding the visible area.

This can be easily seen in Muchi Muchi Pork, which has a VRAM viewer in Special mode (Object Test), which will show 32 px drawn around the visible areas of framebuffers.
See: https://www.arcade-projects.com/threads/research-into-cv1000-blitter-performance-and-behavior.24385/#post-364121

For most gamers, this wont really matter at all... except for in Muchi Muchi Pork, where changing this actually fixes a bug for Rafute (yellow ship).

When Bombing with Rafute (not the other two character), the screen background will go wavy in a sine-like pattern. Currently in mame, the top of screen will show black pixels when this happens.

Like this: http://img.buffis.com/mmppatch/without_patch.png

With this fix for clipping, the background will instead be visible correctly.
Like this: http://img.buffis.com/mmppatch/with_patch.png

Theres still some potential for artifacts though, both at top and on right side of screen... but this happens on PCB too.

Mame with patch, top artifact: http://img.buffis.com/mmppatch/mame_top_artifacts.png
Mame with patch, right artifact: http://img.buffis.com/mmppatch/mame_top_artifacts.png

PCB top artifact: http://img.buffis.com/mmppatch/pcb_top.jpg and http://img.buffis.com/mmppatch/pcb_top2.jpg
PCB right artifact: http://img.buffis.com/mmppatch/pcb_right.png

PCB footage of some bombs for reference: http://img.buffis.com/mmppatch/pcb_footage.mp4

Also renamed the "scroll registers" to have it more clear which one of these are actually used as a "scroll register" (or rather offset for drawing), and which one is strictly used for clipping.

PR #11295, opened 2023-05-31
video/epic12: Apply clipping to Blitter calculations as well

buffis:
Fix issue where some games do very large over-draws which when utilized in Blitter calculations will cause more delay than whats seen on hardware. The fog in S1 of Mushi Futari 1.5 is a good example of this.

On Mushi Futari 1.5, in the section before the boss, the Blitter delay seen on PCB is aprox 12 milliseconds at peaks. With this fix, it matches that quite well!

It's not 100% clear to me this is exactly how things behave, but this should be closer to the truth, until I can figure out a more reliable and reproducible test-case.

See discussions on this PR for background : https://github.com/mamedev/mame/pull/11227

For reference, Saleae logic analyzer output is available in: http://img.buffis.com/misc/futaris.sal

### DIFF
--- a/src/devices/video/epic12.cpp
+++ b/src/devices/video/epic12.cpp
@@ -436,6 +436,18 @@ inline void epic12_device::gfx_draw_shadow_copy(address_space &space, offs_t *ad
 		return;
 	}
 
+	// Clip the blitter operations, to have the calculations only respect the area being written.
+	// It's not 100% clear this is how this is performed, but it is clear that there should be some amount of clipping
+	// applied here to match the hardware. This way seems most likely, and maps well to the delays seen on hardware.
+	// One example of this being utilized heavily is the transparent fog in Mushihimesama Futari Stage 1. This is drawn as
+	// 256x256 sprites, with large parts clipped away.
+	dst_x_start = std::max(dst_x_start, (u16)m_clip.min_x);
+	dst_y_start = std::max(dst_y_start, (u16)m_clip.min_y);
+	dst_x_end = std::min(dst_x_end, (u16)m_clip.max_x);
+	dst_y_end = std::min(dst_y_end, (u16)m_clip.max_y);
+	src_dimx = dst_x_end - dst_x_start + 1;
+	src_dimy = dst_y_end - dst_y_start + 1;
+
 	m_blit_idle_op_bytes = 0;  // Blitter no longer idle.
 
 	// VRAM data is laid out in 32x32 pixel rows. Calculate amount of rows accessed.


dinkc64:
Found another testcase: the titlescreen of Futari - it animates slower/jerkier with the new fix disabled.

buffis:
Yes, that also uses the same 256x256 sprites for transparent fog as in S1.

Without the fix in this commit, there will be a lot of simulated time for "clipped away" data of that.

The change to the clipping margin I submitted earlier actually made this a lot more visible, since before it, some of those draws where fully outside clipping area and optimized away.

Basically, the state of things were:

- With the new Blitter code change but not the clipping margin fix: Incorrect, but not too bad since some of the bigger draws were still dropped
- With new Blitter code change, and clipping margin fix (for MMP mainly), too much delay for some games like Futari which does large writes like the fog.
- With this fix as well: Should be fine.

dinkc64:
@buffis Hi there,
el_rika has been doing tons of testing with this fix, he found:
while this fixed the unwanted slowdown in futari15, it also removed most of the wanted slowdown in the game - especially the type that happens when the screen fills with bullets right before a cancel.  I'm currently waiting for him to get access to a pc and do testing with a current MAME git-build to rule out any goofs on my part. (and also make some comparison videos)

best regards,
- dink

buffis:
Current mame build doesnt have the latest clipping fix btw, so probably not helpful to test on that. Either needs a binary built from head or next release

cuavas:
If someone says a “current MAME build” they likely mean one built from latest master, either compiled themselves or from GitHub CI.

dinkc64:
The MAME build used for testing is compiled from git on Jun 3rd, which includes this PR.

"Here is a quick gameplay of the level 3 Futari maniac with Palm Abn, that shows a couple of instances where the game should slow down considerably, though it does not:
https://youtu.be/l_nbXeNEkJk
0:15 and 0:53"

buffis:
I’ll look at it. Could also be cpu slowdown. It being in s3 maniac is a bit awkward since it makes it trickier to repro on pcb while capturing. Unfortunately not too much pcb reference footage either of 1.5 maniac.

dinkc64:
Thanks :)
I got him to find a scene in level 1 where slowdown is missing for easier testing:
(the MAME he's referring to is a git build from June 3rd)
"So here we have a small section towards the end of level 1 in Futari 1.5 (Normal Reco maniac), where the game slows down on pcb, but it does not at all, on buffi blitter.
This is very easy to reproduce, and i think that if the blitter can be tweaked to produce this slowdown, we are a big step in the right direction.

Here is me playing in buffi mame (the version you provided):
https://youtu.be/G7KrRKzG7nU
- 0:35 to 0:38 - completely full speed

Here is the pcb run:
https://youtu.be/ueWRfrNNuR4
- 1:37 to 1:40 - when shot with laser (keep pressed laser till both birds explode), the resulting blue gems cause slowdown.
"

best regards,
- dink

dinkc64:
@buffis 
Maybe it's not the place for this, but, I've been experiencing with simulating the cycles of the sh-3 pipeline.  With el_rika's help, he found an a part which is 100% dependant on proper cpu timing.  I swear, this guy literally the human reference for Mushihime-Sama (&Futari).
note: also includes a fix to the cycle consumption for mul* insns
note2: inv. cheat is enabled for this test

https://www.youtube.com/watch?v=ICteQPSqEN4

best regards,
- dink

buffis:
I havent had time to check out the stuff on the previous message on pcb yet, but will get to it some time.

Regarding simulating CPU slowdown, it is quite awkward. It is described very well in the SH3 datasheet, but actually implementing it quite hard. Doing it properly likely requires disabling DRC which tanks performance.

I have some real early proof-of-concept stuff in
https://github.com/mamedev/mame/commit/7e1dc50940747e92680abe3f2613c85b6a77c4f6
but it doesnt really work properly right now.

Curious how you are approaching it.

In general, CPU slowdown not related to blitter is very common on these games. Its not too hard to see when looking with logic analyzer. Example> https://buffis.com/research/cv1000-cpu-slowdown-investigated/

dinkc64:
Regarding https://github.com/mamedev/mame/commit/7e1dc50940747e92680abe3f2613c85b6a77c4f6
Your impl. is lightyears ahead of mine, would you mind if I try to get it working properly?  Do you have any ideas what went wrong?

buffis:
Feel free to mess around with it, but the main issue is that to be able to get the cache simulation to work, you need to disabled dynamic recompilation + all memory speedups, to get the caching simulation logic to trigger. This wont run on fullspeed even on my modern gaming machine, which causes major issues like sound glitching.

I'm also not fully confident that the caching is 100% correctly implemented, since I havent really double checked it. I started it as a WIP and then didnt look into it further.

PR #11299, opened 2023-06-01
Cleanup of Markham.cpp documentation

ToastmanJack:
Resubmission of #11294 since I renamed my branch to something more fitting, this removes the games list from the headers in markham.cpp, markham.h and markham_v.cpp and instead only specifies the Hardware supported by the driver.


happppp:
Ok I'll merge it for the work you've put into it. But like I said before, I don't think it's necessary to go through all of the drivers for these kind of comment refactors, unless something's actually wrong.

ToastmanJack:
That's fair, all this started because I was confused about the parent/clone status of BanBam and Pettan Pyuu anyway, and I really veered away from that thread.

PR #11307, opened 2023-06-03
z80.cpp - Add EI to IM0 handler

mgarlanger:
The heath/h89.cpp with a z37 soft-sectored controller uses IM0 and places an EI instruction on the bus when DRQ signal is received from the WD disk controller.

I'm putting this part up early to get comments. Maybe someone with more experience with the z80 code could implement the top TODO in this file
```
- Interrupt mode 0 should be able to execute arbitrary opcodes
```

### DIFF
--- a/src/devices/cpu/z80/z80.cpp
+++ b/src/devices/cpu/z80/z80.cpp
@@ -3323,6 +3323,8 @@ void z80_device::take_interrupt()
 					if (irq_vector == 0xfb)
 					{
 						// EI
+						CC(op, 0xfb);
+						T(m_icount_executing);
 						ei();
 					}
 					else if ((irq_vector & 0xc7) == 0xc7)
--- a/src/devices/cpu/z80/z80.cpp
+++ b/src/devices/cpu/z80/z80.cpp
@@ -3320,11 +3320,22 @@ void z80_device::take_interrupt()
 					PCD = irq_vector & 0xffff;
 					break;
 				default:        /* rst (or other opcodes?) */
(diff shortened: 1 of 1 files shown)


happppp:
It's not as easy as it sounds since AFAIK multi-byte opcodes fetch each byte from the databus as well.
This ei addition: it's missing cycle deduction. CC(xxx) and then T(m_icount_executing)

mgarlanger:
Maybe the first step for that TODO is just to just allow all the one-byte opcodes. The case block above this one assume the multi-byte opcodes for CALL/JP assumes `irq_vector` has all the bytes to the opcode, so it seems some other part of the code is handling assembling the bytes.  

Added the CC() and T() macros.

src/devices/cpu/z80/z80.cpp
```diff
@@ -3320,11 +3320,24 @@ void z80_device::take_interrupt()
...
-					wm16_sp(m_pc);
-					PCD = irq_vector & 0x0038;
+					if (irq_vector == 0xfb)
+					{
+						// EI
+						CC(op, 0xfb);
+						T(m_icount_executing);
+						ei();
+					}
+					else if ((irq_vector & 0xc7) == 0xc7)
```
mgarlanger, line 3330: This was added so that it won't silently execute the RST instruction, if it is not an RST instruction, so if anyone else runs into this, there will be a clear error log shown.

qeeg:
Oh no, why was this merged, this was quite clearly a hack to get one thing working. This could definitely break things.

PR #11317, opened 2023-06-06
js_sound.js: fix sound level default for underrunning case

holub:
this fix removes crackling of sound


src/osd/modules/sound/js_sound.js
```diff
@@ -174,9 +174,10 @@ function tick (event) {
 		}
 	}
 	//Pad with silence if we're underrunning:
```
DopefishJustin, line 176: This comment is no longer accurate, you're now padding with the previous sample and not silence

3rdparty/bx/include/bx/mutex.h
```diff
@@ -31,6 +31,7 @@ namespace bx
 		void unlock();
 
 	private:
+		[[maybe_unused]]
```
DopefishJustin, line 34: This is third-party code, so it would be better to turn off the warning for the "bx" project in scripts/src/3rdparty.lua (look for other uses of -Wno-unused) and that way the next pull from upstream won't overwrite it.

DopefishJustin (changes requested):

cuavas:
This will produce a DC level when paused, won’t it?  The usual recommended practice is to produce zero fill on underruns.

galibert:
The usual recommended practice is stupid then.  Crackling sucks hard.

holub:
> This will produce a DC level when paused, won’t it?  The usual recommended practice is to produce zero fill on underruns.

Maybe we can do latest on partial buffer and 0 if it entirely empty?
Would it satisfy both cases?

galibert:
No, it would crackle like hell too/

holub:
I tried pause during sound playing and didn't noticed crackling.

DopefishJustin (approved):

DopefishJustin:
I do get a popping sound on pause with jaguar tempst2k which is one of my usual test cases. Arguably better than crackling during the emulation though.

holub:
gents?

PR #11324, opened 2023-06-08
anes.cpp - allow to boot, preliminary video and inputs [Luca Elia, David Haywood]

DavidHaywood:
- added preliminary support for the blitter
- added a workaround for delayed banking
- added preliminary inputs
- added palette handling


src/mame/misc/anes.cpp
```diff
@@ -38,138 +39,447 @@ namespace {
...
+
+uint8_t anes_state::banked_rom_r(offs_t offset)
+{
+	uint8_t ret = m_maincpu->space(AS_PROGRAM).read_byte(offset + 0x2000);
+	if (!machine().side_effects_disabled())
+	{
+		if (m_bank_delay)
+		{
+			logerror("reading banked rom area with bank delay %04x: %02x\n", offset + 0x2000, ret);
+			m_banktimer->adjust(attotime::from_usec(1));
```
angelosa, line 246: Note: smc777.cpp uses Z80 R register because it's more reliable than using a timer. 
Even better, it's probably possible to use Z80 `refresh_cb` for both nowadays, likely this uses M1 cycles too.

DavidHaywood: I think poking directly into CPU registers is a pretty gross hack compared to a timer...

DavidHaywood: well M1 stuff is what MAME does through AS_OPCODES, which is where the code to start the timer is hooked up.

I don't know what refresh_cb is in this context, does it happen after an opcode?

angelosa: https://github.com/mamedev/mame/blob/8f3716ad43b45fbef9cb35e20bc84c078a727c60/src/devices/cpu/z80/z80.cpp#L522

z80 core reads the opcode, sends the callback to client finally updates the internal state.

happppp: refresh_cb is preliminary emulation of RFSH pin.
AS_OPCODES is (a weird way instead of a cb) for M1. Though I'd use this one since it's more likely how it works on hardware. And instead of a timer, add a counter? This is also more likely from hw point of view. A 1usec timer will have problems when/if the z80 core emulation gets more accurate.

DavidHaywood: I don't really see how a counter can work when it must execute after a single byte 'ret' or a 3 byte 'jmp'

happppp: because M1 is triggered once per opcode, unless you use prefixed opcodes.

DavidHaywood: but doesn't it happen when the first byte is fetched? meaning it'll fetch the wrong address on a jmp if we switch straight away, or fetch 2 wrong opcodes after a ret if we force a counter based delay?

happppp: Ah, AS_OPCODES is for opcode arg too? Then yes I'm wrong and it means our Z80 core doesn't emulate the M1 pin.

holub: Delayed bankswitch can be done with m1 counting:
https://github.com/mamedev/mame/blob/8f3716ad43b45fbef9cb35e20bc84c078a727c60/src/mame/sinclair/pentevo.cpp#L671

galibert:
No, AS_OPCODES is m1 only.  And the switch is probably on m1, otherwise
you'd have different pages for the jump opcode and the address to jump to.

src/mame/misc/anes.cpp
DavidHaywood: so can somebody just give me a (clean) solution here that will work for THIS case.

1. bankswitch value is written to port 0xfe
2. next opcode must be fetched and execute in full, including the oprand if present ('ret' or 'jmp' or 'call')
3. bank must change (so code after ret / jmp / call has been executed is in the new bank)

I volunteered to look at this because I wanted to concentrate on the video emulation, not this, this is putting me off.

DavidHaywood:
> No, AS_OPCODES is m1 only. And the switch is probably on m1, otherwise you'd have different pages for the jump opcode and the address to jump to.

So you're rejecting this approach even if it could allow me to continue to figure out the video? or you're disagreeing with some of the suggestions above?

src/mame/misc/anes.cpp
happppp: operands don't go through AS_OPCODES.

happppp: I removed my local anes.cpp so I can't check, but yeah I can see why the debugger will go wrong.

now:
```
debugger_instruction_hook(PCD);
uint8_t opcode = rop();
````

should be:
```
assert M1
callback to anes bankswitch here
debugger_instruction_hook(PCD);
uint8_t opcode = rop();
clear M1
````

Comment (DavidHaywood) [reply src/mame/misc/anes.cpp:246]: yes, oprands don't go through AS_OPCODES
but the opcode comes first, I don't see how the logic you've presented would know to not switch the bank before the oprands have been fetched, or in the case of a ret, know that none are needed?

**Comment by galibert** [comment]
And the multiple-m1-cycles is probably not handled by the hardware in the
first place :-)

File: src/mame/misc/anes.cpp [hunk 2]
```diff
@@ -38,150 +39,461 @@ namespace {
...
 void anes_state::prg_map(address_map &map)
 {
-	map(0x0000, 0xefff).rom();
-	map(0xf000, 0xffff).ram();
+	map(0x0000, 0x1fff).rom();
+	map(0x2000, 0xefff).bankr(m_rombank);
+	map(0x0000, 0xefff).w(FUNC(anes_state::blit_rom_w));
+	map(0xf000, 0xf7ff).ram();
+	map(0xf900, 0xffff).ram();
 }
```
cuavas, line 266: Can you use a single function covering the entire space for `AS_OPCODES` that reads the actual data from the `AS_PROGRAM` space and updates the bank switch status?  See src/mame/osborne/osborne1.cpp (`osborne1_state::opcode_r`) for an example of a Z80 system that does bank switching on M1 cycles.

happppp: This was already discussed above. I'm doing exactly that in the solution I gave to Haze.
The only problem is that the debugger shows the wrong opcode, ideally the Z80 emulation would have an M1 devcb.

cuavas: Using separate spaces is more useful to handle systems that have encryption only applied to M1 fetches, etc.  Using a devcb would also cause a significant performance penalty.  At any rate, using a timer isn’t a good approach for something that’s most likely triggered by M1 fetches.

DavidHaywood:
I think there's enough here for this to be worth including as a progress checkpoint.  It boots, and allows you to coin up / play, so should be a better base for anybody wanting to find a better way to handle the delayed bank

DavidHaywood:
If anybody is more familiar with Mahjong etc. and wants to verify inputs, that would also be appreciated.  I'm sure I had it requiring a 2nd set of controls at some point, which aren't mapped.   I can't find any way to get into a test mode.

happppp:
"a better way to handle the delayed bank", you don't accept the one I posted above?
Honestly, I don't know a way to do it much better than that. The issue with the debugger is not the driver's fault.

src/mame/misc/anes.cpp
```diff
@@ -38,149 +39,597 @@ namespace {
...
 
 void anes_state::prg_map(address_map &map)
 {
-	map(0x0000, 0xefff).rom();
-	map(0xf000, 0xffff).ram();
+	map(0x0000, 0x1fff).rom();
+	map(0x2000, 0xefff).bankr(m_rombank);
+	map(0x0000, 0x01ff).w(FUNC(anes_state::blit_rom_w));
+	map(0xf000, 0xf7ff).ram();
+	map(0xf900, 0xffff).ram();
```
angelosa, line 402: Is there a real reason about this RAM hole?

src/mame/misc/anes.cpp
```diff
@@ -38,149 +39,597 @@ namespace {
...
+uint8_t anes_state::key_r(offs_t offset)
+{
+	for (int i = 0; i < 5; i++)
+		if (!BIT(m_mux, i))
+			return m_key[offset][i]->read();
+
+	//  logerror("%s: Unknown key read with mux = %02X\n", machine().describe_context(), m_mux);
+	//  return 0xff;
+	return offset ? m_coin->read() : 0xff;
+}
```
angelosa, line 292: This is a fairly common mahjong key matrix, not a real multiplexer (i.e. the mux_w fn name is misleading).

DavidHaywood: this was from the code I was sent

src/mame/misc/anes.cpp
DavidHaywood: this is how it was sent to me, I think these are the ranges it tests

DavidHaywood:
> "a better way to handle the delayed bank", you don't accept the one I posted above? Honestly, I don't know a way to do it much better than that. The issue with the debugger is not the driver's fault.

I'm not 100% convinced it's better, but I'll go with that then.

src/mame/misc/anes.cpp
```diff
@@ -38,149 +38,587 @@ namespace {
...
+		int offs = (m_palette_offset + (m_palette_offset_msb << 8)) & 0x1fff;
+		m_palette_data_msb[offs] = data;
+		set_color(offs);
+	}
+}
+
+
+void anes_state::opcodes_map(address_map &map)
+{
+	map(0x0000, 0xefff).r(FUNC(anes_state::m1_rom_r));
```
happppp, line 386: Why limit to 0xefff? That wasn't in the one I gave you.
You see, if it happens to copy a function with bankswitch to RAM, and runs it from there, the bankswitch won't work. That can't be right.

galibert:
The code won't work at all since every first byte of the opcodes will be
zero, cause reading an unmapped zone.

angelosa:
> change to 'matrix style' input (doesn't seem to coin up reliably now though)

That's because only bits 0-5 are connected to the key matrix, coins maps outside of it (and this can't possibly have 5 coin slots).

DavidHaywood:
> That's because only bits 0-5 are connected to the key matrix, coins maps outside of it (and this can't possibly have 5 coin slots).

payout and coin are on the same bit though, so must be multiplexed somehow, or it isn't a key matrix in the first place, but a mux.

src/mame/misc/anes.cpp
```diff
@@ -38,149 +38,587 @@ namespace {
...
-	map(0x13, 0x13).portr("DSW3");
-	map(0x14, 0x15).nopr(); // mux read
-	map(0x16, 0x16).portr("IN0").nopw();
-//  map(0xfe, 0xfe) banking? unknown ROM range
+
+	map(0x10, 0x10).portr("SW1");
+	map(0x11, 0x11).portr("SW2");
+	map(0x12, 0x12).portr("SW3");
+	map(0x13, 0x13).portr("SW4");
+	map(0x14, 0x15).r(FUNC(anes_state::key_r));
```
happppp, line 411: more likely:
map(0x14, 0x14).r(FUNC(anes_state::key_r));
map(0x15, 0x15).portr("COIN");

COIN 0x40 = coin (other bits unknown)

DavidHaywood: again, it's code I was sent, so I don't know why it's like that.

DavidHaywood:
remaining video issues are mainly with how transparencies are handled, as sometimes data is being cleared when it shouldn't be / solid fills aren't being handled properly.

PR #11327, opened 2023-06-11
skeleton/happyvideo.cpp: Add manufacturer name

clawgrip:
464dbc55-08f8-44d4-81c4-16d755117b78


cuavas:
Are you sure the game system is from Leyaoyao?  A look at their web site suggests they make payment terminals and communication equipment, but not actual amusement machines.  Their business is providing module that other companies can integrate.

Are you sure it isn’t just using modules from Leyaoyao for handling payment and bookkeeping?

clawgrip:
The PCB was found inside the machine on the pic I posted, which has this logo:

image

It’s hard to tell for sure, but I’d say the Pcb and the cab went together

cuavas:
Guangzhou Leyaoyao Information Technology (广州乐摇摇信息科技有限公司) seems to have made some of their own claw machines/crane games accepting WeChat and AliPay payment rather than using coins beginning in 2017, but I can’t find any reference for them making kiddy rides.  They were founded in 2015 as Shenzhen Leyaoyao Information Technology, and their business always seemed to be connected payment systems, IoT, etc.  They also don’t seem to have ever used that logo with the moon and two birds.

These kiddy rides are generically referred to as “yaoyaoche” (摇摇车, rocking car) in China.

Annoyingly, there seems to be another unrelated “LEYAOYAO” brand used for consumer electronics products sold on Amazon.

The problem with just putting Leyaoyao as the manufacturer is that people are going to assume it’s the same as Guangzhou Leyaoyao, which will cause confusion when we inevitably get dumps of their products.

cuavas:
The only search results I can find for this thing are dead links on jd.com – it’s worse than Yahoo or eBay for making it impossible to see stuff that’s no longer listed, and the Wayback Machine hasn’t archived any of it.  It’s really hard because Baidu search results are polluted by videos of happy children in kiddy rides, Google somehow thinks I’m looking for porn, and DuckDuckGo just wants to to show me the video for Happy by Pharell Williams.

clawgrip:
Maybe we can use "Leyaoyao Happy Kids" for this machine. Or adding a question mark, like "Leyaoyao Happy Kids?"...

cuavas:
I updated the notes in the file: 89be017b386236bd1a11ae5fc410132c3d165ad3

clawgrip:
Good, thanks!

PR #11330, opened 2023-06-11
heath/h19.cpp Implement Shift-Reset functionality

mgarlanger:
The H19 allowed the user to press the right-shift and reset key to reset the system. Since the h89 shared the common terminal board in the H19, this reset also was routed to the h89 cpu board and reset the entire computer. The actual reset occurs on the release of the keys. 

In mame, the default key for reset is F10. Using the right-shift and F10 now causes the machine (either H19 or H89) to be reset.

A few other general cleanup of comments, formatting, etc. is also done in this PR.

### DIFF
--- a/src/mame/heathkit/h89.cpp
+++ b/src/mame/heathkit/h89.cpp
@@ -123,6 +123,7 @@ private:
 	uint8_t raise_NMI_r();
 	void raise_NMI_w(uint8_t data);
 	void console_intr(uint8_t data);
+	void reset_line(uint8_t data);
 };
 
 /*
@@ -362,8 +363,14 @@ void h89_state::machine_reset()
 
 	update_gpp(0);
 	update_mem_view();
+	m_h37->reset();
+	m_intr_cntrl->reset();
+	m_console->reset();
+	m_serial1->reset();
+	m_serial2->reset();
+	m_serial3->reset();
+	m_maincpu->reset();
 }
-
 
 uint8_t h89_state::raise_NMI_r()
 {
@@ -386,6 +393,14 @@ void h89_state::console_intr(uint8_t data)
 	else
 	{
 		m_intr_cntrl->raise_irq(3);
+	}
+}
+
+void h89_state::reset_line(uint8_t data)
+{
+	if (data == ASSERT_LINE)
+	{
+		reset();
 	}
 }
 
@@ -447,6 +462,8 @@ void h89_state::h89(machine_config & config)
 	// Connect the console port on CPU board to serial port on TLB
 	m_console->out_tx_callback().set(m_tlb, FUNC(heath_tlb_device::cb1_w));
 	m_tlb->serial_data_callback().set(m_console, FUNC(ins8250_uart_device::rx_w));
+
+	m_tlb->reset_cb().set(FUNC(h89_state::reset_line));
 
 	HEATH_Z37_FDC(config, m_h37);
 	m_h37->drq_cb().set(m_intr_cntrl, FUNC(z37_intr_cntrl::set_drq));
--- a/src/mame/heathkit/intr_cntrl.cpp
+++ b/src/mame/heathkit/intr_cntrl.cpp
@@ -31,6 +31,11 @@ void heath_intr_cntrl::device_start()
 	m_intr_lines = 0;
 }
 
+void heath_intr_cntrl::device_reset()
+{
+	m_intr_lines = 0;
+}
+
 void heath_intr_cntrl::device_resolve_objects()
 {
 	m_irq_line.resolve_safe();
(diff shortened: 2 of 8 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -153,26 +157,40 @@ static constexpr uint8_t KB_STATUS_KEYBOARD_STROBE_MASK = 0x80;
...
+	m_strobe = false;
+	m_keyclickactive = false;
+	m_bellactive = false;
+	m_reset_pending = false;
+
+	m_crtc->reset();
+	m_ace->reset();
+	m_beep->reset();
+	m_mm5740->reset();
+	m_maincpu->reset();
```
cuavas, line 192: Once again, you shouldn’t be resetting child devices from `device_reset`.

src/mame/heathkit/tlb.cpp
```diff
@@ -242,16 +277,18 @@ uint8_t heath_tlb_device::kbd_flags_r()
...
 int heath_tlb_device::mm5740_control_r()
 {
-	return ((m_kbspecial->read() ^ 0x10) & 0x10) ? ASSERT_LINE: CLEAR_LINE;
+	return (m_kbspecial->read() & 0x10) ? CLEAR_LINE : ASSERT_LINE;
 }
 
 void heath_tlb_device::mm5740_data_ready_w(int state)
 {
+	check_for_reset();
+
```
cuavas, line 291: Is this when the reset key combination is actually checked, or does it happen out-of-band?  If it happens out-of-band, you should attach callbacks to the relevant keys so you get a notification when they’re changed.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -31,6 +31,11 @@ void heath_intr_cntrl::device_start()
 	m_intr_lines = 0;
 }
 
+void heath_intr_cntrl::device_reset()
+{
+	m_intr_lines = 0;
+}
+
```
cuavas, line 38: This value comes from inputs.  You shouldn’t change it on reset.  If the devices that produce those inputs deassert their outputs on reset, it’s their responsibility to update the outputs on reset.

src/mame/heathkit/h89.cpp
```diff
@@ -362,9 +363,15 @@ void h89_state::machine_reset()
 
 	update_gpp(0);
 	update_mem_view();
+	m_h37->reset();
+	m_intr_cntrl->reset();
+	m_console->reset();
+	m_serial1->reset();
+	m_serial2->reset();
+	m_serial3->reset();
+	m_maincpu->reset();
```
cuavas, line 372: Once again, you shouldn’t be resetting child devices here – that happens automatically.

src/mame/heathkit/h89.cpp
```diff
@@ -389,6 +396,14 @@ void h89_state::console_intr(uint8_t data)
 	}
 }
 
+void h89_state::reset_line(uint8_t data)
+{
+	if (data == ASSERT_LINE)
+	{
+		reset();
+	}
+}
```
cuavas, line 405: Read/write line members conventionally use `int` values.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -171,6 +176,15 @@ void z37_intr_cntrl::device_start()
 	m_fd_irq_raised = false;
 }
 
+void z37_intr_cntrl::device_reset()
+{
+	heath_intr_cntrl::device_reset();
+
+	m_interrupts_blocked = false;
+	m_drq_raised = false;
+	m_fd_irq_raised = false;
```
cuavas, line 185: You shouldn’t be changing inputs on reset.  If the devices connected to those inputs deassert them on reset, it’s their responsibility to push the change.

cuavas (commented):
There’s too much wrong with this.  I’m going to revert it.

happppp:
FYI to do a full system reset:
machine().schedule_soft_reset();

cuavas:
> FYI to do a full system reset: machine().schedule_soft_reset();

But that will also reset things attached to serial ports, etc.  Calling `reset()` on the root device has almost the same effect, although it doesn’t wait to return to the scheduler, which can upset some CPUs.

mgarlanger:
Reverting seems a little excessive when the machine runs fine with the new functionality _and_ there is existing code that does the same things that this PR does wrong in such popular systems as C64/C128- https://github.com/mamedev/mame/blob/master/src/mame/commodore/c64.cpp#L1447 (and there are may other examples of it in the code). 

Anyways, new PR https://github.com/mamedev/mame/pull/11339 addresses the comments raised in this PR.

PR #11333, opened 2023-06-12
Address some limitations of delegates and callbacks

cuavas:
This is mostly a response to comments from @rb6502 and @galibert.  This is split up into three commits to make it a bit easier to see what’s going on:
* The actual functional changes in src/emu (not possible to build as it lacks the corresponding updates to the rest of the code)
* Updating the rest of the code for the changes to devcb.h
* Hacks to make three devices I haven’t sorted out yet build

The big change answers the question, “Why does devcb have `resolve` as well as `resolve_safe`?” by making removing them both altogether.  This entails a number of changes:
* Device callbacks are now automatically resolved at the same time as object finders.
* Read callbacks now need a default return value specified at construction, whether you plan to use it or not.
* The `isnull()` accessor and cast to `bool` operator have been removed.
* There’s a new `isunset()` accessor that tells you if the callback wasn’t configured, even when it’s callable after being resolved.

I can think of arguments against this:
* You no longer have to make a conscious decision on whether you’re going to do things differently depending on whether the callback is configured or not.  Currently, if you use `resolve` you need to check that the callback is configured every time.  With this change you play fast and loose.
* You always need to supply a default return value for read callbacks, even if you’re going to check them with `isunset` every time and do something different if they aren’t configured.

This *does* make the code tidier, though:
* One less thing to remember to do when using device callbacks
* Many devices no longer need to implement `device_resolve_objects`
* Over 6,000 lines net reduction


rb6502:
_Now it would be possible to go a step further with this and remove the need for resolve on callbacks altogether, but that would mean moving the default return value to the constructor arguments rather than setting it in device_start. However, it would reduce total line count and avoid possible confusion between callback and device delegate resolve semantics._

I like this idea a lot.  Combine it with requiring the default value to be explicitly set (I also don't like the default zero) and I think it's a winner.

cuavas:
I’ll separate the device delegate part since that isn’t controversial anyway.  Then I can mess around with getting rid of `resolve` on callbacks.  (Device delegates will still have `resolve`, but callbacks won’t need it.)

galibert:
I don't think there's anything controversial in there, only cool stuff.

One thing I would find nice, if you go and tweak all devdb constructors, is to add a name to them.  Multiple times I would have liked a gtkwave-like display of live devcbs in the debugger (for instance to debug adb) but without naming it's impractical.

cuavas:
Sorry, named devcbs aren’t going to make it this time.  I’ll put it on my TODO list.

galibert:
No worries :-)

rb6502:
I love the named devcbs idea, but it's fine to wait on that too.

PR #11339, opened 2023-06-15
heath/h19.cpp Implement Shift-Reset functionality

mgarlanger:
Attempt 2 of implementing reset functionality for H19/H89 previous pr - https://github.com/mamedev/mame/pull/11330 was reverted. This PR address comments on the reverted one.
Original Info from reverted PR:
```
The H19 allowed the user to press the right-shift and reset key to reset the system. Since the h89 shared the common terminal board in the H19, this reset also was routed to the h89 cpu board and reset the entire computer. The actual reset occurs on the release of the keys.

In mame, the default key for reset is F10. Using the right-shift and F10 now causes the machine (either H19 or H89) to be reset.

A few other general cleanup of comments, formatting, etc. is also done in this PR.
```


src/mame/heathkit/tlb.cpp
```diff
@@ -262,6 +278,34 @@ void heath_tlb_device::mm5740_data_ready_w(int state)
...
+{
+	if (m_reset_key && m_right_shift)
+	{
+		m_reset_pending = true;
+	}
+	else if (m_reset_pending)
+	{
+		m_reset_pending = false;
+		reset();
+		m_reset(ASSERT_LINE);
```
mgarlanger, line 291: Should this be immediately followed by an `m_reset(CLEAR_LINE);`? I've seen that for some other systems.

src/mame/heathkit/z37_fdc.cpp
```diff
@@ -170,6 +170,16 @@ void heath_z37_fdc_device::device_start()
...
 
+void heath_z37_fdc_device::device_reset()
+{
+	m_control_reg = 0;
+	m_interface_reg = 0;
+
+	m_intrq_allowed = false;
+	m_drq_allowed = false;
+	m_access_track_sector = false;
+}
```
cuavas, line 181: If the interrupt and/or DMA request outputs were asserted before the device was reset, they should be deasserted due to `m_intrq_allowed` and `m_drq_allowed` becoming false here, shouldn’t they?  You need to call the callbacks to notify whatever’s connected of the change.

src/mame/heathkit/tlb.cpp
mgarlanger: I looks like it is actually held in reset while the keys are both pressed. When one of the keys is released, the reset line is cleared.

PR #11340, opened 2023-06-15
mt32: Add rom version 2.0.7

bmunger:
This change adds version 2.0.7 of the MT32 rom.  It's a 128k rom that will require banking to work in the future.

### DIFF
--- a/src/mame/roland/roland_mt32.cpp
+++ b/src/mame/roland/roland_mt32.cpp
@@ -418,6 +418,10 @@ ROM_START( mt32 )
 	ROMX_LOAD(       "mt32_2.0.4.ic28",              0,   0x10000, CRC(59a49d5c) SHA1(2c16432b6c73dd2a3947cba950a0f4c19d6180eb), ROM_BIOS(6) )
 	ROM_IGNORE(0x10000)  // banking needs to be implemented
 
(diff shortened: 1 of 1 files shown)


src/mame/roland/roland_mt32.cpp
```diff
@@ -418,6 +418,10 @@ ROM_START( mt32 )
 	ROMX_LOAD(       "mt32_2.0.4.ic28",              0,   0x10000, CRC(59a49d5c) SHA1(2c16432b6c73dd2a3947cba950a0f4c19d6180eb), ROM_BIOS(6) )
 	ROM_IGNORE(0x10000)  // banking needs to be implemented
 
+	ROM_SYSTEM_BIOS( 7, "207", "Firmware 2.0.7" )
+	ROMX_LOAD(       "mt32_2.0.7.ic28",              0,   0x10000, CRC(a016b607) SHA1(47b52adefedaec475c925e54340e37673c11707c), ROM_BIOS(6) )
```
angelosa, line 422: Should be `ROM_BIOS(7)`

angelosa (changes requested):

src/mame/roland/roland_mt32.cpp
bmunger: Good catch!  Missed that on my copy/paste.  Will fix.

PR #11345, opened 2023-06-17
zx81_cass.xml: New working entry

ICEknigh7:
- AMSAT AMS-81 Tracking System (Version N5BF 1.4) [Sarah Rose Giddings, ICEknight, retroLEL]

Also: Added general loading instructions for the ZX81 and other info, made entry `8progs` a bit more complete and correct.

### DIFF
--- a/hash/zx81_cass.xml
+++ b/hash/zx81_cass.xml
@@ -2,35 +2,34 @@
 <!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
 <!--
 license:CC0-1.0
+
+Sinclair ZX81 cassettes
+
+Loading instructions:
+
+At the prompt, press the following sequence to execute the LOAD "" command: J, Shift+P, Shift+P, Enter
+Enable UI controls (Scroll Lock) and start the tape (F2).
+If the program doesn't start automatically after loading, execute the RUN command by pressing: R, Enter
+
+Notes about the dumps:
+
+Due to the way the ZX81 saves to tape, different recordings of identical data (such as those in both sides of a cassette)
+might be stored along with a few different bytes in them. Additional dumps of the same releases should match those differences.
+TZX files for ZX81 are just the .P dumps with an additional header on top, so they may not be the cleanest format for archival.
+
+Known dumps not yet added:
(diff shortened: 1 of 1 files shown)


hash/zx81_cass.xml
```diff
@@ -209,26 +220,38 @@ This list was compiled from the archive at http://www.zx81stuff.org.uk/. Updated
...
+	Oscar10			LOAD "G"
+	TenPlan			LOAD "H"
+]]></notes>
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="56578">
-				<rom name="8Programs.tzx" size="56578" crc="bb788390" sha1="e2185c5962d8c89bb020028a22e25470405bef0f"/>
+				<rom name="8Programs.tzx" size="56578" crc="bb788390" sha1="e2185c5962d8c89bb020028a22e25470405bef0f" />
 			</dataarea>
 		</part>
 	</software>
```
cuavas, line 257: This item doesn’t follow the DTD and is causing a build failure.

ICEknigh7:
Just replaced those two notes with comments, should be good to go.

hash/zx81_cass.xml
```diff
@@ -2,21 +2,32 @@
 <!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
 <!--
 license:CC0-1.0
-
-Sinclair ZX81 Cassettes
-
-This list was compiled from the archive at http://www.zx81stuff.org.uk/. Updated to 22/08/2016
```
Pernod70, line 8: This comment served a purpose, the site is the main source for this softlist and the date tells us which site update the softlist was last synced with.

ICEknigh7: Ok, so I've just created a list of missing dumps which includes all entries published since the latest softlist update, up to the last site update three years ago. I think this is more in line with what's currently being done in other softlists?

cuavas:
@Pernod70 can you please look at this?  You can back out the comment change if you don’t like it, or move it back to the top of the file how it was before.

Pernod70 (approved):

PR #11348, opened 2023-06-18
yamaha/ympsr150.cpp: add DD-9

devinacker:
New working machine
----------
Yamaha DD-9 Digital Percussion [BCM, Devin Acker]

### DIFF
--- a/src/mame/layout/dd9.lay
+++ b/src/mame/layout/dd9.lay
@@ -1,0 +1,166 @@
+<?xml version="1.0"?>
+<!--
+license:CC0-1.0
+-->
+<mamelayout version="2">
+	<!-- general grey backdrop -->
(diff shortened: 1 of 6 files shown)


src/mame/layout/dd9.lay
```diff
@@ -0,0 +1,166 @@
...
+			<bounds left="0" top="0" right="1" bottom="1" />
+			<color red="0" green="0" blue="0" />
+		</rect>
+	</element>
+
+	<element name="white">
+		<rect>
+			<bounds left="0" top="0" right="1" bottom="1" />
+		</rect>
+	</element>
```
cuavas, line 32: There’s no point putting the implicit unit square bounds when you just want to fill the element.

PR #11349, opened 2023-06-18
neocd.xml: New working software list item

FoxhackDN:
Double Dragon (Japan) [redump.org]

### DIFF
--- a/hash/neocd.xml
+++ b/hash/neocd.xml
@@ -444,6 +444,44 @@ license:CC0-1.0
 		</part>
 	</software>
 
+	<software name="doubledra" cloneof="doubledr">
+		<!-- Source: Redump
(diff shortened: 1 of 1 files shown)


FoxhackDN:
This is made from the Japanese version of the game. The current version in MAME appears to be the export version (English packaging) and it appears that that particular pressing is bad.

[This post at Neo-Geo.com](https://www.neo-geo.com/forums/index.php?threads/neo-geo-cd-system-resets.22053/) mentions that some Double Dragon CDs cause system resets after they load 1/3 of the bar at startup, which is similar to what happens in MAME. I grabbed the two existing Redump sets: One (dumped and verified by two different dumpers, with a track 01 CRC of `cf083610`, serial `NGCD-082E`) causes the same crashes as `doubledr`; the other dump (track 01 CRC of `3bddb81a`, serial `NGCD-082`, verified by three different dumpers) works perfectly fine. So `doubledr` is a good dump, but a bad pressing.

PR #11352, opened 2023-06-18
heathkit/h19.cpp implement repeat key, fix intr handling

mgarlanger:
Implements the repeat key functionality for the H19, and updates MM5740 to support the repeat input line.
Updates the interrupt handling to properly control the line based on OR-ing the keyboard and serial 8250 interrupt inputs.

### DIFF
--- a/src/devices/machine/mm5740.cpp
+++ b/src/devices/machine/mm5740.cpp
@@ -43,7 +43,7 @@ const tiny_rom_entry *mm5740_device::device_rom_region() const
 //  mm5740_device - constructor
 //-------------------------------------------------
 
-mm5740_device::mm5740_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+mm5740_device::mm5740_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock) :
 	device_t(mconfig, MM5740, tag, owner, clock),
 	m_read_x(*this, 0x3ff),
 	m_read_shift(*this, 0),
@@ -88,16 +88,6 @@ void mm5740_device::device_start()
 }
 
 //-------------------------------------------------
-//  device_start - device-specific reset
-//-------------------------------------------------
-
-void mm5740_device::device_reset()
-{
-	m_repeat = 0;
-	m_trigger_repeat = false;
-}
-
-//-------------------------------------------------
 //  perform_scan - scan the keyboard matrix
 //-------------------------------------------------
 
@@ -108,6 +98,17 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 	for (int x = 0; x < 9; x++)
 	{
 		u16 data = m_read_x[x]() ^ 0x3ff;
+
+		if (data)
+		{
+			ako = true;
+		}
+
+		if ((data ^ m_x_mask[x]) == 0)
+		{
+			// bail early if nothing has changed.
+			continue;
+		}
 
 		for (int y = 0; y < 10; y++)
 		{
@@ -125,8 +126,6 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 
 				u16 b = (((common & 0x10) << 4) | ((uniq & 0x0f) << 4) | (common & 0x0f)) ^ 0x1ff;
 
-				ako = true;
-
 				// Check for a new keypress
 				if (!BIT(m_x_mask[x], y))
 				{
@@ -139,17 +138,6 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 						return;
 					}
 				}
-				else
-				{
-					// check for repeat functionality
-					if (m_trigger_repeat)
-					{
-						m_write_data_ready(ASSERT_LINE);
-						m_trigger_repeat = false;
-
-						return;
-					}
-				}
 			}
 			else
 			{
@@ -157,6 +145,12 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 				m_x_mask[x] &= ~(1 << y);
 			}
 		}
+	}
+
+	if (m_trigger_repeat && (m_b != -1))
+	{
+		m_write_data_ready(ASSERT_LINE);
+		m_trigger_repeat = false;
 	}
 
 	if (!ako)
--- a/src/devices/machine/mm5740.cpp
+++ b/src/devices/machine/mm5740.cpp
@@ -73,8 +73,8 @@ void mm5740_device::device_start()
 {
 	std::fill(std::begin(m_x_mask), std::end(m_x_mask), 0);
 	m_b = -1;
-	m_repeat = 0;
-	m_trigger_repeat = false;
+	m_repeat = false;
+	m_last_repeat = false;
 
 	// allocate timers
 	m_scan_timer = timer_alloc(FUNC(mm5740_device::perform_scan), this);
@@ -84,7 +84,7 @@ void mm5740_device::device_start()
 	save_item(NAME(m_b));
 	save_item(NAME(m_x_mask));
 	save_item(NAME(m_repeat));
-	save_item(NAME(m_trigger_repeat));
+	save_item(NAME(m_last_repeat));
 }
 
 //-------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/devices/machine/mm5740.cpp
```diff
@@ -51,13 +51,12 @@ mm5740_device::mm5740_device(const machine_config &mconfig, const char *tag, dev
...
 {
-	std::fill(std::begin(m_x_mask), std::end(m_x_mask), 0);
 }
 
-uint32_t mm5740_device::calc_effective_clock_key_debounce(uint32_t capacitance)
+u32 mm5740_device::calc_effective_clock_key_debounce(u32 capacitance)
 {
 	// calculate key debounce based on capacitance in pF
-	uint32_t key_debounce_msec = capacitance / 125;
+	u32 key_debounce_msec = capacitance / 125;
```
cuavas, line 59: The rest of this file uses the POSIX-style `uint32_t`, `uint16_t`, etc.  Please don’t change it to mix POSIX-style and abbreviated style.

src/devices/machine/mm5740.cpp
```diff
@@ -72,22 +71,30 @@ uint32_t mm5740_device::calc_effective_clock_key_debounce(uint32_t capacitance)
...
-
 //-------------------------------------------------
 //  device_start - device-specific reset
 //-------------------------------------------------
 
 void mm5740_device::device_reset()
 {
+	m_repeat = 0;
+	m_trigger_repeat = false;
 }
```
cuavas, line 98: The MM5740 has no reset input as far as I can tell.  It shouldn’t do this here.

src/devices/machine/mm5740.cpp
```diff
@@ -100,32 +107,27 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
...
-		uint16_t data = m_read_x[x]() ^ 0x3ff;
-
-		if ((data ^ m_x_mask[x]) == 0)
-		{
-			// bail early if nothing has changed.
-			continue;
-		}
+		u16 data = m_read_x[x]() ^ 0x3ff;
 
 		for (int y = 0; y < 10; y++)
```
cuavas, line 112: The changes to the logic here are dubious.  You should still skip the scan loop if the data hasn’t changed.  The encoder just generates additional data strobe pulses when it detects positive edges on the repeat clock while the data output is valid.

Also, the repeat input is only sampled on falling edges of the clock, and there’s a two cycle latency from detecting a rising edge on the repeat input to generating a data strobe pulse (see fifth page of the datasheet).

cuavas (changes requested):

src/devices/machine/mm5740.cpp
mgarlanger: There was only one remaining posix style in this file, the constructor, I'll update that to be consistent with the rest of the file.

src/devices/machine/mm5740.cpp
mgarlanger: I saw that there was a 2 cycle latency, but due to the `calc_effective_clock_key_debounce()` calculation, the chip is only operating at 25 Hz. So it would be a much bigger effort to accurately model the 2 cycle delay.

mgarlanger: Based on testing with the current the capacitance at 5000 pf (as shown in the schematics for terminal board), which provides the effective clock of 25 Hz, seems to behave differently than a real H19/H89. When playing some games, it was missing fast key presses, something that I've never noticed on a real system.  The datasheet for the MM5740, shows that it is linear between 0 pf -> 500 pf, going from 0 mSec -> 4mSec. I'm wondering if it is still linear all the way to 5000 pf. 25Hz seems _really_ low slow for key handling, and seems questionable that would be what it was designed as.

src/devices/machine/mm5740.cpp
```diff
@@ -138,26 +139,39 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
...
 		m_b = -1;
 	}
 }
 
+void mm5740_device::repeat_line_w(int state)
+{
+	bool repeat = (state == ASSERT_LINE);
+	m_trigger_repeat = repeat && (repeat != m_repeat);
+	m_repeat = repeat;
+}
```
cuavas, line 168: This input isn’t edge-sensitive.  The repeat input is tested on falling edges of the clock input.  You should just be storing the value here.

In `perform_scan`, which is called at the input clock frequency, you should test whether the current repeat input is high after seeing it low on the previous call to `perform_scan` and trigger a an additional data strobe (calling `m_write_data_ready`) if appropriate.

src/devices/machine/mm5740.cpp
```diff
@@ -138,26 +139,39 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 					}
 				}
 			}
-			else    // key released, unmark it from the "down" info
+			else
 			{
+				// key released, unmark it from the "down" info
 				m_x_mask[x] &= ~(1 << y);
 			}
```
cuavas, line 146: You need to test here whether the released key is the most recently detected key-down, and if it is, you should stop generating additional repeat events.

src/devices/machine/mm5740.cpp
```diff
@@ -100,7 +97,12 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
 
 	for (int x = 0; x < 9; x++)
 	{
-		uint16_t data = m_read_x[x]() ^ 0x3ff;
+		u16 data = m_read_x[x]() ^ 0x3ff;
```
cuavas, line 100: Pre-existing issue, but this shouldn’t be scanning all X lines on every active clock edge, it should only be scanning one line on each active clock edge.

mgarlanger: If we need to only do one X line scan per clock, this class will need to be refactored to handle debounce differently. Currently it's using the debounce capacitor value to determine the key bounce mask time, based on the H19 schematics, the capacitor is 5000 pF, and based on a linear slope shown in the datasheet, this would put the bounce mask time at 40 mSec. The effective clock for the chip is just 25 Hz, as I've mentioned in another comment on this PR - https://github.com/mamedev/mame/pull/11352#discussion_r1248874602. If we only scanned one line on each clock, that would be less than 3 scans per line per second.

cuavas: Due to the way MAME works, the emuated inputs are only updated once per emulated video frame anyway (every 16 ms at 60 Hz).  It drives up input latency when debounce is emulated on top of that (as is the case on a number of .  Is the debounce emulation even implemented realistically here?

Anyway, probably something to worry about another time.

src/devices/machine/mm5740.cpp
```diff
@@ -110,22 +112,21 @@ TIMER_CALLBACK_MEMBER(mm5740_device::perform_scan)
...
-
-				ako = true;
+			u16 b = (((common & 0x10) << 4) | ((uniq & 0x0f) << 4) | (common & 0x0f)) ^ 0x1ff;
 
+			if (BIT(data, y))
+			{
+				// Check for a new keypress
 				if (!BIT(m_x_mask[x], y))
 				{
 					m_x_mask[x] |= (1 << y);
```
cuavas, line 132: Only the calculation of `offset` is needed when a key is released, so the rest of these locals can stay inside the inner `if` block.

mgarlanger: But to compare offset with the offset used for the key down, I'll need a new class variable to store offset as calculated prior to it being updated with the shift/ctrl offset and the `* 90`. And it only should be saved when it passes all the if checks and save it at the same spot as the `m_b = b;` statement. Is that correct?

mgarlanger:
>After looking at this some more, the debounce emulation, scanning, and relative timing of data and strobe outputs are all wrong, but fixing any of that will likely require a pile of downstream changes in drivers to avoid regressions.

Only heathkit is using the mm5740 device, so if you have time to explain what changes are needed to fix these issues, I can look at addressing them in a future PR. BTW, this was original based on kb3600.cpp, so any issues may also apply to that one too.

cuavas (commented):
I’ll merge this now, and try to sort out the issues with the MM5740 and the device whose implementation it’s based on later.  It’s not going to hurt anything in this this state, anyway.

PR #11356, opened 2023-06-19
astropc.cpp: Add notes, correct release years

einstein95:
Also added version info for the dslayrr sets from the files


Cristiano-MDQ:
I have an image from a flash memory card, from the Arabian Nights game... the image is in vhd extension, the image was made using Winmage. If necessary, provide me with an email where I can send the files.

PR #11357, opened 2023-06-20
Dumped and added a skeleton driver for a Spanish electromechanical air hockey

clawgrip:
New NOT_WORKING machine
Caramball [Retromaniacs, Recreativas.org]

### DIFF
--- a/src/mame/skeleton/caramball.cpp
+++ b/src/mame/skeleton/caramball.cpp
@@ -94,7 +94,7 @@ ROM_START(caramball)
 	ROM_LOAD("crmbal_4.0.i7", 0x0000, 0x8000, CRC(6ff5eca9) SHA1(8a926509c6143aced4d2d3a024965d8e33ef1074))
 
 	ROM_REGION(0x800, "nvram", 0)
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/caramball.cpp
```diff
@@ -0,0 +1,106 @@
...
+ROM_START(caramball)
+	ROM_REGION(0x8000, "maincpu", 0)
+	ROM_LOAD("crmbal_4.0.i7", 0x0000, 0x8000, CRC(6ff5eca9) SHA1(8a926509c6143aced4d2d3a024965d8e33ef1074))
+
+	ROM_REGION(0x800, "nvram", 0)
+	ROM_LOAD( "mk48z02b-20.i8", 0x000, 0x800, CRC(8125cb9d) SHA1(ac76f81edb092e8f16fae577a0702a1ec4ab4503) )
+
+	ROM_REGION(0x117, "plds", 0)
+	ROM_LOAD("palce16v8h.i9", 0x000, 0x117, NO_DUMP) // Protected and registered
+ROM_END
```
cuavas, line 101: Spacing isn’t consistent – you’ve got spaces inside the parentheses on the second region, but not the first or third region.

clawgrip: You're right. Thx, fixed!

PR #11362, opened 2023-06-22
pda600: iterators are not required to be pointers

pmackinlay:


src/mame/amstrad/pda600_copro.cpp
```diff
@@ -445,7 +445,7 @@ void pda600_copro_device::exec_beep()
 	m_resp_type = m_resp_data = 0;  // Empty response
 	m_state = STATE_PLAY_TONE;
 	m_buf_size -= 4;
-	std::memmove(m_buf.begin(), m_buf.begin() + 3, m_buf_size);
+	std::memmove(&m_buf[0], &m_buf[3], m_buf_size);
 	m_update_timer->adjust(attotime::zero);
```
cuavas, line 449: This seems a little odd – it drops 4 from the size of data in the buffer, but shifts the content by 3.  @sronco am I missing something here?

sronco: This code removes the 3 byte header and 1 byte checksum from the frame, leaving only the tone data which is later used by a timer.

happppp: Can m_buf_size be 3 when entering this function?
I noticed this check:
else if (m_buf_size == m_buf[1] + 3)
So, if m_buf[1] happens to be 0, it would try to memmove 255 bytes?

sronco: Sorry for the delay, but I have been on vacation for two weeks with very limited Internet access.

And yes, you are right, this can occur with a malformed frame. I'll add a check to avoid this.

PR #11363, opened 2023-06-22
new NOT WORKING machines

DavidHaywood:
new NOT WORKING machines
-------
DT7 (prototype) [Graham Noble]

as per @galibert  "Let's close it for now, we can revisit afterwards when things are clearer. Let's say next summer for instance?"

It is now officially summer, so I'll leave this clean PR here as the old PR no longer compiled, and the branch had already been deleted.

I won't be doing any further work on this.  Credit goes to the person who owned the PCB.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -43843,6 +43843,9 @@ gizmondo                        //
 kontiki                         //
 tiki100                         //
 
+@source:toaplan/dt7.cpp
+dt7                             // (c) 1993 Toaplan
+
 @source:toaplan/mjsister.cpp
 mjsister                        // (c) 1986 Toaplan
 
--- a/src/mame/toaplan/dt7.cpp
+++ b/src/mame/toaplan/dt7.cpp
@@ -1,0 +1,661 @@
+// license:BSD-3-Clause
+// copyright-holders:David Haywood
+
+/* This is derived from toaplan2.cpp, but there are enough hardware
+   differences to keep it separate
+
+   TODO:
+    - verify audio CPU opcodes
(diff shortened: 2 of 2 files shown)


happppp:
This driver is not worth the drama it will cause.
Maybe he meant the summer of 2033 ;p

rb6502:
Is this in FBN yet?  If not, yeah, we don't wanna be the ones sticking our heads up.

DavidHaywood:
it's in HBMAME/Arcade

Robbbert:
The code is in HBMAME, but the game is not enabled. It isn't in ARCADE64 (if that's what you meant) at all.

DavidHaywood:
Ok, it's the code is being distributed in HBMAME then, excuse me for not knowing what series of illogical decisions led to a non-homebrew game being in HBMAME rather than Arcade.  I tend to group them together.

arcadez2003:
> Is this in FBN yet? If not, yeah, we don't wanna be the ones sticking our heads up.



> Is this in FBN yet? If not, yeah, we don't wanna be the ones sticking our heads up.

Someone i know was working on it a while back maybe it'll turn up soon in FBN maybe it wont, but at the end of the day 
it's not one of Toaplan's best efforts their earlier Rally bike title was a far better game.

arcadez2003:
> This driver is not worth the drama it will cause. Maybe he meant the summer of 2033 ;p

If our forefathers had thought like that eg skip everything considered to be problematic that would have put a serious 
dent on our advancement as a species.

Then again we seem to be going back the way so much so i pine for the Scotland i was born into everything was simpler 
and IMO much better back then but hey ho i digress :)

PR #11369, opened 2023-06-24
hash/a800.xml: add aDawilah releases

angelosa:
a800.xml: Akhtabar Dhakayik, Altaaqa, Bank Almaelumat, Qamus Najm, Kayf Yaemal Alkumbiutar [archive.org]

### DIFF
--- a/hash/a800.xml
+++ b/hash/a800.xml
@@ -7288,4 +7288,95 @@ Garbage row strip at bottom of screen
 			</dataarea>
 		</part>
 	</software>
+
+	<!-- aDawilah releases, QA presumes using a65xea main set -->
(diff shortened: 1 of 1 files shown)


angelosa:
- [archive.org] being the closest possible match for these I can find, they originally came from a "Atari8bitPreservedSoftware" package which is about the same level of distribution (i.e. they just collected it rather than being real dumpers).

cuavas:
This looks pretty much OK.  The only question would be whether we want to hyphenate the “al” article prefix, e.g. “Al-Taaqa”, “Al-Kumbiutar”, etc. or treat it as part of the word as you’re doing now.  Probably not worth holding up merging though.

PR #11370, opened 2023-06-24
edevices/stlforce.cpp: Fix sound clock and add placeholder for PLDs for Mortal Race

clawgrip:


clawgrip:
image

src/mame/edevices/stlforce.cpp
```diff
@@ -310,7 +309,7 @@ void stlforce_state::stlforce(machine_config &config)
 	// sound hardware
 	SPEAKER(config, "mono").front_center();
 
-	OKIM6295(config, "oki", XTAL(32'000'000) / 32, okim6295_device::PIN7_HIGH).add_route(ALL_OUTPUTS, "mono", 1.0);
+	OKIM6295(config, "oki", 32_MHz_XTAL / 32, okim6295_device::PIN7_HIGH).add_route(ALL_OUTPUTS, "mono", 1.0);
```
cuavas, line 312: If it’s really a 32.220 MHz oscillator module as suggested by the photograph, you should add that to emu/xtal.cpp if necessary and use the correct frequency here.

clawgrip: But, if you see the latests Dumping Union messages, there’s a 32MHz xtal on Steel Force…

What I can tell for sure is the one on Mortal Race (32.220), as I still keep the PCB.

Also, about Steel Force, PCB pics on eBay and Recreativas.org shows what seems to be a 32.275 xtal.

I’m starting to think that they used the cheapest component with a similar value the found.

image

PR #11375, opened 2023-06-25
machine/wd_fdc.cpp: Simplify/optimize how bus inverting is handled

mgarlanger:
Simplifies how bus inverting is handled, no longer has an if check, but always perform an xor with the value defined in the class. For chips that do inverting, the value will be set to 0xff. For chips without inverting, it is set to 0x00.

### DIFF
--- a/src/devices/machine/wd_fdc.h
+++ b/src/devices/machine/wd_fdc.h
@@ -99,7 +99,7 @@ protected:
 	bool disable_mfm;
 	bool enmf;
 	bool has_enmf;
-	bool inverted_bus;
+	uint8_t bus_invert_value;
 	bool side_control;
 	bool side_compare;
 	bool head_control;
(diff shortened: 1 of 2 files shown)


src/devices/machine/wd_fdc.h
```diff
@@ -99,7 +99,7 @@ class wd_fdc_device_base : public device_t {
 	bool disable_mfm;
 	bool enmf;
 	bool has_enmf;
-	bool inverted_bus;
+	uint8_t bus_invert_value;
```
mgarlanger, line 102: I wasn't really happy with the name of the variable, if someone has a better idea, let me know.

cuavas:
@galibert do you have an opinion on this?  It looks correct, and it’s a net reduction in source code lines.  I wouldn’t count on it improving performance – a test/conditional move is likely just as fast as loading the value from a data member.

galibert:
Yeah, I don't see that as having any performance impact, whatever :-)

PR #11378, opened 2023-06-27
dcinva instruction in the i960 disassembler is off by one

DrItanium:
The mnemonic table for the primary opcodes of the i960 instruction set (I consider REG format instructions to be composed of a primary and secondary opcode) was declaring dcinva as 0xAC when the i960 Hx Microprocessor Developer's manual states it is 0xAD (see page 6-45).

As an aside, I put a todo for the future to compose the tables with something like X-macros (or something like it) to make subtle errors like this one a thing of the past.


cuavas:
The TODO comment seems unwarranted as we don’t seem to have a major problem with offsets.  Even just tabulating the opcode definitions would make it more readable.

DrItanium:
My tendency to have a single knowledge source for things like opcodes are
showing through. I'll take the comment out. What do you mean by tabulation?
Do you want me to reformat the table?

PR #11379, opened 2023-06-27
i960 movre instruction will result in out of bounds read/write on platforms where double is 8-bytes in size.

DrItanium:
Given the following instruction sequence:

movre g0, fp0 

The old implementation used a double as the underlying storage for fp0:4. On most platforms this is going to be 8 bytes but an extended real is 10 bytes in length. This will result in data corruption when this instruction is used. Reading or writing to fp3 will result in the next 32-bit number being pulled in. 

This pull request fixes this by making sure that each floating point register is _at least_ 96-bits wide. To do this, a union is introduced which is comprised of a long double and a array of three Ordinals. On platforms which treat double and long double the same (most non x86 platforms) then it will treated as a double. On x86 based platforms, the underlying type will be a proper long double in the same format as real hardware. 

With the use of the union, it is now safe to just use the pointer decay of the array for the transfer. Regardless of platform, this will just work now!

NOTE: you will also see some @todo statements around checking to see what a real i960 will do with reserved floating point literals. I have not done the test because my own i960 single board computer is currently non functioning. I will see what the chip does if you give it bogus floating point literal values once I get it back up and functioning.

### DIFF
--- a/src/devices/cpu/i960/i960.cpp
+++ b/src/devices/cpu/i960/i960.cpp
@@ -2370,24 +2370,32 @@ std::unique_ptr<util::disasm_interface> i960_cpu_device::create_disassembler()
 	return std::make_unique<i960_disassembler>();
 }
 
-void i960_cpu_device::movre() 
+void i960_cpu_device::movre(uint32_t opcode) 
 {
     uint32_t *src=nullptr, *dst=nullptr;
 
+    /// @todo figure out why we are subtracting not adding
     m_icount -= 8;
 
     if(!(opcode & 0x00000800)) {
-        src = (uint32_t *)&m_r[opcode & 0x1e];
+        /// @todo I don't believe that the i960SB does auto alignment of
+        /// registers to quad register boundaries. Check and see with real
+        /// hardware if that is the case.
+        src = static_cast<uint32_t*>(&m_r[opcode & 0x1e]);
     } else {
-        int idx = opcode & 0x1f;
-        if(idx < 4)
-            src = (uint32_t *)&m_fp[idx];
+        if(auto idx = opcode & 0x1f; idx < 4) {
+            src = static_cast<uint32_t*>(&m_fp[idx]);
+        }
     }
 
-    if(!(opcode & 0x00002000)) {
-        dst = (uint32_t *)&m_r[(opcode>>19) & 0x1e];
-    } else if(!(opcode & 0x00e00000))
-        dst = (uint32_t *)&m_fp[(opcode>>19) & 3];
+    if(auto srcDestIndex = opcode >> 19 & 0x1f; !(opcode & 0x00002000)) {
+        /// @todo I don't believe that the i960SB does auto alignment of
+        /// registers to quad register boundaries. Check and see with real
+        /// hardware if that is the case.
+        dst = static_cast<uint32_t*>(&m_r[srcDestIndex & 0b11100]);
+    } else if(!(opcode & 0x00e00000)) {
+        dst = static_cast<uint32_t*>(&m_fp[srcDestIndex & 0b00011]);
+    }
 
     dst[0] = src[0];
     dst[1] = src[1];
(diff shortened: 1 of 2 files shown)


cuavas (changes requested):
If this needs to implement 80-bit floating point values, it needs to use SoftFloat to do it properly, not bake its own implementation.

Also, this doesn’t follow the style of the rest of the codebase and has general issues:
* Added lines with spaces for initial indent.
* Using `///` Doxygen comment delimiter for things that aren’t documentation comments.
* Polluting the global namespace.
* It doesn’t use MAME conventions for identifier formatting.
* Your claim that `long double` aliases to `double` on non-x86 is very much wrong.  It’s IEEE 128 bit on PPC64LE and SPARC, and it’s a non-standard 128-bit extended precision format on PowerPC.

DrItanium:
Can't fix this right now as I'm at work but I have some questions for you:

1. If I were to wrap the types in a namespace would that be sufficient to
not pollute the global namespace?

2. Where can I get the style guidelines for MAME?

Interesting, on what compiler families (GCC, clang, etc) is the float128
for ppc64le and sparc aliased to long double? When we bring up support for
a compiler at work nearly all targets except for x86 based ones are long
double is an alias to double. Regardless, I was originally planning to use
type_traits to be more nuanced on selecting the underlying extended real
type. I'll go back to that.

Thanks for the feedback.

DrItanium:
Just saw the comment about SoftFloat. If you have a consistent 80-bit Intel
format floating point implementation in software then that should be used.
The Numerics Architecture of the i960 is very nuanced and having SoftFloat
for fp80 is necessary.

I will dive in and update the implementation to use the SoftFloat type.

Thanks for the feedback

rb6502:
One additional note - we aren't setting a great example on it yet, but please use SoftFloat 3.  Both 2 and 3 are available in MAME but we're phasing out version 2 because it has a license ambiguity (and 3 is better, of course).

DrItanium:
Thanks for the heads up. I'll make sure to use SoftFloat3 when I can get
to making the fixes.

DrItanium:
So I believe I have fixed nearly all of the issues stated by @cuavas except one.

I was unable to take advantage of the 80 bit extended floating point support from softfloat3 because it is missing implementations of sin/cos/tan/atan etc. The current i960 cpu implementation depends on the standard library implementation of these functions and I am _not_ the person that could reimplement them with softfloat3 (I am terrible at trigonometry). So I designed the ExtendedReal union to still hold onto a double (to minimize the number of breaking changes) but allocate enough space to prevent the previous buffer overrun. 

Once someone else has implemented these missing floating point routines using the softfloat3 library then it will be relatively trivial to migrate over to using it (just change the using declaration and update the parts of the code which were using the native double operations). I did have another idea how to potentially solve this problem now but it is not guaranteed to be portable without some amount of testing on a big endian machine.

Let me know if I missed anything.

MooglyGuy:
Your member and type naming is still problematic. Please refer to the following guideline document, specifically the "Naming conventions" section: https://docs.mamedev.org/contributing/cxx.html

You're using LlamaCase for type names, when snake_case is strongly preferred. Similarly, you're using camelCase for member names, when again, snake_case is strongly preferred. More broadly, guidelines aside, changes to an existing file should mimic the style in that file as closely as possible.

MooglyGuy:
Some of your comments are still using Doxygen-style `///` specifiers for things that aren't documentation, and are instead questions that can be answered:

```/// @todo figure out why we are subtracting not adding```

We subtract from `m_icount` because that's how the CPU cores work in MAME. At the beginning of a timeslice, `m_icount` is given a number of instruction cycles to execute. Instructions deduct from that count, and `exceute_run` will exit when that count is at or below zero. If it under-runs to a negative value, those additional cycles can be accounted for by the core scheduler.

DrItanium:
thank you for the coding guidelines, I assumed that they didn't exist since I wasn't given the link. Appreciate it :). 

Thank you also for answering my question about the subtraction. 

I will have the fixes up shortly.

PR #11382, opened 2023-06-28
bus/a2bus: Add Uthernet II support Wiznet w5100

ksherlock:
a2RetroSystems Uthernet II / Wiznet W5100/W5100S emulation

Wiznet W5100

Used in: Uthernet II (Apple II), Spectranet (ZX Spectrum)

Based on:

W5100, W5100S, W5200, W5500 datasheets
WIZnet ioLibrary Driver (https://github.com/Wiznet/ioLibrary_Driver) https://docs.wiznet.io/Product/iEthernet/W5100/
Uthernet II User's and Programmer's Manual
Testing with Uthernet II and W5100S-EVB-Pico

High-level UDP/TCP/ICMP emulation is included, which attempts to match the quirks.

With the Uthernet II it's often only used as a raw ethernet device (ip65, marinetti, a2osx) but there is some software that uses the TCP stack, eg [A2stream](https://github.com/oliverschmidt/A2Stream)

    mame apple2ee -sl3 uthernet2  -flop1 A2Stream.dsk

### DIFF
--- a/src/devices/machine/w5100.cpp
+++ b/src/devices/machine/w5100.cpp
@@ -38,6 +38,7 @@
 #include "machine/w5100.h"
 #include "machine/w5100_socket.h"
 #include "util/internet_checksum.h"
+#include "multibyte.h"
 
 // #define LOG_GENERAL (1U << 0)
 #define LOG_PACKETS (1U << 1)
@@ -246,25 +247,6 @@ static const uint8_t ETHERNET_MULTICAST[6] = { 0x01, 0x00, 0x5e, 0x00, 0x00, 0x0
 	return buffer;
 }
 
-
-static inline uint32_t be_read(const uint8_t *base, int numbytes)
-{
-	uint32_t result = 0;
-	while (numbytes--)
-		result = (result << 8) | *base++;
-	return result;
-}
-
-static inline void be_write(uint8_t *base, uint32_t value, int numbytes)
-{
-	base += numbytes;
-	while (numbytes--)
-	{
-		*--base = value;
-		value >>= 8;
-	}
-}
-
 static wiznet_eth_info parse_eth(const uint8_t *buffer, int length)
 {
 	wiznet_eth_info info{};
@@ -273,7 +255,7 @@ static wiznet_eth_info parse_eth(const uint8_t *buffer, int length)
 	{
 		memcpy(info.dest_mac, buffer + o_ETHERNET_DEST, 6);
 		memcpy(info.src_mac, buffer + o_ETHERNET_SRC, 6);
-		info.type = be_read(buffer + o_ETHERNET_TYPE, 2);
+		info.type = get_u16be(buffer + o_ETHERNET_TYPE);
 	}
 	return info;
 }
@@ -284,9 +266,9 @@ static wiznet_udp_info parse_udp(const uint8_t *buffer, int length)
 
 	if (length >= 8)
 	{
-		info.udp_length = be_read(buffer + o_UDP_LENGTH, 2);
-		info.src_port = be_read(buffer + o_UDP_SRC_PORT, 2);
-		info.dest_port = be_read(buffer + o_UDP_DEST_PORT, 2);
+		info.udp_length = get_u16be(buffer + o_UDP_LENGTH);
+		info.src_port = get_u16be(buffer + o_UDP_SRC_PORT);
+		info.dest_port = get_u16be(buffer + o_UDP_DEST_PORT);
 	}
 	return info;
 }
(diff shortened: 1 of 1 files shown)


src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+ Timeouts:
+ RTR- retry timeout register - is in units of 100us. Default value (2000) is 200ms
+ Sn_MR_ND - No Delayed Ack (TCP only) -
+ Sn_IR_TIMEOUT 0
+
+*/
+
+
+static const int MAX_FRAME_SIZE = 1514;
+static uint8_t ETHERNET_BROADCAST[6] = { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff };
```
cuavas, line 523: A global non-constant array is usually a bad sign.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+	save_item(m_tx_buffer, "TX Buffer Memory");
+	save_item(m_rx_buffer, "RX Buffer Memory");
+
+	save_item(NAME(m_idm));
+	save_item(NAME(m_identification));
+	save_item(NAME(m_irq_state));
+
+	save_pointer(STRUCT_MEMBER(m_tcp, snd_una), 4);
+	save_pointer(STRUCT_MEMBER(m_tcp, snd_nxt), 4);
+	save_pointer(STRUCT_MEMBER(m_tcp, rcv_nxt), 4);
```
cuavas, line 856: I can’t see how using `save_pointer` on structures like this is going to work properly.  It’s for saving indeterminate length arrays.  The count is the number of elements in the array.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+	for (int sn = 0; sn < 4 ; ++sn)
+	{
+		emu_timer *t;
+		t = timer_alloc(FUNC(w5100_base_device::delayed_ack_timer), this);
+		t->set_param(sn);
+		m_delayed_ack_timers[sn] = t;
+
+		t = timer_alloc(FUNC(w5100_base_device::keep_alive_timer), this);
+		t->set_param(sn);
+		m_keep_alive_timers[sn] = t;
```
cuavas, line 880: What’s the purpose of the local `t`?  It just makes the code more error-prone because you’ve got more local mutable state.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+
+				if ((m_common_registers[0] & MR_AI) && !machine().side_effects_disabled())
+				{
+					m_idm++;
+					if (m_device_type == dev_type::W5100 && (m_idm & 0x1fff) == 0x0000)
+						m_idm -= 0x2000;
+				}
+				break;
+		}
+	}
```
cuavas, line 1355: `break` following `return` doesn’t make sense.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+				offset = m_idm;
+				if (m_common_registers[0] & MR_AI)
+				{
+					m_idm++;
+					if (m_device_type == dev_type::W5100 && (m_idm & 0x1fff) == 0x0000)
+						m_idm -= 0x2000;
+				}
+				break;
+		}
+	}
```
cuavas, line 1390: `break` following `return` doen’t make sense.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+
+	offset &= 0x7fff;
+
+	return m_address_space->read_byte(offset);
```
cuavas, line 1359: Using an `address_space` directly is inefficient – you usually want to use an `address_space_specific` or `address_space_cache`.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+		case Sn_PORT0:
+		case Sn_PORT1:
+		case Sn_PROTO:
+		case Sn_TOS:
+		case Sn_TTL:
+		case Sn_FRAGR0:
+		case Sn_FRAGR1:
+			/* these update immediately */
+			socket[offset] = data;
+			break;
```
cuavas, line 1445: If you’re using big switch statements inside your address space handlers, it usually means you have a problem.  You should be getting the memory system to do the heavy lifting for you and using on handler for each logical function.

src/devices/machine/w5100.cpp
```diff
@@ -0,0 +1,4320 @@
...
+		m_address_space->install_read_handler(0x0400, 0x07ff, 0, 0x3800, 0, read8sm_delegate(*this, FUNC(w5100_base_device::read_socket_register)));
+		m_address_space->install_write_handler(0x0400, 0x07ff, 0, 0x3800, 0, write8sm_delegate(*this, FUNC(w5100_base_device::write_socket_register)));
+	}
+	else
+	{
+		// on the w5100, socket 3 is mirrored.
+		m_address_space->install_read_handler(0x0400, 0x04ff, 0x00ff, 0x3800, 0, read8sm_delegate(*this, FUNC(w5100_base_device::read_socket_register_3)));
+		m_address_space->install_write_handler(0x0400, 0x04ff, 0x00ff, 0x3800, 0, write8sm_delegate(*this, FUNC(w5100_base_device::write_socket_register_3)));
+
+		m_address_space->install_read_handler(0x0400, 0x07ff, 0, 0x0000, 0, read8sm_delegate(*this, FUNC(w5100_base_device::read_socket_register)));
```
cuavas, line 905: Why aren’t you supplying an address map function when configuring the address space as this is all static and installed on start.

cuavas (changes requested):
Look, I haven’t gone through all the code yet, but there are already a bunch of concerning things, and it‘s going to need a fairly major rework which will result in it needing to be reviewed from scratch anyway.

Before we go any further, please refactor the code to at least address these points:
* Don’t use a “device type” member.  Provide implementations for differing functionality in the derived classes, and install the appropriate handlers, etc.
* Use address map functions to set up your internal address space rather than installing handlers in `device_start`.
* Either allocate your buffers dynamically on start or let the memory system do it for you (e.g. with `.ram()` in address map functions or using `memory_share_creator`).
* Use the memory system to dispatch to handlers for specific functions rather than using big switch statements inside your handlers.
* If you can, split up the big “registers” array into meaningful pieces.

ksherlock:
I've rewritten it to move all the registers into discrete fields and split the sockets into a separate device to simplify their registers and tcp/udp support. The w5100s support was also removed to keep things simple since it's not needed for the Uthernet II card.

PR #11390, opened 2023-07-01
pc.cpp: Mark mc1702 bios as BAD_DUMP

einstein95:
When comparing the internal font to the font in the one in the ibm5150 rev1 bios, there are a number of errors that appear to have been caused by some nibbles being duplicated or flipped. Most notably is that part of `8` is gone and a garbage line appears at the top of `♂`.
| ibm5150_rev1 | mc1702 |
|--------|--------|
| ibm5150_rev1 | mc1702 |


rfka01:
Xolod just sent me a new dump for the MC1702 and the host machine, a DEC Rainbow Pro-350 clone. I'm checking if the issues are gone and then submit a PR.

PR #11396, opened 2023-07-02
heathkit/tlb.cpp: Clear screen lines when DE is deasserted

mgarlanger:
On the H19, when turning off line 25, it should no longer be seen on the screen. On a real H19, that line is no longer scanned out on the CRT and will fade quickly. In mame, the screen still shows the line. This fixes it by putting black on the screen when the `de` value is set.

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -338,11 +338,6 @@ MC6845_UPDATE_ROW(heath_tlb_device::crtc_update_row)
 	}
 	else
 	{
-		// for the additional scan lines we need to clear them to properly support
-		// line 25 of the H19. On the real H19, the crtc is reprogramed to not
-		// display these lines and the memory remains the same. Only when the line
-		// 25 is reenabled will the memory be cleared. On a real CRT the screen
-		// will fade, but here, have to explicitly clear them from the screen.
 		const rgb_t color = palette[0];
 
 		for (uint16_t x = 0; x < x_count; x++)
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -305,38 +305,57 @@ void heath_tlb_device::right_shift_w(int state)
 
 MC6845_UPDATE_ROW(heath_tlb_device::crtc_update_row)
 {
-	if (!de)
-	{
-		return;
-	}
-
 	rgb_t const *const palette = m_palette->palette()->entry_list_raw();
 	uint32_t *p = &bitmap.pix(y);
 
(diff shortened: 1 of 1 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -305,38 +305,57 @@ void heath_tlb_device::right_shift_w(int state)
...
-	for (uint16_t x = 0; x < x_count; x++)
-	{
-		uint8_t inv = (x == cursor_x) ? 0xff : 0;
-
-		uint8_t chr = m_p_videoram[(ma + x) & 0x7ff];
+	if (de) {
 
-		if (chr & 0x80)
+		for (uint16_t x = 0; x < x_count; x++)
 		{
```
cuavas, line 314: Please be consistent with your formatting.  You’ve got mixed K&R and Allman here.

src/mame/heathkit/tlb.cpp
```diff
@@ -305,38 +305,57 @@ void heath_tlb_device::right_shift_w(int state)
...
+			*p++ = palette[BIT(gfx, 0)];
+		}
+	}
+	else
+	{
+		// for the additional scan lines we need to clear them to properly support
+		// line 25 of the H19. On the real H19, the crtc is reprogramed to not
+		// display these lines and the memory remains the same. Only when the line
+		// 25 is reenabled will the memory be cleared. On a real CRT the screen
+		// will fade, but here, have to explicitly clear them from the screen.
```
cuavas, line 345: This comment seems to misunderstand how emulated video works.  The system is drawing black if DE is inactive, so you need to fill the relevant part of the bitmap with black.  The output bitmap represents beam intensity.  It isn’t automatically cleared between frames.  If the system outputs black, you need to fill the applicable parts of the output bitmap with black.  Phosphor persistence, beam focus, video bandwidth limitations, etc. is handled in shaders.

mgarlanger: I've removed the comment, but thinking about it, it seems since screen_device represents the beam intensity, it should start each scan out with an zeroed array.

cuavas: It doesn’t because that would be additional overhead when most systems are going to be rendering complete lines and overwriting it anyway.  There’s no reason to fill it before telling the system to update the dirty area.

Also, there’s more than one kind of raster bitmap display.  Consider a passive matrix LCD (e.g. the Nintendo Game Boy).  In that case, the undriven state is “whiter than white”, and driven pixels are darkened.

TL;DR the system/device is responsible for filling the area it’s told to update, and must not make assumptions about what it contains on entry.

PR #11397, opened 2023-07-02
capcom/lwings.cpp: hooked up MCU for avengers [David Haywood]

Osso13:
If/when this gets merged I (Ivan) will probably do a round of cleanups (i.e. moving avengers in a subclass since it's the only set in the driver that needs the MCU emulation code). Just didn't want to make it more difficult to check the actual changes for this PR.

### DIFF
--- a/src/mame/capcom/lwings.cpp
+++ b/src/mame/capcom/lwings.cpp
@@ -78,17 +78,21 @@ Notes:
 
 void lwings_state::avengers_adpcm_w(uint8_t data)
 {
+	//logerror("%s: avengers_adpcm_w %02x\n", machine().describe_context().c_str(), data);
+
 	m_adpcm = data;
 }
 
 uint8_t lwings_state::avengers_adpcm_r()
 {
+//  logerror("%s: avengers_adpcm_r\n", machine().describe_context().c_str());
+
 	return m_adpcm;
 }
 
 void lwings_state::lwings_bankswitch_w(uint8_t data)
 {
-//  if (data & 0xe0) printf("bankswitch_w %02x\n", data);
+//  if (data & 0xe0) logerror("bankswitch_w %02x\n", data);
 //  Fireball writes 0x20 on startup, maybe reset soundcpu?
 	m_sprbank = (data & 0x10)>>4; // Fireball only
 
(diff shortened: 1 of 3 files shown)


src/mame/capcom/lwings.cpp
```diff
@@ -121,161 +125,23 @@ void lwings_state::avengers_interrupt(int state)
...
 	{
-		m_param[3] = data;
-	}
-	else if (pc == 0x0445)
-	{
-		m_soundstate = 0x80;
-		m_soundlatch->write(data);
+		m_mcu->set_input_line(0, ASSERT_LINE);
+		m_irq_ack_timer->adjust(attotime::from_msec(1000/60));
 	}
```
cuavas, line 134: You probably want to either use a latch device, or use a sync barrier so the microcontroller sees the interrupt and updated value at the same time.

src/mame/capcom/lwings.cpp
```diff
@@ -1062,15 +932,83 @@ void lwings_state::trojan(machine_config &config)
...
+	// possibly 0xbf -> 0xff transition - NO can't be, 0xbf doesn't happen on param writes for angle?
+	//if (data == 0xff)
+	//  m_mcu->set_input_line(0, CLEAR_LINE);
+}
+
+
+TIMER_CALLBACK_MEMBER(lwings_state::irq_ack)
+{
+	m_mcu->set_input_line(0, CLEAR_LINE);
+}
```
cuavas, line 986: This seems a bit hacky.  What does the actual interrupt handler code look like?

src/mame/capcom/lwings.cpp
```diff
@@ -1062,15 +932,83 @@ void lwings_state::trojan(machine_config &config)
...
+void lwings_state::mcu_p1_w(u8 data)
+{
+}
+
+void lwings_state::mcu_p2_w(u8 data)
+{
+	m_mcu->set_input_line(0, CLEAR_LINE); // not sure, we also clear on a timed ack
+	m_soundstate = 0x80;
+	m_soundlatch->write(data);
+}
```
cuavas, line 972: The I/O ports aren’t strobed, so clearing the interrupt on any write isn’t possible.

ajrhacker:
I've put together a somewhat less hacky version of these changes. I've avoided setting bizarre and inexplicable CPU clock speeds by doing something a bit more clever (but intensive).

PR #11399, opened 2023-07-03
bfm/bfm_sc6.cpp: Added skeleton driver to hold Scorpion 6 Flash image files

james-wallace-ghub:
Deal or No Deal Easy Money (Scorpion 6, 9561082)
Deal or No Deal Easy Money (Scorpion 6, 9561082, protocol)
Deal or No Deal Easy Money (Scorpion 6, 9560933)
Deal or No Deal Easy Money (Scorpion 6, 9560933, protocol)


james-wallace-ghub:
With Haze closing his account, this fell on the floor. Such as it is, it's a perfectly normal skeleton, and Easy Money definitely doesn't violate the 3 year rule. I don't want any credit for this as I've literally just CTRL-C and V'd as appropriate.

cuavas:
The previous incarnation of this PR was marked as a draft, so it wasn’t mergeable without bypassing that.  Was there a reason it was never marked ready for review?

cuavas (changes requested):
At the very least, you’re going to need to fix the mame.lst changes so you aren’t completely nuking it.  There’s something very wrong there.

qeeg:
My guess there is that this person's editor fucked up on the whitespace or something.

cuavas:
> My guess there is that this person's editor fucked up on the whitespace or something.

It’s more likely they simply copied a file from a branch made before mame.lst was sorted by subproject hierarchy over the top of the file in their new branch, not considering that the actual changes need to be applied to mame.lst in its current state.

qeeg:
That is also entirely plausible. Unfortunately though, I can't know without checking out this branch on my own machine, and honestly? Fuuuuuck that lol.

cuavas:
> That is also entirely plausible. Unfortunately though, I can't know without checking out this branch on my own machine, and honestly? Fuuuuuck that lol.

Maybe it would be better to think before making irrelevant comments.

cuavas:
You’re going to need to completely replace the mame.lst with the one from master and just add the Scorpion 6 games.  It’s still completely messed up, and doing more merges on top is just going to make things worse in terms of potential breakage.

james-wallace-ghub:
Seeing as mame.lst is playing silly beggars with the line endings, I've put this in draft pending a decent fix.

As for why it never got to ready, I think it was literally the account deletion occurred before it could be swapped in, it having the effect of not just closing the old PR, but killing access

cuavas:
> As for why it never got to ready, I think it was literally the account deletion occurred before it could be swapped in, it having the effect of not just closing the old PR, but killing access

No, even before Haze closed his account **it was marked as a draft pull request and hence couldn’t be merged**.  I remember this distinctly, as I’d looked at it and didn’t see anything that was obviously blocking it at a glance.

james-wallace-ghub:
Reworked this to not clobber the line endings.

Don't know why this was set to draft before closure, but as a skeleton it is what it is. My only concern is the naming convention - strictly speaking, I think some protocol sets we've given 'p' suffixes to, but in this case I don't know if it's best to follow that at this stage.

cuavas:
Remember to srcclean the driver file.  It’s currently missing the newline at the end of the file, and make sure you have your git `core.autocrlf` setting set to `true` to make sure line endings are normalised.

I don’t think there’s anything particularly objectionable about the driver.  However, it’s one that’s inevitably going to be problematic due to encryption:
* 128-bit AES is effectively impossible to break with current understanding of cryptography.
* If the keys are volatile and difficult to exfiltrate, we’ll have a hard time ever being able to decrypt any of these games without a leak.
* Do we even know whether the keys are game-specific or common across a platform?

The ColdFire CPU is in the 68000 family, but without any unencrypted code at all, adding a CPU to the skeleton won’t add much value as disassembly will be garbage anyway.

blueonesarefaster:
> I don’t think there’s anything particularly objectionable about the driver. However, it’s one that’s inevitably going to be problematic due to encryption:
> 
> * 128-bit AES is effectively impossible to break with current understanding of cryptography.
> * If the keys are volatile and difficult to exfiltrate, we’ll have a hard time ever being able to decrypt any of these games without a leak.
> * Do we even know whether the keys are game-specific or common across a platform?
> 
> The ColdFire CPU is in the 68000 family, but without any unencrypted code at all, adding a CPU to the skeleton won’t add much value as disassembly will be garbage anyway.

The Reflex version uses 256-bit AES to encrypt the update files. Keys are fixed in the bootloader so don't change. IV changes for each update and is stored in the update file.
Might be worth adding the few Reflex updates which are available ?

BFG version is completely different.

PR #11407, opened 2023-07-05
capcom/cps1.cpp: Updated qadjr DIP switch settings

Abadede:
- Changed qadj difficulty and demo sound settings according to the user manual
- Fixed capitalization inconsistency in daimakair level select strings.


cuavas:
Enabling demo sounds by default even when the manual suggests disabling it is usually an intentional choice, as it simplifies a basic go/no-go sound emulation test.  If demo sounds are enabled, you can let the attract loop run and record the sound output without needing to play back an input recording.

PR #11413, opened 2023-07-10
Fix Qt6 detection on Linux

jmakovicka:
- Look for MOC in QT_HOME/libexec if not found in QT_HOME/bin

- Use the Qt6* libraries if the Qt version is >= 6

### DIFF
--- a/scripts/src/osd/modules.lua
+++ b/scripts/src/osd/modules.lua
@@ -378,12 +378,19 @@ function qtdebuggerbuild()
 			MOC = "moc"
 		else
 			if _OPTIONS["QT_HOME"]~=nil then
-				QMAKETST = backtick(_OPTIONS["QT_HOME"] .. "/bin/qmake --version 2>/dev/null")
-				if (QMAKETST=='') then
-					print("Qt's Meta Object Compiler (moc) wasn't found!")
-					os.exit(1)
+				MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/bin/moc --version 2>/dev/null")
+				if (MOCTST=='') then
+					MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/libexec/moc --version 2>/dev/null")
+					if (MOCTST=='') then
+						print("Qt's Meta Object Compiler (moc) wasn't found!")
+						os.exit(1)
+					else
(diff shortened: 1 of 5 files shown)


rb6502:
I think using Qt 6 if it's available is a good idea, but yeah, we don't want to break Qt 5 yet.

jmakovicka:
Pardon me, I forgot about the include changes when adding the Qt5/Qt6 conditionals. Should build on Qt5 now.

scripts/src/osd/modules.lua
```diff
@@ -378,12 +378,19 @@ function qtdebuggerbuild()
...
+						print("Qt's Meta Object Compiler (moc) wasn't found!")
+						os.exit(1)
+					else
+						MOC = _OPTIONS["QT_HOME"] .. "/libexec/moc"
+					end
+				else
+					MOC = _OPTIONS["QT_HOME"] .. "/bin/moc"
 				end
-				MOC = _OPTIONS["QT_HOME"] .. "/bin/moc"
+				MOC = _OPTIONS["QT_HOME"] .. "/libexec/moc"
```
cuavas, line 393: Why does this immediately change `MOC` after assigning it inside both the `if` and `else` branches?  Surely line 393 should be removed since it always cancels out the assignment at line 391.

belegdol:
Building with Qt 6 only works when QT_HOME is specified. If it is not, the autodetection only looks for Qt 5. Is this expected?

jmakovicka:
Sort of - I only wanted to add an option to build with Qt6 without patching the source, and kept the default unchanged to avoid breakage. Moreover, on my system (Debian Linux) the Qt binaries are not present in the $PATH anyway, so my ability to test the code path without QT_HOME is limited.

cuavas:
> Fair enough. Autodetection works for Qt 5 on Fedora. @cuavas, would having autodetection work for Qt 6 in addition be of interest?

Yes, with the usual conditions:
* Don’t break Qt5, don’t break MSYS2 MinGW, don’t break other Linux distros.
* Leave Qt5 as the default for now.
* No code that makes me puke ;)

belegdol:
OK, turns out it is more difficult than I thought, especially given the third condition ;) I think it makes more sense to define `QT_HOME` for now and switch to Qt6 entirely in the future. The complexity needed to autodetect both seems not to be worth it.

PR #11418, opened 2023-07-14
new working clone Scorpion ZS-256 TURBO+

holub:
spectrum/zxbus.cpp: add SMUC
sinclair/scorpion.cpp: derive Scorpion ZS-256 TURBO+


src/devices/bus/spectrum/zxbus.h
```diff
@@ -81,24 +81,37 @@ DECLARE_DEVICE_TYPE(ZXBUS_SLOT, zxbus_slot_device)
...
 	}
 
+	u8 io_r(offs_t offset);
+	void io_w(offs_t offset, u8 data);
+
 	void add_slot(zxbus_slot_device &slot);
+	void set_custom_spaces() { m_allocspaces = true; }
+
+	space_config_vector memory_space_config() const override;
+	const address_space_config m_io_config;
```
cuavas, line 103: Should this member really be public?

src/devices/bus/spectrum/zxbus.h
```diff
@@ -107,15 +120,14 @@ DECLARE_DEVICE_TYPE(ZXBUS, zxbus_device)
 
 class device_zxbus_card_interface : public device_interface
 {
-public:
-	virtual ~device_zxbus_card_interface();
-
-	void set_zxbusbus(zxbus_device &bus) { assert(!device().started()); m_zxbus = &bus; }
+  friend class zxbus_slot_device;
```
cuavas, line 123: Indentation is off on this line.

src/devices/bus/spectrum/zxbus.h
```diff
@@ -81,24 +81,37 @@ DECLARE_DEVICE_TYPE(ZXBUS_SLOT, zxbus_slot_device)
...
+
+	space_config_vector memory_space_config() const override;
+	const address_space_config m_io_config;
 
 protected:
 	zxbus_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock);
 
-	virtual void device_start() override;
+	void device_config_complete() override;
+	void device_start() override;
```
cuavas, line 109: Please include the `virtual` qualifier on overriden virtual member functions.  The vast majority of MAME includes it.

src/devices/bus/spectrum/zxbus.h
```diff
@@ -107,15 +120,14 @@ DECLARE_DEVICE_TYPE(ZXBUS, zxbus_device)
...
-	void set_zxbusbus(zxbus_device &bus) { assert(!device().started()); m_zxbus = &bus; }
+  friend class zxbus_slot_device;
 
 protected:
 	device_zxbus_card_interface(const machine_config &mconfig, device_t &device);
 
-	virtual void interface_pre_start() override;
+	void interface_pre_start() override;
+
+	void set_zxbusbus(zxbus_device *zxbus) { assert(!device().started()); m_zxbus = zxbus; }
```
cuavas, line 130: The bus should never be null - the parameter should be a reference rather than a pointer.

src/devices/bus/spectrum/zxbus.cpp
```diff
@@ -41,12 +40,40 @@ zxbus_device::zxbus_device(const machine_config &mconfig, const char *tag, devic
...
+
+u8 zxbus_device::io_r(offs_t offset)
+{
+	return m_iospace->read_byte(offset);
+}
+
+void zxbus_device::io_w(offs_t offset, u8 data)
+{
+	m_iospace->write_byte(offset, data);
 }
```
cuavas, line 77: You should use a `memory_access<...>::specific` for this if possible.  It might also be worth inlining this in the header.

src/devices/bus/spectrum/zxbus.cpp
```diff
@@ -41,12 +40,40 @@ zxbus_device::zxbus_device(const machine_config &mconfig, const char *tag, devic
...
+	, m_allocspaces(false)
+{
+}
+
+device_memory_interface::space_config_vector zxbus_device::memory_space_config() const
 {
+	return space_config_vector {
+		std::make_pair(AS_IO, &m_io_config),
+	};
+}
```
cuavas, line 55: If only want the space sometimes, you don’t want to return the configuration for it if you don’t want it.

src/devices/bus/spectrum/zxbus.h
```diff
@@ -81,24 +81,37 @@ DECLARE_DEVICE_TYPE(ZXBUS_SLOT, zxbus_slot_device)
...
+	template<typename T> void install_device(offs_t addrstart, offs_t addrend, T &device, void (T::*map)(class address_map &map), u64 unitmask = ~u64(0))
 	{
 		m_iospace->install_device(addrstart, addrend, device, map, unitmask);
 	}
 
+	u8 io_r(offs_t offset);
+	void io_w(offs_t offset, u8 data);
+
 	void add_slot(zxbus_slot_device &slot);
+	void set_custom_spaces() { m_allocspaces = true; }
```
cuavas, line 100: Is adding `device_memory_interface` here the best approach, or would it be better for the system using it to create a lightweight device with an address space to use?  For example the driver can instantiate an `address_map_bank_device` just for the purpose of attaching the `zxbus_device` to its address space, and ignore its banking features entirely.  That would reduce complexity and potential for misconfiguration in this device.

src/devices/bus/spectrum/smuc.cpp
```diff
@@ -0,0 +1,140 @@
...
+// copyright-holders:Andrei I. Holub
+
+#include "emu.h"
+#include "smuc.h"
+
+#include "bus/ata/atapicdr.h"
+#include "bus/ata/ataintf.h"
+#include "machine/ds17x85.h"
+#include "bus/ata/hdd.h"
+#include "machine/i2cmem.h"
```
cuavas, line 11: Please sort the `#include` statements.

src/devices/bus/spectrum/smuc.cpp
```diff
@@ -0,0 +1,140 @@
...
+		, device_zxbus_card_interface(mconfig, *this)
+		, m_rtc(*this, "rtc")
+		, m_eeprom(*this, "eeprom")
+		, m_ata(*this, "ata")
+    { }
+
+protected:
+	void device_add_mconfig(machine_config &config) override;
+	void device_start() override;
+	void device_reset() override;
```
cuavas, line 31: Please include the `virtual` qualifier on overridden virtual member functions.

cuavas (changes requested):

src/devices/bus/spectrum/zxbus.h
holub: I'm thinking maybe zxbus can be this device? It's fit perfectly for this purpose. Just need some time to refactor existing drivers to use zxbus this way and then we can get rid of this "custom" spaces.

The only challenge I have now is how to prioritize customizes and make driver to contribute first to this spaces. Currently driver adds it's configuration last.

holub: > Is adding `device_memory_interface` here the best approach, or would it be better for the system using it to create a lightweight device with an address space to use? For example the driver can instantiate an `address_map_bank_device` just for the purpose of attaching the `zxbus_device` to its address space, and ignore its banking features entirely. That would reduce complexity and potential for misconfiguration in this device.

That is really awesome. Thank you a lot for the tip!

src/mame/sinclair/scorpion.cpp
```diff
@@ -126,67 +186,106 @@ void scorpion_state::port_1ffd_w(offs_t offset, u8 data)
...
+
+	map(0xa021, 0xa021).mirror(0x1fdc) // BFFD | 101xxxxxxx1xxx01
+		.lw8(NAME([this](u8 data) { m_ay[m_ay_selected]->data_w(data); }));
+	map(0xe021, 0xe021).mirror(0x1fdc) // FFFD | 111xxxxxxx1xxx01
+		.lr8(NAME([this]() { return m_ay[m_ay_selected]->data_r(); })).w(FUNC(scorpion_state::ay_address_w));
+
+	// Mouse
+	map(0xfadf, 0xfadf).lr8(NAME([this]() { return 0x80 | (m_io_mouse[2]->read() & 0x07); }));
+	map(0xfbdf, 0xfbdf).lr8(NAME([this]() { return m_io_mouse[0]->read(); }));
+	map(0xffdf, 0xffdf).lr8(NAME([this]() { return ~m_io_mouse[1]->read(); }));
```
cuavas, line 271: Due to C integer promotion rules, the second lambda here won’t return `u8` natively and may end up needing to instantiate an adapter function.  Please force it to return `u8`, either like:
```c++
	map(0xffdf, 0xffdf).lr8(NAME([this] () -> u8 { return ~m_io_mouse[1]->read(); }));
```

or like:
```c++
	map(0xffdf, 0xffdf).lr8(NAME([this] () -> { return u8(~m_io_mouse[1]->read()); }));
```

src/mame/sinclair/scorpion.cpp
```diff
@@ -201,21 +300,27 @@ void scorpion_state::machine_start()
 	spectrum_128_state::machine_start();
 
 	save_item(NAME(m_port_1ffd_data));
+	save_item(NAME(m_nmi_pending));
+	save_item(NAME(m_ay_selected));
+	save_item(NAME(m_ram_banks));
+
+	m_maincpu->space(AS_PROGRAM).specific(m_program);
```
cuavas, line 307: Can you also get a `memory_access<...>::specific` for `m_bankio`?  Since it implements `device_memory_interface`, it should be possible to call `m_bankio->space(0).specific(...);` so you can avoid using the `read8` and `write8` trampolines.  Since you aren’t actually using the banking functionality, the logic in `read8` and `write8` can be ignored.

cuavas (commented):
This looks close to being ready.

PR #11419, opened 2023-07-15
modules.lua: Remove stray MOC setting after the Qt6 Linux support change

jmakovicka:


cuavas:
This passed the CI builds.  Is there a reason you’ve still got it marked as a draft?

PR #11424, opened 2023-07-18
taito/lkage.cpp: Add variant of The Legend of Kage with MSM5232 [Takahiro Nogi]

Osso13:
New working clones
------------------
The Legend of Kage (MSM5232 ver) [Takahiro Nogi]


Osso13:
Preliminary work, opening for comments.
To be checked:
- can the draw_sprites methods (and thus the screen_update methods) be merged?
- I don't think the soundlatch could be correct as implemented, but I'm not sure

cuavas:
The only differences between the two `draw_sprites` implementations are a fixed X offset and different sprite list size.  You could calculate the end offset from the size of the sprite list RAM share, couldn’t you?  You could pass in the X offset as a parameter, or keep it in an instance data member.  At any rate, there’s definitely no need to copy/paste the entire functions.

PR #11426, opened 2023-07-20
nscsi: Correct SCSI command names and add missing commands

mietek:
This commit corrects the spelling of some SCSI command names and adds a number of missing commands, following the SCSI-2 draft specification.  This work is done in preparation for [adding SCSI tape drive support](https://github.com/mietek/mame/compare/mietek:mame:pretape...mietek:tape), based on [the SIMH tape image format](http://simh.trailing-edge.com/docs/simh_magtape.pdf).

CC @galibert


cuavas:
@mietek has a trailing U+200F right-to-left marker character in their “real name” used on git commits.  This seems malicious, to mess with display of git logs and confuse scripts.

mietek:
> @mietek has a trailing U+200F right-to-left marker character in their “real name” used on git commits. This seems malicious, to mess with display of git logs and confuse scripts.

I most certainly don’t.  It would be nice if you could check your scripts before trying to imply malice.

cuavas:
My scripts have been used every month for years.  They use well-tested Python modules to read the git log.

The problematic character is present in the git log without using my scripts.  Use an editor that highlights control characters like vim to see it.  For example `git log -1 src/devices/bus/nscsi_bus.h | vim -` – it’s displayed as `Author: Miëtek Bak<200f> <mietek@bak.io>` in vim.

After some more investigation, it seems to come from your display name on GitHub which is used for the “real name” on commits from merged pull requests.  It’s also present in the display name span on this page: https://github.com/mietek (can be seen by saving the HTML and opening in a text editor that will show it).  Please fix it.

I assumed it was malicious because use of this specific character was a common attack on web forums after Unicode support was initially rolled out.  Slashdot in particular had suffered many attacks where people would use this character in the comment subject field to alter web page display in a malicious way.  This and similar attacks are the reason Slashdot still blacklist most Unicode blocks.

eschaton:
I know Mietek in real life and am willing to vouch that, if this code point is present in their username, it isn’t intentional. Always been a stand-up person by me.

mietek:
I reset my GitHub display name to an ASCII string, then back to my name.  Hopefully, this fixes that.

PR #11429, opened 2023-07-21
snes.xml: Metadata fixed

ArcadeShadow:
- Renamed description "Val d'Isère Championship" to match the box title.
- Toy Story: Promoted parent set to working (it works as well as the clones).

### DIFF
--- a/hash/snes.xml
+++ b/hash/snes.xml
@@ -31427,7 +31427,7 @@ Alternate board (XL-1)
 		</part>
 	</software>
 
-	<software name="toystory" supported="no">
+	<software name="toystory">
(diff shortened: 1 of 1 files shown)


angelosa:
> Toy Story: Promoted parent set to working (it works as well as the clones).

I'm gonna assume you tested with `snespal` right? 
Regardless needs a note about being (most likely) picky with main/sound comms timings, something that I really need to multiple e2e someday ...

ArcadeShadow:
> I'm gonna assume you tested with `snespal` right? 

Correct.

angelosa (approved):

PR #11430, opened 2023-07-21
nscsi: Add SCSI tape device based on SIMH tape image format

mietek:
This commit adds SCSI tape drive support, following [the SCSI-2 draft specification](https://www.staff.uni-mainz.de/tacke/scsi/SCSI2.html), and based on [the SIMH tape image format](http://simh.trailing-edge.com/docs/simh_magtape.pdf).

This work has been tested mostly with `macii`, `macqd700`, and `macqd800`, using System 7.1.1 Pro and [DeskTape 2.0](https://macintoshgarden.org/apps/desktape).  Example usage:

    mame macii mac711 -ramsize 8m -nothrottle -scsi:5 tape

CC @galibert 

## Review requests

- [x] `simh_tape_image.h`: Is the string used in `simh_tape_image_device::image_type_name` correct?
- [x] `simh_tape_image.h`: Is the string used in `simh_tape_image_device::image_brief_type_name` correct?
- [x] `simh_tape_image.cpp`: Are the strings used in `DEFINE_DEVICE_TYPE` correct?
- [ ] `simh_tape_image.cpp`: Are the strings used in `simh_tape_image_device::device_config_complete` correct?
- [x] `tape.cpp`: Are the strings used in `DEFINE_DEVICE_TYPE` correct?
- [x] `tape.cpp`: Is `osd_printf_error` a correct method of reporting exceptions?
- [x] `tape.cpp`: Is the string used in `nscsi_tape_device::nscsi_tape_device` correct?
- [x] `tape.cpp`: Is the string used in `nscsi_tape_device::device_add_mconfig` correct?
- [x] `tape.cpp`: Should we save all items in `nscsi_tape_device::device_start`?

## Limitations

We support all commands defined as mandatory for sequential devices by the SCSI-2 draft specification, and some commands defined as optional.

We self-identify in response to `INQUIRY` as `MAME SCSI tape drive 1.0`.  Some emulated software may require other vendor, product, and revision values.

We only pretend to support `PREVENT`/`ALLOW MEDIUM REMOVAL`, since the MAME UI does not support preventing the user from removing tape images.  Tape images should be unmounted in the guest OS before being removed by the user to avoid odd guest OS behavior.

We only pretend to support `RESERVE`/`RELEASE UNIT`, since MAME does not support multiple initiators.

Our handling of `MODE SELECT` and `MODE SENSE` is minimal, and may need to be extended in order to support additional emulated software.  In particular, in response to `MODE SENSE`, we only report a zero-length page 0, and a density code of 0 in the block descriptor, since the SIMH tape image specification does not include tape density.

We default to a block length of 512 bytes for fixed-length blocks.  The block length can be changed with `MODE SELECT` up to 16MB.  Variable-length blocks are supported up to the same block length limit.

We default to mounting tape images for reading and writing.  Read-only mounting is accomplished by ensuring the tape image file is write-protected, for example with `chmod -w`.

The medium size is defined by the size of the tape image file.  We assume the early-warning region starts at 32KB from end-of-medium, instead of the 10ft required by ANSI, since the SIMH tape image specification does not include tape length.

We support creating blank tape images from within MAME, and default to a medium size of 60MB, since the MAME UI does not support specifying creation parameters.  Any file consisting of a sequence of `0xFF` bytes is acceptable as a blank tape image file.

We support filemarks, but not setmarks, because setmarks are not defined by the SIMH tape image specification. We do not support spacing over sequential filemarks, since it is defined as optional by the SCSI-2 draft specification, and we do not have a good way to test any filemark-related functionality.

Our handling of the SIMH tape image format conforms to the standard format, as defined by the SIMH specification, but we do not support seeking through erase gaps.  In other words, any data blocks following an erase gap in a tape image are inaccessible.

## Changes to nscsi

`nscsi_bus.h` only provided a few SCSI Additional Sense Codes as 8-bit `SK_ASC_...` constants.  We provide all SCSI Additional Sense Codes and SCSI Additional Sense Code Qualifiers defined by the SCSI-2 draft specification (with SCSI-3 additions) as 16-bit `SKC_...` constants, to match their actual usage.

We augment `nscsi_full_device::sense` with the more expressive `nscsi_full_device::set_sense_data` and associated helpers.

We change the minimum length of response to `REQUEST SENSE` from 4 to 18 in `nscsi_full_device`, as this seems correct.

## Future work

In `nscsi_full_device`, `nscsi_harddisk_device`, and `nscsi_cdrom_device`:

- Replace uses of `nscsi_full_device::sense` with appropriate helpers.
- Reject commands other than `INQUIRY` and `REQUEST SENSE` that are addressed to LUNs other than 0.

In `nscsi_harddisk_device` and `nscsi_cdrom_device`:

- Subtract 1 from the mode data length in response to `MODE SENSE`.

In `nscsi_harddisk_device`:

- Correct `scsi_put_data` to stop treating incoming `MODE SELECT` data as `WRITE` data, which could lead to data loss in unlikely circumstances.


einstein95:
Windows builds failing
```
Compiling src/lib/util/simh_tape_file.cpp...
../../../../../src/lib/util/simh_tape_file.cpp:12:10: fatal error: 'sys/mman.h' file not found
#include <sys/mman.h>
         ^~~~~~~~~~~~
1 error generated.
```

This file doesn't exist in the Windows build environments

galibert:
You know, if you see errors/limitations in scsi_full_device, don't just override in your device, fix the full device instead :-)

src/devices/bus/nscsi/tape.cpp
```diff
@@ -0,0 +1,1028 @@
...
+
+#include <cassert>
+#include <cstring>
+#include "emu.h"
+#include "multibyte.h"
+#include "tape.h"
+#include "simh_tape_file.h"
+
+#if 0
+#define LOG(...)    fprintf(stderr, __VA_ARGS__)
```
pmackinlay, line 15: Please use logmacro.h for logging.

eschaton:
> Windows builds failing
> 
> ```
> Compiling src/lib/util/simh_tape_file.cpp...
> ../../../../../src/lib/util/simh_tape_file.cpp:12:10: fatal error: 'sys/mman.h' file not found
> #include <sys/mman.h>
>          ^~~~~~~~~~~~
> 1 error generated.
> ```
> 
> This file doesn't exist in the Windows build environments

I think the solution here might be to switch to `core_file` like the rest of the stuff in `imagedev` uses instead of raw C/POSIX I/O.

src/devices/bus/nscsi/tape.h
```diff
@@ -0,0 +1,115 @@
...
+
+protected:
+	// basic state
+	int m_fixed_block_len; // fixed-length block length
+	required_device<simh_tape_image_device> m_image; // tape image
+	int m_sequence_counter; // tape image identifier
+	bool m_tape_changed; // should we report unit attention next time we receive command
+
+	// state for READ and WRITE
+	uint8_t m_rw_buf[32768]; // read/write buffer; TODO: adjust size or make it configurable, once we encounter tape images that use larger block lengths
```
rb6502, line 30: MAME prefers a std::unique_ptr<uint8_t> for this kind of buffer.  That way you can dynamically allocate it to match the block size as needed.  (See m_vram in src/mame/apple/dafb.cpp for an example).

src/devices/bus/nscsi/tape.cpp
```diff
@@ -0,0 +1,1028 @@
...
+	scsi_cmdbuf[3] = bd_disable ? 0 : 8; // block descriptor length
+	if (!bd_disable) { // respond with block descriptor
+		scsi_cmdbuf[4] = m_image->get_file() ? m_image->get_file()->get_density_code() : 0; // density code
+		put_uint24be(&scsi_cmdbuf[9], m_fixed_block_len); // block length
+	}
+	scsi_data_in(SBUF_MAIN, resp_len);
+	scsi_status_complete(SS_GOOD);
+}
+
+void nscsi_tape_device::handle_request_sense(const int lun) // mandatory; SCSI-2 section 8.2.14
```
pmackinlay, line 500: Please correct nscsi_full_device::scsi_command() with this logic instead of reimplementing here.

src/devices/bus/nscsi/tape.cpp
```diff
@@ -0,0 +1,1028 @@
...
+{
+	const int alloc_len = scsi_cmdbuf[4]; // allocation length
+	LOG("REQUEST SENSE lun=%d alloc_len=%d\n", lun, alloc_len);
+	if (scsi_cmdbuf[1] & 0x1f || scsi_cmdbuf[2] || scsi_cmdbuf[3]) // error: reserved bits set
```
pmackinlay, line 504: Please add parentheses around the binary and; although not required, some compilers will produce warnings here.

src/devices/bus/nscsi/tape.cpp
```diff
@@ -0,0 +1,1028 @@
...
+		default:
+			return nscsi_full_device::scsi_put_data(id, pos, data);
+	}
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+// condition reporting
+
+void nscsi_tape_device::set_sense_data(const int sense_key, const int sense_key_code, const sense_data data) // TODO: move this function to nscsi_full_device
```
pmackinlay, line 209: I think this (and the report_x() helpers) should be promoted into nscsi_full_device.

src/lib/util/simh_tape_file.cpp
```diff
@@ -0,0 +1,624 @@
...
+// copyright-holders:Mietek Bak
+
+// best read together with SIMH magtape spec (rev 17 Jan 2022)
+// http://simh.trailing-edge.com/docs/simh_magtape.pdf
+
+#include <cassert>
+#include <cstring>
+#include <stdexcept>
+#include <errno.h>
+#include <fcntl.h>
```
rb6502, line 11: As was mentioned in global comments, please use MAME's file I/O rather than POSIX or other direct platform calls.

src/devices/bus/nscsi/tape.cpp
```diff
@@ -0,0 +1,1028 @@
...
+		case SC_MODE_SELECT_10:
+		case SC_MODE_SENSE_10:
+		case SC_READ_BUFFER:
+		case SC_READ_REVERSE:
+		case SC_RECEIVE_DIAGNOSTIC_RESULTS:
+		case SC_RECOVER_BUFFERED_DATA:
+		case SC_VERIFY_6:
+		case SC_WRITE_BUFFER:
+		default: // error: unsupported command
+			return report_bad_cmd(cmd);
```
pmackinlay, line 149: I think this should fall back to nscsi_full_device::scsi_command() after promoting set_sense_data() and helpers.

src/lib/util/simh_tape_file.cpp
```diff
@@ -0,0 +1,624 @@
...
+	}
+	if (!create) { // load existing file
+		const int result = lseek(m_fd, 0, SEEK_END);
+		if (result == -1) // error: we failed to seek to end of file
+			throw std::runtime_error(std::string("lseek: ") + strerror(errno));
+
+		m_buf_len = result;
+	}
+	else { // create new file
+		m_buf_len = 62914560; // we default to 60MB; TODO: allow specifying tape image size, once MAME supports it
```
rb6502, line 82: The specific mechanism isn't directly applicable here (@galibert may have some ideas), but MAME does support multiple options for creating floppy images (400K, 800K, and 1.44M for Apple 3.5" drives, for instance).  So you could choose Create and then have a choice of 60, 120, and 240 MB tapes, or whatever the standard sizes are there.

rb6502 (changes requested):
Aside from what's been commented this is not bad for a first-time submission that touches this many things.  I appreciate the pervasive usage of assert() to make sure everything's OK, that's something we probably should encourage more in MAME.

Also, there are 5 new 68040 Mac models (Quadra 700, Quadra 800, Centris 610, Centris 650, Quadra 610, Quadra 650) in current master that use a more advanced SCSI controller chip (53C96 instead of 5380/53C80) that you should verify compatibility with.  I don't expect problems but you never know.

src/devices/imagedev/simh_tape_image.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "opresolv.h"
+#include "simh_tape_image.h"
+
+#if 0
+#define LOG(...)    fprintf(stderr, __VA_ARGS__)
+#else
+#define LOG(...)    ([](auto &&... args){})(__VA_ARGS__)
+#endif
+
+DEFINE_DEVICE_TYPE(SIMH_TAPE_IMAGE, simh_tape_image_device, "simh_tape_image", "SIMH tape image"); // TODO: not sure these strings are correct
```
pmackinlay, line 14: The strings are OK.

src/lib/util/multibyte.h
```diff
@@ -0,0 +1,163 @@
...
+
+#pragma once
+
+#include <stdint.h>
+
+//////////////////////////////////////////////////////////////////////////////
+
+// unsigned big-endian
+
+inline const uint16_t get_uint16be(const uint8_t *const buf)
```
pmackinlay, line 15: I think the principle of defining these helpers for endian-specific numeric I/O is good, the specific implementation and approach needs to be reviewed and sanctioned by @cuavas. Once that's done, probably the other nscsi code (and probably elsewhere in MAME such as chd) should be reviewed and updated to use them as well.

src/lib/util/simh_tape_file.cpp
```diff
@@ -0,0 +1,624 @@
...
+
+inline const int get_marker_class(const int marker) { return marker >> 28; }
+inline const int get_marker_value(const int marker) { return marker & 0x0fffffff; }
+inline const bool is_marker_half_gap_reverse(const int marker) { return marker >= 0xffff0000 && marker <= 0xfffffffd; }
+
+//////////////////////////////////////////////////////////////////////////////
+
+// construction
+
+simh_tape_file::simh_tape_file(const std::string_view path, const bool create)
```
pmackinlay, line 53: I think being able to read/write SIMH format is perfectly reasonable, but I'm not sure it's the best candidate for a canonical MAME tape format; I don't have any suggestions other than perhaps reviewing whether CHD might be able to be enhanced for the purpose? One significant gap in my mind is that the SIMH format does not appear to have any way to represent what kind of tape or tape drive the image corresponds to. It should not be possible to insert a DAT tape in a QIC drive, for example.

src/lib/util/simh_tape_file.cpp
```diff
@@ -0,0 +1,624 @@
...
+
+// best read together with SIMH magtape spec (rev 17 Jan 2022)
+// http://simh.trailing-edge.com/docs/simh_magtape.pdf
+
+#include <cassert>
+#include <cstring>
+#include <stdexcept>
+#include <errno.h>
+#include <fcntl.h>
+#include <sys/mman.h>
```
pmackinlay, line 12: As already mentioned, mmap() is different on Windows. Either replace with conventional file I/O or propose additions to OSD to support it. Is it really beneficial or necessary given typical memory capacity and "legacy" tape capacities?

pmackinlay (changes requested):
I concur with @rb6502; this is a fantastic first contribution, and we welcome your involvement!

I'm generally OK with the nscsi-related parts, quite unsure about the UI feedback for error handling, dubious about mmap() and file descriptor I/O, and want Vas to comment on the endian-specific numeric encoding library functions.

mietek:
Thank you for all the kind words.

> Also, there are 5 new 68040 Mac models (Quadra 700, Quadra 800, Centris 610, Centris 650, Quadra 610, Quadra 650) in current master that use a more advanced SCSI controller chip (53C96 instead of 5380/53C80) that you should verify compatibility with.

To do this, I’ll need access to the newly required ROM dumps.  I can be found on IRC and Discord under the same nickname.

src/lib/util/multibyte.h
mietek: Meanwhile, I’ll note there are many places in MAME that could use multibyte accessors, and some that already do.  For example:
https://github.com/mamedev/mame/blob/a4e202ae77cc605ad1a8bc006d0b9d80e8952ccf/src/lib/util/aviio.cpp#L577-L715
https://github.com/mamedev/mame/blob/a4e202ae77cc605ad1a8bc006d0b9d80e8952ccf/src/tools/imgtool/modules/ti990hd.cpp#L53-L75

src/lib/util/simh_tape_file.cpp
ajrhacker: The `mmap` approach is not used elsewhere in MAME for file format code (its only use is in `src/mame/gaelco/gaelco3d_m.cpp` for some rather idiosyncratic networking code), and smells very bad to me.

src/lib/util/multibyte.h
ajrhacker: This sort of thing shows up in the filesystem code as well. I did several experiments with utility functions and classes to help deal with this sort of thing (the `sext` function I added to `src/lib/util/coretmpl.h` was a byproduct of one of those), but was not quite convinced that both the API or the implementation details were as correct as they needed to be for the whole MAME project.

src/lib/util/simh_tape_file.cpp
mietek: Why?  To me, the `mmap` approach seems like a natural fit for image files in general, considering their size and typical access patterns.  Loading a ~700MB CD-ROM image file in MAME results in a significant delay.  Tape images can potentially run into multiple GBs.

src/lib/util/simh_tape_file.cpp
rb6502: I don't think SIMH needs to be declared canonical.  If you have something else in mind, go for it.  But interoperating with SIMH is definitely nice as there are hundreds of tapes out there in its format.

src/lib/util/simh_tape_file.cpp
rb6502: MAME does not load an entire CD-ROM or hard disk image all at once, it just reads and writes the current block.  The startup delay is due to an integrity check.  As has been noted, we don't use mmap() for any other file I/O.  If you want to propose a portable mmap() implementation for MAME's file I/O layer that's a thing you can do in a separate PR, but otherwise this one needs to use MAME's file I/O.

src/lib/util/simh_tape_file.cpp
mietek: It’s not clear to me how to provide such a choice within the confines of the MAME UI.  Please advise.

src/lib/util/simh_tape_file.cpp
mietek: I’ll make it clear that the SIMH format is one of potentially many supported formats by introducing an abstract tape file interface.
```c++
enum class tape_status : int {
	OK,
	BOM, // beginning of medium
	EW, // early warning
	FILEMARK,
	FILEMARK_EW, // filemark and early warning
	SETMARK,
	SETMARK_EW, // setmark and early warning
	EOD, // end of data
	EOD_EW, // end of data and early warning
	UNKNOWN,
	UNKNOWN_EW, // unknown and early warning
	EOM // end of medium
};

class tape_file_interface
{
public:
	// construction
	virtual ~tape_file_interface() {}

	// position-preserving operations
	virtual const bool is_read_only() const = 0;
	virtual const bool is_ew() const = 0;
	virtual const uint8_t get_density_code() const = 0;
	virtual const std::pair<const tape_status, const int> read_position() const = 0;

	// non-destructive operations
	virtual void rewind(const bool eom) = 0;
	virtual const tape_status locate_block(const int req_block_addr) = 0;
	virtual const tape_status space_eod() = 0;
	virtual const std::pair<const tape_status, const int> space_blocks(const int req_blocks_num) = 0;
	virtual const std::pair<const tape_status, const int> space_filemarks(const int req_marks_num, const bool setmarks = false, const bool sequential = false) = 0;
	virtual const std::pair<const tape_status, const int> space_blocks_reverse(const int req_blocks_num) = 0;
	virtual const std::pair<const tape_status, const int> space_filemarks_reverse(const int req_marks_num, const bool setmarks = false, const bool sequential = false) = 0;

	virtual const std::pair<const tape_status, const int> read_block(uint8_t *buf, const int buf_size) = 0;

	// destructive operations
	virtual void erase(const bool eom) = 0;
	virtual const tape_status write_block(const uint8_t *const buf, const int len) = 0;
	virtual const tape_status write_filemarks(const int req_marks_num, const bool setmarks = false) = 0;
};
```

src/lib/util/multibyte.h
mietek: I’ll add 64-bit accessors to make `"multibyte.h"` easier to reuse in the future.

src/lib/util/multibyte.h
```diff
@@ -0,0 +1,163 @@
...
+inline void put_int32be(uint8_t *buf, const int32_t data)
+{
+	put_uint32be(buf, data);
+}
+
+//////////////////////////////////////////////////////////////////////////////
+
+// unsigned little-endian
+
+inline const uint32_t get_uint16le(const uint8_t *const buf)
```
mietek, line 90: Oops, found a typo.

src/lib/util/multibyte.h
```diff
@@ -0,0 +1,163 @@
...
+{
+	return (buf[2] << 16) | (buf[1] << 8) | buf[0];
+}
+
+inline const uint32_t get_uint32le(const uint8_t *const buf)
+{
+	return (buf[3] << 24) | (buf[2] << 16) | (buf[1] << 8) | buf[0];
+}
+
+inline void put_uint16le(uint8_t *buf, const uint32_t data)
```
mietek, line 105: Oops, found a typo.

src/devices/bus/nscsi/tape.cpp
mietek: I can do that, but there isn’t much point to it.  All `nscsi_full_device::scsi_command()` does is:
1. handle `SC_REZERO_UNIT`, which is a synonym for `SC_REWIND`, which we must handle on our own;
2. handle `SC_REQUEST_SENSE`, which we must handle on our own as well, due to the medium changing logic.

src/lib/util/simh_tape_file.cpp
mietek: Since there have been no further comments, I’ll leave this to future work.

mietek:
I’ve addressed all of the comments, except that `simh_tape_file.cpp` still needs to be rewritten to use MAME file I/O.

rb6502:
Just to clarify: you are going to rewrite to use MAME file I/O and you just want the rest reviewed while that happens?

mietek:
Yes.  It would be helpful to know what is considered a good usage example of MAME file I/O.  For example, `src/lib/util/cdrom.cpp` (used by `nscsi_cdrom_device`) still uses POSIX calls.

simzy39:
@rb6502

rb6502:
I don't know what would be considered a *good* example (that's @cuavas territory) but a nice simple example would be lib/util/harddisk.cpp in the non-CHD path for each function.

mietek:
`src/lib/util/harddisk.cpp` uses `src/lib/util/ioprocs.h`, which does not seem to define a method for creating files.  However, I also received a suggestion to use `src/osd/osdfile.h`, which does define such a method, and so I’m going to try using that.

src/lib/util/multibyte.h
```diff
@@ -0,0 +1,260 @@
...
+
+inline const s16 get_s16be(const u8 *const buf)
+{
+	return get_u16be(buf);
+}
+
+inline const s32 get_s24be(const u8 *const buf)
+{
+	const struct {
+		const s32 sign_extended:24;
```
pmackinlay, line 102: Please consider replacing this bitfield (and the others below) with util::sext() instead.

pmackinlay:
I have no objections to this being merged now.

mietek:
Thank you.  I believe I’ve addressed all comments now.

mietek:
Now updated to the latest `master`.

simzy39:
@rb6502

rb6502 (approved):

PR #11437, opened 2023-07-24
Add an IPT_LEVER device class.

Hydreigon223:
This was suggested by Ville years back and am intending to make this suggestion come true for racingj, racingj2 and thrilld. The handbrake lever has previously been configured by an IPT_PEDAL (gticlub) or an AD_STICK_Y (racingj, racingj2, thrilld). The latter was never ideal due to the fact the handbrake starts/autocenters at minimum range on the actual cabinet.


Hydreigon223:
The suggested changes may trigger some devs in the process and I am aware of the consequences of attempting a stunt like this.

happppp:
When I think "lever", I think of one that auto sticks to each edge. Like the digital stick shifter on racing games. We don't have an analog input type for those.
But what you have here can be solved with IPT_PEDAL3, no need for a new input type.

Hydreigon223:
> When I think "lever", I think of one that auto sticks to each edge. Like the digital stick shifter on racing games. We don't have an analog input type for those. But what you have here can be solved with IPT_PEDAL3, no need for a new input type.

In nwk-tr, IPT_PEDAL3 was bound to the clutch pedal.

happppp:
If you need more than 3, borrow one from 'player 2'?

Hydreigon223:
> If you need more than 3, borrow one from 'player 2'?

As a matter of fact, I attempted this change in a previous pull request two years ago but that got shut down due to a disagreement of some sort. I'll be more than happy to borrow a pedal input from 'player 2' like the last time. Thanks for agreeing with me.

Hydreigon223:
Going to close this because something went wrong when attempting to revert changes I made to the core input files. Will reopen under a new PR soon.

PR #11439, opened 2023-07-24
sinclair/scorpion.cpp: Added Scorpion GMX.

holub:


src/mame/mame.lst
```diff
@@ -40930,6 +40930,7 @@ profi                           //
 quorum                          //
 scorpio                         //
 scorpiontb                      //
+scorpiongmx                     //
```
cuavas, line 40933: Please keep lines sorted in this file.

src/mame/sinclair/scorpion.cpp
```diff
@@ -23,15 +23,25 @@
...
 namespace {
 
+#define ROM_PAGE_SYS 2
+#define ROM_PAGE_DOS 3
+
+#define DOS        m_beta->is_active()
+#define ROMRAM     BIT(m_port_1ffd_data, 0)
+#define ROM1       BIT(m_port_7ffd_data, 4)
+
+#define GMX_FIXROM BIT(m_port_00_data, 4)
```
cuavas, line 33: Please replace these with private member functions, e.g.
```c++
    bool dos() const { return m_beta->is_active(); }
    bool romram() const { return BIT(m_port_1ffd_data, 0); }
```
etc.

Macros always give worse error messages, etc. when things go wrong.

src/mame/sinclair/scorpion.cpp
```diff
@@ -23,15 +23,25 @@
 
 namespace {
 
+#define ROM_PAGE_SYS 2
+#define ROM_PAGE_DOS 3
```
cuavas, line 27: Please use `constexpr` for constants rather than preprocessor macros.

src/mame/sinclair/scorpion.cpp
```diff
@@ -168,7 +229,25 @@ INPUT_CHANGED_MEMBER(scorpion_state::on_nmi)
...
+	m_nmi_pending = 0;
+}
+
+u8 scorpion_state::port_ff_r()
+{
+	if (m_magic_lock)
+	{
+		LOGIO("Registers lock released\n");
+		m_magic_lock = 0;
+	}
```
cuavas, line 246: You need to add ` && !machine().side_effects_disabled()` to the condition here, or it will screw up when a memory window is open.

src/mame/sinclair/scorpion.cpp
```diff
@@ -487,6 +563,21 @@ u8 scorpiontb_state::ay_data_r()
 		: m_ay[m_ay_selected]->data_r();
 }
 
+INPUT_CHANGED_MEMBER(scorpiontb_state::turbo_changed)
+{
+	m_turbo = !m_turbo;
+	m_maincpu->set_clock_scale(1 << m_turbo);
+	popmessage("Turbo %s\n", m_turbo ? "ON" : "OFF");
+}
```
cuavas, line 571: You should really be checking `newval` here to make sure you don’t get out sync with the state of the toggle input, as it’s saved in the CFG file for the machine (also remember to check it on `machine_reset` to make sure you’re in the right state when starting).

src/mame/sinclair/scorpion.cpp
```diff
@@ -558,16 +651,242 @@ void scorpiontb_state::scorpion_ioext(address_map &map)
...
+
+u8 scorpiongmx_state::port_78fd_r()
+{
+	u8 tmp = (BIT(m_port_fe_data, 1) << 7) // BRD1
+		| (m_port_78fd_data & 0x7f);
+	tmp |= (m_magic_shift & 1);
+	m_magic_shift >>= 1;
+	return tmp;
+}
```
cuavas, line 766: You need to check `!machine().side_effects_disabled()` before updating `m_magic_shift` or bad things will happen.

cuavas (changes requested):

src/mame/sinclair/scorpion.cpp
holub: Converted to ioport. Per my understanding it should provide proper handling of store state out of the box. Can you take a closer look please?

src/mame/sinclair/scorpion.cpp
```diff
@@ -520,6 +614,7 @@ void scorpiontb_state::machine_reset()
 
 	scorpion_state::machine_reset();
 	m_is_m1_even = 0;
+	m_io_turbo->field(0x01)->live().value = 0;
 	m_maincpu->set_clock_scale(1);
```
cuavas, line 618: You really shouldn’t be messing with the value of toggle controls like this – it’s by design that they “stick” across runs.  You switch the machine off, the toggle switch stays in the same position.

cuavas (commented):
Is the turbo switch actually a physical toggle switch, or is it a momentary pushbutton that causes the machine to switch between normal and turbo modes?

If it’s a physical toggle switch, you shouldn’t be trying to change its state on start, as the switch will stay in the same position while the machine is not being used.  Trying to force it to a particular state breaks expectations.

If it’s a momentary pushbutton that switches speed and the system always starts in “normal” speed, you should use `IPT_KEYPAD` or `IPT_OTHER` or something for it, and then switch speed when `newval == 1`.

holub:
> Is the turbo switch actually a physical toggle switch, or is it a momentary pushbutton that causes the machine to switch between normal and turbo modes?
> 
> If it’s a physical toggle switch, you shouldn’t be trying to change its state on start, as the switch will stay in the same position while the machine is not being used. Trying to force it to a particular state breaks expectations.
> 
> If it’s a momentary pushbutton that switches speed and the system always starts in “normal” speed, you should use `IPT_KEYPAD` or `IPT_OTHER` or something for it, and then switch speed when `newval == 1`.

It's actually both. Commonly switch done by software, but if hard button pressed it's really useful to draw the popup message.
But I better get rid of popup rather than introduce new variable to glue both together if there is no beautiful solution for that.

holub:
Oh. I think I've got it. Must be similar to NMI we already have. Let me complete this change.

PR #11440, opened 2023-07-24
Added support for the Roland D70

giulioz:
Boots and all the menus work with proper controls. Can also produce some random sounds when playing the demo song.

image

image

### DIFF
--- a/src/mame/roland/roland_d70.cpp
+++ b/src/mame/roland/roland_d70.cpp
@@ -3,6 +3,7 @@
 /****************************************************************************
 
     Driver for Roland D-70 synthesizer.
+    Derived by the CM32P driver by ValleyBell
 
 ****************************************************************************/
 
@@ -41,14 +42,15 @@ namespace {
 
 #define UNSCRAMBLE_DATA(_data) bitswap<8>(_data, 1, 2, 7, 3, 5, 0, 4, 6)
 
-#define CONT_MASK 0b10000000
-#define RW_MASK 0b01000000
-#define AD_MASK 0b00100000
-#define SCK_MASK 0b00010000
-#define SI_MASK 0b00001000
-#define SO_MASK 0b00000100
-#define ACK_MASK 0b00000010
-#define ENCO_MASK 0b00000001
+// Bitmasks for the display board interface via PORT1
+static constexpr uint8_t CONT_MASK = 0b10000000;
+static constexpr uint8_t RW_MASK = 0b01000000;
+static constexpr uint8_t AD_MASK = 0b00100000;
+static constexpr uint8_t SCK_MASK = 0b00010000;
+static constexpr uint8_t SI_MASK = 0b00001000;
+static constexpr uint8_t SO_MASK = 0b00000100;
+static constexpr uint8_t ACK_MASK = 0b00000010;
+static constexpr uint8_t ENCO_MASK = 0b00000001;
 
 static INPUT_PORTS_START(d70)
   PORT_START("KEY0")
@@ -136,9 +138,9 @@ class roland_d70_state : public driver_device {
 public:
   roland_d70_state(const machine_config &mconfig, device_type type,
                    const char *tag)
-      : driver_device(mconfig, type, tag), cpu(*this, "maincpu"),
-        pcm(*this, "pcm"), lcd(*this, "lcd"), midi_timer(*this, "midi_timer"),
-        ram(*this, "ram"), dsp_ram(*this, "dsp_ram"), m_keys(*this, "KEY%u", 0),
+      : driver_device(mconfig, type, tag), m_cpu(*this, "maincpu"),
+        m_pcm(*this, "pcm"), m_lcd(*this, "lcd"), m_midi_timer(*this, "midi_timer"),
+        m_ram(*this, "ram"), m_dsp_ram(*this, "dsp_ram"), m_keys(*this, "KEY%u", 0),
         sw_scan_index(0), sw_scan_bank(-1), sw_scan_state(0xFF), sw_scan_mode(0),
         sw_scan_current_out(0xFF), m_midi_rx(0), m_midi_pos(0) {}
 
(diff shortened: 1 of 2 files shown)


src/mame/roland/roland_d70.cpp
```diff
@@ -0,0 +1,562 @@
...
+              6, 5, 4, 0)
+
+// PCM cards use a different address line scrambling
+#define UNSCRAMBLE_ADDR_EXT(_offset)                                           \
+  bitswap<19>(_offset, 18, 17, 8, 9, 16, 11, 12, 7, 14, 10, 13, 15, 3, 2, 1,   \
+              6, 4, 5, 0)
+
+#define UNSCRAMBLE_DATA(_data) bitswap<8>(_data, 1, 2, 7, 3, 5, 0, 4, 6)
+
+#define CONT_MASK 0b10000000
```
rb6502, line 44: For this kind of thing we prefer static constexpr over #define.  Gives some type safety and it's more modern C++.

src/mame/roland/roland_d70.cpp
```diff
@@ -0,0 +1,562 @@
...
+  virtual void machine_reset() override;
+
+private:
+  required_device<i8x9x_device> cpu;
+  required_device<mb87419_mb87420_device> pcm;
+  required_device<t6963c_device> lcd;
+  required_device<timer_device> midi_timer;
+  required_device<ram_device> ram;
+  required_device<ram_device> dsp_ram;
+  required_ioport_array<8> m_keys;
```
rb6502, line 160: We prefer but don't require the m_ prefix for member variables.  Consistency is more important though, and I'm flagging this because having one m_ and the rest not stands out a bit.

src/mame/roland/roland_d70.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+void roland_d70_state::lcd_palette(palette_device &palette) const {
+  palette.set_pen_color(0, rgb_t(138, 146, 148));
+  palette.set_pen_color(1, rgb_t(69, 62, 66));
+}
+
+void roland_d70_state::d70_map(address_map &map) {
+  map(0x0100, 0x0100).w(FUNC(roland_d70_state::bank_w));
+  map(0x0400, 0x07ff).rw(FUNC(roland_d70_state::ksga_io_r), FUNC(roland_d70_state::ksga_io_w));
+  map(0x0800, 0x0800).rw(FUNC(roland_d70_state::lcd_ctrl_r), FUNC(roland_d70_state::lcd_ctrl_w));
```
rb6502, line 431: For cases like this where an 8-bit peripheral is on one byte lane of a 16-bit data bus you can map it directly and add .umask16(0xff00) or .umask16(0x00ff) to choose which lane it's on.  Then you don't need the trampoline function.

src/mame/roland/roland_d70.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+  TIMER(config, "samples_timer")
+      .configure_periodic(FUNC(roland_d70_state::samples_timer_cb),
+                          attotime::from_hz(32000 * 2));
+
+  midi_port_device &mdin(MIDI_PORT(config, "mdin", midiin_slot, "midiin"));
+  mdin.rxd_handler().set([this](int state) {
+    // logerror("midi bit %x pos %x\n", state, m_midi_pos);
+
+    if (m_midi_pos == 0 || m_midi_pos == 9) {
```
rb6502, line 491: This isn't ideal but it's not your fault since the CPU device doesn't have proper serial yet.

rb6502 (changes requested):
This is great work for a first-time submission, and it's always great to see more people doing stuff with synths.  Just a few notes and we should be able to get this in.

giulioz:
@rb6502 thanks for the review! I applied your suggestions and added some comments

rb6502 (approved):
That's much better, thanks!  I'll let the CI builds run to make sure it compiles everywhere, and give @ajrhacker and @devinacker a chance to take a look but I think we're good otherwise.  Please send the necessary ROMs as described at https://wiki.mamedev.org/index.php?title=Submitting_Source_Code if you haven't yet.

devinacker:
The driver seems to be directly based on roland_cm32p.cpp - do you think they're similar enough to be made into a single driver?

Aside from that, I'm not familiar with the D70 at all so I can't really offer anything in the way of useful feedback.

rb6502:
I'm not sure on the merging, looking at the other driver that's definitely a possibility.  I'll have to look closer.  

Meanwhile the CI does show some issues though: descramble_rom_external() and midi_w() are unused, and -validate fails because the ROM region doesn't match the CPU data bus width and endianness.  The easy fix for that one is to give it the same name as the CPU, like the CM32P driver ("maincpu" instead of "program").  You may need to change how the program ROM is loaded then to swap bytes.

rb6502:
Hi Giulio, I'm *extremely* sorry about this dragging on so long.  Totally my fault, no excuses, but let's get things going again.  I'm in the process of getting it in (I am merging it with roland_cm32p.cpp as Devin suggested because they're very similar, including sharing 4 of the wave ROMs).  Can you point me to the ROMs?  I don't see a complete set on the usual places like DBWBP.

giulioz:
Hi @rb6502 no problem! :)
Here there should be all of them: https://dbwbp.com/synthbin/roland_d70.zip

rb6502:
Thanks @giulioz !  When I run it I get the initial logo animation, then parts of the display clear and nothing else happens.  I assume your driver as-is does more than that?

rb6502:
The cleaned up version was committed in https://github.com/mamedev/mame/commit/d5aebf8557e8a1f6ead5e32fe39c996d867397e5 so hopefully we can go forward from there.  Thanks again for submitting this in the first place!

giulioz:
@rb6502 thank you!

Axis4s:
does this even work? i have the roms

giulioz:
It boots but no sound still, for that an accurate emulation of 4 (!) chips is needed
- Address generator
- DPCM decoder
- TVF (same as JD800/S770)
- RCC (chorus/reverb, same as CM32P/U220)

PR #11451, opened 2023-07-30
Add initial support for wayland to bgfx renderer

belegdol:
This is an initial working prototype of wayland working with bgfx renderer. It only works with `-bgfx_backend opengl` due to https://github.com/bkaradzic/bgfx/pull/3143 still needing work. Proper conditionalisation is also required so that X11 or Wayland are included and linked only when required.
~~In order to get to this stage, update to the libraries was required:~~
~~- bgfx to 3101a0d93f024e4278caaab29566647cd3e2bf84~~
~~- bx to 198cd120e4941d8aeb677db945ef4291a1ced291~~
~~- bimg to ec02df824a763b2e2ae31e19c674ba0bc88c0695~~
Libs update moved to #11493 for clarity.

While still hacky, it allows for bgfx in a native Wayland window:

Bildschirmfoto vom 2023-07-30 21-42-47


belegdol:
Now vulkan renderer is working on wayland as well. I am not sure whether this is ready to be merged due to the following:

- with vulkan renderer on wayland the window is blinking with `umk3` on my nvidia system. Not sure where the issue is coming from
- in comparison with current state, attempting to start mame with `-videodriver wayland` when it is not built in results in segmentation fault
- with wayland support built in, attempting to start mame with `-videodriver x11` results in segmentation fault due to `-bgfx_backend vulkan` being default

Nevertheless, it shows that native wayland is possible.

scripts/src/3rdparty.lua
```diff
@@ -1514,8 +1518,6 @@ end
 	}
 	if _OPTIONS["targetos"]=="macosx" then
 		files {
-			MAME_DIR .. "3rdparty/bgfx/src/glcontext_eagl.mm",
-			MAME_DIR .. "3rdparty/bgfx/src/glcontext_nsgl.mm",
 			MAME_DIR .. "3rdparty/bgfx/src/renderer_mtl.mm",
```
cuavas, line 1521: Will this potentially cause issues for older Apple hardware that has poor Metal support?

src/osd/modules/render/drawbgfx.cpp
cuavas, line 1: You really should follow the code style of files you’re modifying.  Also, why are you casting data pointers through `uintptr_t`?  That’s only necessary when coercing pointer-to-code to pointer-to-data or vice versa.

scripts/src/3rdparty.lua
belegdol: Probably, but OpenGL support was dropped from upstream bgfx:
https://github.com/bkaradzic/bgfx/commit/928800feab95d7bc3f4680d91c3e4754c852d783

If the PR is interesting but we want to keep OpenGL on Macs, I can see if the necessary commits can be cherry-picked instead.

cuavas: @rb6502 how many people are going to want my head if bgfx via OpenGL is no longer possible on macOS?

rb6502: I just got done promising Mac users I wouldn't lift the system requirements until January, so I would strongly prefer to hold off on anything that would affect that until then.

rb6502: Particularly as Belegdol says this doesn't actually work that well (which describes the totality of Wayland in my experience, to be fair).

belegdol: Fair enough. Maybe we can leave this PR open in the draft state as a proof-of-concept?

belegdol:
I have now pushed an update allowing wayland and X11 to be compiled in simultaneously. The patches needed to achieve this are not merged to upstream bgfx yet, but I am putting these here in case someone is curious and would like to test it themselves.
SDL upstream have also clarified that the `egl_window` pointer being defined the way it was was an accident meaning that creating it with `wl_egl_window_create` is the way to go.
With nvidia binary driver this is not really usable: with vulkan renderer the window is blinking whereas opengl one falls back to software rendering [1], but I would tend do blame these on generally subpar wayland support with the nvidia drivers.

[1] https://forums.developer.nvidia.com/t/hardware-egl-not-working-on-wayland-libegl-warning-egl-failed-to-create-dri2-screen/262167

belegdol:
I moved the update to the 3rdparty libraries to a separate PR (#11493) for clarity.

cuavas:
> I moved the update to the 3rdparty libraries to a separate PR (#11493) for clarity.

It’s still showing as having all those 3rdparty files changed (1063 files changed).

belegdol:
> > I moved the update to the 3rdparty libraries to a separate PR (#11493) for clarity.
> 
> It’s still showing as having all those 3rdparty files changed (1063 files changed).

Apologies. This is either me failing at github or github not being able to create a pr based on another PR. If I understand correctly, it would work if the b* libs update could be temporarily put in a mamedev branch.
Alternatively I could rebase against master but this would make testing slightly more complicated.

belegdol:
Apologies if it is a dumb question, but is the code in https://github.com/mamedev/mame/blob/36f5b6e3be7a2b4e7f9539d1bc9ddc683f7b95f2/src/osd/modules/render/drawbgfx.cpp#L490-L528 actually used? I put a breakpoint at https://github.com/mamedev/mame/blob/36f5b6e3be7a2b4e7f9539d1bc9ddc683f7b95f2/src/osd/modules/render/drawbgfx.cpp#L515 but it seems not to get hit.
If this code is not being used, making the change cleaner would be much simpler. Thanks!

belegdol:
> > > I moved the update to the 3rdparty libraries to a separate PR (#11493) for clarity.
> > 
> > 
> > It’s still showing as having all those 3rdparty files changed (1063 files changed).
> 
> Apologies. This is either me failing at github or github not being able to create a pr based on another PR. If I understand correctly, it would work if the b* libs update could be temporarily put in a mamedev branch. Alternatively I could rebase against master but this would make testing slightly more complicated.

I rebased onto master, will deal with the dependencies locally.

belegdol:
@cuavas, is this better style-wise? Also, if `sdlNativeWindowHandle()` is actually needed, which namespace would be the best to put `create_wl_egl_window` into?

cuavas:
> Apologies if it is a dumb question, but is the code in
> 
> https://github.com/mamedev/mame/blob/36f5b6e3be7a2b4e7f9539d1bc9ddc683f7b95f2/src/osd/modules/render/drawbgfx.cpp#L490-L528
> actually used? I put a breakpoint at
> 
> https://github.com/mamedev/mame/blob/36f5b6e3be7a2b4e7f9539d1bc9ddc683f7b95f2/src/osd/modules/render/drawbgfx.cpp#L497
> but it seems not to get hit.
> If this code is not being used, making the change cleaner would be much simpler. Thanks!

It’s used by `renderer_bgfx::create` if you have multiple output windows/screens.  It definitely *is* needed.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -423,10 +449,11 @@ bool video_bgfx::set_platform_data(bgfx::PlatformData &platform_data, osd_window
 		platform_data.nwh = wmi.info.cocoa.window;
 		break;
 #endif
-#if defined(SDL_VIDEO_DRIVER_WAYLAND) && SDL_VERSION_ATLEAST(2, 0, 16)
+#if defined(SDL_VIDEO_DRIVER_WAYLAND) && SDL_VERSION_ATLEAST(2, 0, 16) && defined(SDLMAME_USE_WAYLAND)
 	case SDL_SYSWM_WAYLAND:
 		platform_data.ndt = wmi.info.wl.display;
-		platform_data.nwh = wmi.info.wl.egl_window;
+		platform_data.nwh = (void *)create_wl_egl_window(dynamic_cast<sdl_window_info const &>(window).platform_window(), wmi.info.wl.surface);
+		platform_data.type = bgfx::NativeWindowHandleType::Wayland;
```
cuavas, line 456: Why are you explicitly casting to `void *`?  You don’t seem to be casting away CV qualifiers, and you aren’t doing a dynamic cast to the canonical object pointer.  You should be able to implicitly convert to `void *` if you just want the pointer value to be preserved.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -382,6 +386,28 @@ bool video_bgfx::init_bgfx_library(osd_window &window)
...
+		int width, height;
+		SDL_GetWindowSize(window, &width, &height);
+		if(!surface)
+			return nullptr;
+		win_impl = wl_egl_window_create(surface, width, height);
+		SDL_SetWindowData(window, "wl_egl_window", win_impl);
+	}
+	return win_impl;
+}
+#endif
```
cuavas, line 408: Please follow the formatting of the file:
* Space before but not after right-associative unary `*`
* Space between flow control keyword and opening parenthesis of controlling expression

Why are you checking `surface` *after* getting the window size?  You won’t use the window size if it’s null.

Can `wl_egl_window_create` fail?

Unlike the other cases in the `switch (wmi.subsystem)` this can fail.  However, you aren’t checking for failure and returning `false` from `video_bgfx::set_platform_data` when that happens.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -513,9 +540,22 @@ static void *sdlNativeWindowHandle(SDL_Window *window)
...
+			{
+				int width, height;
+				SDL_GetWindowSize(window, &width, &height);
+				struct wl_surface* surface = wmi.info.wl.surface;
+				if(!surface)
+					return nullptr;
+				win_impl = wl_egl_window_create(surface, width, height);
+				SDL_SetWindowData(window, "wl_egl_window", win_impl);
+			}
+			return (void*)win_impl;
```
cuavas, line 557: This is non-trivial copy/pasta – shouldn’t it be factored out?

src/osd/modules/render/drawbgfx.cpp
belegdol: Implicit conversion seems to be working, will fix it.

src/osd/modules/render/drawbgfx.cpp
belegdol: Should be fixed now.

src/osd/modules/render/drawbgfx.cpp
belegdol: Done.

belegdol:
I have attempted to implement your feedback to the best of my abilities. Thanks to your hint about multiple windows, it appears that there is an issue with vulkan on wayland whereby second window is not being shown. It is with the opengl renderer strangely enough. I need to investigate further.
Thank you for your patience so far.

belegdol:
> I have attempted to implement your feedback to the best of my abilities. Thanks to your hint about multiple windows, it appears that there is an issue with vulkan on wayland whereby second window is not being shown. It is with the opengl renderer strangely enough. I need to investigate further. Thank you for your patience so far.

Multiple windows needs to be fixed in bgfx:
https://github.com/bkaradzic/bgfx/pull/3162

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -382,6 +386,36 @@ bool video_bgfx::init_bgfx_library(osd_window &window)
...
+		win_impl = wl_egl_window_create(surface, width, height);
+		if (!win_impl)
+			{
+				osd_printf_error("Creating wayland window failed\n");
+				return nullptr;
+			}
+		SDL_SetWindowData(window, "wl_egl_window", win_impl);
+	}
+	return win_impl;
+}
```
cuavas, line 415: Indentation is off in two places – you’ve bumped two levels for some of the `if` statements.  The braces should be at the level of the `if` statement itself.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -423,10 +457,11 @@ bool video_bgfx::set_platform_data(bgfx::PlatformData &platform_data, osd_window
 		platform_data.nwh = wmi.info.cocoa.window;
 		break;
 #endif
-#if defined(SDL_VIDEO_DRIVER_WAYLAND) && SDL_VERSION_ATLEAST(2, 0, 16)
+#if defined(SDL_VIDEO_DRIVER_WAYLAND) && SDL_VERSION_ATLEAST(2, 0, 16) && defined(SDLMAME_USE_WAYLAND)
 	case SDL_SYSWM_WAYLAND:
 		platform_data.ndt = wmi.info.wl.display;
-		platform_data.nwh = wmi.info.wl.egl_window;
+		platform_data.nwh = create_wl_egl_window(dynamic_cast<sdl_window_info const &>(window).platform_window(), wmi.info.wl.surface);
+		platform_data.type = bgfx::NativeWindowHandleType::Wayland;
```
cuavas, line 464: Because `create_wl_egl_window` can fail (unlike the other branches of this `if` statement), you need to check whether it succeeded here, and `return false` if it didn’t.

src/osd/modules/render/drawbgfx.cpp
belegdol: Done

src/osd/modules/render/drawbgfx.cpp
belegdol: Should be fixed now. Is it also needed for `sdlNativeWindowHandle()`?

cuavas: > Should be fixed now. Is it also needed for `sdlNativeWindowHandle()`?

It is, but the function would need to be rearranged a bit.  I’ll worry about that later.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -423,10 +457,16 @@ bool video_bgfx::set_platform_data(bgfx::PlatformData &platform_data, osd_window
...
 	case SDL_SYSWM_WAYLAND:
 		platform_data.ndt = wmi.info.wl.display;
-		platform_data.nwh = wmi.info.wl.egl_window;
+		platform_data.nwh = create_wl_egl_window(dynamic_cast<sdl_window_info const &>(window).platform_window(), wmi.info.wl.surface);
+		if (!platform_data.nwh)
+		{
+			osd_printf_error("BGFX: Error creating a Wayland window\n");
+			return false;
+		}
+		platform_data.type = bgfx::NativeWindowHandleType::Wayland;
```
cuavas, line 469: Should the rest of the branches in this `switch` statement be setting `platform_data.type` to `bgfx::NativeWindowHandleType::Default` to be explicit?  The bgfx examples all seem to be doing it now.

cuavas:
@belegdol can this be marked as ready for review now that the bgfx changes have been merged, or are there other things that would hold it up?

belegdol:
No other things holding up, just wanted to make sure things are merged in the correct order.

PR #11460, opened 2023-08-02
heathkit/h89.cpp: clean up of ASSERT_LINE/CLEAR_LINE usage and variable names

mgarlanger:


src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -38,7 +38,7 @@ void heath_intr_cntrl::device_reset()
 void heath_intr_cntrl::update_intr_line()
 {
 
-	m_irq_line(m_intr_lines == 0 ? CLEAR_LINE : ASSERT_LINE);
+	m_irq_line(m_intr_lines == 0 ? 0 : 1);
 }
```
cuavas, line 42: Can you please put parentheses around the condition in the ternary conditional operator like `(m_intr_lines == 0) ? 0 : 1`?  I know it works as-is, but it’s one of the things that’s more likely to be misinterpreted when someone’s debugging code while sleep-deprived.  Also, the blank line inside the function probably isn’t necessary.

src/mame/heathkit/z37_fdc.cpp
```diff
@@ -49,21 +49,21 @@ void heath_z37_fdc_device::ctrl_w(u8 val)
 {
 	bool motor_on = bool(BIT(val, ctrl_MotorsOn_c));
 
-	m_intrq_allowed = bool(BIT(val, ctrl_EnableIntReq_c));
+	m_irq_allowed = bool(BIT(val, ctrl_EnableIntReq_c));
 	m_drq_allowed = bool(BIT(val, ctrl_EnableDrqInt_c));
-	m_fdc->dden_w(BIT(val, ctrl_SetMFMRecording_c) ? CLEAR_LINE : ASSERT_LINE);
+	m_fdc->dden_w(!BIT(val, ctrl_SetMFMRecording_c));
```
cuavas, line 54: In other places you seem to be avoiding implicit Boolean-to-integer conversion, but you’ve added one here.  Should it be `m_fdc->dden_w(BIT(~val, ctrl_SetMFMRecording_c));` for consistency with the pattern of avoiding the implicit conversions?

src/mame/heathkit/z37_fdc.cpp
```diff
@@ -226,14 +226,14 @@ void heath_z37_fdc_device::device_add_mconfig(machine_config &config)
 
 void heath_z37_fdc_device::set_irq(u8 data)
 {
-	LOGLINES("set irq, allowed: %d data: %d\n", m_intrq_allowed, data);
+	LOGLINES("set irq, allowed: %d data: %d\n", m_irq_allowed, data);
 
-	m_fd_irq_cb(m_intrq_allowed ? data : CLEAR_LINE);
+	m_irq_cb(m_irq_allowed ? data : 0);
 }
```
cuavas, line 232: I just noticed now, but this is attached to a write line callback on the FDC chip – it works because devcb instantiates adapters automatically, but conventionally we would use `set_irq(int state)` since it’s a single line rather than multiple lines in the bits.  Same applies for `set_drq`.

PR #11466, opened 2023-08-04
sinclair/scorpion.cpp: New ProfROM v4/5.xx.31

holub:
scoriontb - Add ProfROM v4.xx.31; scrpiongmx - Add ProfROM v5.xx.31

Not sure what are recommendations in this situation? 
ProfROM is currently under development. New versions have very short release cycle. There is not too much sense to preserve all of them.
I talked to the ProfROM's main developer regarding this topic and he recommends to drop all the version except latest.

Let me know if there are no concerns to cleanup versions what have been already added.


ajrhacker:
MAME generally supports preserving all versions of software, no matter what their developers say.

PR #11467, opened 2023-08-05
samcoupe_flop.xml: Added 2 working items

ArcadeShadow:
Rename "Sam Coupe" to "Sam Coupé"

Rick Dangerous [Spectrum Computing]
Street Fighter 1 [Spectrum Computing]

### DIFF
--- a/hash/samcoupe_flop.xml
+++ b/hash/samcoupe_flop.xml
@@ -35,6 +35,7 @@ license:CC0-1.0
 			</dataarea>
 		</part>
 	</software>
+
 	<software name="samdos1">
 		<description>Sam Coupe System Disk (SAMDOS v1.1)</description>
 		<year>1990</year>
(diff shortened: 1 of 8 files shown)


cuavas (changes requested):
You need to explicitly qualify UTF-8 strings in C++ source code for portability.

ArcadeShadow:
> You need to explicitly qualify UTF-8 strings in C++ source code for portability.

Please, check if the changes made are the way it must be.

ArcadeShadow (commented):
Forced UTF8 encoding

cuavas (changes requested):
Escaped encoded UTF-8 in strings is unreadable.  Why not qualify them as UTF-8 strings as I suggested?

ArcadeShadow:
> Escaped encoded UTF-8 in strings is unreadable. Why not qualify them as UTF-8 strings as I suggested?

Now I think it's fixed like the way it should be.

PR #11468, opened 2023-08-06
dataeast/dec8.cpp: Split classes for reduce optional shared functions, Reduce runtime tag lookups

cam900:
dataeast/dec0.h: remove unused address map

### DIFF
--- a/src/mame/dataeast/dec8.cpp
+++ b/src/mame/dataeast/dec8.cpp
@@ -58,9 +58,9 @@ To do:
 /******************************************************************************/
 
 
-void dec8_state::dec8_mxc06_karn_buffer_spriteram_w(uint8_t data)
+void base_state::buffer_spriteram16_w(u8 data)
 {
-	uint8_t* spriteram = m_spriteram->live();
+	u8* spriteram = m_spriteram->live();
 	// copy to a 16-bit region for the sprite chip
 	for (int i=0;i<0x800/2;i++)
 	{
@@ -69,26 +69,26 @@ void dec8_state::dec8_mxc06_karn_buffer_spriteram_w(uint8_t data)
 }
 
 /* Only used by ghostb, gondo, garyoret, other games can control buffering */
-void dec8_state::screen_vblank_dec8(int state)
+void lastmisn_state::screen_vblank(int state)
 {
 	// rising edge
 	if (state)
 	{
-		dec8_mxc06_karn_buffer_spriteram_w(0);
+		buffer_spriteram16_w(0);
 	}
 }
 
-uint8_t dec8_state::i8751_h_r()
+u8 mcu_state::i8751_h_r()
 {
 	return m_i8751_return >> 8; /* MSB */
 }
 
-uint8_t dec8_state::i8751_l_r()
+u8 mcu_state::i8751_l_r()
 {
 	return m_i8751_return & 0xff; /* LSB */
 }
 
-void dec8_state::i8751_reset_w(uint8_t data)
+void mcu_state::i8751_reset_w(u8 data)
 {
 	// ? reset the actual MCU?
 	//m_i8751_return = 0;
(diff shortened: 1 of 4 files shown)


src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
 #include "screen.h"
 #include "tilemap.h"
 
-class dec8_state : public driver_device
+class base_state : public driver_device
```
cuavas, line 20: The name `base_state` is too generic for a class in the global namespace.  You can only use generic names like that in anonymous namespaces.  Please use `dec8_state_base` instead.

src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
...
-	TILE_GET_INFO_MEMBER(get_ghostb_fix_tile_info);
-	TILE_GET_INFO_MEMBER(get_oscar_fix_tile_info);
+	void allocate_buffered_spriteram16();
+	void dec8_s_map(address_map &map);
+	void oscar_s_map(address_map &map);
+
+};
+
+// with I8751 MCU
+class mcu_state : public base_state
```
cuavas, line 98: Similarly, this needs to be something more specific, like `dec8_mcu_state_base`.

src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
...
-		m_mainbank(*this, "mainbank"),
-		m_nmigate(*this, "nmigate"),
 		m_tilegen(*this, "tilegen%u", 1),
-		m_spritegen_mxc(*this, "spritegen_mxc"),
+		m_mainbank(*this, "mainbank"),
 		m_videoram(*this, "videoram"),
-		m_bg_data(*this, "bg_data"),
-		m_coin_port(*this, "I8751")
+		m_bg_ram(*this, "bg_ram")
 	{ }
```
cuavas, line 37: Since this is an abstract class, the constructor can be protected.

src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
...
+// with I8751 MCU
+class mcu_state : public base_state
+{
+public:
+	mcu_state(const machine_config &mconfig, device_type type, const char *tag) :
+		base_state(mconfig, type, tag),
+		m_mcu(*this, "mcu"),
+		m_coin_port(*this, "I8751")
+	{
+	}
```
cuavas, line 106: This also appears to be an abstract class that can have a protected constructor.

src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
...
+	void i8751_reset_w(u8 data);
+
+	// MCU communication
+	u8  m_i8751_p2 = 0;
+	int m_i8751_port0 = 0;
+	int m_i8751_port1 = 0;
+	int m_i8751_return = 0;
+	int m_i8751_value = 0;
+
+	emu_timer *m_i8751_timer = nullptr;
```
cuavas, line 136: Please keep data members grouped – you’ve got `m_mcu` and `m_coin_port` separate from the rest of the data members.

src/mame/dataeast/dec8.h
```diff
@@ -17,216 +17,299 @@
...
+	void gondo_scroll_w(offs_t offset, u8 data);
+	void gondo_mcu_to_main_w(u8 data);
+
+	TILE_GET_INFO_MEMBER(get_gondo_fix_tile_info);
+	TILE_GET_INFO_MEMBER(get_gondo_tile_info);
+
+	DECLARE_VIDEO_START(lastmisn);
+	uint32_t screen_update_lastmisn(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect);
+
+	void screen_vblank(int state);
```
cuavas, line 202: Please keep the data members grouped and member functions grouped within the `protected` section – you have the `m_spritegen_krn` and `m_nmigate` intermixed with the member functions.

src/mame/dataeast/dec8.cpp
```diff
@@ -226,230 +213,230 @@ void csilver_state::csilver_adpcm_int(int state)
...
+	map(0x180c, 0x180c).w(FUNC(lastmisn_state::sound_w));
+	map(0x180d, 0x180d).w(FUNC(lastmisn_state::lastmisn_control_w)); /* Bank switch + Scroll MSB */
+	map(0x180e, 0x180f).w(FUNC(lastmisn_state::i8751_w));
+	map(0x2000, 0x27ff).ram().w(FUNC(lastmisn_state::videoram_w)).share(m_videoram);
 	map(0x2800, 0x2fff).ram().share("spriteram");
 	map(0x3000, 0x37ff).ram().share("share2");
-	map(0x3800, 0x3fff).rw(FUNC(dec8_state::dec8_bg_data_r), FUNC(dec8_state::dec8_bg_data_w)).share("bg_data");
-	map(0x4000, 0x7fff).bankr("mainbank");
+	map(0x3800, 0x3fff).rw(FUNC(lastmisn_state::bg_ram_r), FUNC(lastmisn_state::bg_ram_w)).share("bg_ram");
+	map(0x4000, 0x7fff).bankr(m_mainbank);
```
cuavas, line 292: You should be able to use `m_bg_ram` here rather than the literal string tag, as well as in the other address map functions.

cuavas (changes requested):
This mostly looks reasonable.  There are just some overly generic class names in the global namespace and a few things that could be cleaned up.

PR #11471, opened 2023-08-09
heathkit/h89.cpp: Correct bad ROMs, add MMS 84B ROM

mgarlanger:
Update H89 ROMS

- MTR-90 (444-84): update to the correct ROM.
- MMS 84A: update to the correct ROM 
- MMS 84B: add rom, add (commented) dip switch settings.
- add chip position for h17 ROM and code roms.
- minor cleanup for consistency in file and across the other heathkit source files.


cuavas:
What do you mean by “update to a working ROM”?  Two ROMs appear to have been removed (SHA1 1b74cafca8213d5c083f16d8a848933ab56eb43b and 03347206dca145ff69ca08435db822b70ce106af).  Where did these actually come from?  Are they real ROM images that *should* work?  Are they ROM images that were released broken?  Are they for a different hardware variant?  Are they simply bad dumps?

mgarlanger:
> What do you mean by “update to a working ROM”? Two ROMs appear to have been removed (SHA1 1b74cafca8213d5c083f16d8a848933ab56eb43b and 03347206dca145ff69ca08435db822b70ce106af). Where did these actually come from? Are they real ROM images that _should_ work? Are they ROM images that were released broken? Are they for a different hardware variant? Are they simply bad dumps?

For SHA1(1b74cafca8213d5c083f16d8a848933ab56eb43b) - the old MMS-84A, it is a bad dump, or was corrupted at some point. The existing dump has byte zero as 0x00. The new MMS-84A has that byte as 0xC3 (JMP), and that is the only difference between .

For SHA1(03347206dca145ff69ca08435db822b70ce106af), I'm not sure where that came from, but it does not match the printed docs that were provided with it the chip or computer (depending if it was bought as an upgrade or came with the computer). I ran the bad ROM with the debugger, and it does not actually set up anything, and quickly goes off into the weeds. Whatever it is, it is not 444-84, not sure it is even for the H89.

PR #11474, opened 2023-08-11
seta/seta.cpp: Added machine configuraton setting for zingzip title language

negunma:
As like Mad Shark, Zing Zing Zip has setting to change title display. It toggles whether display Chinese title.


src/mame/seta/seta.cpp
```diff
@@ -6422,10 +6422,18 @@ static INPUT_PORTS_START( zingzip )
...
-	PORT_BIT( 0x0080, IP_ACTIVE_LOW, IPT_UNKNOWN  )
+	PORT_CONFNAME( 0x0010, 0x0010, DEF_STR( Unknown ) )
+	PORT_CONFSETTING(      0x0010, DEF_STR( Off ) )
+	PORT_CONFSETTING(      0x0000, DEF_STR( On ) )
+	PORT_CONFNAME( 0x0020, 0x0020, DEF_STR( Unknown ) )
+	PORT_CONFSETTING(      0x0020, DEF_STR( Off ) )
+	PORT_CONFSETTING(      0x0000, DEF_STR( On ) )
+	PORT_CONFNAME( 0x0040, 0x0040, DEF_STR( Unknown ) )
+	PORT_CONFSETTING(      0x0040, DEF_STR( Off ) )
+	PORT_CONFSETTING(      0x0000, DEF_STR( On ) )
```
cuavas, line 6433: Without locations or other information, these will be confusing.  Can you at least give the bit numbers or something?  (For example `"Unknown (bit 4)"`, `"Unknown (bit 5)"`, etc.)

src/mame/seta/seta.cpp
```diff
@@ -6422,10 +6422,18 @@ static INPUT_PORTS_START( zingzip )
...
+	PORT_CONFNAME( 0x0080, 0x0080, "Title" )           // Not listed in manual
+	PORT_CONFSETTING(      0x0080, "With Chinese" )
+	PORT_CONFSETTING(      0x0000, "Without Chinese" ) // Without 真真急炮
```
cuavas, line 6436: Maybe make this more descriptive.  Could you call this setting “Title Language”, and call the options “English and Chinese” and “English only”?

PR #11475, opened 2023-08-12
heathkit/h19.cpp: Add dte rs232 port, expose all signals of terminal

mgarlanger:
Enhancements to h19

- Add dte rs232 port to all connecting terminal to other systems
- expose all the signals of the 8250 on the tlb
- specify ROM chip positions
- other minor cleanup

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -812,17 +812,17 @@ const tiny_rom_entry *heath_tlb_device::device_rom_region() const
 	return ROM_NAME(h19);
 }
 
-void heath_tlb_device::serial_out_b(uint8_t data)
+void heath_tlb_device::serial_out_b(int data)
 {
 	m_write_sd(data);
 }
 
-void heath_tlb_device::dtr_out(uint8_t data)
+void heath_tlb_device::dtr_out(int data)
 {
 	m_dtr_cb(data);
 }
 
-void heath_tlb_device::rts_out(uint8_t data)
+void heath_tlb_device::rts_out(int data)
 {
 	m_rts_cb(data);
 }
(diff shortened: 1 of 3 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -834,11 +817,36 @@ void heath_tlb_device::serial_out_b(uint8_t data)
...
 
+void heath_tlb_device::dtr_out(uint8_t data)
+{
+	m_dtr_cb(data);
+}
+
+void heath_tlb_device::rts_out(uint8_t data)
+{
+	m_rts_cb(data);
+}
```
cuavas, line 828: These are “line” handlers – argument should be `int`.

cuavas (commented):
Is it normal to have a loopback plug connected to the DTE port, or would it usually be disconnected if there isn’t a real peripheral connected?

mgarlanger:
> Is it normal to have a loopback plug connected to the DTE port, or would it usually be disconnected if there isn’t a real peripheral connected?

It would usually be connected to some device, an H8, H11, a modem, etc. I saw the wyse50 defaulted to loopback, and figured it by having that, it would help test any ESC codes that trigger the terminal to send characters. It behaves slightly different than the "offline mode", in offline mode the key presses go to the terminal, but sent characters still are sent out the uart. Having the loopback on there, if the user doesn't specify something else, seemed like the best option, but I can remove it if you want it removed.

cuavas:
I’d have to check for specifics for the WY50 terminal.  Most systems in MAME have nothing connected to their serial ports by default if it’s a “user peripheral” port and there isn’t some standard device required to use the system.

For example Osborne 1 has nothing connected to the serial port by default in MAME.  In practice, it was often used to connect a modem for uploading files to head office (the Osborne 1 was often used by traveling employees), but most of the time this isn’t required.

On the other hand, the INTELLEC 4 systems have a terminal connected to the serial port by default, so you can use the monitor program (in practice it was more likely to have been a teletype, like a Model 33).

We can leave the loopback connector for now, it can be changed easily enough.

PR #11486, opened 2023-08-16
heathkit/tlb.cpp: fix nmi handling for H19

mgarlanger:
Fix NMI handling, to only trigger after the proper port is accessed.  Only when the ROM needs to update the CRTC registers does it write to the special port to allow the next VSYNC to trigger a NMI. 

The change fixes the Self-diagnostic test in the UltraROM. It was failing because the scratchpad memory is modified during the NMI interrupt routine, and that was conflicting with the memory test of the scratchpad.
Screenshot 2023-08-15 at 8 44 35 PM

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -15,8 +15,6 @@
     Input can also come from the serial port.
 
   TODO:
-    - determine why ULTRA ROM's self-diag (ESC |) fails for the ROM and
-      scratchpad memory
     - INS8250 needs to implement "Set Break" (LCR, bit 6) before Break key
       will function as expected.
     - 49/50 row mode does not work when DOT clocks are programmed as documented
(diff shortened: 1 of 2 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -280,6 +282,32 @@ void heath_tlb_device::mm5740_data_ready_w(int state)
...
+{
+	m_allow_vsync_nmi = bool(BIT(reg,2));
+
+	m_crtc->register_w(val);
+}
+
+void heath_tlb_device::crtc_vsync_w(int val)
+{
+	m_maincpu->set_input_line(INPUT_LINE_NMI, m_allow_vsync_nmi ? val: CLEAR_LINE);
+}
```
cuavas, line 309: Please clean up the spacing – `BIT(reg, 2)` (space after comma) and `m_allow_vsync_nmi ? val : CLEAR_LINE` (spaces on both sides of the question mark and colon of the conditional operator).

cuavas (changes requested):
Only some inconsistent spacing and this will be good to go (it’s trivial enough that I would’ve done it myself, but it appears you turned off the “allow edits by maintainers” option on this PR).

cuavas:
I’ve got some other accumulated cleanup to do anyway, I’ll merge this now.

mgarlanger:
> Only some inconsistent spacing and this will be good to go (it’s trivial enough that I would’ve done it myself, but it appears you turned off the “allow edits by maintainers” option on this PR).

I only turn it off, since the warning from github makes it sound bad to keep it enabled.  Not sure why github can't just limit it to the PR, and not have it allow other things.

Screenshot 2023-08-16 at 2 15 32 PM

rb6502:
mamedev/mame is pretty locked down in terms of who has write access, and unless you have secret passwords checked into your repository I think it's fine.  And obviously if someone did screw around with someone else's repo through that mechanism they would lose write access to mamedev/mame.

PR #11487, opened 2023-08-16
apple/apple3_m.cpp: Trigger reset/NMI when reset key combinations are pressed

robjustice:


cuavas:
I’ve merged this, but I’m not entirely sure it’s 100% sound.  Consider the following sequence:
* Press Ctrl – nothing happens
* Press Reset – sets `m_reset_latch` and asserts reset
* Release Ctrl – sets `m_nmi_latch` and asserts NMI (reset still asserted)
* Release Reset – clears `m_reset_latch` and releases reset (NMI still asserted)
* No further input changed event happen, NMI never released

robjustice:
Thanks.

Yes, it's not handling that sequence properly. I checked on real hardware
and that sequence leads to a reset as soon as the ctrl key is released.

I added an extra condition to the NMI so that if the reset latch has been
set, then don't trigger NMI. That's working ok now.

    // check for reset key only (NMI)
    else if ((m_kbspecial->read() & 0x80) && (m_via_0_a & ENV_NMIENABLE) &&
!m_reset_latch)
    {
        if (!m_nmi_latch)
        {
            m_nmi_latch = true;
            m_maincpu->set_input_line(INPUT_LINE_NMI, ASSERT_LINE);
        }
    }

I also tested:
 - press and hold reset - NMI is triggered
 - press ctrl
 - release ctrl - Reset is triggered

This is matching the real hw behaviour.

rb6502:
That matches what I'm seeing on the schematic, so yay my ability to read it :-)  I'll merge in the additional changes.

cuavas:
Thanks for following up @robjustice and @rb6502.

PR #11488, opened 2023-08-17
heathkit/h89.cpp: Comment out the extra dip switch setting

mgarlanger:
Needing to comment out the unused dip switch settings, I did not realize that mame would show all dip switch settings even if they were not used. This makes it nearly impossible for the end user to know which of the duplicated settings actually changed the system. Commenting out for now. Looks like I will need to create clones instead of alt BIOS, for the ROMs in which the dip switch settings are different, in order for users to see the correct information about the switches.

Screenshot 2023-08-16 at 10 09 45 PM


src/mame/heathkit/h89.cpp
```diff
@@ -297,14 +297,14 @@ static INPUT_PORTS_START( h89 )
 	PORT_DIPSETTING( 0x10, "Disk I/O #1" )
 	PORT_DIPNAME( 0x20, 0x20, "Perform memory test at start" )  PORT_DIPLOCATION("SW501:6")
 	PORT_DIPSETTING( 0x20, DEF_STR( Off ) )
-	PORT_DIPSETTING( 0x00, DEF_STR( On ) )
+	PORT_DIPSETTING( 0x00, "On" )
```
cuavas, line 300: If you’re using `DEF_STR` for one of the on/off pair, you should probably use it for both.

PR #11491, opened 2023-08-18
apple/apple3_v.cpp: Add support for flashing chars and remove smooths…

robjustice:
…cr from text modes

### DIFF
--- a/src/mame/apple/apple3_v.cpp
+++ b/src/mame/apple/apple3_v.cpp
@@ -140,7 +140,6 @@ void apple3_state::text40(bitmap_ind16 &bitmap, const rectangle &cliprect)
 {
 	uint8_t const *const ram = m_ram->pointer();
 	uint32_t const ram_size = m_ram->size();
-	int const smooth = m_va | (m_vb << 1) | (m_vc << 2);
 	int const beginrow = (cliprect.top() - (cliprect.top() % 8)) / 8;
 	int const endrow = (cliprect.bottom() - (cliprect.bottom() % 8) + 7) / 8;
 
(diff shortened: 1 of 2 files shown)


src/mame/apple/apple3_v.cpp
```diff
@@ -167,33 +166,25 @@ void apple3_state::text40(bitmap_ind16 &bitmap, const rectangle &cliprect)
...
 			{
 				for (int col = 0; col < 7; col++)
 				{
-					int lc;
-					if (m_smoothscr)
+					uint16_t *const dest = &bitmap.pix(y * 8 + row, x * 14 + col * 2);
+
+					// Flash or Inverse
+					if ((!(ch & 0x80) && (char_data[row] & 0x80) && m_flash) || (!(ch & 0x80) && !(char_data[row] & 0x80)))
 					{
```
cuavas, line 179: This seems to have excess parentheses and common subexpressions.  Wouldn’t it be clearer as:
```c++
 		if (!(ch & 0x80) && (!(char_data[row] & 0x80) || m_flash))
```

That’s equivalent logic after minimising the truth table.

src/mame/apple/apple3_v.cpp
```diff
@@ -219,44 +209,50 @@ void apple3_state::text80(bitmap_ind16 &bitmap, const rectangle &cliprect)
...
 			{
 				for (int col = 0; col < 7; col++)
 				{
-					int lc;
-					if (m_smoothscr)
+					uint16_t *const dest = &bitmap.pix(y * 8 + row, x * 14 + col + 0);
+
+					// Flash or Inverse
+					if ((!(ch & 0x80) && (char_data[row] & 0x80) && m_flash) || (!(ch & 0x80) && !(char_data[row] & 0x80)))
 					{
```
cuavas, line 227: Same here – it can use the same minimised expression.

src/mame/apple/apple3_v.cpp
```diff
@@ -219,44 +209,50 @@ void apple3_state::text80(bitmap_ind16 &bitmap, const rectangle &cliprect)
...
 			for (int row = 0; row < 8; row++)
 			{
 				for (int col = 0; col < 7; col++)
 				{
 					uint16_t *const dest = &bitmap.pix(y * 8 + row, x * 14 + col + 7);
-					*dest = (char_data[row] & (1 << col)) ? fg : bg;
+
+					// Flash or Inverse
+					if ((!(ch & 0x80) && (char_data[row] & 0x80) && m_flash) || (!(ch & 0x80) && !(char_data[row] & 0x80)))
+					{
```
cuavas, line 249: And this one as well.

cuavas (changes requested):

src/mame/apple/apple3_v.cpp
robjustice: yes, that's much better.

robjustice: updated

src/mame/apple/apple3_v.cpp
robjustice: updated this as well

src/mame/apple/apple3_v.cpp
robjustice: updated

PR #11492, opened 2023-08-18
sinclair/sprinter.cpp: default layout with Turbo,NeoGS,FDD,HDD leds

holub:
Would such changes to atahle acceptable or can you suggest better place?

### DIFF
--- a/src/devices/bus/spectrum/neogs.cpp
+++ b/src/devices/bus/spectrum/neogs.cpp
@@ -81,6 +81,7 @@ public:
 		, m_view(*this, "view")
 		, m_dac(*this, "dac%u", 0U)
 		, m_sdcard(*this, "sdcard")
+		, m_neogs_led(*this, "neogs_led")
 	{ }
 
 protected:
@@ -106,6 +107,7 @@ protected:
 	memory_view m_view;
 	required_device_array<dac_word_interface, 2> m_dac;
 	required_device<spi_sdcard_sdhc_device> m_sdcard;
+	output_finder<> m_neogs_led;
 
 private:
 	TIMER_CALLBACK_MEMBER(spi_clock);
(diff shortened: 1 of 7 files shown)


cuavas:
The trouble with adding outputs in atahle is there can be more than one of them in a system, and they’ll stomp on each other’s outputs.  Eventually outputs will be per-device, but that requires `output_manager::set_value` to be phased out first, and no-one is actively working on removing uses of it.

holub:
> The trouble with adding outputs in atahle is there can be more than one of them in a system, and they’ll stomp on each other’s outputs. Eventually outputs will be per-device, but that requires `output_manager::set_value` to be phased out first, and no-one is actively working on removing uses of it.

Yeah. That's my current trouble which I'm ignoring. It's fine to have one output for master+slave but second hdd will ruin everething.

Is it possible to have reference such as `ata0:ata_busy_led`?
What options do I have? Revert ATA part in the PR?

holub:
By the way, here I exposed BUSY. I don't see a technical way to expose ACTIVE at the current state.

cuavas:
I’d rather you remove the LED from the ATA HLE thing in MAME at this point.

As an aside, ATA activity LEDs in PCs were usually connected to the ATA interface card, or the motherboard if the ATA interface was integrated in the chipset (via a 2-pin 100 mil pitch header), not the drives themselves.  (SCSI activity LEDs were connected to the drives themselves via a 2-pin header with narrower pitch.)

holub:
> I’d rather you remove the LED from the ATA HLE thing in MAME at this point.
> 
> As an aside, ATA activity LEDs in PCs were usually connected to the ATA interface card, or the motherboard if the ATA interface was integrated in the chipset (via a 2-pin 100 mil pitch header), not the drives themselves. (SCSI activity LEDs were connected to the drives themselves via a 2-pin header with narrower pitch.)

removed

PR #11493, opened 2023-08-18
Update bgfx, bx and bimg to current upstream versions

belegdol:
Update to the libraries to current upstreams:
- bgfx to c18122226e9bce55b19cae5bc630e4f25e360e1e
- bx to 198cd120e4941d8aeb677db945ef4291a1ced291
- bimg to ec02df824a763b2e2ae31e19c674ba0bc88c0695

This is needed by #11451.


belegdol:
Feedback from #11451 from @cuavas and @rb6502: as this drops opengl support on macOS, it should not be merged prior to January 2024.

I will investigate if the commit dropping support can be reverted.

belegdol:
Could someone with access to macOS please test whether opengl renderer for bgfx is working with the two commits reverted? Thanks!

rb6502:
The macOS CI build failed, so that would be an important first step :)

belegdol:
There is a PR fixing this, I will resync once it is merged:
https://github.com/bkaradzic/bgfx/pull/3116
If you have a new enough macOS this failure should not affect you.

cuavas:
Does this PR preserve OpenGL support for macOS, or does it remove it like the other one does/did?

cuavas:
At any rate, this should probably be held off until after freeze.

belegdol:
> Does this PR preserve OpenGL support for macOS, or does it remove it like the other one does/did?

The two reverts, [9f893d0](https://github.com/mamedev/mame/pull/11493/commits/9f893d06fd28e761fd528f53392eae494f6fb73d) and [467e2ff](https://github.com/mamedev/mame/pull/11493/commits/467e2ff8f9a7c8ec830a5dd46cd2f14c6a45663b) are supposed to add it back, as long as it was merely removed for being old and the subsequent commits still work. I do not have access to a mac so I cannot test it myself.


> At any rate, this should probably be held off until after freeze.

Definitely. It should not be merged until the PR fixing building on macOS earlier than 13.0 is merged upstream.

cuavas:
Is the macOS CI failure real this time?  It seems to be complaining about using a 10.13 API feature with 10.12 deployment target.

My preferred strategy for getting this in would be to merge it fairly soon after a freeze, then after confirming that it isn’t breaking the major targets (including testing OpenGL and Metal backends on macOS), worry about the other PR with Wayland support for Linux (unless @rb6502 has other ideas).  We’re only a few days away from freeze for the August release, so if you want to get this in for the September cycle, it needs to be sorted out soon.

Also, if you could trigger the shader recompilation CI job, grab the artefact, and check in the rebuilt bgfx shaders, that would be helpful.

belegdol:
> Is the macOS CI failure real this time? It seems to be complaining about using a 10.13 API feature with 10.12 deployment target.

This is supposed to be fixed by https://github.com/bkaradzic/bgfx/pull/3116 which unfortunately has not been merged yet.

> 
> My preferred strategy for getting this in would be to merge it fairly soon after a freeze, then after confirming that it isn’t breaking the major targets (including testing OpenGL and Metal backends on macOS), worry about the other PR with Wayland support for Linux (unless @rb6502 has other ideas). We’re only a few days away from freeze for the August release, so if you want to get this in for the September cycle, it needs to be sorted out soon.
> 

If the PR fixing building with older XCode does not get merged soon, I could theoretically just add the PR patch manually. I would prefer not to just in case the final code ends up being different. Next time we do b* libs update noone will remember why our code is different especially given that the individual commits get squashed when a PR is merged.

> Also, if you could trigger the shader recompilation CI job, grab the artefact, and check in the rebuilt bgfx shaders, that would be helpful.

Not a problem, will do it once it is clear which snapshot we are going to ship.

belegdol:
It unfortunately appears that the PR does not fix the build failure we are hitting (XCode 14 building for macOS). So waiting is the only option at the moment.

belegdol:
Since there was no progress on the upstream PR and further upstream commits have worsened the pre-macOS 13 situation further, I have decided to revert the offending commits myself. I also added a README file explaining the deviations so that these reverts can be undone once we no longer care for pre-Ventura.
Please note that I have no way of testing macOS myself.

cuavas:
@rb6502 should I merge this?  Will you be able to test macOS with the OpenGL and Metal backends?

cuavas:
On Windows, most things work as well as they did previously.  The main things that are wonky are:
* None of the xBR Multipass chains seem to work with any backend.  They all produce a black screen.
* None of the xBR chains work properly with the `opengl` backend.  Most of them produce a black screen or a monochrome image.
* Attempting to use the xBR-mlv4 Multipass + Dilation chain with the `d3d12` or `vulkan` backend freezes my PC so had I need to hold down the power button to turn it off – no response to any other input.  I think this is worse than previous versions of bgfx, IIRC I could run through all the chains in the past and the worst thing that happened was bad video.

belegdol:
> On Windows, most things work as well as they did previously. The main things that are wonky are:
> 
>     * None of the xBR Multipass chains seem to work with any backend.  They all produce a black screen.
> 
>     * None of the xBR chains work properly with the `opengl` backend.  Most of them produce a black screen or a monochrome image.
> 
>     * Attempting to use the xBR-mlv4 Multipass + Dilation chain with the `d3d12` or `vulkan` backend freezes my PC so had I need to hold down the power button to turn it off – no response to any other input.  I think this is worse than previous versions of bgfx, IIRC I could run through all the chains in the past and the worst thing that happened was bad video.

Thank you for testing. This is unfortunate. I am travelling this week so I will not be able to check this. Truth be told, shader code is black magic to me so if this is an unacceptable regression we might need to revert.

cuavas:
@angelosa is reporting crashes on Ubuntu Linux with default configuration:
```
   from /lib/x86_64-linux-gnu/libnvidia-glcore.so.470.199.02
(gdb) bt
#0  0x00007ffff13008b7 in ?? ()
   from /lib/x86_64-linux-gnu/libnvidia-glcore.so.470.199.02
#1  0x000055556630af5b in bgfx::vk::RendererContextVK::submit(bgfx::Frame*, bgfx::ClearQuad&, bgfx::TextVideoMemBlitter&) ()
#2  0x000055556628e2d5 in bgfx::Context::renderFrame(int) ()
#3  0x00005555662984ca in bgfx::Context::renderThread(bx::Thread*, void*) ()
#4  0x00005555663a5493 in bx::ThreadInternal::threadFunc(void*) ()
#5  0x00007ffff7e38609 in start_thread (arg=<optimized out>)
    at pthread_create.c:477
#6  0x00007ffff64c2133 in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

It appears to be using Vulkan by default with NVIDIA binary drivers.  He also said that using `-bgfx_backend opengl` results in a segmentation fault on exit.

I can remove the xBR Multipass shaders if that’s the only major problem, but crashes on Linux are somewhat concerning.

belegdol:
Crashes I can look into next week.

cuavas:
Since the major motivation for this is to support Wayland on Linux if it’s crashy on Linux it defeats the whole purpose.

angelosa:
> He also said that using -bgfx_backend opengl results in a segmentation fault on exit.

With MAME symbols build (?), on application exit (including on MAME info/nag screens) and `-verbose`:
```gdb
No more renderers - shutting down BGFX library

Thread 2 "decosym" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeca9a700 (LWP 197320)]
0x00007fffec28aa60 in ?? ()
(gdb) bt
#0  0x00007fffec28aa60 in ?? ()
#1  0x00007ffff7e375a1 in __nptl_deallocate_tsd () at pthread_create.c:301
#2  0x00007ffff7e3862a in __nptl_deallocate_tsd () at pthread_create.c:256
#3  start_thread (arg=<optimized out>) at pthread_create.c:488
#4  0x00007ffff64c2133 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

```

```
#
# CORE SCREEN OPTIONS
#
brightness                0.95
contrast                  1.0
gamma                     0.95
pause_brightness          0.65
effect                    none
video                     bgfx
numscreens                1
window                    1
maximize                  1
waitvsync                 0
syncrefresh               0
monitorprovider           auto

# BGFX POST-PROCESSING OPTIONS
#
bgfx_path                 bgfx
bgfx_backend              opengl
bgfx_debug                0
bgfx_screen_chains        unfiltered
bgfx_shadow_mask          slot-mask.png
bgfx_lut                  
bgfx_avi_name             auto

#
# OpenGL-SPECIFIC OPTIONS
#
gl_forcepow2texture       0
gl_notexturerect          0
gl_vbo                    1
gl_pbo                    1
gl_glsl                   0
gl_glsl_filter            1
```

rb6502:
This looks good on my Mac running the latest Sonoma beta, with both OpenGL and Metal backends.  It's probably going to break my unofficial patch that re-enabled back compatibility to 10.13 though, which as I noted earlier I promised we wouldn't do before January.

cuavas:
If this can't be patched to work on macOS 10.13, or it's too crashy on Ubuntu-like Linux distros, I'll just revert it. Otherwise I'll remove the xBR Multipass chains.

belegdol:
Given the plethora of issues with this update: would it be worth cherry-picking the minimum number of patches needed to get Wayland working?

cuavas:
> Given the plethora of issues with this update: would it be worth cherry-picking the minimum number of patches needed to get Wayland working?

I don’t know how practical that would be, given how much code churn there seems to be on upstream bgfx.

belegdol:
> > Given the plethora of issues with this update: would it be worth cherry-picking the minimum number of patches needed to get Wayland working?
> 
> I don’t know how practical that would be, given how much code churn there seems to be on upstream bgfx.

Let me see next week, maybe it will work. For now please feel free to revert.

PR #11494, opened 2023-08-19
heathkit/tlb.cpp: Convert tlb into a slot device.

mgarlanger:
Change how the h19 and h89 is using the terminal logic board(tlb) and it's variants. It now has a connector/device_single_card_slot_interface, to allow user to pick which variant of the board is used at run-time. This simplifies the h19, and removes the unneeded clones. This also allows the h89 to use any of the terminal logic boards, just like a real system would be able to.

This PR also changes the interrupt controller to be a slot device, so it can be picked between the original logic, and the updated logic when the h37 controller is installed. If user wants to use the Z-89-37 soft-sectored controller, they must use the updated h37 interrupt controller.


src/mame/heathkit/intr_cntrl.h
```diff
@@ -73,4 +92,51 @@ class z37_intr_cntrl : public heath_intr_cntrl
...
+	IRQ_CALLBACK_MEMBER(irq_callback);
+
+	auto irq_line_cb() { return m_irq_line.bind(); }
+	virtual uint8_t get_instruction();
+
+	void raise_irq(int state) { m_irq_line(state); }
+
+	void set_irq(int state) { if (m_cntrl) { m_cntrl->set_irq(state); }};
+	void set_drq(int state) { if (m_cntrl) { m_cntrl->set_drq(state); }};
+	void block_interrupts(uint8_t data) { if (m_cntrl) { m_cntrl->block_interrupts(data); }};
```
cuavas, line 126: Is there a reason `set_irq_level` and `raise_irq` are virtual?  They shouldn’t need to be since they’re just passed on to the connected device.

src/mame/heathkit/intr_cntrl.h
```diff
@@ -73,4 +92,51 @@ class z37_intr_cntrl : public heath_intr_cntrl
...
+
+	heath_intr_socket(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock = 0);
+	virtual ~heath_intr_socket();
+
+	// required operation overrides
+	virtual void set_irq_level(uint8_t level, int state) { if (m_cntrl) { m_cntrl->set_irq_level(level, state); }};
+
+	IRQ_CALLBACK_MEMBER(irq_callback);
+
+	auto irq_line_cb() { return m_irq_line.bind(); }
```
cuavas, line 119: Since `irq_line_cb` is a configuration function as opposed to a live signal function, please put it above the live signal functions.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -168,14 +174,58 @@ void z37_intr_cntrl::device_start()
...
+	device_heath_intr_interface *const connected = dynamic_cast<device_heath_intr_interface *>(get_card_device());
+	if (connected)
+	{
+		return connected;
+	}
+	else
+	{
+		return nullptr;
+	}
+}
```
cuavas, line 231: You shouldn’t need this – `device_single_card_slot_interface` already has the glue to downlcast the connected device in its implementation of `get_card_device`.

src/mame/heathkit/tlb.cpp
```diff
@@ -1174,3 +1182,37 @@ ioport_constructor heath_gp19_tlb_device::device_input_ports() const
...
+	device_heath_tlb_card_interface *const connected = dynamic_cast<device_heath_tlb_card_interface *>(get_card_device());
+	if (connected)
+	{
+		return connected;
+	}
+	else
+	{
+		return nullptr;
+	}
+}
```
cuavas, line 1218: This also shouldn’t be necessary when you’re deriving from `device_single_card_slot_interface`.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -71,24 +91,7 @@ uint8_t heath_intr_cntrl::get_instruction()
...
-		}
-		level++;
-		mask <<= 1;
-	}
-
-	if (level > 7)
-	{
-		logerror("bad level: %d\n", level);
-	}
+	uint8_t level = ffs(m_intr_lines) - 1;
```
cuavas, line 94: `ffs` isn’t available on Windows.  You’ll need the loop for now.

mgarlanger: That's disappointing, builds showed it was fine with Windows and CLANG, just not Windows and GCC.

src/mame/heathkit/tlb.cpp
```diff
@@ -93,6 +94,14 @@ DEFINE_DEVICE_TYPE(HEATH_ULTRA, heath_ultra_tlb_device, "heath_ultra_tlb", "Heat
 DEFINE_DEVICE_TYPE(HEATH_GP19, heath_gp19_tlb_device, "heath_gp19_tlb", "Heath Terminal Logic Board plus Northwest Digital Systems GP-19")
 
 
+
+device_heath_tlb_card_interface::device_heath_tlb_card_interface(const machine_config &mconfig, device_t &device) :
+	device_interface(device, "heathtlbdevice")
+{
+	m_slot = dynamic_cast<heath_tlb_connector *>(device.owner());
+}
```
cuavas, line 102: If you move setting `m_slot` to the initialiser list, you can make it a `const` pointer.

src/mame/heathkit/intr_cntrl.h
```diff
@@ -13,32 +13,49 @@
...
+	virtual void set_irq(int state) {};
+	virtual void block_interrupts(uint8_t data) {};
+
+protected:
+	// construction/destruction
+	device_heath_intr_interface(const machine_config &mconfig, device_t &device);
+
+	virtual uint8_t get_instruction() = 0;
+
+	heath_intr_socket  *m_socket;
```
cuavas, line 34: You can make this pointer `const`, too.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -10,17 +10,34 @@
 
 #include "intr_cntrl.h"
 
+#include <strings.h>
```
cuavas, line 13: Better to avoid pulling in this header.

src/mame/heathkit/intr_cntrl.cpp
```diff
@@ -168,14 +180,45 @@ void z37_intr_cntrl::device_start()
...
+
+IRQ_CALLBACK_MEMBER(heath_intr_socket::irq_callback)
+{
+	// assume NO-OP
+	uint8_t instr = 0x00;
+
+	if (m_cntrl)
+	{
+		 instr = m_cntrl->get_instruction();
+	}
```
cuavas, line 221: There’s a stray extra space in the indent in here.

src/mame/heathkit/intr_cntrl.cpp
cuavas: That’s only because the Windows clang build only includes a minimal set of systems.  It’s mostly for testing that the Windows-specific code in src/osd can be compiled with clang (the rest of MAME is built with clang in Linux and macOS CI jobs).  This code would still fail to build with MinGW clang on Windows.

src/mame/heathkit/tlb.h
```diff
@@ -190,4 +214,61 @@ DECLARE_DEVICE_TYPE(HEATH_SUPER19, heath_super19_tlb_device)
...
+
+	device_heath_tlb_card_interface *m_tlb;
+};
+
+
+// device type definition
+DECLARE_DEVICE_TYPE(HEATH_TLB_CONNECTOR, heath_tlb_connector)
+
+extern template class device_finder<heath_tlb_connector, false>;
+extern template class device_finder<heath_tlb_connector, true>;
```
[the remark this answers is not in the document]
mgarlanger: I wasn't sure what the lines were for, and I think I did pull this from a2bus, since it was at the top of the list as I was searching for examples to see how to make this into a slot device. I'll remove them.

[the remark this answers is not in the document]
cuavas: The only device header I can think of off the top of my head that does something like this in the header is the floppy drive connector abstraction.  It’s unorthodox in several ways.

cuavas:
With the additional slot options and the dependency between the floppy controller and interrupt controller, these systems are becoming more complex to use.  Please consider adding a guide on the wiki to help other developers and would-be users.  You can add another item here: https://wiki.mamedev.org/index.php/System-Specific_Setup_and_Information

mgarlanger:
> With the additional slot options and the dependency between the floppy controller and interrupt controller, these systems are becoming more complex to use. Please consider adding a guide on the wiki to help other developers and would-be users. You can add another item here: https://wiki.mamedev.org/index.php/System-Specific_Setup_and_Information

Sounds good. I'll look at adding a guide soon. Currently, the defaults for the slot options provide the best experience, so hopefully people won't mess with them unless they know what they are doing.

PR #11498, opened 2023-08-20
barcrest/mpu1.cpp: Added MPU2 support

SomeRandomGuyIdk:
- Added MPU2 support to mpu1.cpp, 14 new playable sets w/ barebones layouts
- Removed separate MPU2 driver, as it directly builds on MPU1
- Minor corrections in jpmsru.cpp and bfm_blackbox.cpp
- Corrected overdumped ROMs for m_gndgit
- Made the existing SRU layouts larger


src/mame/barcrest/mpu1.cpp
```diff
@@ -35,85 +53,233 @@
...
 		m_dac(*this, "dac"),
-		m_samples(*this, "samples")
+		m_samples(*this, "samples"),
+		m_pia3(*this, "pia3"),
+		m_pia4(*this, "pia4"),
+		m_nvram(*this, "nvram", 0x100, ENDIANNESS_BIG)
 	{ }
 
-	DECLARE_INPUT_CHANGED_MEMBER(coin_input);
+	void mpu12_base(machine_config &config);
```
cuavas, line 92: If this is an abstract class, this constructor and member function should be in the `protected` section.

src/mame/barcrest/mpu1.cpp
```diff
@@ -35,85 +53,233 @@
...
-	output_finder<13> m_lamps;
-	required_device_array<em_reel_device, 4> m_reels;
+	output_finder<56> m_lamps;
 	required_device<dac_1bit_device> m_dac;
 	required_device<fruit_samples_device> m_samples;
+
+	// MPU2 devices
+	optional_device<pia6821_device> m_pia3;
+	optional_device<pia6821_device> m_pia4;
+	memory_share_creator<uint8_t> m_nvram;
```
cuavas, line 154: If these are for the MPU2, can you move them out of the base class and make them `required_device` rather than `optional_device`?  It’s better to avoid `optional_device` if it isn’t really necessary.

src/mame/barcrest/mpu1.cpp
```diff
@@ -35,85 +53,233 @@
...
+
+class mpu1_state : public mpu12_base_state
+{
+public:
+	mpu1_state(const machine_config &mconfig, device_type type, const char *tag) :
+		mpu12_base_state(mconfig, type, tag),
+		m_reels(*this, "emreel%u", 1U)
+	{ }
+
+	void add_em_reels(machine_config &config, int symbols, attotime period);
```
cuavas, line 165: `add_em_reels` should be in the `protected` section.

cuavas (commented):
This mostly looks OK.  `awp_draw_reel` really needs to be replaced at some point, but that’s old technical debt that’s in no way your fault, so you don’t need to worry about it for this PR.

src/mame/barcrest/mpu1.cpp
SomeRandomGuyIdk: Fixed.

src/mame/barcrest/mpu1.cpp
SomeRandomGuyIdk: Fixed.

src/mame/barcrest/mpu1.cpp
SomeRandomGuyIdk: Fixed.

PR #11131, opened 2023-04-20
namcos10: New dumps/redumps, support CD drive and Taiko I/O

987123879113:
- Added new dump of Seishun-Quiz Colorful High School (CHS1 Ver.A) [Guru]
- Added redump of Point Blank 3 from Guru. This was initialized on the machine before redumping. The redump confirmed that the old dump was a good dump so the only thing that changed in this was the EEPROM section data (0x4200-0x8400).
- Tsukkomi Yousei Gips Nice Tsukkomi doesn't appear to have a CD after all after more testing by Guru
- Added some notes about the MEM(P3) board

For taiko4:
- Taiko no Tatsujin 4 decrypter code by Samuel Neves, Peter Wilhelmsen. The message I received makes me think they're trying to find a different method that doesn't rely on ignoring the IV/SBOX so I called it preliminary, but it's fully working from what I've tested. However the code may change again in the future.
- Implemented CD-ROM support for Taiko and was able to document some more interrupt and DMA registers with it
- Added a new command to t10mmc that Taiko relies on to read data. The actual command itself is much more complex (it's apparently meant to be a "reads everything" type of command that can do everything from subchannel data to headers to plain data and more) but I only implemented the parts I could test.

Lastly, I added a way to force the PSX SPU to use `STREAM_SYNCHRONOUS` when requested. I think probably any game using PSX hardware that relies on strict SPU IRQ timings for DMAs will probably benefit from `STREAM_SYNCHRONOUS` but I also understand it's not an ideal solution so I made it a configurable flag you must set to use. For streaming BGMs in Taiko no Tatsujin, the game is only requesting 256 bytes total to be DMA'd from PSX memory into the SPU at a time and only seems to send more when the IRQ is triggered. Sound effects are DMA'd in at 1024 bytes at a time so don't seem to have any issues. I believe it's running into an issue when playing BGMs where there's not enough data to fill the stream buffer at the rate requested which causes lots of audio glitches and the audio to not play at the correct speed. I think because `STREAM_SYNCHRONOUS` makes the sound stream only update 1 sample at a time, IRQs can be triggered and handled by the game sooner than if the IRQ is triggered sometime during a larger stream update.

PR #11499, opened 2023-08-21
t10mmc: Implement most features of T10MMC_CMD_READ_CD

987123879113:
Follow up to code I implemented previously for Namco System 10 Taiko games in https://github.com/mamedev/mame/pull/11131. I found a usecase that requires data other than just the user data so I implemented the other fields to the best of my understanding.

Namco System 12's Truck Kyosokyoku (`truckk`) uses the READ CD command requesting subheader + user data and uses the submode byte of the subheader to sift the data stream's video and audio data and handles them accordingly so it's not optional. `truckk` isn't playable yet but this change alone is fairly large due to the nature of the command itself so I wanted to throw it out as its own PR. I've tested the Taiko games to make sure they're still working, and the `truckk` usage is working in my WIP code.

Reference: mmc3r10g.pdf, PDF pages 192 through 200 (physical numbered pages 156 through 164) describe the READ CD command. Table 153 I think is the most useful for reviewing this PR:
image


src/devices/machine/t10mmc.cpp
```diff
@@ -539,58 +541,220 @@ void t10mmc::ExecCommand()
...
+				|| (read_cd_flags == (T10MMC_READ_CD_FIELD_SYNC | T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_SUBHEADER | T10MMC_READ_CD_FIELD_ECC));
+
+			// Mode 2 form 1/2 sectors have additional restrictions that CDDA, mode 1, and mode 2 formless tracks don't
+			if (track_type == cdrom_file::CD_TRACK_MODE2_FORM1 || track_type == cdrom_file::CD_TRACK_MODE2_FORM2 || mode2_form > 0)
+			{
+				is_illegal_combo |= (read_cd_flags == (T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_USER_DATA))
+					|| (read_cd_flags == (T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_USER_DATA | T10MMC_READ_CD_FIELD_ECC))
+					|| (read_cd_flags == (T10MMC_READ_CD_FIELD_SYNC | T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_USER_DATA))
+					|| (read_cd_flags == (T10MMC_READ_CD_FIELD_SYNC | T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_USER_DATA | T10MMC_READ_CD_FIELD_ECC));
+			}
```
cuavas, line 641: Using `|=` on a `bool` is kinda bad form as it technically involves implicit conversion to `int` and back.  You don’t even need the extra tests if you already have an illegal combo, so you could do it like:

```c++
if (!is_illegal_combo && (track_type == cdrom_file::CD_TRACK_MODE2_FORM1 || track_type == cdrom_file::CD_TRACK_MODE2_FORM2 || mode2_form > 0))
	is_illegal_combo = (read_cd_flags == (T10MMC_READ_CD_FIELD_HEADER | T10MMC_READ_CD_FIELD_USER_DATA))
...
```

cuavas (commented):
This looks sane enough to me.  I’ll leave it open a bit so other people (e.g. @galibert and @rb6502) can have a bit of a look if they want.

src/devices/machine/t10mmc.cpp
987123879113: Funnily enough that's how I had it originally but wanted to reduce the amount of already huge conditionals in if statements in that area so I rewrote it to how it is now. I don't have a particular preference so I'll revert that specific change. Also noted about the `|=` on bools, hadn't considered that part.

PR #11500, opened 2023-08-21
sh2: Refactor code into SH2 base CPU class + separate CPU devices

987123879113:
sh2.cpp doesn't have separation between the SH-2 CPU core code and the on-chip peripherals, so I separated the core code to make adding new SH-2 CPUs easier.

This is only a draft PR for now as I haven't done a lot of testing to make sure nothing is broken, but I would like feedback on the overall refactoring work if there's something noticeably wrong with how I'm going about this refactor.

I split off sh2.cpp/sh2comn.cpp into separate CPU devices with non-generic names into their own separate files, and tried to make some small refactoring changes to allow each individual CPU to handle things like IRQs in their own way as needed (like the vector mode flag requirement for only SH7604) but have otherwise tried to keep the actual logic the same. Unused variables and defines and such were deleted when found for a small bit of cleanup. I also tried to credit authors in the headers for the separated CPU cores when they were obvious but the code dates back pre-git so I gave up for now. Feel free to let me know if someone should or shouldn't be included in any of the specific files.

What inspired this refactor/what I'm trying to fix with this refactor: While trying to implement a SH-2 CPU with a completely different set of non-matching peripherals, I quickly ran into many issues such as clashing peripheral variable names (of different type sizes especially), `device_start` registering a bunch of save items that would never be used (and also causing name collisions), creation of unnecessary timers, unnecessary callbacks, issues with the address mask used for the existing SH-2 cores (for example it's impossible to set up a DMA to read data into a on-chip serial port which is a use case I was seeing in my CPU core, due to the source/dest addresses being masked and in general all reads/writes mask using `SH12_AM`), vector mode/m_vecmd not being required in the exception handler, etc.


src/devices/cpu/sh/sh2.cpp
```diff
@@ -12,269 +12,109 @@
...
-
-    20020218 Added save states (mokona@puupuu.org)
-
- *****************************************************************************/
 
 #include "emu.h"
+#include "cpu/drcumlsh.h"
 #include "sh2.h"
-#include "sh2comn.h"
 #include "sh_dasm.h"
```
cuavas, line 19: Module headers before library headers – `cpu/drcumlsh.h` should come after `sh2.h` and `sh_dasm.h` (if that doesn’t work, on of the headers needs to `#include` it).

src/devices/cpu/sh/sh2.cpp
```diff
@@ -12,269 +12,109 @@
...
+				ptr++;
+			return ptr;
+		};
+	else
+		m_prptr = [this](offs_t address) -> const void * {
+			const u16 *ptr = static_cast<u16 *>(m_cache32.read_ptr(address & ~3));
+			if(address & 2)
+				ptr++;
+			return ptr;
+		};
```
cuavas, line 82: `reinterpret_cast` is probably more appropriate than `static_cast` here.

src/devices/cpu/sh/sh7021.cpp
```diff
@@ -0,0 +1,162 @@
...
+	std::fill(std::begin(m_sh7021_regs), std::end(m_sh7021_regs), 0);
+
+	m_dmaor = 0;
+
+	for (int i = 0; i < 4; i++) {
+		m_dma[i].sar = 0;
+		m_dma[i].dar = 0;
+		m_dma[i].tcr = 0;
+		m_dma[i].chcr = 0;
+	}
```
cuavas, line 39: Looks like K&R sneaking into an Allman file?

src/devices/cpu/sh/sh7021.h
```diff
@@ -0,0 +1,54 @@
...
+class sh2a_sh7021_device : public sh2_device
+{
+public:
+	// construction/destruction
+	sh2a_sh7021_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override;
+	virtual void device_reset() override;
```
cuavas, line 20: Can we please lose this copy/pasted comment?  If you want to label the `device_t` implementation, make the comment read “// device_t implementation”.

src/devices/cpu/sh/sh7604.h
```diff
@@ -0,0 +1,283 @@
...
+	template <typename... T> void set_ftcsr_read_callback(T &&... args) { m_ftcsr_read_cb.set(std::forward<T>(args)...); }
+
+	void sh2_notify_dma_data_available();
+
+protected:
+	sh2_sh7604_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, int cpu_type,address_map_constructor internal_map, int addrlines);
+
+	// device-level overrides
+	virtual void device_start() override;
+	virtual void device_reset() override;
```
cuavas, line 41: Same here, that comment was copy/pasted too many times already.

src/devices/cpu/sh/sh7604.h
```diff
@@ -0,0 +1,283 @@
...
+	void rtcnt_w(offs_t offset, uint32_t data, uint32_t mem_mask = ~0);
+
+	virtual void set_frt_input(int state) override;
+
+	// SCI
+	uint8_t m_smr = 0, m_brr = 0, m_scr = 0, m_tdr = 0, m_ssr = 0;
+	// FRT / FRC
+	uint8_t m_tier = 0, m_ftcsr = 0, m_frc_tcr = 0, m_tocr = 0;
+	uint16_t m_frc = 0;
+	uint16_t m_ocra = 0, m_ocrb = 0, m_frc_icr = 0;
```
cuavas, line 201: Can you please move initialisation for data members that aren’t part of mini nested structures/classes to the constructor initialiser list?  Having them in the header increases compilation overhead and fragility.

src/devices/cpu/sh/sh7604.h
```diff
@@ -0,0 +1,283 @@
...
+	{
+		if(ACCESSING_BITS_0_7)
+		{
+			uint8_t old;
+			old = m_dmaor & 0xf;
+			m_dmaor = (data & ~6) | (old & m_dmaor & 6); // TODO: should this be old & data & 6? bug?
+			sh2_dmac_check(0);
+			sh2_dmac_check(1);
+		}
+	}
```
cuavas, line 164: Do all these helpers need to be inlined here?  Are they only used in sh7604.cpp?  If they’re only used there, they can be moved to that file.  If they need to be instantiated from elsewhere, can they be moved to a .ipp file?

src/devices/cpu/sh/sh7604.cpp
```diff
@@ -1,25 +1,310 @@
 // license:BSD-3-Clause
-// copyright-holders:R. Belmont
+// copyright-holders:Juergen Buchmueller, R. Belmont, Angelo Salese, ...
```
cuavas, line 2: The ellipsis here doesn’t look good…

cuavas (commented):
I like this in principle.  The SH2 device shouldn’t implement a particular set of peripherals given how many chips used the core.

src/devices/cpu/sh/sh7021.cpp
987123879113: I threw in some additional save state stuff and initialization in some of these split off CPUs so that's probably my fault. 😇

src/devices/cpu/sh/sh7604.cpp
987123879113: The ones with ellipsis are the credits I want to go back to double check if I missed any obvious (major) contributors. I'm not sure how strict I should be on the copyright holders list. Would it be better to just let anyone who contributed add themselves if they wanted at this point?

src/devices/cpu/sh/sh7604.h
987123879113: They are only used by SH7604 and I don't think they really need to be in the header. I can move them out into sh6704.cpp.

src/devices/cpu/sh/sh7604.cpp
cuavas: Well, you know copyright rules – it has to be a substantial, creative contribution for copyright claims.  I wouldn’t worry too much about finding everyone who’s made any contribution at all.

src/devices/cpu/sh/sh7604.h
987123879113: Yeah, sounds good to me. I debated with myself on whether I should make that change in this PR since I vaguely remember you commenting something similar in the past on someone else's commit.

src/devices/cpu/sh/sh7604.cpp
987123879113: Ok. MAME's been using git for a while now so if someone within the past decade or 2 has made any substantial contributions and wants their name added then it should be easily verifiable at least.

987123879113: Ok, changes made to the credits to remove you from sh7604 and made you the copyright holder for sh7021 and sh7032.

987123879113:
No worries, wasn't trying to get this in for this month's release. Primarily just wanted the feedback to make sure this was a step in the right direction. 

I've tested a number of machines and everything is working fine so far from what I can tell, so I don't think the refactoring introduced any unexpected bugs. I tested a few games on the machines with a lot of games, and all of the games on machines with few games. I'll test some games on the remaining machines tomorrow.

- [x] src/mame/seibu/feversoc.cpp
- [x] src/mame/sega/coolridr.cpp
- [ ] src/mame/sega/saturn.cpp
- [ ] src/mame/sega/saturn_cdb.cpp
- [x] src/mame/sega/stv.cpp
- [x] src/mame/psikyo/psikyo4.cpp
- [x] src/mame/psikyo/psikyosh.cpp
- [x] src/mame/misc/hideseek.cpp (broken, but is marked as NOT_WORKING and is really a SH7045 and not SH7604 so was never going to work)
- [x] src/mame/kaneko/suprnova.cpp
- [x] src/mame/dataeast/deco_mlc.cpp
- [ ] src/mame/casio/casloopy.cpp (not sure how to test, the SH7021 CPU is already being worked on seperately so not concerned)
- [x] src/mame/capcom/cps3.cpp
- [ ] src/mame/shared/mega32x.cpp

angelosa:
- `sega/saturn_cdb.cpp` is a stub, it needs the SH1 serial to proceed and I don't even think it's accessible atm;
- `misc/hideseek.cpp` lacks the internal ROM dump;

987123879113:
Thanks for the info. Yeah, looks like saturn_cdb isn't even accessible right now from the looks of it. I did some quick testing of the other remaining consoles and things were booting (BIOS, some games) so I think things are working as expected.

PR #11503, opened 2023-08-23
apple/apple3.cpp: Add keyboard 2spd autorepeat

robjustice:
Adding the keyboard auto repeat support. The Apple /// has a faster repeat speed that is invoked with the closed apple key. This is based on the Apple2 code.

### DIFF
--- a/src/mame/apple/apple3.cpp
+++ b/src/mame/apple/apple3.cpp
@@ -90,6 +90,10 @@ void apple3_state::apple3(machine_config &config)
 	m_ay3600->shift().set(FUNC(apple3_state::ay3600_shift_r));
 	m_ay3600->control().set(FUNC(apple3_state::ay3600_control_r));
 	m_ay3600->data_ready().set(FUNC(apple3_state::ay3600_data_ready_w));
+	m_ay3600->ako().set(FUNC(apple3_state::ay3600_ako_w));
+
+	/* repeat timer */
+	TIMER(config, "repttimer").configure_generic(FUNC(apple3_state::ay3600_repeat));
 
 	/* slot bus */
 	A2BUS(config, m_a2bus, 0);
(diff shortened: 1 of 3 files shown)


src/mame/apple/apple3.cpp
```diff
@@ -90,6 +90,10 @@ void apple3_state::apple3(machine_config &config)
 	m_ay3600->shift().set(FUNC(apple3_state::ay3600_shift_r));
 	m_ay3600->control().set(FUNC(apple3_state::ay3600_control_r));
 	m_ay3600->data_ready().set(FUNC(apple3_state::ay3600_data_ready_w));
+	m_ay3600->ako().set(FUNC(apple3_state::ay3600_ako_w));
+
+	/* repeat timer */
+	TIMER(config, "repttimer").configure_generic(FUNC(apple3_state::ay3600_repeat));
```
cuavas, line 96: Please use `m_repttimer` in place of the literal tag here.  It’s less error-prone.

src/mame/apple/apple3_m.cpp
```diff
@@ -1294,6 +1294,46 @@ void apple3_state::ay3600_data_ready_w(int state)
...
 
+void apple3_state::ay3600_ako_w(int state)
+{
+	m_anykeydown = (state == ASSERT_LINE) ? true : false;
+
+	if (m_anykeydown)
+	{
+		m_repeatdelay = 5;
+		m_repttimer->adjust(attotime::from_hz(10));
+	}
```
cuavas, line 1305: Wouldn’t it be simpler to just set the initial delay of half a second here rather than counting five ticks?  Also, do you want to cancel the timer when AKO is deasserted?

src/mame/apple/apple3.cpp
robjustice: ok, updated

src/mame/apple/apple3_m.cpp
robjustice: Ahh, yes, that's a result of me copying it over and then adding the faster speed later and not thinking it over again. It's much cleaner coded like that.  I had thought no need to cancel it, but it makes it nice and clean so have added it. Thanks for your patience working through these updates with me.

cuavas: No problem.  Thanks for following up.

PR #11505, opened 2023-08-25
apple/apple3_m.cpp: Added cpu clock changing for video raster scanning

robjustice:
This updated is based on the schematic and logic analyser traces to confirm behavior on real hardware. The CPU speed is down changing to 1MHz when reading the video memory, and then back to 2MHz during the blanking periods. The video screen can also be turned off, resulting in 2MHz speed running full time.

The other thing impacting the CPU speed is slowdowns for dram refresh cycles. There is a logic rom that takes input from the video counters to generate the refresh signal, it's not a simple asymmetric clock signal. I have used an average frequency measured from a logic analyser trace for now for the resultant higher clock speed.

With these changes in, the game Atomic Defence speech is very close in sound now to the real hardware. Previously it was running much faster in MAME. And the smooth scrolling intro is scrolling at the same rate now.

### DIFF
--- a/src/mame/apple/apple3_m.cpp
+++ b/src/mame/apple/apple3_m.cpp
@@ -63,8 +63,14 @@
 #define ENV_ROMENABLE   (0x01)
 
 // 14M / 14, but with every 65th cycle stretched which we cannot reasonably emulate
-// 2 MHz mode probably has every 33rd cycle stretched but this is currently unproven.
 static constexpr XTAL APPLE2_CLOCK(1'021'800);
+
+// 2 MHz mode has every ~130th cycle stretched, and there are also dram refresh cycles
+// happening which slows the clock back to 1MHz for some periods.
+// There is mentioned in Apple documentation the result is ~1.8MHz
+// PH0 Measured on a logic analyser when running a loop in ram with the screen off 
+// averaged over a long capture is 1.905MHz.
+static constexpr XTAL A3_2MHZ_CLOCK(1'905'000);
 
 uint8_t apple3_state::apple3_c0xx_r(offs_t offset)
 {
@@ -153,7 +159,7 @@ uint8_t apple3_state::apple3_c0xx_r(offs_t offset)
 
 		case 0x50: case 0x51: case 0x52: case 0x53:
 		case 0x54: case 0x55: case 0x56: case 0x57:
-			m_screen->update_partial(m_screen->vpos());
+			//m_screen->update_partial(m_screen->vpos());
 			/* graphics softswitches */
 			if (offset & 1)
 				m_flags |= 1 << ((offset - 0x50) / 2);
@@ -368,7 +374,7 @@ void apple3_state::apple3_c0xx_w(offs_t offset, uint8_t data)
 
 		case 0x50: case 0x51: case 0x52: case 0x53:
 		case 0x54: case 0x55: case 0x56: case 0x57:
-			m_screen->update_partial(m_screen->vpos());
+			//m_screen->update_partial(m_screen->vpos());
 			/* graphics softswitches */
 			if (offset & 1)
 				m_flags |= 1 << ((offset - 0x50) / 2);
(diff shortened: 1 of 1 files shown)


src/mame/apple/apple3_m.cpp
```diff
@@ -153,7 +159,7 @@ uint8_t apple3_state::apple3_c0xx_r(offs_t offset)
 
 		case 0x50: case 0x51: case 0x52: case 0x53:
 		case 0x54: case 0x55: case 0x56: case 0x57:
-			m_screen->update_partial(m_screen->vpos());
+			//m_screen->update_partial(m_screen->vpos());
```
rb6502, line 162: Is there some reason these are commented out?  That will pretty much disable screen mode splitting like Atomic Defense does.

robjustice: This does need some discussion. I did comment those out there and uncommented the update_partial that was in the scan-line callback. With the A3 running slower now, it really needs the screen to be draw a line at a time like the real raster would do. Without that change, there was artifacting on the screen during the smooth scroll. 

Its taking time to move all the memory for the lines that are scrolling, and it looks like it follows down behind the raster doing these changes so it has the full time to of a screen refresh period to do this.

rb6502: That's fine as long as nobody tries to do a mid-scan line split like the A2 French Touch demos.  Go ahead and just remove those calls instead of commenting them out then and we'll be good.

happppp: Note that update_partial(scanline) is inclusive. I mean, if you're at the start or at the middle of the scanline, it will update the whole scanline. Maybe that's what you ran into.

cuavas:
Fudging CPU frequencies is an approach we’re trying to move away from for the most part.  Is it not possible to use timers to trigger CPU clock frequency changes at the right time?

src/mame/apple/apple3_m.cpp
robjustice: Thanks for explaining, understood now. I have uncommented these and kept them there now. I think I had thought it only needed to be updated in the scanline part.

robjustice:
Yes, timers should be possible. I Looked at the logic and thought that's going to take a bit of time to work through, so went with the averaged cpu for now.
Maybe once its sorted it may simplify down ok. This is the logic that generates the refresh signal:

RRFSH	= /H2 * /H3 * /H4 * /VA * /VB * /VC 
        + H2 * /H3 * /H4 * VA * /VB * /VC 
        + /H2 * H3 * /H4 * /VA * VB * /VC 
        + H2 * H3 * /H4 * VA * VB * /VC 
        + /H2 * /H3 * H4 * /VA * /VB * VC 
        + H2 * /H3 * H4 * VA * /VB * VC * /VBL 
        + /H2 * H3 * H4 * /VA * VB * VC 
        + H2 * H3 * H4 * VA * VB * VC 
        + /H3 * /H4 * /H5 * VA * /VB * /VC * /V0 * /V1 * VBL 
        + H2 * /H3 * H4 * VA * /VB * VC * /V1 
        + H2 * H3 * H4 * H5 * VA * /VB * /VC * V0 * /V1 * VBL 
        + /H2 * /H4 * /H5 * /VA * VB * /VC * V0 * /V1 * VBL 
        + /H3 * /H4 * /H5 * VA * VB * /VC * V0 * /V1 * VBL 
        + H2 * /H4 * /H5 * VA * VB * /VC * V0 * /V1 * VBL 
        + H2 * H3 * H5 * VA * VB * /VC * /V0 * V1 * VBL 
        + /H2 * /H3 * /H5 * /VA * /VB * VC * /V0 * V1 * VBL 
        + /H3 * /H4 * /H5 * VA * /VB * VC * /V0 * V1 * VBL 
        + H2 * /H3 * /H5 * VA * /VB * VC * /V0 * V1 * VBL 
        + H2 * /H3 * H4 * VA * /VB * VC * /V0 
        + H2 * /H3 * H4 * VA * /VB * VC * V0 
        + H2 * H4 * H5 * VA * /VB * VC * V0 * V1 * VBL 
        + /H2 * /H3 * /H4 * /H5 * VB * VC * V0 * V1 * VBL 
        + /H3 * /H4 * /H5 * VA * VB * VC * V0 * V1 * VBL

cuavas:
The is also one of the CPUs in MAME that’s been updated to handle memory access delays, etc. so it should be in a good state for implementing this properly.

PR #11509, opened 2023-08-26
Tidied up the Amstrad Action covertapes, promoted amsac004 to working status

ICEknigh7:
Added the actual titles that these covertapes were given and tidied up their descriptions in general.
Added "Future Publishing" as publisher.
Promoted `amsac004` to working status (others I tested still froze at some point, so might not be worth testing further until the core gets some fixes).

### DIFF
--- a/hash/cpc_cass.xml
+++ b/hash/cpc_cass.xml
@@ -41157,19 +41157,21 @@ Covertapes
 -->
 
 
-	<software name="amsac004" supported="no">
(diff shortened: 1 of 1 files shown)


cuavas (commented):
Wouldn’t the publisher be “Future Publishing” or its registered business name “Future plc”?

ICEknigh7:
Was going with what the rest of the softlist said but yeah, you're right. I'll change all those.

PR #11512, opened 2023-08-27
ppc: Don't overwrite RA in lmw instruction

987123879113:
Findings around this can be found here: https://github.com/mamedev/mame/commit/62f77513f1a4297dd15971f8909e0922918a53cc#r125639616

For 4xx series and 601 this change will skip over the register update which is what the CPU manuals says should happen. For any other flavor it will raise an illegal instruction exception since the behavior isn't defined but it does state it's an invalid form.

Only known game so far to rely on this functionality is `tropchnc` which explodes early on in boot process without this fix.

### DIFF
--- a/src/devices/cpu/powerpc/ppcdrc.cpp
+++ b/src/devices/cpu/powerpc/ppcdrc.cpp
@@ -2256,12 +2256,27 @@ bool ppc_device::generate_opcode(drcuml_block &block, compiler_state *compiler, 
 		case 0x2e:  /* LMW */
 			UML_MAPVAR(block, MAPVAR_DSISR, DSISR_IMMU(op));                                // mapvar  dsisr,DSISR_IMMU(op)
 			UML_MOV(block, mem(&m_core->tempaddr), R32Z(G_RA(op)));                  // mov     [tempaddr],ra
-			for (int regnum = G_RD(op); regnum < 32; regnum++)
+
+			if (G_RD(op) <= G_RA(op) && !(m_cap & PPCCAP_4XX) && m_flavor != PPC_MODEL_601)
 			{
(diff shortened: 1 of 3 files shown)


src/devices/cpu/powerpc/ppcdrc.cpp
```diff
@@ -2261,7 +2261,20 @@ bool ppc_device::generate_opcode(drcuml_block &block, compiler_state *compiler,
...
+				}
+				else if (!(m_cap & PPCCAP_4XX) && m_flavor != PPC_MODEL_601)
+				{
+					// RA being in the range is an invalid form
+					// 403 and 405 manual allows writes where regnum != RA || regnum == 31, regnum == RA is skipped
+					// 601 user manual lists no special rules for R31 being valid but regnum == RA is skipped
+					// 603 and above just says it's an invalid form
+					// tropchnc (403GA) is known to use this behavior
+					UML_EXH(block, *m_exception[EXCEPTION_PROGRAM], 0x80000);          // exh    exception_program,0x80000
+				}
```
cuavas, line 2277: The forms that generate an exception shouldn’t get into the loop at all.  The check should be done before the loop.

I’m pretty sure the chips that generate an exception will do so when RT is zero – from the architecture manual, “If RA is in the range of registers to be loaded, including the case in which RA=0, the instruction form is invalid.”  Even though RA=0 is treated as constant zero for D-form instructions, it’s still invalid.

I’d also guess that the 403/405/601 skip loading R0 when RA=0.

cuavas (changes requested):

src/devices/cpu/powerpc/ppcdrc.cpp
987123879113: Attempt 2. Doing it out of the loop makes a lot more sense.

PR #11513, opened 2023-08-28
heathkit/h89.cpp: add optional CPU speed upgrade modifications

mgarlanger:
Several companies offered CPU speed upgrades for the H89. These used a bit in the GPP to toggle between the standard 2.048 MHz speed and the upgraded speeds. Upgrades at least to 6 MHz were available.
Options for 4 MHz and 6 MHz were added.

Also updated how changes in GPP are detected.

Set the interrupt controller slot device to a "fixed" slot, to avoid users breaking the H89 by changing it. Once additional add-on boards are available and the soft-sectored controller is made into a slot device itself, the program will be able to properly set the interrupt controller based on what boards are installed.

### DIFF
--- a/src/mame/heathkit/h89.cpp
+++ b/src/mame/heathkit/h89.cpp
@@ -102,7 +102,7 @@ private:
 	bool m_timer_intr_enabled;
 	bool m_floppy_ram_wp;
 
-	int m_cpu_speed_multiplier;
+	uint32_t m_cpu_speed_multiplier;
 
 	// Clocks
 	static constexpr XTAL H89_CLOCK = XTAL(12'288'000) / 6;
@@ -336,11 +336,10 @@ static INPUT_PORTS_START( h89 )
 */
 
 	PORT_START("CONFIG")
-	PORT_CONFNAME(0x03, 0x00, "CPU Clock Speedup Upgrade")
+	PORT_CONFNAME(0x03, 0x00, "CPU Clock Speed Upgrade")
 	PORT_CONFSETTING(0x00, DEF_STR( None ) )
 	PORT_CONFSETTING(0x01, "2 / 4 MHz")
 	PORT_CONFSETTING(0x02, "2 / 6 MHz")
-	PORT_CONFSETTING(0x03, DEF_STR( None ) )
 INPUT_PORTS_END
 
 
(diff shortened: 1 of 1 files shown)


src/mame/heathkit/h89.cpp
```diff
@@ -100,12 +102,15 @@ class h89_state : public driver_device
 	bool m_timer_intr_enabled;
 	bool m_floppy_ram_wp;
 
+	int m_cpu_speed_multiplier;
```
cuavas, line 105: If you don’t register `m_cpu_speed_multiplier` for save states (and change it to an explicitly sized integer), you’ll get strange behaviour.  The next time “fast” mode is switched on, the multiplier will come from the config setting on the most recent reset, not the reset before the saved state was created.

src/mame/heathkit/h89.cpp
```diff
@@ -329,6 +334,13 @@ static INPUT_PORTS_START( h89 )
     PORT_DIPSETTING( 0x00, DEF_STR( Normal ) )
     PORT_DIPSETTING( 0x80, "Auto" )
 */
+
+	PORT_START("CONFIG")
+	PORT_CONFNAME(0x03, 0x00, "CPU Clock Speedup Upgrade")
+	PORT_CONFSETTING(0x00, DEF_STR( None ) )
+	PORT_CONFSETTING(0x01, "2 / 4 MHz")
+	PORT_CONFSETTING(0x02, "2 / 6 MHz")
+	PORT_CONFSETTING(0x03, DEF_STR( None ) )
```
cuavas, line 343: Better not to have two redundant settings for no speed-up.  It’s fine not to expose every bit pattern for these configuration inputs.

src/mame/heathkit/h89.cpp
```diff
@@ -386,6 +398,30 @@ void h89_state::machine_reset()
 	m_timer_intr_enabled = true;
 	m_floppy_ram_wp = false;
 
+	ioport_value const cfg(m_config->read());
+
+	// CPU clock speed
+	uint8_t selected_clock_upgrade = cfg & 0x3;
```
cuavas, line 404: You can make this `const` as well.

PR #11514, opened 2023-08-28
konami/twin16.cpp: Dumped priority PROM for The Final Round/Hard Puncher

jotego:
We have dumped and measured PROM E16 used for priority encoding from a GX870 PWB351311A board. I have added the information to the driver.


Tafoid:
@jotego If you could be so kind as to send a copy of this PROM data to the submission address please?   
Submit-email

PR #11517, opened 2023-08-29
NWK-TR, Hornet fixed [sjy96525]

sjy55525:
Racing Jam: Fix timekeeper mismatch.
Thrill Drive: Add Hand Brake Lever control for Thrill Drive (ver GN713JAB)


Hydreigon223:
Good eyes. Let this be a lesson to always delete your nvram before attempting to add these timekeeper rtc dumps. I apparently failed to do so.

PR #11522, opened 2023-09-01
cpu/z80/z80.cpp: timings tables removal; undocumented flags improvements

holub:
Credits:
    undocumented flags - Manuel Sainz de Baranda y Goñi, Peter Helcmanovsky, Patrik Rak
    tables removal - smf, happppp, Lord Nightmare

Ported from new disaster core branch:
* removal of timing tables
* undocumented flags improvements


src/mame/amstrad/amstrad_m.cpp
```diff
@@ -2174,6 +2180,18 @@ The exception is the case where none of b7-b0 are reset (i.e. port &FBFF), which
...
+	if (!machine().side_effects_disabled())
+		m_maincpu->adjust_icount(-((4 - m_maincpu->total_cycles() % 4) % 4));
+	return ((u8*)m_banks[(offset >> 13)]->base())[offset & 0x1fff];
+}
+
+void amstrad_state::amstrad_cpc_mem_w(offs_t offset, uint8_t data)
+{
+	m_maincpu->adjust_icount(-((4 - m_maincpu->total_cycles() % 4) % 4));
+	((u8*)m_banks[(offset >> 13) + 8]->base())[offset & 0x1fff] = data;
+}
```
cuavas, line 2194: The rest of this file uses unabbreviated `uint8_t` – better to be consistent with the current code.

src/devices/cpu/z80/z80.cpp
```diff
@@ -478,43 +362,44 @@ uint8_t z80_device::rm_reg(uint16_t addr)
...
+
+inline void z80_device::wm(u16 addr, u8 value)
 {
-	// As we don't count changes between read and write, simply adjust to the end of requested.
-	if(m_icount_executing != MTM) T(m_icount_executing - MTM);
-	m_data.write_byte(translate_memory_address(addr), value);
-	T(MTM);
+	data_write(addr, value);
+	T(m_memrq_cycles);
 }
```
cuavas, line 384: You’ve got an inconsistent mix of K&R and Allman style on two consecutive functions here.  Please make it use Allman style consistently.

src/devices/cpu/z80/z80.cpp
```diff
@@ -566,13 +462,13 @@ uint16_t z80_device::arg16()
...
+	m_ea = (u32)(u16)(IX + (int8_t)arg());
 	WZ = m_ea;
 }
 
 inline void z80_device::eay()
 {
-	m_ea = (uint32_t)(uint16_t)(IY + (int8_t)arg());
+	m_ea = (u32)(u16)(IY + (int8_t)arg());
 	WZ = m_ea;
 }
```
cuavas, line 473: You’re mixing the unabbreviated `int8_t` with the abbreviated `u32` and `u16` here now.

src/devices/cpu/z80/z80.cpp
```diff
@@ -910,101 +803,84 @@ inline void z80_device::neg()
...
  ***************************************************************/
 inline void z80_device::exx()
 {
-	PAIR tmp;
-	tmp = m_bc; m_bc = m_bc2; m_bc2 = tmp;
-	tmp = m_de; m_de = m_de2; m_de2 = tmp;
-	tmp = m_hl; m_hl = m_hl2; m_hl2 = tmp;
+	std::swap(m_bc, m_bc2);
+	std::swap(m_de, m_de2);
+	std::swap(m_hl, m_hl2);
```
cuavas, line 871: The standard C++ idiom for swappable types is to do this as:
```c++
	using std::swap;
	swap(m_bc, m_bc2);
	swap(m_de, m_de2);
	swap(m_hl, m_hl2);
```

The reason being that if you do it this way, then if a specialisation for `swap` is defined in the namespace containing `PAIR`, dependent name lookup will prefer it over the blind implementation of `std::swap` using a temporary.  (I realise it won’t really make a difference in practice for this particular case.)

src/devices/cpu/z80/z80.cpp
```diff
@@ -1461,6 +1395,23 @@ inline void z80_device::ei()
...
+inline void z80_device::illegal_1() {
+	LOGUNDOC("ill. opcode $%02x $%02x ($%04x)\n",
+			m_opcodes.read_byte(translate_memory_address((PCD-1)&0xffff)), m_opcodes.read_byte(translate_memory_address(PCD)), PCD-1);
+}
+
+inline void z80_device::illegal_2()
+{
+	LOGUNDOC("ill. opcode $ed $%02x\n",
+			m_opcodes.read_byte(translate_memory_address((PCD-1)&0xffff)));
+}
```
cuavas, line 1413: More K&R formatting.  Also, you changed most of the file to have spaces around binary operators – why no spaces in the expressions here?

src/devices/cpu/z80/z80.cpp
```diff
@@ -3356,7 +3287,7 @@ void z80_device::nomreq_ir(s8 cycles)
 
 void z80_device::nomreq_addr(u16 addr, s8 cycles)
 {
-	for (; cycles; cycles--)
+	while(cycles--)
 	{
```
cuavas, line 3291: `while` isn’t a function – it should have a space like the other control flow keywords like `while (...)`

PR #11523, opened 2023-09-01
spectrum/zxbus.cpp: Nemo IDE Controller

holub:


src/devices/bus/spectrum/nemoide.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+void nemoide_device::map_io(address_map &map)
+{
+	map(0x00011, 0x00011).mirror(0xff00).lrw8(NAME([this]() { return m_ata_data_latch; })
+		, NAME([this](offs_t offset, u8 data) { m_ata_data_latch = data; }));
+	map(0x00010, 0x00010).select(0xffe0).lrw8(NAME([this](offs_t offset) { return ata_r(offset >> 5); })
+		, NAME([this](offs_t offset, u8 data) { ata_w(offset >> 5, data); }));
+	map(0x000c8, 0x000c8).mirror(0xff00).lrw8(NAME([this]() { return m_ata->cs1_r(6); })
+		, NAME([this](offs_t offset, u8 data) { m_ata->cs1_w(6, data); }));
```
cuavas, line 62: Why have the shifts here, rather than doing them in `ata_r` and `ata_w`?  Those methods aren’t used anywhere else anyway, and putting the shifts here clutters the address map.

src/devices/bus/spectrum/nemoide.cpp
```diff
@@ -0,0 +1,91 @@
...
+void nemoide_device::device_start()
+{
+	save_item(NAME(m_ata_data_latch));
+
+	m_zxbus->install_device(0x00000, 0x1ffff, *this, &nemoide_device::map_io);
+}
+
+void nemoide_device::device_reset()
+{
+}
```
cuavas, line 85: You don’t actually need an empty `device_reset` – there’s an empty one in `device_t` already (unlike `device_start` that you always need to implement).

src/devices/bus/spectrum/nemoide.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+} // anonymous namespace
+
+} // namespace bus::spectrum::zxbus
+
+DEFINE_DEVICE_TYPE_PRIVATE(ZXBUS_NEMOIDE, device_zxbus_card_interface, bus::spectrum::zxbus::nemoide_device, "nemoid", "Nemo IDE Controller")
```
cuavas, line 91: The short name seems overly short and generic.  Should it be more explicit?

holub: typo: nemoid > nemoide
can't think anything more specific

Pernod70: By no means a requirement but personally I would prefix the device shortname with 'zxbus_', so 'zxbus_nemoide', and do the same for smuc and neogs. This keeps them distinct from potential machine shortnames. You should find that all spectrum interface devices are similarly prefixed with 'spectrum_'.

holub: Nice catch! I'll do so in one if the fillowing PRs

PR #11524, opened 2023-09-02
igs/pgm.cpp: Fix zooming behavior, Add notes

cam900:


qeeg:
Is there anything justifying this change?

cam900:
because current mame's behavior has slightly wrong PGM zoom at most games(when last entry of zoom RAM is all 0).

cuavas:
Which games are affected by this?  Can you show a before/after example?

cuavas:
FWIW, “all bits set” being hard-coded to 1 seems plausible.

Is there any case where a game sets zero at another place in the zoom table and expects to actually use zero?  If not, it could also be that any zero entry in the table is treated as 1.

cam900:
The zoom table stored in PGM bios has zero in the last entry, and most games use the table directly or duplicated.
But some games (ex: dw2/3/pc) using slightly different table and its last entry is 1.

Comparision(ddp3):

before
before

after
after

happppp:
Is 0 an invalid zoom value, and if so does this make more sense?
zoomval = zoomtable[];
zoom = zoomval ? zoomtable[] : 1;

cam900:
> Is 0 an invalid zoom value, and if so does this make more sense? zoomval = zoomtable[]; zoom = zoomval ? zoomval : 1;

It needs to more hardware test. I guess last zoom table entry is buggy or hardcoded.

cuavas:
I’ll merge this for now, with the altered comment.  If no game expects the current behaviour for a zero in the table, I think it’s at least as likely that zero and 1 just do the same thing.  It really needs a test on hardware.

PR #11527, opened 2023-09-02
heathkit/tlb.cpp: Add options for CPU speed, CRT color

mgarlanger:
- Added options for both CPU speed and CRT Colors
- Minor changes to keyboard codes
- Update ROM Regions to actual size of ROMS

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -71,6 +71,10 @@ static constexpr XTAL GP19_DOT_CLOCK_3 = XTAL(10'644'000);   // Graphics mode
 
 // Standard H19 used a 2.048 MHz clock for Z80
 static constexpr XTAL H19_CLOCK = MASTER_CLOCK / 6;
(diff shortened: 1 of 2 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -208,6 +213,42 @@ void heath_tlb_device::device_reset()
...
+	case 0x00:
+	default:
+		// Standard Clock
+		m_maincpu->set_clock(H19_CLOCK);
+		break;
+	}
+
+	// Set screen color
+	switch(BIT(cfg,2, 2))
+	{
```
cuavas, line 240: This `switch` is missing the space between the keyword and the opening parenthesis of the control expression (unlike the one above it).

mgarlanger: Spacing fixed, in case you didn't notice, I have enabled the `allow edits by maintainers` in the last several PRs I've submitted.

PR #11528, opened 2023-09-03
machine/rp5c15.cpp: implement set_epoch setter, fix x68k default year to be current rather than -20

angelosa:


cuavas:
Changing the epoch when the year is written seems like baking in X68000-specific hackery.  How does the X68000 decide what the RTC’s year is relative to?  Does it store a base epoch elsewhere?

angelosa:
I agree and I'm not really keen about this.
x68k expects that year covers a range of 00 to 99 where:
- 00 = 1980
- 20 = 2000
- 43 = 2023
- 99 = 2079

It may be simplified to have just a one time setup in device_start depending on how cooperative device_rtc_interface is.

angelosa:
There's also this that expects the year to be +20, from kenkous4 -> overtake

ajrhacker:
> It may be simplified to have just a one time setup in device_start depending on how cooperative device_rtc_interface is.

If you're curious, `rtc_clock_updated` isn't called until after devices have been started, and it also follows the NVRAM loading phase (since it's very common for RTCs to include NVRAM blocks or have static logic registers that can also be used as NVRAM).

PR #11530, opened 2023-09-04
machine/rp5c15.cpp: implement set_year_offset setter, fix x68k default year setup to be current rather than -20

angelosa:


angelosa:
Looks like `ds1386_device` already overrides set_current_time, I'll deal with this later.

PR #11540, opened 2023-09-10
capcom/cps2.cpp: Support for CPS2 debug DIP switches

drojaazu:
Adds support for the debugging DIP switches which are referenced in many CPS2 games and which apparently existed on dev boards thanks to images that have surfaced. Implemented as a machine configuration.

A write-up detailing the switches is posted here:
https://sudden-desu.net/entry/cps2-debug-switches-and-the-games-that-love-them

Note that in the article, I mentioned deciding against submitting a PR as it should technically be a whole new driver as it is different hardware, but given a conversation with @MooglyGuy on twitter, I've decided to go ahead and let the devs decide if it should be accepted.
https://twitter.com/TheMogMiner/status/1588881175309086730

### DIFF
--- a/src/mame/capcom/cps2.cpp
+++ b/src/mame/capcom/cps2.cpp
@@ -668,7 +668,7 @@ public:
 		, m_output(*this, "output")
 		, m_io_in0(*this, "IN0")
 		, m_io_in1(*this, "IN1")
-		, m_dsw{ { *this, "DSWA" }, { *this, "DSWB" }, { *this, "DSWC" } }
+		, m_dsw(*this, "DSW%c", 'A')
 		, m_cps2_dial_type(0)
 		, m_ecofghtr_dial_direction0(0)
 		, m_ecofghtr_dial_direction1(0)
@@ -741,7 +741,7 @@ private:
 
 	optional_ioport m_io_in0;
 	optional_ioport m_io_in1;
-	optional_ioport m_dsw[3];
+	optional_ioport_array<3> m_dsw;
 
 	std::unique_ptr<uint16_t[]> m_cps2_buffered_obj;
 	std::unique_ptr<uint16_t[]> m_gigaman2_dummyqsound_ram;
--- a/src/mame/capcom/cps2.cpp
+++ b/src/mame/capcom/cps2.cpp
@@ -668,6 +668,7 @@ public:
 		, m_output(*this, "output")
 		, m_io_in0(*this, "IN0")
 		, m_io_in1(*this, "IN1")
+		, m_dsw{ { *this, "DSWA" }, { *this, "DSWB" }, { *this, "DSWC" } }
 		, m_cps2_dial_type(0)
 		, m_ecofghtr_dial_direction0(0)
 		, m_ecofghtr_dial_direction1(0)
@@ -728,8 +729,6 @@ private:
 	void init_cps2_video();
 	void init_cps2crypt();
 
-	uint8_t cps2_dsw_r(offs_t offset);
-
 	optional_shared_ptr<uint16_t> m_decrypted_opcodes;
 	optional_memory_region m_region_key;
 
@@ -742,6 +741,7 @@ private:
 
 	optional_ioport m_io_in0;
 	optional_ioport m_io_in1;
+	optional_ioport m_dsw[3];
 
 	std::unique_ptr<uint16_t[]> m_cps2_buffered_obj;
 	std::unique_ptr<uint16_t[]> m_gigaman2_dummyqsound_ram;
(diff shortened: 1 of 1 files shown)


src/mame/capcom/cps2.cpp
```diff
@@ -1417,7 +1426,7 @@ void cps2_state::cps2_map(address_map &map)
 	map(0x804030, 0x804031).r(FUNC(cps2_state::cps2_qsound_volume_r));                                                                // Master volume. Also when bit 14=0 addon memory is present, when bit 15=0 network adapter present.
 	map(0x804040, 0x804041).w(FUNC(cps2_state::cps2_eeprom_port_w));                                                                  // EEPROM
 	map(0x8040a0, 0x8040a1).nopw();                                                                                                   // Unknown (reset once on startup)
-	map(0x8040b0, 0x8040b3).r(FUNC(cps2_state::kludge_r));                                                                            // Unknown (xmcotaj hangs if this is 0)
+	// 0x8040b0 to 0x8040b2 are the debug DIPs present on development hardware, implemented as read handler cps2_dsw_r
```
galibert, line 1429: Why not just put the handler there instead of commenting?

src/mame/capcom/cps2.cpp
```diff
@@ -1456,7 +1465,7 @@ void cps2_state::dead_cps2_map(address_map &map)
...
```
galibert, line 1468: Same here.

src/mame/capcom/cps2.cpp
```diff
@@ -10821,6 +10915,8 @@ void cps2_state::init_cps2()
 	// Decrypt the game - see machine/cps2crypt.cpp
 	init_cps2crypt();
 	init_cps2nc();
+
+	m_maincpu->space(AS_PROGRAM).install_read_handler(0x8040b0, 0x8040b2, read8sm_delegate(*this, FUNC(cps2_state::cps2_dsw_r)));
```
galibert, line 10919: Much better to have that directly in the memory map.  If you have a compelling reason to do that install here anyway, use emu::rw_delegate instead of read8sm_delegate though.  We're gonna have to batch-convert those at some point.

src/mame/capcom/cps2.cpp
```diff
@@ -1313,13 +1314,8 @@ uint16_t cps2_state::cps2_qsound_volume_r()
...
  *
  *************************************/
-
-uint16_t cps2_state::kludge_r()
-{
-	return 0xffff;
-}
-
 uint16_t cps2_state::joy_or_paddle_r()
+
```
galibert, line 1318: Bad added blank line.

galibert (changes requested):

src/mame/capcom/cps2.cpp
drojaazu: I had tried adding it directly to the map a couple of different ways, but it always complained about the alignment. I couldn't find a way to do byte reads, and always got some variation on "wrong read handler" coming from the maincpu component. I found the delegates as a method while looking at other drivers for examples.

Can you suggest a method for putting it directly in the map to work with single byte reads? Or point me to an existing driver that does something similar? I admit I haven't done tons of MAME development, so I'm mainly looking at existing code for examples.

angelosa: If it's an 8-bit port then you'd need a lambda over a `required_ioport`, cfr. this example:
https://github.com/mamedev/mame/blob/9eab73cc4c6b901e91490f075615d7e1b8bc6641/src/mame/capcom/alien.cpp#L186

drojaazu: Implemented directly into the memory map as a lambda. Includes a check against the HW_TYPE machine configuration to accurately reflect standard (final production) hardware.

src/mame/capcom/cps2.cpp
```diff
@@ -668,6 +668,7 @@ class cps2_state : public cps_state
 		, m_output(*this, "output")
 		, m_io_in0(*this, "IN0")
 		, m_io_in1(*this, "IN1")
+		, m_dsw{ { *this, "DSWA" }, { *this, "DSWB" }, { *this, "DSWC" } }
```
galibert, line 671: Nitpicking level: very high.

Please use a optional_ioport_array instead.

drojaazu: Nitpicking makes for quality. :)

I saw optional_ioport_array and considered it, but it looks like it expects the ports to be differentiated numerically since it uses integer_sequence. The board specifically names the banks DIPSWA, DIPSWB, DIPSWC, so I thought for accuracy's sake, the naming should reflect the board labeling to some extent. The cps1 driver also has them named A/B/C for whatever that is worth.

If you think optional_ioport_array is still the right choice, let me know and I'll go ahead and implement it that way.

Osso13: you can also use letters as in: https://git.redump.net/mame/tree/src/mame/atari/triplhnt.cpp#n52

drojaazu: ... I really should have realized that would work with %c.

Thanks for everyone's help. Updated and pushed.

galibert (approved):

PR #11542, opened 2023-09-10
sinclair/sprinter.cpp: fix audio chanels for 16b mode

holub:


src/mame/sinclair/sprinter.cpp
```diff
@@ -1513,9 +1513,15 @@ TIMER_CALLBACK_MEMBER(sprinter_state::irq_off)
 
 TIMER_CALLBACK_MEMBER(sprinter_state::cbl_tick)
 {
-	const u16 left = m_cbl_data[m_cbl_cnt++];
+	u16 left = m_cbl_data[m_cbl_cnt++];
+	u16 right = cbl_stereo() ? m_cbl_data[m_cbl_cnt++] : left;
+	if (cbl_mode16())
+	{
+		using std::swap;
+		swap(left, right);
```
galibert, line 1521: Urgh, why not std::swap(left, right); instead?

holub: https://github.com/mamedev/mame/pull/11522#discussion_r1313253334

galibert: Oh wow.

Issue #7922, opened 2021-04-03
Launch on primary display in Windows


mydoghasworms:
In Windows, mame does not respect the primary display setting to launch on the primary display.

I have a Windows 10 laptop with a monitor connected via HDMI. In the Windows settings, I have set the monitor to the primary display, but mame always launches on the laptop display.

I have tried the command line flags with little or no success, but it would be nice not to have to tinker with those and have mame just working "as expected".

There is a similar bug for Linux: #7532

john-iv:
Run mame with the -v command to get the displays interrogated:


Mame dkong -v|find \\.\DISPLAY<file://./DISPLAY>

In my 3 monitor case it produces:
Video: Monitor 65539 = "\\.\DISPLAY1"
Video: Monitor 65537 = "\\.\DISPLAY2" (primary)
Video: Monitor 65541 = “\\.\DISPLAY3<file://./DISPLAY3>”

Then use mame dkong -screen \\.\DISPLAY2<file://./DISPLAY2> in this case.

I agree though, it would be nice for MAME to default to the primary.


From: Martin Ceronio ***@***.***>
Sent: Saturday, April 3, 2021 6:59 AM
To: mamedev/mame ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [mamedev/mame] Launch on primary display in Windows (#7922)


In Windows, mame does not respect the primary display setting to launch on the primary display.

I have a Windows 10 laptop with a monitor connected via HDMI. In the Windows settings, I have set the monitor to the primary display, but mame always launches on the laptop display.

I have tried the command line flags with little or no success, but it would be nice not to have to tinker with those and have mame just working "as expected".

There is a similar bug for Linux: #7532<https://github.com/mamedev/mame/issues/7532>

DavidHaywood:
yeah MAME's behavior here is against the grain, especially as the default option is 'auto' which would suggest an intelligent choice.  In reality, it just always chooses 'screen1'

Davidian1024:
For what it's worth I found a fix that's working for me on Ubuntu 20.04.6.  Maybe it'll work for you as well.
https://github.com/mamedev/mame/issues/7532#issuecomment-1713880191

Davidian1024:
This might be fixed as of 0.259

birdybro:
I can confirm it still is not fixed if you are in windowed mode. Windowed mode always launches on Display 1 in Windows no matter if it's primary at least for me.

[...]

Issue #7532, opened 2020-12-07
Cannot make MAME to show on primary display (Linux, Gnome, MAME 0.226)


TheCodeTherapy:
MAME 0.226, compiled from source on PopOS 20.04 (Ubuntu-based distro). This is somehow related to #7309 but it doesn't work properly.

The setup used to reproduce is a triple-monitor setup, with 3 monitors connected by DisplayPort.

I tried the suggested command line parameters provided on #7309 however:
-screen screen0 makes MAME appear on my right monitor
-screen screen1 does the same (MAME appears on right monitor)
-screen screen2 makes MAME appear on my left monitor

All other possible screenX provided make MAME appear on the right monitor, and so far I couldn't find a way for it to appear on my center monitor (which is the primary display on my distro configuration).

Worth mentioning that the issue only occurs on Gnome (on KDE Plasma, maybe due to differences in how KDE Plasma manages composition or display management, the issue does not occur, and MAME always appears on the primary display).

Worth mentioning as well that there's no information on MAME docs on how the OSD  PER-WINDOW VIDEO OPTIONS should be configured on Linux machines. The docs only mention the \\.\DISPLAYn, which is a Windows-only notation.

mame64 -verbose:

`
Attempting load of mame.ini
Parsing mame.ini
Attempting load of mame.ini
Parsing mame.ini
Reading translation file English: 1 strings, original table at word offset 7, translated table at word offset 9
Loaded 0 translations from file English
Starting plugin data...
Attempting load of mame.ini
Parsing mame.ini
Attempting load of mame.ini
Parsing mame.ini
Available videodrivers: x11 wayland dummy 
Current Videodriver: x11
	Display #0
		Renderdrivers:
			    opengl (0x0)
			 opengles2 (0x0)
			  software (0x0)
	Display #1
		Renderdrivers:
			    opengl (0x0)
			 opengles2 (0x0)
			  software (0x0)
	Display #2
		Renderdrivers:
			    opengl (0x0)
			 opengles2 (0x0)
			  software (0x0)
Available audio drivers: 
	pulseaudio          
	alsa                
	sndio               
	dsp                 
	disk                
	dummy               
Build version:      0.226 (mame0226-503-g3a955d2df8c)
Build architecure:  
Build defines 1:    SDLMAME_UNIX=1 SDLMAME_X11=1 SDLMAME_LINUX=1 
Build defines 1:    LSB_FIRST=1 PTR64=1 
[...]

PR #11546, opened 2023-09-11
removes SDL_WINDOW_BORDERLESS flag

Davidian1024:
Meant to address the issue reported here:
https://github.com/mamedev/mame/issues/7532#issuecomment-1713880191 And possibly here:
https://github.com/mamedev/mame/issues/7922#issuecomment-1713887214

### DIFF
--- a/src/osd/sdl/window.cpp
+++ b/src/osd/sdl/window.cpp
@@ -593,7 +593,7 @@ int sdl_window_info::complete_create()
 	// create the SDL window
 	// soft driver also used | SDL_WINDOW_INPUT_GRABBED | SDL_WINDOW_MOUSE_FOCUS
 	m_extra_flags |= (fullscreen() ?
-			SDL_WINDOW_BORDERLESS | SDL_WINDOW_INPUT_FOCUS | SDL_WINDOW_FULLSCREEN : SDL_WINDOW_RESIZABLE);
+			SDL_WINDOW_INPUT_FOCUS | SDL_WINDOW_FULLSCREEN : SDL_WINDOW_RESIZABLE);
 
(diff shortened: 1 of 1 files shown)


cuavas:
Does this have a detrimental effect on any platforms?  What happens on Windows and macOS when this flag is not specified?

Davidian1024:
I honestly don't know.  To be completely honest, I don't fully understand what a borderless window is.

All that said, these days I see borderless as optional in probably most games nowadays.

Currently MAME seems to force it's fullscreen window to always be of the borderless variety at least when SDL is being used.

cuavas:
I tested this on Windows with two displays, and it seemed to work.  I think you’re supposed to either specify full screen, or the combination of borderless and maximised, but not a combination.

A borderless window won’t have “decorations” like a title bar, window frame, shadow, etc.  Depending on the OS and window manager, there may be a distinction between a borderless window that covers an entire display, and an application taking exclusive control of a display.

Davidian1024:
Oh ok cool.  :)

rb6502:
FWIW, non-borderless full screen is deprecated by both Microsoft and Apple and we no longer use it at work.  I think this is probably an SDL bug and I greatly oppose this change.

cuavas:
> FWIW, non-borderless full screen is deprecated by both Microsoft and Apple and we no longer use it at work. I think this is probably an SDL bug and I greatly oppose this change.

You're missing the point - SDL always creates a borderless window when you specify fullscreen if it isn't using exclusive mode. If you want it to always use borderless fullscreen (never use exclusive mode, never change resolution), you need to specify borderless and maximised. Specifying a combination of fullscreen and borderless confuses it.

PR #11547, opened 2023-09-11
osd/sdl: Fixed interpretation of result of SDL_GetDesktopDisplayMode

Davidian1024:
SDL_GetDesktopDisplayMode returns 0 on success.  The current logic seems to assume that it returns true on success.  Updated the logic so that the lines printing the desktop mode and current display mode info can be reached.


Davidian1024:
Cool, thanks alot.  :)

PR #11549, opened 2023-09-13
Update zaxxon.cpp

adamtoth:
Changed rotation for zaxxon, super zaxxon, future spy, and congo bongo. Confirmed on original arcade pcbs. I do not have a copy of Jackson bootleg to confirm its default orientation. I have multiple Zaxxon pcbs as well, and all are rotation 270. Confirmed on both 2 and 3 pcb congo bongo stacks. Having incorrect orientation in MAME makes advanced searching on other sites like adb.arcadeitalia.com based on rotation incorrect.

### DIFF
--- a/src/mame/sega/zaxxon.cpp
+++ b/src/mame/sega/zaxxon.cpp
@@ -1611,17 +1611,17 @@ void zaxxon_state::init_razmataz()
  *************************************/
 
 /* these games run on standard Zaxxon hardware */
-GAME( 1982, zaxxon,   0,      zaxxon,    zaxxon,   zaxxon_state,   empty_init,    ROT90,  "Sega",    "Zaxxon (set 1, rev D)",        MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE )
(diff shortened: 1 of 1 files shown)


happppp:
This would make the emulated screen upside-down.
If cocktail mode emulation is currently correct, then fixing it is simply a matter of setting m_flip_screen in zaxxon_v.cpp to bool(state);

cuavas:
Please give the pull request a meaningful title.  You should describe what it actually does.

rb6502:
As hap said, just changing the ROT flag will make the screen upside down, which is not something we'd want to ship.  The rendering also needs to be reworked so the games are right-side up again.

adamtoth:
If you were being 100% accurate to the hardware, yes the game would be upside down if you assumed that most players setting up a vertical monitor would use 90 degree rotation as a standard. Is that how all MAME vertical games work? Games with ROT90 are left alone, but all ROT270 games have some extra code to flip it to 90? Wouldn't you want it accurate to the hardware and then let a user use the flipx/y or rotation video settings to adjust to their own monitor rotation? Not trying to argue, just to understand how it is done in MAME, as there are other games that I'd like to submit similar adjustments to. I'm happy to change the m_flip_screen variable as hap suggested.

PR #11550, opened 2023-09-16
luaengine.cpp: fix minor typo offfset -> offset

goldnchild:
Noticed a minor typo in luaengine.cpp


angelosa:
imho don't think that's minor.

PR #11551, opened 2023-09-16
seta/seta.cpp: Input updates

cam900:
- Add Romanization of Chinese title in zingzip, Set default title as English only
- Fix rezon input, Set Licensor text selectable


mamehaze:
the reason the 1991 Rezon set is hardcoded to Taito and the 1990 set is hardcoded to Not-Taito is because we've only seen those configurations in the wild.  It's a single byte change in ROM, but it seems that all versions distributed by Taito were the 1991 version, and all not were the 1990 version.

This PR loses that information.

cam900:
Some taito licensed version of rezon has 1991 copyright date instead 1992, ex: https://www.youtube.com/live/VHajRAX6YSY?si=FG6kI_NTFN5uo1fx (notice:glitch)

PR #11552, opened 2023-09-16
seta/simple_st0016.cpp: Add/Fix Romanized titles and original Japanese title

cam900:


src/mame/seta/simple_st0016.cpp
```diff
@@ -583,7 +583,7 @@ void st0016_state::renju(machine_config &config)
  *************************************/
 
 /*
-Renjyu Kizoku
+KIRA KIRA 五目ならべ(Gomoku Narabe) 連珠貴族(Renju Kizoku)
```
cuavas, line 586: If you’re going to put transliterations as well as the original script, please put the whole transliterated title together – mixing it together like this is confusing.

PR #11564, opened 2023-09-21
nltool: report line number when an exception occurs

stonedDiscord:
I tried to convert a KiCad netlist and when a component value was missing or not a number nltool would only report `Exception caught: Error converting string to number:` but not what component or line was meant.

This PR counts the lines, prints the line number to stderr and passes on the exception.


stonedDiscord:
I could fix those up when I'm back from my vacation in a wekk unless someone else wants to beat me to it

PR #11580, opened 2023-09-30
bcheetah.lay: add keypad

goldnchild:
Added a keypad to the bcheetah layout like the one on the Incredible Brain Buggy.

### DIFF
--- a/src/mame/layout/bcheetah.lay
+++ b/src/mame/layout/bcheetah.lay
@@ -23,10 +23,157 @@ license:CC0-1.0
 	<element name="text_m2"><text string="motor 2:"><color red="0.7" green="0.7" blue="0.7" /></text></element>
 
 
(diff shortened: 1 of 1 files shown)


happppp:
The Bandai Cheetah button panel doesn't look like that. Buttons are red, labels are above buttons.
Neat playing around with .lay features. It can be made so much simpler if you just stick to the basics.

goldnchild:
Took me awhile to find a picture of the bandai cheetah:


http://www.computercars.de/car/bandai_cheetah

In the picture it says "System Contorol Car".

Ok, I can fix the button labels and the colors.

happppp:
Hmm, ok I'll simplify it if you don't mind

goldnchild:
wow, that is seriously condensed.  That cuts it down from 9911 to 5915 bytes.

PR #11581, opened 2023-09-30
namco/namcops2.cpp: correct notes with regards of THE IDOLM@STER & V300 I/O boards

Keropon:
Updated the dump status of THE IDOLM@STER & Minna de Kitaeru Zenno Training.

Also added some info about the [V300 FCB PCB](https://twitter.com/lsylsy2/status/1071616172263534592) found on the actual THE IDOLM@STER cabinets.

### DIFF
--- a/src/mame/namco/namcops2.cpp
+++ b/src/mame/namco/namcops2.cpp
@@ -125,8 +125,8 @@ NM00053 ?
 NM00054 ?
 NM00055 ?
 
-Note about :
-This is a general guideline flag showing compatibility between the different revisions of  246 hardware.
+Note about System:
+This is a general guideline flag showing compatibility between the different revisions of System 246 hardware.
 1-7: Tested Working
 X  : Tested Not Working
 -  : Not Tested
@@ -625,7 +625,7 @@ Texas Instruments TMS32VC540x DSP, TSOP32 flash ROM and other components.
 V300 FCB PCB
 2606960100 (2582960101)
 
-Notes: V300 FCB PCB is almost identical to the V290 FCB PCB. The main differences are changed internal MCU 
+Notes: V300 FCB PCB is almost identical to the V290 FCB PCB. The main differences are changed internal MCU
 code & PIC code, more output transistors and an traded the RS-232 for a RS-232C IC.
 [to-do]
 
--- a/src/mame/namco/namcops2.cpp
+++ b/src/mame/namco/namcops2.cpp
@@ -35,18 +35,21 @@ Capcom Fighting Jam/Capcom Fighting Evolution... XXXX56X  JAM1 DVD0            D
 Cobra The Arcade................................ XXXX56X  CBR1-HA              HDD (40GB)   NM00021   CBR1 Ver.B             Namco 2004                    Requires RAYS I/O PCB and IR guns and IR sensors. HDD: Maxtor DiamondMax Plus 8 40GB 6E040L0
 Dragon Chronicles (satellite)................... ------X *DCO31-TS CD0        *CD           NM00020   DC001 Ver.A            Namco 2002                    \
 Dragon Chronicles Legend of the Master Ark (sat) ------X *DGC11 CD0           *CD          *NM00014  *DGC11 Ver.A1           Namco 200?                    | server is a custom PC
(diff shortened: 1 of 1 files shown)


src/mame/namco/namcops2.cpp
```diff
@@ -122,8 +125,8 @@ NM00053 ?
 NM00054 ?
 NM00055 ?
 
-Note about System:
-This is a general guideline flag showing compatibility between the different revisions of System 246 hardware.
+Note about :
+This is a general guideline flag showing compatibility between the different revisions of  246 hardware.
 1-7: Tested Working
```
Gurudumps: I wrote this documentation. No sorry, you can not remove the word System here. 'Note about:' means nothing. The notes are about "SYSTEM 246 hardware", not "246 hardware".

src/mame/namco/namcops2.cpp
```diff
@@ -614,11 +617,17 @@ V290 FCB PCB
...
 It supports a serial touchscreen interface, card readers and buttons.
 The additional devices are supported via J108 which connects to another PCB 'XOU020-A' which contains a
 Texas Instruments TMS32VC540x DSP, TSOP32 flash ROM and other components.
 
+V300 FCB PCB
+2606960100 (2582960101)
+
+Notes: V300 FCB PCB is almost identical to the V290 FCB PCB. The main differences are changed internal MCU 
+code & PIC code, more output transistors and an traded the RS-232 for a RS-232C IC.
+[to-do]
```
Gurudumps, line 631: "and an traded the RS-232 for a RS-232C IC". What? Come on at least write proper English/grammar. Anyway this part quoted is not required. In practical terms for emulation purposes there is no difference between RS-232 and RS-232C

src/mame/namco/namcops2.cpp
happppp: Already corrected https://github.com/mamedev/mame/pull/11581/commits/1a5e1cd4d2384f5b93e7ea7fdf922572cb1dbacb

src/mame/namco/namcops2.cpp
Keropon: Thanks for the _destructive_ commentary. English is not my first language. 
I copied a note that was similar and lots of people look at MAME's documentation to find quirks or details about some hardware, so I included it for completion's sake. No need to be like that.
Will edit.

src/mame/namco/namcops2.cpp
Keropon: @Gurudumps Missclicked and deleted the word System by mistake, this is **not** an attack to your work. Will correct.

src/mame/namco/namcops2.cpp
Keropon: There are other emulators that utilize MAME drivers and documentation as a basis for naming conventions and guidelines regarding ROM/dump names. I've used them myself as a reference to identify the necessary components for running my own 2x6 hardware. I don't believe it's 'pointless', and I hope other reviewers share the same perspective.

angelosa: > Anyway your whole addition is pointless as PS2 emulation will never be working in MAME to any level of playability 

Nobody in the dev circle really cared so far, already replied to this in the past, it's not a `never` ...

> In future if you are going to make changes to existing docs

I'm honestly of the opinion that these super verbose docs can live as separate markdown files in 2023. They don't help at all when you have to jump back and forth between multi thousand lines of headers.

cuavas:
@Keropon do you have photos of the V290 and V300 I/O boards with PCB and chip silkscreen readable?

Gurudumps:
I documented V290 in the src....
https://github.com/mamedev/mame/blob/master/src/mame/namco/namcops2.cpp#L585

V300 is the same board with a sticker over the top of V290....
https://imgur.com/a/isBNG62
Differences are the number of driver transistors and connectors populated and possibly the PIC and MCU code.
All *extremely* unimportant given that PS2 h/w isn't emulated/working and the games will boot (for emulation purposes) with a number of whitelisted I/O boards.

cuavas:
> All _extremely_ unimportant given that PS2 h/w isn't emulated/working and the games will boot (for emulation purposes) with a number of whitelisted I/O boards.

MAME isn't just about playing games on your PC. You of all people should know that. Also what isn't emulated changes over time, and having more complete information helps down the road.

Gurudumps:
Please re-read the above, this time more slowly. The 290 and 300 boards are the same physical PCB with a few parts populated/not populated.

angelosa (approved):

PR #11582, opened 2023-09-30
dgpix: Use intelfsh + improvements to sound emulation

987123879113:
- Removed the fake intelfsh emulation code and turned it all into real intelfsh devices. Each of the ROM board types have up to 8 flash ROMs but usually only 2 are populated, except for kdynastg which had 4. This involved some endianness changes so the patches were also fixed.
- Implemented BMkey ROM PCB and refactored out the old Type A ROM PCB code into separate devices because they have slightly different memory mappings and usages
- Implemented/documented a new rotate opcode in the KS0164 CPU which was in a codepath used by btplay2k
- Fixed issue where shifts would not disassemble due to bit 3 being set (the actual CPU emulation code does not use bit 3 for those opcodes)
- Swapped bges/bles opcodes which was causing some code to be executed improperly (fixes sound effects in a lot of games)
- Stop sounds once they finish playing + mark the possible loop flag (unused by any game though)
- Fixed controls for btplay2k and letsdnce

All games tested to make sure they're still at least booting. letsdnce is mostly playable with this but has timing issues, and btplay2k plays sounds properly but there's a timer issue somewhere causing songs to randomly pause or end early.

funkball, which uses the same ks0164 chip, plays sound effects now with the bges/bles swap change in this PR.

### DIFF
--- a/src/devices/cpu/ks0164/ks0164.cpp
+++ b/src/devices/cpu/ks0164/ks0164.cpp
@@ -253,8 +253,8 @@ void ks0164_cpu_device::execute_run()
 			case 0x9: cond = !(m_r[R_PSW] & F_Z) && !(m_r[R_PSW] & F_C); break;
 			case 0xa: cond = (m_r[R_PSW] & F_Z) || (m_r[R_PSW] & F_C); break;
 			case 0xb: cond = (!(m_r[R_PSW] & F_Z) && (m_r[R_PSW] & F_N) && (m_r[R_PSW] & F_V)) || (!(m_r[R_PSW] & F_Z) && !(m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
-			case 0xc: cond = (m_r[R_PSW] & F_Z) || ((m_r[R_PSW] & F_N) && !(m_r[R_PSW] & F_V)) || ((m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
-			case 0xd: cond = ((m_r[R_PSW] & F_N) && (m_r[R_PSW] & F_V)) || (!(m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
+			case 0xc: cond = ((m_r[R_PSW] & F_N) && (m_r[R_PSW] & F_V)) || (!(m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
+			case 0xd: cond = (m_r[R_PSW] & F_Z) || ((m_r[R_PSW] & F_N) && !(m_r[R_PSW] & F_V)) || ((m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
 			case 0xe: cond = ((m_r[R_PSW] & F_N) && !(m_r[R_PSW] & F_V)) || ((m_r[R_PSW] & F_V) && !(m_r[R_PSW] & F_N)); break;
 			case 0xf: default: cond = true; break;
 			}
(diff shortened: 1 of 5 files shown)


cuavas:
So are those games added as working actually not working?  Rhythm games with timing issues shouldn’t be marked working.

mamehaze:
They didn't even have working audio in the release build, as I mentioned prior to the release.

People outside of the circle of the team are laughing at the lack of standards here; as far as they're concerned games like Taiko have no desecernable problems, but are marked as NOT WORKING yet music games where the audio doesn't even correspond to the track being played get a working flag.  It's being used as justification to strip the warning messages.

This changeset makes things much, much better, but still not even close to the quality of the Taiko emulation which apparently doesn't meet the criteria for a working flag.

angelosa:
The problem is at root: `MACHINE_NOT_WORKING` (or even the `GAME_NOT_WORKING`) originally meant that the game was a straight black screen or crashes during gameplay (not even service mode), the _philosophical_ aspects of the _immersion ruined_ have been added later, once we developed a QA over the years.

Now, this "QA" is an opinion based where i.e. almost every time I have to touch a SW list I have to make non-trivial demoting decisions, which in turn made me realize that "demoting" isn't the end of the world like you are putting it here but it's just a part of a "game". 

Game that you or those so called "people outside of the circle" definitely have no interest to either improve or even discuss somewhere so to lead to an official updated spec, and notice that in a normal company scenario it's the user that is supposed to _flag_ the product as working or not, definitely not the programmer(s).

987123879113:
>So are those games added as working actually not working? Rhythm games with timing issues shouldn’t be marked working.

I don't think they should've been marked as working to begin with to be honest. letsdnce is the most passible out of the two but it definitely has some sync issues. btplay2k is still broken enough to be marked not working. 

I played around with various parameters to see if they affected timing (video timings, sound CPU clocks, etc) but the only things that actually had an effect was the main CPU's clock which I don't think is wrong in the code, so it's something further internal that's the cause. The game sends MIDI commands to the sound CPU to play sounds so there might be some small delay between when the command is sent, processed, and actual sound is played but I'm not really sure.

cuavas:
They obviously shouldn’t have been marked working.  I’m getting really sick of premature promotions things being marked working when they clearly aren’t playable.

@mamehaze on the one hand you accuse me of gatekeeping and tell me I need to trust people more, then on the other you attack me when stuff slips past like the ADSP2100 regression, CHD regression, and these games being marked working in a release.  These were all done by “trusted” people with direct commit access.  I spent most of a week going through the rest of the changes to the image formats to find other regressions from those changes and fix them before they made it into a release, but you don’t see or care about that.  You just want to throw stones.

All I can do is demote these games now.

mamehaze:
I'm not attacking *you* I'm saying these things shouldn't have happened.

Trusting people should be the default, but if people screw up they screw up, and it gets pointed out.

rb6502:
Do we know how good the sync is for these on hardware?  There have certainly been rhythm games released that are off as shipped, and given the general bootleggy appearance of a lot of these dgpix games I wouldn't be surprised if they were off already.

987123879113:
That I'm not sure of but I could believe the game itself just isn't synced well. I replayed Let's Dance again tonight to double check the timing and on second glance I think it's probably at a personal judgement level now. The only things of note I found were sometimes the background videos felt like they were causing internal frame rate issues (not sure how that is on real hardware, it wasn't consistent and real PCB pics show the game isn't exactly buttery smooth to begin with), and there were also some small glitching with audio (sound effects on the result screen have a pop at the end which isn't there on real hardware videos) but there were no major gamebreaking issues. I wouldn't find it unreasonable to mark just Let's Dance as working now.

My opinion on Beat Player 2000 remains the same though. It has actual gamebreaking bugs that make it unplayable.

PR #11584, opened 2023-09-30
nmk/ddealer.cpp: replace MCU simulation in ddealer with emulated NMK-110 MCU (also tdragon1, hachamf, hachamfa in nmk16.cpp) [Sergio Galiano, Recreativos Piscis, David Haywood, hammy]

mamehaze:
nmk/ddealer.cpp: replace MCU simulation in ddealer with emulated NMK-110 MCU [Sergio Galiano, Recreativos Piscis, David Haywood, hammy]
nmk/nmk16.cpp: replace MCU simulation in tdragon1 with emulated NMK-110 MCU [Sergio Galiano, Recreativos Piscis, David Haywood, hammy]
nmk/nmk16.cpp: replace MCU simulation in hachamf, hachamfa with emulated NMK-113 MCU [Sergio Galiano, Recreativos Piscis, David Haywood, hammy]


src/mame/nmk/ddealer.cpp
```diff
@@ -186,13 +112,68 @@ class ddealer_state : public driver_device
...
+
+void ddealer_state::machine_reset()
+{
+	m_bus_status = 0x04;
+}
+
+void ddealer_state::mcu_port6_w(u8 data)
+{
+	// the actual mechanism is a little more complex, but these are written at the
+	// start and end of the take / release bus function
```
angelosa, line 132: Can you elaborate a bit? Connected to bus request I presume?

angelosa (approved):

src/mame/nmk/ddealer.cpp
mamehaze: the notes I was sent

writes 0x03:  0011 (important thing is bit 3 is zero, asking for bus request)
reads bit 2, until it's zero (receiving bus grant signal from 68k)
writes 0x01:  0001  (important thing here is bit 1 is zero, sending bus grant acknowledge signal to 68k)
writes 0x08: 1000  (important thing here is bit 0 is zero, enabling the memory access logic in the pcb)
wirites 0x09: 1001 (important thing here is bit 0 is one, disabling the memory access logic in the pcb)

sergiopolog: @angelosa yes, it's 68k bus arbitration lines. Device is hooked to the main CPU in this way:

image

- Bus request line is bit 3 of Port 6 and configured as output
- Bus grant line is bit 2 of the Port 6 and it's configured as input
- Bus grant acknowledge line is bit 1 of the Port 6 and it's configured as output
- Also there is another pin that enables the pcb logic to access the bus after all of above it's done, that is the bit 0 of the Port 6, configured as output, too

src/mame/nmk/ddealer.cpp
```diff
@@ -186,13 +112,68 @@ class ddealer_state : public driver_device
...
+	}
+	else if (data == 0xb)
+	{
+		logerror("%s: mcu_port6_w 68k bus returned, 68k started (data %02x)\n", machine().describe_context(), data);
+		m_maincpu->set_input_line(INPUT_LINE_HALT, CLEAR_LINE);
+	}
+	else
+	{
+		logerror("%s: mcu_port6_w 68k bus access (data %02x)\n", machine().describe_context(), data);
+	}
```
cuavas, line 146: Please replace these with `LOG` – having the logging always enabled here is a bit heavy for something attached to bus arbitration.

Also, @galibert can you check if this is this the right way to be doing this?  I want to be sure it’s up-to-date for the new, interruptible 68k.

src/mame/nmk/ddealer.cpp
```diff
@@ -186,13 +112,68 @@ class ddealer_state : public driver_device
...
+{
+	return m_screen->vpos() >> 2;
+}
+
+u8 ddealer_state::mcu_port6_r()
+{
+	// again this is simplified for now
+	m_bus_status ^= 0x04;
+	return m_bus_status;
+}
```
cuavas, line 159: Once again @galibert is there a clean way to do this?  Magic side effects of reading ports that aren’t strobed is a fairly obvious hack.

mamehaze:
@angelosa I don't believe any of your code remains in ddealer.cpp now (it was mostly the simulation?)  

From what I can see I recognize all the code in there as my own, including the weird (probably needs more research) video implementation.  I propose to relicense it as BSD-3-Clause.

angelosa:
> @angelosa I don't believe any of your code remains in ddealer.cpp now (it was mostly the simulation?)

I definitely remember spotting palette among other things from the initial Tomasz stub. It's otherwise fine to downgrade to BSD to me.

rb6502:
For the record, OG said he'd take a look at this once he figures out how the bus request stuff should be done.

PR #11578, opened 2023-09-29
Documentation fixes part 2

Tomcariello:
Various typos & other grammar fixes.

PR #11575, opened 2023-09-27
Various Documentation fixes

Tomcariello:
Fixes of various typos, misspellings, and other minor errors in the website Documentation.

PR #11586, opened 2023-10-01
Fix documentation compilation warning

Firehawke:
While updating my toolset and getting things working again, I noticed there were four warnings in the documentation compilation process. Three of them are sections of the Lua scripting that haven't been written yet. This warning about ``sdlconfig.rst:2: WARNING: Title underline too short.`` could be fixed in one byte so I might as well just get it while I'm here.

### DIFF
--- a/docs/source/commandline/sdlconfig.rst
+++ b/docs/source/commandline/sdlconfig.rst
@@ -1,5 +1,5 @@
 SDL-Specific Command-line Options
-================================
+=================================
 
 This section contains configuration options that are specific to any build
 supported by SDL (including Windows when built with SDL instead of native).


cuavas:
Thanks.  The broken cross-references in the Lua documentation are to remind me that I need to re-think that part of the API and then document it once it’s in a supportable state.  They get rendered as plain text when the target isn’t found.

Firehawke:
Yeah, I noticed that it was rendering them as plain text when I went and checked the results, and that's why I didn't do anything with those. I figured it was a "eventually this will get dealt with" thing since the Lua stuff is still very much in a state of flux in general. As it didn't have any negative impact on the actual output, I figured it didn't need immediate handling.

I'll look into the state of the docs in the next day or so and see where things stand and what needs to be worked on.

cuavas:
There have been a few fairly significant changes to the docs since you last looked.  I think some of the big ones are:
* Lua interface documentation moved to its own section, so that one massive page could be split up.
* Full debugger command reference added.
* Most of the included plugins documented.
* Preliminary “contributing” section added.
* More documentation for the internal UI (Basic MAME Usage section).
* A few more things added in tech specs.

@Tomcariello has also been proof-reading things and catching grammar and editing errors (see e.g. #11574, #11575 and #11578).

PR #11588, opened 2023-10-01
ioprocs.h: Add read_stream::alloc_read and random_read::alloc_read_at…

ajrhacker:
…. These work just like read and read_at, but take a std::unique_ptr to a uint8_t buffer and attempt to newly allocate it with the requested length. If the allocation fails, these return an error code instead of throwing an exception.

* formats/trs80_dsk.cpp: Test the JV3 writable byte without slurping in the entire file.

* diimage.h: Retire image_feof and the pointer-allocating versions of fread.

### DIFF
--- a/src/devices/bus/adam/exp.cpp
+++ b/src/devices/bus/adam/exp.cpp
@@ -76,13 +76,18 @@ void adam_expansion_slot_device::device_start()
 
 std::pair<std::error_condition, std::string> adam_expansion_slot_device::call_load()
 {
+	std::error_condition err;
+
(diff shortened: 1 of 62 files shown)


cuavas:
Well, I haven’t read the entire change, but this breaks Odyssey2 software list loading – it always returns `image_error::UNSPECIFIED`.  This doesn’t inspire confidence.

This also doesn’t address the desire for wrappers to deal with attempting to complete a read/write that returns short.  #11608 is more along the lines of where I wanted to go with this.

PR #11593, opened 2023-10-03
workaround compile issue with tms57002.h / tms57002.hxx

mamehaze:
I don't know if there's a better way to do this, but I can't see one with genie

The problem arises when a file includes tms57002.h before tms57002.hxx has been built, which on a sufficiently quick to compile system and high enough -j value is guaranteed to happen, causing compile failure.

None of the other generated cores appear to include the hxx from the header in this way.

Creating this PR to log the issue as much as anything else.


rb6502:
@galibert Do you have anything better for this?

galibert:
With genie, no.  This is why I comitted the m68k generated files too, there's no same way.

Issue #11567, opened 2023-09-22
MC6847 VDG and screen rapidly drift out of sync (coco)


dave-br:
### MAME version

0.251, still occurs in my fork from around Sep 22 2023 / c92ee7270bddda658bfd165cc15520a76839f488

### System information

Windows 10 x64, Intel Core i5-4690 @ 3.5Ghz, Intel HD Graphics 4600.  English US

### INI configuration details

```INI
#
# CORE CONFIGURATION OPTIONS
#
readconfig                1
writeconfig               0

#
# CORE SEARCH PATH OPTIONS
#
homepath                  .
rompath                   roms
hashpath                  hash
samplepath                samples
artpath                   artwork
ctrlrpath                 ctrlr
inipath                   .;ini;ini/presets
fontpath                  .
cheatpath                 cheat
crosshairpath             crosshair
pluginspath               plugins
languagepath              language
swpath                    software

#
# CORE OUTPUT DIRECTORY OPTIONS
#
cfg_directory             cfg
nvram_directory           nvram
input_directory           inp
state_directory           sta
snapshot_directory        snap
diff_directory            diff
comment_directory         comments
share_directory           share

#
# CORE STATE/PLAYBACK OPTIONS
#
state                     
autosave                  0
rewind                    0
rewind_capacity           100
playback                  
record                    
exit_after_playback       0
mngwrite                  
aviwrite                  
wavwrite                  
snapname                  %g/%i
snapsize                  auto
snapview                  auto
snapbilinear              1
statename                 %g
burnin                    0

#
# CORE PERFORMANCE OPTIONS
#
autoframeskip             0
frameskip                 0
seconds_to_run            0
throttle                  1
sleep                     1
speed                     1.0
refreshspeed              0
lowlatency                0

#
# CORE RENDER OPTIONS
#
keepaspect                1
unevenstretch             1
unevenstretchx            0
unevenstretchy            0
autostretchxy             0
intoverscan               0
intscalex                 0
intscaley                 0

#
# CORE ROTATION OPTIONS
#
rotate                    1
ror                       0
rol                       0
autoror                   0
autorol                   0
flipx                     0
flipy                     0

#
# CORE ARTWORK OPTIONS
#
artwork_crop              0
fallback_artwork          
override_artwork          

#
# CORE SCREEN OPTIONS
#
brightness                1.0
contrast                  1.0
gamma                     1.0
pause_brightness          0.65
effect                    none

#
# CORE VECTOR OPTIONS
#
[...]

PR #11594, opened 2023-10-03
Fix synchronization between MC6847 video display generator and screen device

dave-br:
@tlindner: FYI
 
- fix: #11567 

PROBLEM: The vertical position of the screen and the physical scanline of the MC6847 video display generator drift out of sync further and further every few frames.

PRIMARY FIX: Initialize the screen device using set_raw, with more precise parameters to describe horizontal and vertical blanking.  (Full list of supporting changes below.)

SCENARIOS FIXED:

1) mame 6809 debugging (my primary objective): Hitting F7 (run until next interrupt) stops at the next FS interrupt.  Before this fix, the debugger reports a beamy (based on the screen's vertical position) that varies all over the frame.  About every 3 frames (3 hits of F7), beamy goes up by one, through the entire range of vertical positions due to a continuous drift between the screen's and 6847's timing.  After this fix, beamy no longer drifts further and further away from the VDG's vertical position.  Beamy reliably rests in the range between the end of 216 and beginning of 217.

2) Dragonfire: Not my primary objective with this fix, but I tried out this game since its code is meticulously timed to allow multiple display mode changes per line.  It is still not perfect, but is significantly better.  The changing video garbage of the margins is gone, and the flashing on the sides of the castle walls is 99% gone.  There are still static fringes of purple around the moat.

3) When viewing logging, it is now clear that the 6847's scanline matches the screen's vpos.  The debugger scenario no longer drifts, but is imperfect due to varying cycles expended between an interrupt getting issued and the debugger stopping.  Logging, on the other hand, reliably shows the exact same screen hpos & vpos each time the 6847 advances to a new scanline or issues the FS timer.

OTHER EFFECTS:

1) The L/R margins were smaller than what a CoCo 1/2 normally generates.  I believe they are now the correct size.

2) Due to how the screen device handles the larger number of pixels I'm feeding it, the active display area now appears a bit more stretched vertically than prior builds.  This was jarring to me at first, but when I compare with my CoCo on an old CRT, it looks similar.

LIST OF CODE CHANGES:

Initialize screen device using raw parameters
Inform screen device of full vblanking and hblanking amounts
All hblanking moved from beginning of line to end of line (vblanking remains at end of field)
New logging on next scanline, more detailed logging elsewhere
Moved some repeatedly-used hard-coded numbers into #defines


src/devices/video/mc6847.cpp
```diff
@@ -76,25 +76,42 @@
...
-#define TIMER_FSYNC_ON_TIME     (TIMER_HSYNC_PERIOD * (TOP_BORDER + 192) + TIMER_HSYNC_ON_TIME)
 
 #define LOG_SCANLINE (1U << 1)
 #define LOG_HSYNC    (1U << 2)
 #define LOG_FSYNC    (1U << 3)
 #define LOG_FLUSH    (1U << 4)
 #define LOG_INPUT    (1U << 5)
+#define LOG_NEXTLINE (1U << 6)
 #define VERBOSE (0)
 #include "logmacro.h"
```
angelosa, line 116: This block should really be just after the header declarations, as conventionally done by basically any device/driver that uses `logmacro.h`

src/devices/video/mc6847.cpp
```diff
@@ -406,6 +432,9 @@ inline void mc6847_friend_device::next_scanline()
 		m_logical_scanline_zone = SCANLINE_ZONE_BOTTOM_BORDER;
 		enter_bottom_border();
 	}
+
+	// This helps to confirm synchronization between the screen device and the 6847
+	LOGMASKED(LOG_NEXTLINE, "mc6847_friend_device::next_scanline(): (vpos,hpos)=(%d,%d), m_physical_scanline='%d', m_logical_scanline_zone='%s', m_logical_scanline='%d'\n", (screen().vpos()), (screen().hpos()), m_physical_scanline, (scanline_zone_string((scanline_zone)m_logical_scanline_zone)), m_logical_scanline);
```
angelosa, line 437: I feel like the comment is redunant, at most it should be on same line of the `LOG_NEXTLINE` definition itself above.

src/devices/video/mc6847.cpp
```diff
@@ -752,11 +793,36 @@ void mc6847_base_device::record_partial_body_scanline(uint16_t physical_scanline
...
 
 int32_t mc6847_base_device::scanline_position_from_clock(int32_t clocks_since_hsync)
 {
-	return (clocks_since_hsync - 20) / 4;
+	// This value determined by experimentation with the Dragon Fire game,
+	// which uses precise timing measurements to change video modes
+	// multiple times mid-line.  This value seems to minimize video
+	// garbage, but doesn't eliminate it, so there are likely timing
+	// issues elsewhere.
+	//
```
angelosa, line 813: Dragon Fire is under `drgnfire` in `coco_cart.xml, please use the romset ID here. I otherwise give it a shot later, I presume it's main system waitstates?

src/devices/video/mc6847.cpp
```diff
@@ -1,4 +1,4 @@
-// license:BSD-3-Clause
+// license:BSD-3-Clause
```
startaq, line 1: You can avoid spurious changes like this by running `srcclean` on the file.

src/devices/video/mc6847.cpp
```diff
@@ -559,11 +588,23 @@ void mc6847_base_device::device_config_complete()
...
-		screen().set_vblank_time(0);
+		// It is preferred to initialize the screen device using set_raw.
+		// Setting screen device's raw parameters requires values in terms of pixels.
+		// Multiplying the 6847's clock by 2 produces a pixel clock that allows
+		// 256 pixels of active video area per line, with apparently no roundoff
+		// error or drift.
+		//
+		// Note: This math is valid for NTSC.  If / when PAL is implemented, m_tpfs
+		// can be used to determine if this 6847 is PAL, and a different set_raw
+		// call can be made if helpful.
```
angelosa, line 599: I wonder what the real 6847 really do in order to pinpoint that is a PAL machine?
From quick glance from the datasheets the VDG seems coupled with a MC1372, which is what drives the clock on a NTSC machine, no idea about PAL atm but regardless I guess that `m_tpfs` is some kind of cheap simplification that should be removed in the end.

src/devices/video/mc6847.cpp
```diff
@@ -76,25 +76,42 @@
...
+#define TIMER_HSYNC_OFF_TIME			(CLOCKS_ACTIVE_VIDEO_AND_BORDERS + CLOCKS_FRONT_PORCH_DURATION)
+#define TIMER_HSYNC_ON_TIME				(TIMER_HSYNC_OFF_TIME + 16.5)
+
+// FSYNC occurs just after active video, just before right border
+#define TIMER_FSYNC_TIME				(CLOCKS_ACTIVE_VIDEO + CLOCKS_L_OR_R_BORDER)
+
+// These units are the screen device's bitmap pixels.  Multiplied by 2 because
+// the pixel clock is the mc6847's clock * 2
+#define BMP_L_OR_R_BORDER				(CLOCKS_L_OR_R_BORDER * 2)
+#define BMP_ACTIVE_VIDEO				(CLOCKS_ACTIVE_VIDEO * 2)
```
cuavas, line 106: Can you use `constexpr` in an anonymous namespace rather than macros?  It’s a lot nicer for compiler error messages and debugging symbols.

src/devices/video/mc6847.cpp
cuavas, line 437: Besides the comment being pointless, this line is ridiculously long and has redundant parentheses.

src/devices/video/mc6847.cpp
```diff
@@ -559,11 +588,23 @@ void mc6847_base_device::device_config_complete()
...
+		// can be used to determine if this 6847 is PAL, and a different set_raw
+		// call can be made if helpful.
+		screen().set_raw(
+			(u32)(clock() * 2),
+			456,										// htotal
+			0,											// hbend
+			BMP_L_OR_R_BORDER * 2 + BMP_ACTIVE_VIDEO,	// hbstart
+			m_tpfs,										// vtotal
+			0,											// vbend
+			LINES_UNTIL_RETRACE);						// vbstart
```
cuavas, line 607: The rest of this file doesn’t use the abbreviated integer types.  Please follow the pattern.

src/devices/video/mc6847.cpp
```diff
@@ -161,9 +178,9 @@ void mc6847_friend_device::device_start()
 	m_hsync_off_timer = timer_alloc(FUNC(mc6847_friend_device::change_horizontal_sync), this);
 	m_fsync_timer = timer_alloc(FUNC(mc6847_friend_device::change_field_sync), this);
 
-	m_frame_timer->adjust(clocks_to_attotime(0), 0, clocks_to_attotime(m_tpfs * TIMER_HSYNC_PERIOD * m_divider));
-	m_hsync_on_timer->adjust(clocks_to_attotime(TIMER_HSYNC_ON_TIME * m_divider), 1, clocks_to_attotime(TIMER_HSYNC_PERIOD * m_divider));
-	m_hsync_off_timer->adjust(clocks_to_attotime(TIMER_HSYNC_OFF_TIME * m_divider), 0, clocks_to_attotime(TIMER_HSYNC_PERIOD * m_divider));
+	m_frame_timer->adjust(clocks_to_attotime(0), 0, clocks_to_attotime(m_tpfs * CLOCKS_HSYNC_PERIOD * m_divider));
+	m_hsync_on_timer->adjust(clocks_to_attotime(TIMER_HSYNC_ON_TIME * m_divider), 1, clocks_to_attotime(CLOCKS_HSYNC_PERIOD * m_divider));
+	m_hsync_off_timer->adjust(clocks_to_attotime(TIMER_HSYNC_OFF_TIME * m_divider), 0, clocks_to_attotime(CLOCKS_HSYNC_PERIOD * m_divider));
```
cuavas, line 183: `clocks_to_attotime` takes an integer parameter, but your `TIMER_HSYNC_ON_TIME` is a `double`.  Playing fast and loose with integer to floating point conversions leads to trouble.  It also won’t work with devices derived from `mc6847_base_device` which initialise `m_divider` as 1 as the 0.5 will get truncated away.

src/devices/video/mc6847.cpp
dave-br: Thanks for the info about the ROM.  I'd been using the disk version from here: https://colorcomputerarchive.com/repo/Disks/Games/Dragonfire%20(Tandy).zip (LOADM + EXEC).  I will look for the ROM and try it out next.  I haven't tried disassembling Dragonfire to see what it's doing, but debugging MAME itself (C++ debugger), I can see that Dragonfire uses the hsync ON interrupt.  My guess is that it responds to that interrupt by clearing a flag somewhere so that the mainline code is released at a known time to do its meticulously-timed processing for the next horizontal line.  Just a guess, though.

src/devices/video/mc6847.cpp
dave-br: Yeah, I wish I had a good answer to that, but I'm brand new to all of this (hardware and MAME).  The original MC6847.cpp code switched off of m_tpfs when setting the refresh rate.  I wanted to preserve that approach in a comment when I changed the code to use set_raw, so that future changes will at least have a starting point to consider.

src/devices/video/mc6847.cpp
dave-br: Nice catch.  m_divider is indeed 1 for the coco.  I think what I need is a clocks_to_attotime which multiplies its argument by m_attoseconds_per_clock using double arithmetic and then casts to an int.  I don't see anything like that--would you advise I create such a thing in device_t?  (Or is that generally too dangerous to legitimize at such a level?)

src/devices/video/mc6847.cpp
dave-br: Tried the cartridge version from https://colorcomputerarchive.com/repo/Cartridges/Dragon%20Fire%20(1984)%20(26-3098)%20(Tandy).ccc.  Performs exactly the same as the disk version before my change (moving & shimmering garbage), and the same as the disk version after my change (nearly no garbage).  Will add "drgnfire" to the comment.

src/devices/video/mc6847.cpp
dave-br: Done.

src/devices/video/mc6847.cpp
dave-br: Makes sense.  I've removed the comment.

src/devices/video/mc6847.cpp
dave-br: I've added "[Dragon Fire listed in coco_cart.xml as drgnfire.]" to the end of the comment.

src/devices/video/mc6847.cpp
dave-br: Nice, ran the tool this time.

src/devices/video/mc6847.cpp
dave-br: Agreed!  Done.

src/devices/video/mc6847.cpp
dave-br: yup.  Comment is gone, and the LOGMASKED call has been broken into multiple lines.

src/devices/video/mc6847.cpp
dave-br: Indeed!  Changed.

src/devices/video/mc6847.cpp
dave-br: Actually, I'm changing 16.5 -> 16 altogether, and explicitly marking as int.  I tried the double arithmetic (multiplying TIMER_HSYNC_ON_TIME * m_attoseconds_per_clock) before explicitly casting, and that threw the timing off for Dragonfire.  I tried several adjustments, but interestingly can't do better than just straight int arithmetic with 16, which is what the code was inadvertently doing anyway.

In fact, it's not clear to me from the data sheet that 16.5 was the correct value to begin with.  There are small factors beyond 16.5 that affect HS width, including HS falling & rising time.  I'm likely missing something here, but 16 seems to be the most accurate number.

src/devices/video/mc6847.cpp
angelosa: That's even more redundant now, just referring it as `coco_cart:drgnfire` or `drgnfire (coco_cart)` is enough. 
It's a QA convention, adding Long Name Game Or Software Title as a comment is not pretty to look at, plus it will eventually become outdated the moment somebody takes this (and PC-6001) to a serious testing level.

dave-br: I see.  Ok, I've replaced "dragon fire" with coco_cart:drgnfire.  There was also a prior comment at the top of the file that had referred to "dragon fire", so I replaced that as well.

dave-br:
Thanks for all the feedback so far.  Please let me know if there is anything else you need from me.  I believe I've addressed all feedback so far, but just wanted to double-check you weren't waiting on me for anything.

angelosa:
I'm gonna apply this if/when CI passes, been forever I've been meaning to QA this system ...

dave-br:
Cool, thanks for the update.

PR #11597, opened 2023-10-04
Update Turkish(Türkçe) strings.po

kadireksi:


language/Turkish/strings.po
```diff
@@ -243,29 +254,30 @@ msgid ""
 "%s\n"
 " removed from favorites list."
 msgstr ""
+" favori listesinden kaldırıldı."
```
cuavas, line 257: You missed the line with `"%s\n"` here – it will cause the name of the system to be missing from the message.

language/Turkish/strings.po
```diff
@@ -843,27 +855,27 @@ msgstr "yazıcı"
 #: src/frontend/mame/ui/info.cpp:55
 msgctxt "emulation-feature"
 msgid "magnetic tape"
-msgstr ""
+msgstr "manyetik kaset"
```
cuavas, line 858: Is this translation appropriate for all forms of magnetic tape?  For example there were open reel tape formats (e.g. IBM [9-track tape](https://en.wikipedia.org/wiki/9-track_tape)) as well as cassette formats.

cuavas (commented):
Thanks for this update.  I noticed one error and have one question.

language/Turkish/strings.po
kadireksi: as you can guess, both "magnetic" and "tape"  doesn't have a real Turkish translation. We mostly call the tape(cassette) player as "teyp" and the "tape" itself is called "kaset" which also comes from "cassette". So it shouldn't be a problem.

language/Turkish/strings.po
```diff
@@ -1980,6 +2034,12 @@ msgid ""
 "If this is your first time using %2$s, please see the %2$s.pdf file in the "
 "docs folder for information on setting up and using %2$s."
 msgstr ""
+"Sistem ROM'u bulunamadı. Lütfen %1$s.ini 'de belirtilen rompath ayarını "
+"kontrol edin.\n"
+"\n"
+"%2$s'yi ilk kez kullanıyorsanız, lütfen %2$s'yi kurma ve kullanma hakkında "
+"bilgi için dokümanlar klasöründeki %2$s.pdf dosyasına bakın."
```
cuavas, line 2041: Oh, one other thing.  In this message, “docs” is the name of the folder containing the documentation when you download and extract MAME.  The word “docs” should be preserved literally in the translated message, as the name of the folder will not be localised.

cuavas (commented):
Sorry, after reading through the changes again, I noticed one other thing.

language/Turkish/strings.po
kadireksi: fixed it

kadireksi: also thanks from me :)
i'll try to finish the rest in month...

PR #11598, opened 2023-10-05
isa/sblaster_lle.cpp: Add LLE Sound Blaster 2.0 device

Luigi30:
Adds a separate LLE Sound Blaster device with support for DSP 2.02, which was recently dumped.

### DIFF
--- a/src/devices/bus/isa/sblaster_lle.cpp
+++ b/src/devices/bus/isa/sblaster_lle.cpp
@@ -1,0 +1,381 @@
+// license:BSD-3-Clause
+// copyright-holders:Katherine Rohl
+/***************************************************************************
+    Sound Blaster LLE device
+    - The "DSP" is an 8051.
+
+    Models:
+    - CT1310 (Sound Blaster 1.0)
+        - In Creative Labs product literature but may not have been released
+        - OPL2, 2x SAA1099, gameport, 22KHz playback/12KHz record
+
+    - CT1320 (Sound Blaster 1.5)
+        - DSP p/n CT1321, version 1.5
+        - Cost-reduced CT1310
+        - OPL2, 2xSAA1099 are now optional, gameport, 22KHz playback/12KHz record
+
+    - CT1350 (Sound Blaster 2.0)
+        - DSP p/n CT1351, version 2.0
+        - OPL2, no SAA1099, 44KHz playback/22KHz record, gameport, auto-initializing DMA
+        - If DSP 2.0x is placed in a CT1320, it becomes functionally equivalent to a CT1350.
+***************************************************************************/
+
+#include "emu.h"
+#include "sblaster_lle.h"
+
+//#define VERBOSE 1
+#include "logmacro.h"
+
+//**************************************************************************
+//  GLOBAL VARIABLES
+//**************************************************************************
+
+#define ym3812_StdClock XTAL(3'579'545)
+#define ymf262_StdClock XTAL(14'318'181)
+#define XTAL_DSP        XTAL(12'000'000)    // XTL1
+
+DEFINE_DEVICE_TYPE(ISA8_SOUND_BLASTER_2_0_LLE, isa8_sblaster_2_0_lle_device, "isa8_sblaster_2_0", "Sound Blaster 2.0 (CT1350, LLE)")
+
+void isa8_sblaster_2_0_lle_device::device_add_mconfig(machine_config &config)
+{
+	I80C51(config, m_dsp, XTAL_DSP); // XTL1
+	m_dsp->set_addrmap(AS_PROGRAM, &isa8_sblaster_2_0_lle_device::map_dsp_program);
+	m_dsp->set_addrmap(AS_IO, &isa8_sblaster_2_0_lle_device::map_dsp_io);
+
+	// No inputs on ports 0 or 1
+	m_dsp->port_in_cb<2>().set(FUNC(isa8_sblaster_2_0_lle_device::dsp_port2_r));
+	m_dsp->port_in_cb<3>().set(FUNC(isa8_sblaster_2_0_lle_device::dsp_port3_r));
+	// No outputs on port 0
+	m_dsp->port_out_cb<1>().set(m_dac, FUNC(dac_byte_interface::data_w));               // DSP port 1 is hooked right up to the DAC.
+	m_dsp->port_out_cb<2>().set(FUNC(isa8_sblaster_2_0_lle_device::dsp_port2_w));
+	m_dsp->port_out_cb<3>().set(FUNC(isa8_sblaster_2_0_lle_device::dsp_port3_w));
+
+	PC_JOY(config, m_joy);
+
+	SPEAKER(config, m_speaker).front_center();
+	MC1408(config, m_dac, 0).add_route(ALL_OUTPUTS, m_speaker, 0.5);
+
+	YM3812(config, m_ym3812, ym3812_StdClock);
+	m_ym3812->add_route(ALL_OUTPUTS, m_speaker, 1.0);
+	/* no CMS */
+}
+
+static INPUT_PORTS_START( sb_dsw )
+	PORT_START("JP1")
+	PORT_CONFNAME( 0x03, 0x01, "Base Address")
+	PORT_CONFSETTING(    0x01, "220h")
+	PORT_CONFSETTING(    0x02, "240h")
+
+	PORT_START("JP4")
+	PORT_CONFNAME( 0x0f, 0x01, "IRQ Line")
+	PORT_CONFSETTING(    0x08, "IRQ2/9" )
+	PORT_CONFSETTING(    0x04, "IRQ3" )
+	PORT_CONFSETTING(    0x02, "IRQ5")
+	PORT_CONFSETTING(    0x01, "IRQ7")
+INPUT_PORTS_END
+
+ioport_constructor isa8_sblaster_2_0_lle_device::device_input_ports() const
+{
+	return INPUT_PORTS_NAME( sb_dsw );
+}
+
+isa8_sblaster_2_0_lle_device::isa8_sblaster_2_0_lle_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+	device_t(mconfig, ISA8_SOUND_BLASTER_2_0_LLE, tag, owner, clock),
+	device_isa8_card_interface(mconfig, *this),
+	m_dsp(*this, "dsp"),
+	m_joy(*this, "pc_joy"),
+	m_ym3812(*this, "ym3812"),
+	m_dac(*this, "dac"),
+	m_speaker(*this, "speaker")
+{
+}
+
+void isa8_sblaster_2_0_lle_device::device_start()
+{
+	set_isa_device();
+}
+
+void isa8_sblaster_2_0_lle_device::device_reset()
+{
+	uint16_t base_port = (ioport("JP1")->read() & 0x02) ? 0x240 : 0x220;
+
+	m_isa->install_device(0x0200, 0x0207,
+		read8smo_delegate(*subdevice<pc_joy_device>("pc_joy"), FUNC(pc_joy_device::joy_port_r)),
+		write8smo_delegate(*subdevice<pc_joy_device>("pc_joy"), FUNC(pc_joy_device::joy_port_w)));
+
+	m_isa->install_device(base_port + 0x6, base_port + 0x7,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_reset_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_reset_w)));
+	m_isa->install_device(base_port + 0x8, base_port + 0x9,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::ym3812_16_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::ym3812_16_w)));
+	m_isa->install_device(base_port + 0xa, base_port + 0xb,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_data_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_data_w)));
+	m_isa->install_device(base_port + 0xc, base_port + 0xd,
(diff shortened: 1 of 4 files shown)


src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+//#define VERBOSE 1
+#include "logmacro.h"
+
+//**************************************************************************
+//  GLOBAL VARIABLES
+//**************************************************************************
+
+#define ym3812_StdClock XTAL(3'579'545)
+#define ymf262_StdClock XTAL(14'318'181)
+#define XTAL_DSP        XTAL(12'000'000)    // XTL1
```
cuavas, line 35: Can you use `constexpr` for these rather than macros?  Also, be consistent and use `SCREAMING_SNAKE_CASE` for things that function like constants.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+	m_isa->install_device(base_port + 0xc, base_port + 0xd,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_wbuf_status_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_cmd_w)));
+	m_isa->install_device(base_port + 0xe, base_port + 0xf,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_rbuf_status_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::dsp_rbuf_status_w)));
+
+	m_isa->install_device(0x0388, 0x0389,
+		read8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::ym3812_16_r)),
+		write8sm_delegate(*this, FUNC(isa8_sblaster_2_0_lle_device::ym3812_16_w)));
```
cuavas, line 124: You need to set a flag after installing the handlers, so that if someone flips the base address jumper and resets, you don’t end up installed at multiple addresses.

Also, can you use `emu::rw_delegate` rather than the explicit types?

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+	set_isa_device();
+}
+
+void isa8_sblaster_2_0_lle_device::device_reset()
+{
+	uint16_t base_port = (ioport("JP1")->read() & 0x02) ? 0x240 : 0x220;
+
+	m_isa->install_device(0x0200, 0x0207,
+		read8smo_delegate(*subdevice<pc_joy_device>("pc_joy"), FUNC(pc_joy_device::joy_port_r)),
+		write8smo_delegate(*subdevice<pc_joy_device>("pc_joy"), FUNC(pc_joy_device::joy_port_w)));
```
cuavas, line 104: You’ve got an object finder for the joystick port.  Use it.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+	if(!m_irq_raised)
+	{
+		LOG("SB raising IRQ %d\n", 5);
+		switch(ioport("JP4")->read())
+		{
+			case 1: m_isa->irq7_w(ASSERT_LINE); break;
+			case 2: m_isa->irq5_w(ASSERT_LINE); break;
+			case 4: m_isa->irq3_w(ASSERT_LINE); break;
+			case 8: m_isa->irq2_w(ASSERT_LINE); break;
+		}
```
cuavas, line 177: Don’t do tag lookups on hot paths.  Use object finders.  Also, you should probably keep the interrupt number stored somewhere, otherwise someone could flip the jumper while it’s asserted, and then you’ll deassert the wrong IRQ.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+		lower_irq();
+		lower_dma();
+	}
+}
+
+uint8_t isa8_sblaster_2_0_lle_device::dsp_rbuf_status_r(offs_t offset)
+{
+	// 0x22E
+	lower_irq();                        // Reading this bit clears the IRQ.
+	return (m_dav_pc ? 0x80 : 0x00);    // Return the data waiting flag in bit 7.
```
cuavas, line 317: You need to guard against debugger accesses by checking `machine().side_effects_disabled()` before doing things with side effects in read handlers.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+// From the perspective of the ISA bus.
+uint8_t isa8_sblaster_2_0_lle_device::dsp_data_r(offs_t offset)
+{
+	uint8_t data = m_dsp_to_host_latch;     // PC retrieves the latched byte.
+	LOG("PC reading DSP data latch: D:%02X\n", data);
+	m_dav_pc = false;                       // Clear the PC has data flag.
+	m_dsp_to_host_latch = 0;                // Clear the latch.
+
+	return data;
```
cuavas, line 334: Same here, you need to guard against debugger accesses.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,381 @@
...
+}
+
+// From the perspective of the DSP.
+uint8_t isa8_sblaster_2_0_lle_device::dsp_latch_r(offs_t offset)
+{
+	uint8_t data = m_host_to_dsp_latch;                 // DSP retrieves the latched byte.
+	m_host_to_dsp_latch = 0x00;                         // Clear the latch.
+	LOG("DSP reading a byte from PC: D:%02X\n", data);
+	m_dav_dsp = false;                                  // Clear the DSP has data flag.
+	return data;
```
cuavas, line 352: Another read with side effects.

src/devices/bus/isa/sblaster_lle.h
```diff
@@ -0,0 +1,91 @@
...
+
+#pragma once
+
+#include "isa.h"
+#include "bus/pc_joy/pc_joy.h"
+#include "cpu/mcs51/mcs51.h"
+#include "sound/dac.h"
+#include "sound/ymopl.h"
+#include "sound/spkrdev.h"
+#include "speaker.h"
```
cuavas, line 14: Do you need to `#include "speaker.h"` here?  I think it’s only needed in the implementation file.

src/devices/bus/isa/sblaster_lle.h
```diff
@@ -0,0 +1,91 @@
...
+	uint8_t dsp_wbuf_status_r(offs_t offset);
+	uint8_t dsp_rbuf_status_r(offs_t offset);
+
+	void    dsp_reset_w(offs_t offset, uint8_t data);
+	void    dsp_data_w(offs_t offset, uint8_t data) {}
+	void    dsp_cmd_w(offs_t offset, uint8_t data);
+	void    dsp_rbuf_status_w(offs_t offset, uint8_t data) {}
+
+	void    map_dsp_program(address_map &map);
+	void    map_dsp_io(address_map &map);
```
cuavas, line 41: I’m pretty sure these don’t need to be public.

cuavas (changes requested):

Luigi30:
OK, addressed the review issues.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+	if(m_installed_base_io != 0)
+	{
+		// Unmap if already mapped.
+		m_isa->unmap_device(m_installed_base_io+0x6, m_installed_base_io+0x7);
+		m_isa->unmap_device(m_installed_base_io+0x8, m_installed_base_io+0x9);
+		m_isa->unmap_device(m_installed_base_io+0xa, m_installed_base_io+0xb);
+		m_isa->unmap_device(m_installed_base_io+0xc, m_installed_base_io+0xd);
+		m_isa->unmap_device(m_installed_base_io+0xe, m_installed_base_io+0xf);
+	}
```
cuavas, line 132: It isn’t safe to do this, because there’s nothing to stop some other device installing itself over the top of your handlers in the mean time, and this will unmap that as well.  The least bad thing to do is to only install handlers the first time, requiring a hard reset to change the base address, etc.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -0,0 +1,415 @@
...
+}
+
+// From the perspective of the ISA bus.
+uint8_t isa8_sblaster_2_0_lle_device::dsp_data_r(offs_t offset)
+{
+	uint8_t data = m_dsp_to_host_latch;     // PC retrieves the latched byte.
+	LOG("PC reading DSP data latch: D:%02X\n", data);
+
+	if(!machine().side_effects_disabled())
+	{
```
cuavas, line 358: You should probably put the `LOG` inside the `if` to avoid spam if a debugger memory window is open.

cuavas (commented):
Given the way MAME works, it really isn’t worth trying to reinstall at a different address on reset – just keep track of whether you’ve installed your handlers, and don’t install again.

Also, I noticed you have no synchronisation barriers on communication between the host CPU and the microcontroller on the card.  Is there potential for this to cause issues?  Should it take a trip through the scheduler and/or temporarily reduce the scheduling quantum when this happens?

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -52,12 +53,27 @@ void isa8_sblaster_2_0_lle_device::device_add_mconfig(machine_config &config)
...
+	if(m_cms_present)
+	{
+		// FM and DAC are mono, but CMS is stereo.
+		SAA1099(config, m_saa1099_1, CLOCK_SAA1099);
+		m_saa1099_1->add_route(0, m_lspeaker, 0.5);
+		m_saa1099_1->add_route(1, m_rspeaker, 0.5);
+		SAA1099(config, m_saa1099_2, CLOCK_SAA1099);
+		m_saa1099_2->add_route(0, m_lspeaker, 0.5);
+		m_saa1099_2->add_route(1, m_rspeaker, 0.5);
+	}
```
cuavas, line 76: This can’t possibly work.  Machine config happens before device start/reset, and can’t depend on the state of I/O ports anyway (and if it could, it would wreak havoc with stuff like `-listxml` and `-listroms`).

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -191,29 +242,32 @@ void isa8_sblaster_2_0_lle_device::raise_irq()
...
 		{
-			case 1: m_isa->irq7_w(ASSERT_LINE); m_irq_raised = 7; break;
-			case 2: m_isa->irq5_w(ASSERT_LINE); m_irq_raised = 5; break;
-			case 4: m_isa->irq3_w(ASSERT_LINE); m_irq_raised = 3; break;
-			case 8: m_isa->irq2_w(ASSERT_LINE); m_irq_raised = 2; break;
+			case 7: m_isa->irq7_w(ASSERT_LINE); m_irq_raised = 7; break;
+			case 5: m_isa->irq5_w(ASSERT_LINE); m_irq_raised = 5; break;
+			case 3: m_isa->irq3_w(ASSERT_LINE); m_irq_raised = 3; break;
+			case 2: m_isa->irq2_w(ASSERT_LINE); m_irq_raised = 2; break;
 		}
```
cuavas: Wouldn’t it be better to `m_irq_raised = m_selected_irq` after the `switch` statement?  It would definitely be less clutter.

src/devices/bus/isa/sblaster_lle.cpp
```diff
@@ -404,9 +459,27 @@ void isa8_sblaster_2_0_lle_device::map_dsp_io(address_map &map)
...
+
+void isa8_sblaster_2_0_lle_device::saa1099_1_16_w(offs_t offset, uint8_t data)
+{
+	m_saa1099_1->write(offset, data);
+}
+
+void isa8_sblaster_2_0_lle_device::saa1099_2_16_w(offs_t offset, uint8_t data)
+{
+	m_saa1099_2->write(offset, data);
+}
```
cuavas, line 475: Are these trampolines needed?  Is there a reason you can’t install the device handlers directly?

src/devices/bus/isa/sblaster_lle.h
```diff
@@ -11,6 +11,7 @@
 #include "sound/dac.h"
 #include "sound/ymopl.h"
 #include "sound/spkrdev.h"
+#include "sound/saa1099.h"
```
cuavas: Please sort the headers you `#include`.

src/devices/bus/isa/sblaster_lle.cpp
Luigi30: The SAA1099s are write-only and m_isa->install_device doesn't handle that case, so I still need the read handler locally but I can change the write handler to hit the device directly.

src/devices/bus/isa/sblaster_lle.cpp
Luigi30: Hmm, the CMS is independent of everything else on the board, so I can leave it not hooked up if the machine config switch is disabled on a hard reset.

cuavas (commented):
I’m still a bit concerned about the lack of synchronisation barriers.  The thing is, you’ll have at least two “CPUs” in the system – the host CPU and the “DSP” microcontroller on the sound card.

With the code the way it is, you can end up with time travel effects where things that need to appear at the same time will appear at substantially different times.

For example suppose the following happens:
* `dsp_latch_w` sets `m_dsp_to_host_latch` and `m_dav_pc`
* `dsp_port3_w` raises the interrupt output

Now because the interrupt is implemented as a `device_execute_interface` input line, it won’t be set directly, it will take a trip through the scheduler, *then* set the input line.  The events will be something like this:
* `m_dsp_to_host_latch` and `m_dav_pc` set
* Scheduler ends DSP’s time slice as quickly as possible
* Scheduler allows main CPU to “catch up” with DSP – new values of `m_dsp_to_host_latch` and `m_dav_pc` visible at the *beginning* of this time slice
* Main CPU interrupt line asserted, main CPU sees interrupt

The issue is that the CPU could potentially see the interrupt substantially later than it sees the updated data and status values from the DSP.

Are you sure this won’t cause issues?  I’d feel better if you were using `machine().scheduler().synchronize(...)` on things that cross scheduled device boundaries to make so-called time travel issues less likely.

PR #11599, opened 2023-10-05
saturn.xml: Metadata cleaning

ArcadeShadow:
Replaced countries' and "Euro" abbreviations

### DIFF
--- a/hash/saturn.xml
+++ b/hash/saturn.xml
@@ -30130,7 +30130,7 @@ Sega logo animation draws offset [VDP2]
 
 	<!-- Identifying 3D Lemmings (E)... -->
 	<software name="3dlemmin" supported="no">
-		<description>3D Lemmings (Euro)</description>
+		<description>3D Lemmings (Europe)</description>
 		<year>1996</year>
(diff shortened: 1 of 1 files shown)


hash/saturn.xml
```diff
@@ -34347,7 +34347,7 @@ Sega logo animation draws offset [VDP2]
 
 	<!-- Identifying Saturn Power N.1 EUR [610-6576] - Core Demo Disc EUR [790-0002-50]... -->
 	<software name="satpowr1" supported="no">
-		<description>Saturn Power No. 1 (Euro) ~ Core Demo Disc (Euro)</description>
+		<description>Saturn Power No. 1 (Europe) ~ Core Demo Disc (Europe)</description>
```
angelosa, line 34350: This has a duplicated dual tag.

angelosa (approved):

hash/saturn.xml
cuavas: It’s because it’s two titles used in Europe both jammed into the description.  One of them should be moved to an `alt_title`.

PR #11602, opened 2023-10-05
Update Turkish strings.po (231005_2315)

kadireksi:


language/Turkish/strings.po
```diff
@@ -5876,31 +5899,31 @@ msgstr ""
...
 
 #: plugins/cheatfind/init.lua:737
 msgid "Slot 1 Value"
-msgstr ""
+msgstr "Slot 1 ,değeri"
 
 #: plugins/cheatfind/init.lua:737
 msgid "Last Slot Value"
-msgstr ""
+msgstr "Son Slot Değeri"
```
cuavas, line 5922: The word “Slot” is used inappropriately in the English UI for the cheat finder plugin, and I don’t want this carried over to translations unnecessarily.

The way the cheat finder works is that it saves the contents of memory, and then scans for values that changed or didn’t change between the saved memory snapshots.  Right now, it refers to these saved snapshots of memory as “slots” in English.  This is a poor choice of terminology, since it doesn’t describe what they are, and it can cause confusion with what MAME calls slot devices.

Can you use a better term than “slot” in Turkish to refer to these snapshots of memory?

language/Turkish/strings.po
```diff
@@ -6148,122 +6171,122 @@ msgstr ""
...
 msgctxt "plugin-inputmacro"
 msgid "Prolong step %d"
-msgstr ""
+msgstr "Uzatmalı adım %d"
 
 #: plugins/inputmacro/inputmacro_menu.lua:381
 msgctxt "plugin-inputmacro"
 msgid "When held"
-msgstr ""
+msgstr "basılı tutunca"
```
cuavas, line 6273: Should this have an initial uppercase letter like the other menu items in this plugin?

language/Turkish/strings.po
```diff
@@ -6341,72 +6364,72 @@ msgstr ""
 #: plugins/timer/init.lua:32
 msgctxt "plugin-timer"
 msgid "Wall clock"
-msgstr ""
+msgstr "Duvar saati"
```
cuavas, line 6367: In this context, “Wall clock” refers to elapsed real time using an independent clock/time reference.  Can you confirm that the Turkish translation is appropriate?

language/Turkish/strings.po
```diff
@@ -3226,58 +3232,58 @@ msgstr "Genel Bilgi"
...
 #, c-format
 msgid "Manufacturer\t%1$s\n"
-msgstr ""
+msgstr "Üretici\t%1$s\n"
 
 #: src/frontend/mame/ui/selmenu.cpp:2944
 #, c-format
 msgid "System is Clone of\t%1$s\n"
-msgstr ""
+msgstr " Sistem \t%1$s'in klonudur\n"
```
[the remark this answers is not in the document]
kadireksi: fixed

language/Turkish/strings.po
```diff
@@ -4345,12 +4368,12 @@ msgstr ""
...
 msgctxt "input-name"
 msgid "High"
-msgstr ""
+msgstr "Yüksek"
 
 #: src/emu/inpttype.ipp:116
 msgctxt "input-name"
 msgid "Low"
-msgstr ""
+msgstr "Düşük"
```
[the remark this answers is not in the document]
kadireksi: never played gambling games in Mame (and real life) .
"3 is higher than 2" = "3 2'den yüksek"
"2 is lower than 3"  = "2 3'ten düşük"
in this case translation is correct.
We can also call "3 2'den büyüktür" = "3 is bigger than 2".
I can test it later with gambling games(i also have to test all translations because sometimes even the translations are correct they don't fit the used situation)

language/Turkish/strings.po
```diff
@@ -5167,202 +5190,202 @@ msgstr ""
...
 msgctxt "input-name"
 msgid "Mouse X 4"
-msgstr ""
+msgstr "Fare x 4"
 
 #: src/emu/inpttype.ipp:844
 msgctxt "input-name"
 msgid "Mouse X 5"
-msgstr ""
+msgstr "Fare x 5"
```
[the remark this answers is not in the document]
kadireksi: firstly wrote wrong case, then copy paste ^^'

language/Turkish/strings.po
kadireksi: fixed, last time i used cheat finder was 2004 when i found the cheats for "Twin Eagle" for Mame 0.81 xD

language/Turkish/strings.po
kadireksi: fixed, most of the typo errors are because of resisting to wear near glasses :)

language/Turkish/strings.po
kadireksi: Will compile Mame tomorrow and check all the translations in place.
A noob question: do i have to make new pull request for the corrections?

cuavas: UI translation files are a bit outdated at the moment, they aren’t quite in sync with the application.  I need to get it back in sync, but I also need to review the Brazilian Portuguese updates before I can do that.  The “Source File” filter was added since the last time I got the UI translation files in sync, so it can’t be localised at the moment.  Sorry.

language/Turkish/strings.po
```diff
@@ -2441,7 +2443,7 @@ msgstr ""
 #, c-format
 msgctxt "gfxview"
 msgid " %1$d×%2$d origin (%3$d %4$d)"
-msgstr ""
+msgstr " %1$d×%2$d köken (%3$d %4$d)"
```
cuavas, line 2446: This is “origin” in the sense of the origin of a coordinate system.  From a quick look at Wikipedia, it suggests it may be called “sıfır noktası” or similar in Turkish?

kadireksi: you're right, fixed it as "sıfır noktası"

language/Turkish/strings.po
kadireksi: that's ok.
meanwhile will check in place transilations from time to time to fix any misuses.Thanks.

mascguy:
Folks, the following fatal errors occur, for the Turkey translation file:

```
language/Turkish/strings.po:1563: missing 'msgstr' section
language/Turkish/strings.po:1573: missing 'msgstr' section
language/Turkish/strings.po:3443:26: invalid control sequence
msgmerge: found 3 fatal errors
```

The errors for lines 1563 and 1573 look legitimate, as there appear to be missing a `msgstr` for both sections.

These potentially appear to be from a prior commit/PR though? Regardless, needs to be fixed.

As for line 3443, the control sequence does indeed look incorrect. Presumably the string should simply end with `\n`, rather than `\in`?

```
msgstr "WAN\tUygulanmadı\in"
```

PR #11603, opened 2023-10-06
Dumped and added a new Korean "Treasure Island" clone named "New HUNTer"

clawgrip:
New WORKING clone
---------------------------
New HUNTer [Taksangs]

Also reformatted single line comments.

### DIFF
--- a/src/mame/subsino/subsino.cpp
+++ b/src/mame/subsino/subsino.cpp
@@ -290,6 +290,7 @@ public:
 	void init_victor5();
 	void init_tisuba();
 	void init_tisubb();
+	void init_newhunter();
 	void init_sharkpye();
 	void init_tisub();
 	void init_mtrainnv();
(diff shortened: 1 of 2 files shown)


src/mame/subsino/subsino.cpp
```diff
@@ -3237,6 +3238,49 @@ ROM_START( tisubb )
 	ROM_LOAD( "n82s129n.u41", 0x200, 0x100, BAD_DUMP CRC(db99f6da) SHA1(d281a2fa06f1890ef0b1c4d099e6828827db14fd) )
 ROM_END
 
+ROM_START( newhunter )
+	// The MCU had its surface scratched out, so we cannot tell for sure if it has internal ROM or not
```
ajrhacker, line 3242: For the record, I believe that all HD647180X-based and H8/3044-based Subsino games actually do use the internal ROM, even though the present emulation forces them to start from external ROM instead. The patches applied to some of the HD647180X games are most likely a workaround for the MMU removing the internal ROM from the memory map after it performs some boot checks. The HD647180X's internal ROM might also be responsible for decrypting the encrypted part of external ROM code, which would then get stored in and executed from the 32Kx8 NVRAM.

PR #11604, opened 2023-10-08
PortAudio: sync to upstream

invertego:
Update PortAudio to pick up the following fix: https://github.com/PortAudio/portaudio/pull/846

Tested with the tiny subtarget on Windows, Linux (Ubuntu), and macOS.

In addition to updating the docs to remove the now obsolete workaround for MSVC builds, I went ahead and enabled PortAudio for Clang-CL. I'll note the Clang-CL build is not in a healthy state, so a few small changes were needed elsewhere to produce a successful build (not included in this PR).

Aside from the fix motivating this PR, not much of note has changed in PortAudio with one major exception: a new PulseAudio backend was introduced. Since MAME has its own PulseAudio backend I opted not to enable the PortAudio wrapper. I did try it out though, and for future reference, here is the change to enable it:

```
diff --git a/scripts/src/3rdparty.lua b/scripts/src/3rdparty.lua
index 2ded17d2ccf..b65cdd79576 100755
--- a/scripts/src/3rdparty.lua
+++ b/scripts/src/3rdparty.lua
@@ -1677,6 +1677,17 @@ project "portaudio"
                        MAME_DIR .. "3rdparty/portaudio/src/hostapi/alsa/pa_linux_alsa.c",
                        MAME_DIR .. "3rdparty/portaudio/src/hostapi/oss/pa_unix_oss.c",
                }
+               if _OPTIONS["NO_USE_PULSEAUDIO"]~="1" then
+                       defines {
+                               "PA_USE_PULSEAUDIO=1",
+                       }
+                       files {
+                               MAME_DIR .. "3rdparty/portaudio/src/hostapi/pulseaudio/pa_linux_pulseaudio_block.c",
+                               MAME_DIR .. "3rdparty/portaudio/src/hostapi/pulseaudio/pa_linux_pulseaudio.c",
+                               MAME_DIR .. "3rdparty/portaudio/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c",
+                               MAME_DIR .. "3rdparty/portaudio/src/common/pa_ringbuffer.c",
+                       }
+               end
        end
        if _OPTIONS["targetos"]=="macosx" then
                defines {
```


cuavas:
@galibert should we enable PortAudio’s PulseAudio backend for Linux?

rb6502:
I think it should be enabled  PA seems to be the best API for systems that use PipeWire (which is all the big names now).

PR #11605, opened 2023-10-09
subsino/subsino.cpp: Dumped and added a "New HUNTer" clone

clawgrip:
New WORKING clone
-------------------------
New HUNTer (bootleg) [Taksangs]

Also fixed the year for 'newhunter' set.

### DIFF
--- a/src/mame/subsino/subsino.cpp
+++ b/src/mame/subsino/subsino.cpp
@@ -3322,8 +3322,8 @@ ROM_START( newhunter )
 	ROM_LOAD( "ht-4.bin",     0x00345, 0x00117, CRC(f4f78925) SHA1(738281c05c0f51c0edcf65bacc18ebc87c2408c1) )
 ROM_END
 
-/*  The PCB have five 8-dipswitches bank sockets, but four of them were unpopulated (only one was present on the PCB).
-    The socket for the reset switch was also unpopulated on the PCB, and there wasn't memory backed RAM on the PCB
+/*  The PCB had five positions for banks of eight DIP switches, but four of them were unpopulated (only one was present on the PCB).
+    The position for the reset switch was also unpopulated, and there wasn't battery-backed RAM on the PCB
     (present on most games on this hardware). */
 ROM_START( newhunterb )
 	ROM_REGION( 0x04000, "mcu", 0 )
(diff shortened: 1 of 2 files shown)


src/mame/subsino/subsino.cpp
```diff
@@ -1020,6 +1023,37 @@ void subsino_state::tisub_map(address_map &map)
...
+	map(0x10000, 0x13fff).rom();
+
+	map(0x150c0, 0x150ff).ram().share("reel_scroll.2");
+	map(0x15140, 0x1517f).ram().share("reel_scroll.1");
+	map(0x15180, 0x151bf).ram().share("reel_scroll.0");
+
+	map(0x15800, 0x159ff).ram().w(FUNC(subsino_state::reel_ram_w<0>)).share("reel_ram.0");
+	map(0x15a00, 0x15bff).ram().w(FUNC(subsino_state::reel_ram_w<1>)).share("reel_ram.1");
+	map(0x15c00, 0x15dff).ram().w(FUNC(subsino_state::reel_ram_w<2>)).share("reel_ram.2");
+}
```
cuavas, line 1055: How different is this to the other address maps?  Can it be derived from a common map with just differences applied??

src/mame/subsino/subsino.cpp
```diff
@@ -3279,7 +3322,44 @@ ROM_START( newhunter )
 	ROM_LOAD( "ht-4.bin",     0x00345, 0x00117, CRC(f4f78925) SHA1(738281c05c0f51c0edcf65bacc18ebc87c2408c1) )
 ROM_END
 
+/*  The PCB have five 8-dipswitches bank sockets, but four of them were unpopulated (only one was present on the PCB).
+    The socket for the reset switch was also unpopulated on the PCB, and there wasn't memory backed RAM on the PCB
+    (present on most games on this hardware). */
```
cuavas, line 3327: “The PCB had five positions for banks of eight DIP switches, but for of them were unpopulated…”
“…there was no battery-backed RAM…”

(Assuming that’s what you meant – “memory backed RAM” doesn’t make a lot of sense.)

clawgrip: Fixed, thanks  :D

src/mame/subsino/subsino.cpp
clawgrip: @Osso13 Can you please help with this question?

Osso13: I'll have a look tomorrow (probably)

PR #11610, opened 2023-10-12
src/devices/sh: Numerous improvements [Ryan Holtz, Phil Bennett]

philipjbennett:
src/devices/sh: Numerous improvements [Ryan Holtz, Phil Bennett]
- General cleanup
- Renamed SH2A to SH7021
- Added (rudimentary) SH7021 DMA, timers and GPIO, and logging for the rest
- Fix NMIs occurring in a delay slot (interpreter only)

### DIFF
--- a/src/devices/cpu/sh/sh.cpp
+++ b/src/devices/cpu/sh/sh.cpp
@@ -169,7 +169,7 @@ void sh_common_execution::ADD(uint32_t m, uint32_t n)
  */
 void sh_common_execution::ADDI(uint32_t i, uint32_t n)
 {
-	m_sh2_state->r[n] += (int32_t)(int16_t)(int8_t)i;
+	m_sh2_state->r[n] += util::sext(i, 8);
 }
 
 /*  code                 cycles  t-bit
@@ -197,24 +197,16 @@ void sh_common_execution::ADDC(uint32_t m, uint32_t n)
  */
 void sh_common_execution::ADDV(uint32_t m, uint32_t n)
 {
-	int32_t dest, src, ans;
-
-	if ((int32_t) m_sh2_state->r[n] >= 0)
-		dest = 0;
-	else
-		dest = 1;
-	if ((int32_t) m_sh2_state->r[m] >= 0)
-		src = 0;
-	else
-		src = 1;
+	int32_t dest = BIT(m_sh2_state->r[n], 31);
+	int32_t src = BIT(m_sh2_state->r[m], 31);
 	src += dest;
+
 	m_sh2_state->r[n] += m_sh2_state->r[m];
-	if ((int32_t) m_sh2_state->r[n] >= 0)
-		ans = 0;
-	else
-		ans = 1;
+
+	int32_t ans = BIT(m_sh2_state->r[n], 31);
 	ans += dest;
-	if (src == 0 || src == 2)
+
+	if (src != 1)
 	{
 		if (ans == 1)
 			m_sh2_state->sr |= SH_T;
@@ -249,11 +241,9 @@ void sh_common_execution::ANDI(uint32_t i)
  */
 void sh_common_execution::ANDM(uint32_t i)
 {
-	uint32_t temp;
-
 	m_sh2_state->ea = m_sh2_state->gbr + m_sh2_state->r[0];
-	temp = i & RB( m_sh2_state->ea );
-	WB( m_sh2_state->ea, temp );
+	uint32_t temp = i & read_byte(m_sh2_state->ea);
+	write_byte(m_sh2_state->ea, temp);
 	m_sh2_state->icount -= 2;
 }
 
@@ -296,7 +286,7 @@ void sh_common_execution::BRA(uint32_t d)
 #if BUSY_LOOP_HACKS
 	if (disp == -2)
 	{
-		uint32_t next_opcode = RW(m_sh2_state->pc & AM);
+		uint32_t next_opcode = read_word(m_sh2_state->pc & m_am);
 		/* BRA  $
 		 * NOP
 		 */
@@ -407,7 +397,7 @@ void sh_common_execution::CMPEQ(uint32_t m, uint32_t n)
  */
 void sh_common_execution::CMPGE(uint32_t m, uint32_t n)
 {
-	if ((int32_t) m_sh2_state->r[n] >= (int32_t) m_sh2_state->r[m])
+	if ((int32_t)m_sh2_state->r[n] >= (int32_t)m_sh2_state->r[m])
 		m_sh2_state->sr |= SH_T;
 	else
 		m_sh2_state->sr &= ~SH_T;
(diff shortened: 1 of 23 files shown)


MooglyGuy:
Kindly direct most feedback on this one in my direction. Among the above listed changes, the general cleanup includes:
- A bracing consistency pass
- A spacing consistency pass
- A naming consistency pass (e.g. `RW` to `read_word`)

The actual changes needed to get the upcoming driver in good shape were fairly minimal, but I couldn't rely on that, so I hooked up stub functions with logging for the rest of the SH-7021 on-board devices. At some point, serial functionality will be beneficial, but is currently worked around.

src/devices/cpu/sh/sh.cpp
```diff
@@ -1547,68 +1508,51 @@ void sh_common_execution::SUBC(uint32_t m, uint32_t n)
...
-		if (ans == 1)
-			m_sh2_state->sr |= SH_T;
-		else
-			m_sh2_state->sr &= ~SH_T;
-	}
+
+	if (src == 1 && ans == 1)
+		m_sh2_state->sr |= SH_T;
 	else
 		m_sh2_state->sr &= ~SH_T;
```
cuavas, line 1523: This changes behaviour.

Previous behaviour:

| src | ans | action |
| --- | --- | --- |
| != 1 | != 1 | no-op |
| != 1 | == 1 | no-op |
| == 1 | != 1 | `m_sh2_state->sr &= ~SH_T` |
| == 1 | == 1 | `m_sh2_state->sr \|= SH_T` |

New behaviour:

| src | ans | action |
| --- | --- | --- |
| != 1 | != 1 | `m_sh2_state->sr &= ~SH_T` |
| != 1 | == 1 | `m_sh2_state->sr &= ~SH_T` |
| == 1 | != 1 | `m_sh2_state->sr &= ~SH_T` |
| == 1 | == 1 | `m_sh2_state->sr \|= SH_T` |

That is, previously it would only do the and/or if `src == 1`, while now it takes the `else` path whenever `src` is not equal to 1.

Is this intentional?

src/devices/cpu/sh/sh7021.cpp
```diff
@@ -1,164 +1,2169 @@
...
 	}
 
-	// channel enable & master enable
-	if((m_dma[ch].chcr & 1) == 0 || (m_dmaor & 1) == 0)
+	// Timer IRQs
+	for (uint32_t i = 0; i < 5; ++i)
+	{
+		int level;
+		if ((m_itu.timer[i].tier & m_itu.timer[i].tsr) & 7)
+		{
```
cuavas, line 576: Does `level` need to be declared outside the `if` block?  It seems to only be used inside.

src/devices/cpu/sh/sh7021.cpp
```diff
@@ -1,164 +1,2169 @@
...
+	LOGMASKED(LOG_SCI_RD, "%s:         Transmit Empty Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 7));
+	LOGMASKED(LOG_SCI_RD, "%s:         Receiver Full Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 6));
+	LOGMASKED(LOG_SCI_RD, "%s:         Overrun Error Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 5));
+	LOGMASKED(LOG_SCI_RD, "%s:         Framing Error Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 4));
+	LOGMASKED(LOG_SCI_RD, "%s:         Parity Error Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 3));
+	LOGMASKED(LOG_SCI_RD, "%s:         Transmit End Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 2));
+	LOGMASKED(LOG_SCI_RD, "%s:         Multiprocessor Bit Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 1));
+	LOGMASKED(LOG_SCI_RD, "%s:         Multiprocessor Bit Transfer Flag: %d\n", machine().describe_context(), Channel, BIT(m_sci[Channel].ssr, 0));
+	m_sci[Channel].ssr_read |= m_sci[Channel].ssr & 0xf8;
+	return (m_sci[Channel].ssr & 0xf9) | 0x04;
```
cuavas, line 1876: You should guard against debugger accesses in read handlers (`machine().side_effects_disabled()`).

cuavas (commented):
I noticed a few things.  As well as that, there are a few files that should be `srcclean`’d as there are tabs beyond the initial indent, so getting it done now will reduce churn.

(Tag @MooglyGuy so hopefully he sees the review.)

src/devices/cpu/sh/sh.cpp
philipjbennett: Hmm, does it? This is the previous code:

```
if (src == 1)
{
	if (ans == 1)
		m_sh2_state->sr |= SH_T;
	else
		m_sh2_state->sr &= ~SH_T;
}
else
	m_sh2_state->sr &= ~SH_T;
```

It was always doing `&= ~SH_T` if `src` was not equal to 1.

cuavas: Ah damn, I missed that the extra `else` branch was already there.

src/devices/cpu/sh/sh4dmac.cpp
```diff
@@ -518,43 +514,51 @@ void sh34_base_device::sh4_dma_ddt(struct sh4_ddt_dma *s)
...
+			{
+				uint32_t *p32bits = (uint32_t *)s->buffer;
+				for (uint32_t pos = 0; pos < s->length; pos++)
+				{
 					m_program->write_dword(s->destination, *p32bits);
 					p32bits++;
 					s->destination = s->destination + 4;
 				}
 			}
 		}
```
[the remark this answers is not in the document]
MooglyGuy: I suspect it was for brevity, and old C-style variable declaration. I've had a look through the small handful of spots where these DMA callbacks are used - all of them call `sh4_dma_ddt` while the `sh4_ddt_dma` struct passed in is a local variable. So, even if a faulty DMA were to result in enough register stomping to trigger a second DMA, the struct itself would contain unique data due to the stack frame being different.

cuavas (commented):
I found one more read handler with side effects that isn’t guarding against debugger accesses.

The only other thing I’d suggest is to guard against debugger accesses for logging anything on reads, otherwise it will spam like mad when a debugger memory window is open.  However, that wouldn’t be a deal-breaker for merging this, as the logging is disabled by default.

src/devices/cpu/sh/sh7021.cpp
```diff
@@ -1,164 +1,2173 @@
...
+	m_bsc.rcr = (uint8_t)data;
+}
+
+uint16_t sh7021_device::bsc_rtcsr_r()
+{
+	LOGMASKED(LOG_BSC_RD, "%s: Refresh Timer Control/Status Register, bsc_rtcsr_r: %04x\n", machine().describe_context(), m_bsc.rtcsr);
+	if (BIT(m_bsc.rtcsr, 7))
+		m_bsc.rtcsr_read = true;
+	return m_bsc.rtcsr;
+}
```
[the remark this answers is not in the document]
MooglyGuy: Fixed.

MooglyGuy:
> I found one more read handler with side effects that isn’t guarding against debugger accesses.
> 
> The only other thing I’d suggest is to guard against debugger accesses for logging anything on reads, otherwise it will spam like mad when a debugger memory window is open. However, that wouldn’t be a deal-breaker for merging this, as the logging is disabled by default.

Both fixed in a branch, waiting on Phil to merge it into the PR branch.

PR #11613, opened 2023-10-15
casio/cz101: promote to working

devinacker:
Systems promoted to working
----------------
Casio CZ-101 [Devin Acker]


src/mame/casio/ra3.h
```diff
@@ -0,0 +1,42 @@
...
+	u8 read(offs_t offset);
+	void write(offs_t offset, u8 data);
+
+	bool present() { return is_loaded(); }
+
+protected:
+	virtual void device_start() override;
+
+private:
+	uint8_t m_ram[0x1000];
```
cuavas, line 36: Should `m_ram` be dynamically allocated on start to reduce allocation size when instantiating the device for `-validate`, `-listxml`, etc.?

src/mame/casio/ra3.cpp
```diff
@@ -0,0 +1,72 @@
...
+/**************************************************************************/
+std::pair<std::error_condition, std::string> casio_ra3_device::call_load()
+{
+	if (length() != 0x1000)
+		return std::make_pair(image_error::INVALIDLENGTH, std::string());
+
+	fseek(0, SEEK_SET);
+	size_t ret = fread(m_ram, 0x1000);
+	if (ret != 0x1000)
+		return std::make_pair(image_error::UNSPECIFIED, "Error reading file");
```
cuavas, line 37: You can use `std::errc::io_error` or something.

src/mame/casio/ra3.cpp
```diff
@@ -0,0 +1,72 @@
...
+		return std::make_pair(image_error::UNSPECIFIED, "Error writing file");
+
+	return std::make_pair(std::error_condition(), std::string());
+}
+
+/**************************************************************************/
+u8 casio_ra3_device::read(offs_t offset)
+{
+	return is_loaded() ? m_ram[offset & 0xfff] : 0xff;
+}
```
cuavas, line 65: You could avoid the `is_loaded()` check here by filling `m_ram` with 0xff on unloading.

src/mame/casio/cz101.cpp
```diff
@@ -14,10 +28,14 @@
 #include "bus/midi/midioutport.h"
 #include "cpu/upd7810/upd7811.h"
 #include "machine/clock.h"
+#include "machine/nvram.h"
+#include "sound/upd933.h"
 #include "video/hd44780.h"
+#include "ra3.h"
```
cuavas, line 34: Please put the local project header `"ra3.h"` before the devices library headers.

src/mame/casio/cz101.cpp
```diff
@@ -105,153 +133,168 @@ void cz101_state::maincpu_map(address_map &map)
...
-	PORT_BIT(0x40, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("SELECT")
-	PORT_BIT(0x80, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("P")
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_KEYPAD) PORT_CODE(KEYCODE_E) PORT_NAME("Solo")
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_KEYPAD) PORT_CODE(KEYCODE_R) PORT_NAME("Tone Mix")
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_KEYPAD) PORT_CODE(KEYCODE_T) PORT_NAME("Key Transpose")
+	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_KEYPAD) PORT_CODE(KEYCODE_Y) PORT_NAME("Write")
+	PORT_BIT(0x10, IP_ACTIVE_LOW, IPT_KEYPAD) PORT_CODE(KEYCODE_U) PORT_NAME("MIDI")
+	PORT_DIPNAME(0x20, 0x20, "Memory Protect")
+	PORT_DIPSETTING(   0x20, DEF_STR( Off ))
+	PORT_DIPSETTING(   0x00, DEF_STR( On ))
```
cuavas, line 229: Is this a real DIP switch, or should it just be `PORT_TOGGLE`?

src/mame/casio/cz101.cpp
```diff
@@ -105,153 +133,168 @@ void cz101_state::maincpu_map(address_map &map)
...
 
+	PORT_START("PB")
+	PORT_BIT(0x0f, IP_ACTIVE_HIGH, IPT_UNUSED)
+	PORT_BIT(0x10, IP_ACTIVE_LOW,  IPT_CUSTOM) PORT_READ_LINE_DEVICE_MEMBER("upd933", upd933_device, rq_r)
+	PORT_BIT(0x60, IP_ACTIVE_HIGH, IPT_UNUSED)
+	PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_POWER_OFF) PORT_NAME("Power") PORT_TOGGLE PORT_CHANGED_MEMBER(DEVICE_SELF, cz101_state, power_w, 0)
+
 	PORT_START("AN1")
-	PORT_BIT(0xff, 0x7f, IPT_PADDLE) PORT_NAME("Pitch Wheel") PORT_SENSITIVITY(100) PORT_KEYDELTA(10) PORT_MINMAX(0x00, 0xff)
+	PORT_BIT(0xff, 0x7f, IPT_PADDLE) PORT_NAME("Pitch Wheel") PORT_SENSITIVITY(100) PORT_KEYDELTA(10) PORT_MINMAX(0x00, 0xff) PORT_CODE_DEC(KEYCODE_PGDN) PORT_CODE_INC(KEYCODE_PGUP)
```
cuavas, line 292: You don’t have “keyboard” inputs, but you’re using Page Down, which is the default “fast forward” UI control on Mac and Linux, as well as SDL builds for Windows.  That may irritate some people.

devinacker: Good point. I chose those sort of arbitrarily just so the defaults wouldn't conflict with the cursor/value keys. Is there a way to have an analog control not have any default inc/dec controls?

src/mame/casio/cz101.cpp
devinacker: It (and the APO switch) are just individual slide switches on the back/side of the unit, so I guess `PORT_TOGGLE` also makes sense.

src/mame/casio/cz101.cpp
cuavas: If there is a way, I can’t think of it right now.

src/mame/casio/cz101.cpp
cuavas: Using `PORT_TOGGLE` has the advantage that it can easily have a key assigned to it.

src/mame/casio/cz101.cpp
angelosa: I don't think there is, you may use a `JOYCODE` as a workaround?

PR #11617, opened 2023-10-15
netdev: Add support for the macOS 10.10+ vmnet virtual network interface [Kelvin Sherlock, R. Belmont]

rb6502:
vmnet is a virtual network passthrough API added in macOS 10.10 and enhanced in 10.15.  It works well for our purposes except that it requires root privileges.  Therefore this includes two implementations: one which works in the straightforward way but requires you to sudo mame, and one which builds a small helper program (when TOOLS=1) which can be set SUID root.  MAME then launches the helper program on startup and sends/receives data through it via pipes.  The helper program drops root as soon as the vmnet connection is open, so it should be all around better for security than sudo'ing MAME.

This was written originally by Kelvin Sherlock for Ample; I've done a fairly extensive cleanup pass on it and also found a few bugs in it which have been fixed.  It now works well for various systems with networking that I've tried; emulated Mac System 7 can mount my netatalk 2.2.10 server, access the outside Internet with FTP, HTTP, and SSH.

Limitations: 
1) vmnet does not pass EtherTalk packets (WiFi routers don't either).  The fix for that will be to support real bit-serial LocalTalk in the 8530 and a bitb-like adapter device that translates it into the community-defined LToUDP (LocalTalk over UDP) standard.  Many other emulators and a new LocalTalk WiFi dongle for real hardware ("AirTalk") now speak LToUDP.
2) vmnet null-routes all traffic until the emulated client sends a DHCP request.  Macs that can run Open Transport are fine, but MacTCP and some of the older workstations MAME emulates use bootp and will require some additional packet intervention code in vmnet_common to translate to and from DHCP.


cuavas:
Haven’t looked at the code yet, but one of the things on my (very long) TODO list is an HDLC async mode bridge device for emulated serial ports, which would allow bridging LocalTalk to other programs, as well as SNA, BITBUS, etc.

rb6502:
@cuavas I'd appreciate it if you could take a look at this when you have time.  I'm confident about the actual functionality (I use it all the time now) but if you have any ideas on making the chown/chmod dance easier or anything else I'd love to hear it.

src/osd/modules/netdev/vmnet.cpp
```diff
@@ -0,0 +1,280 @@
...
+		/* todo -- skip multicast messages based on flag? */
+		type = classify_mac(m_buffer);
+		if (type == 2) continue; /* multicast */
+		break;
+	}
+
+	if (count < 1) return 0;
+
+	osd_printf_verbose("MAC is %02x:%02x:%02x:%02x:%02x:%02x\n", m_mac[0], m_mac[1], m_mac[2], m_mac[3], m_mac[4], m_mac[5]);
+	if (memcmp(m_mac, m_vmnet_mac, 6) != 0) {
```
cuavas, line 249: This will get really spammy if there’s significant traffic.  Also, isn’t the only place the contents of `m_mac` can change in `set_mac`?  What’s the point of logging it on every received packet?

src/osd/modules/netdev/vmnet_common.h
cuavas, line 1: Please use `SCREAMING_SNAKE_CASE` for constants following the MAME convention.

src/osd/modules/netdev/vmnet_common.cpp
cuavas, line 1: This file appears to not be using one tab indent per scope level.

src/osd/modules/netdev/vmnet_common.cpp
```diff
@@ -0,0 +1,234 @@
...
+	&& packet[12] == 0x08 && packet[13] == 0x06 /* ARP */
+	&& packet[14] == 0x00 && packet[15] == 0x01 /* ethernet */
+	&& packet[16] == 0x08 && packet[17] == 0x00 /* ipv4 */
+	&& packet[18] == 0x06 /* hardware size */
+	&& packet[19] == 0x04 /* protocol size */
+  ;
+}
+
+int vmnet_common::is_broadcast(const uint8_t *packet, u32 size) {
+  return !memcmp(packet + 0, ff, 6);
```
cuavas, line 21: You should `#include <cstring>` if you’re going to use `memcmp` etc.

src/osd/modules/netdev/vmnet_common.cpp
```diff
@@ -0,0 +1,234 @@
...
+   */
+  if (length < ETHERNET_MIN_FRAME - 4)
+  {
+	std::fill_n(&buf[length], ETHERNET_MIN_FRAME - length - 4, 0);
+
+	length = ETHERNET_MIN_FRAME - 4;
+  }
+
+  // compute and append the frame check sequence
+  const u32 fcs = util::crc32_creator::simple(buf, length);
```
cuavas, line 225: You should `#include` the header for this.

src/osd/modules/netdev/vmnet_common.cpp
```diff
@@ -0,0 +1,234 @@
+// license:BSD-3-Clause
+// copyright-holders: Kelvin Sherlock, R. Belmont
+
+#include "emu.h"
+#include "vmnet_common.h"
```
cuavas, line 5: You shouldn’t `#include "emu.h"` here – we’re trying to reduce circular dependencies.  I realise it’s still needed in netdev modules, but it really shouldn’t be needed in a utility file like this.

cuavas (commented):
Before actually looking at how the communication works, there are a bunch of general things with the code:
* Changes to build scripts look wrong
* An info property list that lacks mandatory keys
* Depending on emu.h in src/osd when it should be avoided because of circular dependencies
* Not putting private implementation stuff in anonymous namespaces
* Variable scope not as tight as it could be
* Not using smart pointers to safely manage dynamically allocated stuff
* Non-`const` static variables
* Not including standard library headers for things that are used
* Why is the helper tool in C using static variables for everything?  C++ is more expressive.

It used to be possible to prompt the user for an admin username and password with Authorization Services, then use `AuthorizationExecuteWithPrivileges` to run a command as root (e.g. to set the owner and SUID bit on a helper tool), but it looks like Apple deprecated that: https://developer.apple.com/documentation/security/1540038-authorizationexecutewithprivileg

I’m not sure what the proper way to do this is now.

src/osd/modules/netdev/vmnet.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+	vmnet_module() : osd_module(OSD_NETDEV_PROVIDER, "vmnet"), netdev_module() {}
+	virtual ~vmnet_module() {}
+
+	virtual int init(osd_interface &osd, const osd_options &options);
+	virtual void exit();
+
+	virtual bool probe() {
+		return true;
+	}
```
cuavas, line 63: This module won’t work without appropriate entitlements.  Can you make this check for the `com.apple.vm.networking` entitlement or running as root, and return `false` if neither is the case?

You can check entitlements with something like:
```c++
	SecCodeRef me;
	if (SecCodeCopySelf(kSecCSDefaultFlags, &me) == errSecSuccess)
	{
		SecStaticCodeRef static_code;
		if (SecCodeCopyStaticCode(me, kSecCSDefaultFlags, &static_code) == errSecSuccess)
		{
			CFDictionaryRef signing;
			if (SecCodeCopySigningInformation(static_code, kSecCSDefaultFlags, &signing) == errSecSuccess)
			{
				CFDictionaryRef entitlements = CFDictionaryRef(CFDictionaryGetValue(signing, kSecCodeInfoEntitlementsDict));
				if (entitlements)
				{
					// value should be 1 if entitlement is available, but the key shouldn't be present at all otherwise
					if (CFDictionaryGetValue(entitlements, CFSTR("com.apple.vm.networking")))
					{
						// entitlement should be available if you get here.
					}
				}
				CFRelease(signing);
			}
			CFRelease(static_code);
		}
		CFRelease(me);
	}
```

You should be able to check if you’re running as root by getting the current uid/euid the usual UNIX way.

src/osd/modules/netdev/vmnet.cpp
```diff
@@ -0,0 +1,288 @@
...
+	uint8_t *m_buffer = 0;
+};
+
+
+netdev_vmnet::netdev_vmnet(const char *name, class device_network_interface *ifdev, int rate)
+	: osd_netdev(ifdev, rate) {
+
+	xpc_object_t dict;
+	dispatch_queue_t q;
+	dispatch_semaphore_t sem;
```
cuavas, line 95: You can use a custom deleted to specialise `std::unique_ptr` to automatically free the `xpc_object_t` and `dispatch_semaphore_t` on unwind.

cuavas:
There’s still a bunch of stuff from the first review.  I assume you or Kelvin are still looking at it.

src/osd/modules/netdev/vmnet.cpp
rb6502: We're forbidden from doing stuff like that at work because it's too complicated.  Is what I did to xpc_object_t even close?  It compiles, at least.

src/osd/modules/netdev/vmnet.cpp
rb6502: Done.

rb6502:
I addressed every bullet point from the original review except the smart pointer stuff (which we're now starting to get into here).  Please detail what else from that you feel isn't done.

src/osd/modules/netdev/vmnet.cpp
ksherlock: `xpc_object_t` and `dispatch_semaphore_t` are already pointers so you could

```
	std::unique_ptr<std::remove_pointer<xpc_object_t>::type, void (*)(xpc_object_t)> dict(
		xpc_dictionary_create( NULL, NULL, 0),
		[](xpc_object_t o) {
			if (o) xpc_release(o);
	});


	__block std::unique_ptr<std::remove_pointer<dispatch_semaphore_t>::type, void (*)(dispatch_semaphore_t)> sem(
		dispatch_semaphore_create(0),
		[](dispatch_semaphore_t s) {
			if (s) dispatch_release(s);
	});
```

Then use `dict.get()` / `sem.get()` to get back the pointers.  The semaphore needs to be `__block` (or moved out of the smart pointer with something like `dispatch_semaphore_t block_sem = sem.get()`) so it can be used in the block.

cuavas:
This is back to being up-to-date with mainline, so adventurous people can try it.  Next step is to do some mechanical cleanup, etc.

PR #11620, opened 2023-10-16
xavix.cpp - treat taikodp drum input as digital to avoid issue with multiple hits (same workaround as namcos10 taiko games)

mamehaze:


src/mame/tvgames/xavix.cpp
```diff
@@ -590,10 +590,20 @@ static INPUT_PORTS_START( taikodp )
 	PORT_INCLUDE(xavix_an)
 
 	PORT_MODIFY("AN0") // 00  (read by one type of function, handled in timer interrupt at 0x1d92 in RAM)
+#if 0 // should be more correct, but causes multiple hits to register
```
987123879113, line 593: Haven't looked at xavix before but based on my experience with the Namco System 10 Taiko games, most likely the reason it's causing multiple hits is because the pedal input will automatically autocenter on release, and each new value is being recognized as a new hit. Ideally drum inputs should be an analog value (to get the registered hit velocity) but should otherwise act more similar to a digital input (once released it immediately goes back to 0, no auto centering or anything like that).

Taiko itself doesn't seem to actually make use of the velocity values in gameplay besides just checking if the drum was hit or not. And the home controllers for Taiko are actually digital inputs fwiw.

PR #11622, opened 2023-10-16
Optimizations for Atmel AVR-8 core

MooglyGuy:
AVR8:
- Optimized general-timer and SPI-timer handling.
- Templatized timer accessor functions in order to flatten branches and reduce runtime load.
- Approximately 1.5x speedup; `uzebox frogger` which ran at 82-83% unthrottled now runs at 146-147% unthrottled. Other AVR-based drivers show a similar speedup.

### DIFF
--- a/src/devices/cpu/avr8/avr8.cpp
+++ b/src/devices/cpu/avr8/avr8.cpp
@@ -104,23 +104,23 @@
 
 enum
 {
-	AVR8_SREG_C = 0,
-	AVR8_SREG_Z,
-	AVR8_SREG_N,
-	AVR8_SREG_V,
-	AVR8_SREG_S,
-	AVR8_SREG_H,
-	AVR8_SREG_T,
-	AVR8_SREG_I,
+	SREG_C = 0,
+	SREG_Z,
+	SREG_N,
+	SREG_V,
+	SREG_S,
+	SREG_H,
+	SREG_T,
+	SREG_I,
 
-	AVR8_SREG_MASK_C = 0x01,
-	AVR8_SREG_MASK_Z = 0x02,
-	AVR8_SREG_MASK_N = 0x04,
-	AVR8_SREG_MASK_V = 0x08,
-	AVR8_SREG_MASK_S = 0x10,
-	AVR8_SREG_MASK_H = 0x20,
-	AVR8_SREG_MASK_T = 0x40,
-	AVR8_SREG_MASK_I = 0x80
+	SREG_MASK_C = 0x01,
+	SREG_MASK_Z = 0x02,
+	SREG_MASK_N = 0x04,
+	SREG_MASK_V = 0x08,
+	SREG_MASK_S = 0x10,
+	SREG_MASK_H = 0x20,
+	SREG_MASK_T = 0x40,
+	SREG_MASK_I = 0x80
 };
 
 // I/O Enums
@@ -176,12 +176,6 @@ enum
 	WGM5_FAST_PWM_OCR
 };
 
-static const char avr8_reg_name[4] = { 'A', 'B', 'C', 'D' };
-
-#define SREG_R(b)   ((m_r[SREG] & (1 << (b))) >> (b))
-#define SREG_W(b,v) m_r[SREG] = (m_r[SREG] & ~(1 << (b))) | ((v) << (b))
-#define NOT(x) (1 - (x))
-
 // Opcode-Parsing Defines
 #define RD2(op)         (((op) >> 4) & 0x0003)
 #define RD3(op)         (((op) >> 4) & 0x0007)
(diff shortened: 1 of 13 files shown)


src/mame/elektor/avrmax.cpp
```diff
@@ -52,8 +52,8 @@ For the German magazine, AVR-Max-Schachzwerg and CC2-Schachzwerg are in 06/2009.
 #include "screen.h"
 
 // internal artwork
-#include "avrmax.lh"
-#include "atm18mcc.lh"
+#include "avrmax.lh" // clickable
+#include "atm18mcc.lh" // clickable
```
happppp, line 57: It looks like you modified an older version of this file and copied it over? See the readded 'clickable' comment, and also the slightly modified comment at the top of the driver.

src/mame/handheld/pensebem.cpp
```diff
@@ -281,11 +281,11 @@ void pensebem2017_state::pensebem2017(machine_config &config)
...
-	m_maincpu->gpio_in<AVR8_IO_PORTB>().set(FUNC(pensebem2017_state::port_b_r));
-	m_maincpu->gpio_out<AVR8_IO_PORTB>().set(FUNC(pensebem2017_state::port_b_w));
-	m_maincpu->gpio_out<AVR8_IO_PORTC>().set(FUNC(pensebem2017_state::port_c_w));
-	m_maincpu->gpio_out<AVR8_IO_PORTD>().set(FUNC(pensebem2017_state::port_d_w));
-	m_maincpu->gpio_out<AVR8_IO_PORTE>().set(FUNC(pensebem2017_state::port_e_w));
+	m_maincpu->gpio_in<atmega88_device::GPIOB>().set(FUNC(pensebem2017_state::port_b_r));
+	m_maincpu->gpio_out<atmega88_device::GPIOB>().set(FUNC(pensebem2017_state::port_b_w));
+	m_maincpu->gpio_out<atmega88_device::GPIOC>().set(FUNC(pensebem2017_state::port_c_w));
+	m_maincpu->gpio_out<atmega88_device::GPIOD>().set(FUNC(pensebem2017_state::port_d_w));
+	m_maincpu->gpio_out<atmega88_device::GPIOE>().set(FUNC(pensebem2017_state::port_e_w));
```
felipesanches, line 288: Shouldn't these be `atmega168_device`?

rb6502:
For the Clang build failures, you probably need to explicitly instantiate the templates.

src/devices/cpu/avr8/avr8ops.hxx
```diff
@@ -1115,93 +1181,105 @@ void avr8_device::op_dec(uint16_t op)
...
 	SREG_W(AVR8_SREG_I, 1);
 }
 
-void avr8_device::op_sleep(uint16_t op)
+template <int NumTimers>
+void avr8_device<NumTimers>::op_sleep(uint16_t op)
 {
-	m_pc--;
+	m_pc = (m_pc << 1) - 2;
 }
```
cuavas, line 1228: Sorry, I assume I’m missing something here, but why does `op_sleep` need to adjust the program counter shift?  I’m assuming `op_ret` and `op_reti` need to because they’re converting a word address from the stack to an internal word address format, but `op_sleep` is just modifying the value in the program counter.  Why does it need adjustment in this case?

src/devices/cpu/avr8/avr8.h
```diff
@@ -33,21 +63,20 @@ class avr8_device : public cpu_device
...
 
 	// ADC
 	template<uint8_t Pin> auto adc_in() { return m_adc_in_cb[Pin].bind(); }
 
 protected:
-	avr8_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock, const device_type type, uint32_t address_mask, address_map_constructor internal_map, int32_t num_timers);
+	avr8_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock, const device_type type, uint32_t address_mask, address_map_constructor internal_map);
 
+	typedef void (avr8_device::*timer_func) ();
 	typedef void (avr8_device::*op_func) (uint16_t op);
```
cuavas, line 80: Would you get any performance gain from making `timer_func` a delegate rather than a member function pointer, and avoiding one conditional branch on invocation?

src/devices/cpu/avr8/avr8.cpp
```diff
@@ -730,44 +840,179 @@ atmega2560_device::atmega2560_device(const machine_config &mconfig, const char *
...
+	m_timer1_tick = m_timer1_ticks[0];
+	m_timer2_tick = m_timer2_ticks[0];
 }
 
+//-------------------------------------------------
+//  avr8_device - constructor
+//-------------------------------------------------
+template class avr8_device<2>;
+template class avr8_device<3>;
+template class avr8_device<6>;
```
cuavas, line 1001: Because compilers can be stupid sometimes, you probably need to move the explicit instantiations to the bottom of the file to fix the linker errors.

cuavas (commented):
How much of the AVR8 code depends on the number of timers?  Would it be possible to create a non-template base class to hold the opcode handlers that don’t depend on the number of timers, so those handlers don’t need to be instantiated for each number of timers?

Like

```c++
template <int NumTimers>
class avr8_device_base : public cpu_device
{
	// parts that don't depend on number of timers go here
};

template <int NumTimers>
class avr8_device : public avr8_device_base
{
	// parts that depend on timers, including main execute loop
};

// concrete device classes
```

cuavas:
> For the Clang build failures, you probably need to explicitly instantiate the templates.

The class templates *are* being instantiated.  The problem is that clang is sometimes dumb and requires the explicit instantiations to come after all the member definitions.  Moving them to the bottom of the file should fix it.

src/devices/cpu/avr8/avr8ops.hxx
MooglyGuy: Should be addressed now.

src/devices/cpu/avr8/avr8.h
MooglyGuy: Should be addressed now.

src/devices/cpu/avr8/avr8.cpp
MooglyGuy: Moved to the bottom of the file as suggested.

src/mame/handheld/pensebem.cpp
MooglyGuy: Good catch, thanks - fixed

MooglyGuy:
All feedback should be addressed now, and I ended up switching over to use delegates.

Tafoid:
Not 100% sure if this is a regression, improvement, or otherwise - display no longer shows nor audio tone plays for this machine:
replica1 - makerbot/replicator.cpp

replica1-diff

kEPoZRP:
No audio for hwchiptn and powernin - homebrew/lft_chiptune.cpp
No audio and video for phasor  - homebrew/lft_phasor.cpp

happppp:
pbem2017 is also broken. When I ran it recently, I paid no mind to it since it's a non-working driver, but it looks like it's in a better state in MAME 0.260.

happppp:
In case you don't have notifications on for the above messages, maybe with this?: @MooglyGuy

happppp:
It should be in better shape again after https://github.com/mamedev/mame/commit/fbabb96113d690ea562a3258cdf52823d799c1e2

replica1: display shows again, but no beep
phasor: works, but screen is now stable instead of wobbly
chiptunes: works
pbem2017: in same state as 0.260

PR #11628, opened 2023-10-18
Update homelab.cpp

linuxforum5:
I changed the original homelab.cpp because i need a workable Aircomp 16 emulator. The original Homaleb 2 code was buggy. Now, the new code can emulate all Homelab 2 (Aircomp 16) graphics capabilities - merged text and graphics mode, and changeable screen size. The ROM dump was NOT bad! The homelab 2 had not a video processor, the CPU generated the video signal - like ZX80. The hardware send special memory data in 0xE000-0xFFFF address range. The keyboard emulation had a bad code for right shift. The serial input (LOAD from tape) did not work, and the quickload can only load htp file with one record. Now the wav file load is working correctly, and it can load htp file with multiple record correctly. I tested the new emulator code with many different homelab2 programs, and the emulator is working perfectly.

### DIFF
--- a/src/mame/homelab/homelab.cpp
+++ b/src/mame/homelab/homelab.cpp
@@ -46,147 +46,233 @@ MB7051 - fuse programmed prom.
 #include "screen.h"
 #include "speaker.h"
 
-
 namespace {
 
 class homelab_state : public driver_device
 {
 public:
-	homelab_state(const machine_config &mconfig, device_type type, const char *tag)
-		: driver_device(mconfig, type, tag)
-		, m_maincpu(*this, "maincpu")
-		, m_bank1(*this, "bank1")
-		, m_p_chargen(*this, "chargen")
-		, m_speaker(*this, "speaker")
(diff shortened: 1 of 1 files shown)


cuavas:
`srcclean` won’t fix this – it’s been expanded to eight spaces per tab.  Please apply your changes again, with your editor set to have tabs at four-space intervals, and set to preserve tabs.  Also, please give this a coherent title and description of what you changed.  The title is currently not descriptive at all, and the description is an incoherent mess.

rb6502:
@linuxforum5 Are you able to clean this up so it can be submitted?  We'd love to have the improvements, but we'd like them formatted reasonably for the rest of the project too :-)

rb6502:
This has been landed in mainline for 0.263 (the end of February release).  I ran a pretty printer on this and then did fairly extensive additional work to get the driver near current MAME standards.  See https://github.com/mamedev/mame/commit/05b623dbe7e602ef015f00ea24407435321a7d26

PR #11629, opened 2023-10-18
jaleco/megasys1.cpp - use dumped MCU for bigstrik [Sergio Galiano, Recreativos Piscis, David Haywood]

mamehaze:
- updated tlcs90 port / irq handling to match hardware, implemented halt support [Sergio Galiano, Recreativos Piscis]
- hooked up MCU support in megasys1 driver [Sergio Galiano, Recreativos Piscis, David Haywood]
- started splitting megasys1 driver into subclasses [David Haywood]
- added 2 clones [unknown]

the clone sets have been kicking around in unemulated sets since the late 90s, probably Arcade ROMs from Japan or Romlist dumps / stuff that was sent to RAINEdev or similar.


src/mame/jaleco/megasys1.cpp
```diff
@@ -398,6 +453,103 @@ void megasys1_state::megasys1C_map(address_map &map)
...
+void megasys1_bc_iosim_state::megasys1C_iosim_map(address_map &map)
+{
+	megasys1C_map(map);
+	map(0x0d8000, 0x0d8001).rw(FUNC(megasys1_bc_iosim_state::ip_select_r), FUNC(megasys1_bc_iosim_state::ip_select_w));
+}
+
+/*
+  Used by IO MCU to read data from different inputs (DSW1, DSW2, P1, P2, COIN) depending on
+  which one was previously selected by the MCU itself using the offset.
+  It also sends the IRQ2 to the main CPU for some case
```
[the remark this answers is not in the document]
mamehaze: some tests were done on the hardware in general, and interrupts occur at fixed screen positions (every 1/4 of the screen) apart from the one from the MCU, which is just comms related.

St Dragon should be a simple raster split based on fixed position IRQ.

src/mame/jaleco/megasys1.cpp
```diff
@@ -365,6 +389,38 @@ void megasys1_state::megasys1B_monkelf_map(address_map &map)
 #define INTERRUPT_NUM_C INTERRUPT_NUM_B
 #define interrupt_C     interrupt_B
 
+void megasys1_state::megasys1c_handle_scanline_irq(int scanline)
```
sergiopolog, line 392: @angelosa SystemC IRQ timings were updated to match what's measured from actual pcb.
Probably St Dragon issue is related to the same thing, but unfortunately I don't have any systemA board to take reading from.

src/devices/cpu/tlcs90/tlcs90.cpp
```diff
@@ -1422,18 +1431,14 @@ void tlcs90_device::execute_set_input(int inputnum, int state)
 
 void tlcs90_device::set_irq_line(int irq, int state)
 {
-	if ( ((m_irq_line_state >> irq)&1) == state ) return;
-
```
cuavas, line 1426: What’s the reason for removing this check?  It looks like it would generate a spurious transition if something calls `set_input_line` with the same values twice in a row.

src/mame/jaleco/megasys1.h
```diff
@@ -14,6 +14,7 @@
 
 #include "machine/gen_latch.h"
 #include "machine/timer.h"
+#include "cpu/tlcs90/tlcs90.h"
 #include "sound/msm5205.h"
 #include "sound/okim6295.h"
 #include "ms1_tmap.h"
```
cuavas, line 20: Please keep headers in alphabetical order for a module, and put local project headers before device headers.

src/mame/jaleco/ms1_tmap.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include "emu.h"
 #include "ms1_tmap.h"
+#include "screen.h"
```
cuavas, line 21: If you’re including `"screen.h"` in the corresponding header, don’t do it here as well.

src/mame/jaleco/megasys1.cpp
```diff
@@ -365,6 +389,38 @@ void megasys1_state::megasys1B_monkelf_map(address_map &map)
...
+{
+	int scanline = param;
+
+	megasys1c_handle_scanline_irq(scanline);
+
+	if(scanline == 0) // end of vblank (rising edge)
+	{
+		LOG("%s: megasys1C_bigstrik_scanline: Send INT1 to MCU: (scanline %03d)\n", machine().describe_context(), scanline);
+		m_iomcu->set_input_line(INPUT_LINE_IRQ1, ASSERT_LINE);
+	}
```
cuavas, line 421: When does this get deasserted?  I can’t see it at a glance.  Is this the reason for the hack in the MCU?

src/mame/jaleco/megasys1.cpp
```diff
@@ -398,6 +453,103 @@ void megasys1_state::megasys1C_map(address_map &map)
...
+/*
+  Write handler triggered by the main CPU at some specific address. Used to write data to port 1 of IO MCU and send irq0
+*/
+void megasys1_bc_iomcu_state::ip_select_bigstrik_w(u16 data) // TO MCU
+{
+	LOG("%s: ip_select_bigstrik_w: Send data to MCU: (data %02x)\n", machine().describe_context(), data);
+
+	m_mcu_input_data = data;
+
+	m_iomcu->set_input_line(INPUT_LINE_IRQ0, ASSERT_LINE);
```
cuavas, line 538: Similarly, where is this being deasserted?

src/mame/jaleco/megasys1.cpp
```diff
@@ -1962,32 +2139,33 @@ void megasys1_state::system_D(machine_config &config)
...
 	m_screen->set_palette(m_palette);
 
 	GFXDECODE(config, m_gfxdecode, m_palette, gfx_z);
 	PALETTE(config, m_palette).set_format(palette_device::RRRRGGGGBBBBRGBx, 0x800/2);
-	MCFG_VIDEO_START_OVERRIDE(megasys1_state,megasys1_z)
 
 	MEGASYS1_TILEMAP(config, m_tmap[0], m_palette, 256*0);
+	m_tmap[0]->set_screen_tag("screen");
 	MEGASYS1_TILEMAP(config, m_tmap[1], m_palette, 256*2);
+	m_tmap[1]->set_screen_tag("screen");
```
cuavas, line 2168: You’ve got a device finder for the screen.  Why aren’t you using it in place of the literal tag here?

src/mame/jaleco/megasys1.cpp
```diff
@@ -1935,10 +2111,11 @@ void megasys1_state::system_D(machine_config &config)
 
 	GFXDECODE(config, m_gfxdecode, m_palette, gfx_abc);
 	PALETTE(config, m_palette, FUNC(megasys1_state::megasys1_palette)).set_format(palette_device::RGBx_555, 0x800/2);
-	MCFG_VIDEO_START_OVERRIDE(megasys1_state,megasys1)
 
 	MEGASYS1_TILEMAP(config, m_tmap[0], m_palette, 256*0);
+	m_tmap[0]->set_screen_tag("screen");
 	MEGASYS1_TILEMAP(config, m_tmap[1], m_palette, 256*1);
+	m_tmap[1]->set_screen_tag("screen");
```
cuavas, line 2118: You’ve got a device finder for the screen. Why aren’t you using it in place of the literal tag here?

src/mame/jaleco/megasys1.cpp
```diff
@@ -1859,11 +2010,13 @@ void megasys1_state::system_Bbl(machine_config &config)
...
 	GFXDECODE(config, m_gfxdecode, m_palette, gfx_abc);
 	PALETTE(config, m_palette, FUNC(megasys1_state::megasys1_palette)).set_format(palette_device::RRRRGGGGBBBBRGBx, 0x800/2);
-	MCFG_VIDEO_START_OVERRIDE(megasys1_state,megasys1)
 
 	MEGASYS1_TILEMAP(config, m_tmap[0], m_palette, 256*0);
+	m_tmap[0]->set_screen_tag("screen");
 	MEGASYS1_TILEMAP(config, m_tmap[1], m_palette, 256*1);
+	m_tmap[1]->set_screen_tag("screen");
 	MEGASYS1_TILEMAP(config, m_tmap[2], m_palette, 256*2);
+	m_tmap[2]->set_screen_tag("screen");
```
cuavas, line 2019: You’ve got a device finder for the screen. Why aren’t you using it in place of the literal tag here?

src/mame/jaleco/megasys1.cpp
```diff
@@ -1723,11 +1873,13 @@ void megasys1_state::system_A(machine_config &config)
...
```
cuavas, line 1882: You’ve got a device finder for the screen. Why aren’t you using it in place of the literal tag here?

cuavas (changes requested):
It looks like this hacks the TLCS90 core to disable edge detection on input lines to avoid needing to deassert interrupt inputs.  That’s going to break other things.  It would be less evil to use `pulse_input_line` or `HOLD_LINE` in the driver.

src/devices/cpu/tlcs90/tlcs90.cpp
sergiopolog: This was done because some of the IRQ of the device are edge-triggered and are automatically cleared as soon as they are acknowledged. 
The code in tlcs90.cpp didn't reset the state of the irq lines (`m_irq_line_state`) when clearing it in that scenario and prevents subsequent (not at the same time) interrupts to be acknowledged, so, I simply removed those checks.
However, on a second look, a better solution would be maintain them and reset `m_irq_line_state` on the `take_interrupt` function, at the same point the irq is cleared.

src/mame/jaleco/megasys1.cpp
sergiopolog: We can change it to `HOLD_LINE`.

src/mame/jaleco/megasys1.cpp
sergiopolog: We can change it to `HOLD_LINE`.

src/devices/cpu/tlcs90/tlcs90.cpp
cuavas: > This was done because some of the IRQ of the device are edge-triggered and are automatically cleared as soon as they are acknowledged.

You aren’t understanding what the code is doing.  `m_irq_line_state` holds the level of the interrupt lines so that it can detect an edge.  If it doesn’t track the current state of the line, it won’t know if it’s seeing an edge or the same value again.

> The code in tlcs90.cpp didn't reset the state of the irq lines (`m_irq_line_state`) when clearing it in that scenario and prevents subsequent (not at the same time) interrupts to be acknowledged, so, I simply removed those checks.

Well yes, the state of the signal on an interrupt input doesn’t magically change when the CPU services the interrupt.  It will only change when the circuit connected to the input changes it.

The interrupt pin must be lowered and raised again before a subsequent edge-triggered interrupt will be detected.  That’s the correct behaviour.  You’ve broken that.

> However, on a second look, a better solution would be maintain them and reset `m_irq_line_state` on the `take_interrupt` function, at the same place the irq is cleared there.

That’s no better.  It’s still magically changing the state of an input that isn’t generated by the CPU.  The code for whatever is attached to the pin is responsible for lowering the line before raising it again to cause another edge-triggered interrupt.

src/mame/jaleco/megasys1.cpp
cuavas: But how does it actually get deasserted in real life on the hardware?  Does it detect an interrupt acknowledge cycle (which is what `HOLD_LINE` is for), or does it generate a fixed length pulse (which can be done `pulse_input_line`), or does something else cause the signal to be deasserted?

sergiopolog: In this case, the IRQ1 signal on the TLCS90 is, in fact, tied to VBlank signal on actual pcb. The interrupt is requested when VBlank ends (scanline == 0) and the line is desasserted when VBlank starts again (scanline == 224).
So, according of what you state above, we may need to simply put a CLEAR_LINE on the timer callback when scanline == 224

cuavas: > So, according of what you state above, we may need to simply put a CLEAR_LINE on the timer callback when scanline == 224

Correct.  That’s the proper way to implement it in this case.

src/mame/jaleco/megasys1.cpp
sergiopolog: In this case, the IRQ0 line is only active during the time spent by the CPU to write the data for the MCU, i.e. only a CPU write cycle.
Then, I don't know how to implement that at code level and where to place the `CLEAR_LINE`.
Does we need to use `pulse_input_line` in this case? How to set the proper lenght of the generated pulse?

cuavas: Yes, you can use `pulse_input_line` in that case.  If it’s for a single bus cycle, you can use `attotime::from_ticks(1, m_maincpu->clock())` or something as the duration.

mamehaze:
all feedback should be addressed

cuavas:
> all feedback should be addressed

Indeed, it appears so.

PR #11643, opened 2023-10-20

A-Noid33:
Some metadata cleanup and consistency checking for Apple II softlists.

Software redumped (apple2_flop_orig.xml)
Wizardry III: Legacy of Llylgamyn (Version 4, 20-Aug-1983 update) [4AM, a2_qkumba]

Software redumped (apple2_flop_clcracked.xml)
Marty's Family Reader (4am crack) [4AM, LoGo]
Marty's Reading Workout (4am crack) [4AM, LoGo]
Greeting Card Maker (4am crack) [4am]
Peanuts Maze Marathon (4am crack) [4am]
The School Tool (4am crack) [4am]

New working software list items (apple2_flop_clcracked.xml) 
A First Program About Insects (4am crack) [4am]
Algebra Word Problems (4am crack) [4am]
Alphabet Zoo (4am crack) [4AM]
Arithmekicks (4am crack) [4am]
Arithmetickle (4am crack) [4am]
Destiny (4am crack) [4am]
Equations I (4am crack) [4am]
Faire le Point-Bac Physique Volume 3 (French) (4am crack) [4am] 
Floppy Teaches: How to Print Letters and Numerals (4am crack) [4am] 
Floppy Teaches: Letter and Numeral Recognition (4am crack) [4am] 
Floppy Teaches: What is Missing? (4am crack) [4am] 
French Classroom Words (4am crack) [4am]
Gobbler (4am and san inc crack) [4AM]
Half Time: Computerized Charting for Marching Bands (version 2.0) (4am crack) [4am] 
Knowledge Master Basics: Earth Science (4am crack) [4am] 
Knowledge Master Basics: Geography (4am crack) [4am] 
Knowledge Master Basics: Language Arts #1 (4am crack) [4am] 
Knowledge Master Basics: Language Arts #2 (4am crack) [4am] 
Knowledge Master Basics: Life Science #1 (4am crack) [4am] 
Knowledge Master Basics: Life Science #2 (4am crack) [4am] 
Knowledge Master Basics: Literature (4am crack) [4am] 
Knowledge Master Basics: Math #1 (4am crack) [4am] 
Knowledge Master Basics: Math #2 (4am crack) [4am] 
Knowledge Master Basics: Physical Science (4am crack) [4am] 
Knowledge Master Basics: Social Studies #2 (4am crack) [4am] 
Map Reading (4am crack) [4am]
Math Football: Decimals (version 3.1.6) (4am crack) [4am] 
Mathematics Word Problems - Grade 7 (4am crack) [4am] 
Mathematics Word Problems - Grade 8 (4am crack) [4am] 
Money (version 4.1.7) (4am crack) [4am]
MultiScribe (version 3.01c) (4am crack) [4AM]
Parachute Parts of Speech (4am crack) [4am]
Percent (version 3.1.6) (4am crack) [4am]
Percent Word Problems (4am crack) [4am]
Perplexing Puzzles (version 1985-07-09) (4am crack) [4am] 
pfs: report (version 1984-03-20) (4am crack) [4AM] 
pfs: report (version 1985-09-10) (4am crack) [4AM] 
Punctuation Baseball Level A (version 4.1.5) (4am crack) [4am] 
Real Number System (4am crack) [4am]
Rendezvous with Rama (4am crack) (Telarium Re-release) [4am] 
Skill Builder: Social Studies - Grade 6 (4am crack) [4am] 
Spider Hunt Spelling (version 4.1.6) (4am crack) [4am] 
Story Writing (4am crack) [4am]
Take 1 (version 06431) (4am crack) [4AM]
The Great Knowledge Race (4am crack) [4am]
The Magic Cash Register (4am crack) [4am]
The Time Tunnel: Sports Edition (4am crack) [4am]
Time Capsule (4am crack) [4am]
Travels with Za-Zoom: The World (revision 2) (4am crack) [4am] 
Word Problems Level C (version 3.1.6) (4am crack) [4am]

### DIFF
--- a/hash/apple2_flop_clcracked.xml
+++ b/hash/apple2_flop_clcracked.xml
@@ -50348,7 +50348,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfse841126" cloneof="pfsreport">
+	<software name="pfsreportf841126" cloneof="pfsreport">
 		<description>pfs: etat (French, version 1984-11-26) (4am crack)</description>
 		<!--French version of pfs: report-->
 		<year>1984</year>
@@ -50366,7 +50366,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsfich841126" cloneof="pfsfile">
+	<software name="pfsfilef841126" cloneof="pfsfile">
 		<description>pfs: fichier (French, version 1984-11-26) (4am crack)</description>
 		<!--French version of pfs: file-->
 		<year>1984</year>
@@ -50384,7 +50384,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsfa00" cloneof="pfsfile">
+	<software name="pfsfilea00" cloneof="pfsfile">
 		<description>pfs: file (version A.00) (4am crack)</description>
 		<year>1983</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50401,7 +50401,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsfb01" cloneof="pfsfile">
+	<software name="pfsfileb01" cloneof="pfsfile">
 		<description>pfs: file (version B.01) (4am crack)</description>
 		<year>1982</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50418,7 +50418,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsfc00" cloneof="pfsfile">
+	<software name="pfsfilec00" cloneof="pfsfile">
 		<description>pfs: file (version C.00) (4am crack)</description>
 		<year>1984</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50452,7 +50452,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsg830112" cloneof="pfsgraph">
+	<software name="pfsgraph830112" cloneof="pfsgraph">
 		<description>pfs: graph (version 1983-01-12) (4am crack)</description>
 		<year>1983</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50469,7 +50469,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsg840131" cloneof="pfsgraph">
+	<software name="pfsgraph840131" cloneof="pfsgraph">
 		<description>pfs: graph (version 1984-01-31) (4am crack)</description>
 		<year>1984</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50503,7 +50503,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsrb00" cloneof="pfsreport">
+	<software name="pfsreportb00" cloneof="pfsreport">
 		<description>pfs: report (version B.00, 1981-01-03) (4am crack)</description>
 		<year>1981</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50520,7 +50520,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsrb02" cloneof="pfsreport">
+	<software name="pfsreportb02" cloneof="pfsreport">
 		<description>pfs: report (version B.02) (4am crack)</description>
 		<year>1982</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50537,7 +50537,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsr820114" cloneof="pfsreport">
+	<software name="pfsreport820114" cloneof="pfsreport">
 		<description>pfs: report (version 1982-01-14) (4am crack)</description>
 		<year>1982</year>
 		<publisher>Software Publishing Corporation</publisher>
@@ -50554,7 +50554,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsr840320" cloneof="pfsreport">
+	<software name="pfsreport840320" cloneof="pfsreport">
 		<description>pfs: report (version 1984-03-20) (4am crack)</description>
 		<year>1984</year>
 		<publisher>Software Publishing Corporation</publisher>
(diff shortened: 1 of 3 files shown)


hash/apple2_flop_clcracked.xml
```diff
@@ -49584,4 +48700,1980 @@ license:CC0-1.0
...
+		<!--productivity program-->
+		<part name="flop1" interface="floppy_5_25">
+			<dataarea name="flop" size="143360">
+				<rom name="pfs file va.00 (4am crack).dsk" size="143360" crc="49980399" sha1="14451182f9d319d2c08a7f7fd68aec10654299b1"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="pfsfb01">
+		<description>pfs: file (version B.01) (4am crack)</description>
```
cuavas, line 50405: Similarly, are the pfs: file versions different revisions of the same software?  Is pfs: fichier the French version of pfs: file?

hash/apple2_flop_clcracked.xml
```diff
@@ -49584,4 +48700,1980 @@ license:CC0-1.0
...
+		<!--productivity program-->
+		<part name="flop1" interface="floppy_5_25">
+			<dataarea name="flop" size="143360">
+				<rom name="pfs report 1981-01-03 (vb.00) (4am crack).dsk" size="143360" crc="47f55268" sha1="a67c4ef6e2591bc0600b565c88f00552eecb5a51"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="pfsrb02">
+		<description>pfs: report (version B.02) (4am crack)</description>
```
cuavas, line 50524: Should alternate versions of pfs: report be clones?

cuavas (changes requested):
* To me, it appears that several of the pfs software packages are missing parent/clone relationships for different versions of the same software.  Can you please check that the different versions of pfs: write, pfs: report, etc. are just different versions of the same software and add parent/clone relationships as appropriate?
* The latest version of a software package should be the parent in most situations.  The parent should have the base `name` and suffixes should be appended to the `name` for clones.  I found at least one case where a later revision was marked as a clone.  Can you please check these?
* You’ve changed the short names of several items in apple2_flop_clcracked.xml.  Can you please ensure the short names match the equivalent uncracked software in apple2_flop_orig.xml?  It makes it easier to do comparisons, etc.

hash/apple2_flop_clcracked.xml
A-Noid33: @cuavas Completed updated the parent-clone relationships.

cuavas:
> I assume that Parent-Clone relationships will be per-software lists, so there will be a parent in cleanly cracked and a separate parent in original.

Correct – there’s no way to specify relationships between software in different software lists at this time.

> What should I do with 800K 3.5" floppies vs 5.25" floppies?  Should I make the 800K floppy the parent, when we have both 3.5 and 5.25 of the same base version?

Interesting question.  My gut instinct says to make the 5.25" version the parent if the software is compatible with 8-bit Apple II machines as 5.25" drives were standard (ignoring the misguided IIc+).  For 16-bit IIgs software, maybe the 3.5" version could be made the parent?  I’m not sure.

Maybe @rb6502 has opinions on this.

hash/apple2_flop_clcracked.xml
```diff
@@ -50384,7 +50384,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsfa00">
+	<software name="pfsfa00" cloneof="pfsfile">
```
cuavas, line 50387: For cases like this, can you please just make the `name` of the clone the name of the parent with a suffix appended?  It makes it easier to see the groups of clones in a sorted listing.  In this case, you can make the clone’s `name` `pfsfilea00`.  There’s no need to force the `name` for clones into eight characters (up to sixteen characters are allowed by validation).

hash/apple2_flop_clcracked.xml
```diff
@@ -50469,7 +50469,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="pfsg840131">
+	<software name="pfsg840131" cloneof="pfsgraph">
```
cuavas, line 50472: Same here – `pfsgraph840131` fits in sixteen characters fine.

cuavas (changes requested):
For the `name` of clones, please append a suffix to the parent `name` where it makes sense.  It’s a lot easier to see the groups of clones in sorted listings that way, and it makes things easier to read.  Up to sixteen characters are allowed.

rb6502:
> Maybe @rb6502 has opinions on this.

My opinion is simple: separate 5.25 and 3.5 software lists.  IIgs and IIc+ (and the Liron card whenever the IWM is restructured to have correct controller/drive boundaries) prioritize the 3.5.  Then you aren't polluting the software list namespace with "aplwks40_35" and you aren't violating the rule of least surprise.

cuavas:
I’m still uncomfortable with artificial splits like that.  There’s already the interface used to match media to a suitable device, it falls over for software that included multiple media types, and it makes it much harder to group releases of the same software on different media.

rb6502:
That split has the advantage of not requiring additional work on the software loading.

To work without the split, multiple media types in the same software needs to work, there needs to be some way to specify which one to try first (so IIgs and IIc+ can get 3.5), and there needs to be a way to specify sets of media types so that a dual-media AppleWorks set wouldn't load both 5.25 and 3.5, but a Windows 95 boot floppy + CD-ROM would.

A-Noid33:
We also have issues with different compatibility between Parents and
Clones.  There are lots of Apple II parents that will target different
ranges of apple II computers than the earlier clones, for example dropping
support for apple II plus and adding apple II e/gs support.

There are also things like 13-sector drives.  We probably need more choices
for the <part/>  than just floppy_3_5 and Floppy_5_25.

Also, there are memory requirements... 48K, 64K, 128K.

rb6502:
Right.  We'd need separate compatibility tags per media set.  I'm less concerned about memory requirements because I think you should be able to -ramsize 16K and have everything get angry at you just like on hardware.

The upshot is that all of this potentially adds a lot of complexity to the software handling, and you can side-step a lot of it with separate lists.

A-Noid33:
Here's a question about parent/clone relationships.  With Crosscountry
California, the version for woz-a-day is revision 2 with a copyright of
1991.  The cracked version is revision 1 with a copyright of 1987.   The
1991 version has different opening graphics and copyright.   SInce they are
in two different software lists and can't have parent / clone
relationships, I think I should still maintain the base name...   so should
I name the latest one, crscali in the apple2_flop_orig.xml and crscalir2 in
the apple2_flop_clcracked list?

This is a much larger issue for MECC, where 4AM has released many different
versions in each of the repositories.

cuavas:
Is there no woz-a-day of revision 1?  At any rate, I agree with you on trying to maintain the same names across lists, although I might call the rev 1 crack `crscalir1` since it’s rev 1 while the `crscali` in the other list is rev 2.

PR #11650, opened 2023-10-22
megasys1.cpp - various cleanups based on new information etc.

mamehaze:


src/mame/jaleco/megasys1.cpp
```diff
@@ -5139,33 +5113,91 @@ void megasys1_state::megasys1A_mcu_hs_w(offs_t offset, u16 data, u16 mem_mask)
...
-
+	m_maincpu->space(AS_PROGRAM).install_read_handler(0x00000, 0x3ffff, read16s_delegate(*this, FUNC(megasys1_state::gatearray_r)));
+	m_maincpu->space(AS_PROGRAM).install_write_handler(base_write, base_write+0x9, write16s_delegate(*this, FUNC(megasys1_state::gatearray_w)));
+	m_gatearray_seq = sequence;
 	m_mcu_hs = 0;
 	memset(m_mcu_hs_ram, 0, sizeof(m_mcu_hs_ram));
-
 	save_item(NAME(m_mcu_hs));
 	save_item(NAME(m_mcu_hs_ram));
 }
```
cuavas, line 5125: Please put the right-associative unary `*` qualifier to the right of the space, and restore the blank lines that separated installing memory handlers, intialising members, and registering items for save states.

src/mame/jaleco/megasys1.cpp
```diff
@@ -5238,105 +5274,44 @@ void megasys1_bc_iosim_state::init_cybattlr()
...
-
-	m_mcu_hs = 0;
-	memset(m_mcu_hs_ram, 0, sizeof(m_mcu_hs_ram));
-
-	save_item(NAME(m_mcu_hs));
-	save_item(NAME(m_mcu_hs_ram));
+	save_item(NAME(m_sprite_bank));
+	save_item(NAME(m_mcu_input_data));
+	save_item(NAME(m_mcu_io_data));
 }
```
cuavas, line 5312: Does this really need an init function?  Can’t the class with the emulated microcontroller just override `machine_start` to register things for save states?

src/mame/jaleco/megasys1.cpp
```diff
@@ -5238,105 +5274,44 @@ void megasys1_bc_iosim_state::init_cybattlr()
...
 	m_ip_select_values[2] = 0x53;
 	m_ip_select_values[3] = 0x54;
 	m_ip_select_values[4] = 0x55;
 
-	m_ip_select_values[5] = 0xfc;
+	m_ip_select_values[5] = 0xf2;
 	m_ip_select_values[6] = 0x06;
 
 	save_item(NAME(m_ip_latched));
+	save_item(NAME(m_sprite_bank));
```
cuavas, line 5304: Can these be saved in `machine_start` rather than repeating it in all the init functions for the class that simulates the microcontroller?

src/mame/jaleco/megasys1.h
```diff
@@ -157,20 +154,19 @@ class megasys1_state : public driver_device
 	int m_mcu_hs = 0;
 	u16 m_mcu_hs_ram[0x10]{};
 
+	const uint16_t* m_gatearray_seq = nullptr;
```
cuavas, line 157: Please put the right-associative unary `*` qualifier to the right of the space.

src/mame/jaleco/megasys1.h
```diff
@@ -191,6 +187,7 @@ class megasys1_state : public driver_device
 	TIMER_DEVICE_CALLBACK_MEMBER(megasys1A_iganinju_scanline);
 	TIMER_DEVICE_CALLBACK_MEMBER(megasys1C_scanline);
 
+	void install_gatearray_overlay(uint32_t base_write, const uint16_t* sequence);
```
cuavas, line 190: Please put the right-associative unary `*` qualifier to the right of the space.

src/mame/jaleco/megasys1.h
mamehaze: I keep forgetting this because it's at odds with standards everywhere else I do work, and at odds with visual studio's default auto reformatting, but fixed now.

src/mame/jaleco/megasys1.cpp
mamehaze: the ip_latched thing can be, for sprite_bank seems to be a 1c specific feature that might get tidied up later

src/mame/jaleco/megasys1.cpp
mamehaze: it now only has the sprite_bank bit as it's type c, the rest is in a machine_start

mamehaze:
fwiw this also fixes the random crash with tshingen https://mametesters.org/view.php?id=8541
rather than doing it on startup, they do it when certain conditions are met.

tshingena has the same code but I'm not sure it's called

even plusalph has the checks in it, but I'm again not sure it's called

src/mame/jaleco/ms1_gatearray.cpp
```diff
@@ -0,0 +1,249 @@
...
+#undef  BITSWAP_2
+#undef  BITSWAP_3
+
+		RAM[i] = y;
+	}
+}
+
+
+/*
+    MCU handshake sequence:
```
angelosa, line 149: If it's not a MCU then this note is outdated.

src/mame/jaleco/ms1_gatearray.cpp
```diff
@@ -0,0 +1,249 @@
...
+		m_mcu_hs_ram[4/2] == _3_ && \
+		m_mcu_hs_ram[6/2] == _4_)
+
+u16 megasys1_gatearray_device::gatearray_r(offs_t offset, u16 mem_mask)
+{
+	uint16_t *m_rom_maincpu = (uint16_t *)m_rom;
+	if(m_mcu_hs && ((m_mcu_hs_ram[8/2] << 6) & 0x3ffc0) == ((offset*2) & 0x3ffc0))
+	{
+		if(MCU_HS_LOG && !machine().side_effects_disabled())
+			printf("MCU HS R (%04x) <- [%02x]\n",mem_mask,offset*2);
```
angelosa, line 170: `printf` can be changed to `logmacro.h` `LOG`

src/mame/jaleco/megasys1.h
```diff
@@ -262,12 +281,20 @@ class megasys1_bc_iosim_state : public megasys1_state
 	void system_C_iosim(machine_config &config);
 
 protected:
+	virtual void machine_start() override;
 	virtual void machine_reset() override;
 
-	u16 m_ip_select_values[7]{}; // System B and C
+	const u8* m_ip_select_values = nullptr; // System B and C
```
cuavas, line 287: Please put the right-associative unary `*` qualifier to the right of the space.

src/mame/jaleco/ms1_gatearray.cpp
```diff
@@ -0,0 +1,245 @@
...
+	{
+		u16 y;
+		u16 x = RAM[i];
+
+// [0] def0 189a bc56 7234
+// [1] fdb9 7531 eca8 6420
+// [2] 0123 4567 ba98 fedc
+#define BITSWAP_0   bitswap<16>(x,0xd,0xe,0xf,0x0,0x1,0x8,0x9,0xa,0xb,0xc,0x5,0x6,0x7,0x2,0x3,0x4)
+#define BITSWAP_1   bitswap<16>(x,0xf,0xd,0xb,0x9,0x7,0x5,0x3,0x1,0xe,0xc,0xa,0x8,0x6,0x4,0x2,0x0)
+#define BITSWAP_2   bitswap<16>(x,0x0,0x1,0x2,0x3,0x4,0x5,0x6,0x7,0xb,0xa,0x9,0x8,0xf,0xe,0xd,0xc)
```
cuavas, line 64: Please use lambdas for these, something along the lines of:
```c++
	auto const BITSWAP_0 = [x] () { return bitswap<16>(x,0xd,0xe,0xf,0x0,0x1,0x8,0x9,0xa,0xb,0xc,0x5,0x6,0x7,0x2,0x3,0x4); };
```
etc. to avoid the need for preprocessor expansion.

src/mame/jaleco/ms1_gatearray.cpp
```diff
@@ -0,0 +1,245 @@
...
+		u16 y;
+		u16 x = RAM[i];
+
+// [0] def0 a981 65cb 7234
+// [1] fdb9 7531 8ace 0246
+// [2] 4567 0123 ba98 fedc
+
+#define BITSWAP_0   bitswap<16>(x,0xd,0xe,0xf,0x0,0xa,0x9,0x8,0x1,0x6,0x5,0xc,0xb,0x7,0x2,0x3,0x4)
+#define BITSWAP_1   bitswap<16>(x,0xf,0xd,0xb,0x9,0x7,0x5,0x3,0x1,0x8,0xa,0xc,0xe,0x0,0x2,0x4,0x6)
+#define BITSWAP_2   bitswap<16>(x,0x4,0x5,0x6,0x7,0x0,0x1,0x2,0x3,0xb,0xa,0x9,0x8,0xf,0xe,0xd,0xc)
```
cuavas, line 98: Same here, please use lambdas to avoid the need for the preprocessor.

src/mame/jaleco/ms1_gatearray.cpp
```diff
@@ -0,0 +1,245 @@
...
+	{
+		u16 y;
+		u16 x = RAM[i];
+
+// [0] d0a9 6ebf 5c72 3814  [1] 4567 0123 ba98 fedc
+// [2] fdb9 ce07 5318 a246  [3] 4512 ed3b a967 08fc
+#define BITSWAP_0   bitswap<16>(x,0xd,0x0,0xa,0x9,0x6,0xe,0xb,0xf,0x5,0xc,0x7,0x2,0x3,0x8,0x1,0x4);
+#define BITSWAP_1   bitswap<16>(x,0x4,0x5,0x6,0x7,0x0,0x1,0x2,0x3,0xb,0xa,0x9,0x8,0xf,0xe,0xd,0xc);
+#define BITSWAP_2   bitswap<16>(x,0xf,0xd,0xb,0x9,0xc,0xe,0x0,0x7,0x5,0x3,0x1,0x8,0xa,0x2,0x4,0x6);
+#define BITSWAP_3   bitswap<16>(x,0x4,0x5,0x1,0x2,0xe,0xd,0x3,0xb,0xa,0x9,0x6,0x7,0x0,0x8,0xf,0xc);
```
cuavas, line 130: As above, lambdas can be used here.

src/mame/jaleco/ms1_gatearray.h
```diff
@@ -0,0 +1,97 @@
...
+	void gatearray_w(offs_t offset, u16 data, u16 mem_mask = ~0);
+
+	int m_gatearray_hs = 0;
+	u16 m_gatearray_hs_ram[0x8]{};
+
+	const u16* m_gatearray_seq = nullptr;
+
+	required_device<cpu_device> m_cpu;
+	u8* m_rom;
+	u32 m_romsize;
```
cuavas, line 37: Please use a `required_memory_region` for this, so it can be configured at machine config time, similarly to the CPU.  Then you can do the decoding on device start instead of having to do it on reset and keep track of whether you’ve done it.

cuavas (changes requested):

PR #11661, opened 2023-10-26
igs/goldstar.cpp: Dumped and added an "El Dorado" clone named "Suns Pecker"

clawgrip:
New NOT_WORKING clone
Suns Pecker (V1.0) [Taksangs]


cuavas:
Can you change the short name to “sunspckr”?  It fits in eight characters and reads a bit better than “pecke”.

mamehaze:
are the PROMs on these going to be dumped eventually?  It's starting to become concerning how many sets are being added without them IMHO.

clawgrip:
> are the PROMs on these going to be dumped eventually? It's starting to become concerning how many sets are being added without them IMHO.

Taksangs is trying to rely on a repair shop to deal with them. We'll see the results this weekend.

clawgrip:
> Can you change the short name to “sunspckr”? It fits in eight characters and reads a bit better than “pecke”.

Done, thanks!!

PR #11664, opened 2023-10-27
nuon.xml: Added 1 item (non working), updated metadata, dumps to DVD …

FoxhackDN:
…CHD format

Jjangguneun Monmallyeo 3: Doraon Jjanggu (Korea) [emceebois]

### DIFF
--- a/hash/nuon.xml
+++ b/hash/nuon.xml
@@ -41,12 +41,12 @@ license:CC0-1.0
 
 	<software name="jm3" supported="no">
 		<description>Jjangguneun Monmallyeo 3: Doraon Jjanggu (Korea)</description>
-		<info name="alt_title" value="짱구는 못말려 3: 돌아온 짱구" />
 		<year>2001</year>
 		<publisher>VM Labs, Inc.</publisher>
 		<notes><![CDATA[
 Region locked, will only work in Korea region players.
 ]]></notes>
+		<info name="alt_title" value="짱구는 못말려 3: 돌아온 짱구" />
 		<part interface="dvdrom" name="dvdrom">
 			<diskarea name="dvdrom">
 				<disk name="jjangguneun monmallyeo 3 - doraon jjanggu (korea)" sha1="ff2bfb2108643610969980112c5938491f3df3bb
--- a/hash/nuon.xml
+++ b/hash/nuon.xml
@@ -12,7 +12,7 @@ license:CC0-1.0
 		<publisher>VM Labs, Inc.</publisher>
 		<part interface="dvdrom" name="dvdrom">
 			<diskarea name="dvdrom">
-				<disk name="ballstic (usa, europe)" sha1="5371a3abfb92bc1841d3416938bc07b2619024f0" />
+				<disk name="ballistic (usa, europe)" sha1="5371a3abfb92bc1841d3416938bc07b2619024f0" />
 			</diskarea>
 		</part>
 	</software>
--- a/hash/nuon.xml
+++ b/hash/nuon.xml
@@ -6,7 +6,7 @@ license:CC0-1.0
 
 <softwarelist name="nuon" description="NUON DVD-ROMs">
 
-	<software name="ballstic">
+	<software name="ballstic" supported="no">
 		<description>Ballistic (USA, Europe)</description>
 		<year>2000</year>
 		<publisher>VM Labs, Inc.</publisher>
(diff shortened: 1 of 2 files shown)


FoxhackDN:
Software list metadata (region and names) has been updated based on Redump's entries and my own research. All existing CHDs were updated to the DVD CHD format by extracting them with `chdman extractcd`, then recompressing them with `chdman createdvd`; no data was lost, the old extracted dumps matched the verified Redump dumps. So, old dumps can be updated with those two commands.

I also went ahead and renamed the files to something that was more readable, but kept the old folder / set names intact.

hash/nuon.xml
```diff
@@ -7,78 +7,96 @@ license:CC0-1.0
...
+		<year>2000</year>
+		<publisher>Toshiba</publisher>
+		<notes><![CDATA[
+Bundled with Toshiba players. Disc also contains DVD Video content that will work on any standard DVD player.
+]]></notes>
 		<part interface="dvdrom" name="dvdrom">
 			<diskarea name="dvdrom">
-				<disk name="toshsamp" sha1="e312c9cec7f11a4594c1a3340fbcc9f66a65e9b1" />
+				<disk name="the next tetris - toshiba sampler (11-1-2000) (usa)" sha1="27051216969f17c6f4e832338c860c58fe8d127f" />
 			</diskarea>
```
cuavas, line 100: Does this use CSS for the DVD-Video content?  If it does, it needs to be marked bad because CHD doesn’t even pretend to support the disc features used for that.

hash/nuon.xml
```diff
@@ -7,78 +7,96 @@ license:CC0-1.0
...
+Region locked, will only work in Korea region players.
+]]></notes>
 		<part interface="dvdrom" name="dvdrom">
 			<diskarea name="dvdrom">
-				<disk name="is3" sha1="47dec1bbb819342cf8bc2fee98fc7b18d1302503" />
+				<disk name="jjangguneun monmallyeo 3 - doraon jjanggu (korea)" sha1="ff2bfb2108643610969980112c5938491f3df3bb
+" />
 			</diskarea>
 		</part>
 	</software>
```
cuavas, line 55: Please include the Korean title in Hangeul (짱구는 못말려 3: 돌아온 짱구) as an `alt_title`.

hash/nuon.xml
```diff
@@ -7,78 +7,96 @@ license:CC0-1.0
...
+		<year>2001</year>
+		<publisher>VM Labs, Inc.</publisher>
+		<part interface="dvdrom" name="dvdrom">
+			<diskarea name="dvdrom">
+				<disk name="iron soldier 3 (usa)" sha1="5e8951552c80fb0148de33fd2c77c29a9a769603" />
+			</diskarea>
+		</part>
+	</software>
+
+	<software name="jm3">
```
angelosa, line 42: If the SW item is "not working" then it should have `supported="no"` flag (this applies to everything in this file really).

angelosa (changes requested):

hash/nuon.xml
FoxhackDN: Gotcha. I'll add the tag to all the items later today.

cuavas: A lot of the software lists for skeleton systems, or systems marked not working where it’s implied nothing works, don’t bother setting `supported` attributes.

hash/nuon.xml
FoxhackDN: I'm not sure... I'll look at it later today and add a note if that's the case. I know it's got Regions 1 to 8, but I don't know if it has proper CSS on it.

FoxhackDN: According to DVD Decrypter, the disc is all region (1 to 8), has no CSS encryption, nor RCE protection. You can just drag and drop the VOB files into normal movie playback software and they'll work without issues, so there's no need to mark this as a bad dump.

hash/nuon.xml
FoxhackDN: Done, all other requested changes have been made too.

hash/nuon.xml
angelosa: In my crusade against `supported=<undefined>` (sometimes with rather [embarrassing later demotions](https://github.com/mamedev/mame/commit/a98cf345cd7d6d0370f69b25c5bcb85de80258ea)) hearing that a 6th gen DVD player with likely performance and structural pitfalls don't need `supported=no` isn't an option, it will be done either in this PR or later by me. It just won't happen that anybody will emulate this in one big commit, and it's baby regex that requires 30 seconds.

PR #11665, opened 2023-10-27
ins8250.cpp: Implement Break functionality

mgarlanger:
Compared behavior between a real H19 and the emulated H19, while holding down the break key. Both behaved identically, other keys pressed while the break key was pressed queued up and processed after the break key was released.


cuavas (changes requested):
This won’t work properly:
* Setting the break bit in the LCR should immediately pull down TxD, whether data transmission is in progress or not.
* Clearing the break bit in the LCR should immediately set TxD back to whatever was last shifted out of the transmit shift register.

You need to:
* Leave `m_txd` containing whatever was shifted out of the transmit shift register.
* Call `m_out_tx_cb` (or `device_serial_interface::rx_w` for loopback cases) immediately when the break bit changes on a register write.
* Check the break bit before calls to `m_out_tx_cb` (and `device_serial_interface::rx_w` for loopback cases) so you don’t change the output when it’s being overridden by the break bit.

PR #11669, opened 2023-10-29
ins8250.cpp: Remove scratchpad register access to 8250 class chips

mgarlanger:
The original 8250 class chips did not have a scratchpad register at offset 7. This was added in the 16450. 
8250 datasheets specify this, and can be found in some old manuals like the Heath H89 or H19 manual. 
Here is the chart in an H89 manual.

Screenshot 2023-10-29 at 1 32 27 AM

It is also documented on this wiki, that the lack of the scratchpad was a way to determine it was an 8250 chip: https://en.wikibooks.org/wiki/Serial_Programming/8250_UART_Programming

### DIFF
--- a/src/devices/machine/ins8250.cpp
+++ b/src/devices/machine/ins8250.cpp
@@ -469,7 +469,10 @@ u8 ins8250_uart_device::ins8250_r(offs_t offset)
 			}
 			break;
 		case 7:
-			data = m_regs.scr;
+			if (m_device_type >= dev_type::NS16450)
+			{
+				data = m_regs.scr;
+			}
 			break;
 	}
 	return data;
--- a/src/devices/machine/ins8250.cpp
+++ b/src/devices/machine/ins8250.cpp
@@ -390,13 +390,16 @@ void ins8250_uart_device::ins8250_w(offs_t offset, u8 data)
 			 */
 			m_regs.msr = (m_regs.msr & 0xf0) | (data & 0x0f);
 
-			if ( m_regs.msr & 0x0f )
+			if (m_regs.msr & 0x0f)
 				trigger_int(COM_INT_PENDING_MODEM_STATUS_REGISTER);
 			else
 				clear_int(COM_INT_PENDING_MODEM_STATUS_REGISTER);
 			break;
 		case 7:
-			m_regs.scr = data;
+			if (m_device_type >= dev_type::INS8250A)
+			{
+				m_regs.scr = data;
+			}
 			break;
 	}
 }
@@ -469,7 +472,7 @@ u8 ins8250_uart_device::ins8250_r(offs_t offset)
 			}
 			break;
 		case 7:
-			if (m_device_type >= dev_type::NS16450)
+			if (m_device_type >= dev_type::INS8250A)
 			{
 				data = m_regs.scr;
 			}


cuavas:
The National Semiconductor PC8250A/INS8250A/INS82C50A *does* have the scratch register – it’s functionally identical to the NS16450 but with a slower CPU bus interface.

Do only the INS8250 and INS8250-B lack the scratch register?  They supposedly only differ in CPU bus timings (the INS8250-B is slower).  They both momentarily drop the interrupt output when a new interrupt condition arrives and they output handshake signals in loopback mode, which the INS8250A/NS16450/etc. don’t do.

Oh, one other thing – did you actually test this with an INS8250 or INS8250-B?  I realise the documentation *says* there’s no register there.  However that could just mean that writing to that location has no effect on the chip.  The scratch register may have already existed as an undocumented register that had no effect, leading to people who discovered it to use it as a scratch register.

The reason I’m wondering is that I have the National Semiconductor document that compares the members of the family (note 493, April 1989), and it makes no mention of the INS8250/INS8250-B lacking the scratch register.  It *does* note the differences in handling multiple pending interrupts and handshake signals in loopback mode.  Lack of the scratch register would be a functional difference, so it would be surprising if they chose not to mention it in the comparison.

mgarlanger:
Back in the late '80s, I used a similar check as on the wikibook link I posted above, to have my modem program properly determine the type of UART, and I know it was able to detect the 8250s I tested. 

I'll setup one of my H89s and test it. Have you tested the National Semiconductor PC8250A/INS8250A/INS82C50A?

cuavas:
> Back in the late '80s, I used a similar check as on the wikibook link I posted above, to have my modem program properly determine the type of UART, and I know it was able to detect the 8250s I tested.
> 
> I'll setup one of my H89s and test it. Have you tested the National Semiconductor PC8250A/INS8250A/INS82C50A?

I’ve tested the 82C50A – the scratch register is present.  I don’t have the XMOS versions to test (I also don’t have examples of the INS8250/INS8250-B to test).

mgarlanger:
My H89 is having some hardware issues, I'll try to work on later today, but I also found that linux uses the scratchpad check to identify the serial port: https://github.com/torvalds/linux/blob/master/drivers/tty/serial/8250/8250_port.c#L945 . And this WD datasheet(http://bitsavers.org/components/westernDigital/_dataSheets/WD8250_82C50_16C450.pdf) shows this difference between their 8250 and 16450 chips: 
Screenshot 2023-10-29 at 12 13 18 PM

cuavas:
OK, that means at least the 8250 clones lack a scratch register.

cuavas:
I updated this to only suppress the scratch register for the 8250, since at least the CMOS 82C50A has the register present.

We still aren’t emulating the differences in loopback behaviour or dealing with multiple pending interrupts.  I guess no-one’s run into a use case that requires it.  (At least some systems with edge-triggered interrupts supposedly don’t work with the 8250A and later chips.)

mgarlanger:
> The reason I’m wondering is that I have the National Semiconductor document that compares the members of the family (note 493, April 1989), and it makes no mention of the INS8250/INS8250-B lacking the scratch register. It _does_ note the differences in handling multiple pending interrupts and handshake signals in loopback mode. Lack of the scratch register would be a functional difference, so it would be surprising if they chose not to mention it in the comparison.

In that document, I did find mention that the scratch register is not present in 8250 and -B. It is in section 3.0, item 2.
Screenshot 2023-10-31 at 10 59 19 PM

PR #11670, opened 2023-10-29
Add Altos 586 emulation

lkundrak:
Hi. This adds incomplete emulation of Altos 586 -- the main board and a hard drive board. It is good enough to boot and pass tests on the diagnostic floppy, but not much else (no XENIX, CCP/M, OASIS). More details in the commit messages of the individual commits and code comments.

I eventually intend to extend the emulation to boot all the supported software, but as I'm pretty new to MAME, I thought it's a good idea to post this for review early, before I go too far doing something stupid. 

This is my first time programming C++ so please go easy on me :)

### DIFF
--- a/src/mame/altos/altos586.cpp
+++ b/src/mame/altos/altos586.cpp
@@ -1,0 +1,777 @@
+// license:BSD-2-Clause
+// copyright-holders:Lubomir Rintel
+
+/***************************************************************************
+
+    Altos 586 computer emulation
+
+    Work in progress. The current goal is to iron our the overall flaws
+    caused by my inexperience with C++ and MAME. The ultimate goal is to
+    make this complete enough for XENIX to run. I guess at that point
+    it would imply CP/M and OASIS working too.
+
+    At this point I've not tried to boot anything but the diags floppy,
+    and it's unlikely to work.
+
+    The tests on the diags floppy work well, including serial, clock,
+    memory management, hard disk & floppy diags and maintenance.
+    What the diags disk doesn't exercise are interrupts, access control of
+    bus access from IOP/HDC and system call machinery.
+
+    Literature:
+
+    [1] 586T/986T System Reference Manual, P/N 690-15813-002, April 1985
+        This describes a slightly different system.
+        690-15813-002_Altos_586T_986T_System_Reference_Apr85.pdf
+
+    [2] Notes on the Altos 586 Computer & Firmware disassembly
+        https://github.com/lkundrak/altos586/
+
+***************************************************************************/
+
+#include "emu.h"
+
+#include "altos586_hdc.h"
+
+#include "bus/rs232/rs232.h"
+
+#include "cpu/i86/i86.h"
+#include "cpu/z80/z80.h"
+
+#include "imagedev/floppy.h"
+
+#include "machine/clock.h"
+#include "machine/mm58167.h"
+#include "machine/pic8259.h"
+#include "machine/pit8253.h"
+#include "machine/ram.h"
+#include "machine/wd_fdc.h"
+#include "machine/z80dma.h"
+#include "machine/z80pio.h"
+#include "machine/z80sio.h"
+
+// TODO: Should this be a separate device? It is on a same board.
+// I've split this so that I've got two device_memory_interface-s --
+// the board object provides the maps and unmanaged memory and I/O
+// spaces, while the MMU provides the managed ones for the bus
+// peripherals (IOP and HDC) via its AS_PROGRAM/AS_IO spaces and
+// cpu_{mem,io}_{r,w} routines for the access from the main CPU.
+// Could there be a better way to structure this?
+class altos586_mmu_device : public device_t, public device_memory_interface
+{
+public:
+	typedef device_delegate<void ()> violation_delegate;
+
+	altos586_mmu_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+	template <typename T>
+	altos586_mmu_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock, T &&board_tag)
+		: altos586_mmu_device(mconfig, tag, owner, clock)
+	{
+		m_board.set_tag(std::forward<T>(board_tag));
+	}
+
+	template <typename... T> void set_violation_callback(T &&... args) { m_violation_callback.set(std::forward<T>(args)...); }
+	auto syscall_handler() { return m_syscall_handler.bind(); }
+
+	// Managed access from the main board CPU
+	u16 cpu_io_r(offs_t offset, u16 mem_mask);
+	void cpu_io_w(offs_t offset, u16 data, u16 mem_mask);
+	u16 cpu_mem_r(offs_t offset, u16 mem_mask);
+	void cpu_mem_w(offs_t offset, u16 data, u16 mem_mask);
+
+	// MMU control registers
+	u16 err_addr1_r(offs_t offset) { return m_err_addr1; }
+	u16 err_addr2_r(offs_t offset) { return m_err_addr2; }
+	u16 clr_violation_r(offs_t offset) { return m_violation = 0xffff; }
+	void clr_violation_w(offs_t offset, u16 data) { clr_violation_r(offset); }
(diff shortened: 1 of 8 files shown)


lkundrak:
I realize I'm asking for your time and effort, but any chances I could get a review?
Anything I can do to help with that?
Thanks!

rb6502:
@cracyc Can you look at the i86 change?  I'm gonna look at the rest, but I definitely want your eyes on that part :-)

cracyc:
A glance at the microcode says that out byte and out word both put all of AX on the bus so seems okay.

cuavas:
I’ll look at this.  It was marked draft, and GitHub doesn’t seem to send a notification when things are marked ready for review, so I didn’t see the change.

src/mame/altos/altos586_hdc.h
```diff
@@ -0,0 +1,83 @@
...
+
+	uint16_t m_cyl_latch;
+	uint16_t m_cyl[2];
+
+	uint8_t m_sector[517];
+	int m_secoffset;
+
+	uint8_t m_drive;
+	uint8_t m_head;
+	const struct hard_disk_file::info* m_geom[2];
```
cuavas, line 78: Please keep the data members and member functions grouped in each section – you’ve got data members above and below the member functions here.

src/mame/altos/altos586_hdc.h
```diff
@@ -0,0 +1,83 @@
...
+	void altos586_hdc_io(address_map &map);
+	void altos586_hdc_mem(address_map &map);
+
+private:
+	required_device<device_memory_interface> m_bus;
+	required_device<i8089_device> m_iop;
+	required_device<harddisk_image_device> m_hdd0;
+	required_device<harddisk_image_device> m_hdd1;
+
+	address_space *m_bus_mem;
```
cuavas, line 47: You should use `required_address_space` in place of `m_bus` and `m_bus_mem`.  It lets the user configure the address space number and takes care of validating that the address space exists.

src/mame/altos/altos586_hdc.cpp
```diff
@@ -0,0 +1,380 @@
...
+	, m_hdd1(*this, "drive1")
+{
+}
+
+bool altos586_hdc_device::sector_exists(uint8_t index)
+{
+	if (!m_geom[m_drive]) {
+		logerror("drive not present: %d", m_drive);
+	} else if (m_geom[m_drive]->sectorbytes != 512) {
+		logerror("expected 512 bytes per sector, got %d", m_geom[m_drive]->sectorbytes);
```
cuavas, line 74: Should this be checked when mounting a disk image rather than on every sector access?

cuavas:
This is structured pretty well and easy to read.  It should be easy enough to clean up and get it ready for merging.

lkundrak:
I've also spent some time figuring out what is needed to make XENIX boot. With a fix to how the 8254 timer is wired up, it appears to sometimes (like 1 in 50 attempts) goes as far loading the kernel and having it print something!
Screenshot from 2024-03-20 17-25-37

src/mame/mame.lst
```diff
@@ -581,6 +581,9 @@ altos486
 @source:altos/altos8600.cpp
 altos8600
 
+@source:altos/altos586.cpp
+altos586
+
 @source:amiga/alg.cpp
```
cuavas, line 587: Please keep this file in order – this should come after the `altos/altos486.cpp` section.

src/mame/altos/altos586_hdc.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+uint16_t altos586_hdc_device::mem_r(offs_t offset, uint16_t mem_mask)
+{
+	return m_bus->read_word_unaligned(offset << 1, mem_mask);
+}
+
+void altos586_hdc_device::mem_w(offs_t offset, uint16_t data, uint16_t mem_mask)
+{
+	m_bus->write_word_unaligned(offset << 1, data, mem_mask);
+}
```
cuavas, line 114: Can you just use `read_word` and `write_word` here?  You’ll only get aligned memory accesses through the address space read/write handlers, so it shouldn’t be necessary to use the ones that have extra code for unaligned accesses.

src/mame/altos/altos586_hdc.cpp
```diff
@@ -0,0 +1,371 @@
...
+	m_geom[index] = &m_hdd[index]->get_info();
+	return std::error_condition();
+}
+
+void altos586_hdc_device::device_add_mconfig(machine_config &config)
+{
+	I8089(config, m_iop, XTAL(15'000'000) / 3);
+	m_iop->set_addrmap(AS_PROGRAM, &altos586_hdc_device::altos586_hdc_mem);
+	m_iop->set_addrmap(AS_IO, &altos586_hdc_device::altos586_hdc_io);
+	m_iop->set_data_width(16);
```
cuavas, line 340: Does this board have its own clock, or does it use the bus clock for this?  If it uses the bus clock, you can use the `DERIVED_CLOCK` helper for the 8089 clock frequency and set the bus clock when instantiating the controller board.

src/mame/altos/altos586_hdc.h
```diff
@@ -0,0 +1,89 @@
...
+
+	uint16_t m_cyl_latch;
+	uint16_t m_cyl[2];
+
+	uint8_t m_sector[517];
+	int m_secoffset;
+
+	uint8_t m_drive;
+	uint8_t m_head;
+	const struct hard_disk_file::info* m_geom[2];
```
cuavas, line 83: You don’t seem to be initialising `m_geom` in the constructor or `device_start`.

src/mame/altos/altos586_hdc.h
```diff
@@ -0,0 +1,89 @@
...
+	// Disk access routines and state.
+	bool sector_exists(uint8_t index);
+	uint32_t sector_lba(uint8_t index);
+	void sector_read(uint8_t index);
+	void sector_write(uint8_t index);
+
+	// Image mount and unmount.
+	std::error_condition hdd_load(device_image_interface &image, uint8_t index);
+	std::error_condition hdd0_load(device_image_interface &image) { return hdd_load(image, 0); }
+	std::error_condition hdd1_load(device_image_interface &image) { return hdd_load(image, 1); }
```
cuavas, line 63: You can avoid the need for having three of these by making the drive index a template parameter.

src/mame/altos/altos586_hdc.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+void altos586_hdc_device::attn_w(uint16_t data)
+{
+	m_iop->ca_w(ASSERT_LINE);
+	m_iop->ca_w(CLEAR_LINE);
+}
+
+void altos586_hdc_device::device_start()
+{
+}
```
cuavas, line 371: You should at least do basic save state registration here (register you data members for save states).

src/mame/altos/altos586.cpp
```diff
@@ -0,0 +1,775 @@
...
+	// Managed I/O access from the bus (IOP and HDC)
+	void bus_io(address_map &map);
+	u16 bus_io_r(offs_t offset, u16 mem_mask);
+	void bus_io_w(offs_t offset, u16 data, u16 mem_mask);
+
+	// device-level overrides
+	virtual void device_start() override;
+
+	// device_memory_interface overrides
+	virtual space_config_vector memory_space_config() const override;
```
cuavas, line 117: Overridden `protected` member functions should be `protected`.

src/mame/altos/altos586.cpp
```diff
@@ -0,0 +1,775 @@
...
+	}
+
+	void altos586(machine_config &config);
+
+private:
+	// machine-level overrides
+	virtual void machine_start() override;
+
+	// device_memory_interface overrides
+	virtual space_config_vector memory_space_config() const override;
```
cuavas, line 402: Overridden `protected` member functions should be `protected`.

cuavas (commented):
This looks pretty good.  Just a few things I noticed.

src/mame/altos/altos586_hdc.cpp
lkundrak: Done, but I'm not sure it's correct -- I haven't yet used save/restore, so I need to figure out how to test it first.

PR #11676, opened 2023-10-29
new skeleton - Evolution  (Help needed to identify CPU type!)

mamehaze:
--------------------
Evolution Max [Team Europe, David Haywood]


src/devices/cpu/evolution/evo.cpp
```diff
@@ -0,0 +1,78 @@
...
+void evo_cpu_device::state_import(const device_state_entry &entry)
+{
+	switch (entry.index())
+	{
+		case EVO_PC:
+		case STATE_GENPCBASE:
+			m_pc = (m_debugger_temp & 0xffffffff);
+			break;
+	}
+}
```
cuavas, line 59: What’s the reason for not exporting `m_pc` directly?

src/devices/cpu/evolution/evo.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void evo_cpu_device::execute_run()
+{
+	while (m_icount > 0)
+	{
+		debugger_instruction_hook(m_pc);
+		m_pc++;
+		m_icount--;
+	}
```
cuavas, line 77: Would it be worth at least skipping two words for the jump instructions that are known to be two words long?  If you just want it to show disassembly and don’t care for execution at all, you can stub out the function with just `m_icount = 0;` and nothing else.

src/devices/cpu/evolution/evo.h
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef MAME_CPU_EVO_EVO_H
+#define MAME_CPU_EVO_EVO_H
+
+#pragma once
+
+enum
+{
+	EVO_PC = STATE_GENPC
+};
```
cuavas, line 12: Can you put this inside `evo_cpu_device` to keep the constants out of the global namespace?

src/devices/cpu/evolution/evo.h
```diff
@@ -0,0 +1,50 @@
...
+{
+public:
+	evo_cpu_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+protected:
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+	virtual uint32_t execute_min_cycles() const noexcept override { return 5; }
+	virtual uint32_t execute_max_cycles() const noexcept override { return 5; }
```
cuavas, line 24: This is potentially going to do strange things when the loop in `execute_run` has granularity of 1 while you’re returning 5 from `execute_min_cycles`.

src/mame/mame.lst
```diff
@@ -41390,6 +41390,9 @@ esprit3                         // Hazeltine Esprit III
 @source:skeleton/eurit.cpp
 eurit30                         //
 
+@source:skeleton/evolution_handheld.cpp
+evolhh
+
 @source:skeleton/eurocom2.cpp
 eurocom2                        //
```
cuavas, line 41397: Please put the source files in alphabetical order (v comes after u).

src/mame/skeleton/evolution_handheld.cpp
```diff
@@ -0,0 +1,82 @@
...
+// copyright-holders:David Haywood
+
+// TODO: identify the CPU type! - there are vectors(?) near the start
+
+#include "emu.h"
+
+#include "speaker.h"
+#include "screen.h"
+
+#include "cpu/evolution/evo.h"
```
cuavas, line 11: Please put headers from src/emu after headers from src/devices, and keep the headers from src/emu in alphabetical order.

PR #11677, opened 2023-10-29
misc/calomega.cpp: ROM '23-02.u6' from 'comg230' is a NO_DUMP, not a BAD_DUMP

clawgrip:
An empty dump should be a NO_DUMP, not a BAD_DUMP. Having it as a BAD_DUMP just causes noise when checking ROMs, and it contains no info at all.


RobertoFresca:
Great!... Now remove the whole Baby Formula set, since ALL of its ROMs are useless bad dumps failing in the same stuck lines, and therefore causing even more noise.
The set could be just listed in the driver.

PR #11688, opened 2023-10-31
nabupc: Add nabu floppy disk formats

brijohn:
This adds a custom format to properly support  NABU floppy disks.

The NABU supports the following disk types:
  * 5.25" SSDD - 200k
  * 5.25" DSDD - 400k
  * 3.5"   DSDD - 800k

Each track contains 5 1024 byte sectors.
  
The in order to determine disk type filesystem layout of a disk the NABU embeds a CP/M Disk Parameter Block in the index gap of track 0. This patch injects the correct DPB data into the disk based on its type when loading raw images.


src/lib/formats/nabupc_dsk.cpp
```diff
@@ -0,0 +1,229 @@
...
+int nabupc_format::identify(util::random_read &io, uint32_t form_factor, const std::vector<uint32_t> &variants) const
+{
+
+	int const format = find_format(io, form_factor, variants);
+
+	if (format == -1) {
+		return 0;
+	}
+
+	return FIFID_SIZE | FIFID_STRUCT;
```
galibert, line 79: That's a lie, you don't have any structural checking, only the size.

PR #11689, opened 2023-10-31
universal/cosmic.cpp: Dumped and added a new Space Panic set

clawgrip:
New WORKING clone
-----------------------------------------
Space Panic (set 4) [David Bosque, ClawGrip]

Also reformatted single line comments.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -44981,6 +44981,7 @@ nomnlndg                        // 8003 (c) [1980?] + Gottlieb
 panic                           // 8015 (c) 1980
 panic2                          // 8015 (c) 1980
 panic3                          // 8015 (c) 1980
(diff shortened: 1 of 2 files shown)


cuavas:
Have you examined the differences between this and the other sets at all?  It looks like all the ROMs are different.  Are the changes in code or data?

cuavas:
I’m just a bit concerned that it’s bit rot with some of these – the existing sets all share the same ROM at position 6, so it’s a surprise to find one that has all unique ROMs.  They’re over forty years old now, so the EPROMs could be losing data even if they’ve been stored properly.  With small changes like that, it could be a minor code update, but it could also be the beginning of bit rot.

I know disassembling the ROMs is a pile of work, but hopefully someone gives it the TCRF treatment at some point.

PR #8835, opened 2021-11-17
savquest: Updated and now boots to DOS prompt.

nickl-:
Still want to do more work on this but life started creeping in and before anything unforeseen happens lets just get these updates in.

As per the added driver comments:
```
Machine boots now, with voodoo card detected, and gives you command prompt after
failing to retrieve HASP key from the dongle. Still can't run SQ05 as the binary is SRM
enveloped and only prompts "HASP key not found". The SQ05 directory structure is also
impaired to a degree refusing to copy files or list files with dir, however tree works.

Voodoo custom PCI Headers are properly implemented now, allows for voodoo card to be
detected, see references for specification. This should probably be ported to library
video/voodoo_2 implementation. The card is initialized on boot but renders incorrectly,
perhaps the actual opengl translation from 3Dfx implementation is missing still.
To reset the display back to vga pass-through after boot run \sb16\diagnose

Without lpt controller added the boot script will go into a loop querying the lpt read and
write. Not convinced that does as advertised leaving it running for 24 hours still had no
HASP key. Adding the lpt controller the boot script detects the missing dongle and aborts,
giving access to the dos prompt.

82371AB not correctly emulated yet, as per previous dev's TODO comment on identification.
This is causing the IRQs to conflict as they are not properly allocated with no allocations
beyond IRQ 7, which appears to be the only usable channel. Tried to use the machine/i2371SB
implementation but at closer inspection noticed that SB is PIIX3 whereas AB is PIIX4 and these
specifications are different, see references.

The primary slave, cdrom, and secondary IDE drives are now detected. Secondary drive mounts the
same image as the primary master but read only, requires a mapped image to be detectable, and
is accessible but only on IRQ7 probably due to incomplete 82371AB emulation. Could not manage to
mount cdrom with any image to be usable, might be IRQ conflicts but could be incompatible with
the ATAPI driver.

Added timings for ISA and PCI bus as per PII Motherboard specification, see references.

Changed driver game flags to only prompt protection not implemented as this will ultimately be
the deciding impediment to playing the game even after the mentioned emulations are completed.
At least this will give us a working PIIX4 implementation booting into DOS6.11 to work from.
```

PR #11696, opened 2023-11-02
atari/mediagx: Adds Preliminary Input Handling, gets machine working for a51site4

montymxb:
First time committer. I was working on this driver for a good portion of last year, and never got around to putting a PR up. The driver still needs to be 'modernized', and uses some older approaches that need to be phased out -- to my understanding. Still, I figured it's helpful to share this and see if it's possible to add it in, or at least iterate from this point forward.
...

This PR adds input handling support for the atari/mediagx driver. Previously the machine was flagged as not working, and was capable of starting up with no real functionality. This PR implements a functional (but still incomplete) mapping for handling inputs properly. In practice this allows starting up & running a51site4  (albeit with performance limitations).

The changes include:
- P1/P2 start buttons work
- P1/P2 coin insertions work
- P1/P2 JGun X/Y tracking & triggers work (but JGuns may need to be calibrated in test menu first)
- Test/Diagnostic menu can be accessed as expected

As a result, a51site4 can be run and played (provided you have the hardware to mitigate performance issues).

Limitations:
- Emulation speed is inconsistent at times, sometimes slowing down or speeding up depending on the computational complexity at hand. Not the worst issue, but something to keep in mind.
- Performance varies based on your hardware, same as before essentially, but overall not very performant. Testing this on some newer hardware (an M1) produced buttery smooth results, but an older machine struggled to even do seconds per frame at times.
- Audio is really bad. It's playing in short fragments, very clippy, not so great. I did play around with the CPU clock rate in an attempt to see if the audio would come out clean (dropping it to some absurdly low values), and it did come out perfectly clean; however that's not really a fix and was just an experiment. I think the sound timer for the AD1847 can be adjusted to rectify this properly, but that's as far I got there.
- Emulator has to be manually stopped (ESC is not bound correctly yet). This is more of a limitation on my understanding on how to do this properly.
- Coin insertion doesn't work without a double tap most of the time, but beyond that works fine.
- Coin tracking is not quite right, i.e. statistics are logged mostly under slot 1, likely needs further refinement.
- Muzzle flash is absent (almost had it at one point with the state logic, but got a bit hairy to keep other things working as expected).
- Saving is still not supported.

### DIFF
--- a/src/mame/atari/mediagx.cpp
+++ b/src/mame/atari/mediagx.cpp
@@ -660,12 +660,43 @@ uint16_t mk_parport_outval(uint8_t v) {
 	return v2 << 8;
 }
 
+//#define READ_0x1_TEST 1
+// #define READ_0x18_TEST 1
+// #define READ_0x2_TEST 1
+// #define READ_0x3_TEST 1
+// #define READ_0x4_TEST 1
+// #define READ_0x5_TEST 1 
+// #define READ_0x6_TEST 1
+//#define READ_0xFF_TEST 1
+
+// for the very end, as if all were switched on, overrides all values as well
+//#define READ_BLANKET_TEST 1
+
+bool lock = false;
+int parallelPointMark = 10;
+
 // Reads from the parallel port, seems to hit what we're looking for?
 // 'offset' seems to be nearly constant in this case
 // the 'mem_mask' changes between (L) 0000FF00 and (H) 00FF0000
 uint32_t mediagx_state::parallel_port_r(offs_t offset, uint32_t mem_mask)
 {
 	uint32_t r = 0;
+
+	uint32_t mp00 = m_ports[0].read_safe(0);
+	if(mp00 == 0x4 && !lock) {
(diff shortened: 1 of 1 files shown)


angelosa:
As per #8835 the driver in general needs a real modernization to PCI specs so we aren't interested in pushing this PR as-is. However I'll use this as a consultation for making a proper parallel port bus slot for the inputs soon enough.

> Emulation speed is inconsistent at times, sometimes slowing down or speeding up depending on the computational complexity at hand. Not the worst issue, but something to keep in mind.

It's pretty rough and really depends on bus penalties (MediaGX is pretty known to not exactly be the most performant x86 arch ever).

montymxb:
@angelosa right I see. Well that still sounds good to me, so long as this helps makes progress overall.

Also, if there's anything I can do to help or clarify the changes I made here, feel free to let me know.

20Phoenix:
Would be aweome to get this nice Lightgun Shooter working next year.

PR #11699, opened 2023-11-03
karnov: interrupt enable feature added as per schematics

jotego:
The PCB can permanently disabled blanking interrupts, and it does so during power up. Writes to certain memory addresses will enable/disable the interrupt. Any write in the range will clear the current interrupt request. The logic in the schematic is not obvious, and it actually spawns to one more page at the end, but its functionality is modelled with this PR.

image

### DIFF
--- a/src/mame/dataeast/karnov.cpp
+++ b/src/mame/dataeast/karnov.cpp
@@ -153,6 +153,7 @@ private:
 
 	void videoram_w(offs_t offset, u16 data, u16 mem_mask = ~0);
 	void playfield_w(offs_t offset, u16 data, u16 mem_mask = ~0);
+	void vintctl_w(offs_t offset, u16 data, u16 mem_mask = ~0);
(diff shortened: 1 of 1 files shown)


src/mame/dataeast/karnov.cpp
```diff
@@ -633,6 +636,13 @@ void karnov_state::playfield_w(offs_t offset, u16 data, u16 mem_mask)
 	m_bg_tilemap->mark_tile_dirty(offset);
 }
 
+void karnov_state::vintctl_w(offs_t offset, u16 data, u16 mem_mask)
+{
+	m_vint_en = (offset & 1)!=0;
+	// writting to any position in the range will clear the line
```
angelosa, line 642: `m_vint_en = bool(offset & 1);` should work, also `writting`

jotego: Isn't `(offset & 1)!=0` a bool type already?

PR #11704, opened 2023-11-04
supracan: Hooked up internal 68k boot ROM, externalized key

MooglyGuy:
supracan: Hooked up internal 68k boot ROM, externalized key. [Ryan Holtz, wxhsplash]

### DIFF
--- a/src/mame/funtech/supracan.cpp
+++ b/src/mame/funtech/supracan.cpp
@@ -126,6 +126,10 @@ public:
 		, m_maincpu(*this, "maincpu")
 		, m_soundcpu(*this, "soundcpu")
 		, m_cart(*this, "cartslot")
+		, m_internal68(*this, "internal68")
+		, m_internal68_view(*this, "internal68")
+		, m_internal68_view_hi(*this, "internal68_hi")
+		, m_umc6650key(*this, "umc6650key")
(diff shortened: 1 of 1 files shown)


MooglyGuy:
There could be a better way of handling the views for the internal ROM. I'm open to suggestions.

angelosa:
> There could be a better way of handling the views for the internal ROM. I'm open to suggestions.

The best way is to move this to a bus slot that `address_map` to a single entity, cfr. bus/a800/a800_slot.cpp. 
It's imho a nice to have unless any of the games also needs EEPROM/backup RAM in the meantime.

MooglyGuy:
> > There could be a better way of handling the views for the internal ROM. I'm open to suggestions.
> 
> The best way is to move this to a bus slot that `address_map` to a single entity, cfr. bus/a800/a800_slot.cpp. It's imho a nice to have unless any of the games also needs EEPROM/backup RAM in the meantime.

The use of a memory view is based on how the Game Boy driver does it, and I'm more likely to believe that one's a more modern implementation. It's not the use of a view that I was wondering about, it's that it looks like the internal ROM is mirrored at 0xF80000 at one point during the last phases of theboot sequence. I'm currently handling it via a second view, but there might be a better way.

PR #11708, opened 2023-11-05
Migrate cv1k and nexus3d drivers to use nandflash.h instead of serflash.h

buffis:
serflash.h looks real old and sortof messy, and is pretty much implementing the basic functionality of K9F1G08U0 (it has that manufacturer/device ID hardcoded).

It's only used by cv1k and nexus3d.cpp (which I'll also migrate in a followup, removing serflash).

nandflash.h seems newer, has support for SAMSUNG_K9F1G08U0B and looks less jank

Migrating to it is pretty straightforward and seems to work.

Only difference is the write to 0x10C00003 which controls the CE pin of U2 (through the U13 CPLD). serflash seems to have a setter for this (flash_enab_w), but I don't really see how it's needed. As far as I can understand, its only used to signal that U2 should be the one to access the shared data-bus, which is a hardware limitation not relevant for emulation.

I assume that's why there's no handling of CE pin in nandflash.h/cpp

Tried a few different games, and seems to work fine


angelosa:
For nexus3d.cpp it's mildly interesting if you can check by patching in debugger the two `CMPS` listed in init and see if it doesn't diverge afterwards, but I honestly don't mind either way.

angelosa:
I checked thru the CI build, it looks like it's not loading the initial bootstrap properly with this.

buffis:
Thanks for spotting.
I've been messing around with the debugger, and I think this is a bug in nandflash.cpp

nexus3d does the following read sequence after a while:

```
00 -> 9c000010 # READ CYCLE1
00 -> 9c000018 # A COL0
00 -> 9c000018 # A COL1
01 -> 9c000018 # A ROW1
00 -> 9c000018 # A ROW2
00 -> 9c000018 # A ROW3
30 -> 9c000010 # READ CYCLE2
70 -> 9c000010 # READ STATUS
e0 <- 9c000000 # STATUS = OK 
00 -> 9c000010 # READ CYCLE1
XX <- 9c000000 # DATA READ ....
XX <- 9c000000 # DATA READ ....
XX <- 9c000000 # DATA READ .... 
...
```

The read status, followed by READCYCLE1 before reading looks kindof weird, but datasheet states under READ STATUS

"The command register remains in Status Read mode until further commands are issued to it. Therefore, if the status register is read during a random read cycle, the read command(00h) should be given before starting read cycles."

which I guess explains this.

nandflash.cpp will however reset the address when this triggers, which seems like a bug.

I'll look into fixing it.

buffis:
TL;DR for above is that this should not be executed in command_w (data=0x00)

```
m_page_addr = 0 
```

m_page_addr should only be updated on the second command cycle of a read command in nandflash. Should be easy to fix, but wont have time today probably

buffis:
Actually, turns out it was pretty easy. Haven't verified other nandflash users yet, which I should, but now for nexus32 with rom "acheart", you can trigger the bootstrapping by doing

```
> bp 0x1234
> run
> R0 = 1
> run
```

There will be some blinking text on the top of screen afterwards then, which seems cool

@angelosa , please take a look

angelosa:
Yeah, iirc after a while it will start dialoguing with the VRender3d FIFO with that debug trick in, never managed to make a sense over that one for the lack of any footage in the first place. :)

buffis:
Actually, the rabbit hole goes deeper I guess.

Some of the NAND types don't use a second read cycle (k9f5608u0d at least does not), which means that this will not work for that, so will have to separate out 1 and 2 cycle read IC's a bit.

I'll do that tomorrow or something. Don't merge this until then

angelosa:
Please mark as non-draft once ready.

galibert:
The call of 8b2cc at 8b834 returns ff in r0, while the code wants c4.  The c4 comes from that part of the flash:
000e9d80: 0000 0000 8000 c4ff ffea 4372 617a 7920  ..........Crazy 
000e9d90: 4172 6361 6465 2042 6e42 2028 5645 4741  Arcade BnB (VEGA
000e9da0: 3329 0000 0000 e813 9fe5 5000 81e5 0ef0  3)........P.....

The code at 8b2cc is rather complicated, I don't have time to analyze it right now :-)

  OG.

mamehaze:
Arcade BnB is... "bballoon          BnB Arcade" so is this some kind of recycled code?

src/mame/misc/cv1k.cpp
```diff
@@ -193,7 +193,7 @@ Timing
 #include "cpu/sh/sh3comn.h"
 #include "cpu/sh/sh4.h"
 #include "machine/rtc9701.h"
-#include "machine/serflash.h"
+#include "machine/nandflash.h"
```
cuavas, line 196: Please remember to get these back in alphabetical order.

buffis:
There, I think this is the cleanest way to support the Read Status during Read as described further up.

I verified that acheart, futari15, espgal2, palmz22, mrdrilrg and mini2440 works as before, to have good coverage of different NAND chips.

PR #11710, opened 2023-11-06
goodejan.cpp: Update sprite rendering

cam900:
goodejan.cpp: Fix sprite coordinate for goodejan, Use single-pass sprite rendering

### DIFF
--- a/src/mame/seibu/goodejan.cpp
+++ b/src/mame/seibu/goodejan.cpp
@@ -9,7 +9,7 @@ Sound: Z80 YM3812 M6295
 OSC: 12.000MHz 16.000MHz 7.15909MHz
 
 
-ToDo:
+TODO:
  Clean up inputs (some missing dips)
  Some sprite flickers on attract mode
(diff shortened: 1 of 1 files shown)


src/mame/seibu/goodejan.cpp
```diff
@@ -276,42 +272,62 @@ TILE_GET_INFO_MEMBER( goodejan_state::seibucrtc_sc3_tile_info )
...
-	int offs,fx,fy,x,y,color,sprite;
-	int dx,dy,ax,ay;
-
-	for (offs = 0x400-4;offs >= 0;offs -= 4)
+	u32 pri_mask[4] = {
+		GFX_PMASK_8,
+		GFX_PMASK_8 | GFX_PMASK_4,
+		GFX_PMASK_8 | GFX_PMASK_4 | GFX_PMASK_2,
+		0
+	};
```
cuavas, line 282: Should this be `constexpr`?

PR #11717, opened 2023-11-08
itech8.cpp: Split class related to sensor hardware and timer

cam900:
itech8.cpp: Fix logging, Use shorter type values

### DIFF
--- a/src/mame/itech/itech8_m.cpp
+++ b/src/mame/itech/itech8_m.cpp
@@ -88,8 +88,8 @@
  *************************************/
 
 #ifdef STANDALONE
-static void sensors_to_words(uint16_t sens0, uint16_t sens1, uint16_t sens2, uint16_t sens3,
-							uint16_t *word1, uint16_t *word2, uint16_t *word3, uint8_t *beams)
+static void sensors_to_words(u16 sens0, u16 sens1, u16 sens2, u16 sens3,
+							u16 *word1, u16 *word2, u16 *word3, u8 *beams)
 {
 	/* word 1 contains the difference between the larger of sensors 2 & 3 and the smaller */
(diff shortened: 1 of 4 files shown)


src/mame/itech/itech8_m.cpp
```diff
@@ -613,13 +630,13 @@ uint32_t itech8_state::screen_update_slikshot(screen_device &screen, bitmap_rgb3
 
 int main(int argc, char *argv[])
```
rb6502, line 631: This STANDALONE stuff can just be eliminated.

src/mame/itech/itech8_v.cpp
```diff
@@ -180,7 +180,7 @@ void itech8_state::palette_w(offs_t offset, uint8_t data)
 void itech8_state::page_w(u8 data)
 {
 	m_screen->update_partial(m_screen->vpos());
-	logerror("%04x:display_page = %02X (%d)\n", m_maincpu->pc(), data, m_screen->vpos());
+	logerror("%s:display_page = %02X (%d)\n", machine().describe_context(), data, m_screen->vpos());
```
rb6502, line 183: This should be updated to the modern logmacro.h / LOGMASKED().

src/mame/itech/itech8_v.cpp
```diff
@@ -477,7 +477,7 @@ void itech8_state::blitter_w(offs_t offset, uint8_t data)
 	}
 
 	/* debugging */
-	if (FULL_LOGGING) logerror("%04x:blitter_w(%02x)=%02x\n", m_subcpu->pcbase(), offset, data);
+	if (FULL_LOGGING) logerror("%s:blitter_w(%02x)=%02x\n", machine().describe_context(), offset, data);
```
rb6502, line 480: This should also be replaced with LOGMASKED().

PR #11724, opened 2023-11-12
Update some notes on dumps on namcos10.cpp

simzy39:
I added ** to the games known on the DU to have bad dumps:
- Gegege no Kitaro Yokai Yokocho Matsuri De Batoru Ja         Japan, GYM1 Ver.A
- Knockdown 2001 (KO 2001)   Japan, KD11 Ver. B
- Pacman Ball   World, PMB2 Ver.A
- "Sekai Kaseki Hakken   Japan, SKH1 Ver.A

 I also wrote down an undumped revision of Puzz Ball (PZBA Ver. C) found this week on ebay: https://www.ebay.com/itm/145259247988

### DIFF
--- a/src/mame/namco/namcos10.cpp
+++ b/src/mame/namco/namcos10.cpp
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:smf,windyfairy
+// copyright-holders:smf, windyfairy
 /***************************************************************************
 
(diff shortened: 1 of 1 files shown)


cuavas:
I don’t think the extra asterisks are helpful – we already have `BAD_DUMP` for bad dumps.  This just adds something that can get out of sync.

PR #11726, opened 2023-11-12
nec/pce*.cpp: Move super system card and arcade card emulation into its own files

cam900:
Use address_space instead read/write handler, Add primary Arcade Card Duo support
pce_cd.cpp: Use devcb for interrupt callback


hash/pce.xml
```diff
@@ -4502,4 +4502,17 @@ https://mametesters.org/view.php?id=6597
 		</part>
 	</software>
 
+	<software name="acardpro">
+		<description>Arcade Card Pro</description>
+		<year>19??</year>
```
angelosa, line 4507: https://necretro.org/Arcade_Card 1994

src/devices/bus/pce/pce_rom.cpp
```diff
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:Fabio Priuli
+// copyright-holders:Fabio Priuli, Charles MacDonald, Wilbert Pol, Angelo Salese
```
angelosa, line 2: Arcade Pro stuff should go to a separate file, and credited to me only (since it still uses the old and outdated i/f), likewise I think the base Super CD system is mostly a @wilbertpol product?

Regardless, considering it eventually needs Expansion bus bindings I believe having both as standalone files is for the best.

angelosa (changes requested):

src/mame/nec/pce_m.cpp
```diff
@@ -23,32 +23,41 @@ void pce_state::init_tg16()
...
+
+		// Install Arcade Card registers
+		if ((m_cartslot->get_type() == PCE_ACARD_DUO)
+			|| (m_cartslot->get_type() == PCE_ACARD_PRO))
+		{
+			m_maincpu->space(AS_PROGRAM).install_readwrite_handler(0x080000, 0x087fff, read8sm_delegate(m_cartslot, FUNC(pce_cart_slot_device::read_ram)), write8sm_delegate(m_cartslot, FUNC(pce_cart_slot_device::write_ram)));
+			m_maincpu->space(AS_PROGRAM).install_readwrite_handler(0x1ffa00, 0x1ffbff, read8sm_delegate(m_cartslot, FUNC(pce_cart_slot_device::peripheral_r)), write8sm_delegate(m_cartslot, FUNC(pce_cart_slot_device::peripheral_w)));
+		}
+	}
+
```
angelosa, line 50: This is hella leaky, the HuCard slot can only access up to `0x0fffff`, the rest belongs to expansion bus which afaik can map itself anywhere in the upper bank.
I feel like moving stuff like this is not much better than the old code ...

hash/pce.xml
```diff
@@ -4502,4 +4502,29 @@ https://mametesters.org/view.php?id=6597
...
+		<description>Arcade Card Duo</description>
+		<year>1994</year>
+		<publisher>NEC</publisher>
+		<part name="cart" interface="pce_cart">
+			<feature name="slot" value="acard_duo" />
+			<dataarea name="rom" size="262144">
+				<!-- bios not included -->
+			</dataarea>
+		</part>
+	</software>
```
angelosa, line 4515: Lacking a BIOS means this option cannot be emulated. You shouldn't guess, given your past history makes me wonder if you even tested all this stuff.

einstein95: The Arcade Card Duo only contains the RAM needed for the Arcade titles as the Duo/Super CD System² came with the System Card 3.0 inbuilt

src/mame/nec/pce_m.cpp
cam900: CPU A20 pin is tied into XCE pin of HuCard slot, so HuCard can accessible to over ~`0x100000` area.
reference: https://gamesx.com/wiki/doku.php?id=hu6280_netlist&s[]=hucard, [TurboGrafx-16 Unit Service Manual](https://gamesx.com/wiki/lib/exe/fetch.php?media=schematics:turbografx-16_unit_service_manual_-_smtg16.pdf)

angelosa: So it's a direct overlay, and my point still stand: this initialization should be local to the cart bus, not handled by the pce_m.cpp (which is a deprecated file bucket that should have stuff removed, not added in any way or form).

hash/pce.xml
angelosa: That again implies the Arcade Card Duo is an option slot for a base system there's no real handling for. 
I also have my doubts wrt this kind of fake entry in the SW list, looks an unnecessary bridge for something doable with `-listslots` alone.

einstein95: I see no reason why not to have it, as it'd take up the cart slot on the console much like RAM carts for like the Saturn

angelosa: Well it doesn't do anything to begin with, it needs a base machine for the Duo that internalizes the CD System Card and make sure it overrides the mapping if a card is present, or overlays in the case you want Arcade stuff.
https://nicole.express/2020/why-do-you-need-an-arcade-card.html

hash/pce.xml
```diff
@@ -4502,4 +4502,29 @@ https://mametesters.org/view.php?id=6597
...
+		<year>1994</year>
+		<publisher>NEC</publisher>
+		<part name="cart" interface="pce_cart">
+			<feature name="slot" value="acard_duo" />
+			<dataarea name="rom" size="1">
+				<!-- bios not included -->
+			</dataarea>
+		</part>
+	</software>
+
```
angelosa, line 4516: Let's put this as a rejected flag so you get it. 
It's an internal slot option, because NEC loves to make external and internal variants, it doesn't need to live as a fake SW entry. Either emulate it as a base `GAME` machine or get rid of it.

angelosa (changes requested):

src/devices/bus/pce/pce_scdsys.cpp
```diff
@@ -0,0 +1,112 @@
...
+		case 0x7: return 0x03;
+	}
+	return 0x00;
+}
+
+
+void pce_cdsys3_device::install_memory_handlers(address_space *space)
+{
+	space->install_rom(0x000000, 0x03ffff, m_rom);
+	space->install_readwrite_handler(0x0d0000, 0x0fffff, emu::rw_delegate(m_cdsys3, FUNC(pce_cdsys3_base_device::ram_r)), emu::rw_delegate(m_cdsys3, FUNC(pce_cdsys3_base_device::ram_w)));
```
cuavas, line 110: Why can’t you use `install_ram` to install the RAM in the appropriate address range?  The trampoline functions are inefficient.

src/devices/bus/pce/pce_scdsys.cpp
```diff
@@ -0,0 +1,112 @@
...
+//  mapper specific start/reset
+//-------------------------------------------------
+
+
+void pce_cdsys3_base_device::device_start()
+{
+	/* Set up Arcade Card RAM buffer */
+	m_ram = make_unique_clear<uint8_t[]>(0x30000);
+
+	save_pointer(NAME(m_ram), 0x30000);
```
cuavas, line 60: If you’re always instantiating the RAM and it’s always the same size, why not use a `memory_share_creator`?

src/devices/bus/pce/pce_scdsys.h
```diff
@@ -0,0 +1,84 @@
...
+	uint8_t register_r(offs_t offset);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override;
+
+private:
+	std::unique_ptr<uint8_t[]> m_ram; // internal RAM
+	bool m_region = false; // Cartridge region
+};
```
cuavas, line 34: Is there really a benefit of making this a full device rather than just a helper class that can be instantiated as a simple member at most?  The only thing it really implements is the “register” read stuff which is fairly trivial.

The RAM can be done using a `memory_share_creator` and installed with `install_ram`.

src/devices/bus/pce/pce_acard.cpp
```diff
@@ -0,0 +1,254 @@
...
+//  mapper specific start/reset
+//-------------------------------------------------
+
+
+void pce_acard_duo_device::device_start()
+{
+	/* Set up Arcade Card RAM buffer */
+	m_dram = make_unique_clear<uint8_t[]>(0x200000);
+
+	save_pointer(NAME(m_dram), 0x200000);
```
cuavas, line 55: You may as well use a `memory_share_creator` for this.

src/devices/bus/pce/pce_scdsys.cpp
```diff
@@ -0,0 +1,91 @@
...
+	, m_region(false)
+{
+}
+
+pce_cdsys3_device::pce_cdsys3_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, bool region)
+	: device_t(mconfig, type, tag, owner, clock)
+	, device_pce_cart_interface(mconfig, *this)
+	, m_cdsys3(*this, "cdsys3")
+{
+}
```
cuavas, line 38: This doesn’t seem to be doing anything with `region`.

angelosa (changes requested):
1004x

`mame pce sapphire` now punts instead of detecting Arcade Card off the bat.

src/devices/bus/pce/pce_acard.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+void pce_acard_duo_device::install_memory_handlers(address_space &space)
+{
+	space.install_readwrite_handler(0x080000, 0x087fff, emu::rw_delegate(*this, FUNC(pce_acard_duo_device::ram_r)), emu::rw_delegate(*this, FUNC(pce_acard_duo_device::ram_w)));
+	space.install_readwrite_handler(0x1ffa00, 0x1ffbff, emu::rw_delegate(*this, FUNC(pce_acard_duo_device::peripheral_r)), emu::rw_delegate(*this, FUNC(pce_acard_duo_device::peripheral_w)));
+}
+
+void pce_acard_pro_device::install_memory_handlers(address_space &space)
+{
+	space.install_rom(0x000000, 0x03ffff, 0x040000, m_rom); // TODO: actually 0x000000-0x07ffff
```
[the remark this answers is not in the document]
cam900: Arcade card BIOS is undumped currently and using scdsys BIOS for now, so needs to check size for it.

src/devices/bus/pce/pce_slot.h
```diff
@@ -96,11 +99,9 @@ class pce_cart_slot_device : public device_t,
 
 	void set_intf(const char * interface) { m_interface = interface; }
 
-	// reading and writing
-	uint8_t read_cart(offs_t offset);
-	void write_cart(offs_t offset, uint8_t data);
-
 protected:
+	required_address_space m_address_space;
```
cuavas, line 103: Please put this with the other data members (below the member function declarations).

src/devices/bus/pce/pce_slot.cpp
```diff
@@ -260,8 +264,8 @@ std::pair<std::error_condition, std::string> pce_cart_slot_device::call_load()
 
 		if (m_type == PCE_POPULOUS)
 			m_cart->ram_alloc(0x8000);
-		if (m_type == PCE_CDSYS3J || m_type == PCE_CDSYS3U)
-			m_cart->ram_alloc(0x30000);
+
+		m_cart->install_memory_handlers(*m_address_space.target());
 	}
```
cuavas, line 269: You should be able to just do `m_cart->install_memory_handlers(*m_address_space);` – object finders are dereferenceable.

src/devices/bus/pce/pce_scdsys.h
```diff
@@ -0,0 +1,84 @@
...
+						public device_pce_cart_interface
+{
+public:
+	// reading and writing
+	virtual void install_memory_handlers(address_space &space) override;
+
+protected:
+	// device-level overrides
+	virtual void device_start() override;
+	virtual void device_reset() override { }
```
cuavas, line 46: You don’t need to override `device_reset` if you aren’t doing anything special.

src/devices/bus/pce/pce_acard.h
```diff
@@ -0,0 +1,114 @@
...
+					m_addr_offset += m_addr_inc;
+				}
+			}
+		}
+
+		inline void adjust_addr()
+		{
+			m_base_addr += m_addr_offset + (BIT(m_ctrl, 3) ? 0xff0000 : 0);
+			m_base_addr &= 0xffffff;
+		}
```
cuavas, line 71: You don’t need the `inline` qualifier on member functions with inline bodies at declaration.

src/devices/bus/pce/pce_acard.cpp
```diff
@@ -0,0 +1,232 @@
...
+			case 2: m_shift = (data << 16)  | (m_shift & 0xff00ffff); break;
+			case 3: m_shift = (data << 24)  | (m_shift & 0x00ffffff); break;
+			case 4:
+				m_shift_reg = data & 0x0f;
+
+				if (m_shift_reg != 0)
+				{
+					m_shift = (m_shift_reg < 8) ?
+					(m_shift << m_shift_reg)
+					: (m_shift >> (16 - m_shift_reg));
```
cuavas, line 180: Please indent the continuation lines, e.g.:
```c++
					m_shift = (m_shift_reg < 8)
							? (m_shift << m_shift_reg)
							: (m_shift >> (16 - m_shift_reg));
```

src/devices/bus/pce/pce_acard.cpp
```diff
@@ -0,0 +1,232 @@
...
+				}
+				break;
+			case 5:
+				m_rotate_reg = data & 0x0f;
+
+				if (m_rotate_reg != 0)
+				{
+					m_shift = (m_rotate_reg < 8) ? 
+					((m_shift << m_rotate_reg) | (m_shift >> (32 - m_rotate_reg)))
+					: ((m_shift >> (16 - m_rotate_reg)) | (m_shift << (32 - (16 - m_rotate_reg))));
```
cuavas, line 190: Please indent the continuation lines, e.g.:
```c++
					m_shift = (m_rotate_reg < 8) 
							? ((m_shift << m_rotate_reg) | (m_shift >> (32 - m_rotate_reg)))
							: ((m_shift >> (16 - m_rotate_reg)) | (m_shift << (32 - (16 - m_rotate_reg))));
```

cuavas (commented):
This looks mostly OK now.  Getting this stuff into slot devices rather than machine configuration switches is a step forward.

@cam900 how many of the games that depend on the Arcade Cards have you tested?  How extensively have you tested them?

@angelosa did you have any other issues with this?  I think at this point it’s an improvement over what we’ve already got, and makes the Arcade Cards more self-contained so it’s easier to work on them in the future.

cuavas:
@cam900 can you please give a brief explanation of what you’ve tested to ensure the changes are working?

@angelosa can you please grab a CI build of this (or compile it yourself) and give it a bit of a test?

If testing looks OK, we can merge it after the freeze this month.

cam900:
@cuavas it's implements bit rotate register in Arcade Card, but only sapphire is tested and confirmed works fine.

angelosa:
> 1004x
> 
> `mame pce sapphire` now punts instead of detecting Arcade Card off the bat.

This still happens btw, I'm not even gonna bother with a full regtest until it's fixed.

cuavas:
There are some titles that can run with the Super CD system card, but will take advantage of the Arcade Card if it is available for enhancements. With the current state of this pull request, these games use the Super CD system card by default.

@cam900 can you please try to identify the software items that can take advantage of an Arcade Card and update them to use an arcade card by default? This will give a better experience.

angelosa (approved):

cuavas:
Let’s YOLO this in and hope it makes the release.  If there are CD titles that need their default system card changed, that’s easy to get in before the release goes out.  If someone finds something horribly broken that we overlooked, we can fix it or back it out temporarily.

Issue #10308, opened 2022-09-05
CHD does not store CD track indexes from cuesheet


Senemu:
<figure>
<figcaption>Example input cuesheet</figcaption>

```cue-sheet
FILE "purikura.bin" BINARY
  TRACK 01 MODE1/2352
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    INDEX 00 15:49:15
    INDEX 01 15:51:15
  TRACK 03 AUDIO
    INDEX 00 18:09:16
    INDEX 01 18:11:15
  TRACK 04 AUDIO
    INDEX 00 20:29:17
    INDEX 01 20:31:16
  TRACK 05 AUDIO
    INDEX 00 22:49:18
    INDEX 01 22:51:17
  TRACK 06 AUDIO
    INDEX 00 25:09:19
    INDEX 01 25:11:18
  TRACK 07 AUDIO
    INDEX 00 27:29:20
    INDEX 01 27:31:19
  TRACK 08 AUDIO
    INDEX 00 29:49:21
    INDEX 01 29:51:20
  TRACK 09 AUDIO
    INDEX 00 31:53:51
    INDEX 01 31:55:50
  TRACK 10 AUDIO
    INDEX 00 33:24:34
    INDEX 01 33:26:33
  TRACK 11 AUDIO
    INDEX 00 35:08:06
    INDEX 01 35:10:05
  TRACK 12 AUDIO
    INDEX 00 37:10:65
    INDEX 01 37:12:64
  TRACK 13 AUDIO
    INDEX 00 39:06:17
    INDEX 01 39:08:16
  TRACK 14 AUDIO
    INDEX 00 40:57:23
    INDEX 01 40:59:22
  TRACK 15 AUDIO
    INDEX 00 43:11:54
    INDEX 01 43:13:53
  TRACK 16 AUDIO
    INDEX 00 45:01:71
    INDEX 01 45:03:70
  TRACK 17 AUDIO
    INDEX 00 45:19:49
    INDEX 01 45:21:48
  TRACK 18 AUDIO
    INDEX 00 45:33:27
    INDEX 01 45:35:26
  TRACK 19 AUDIO
    INDEX 00 45:43:17
    INDEX 01 45:45:16
  TRACK 20 AUDIO
    INDEX 00 45:51:66
    INDEX 01 45:53:65
  TRACK 21 AUDIO
    INDEX 00 46:02:64
    INDEX 01 46:04:63
  TRACK 22 AUDIO
[...]

Issue #5867, opened 2019-11-06
chdman extractcd output is not the same as input on multiple audio bins


luke-70it:
It would be nice to obtain same cue+bin files used to convert to CHD, when extracting it.
Original CD dump consisting in multiple bins are converted back in a single bin file.

Zenjir0:
I too would like to see the output the same as the input, instead of having a merged single BIN file.

oblivioncth:
Would it be acceptable to add a switch for the extractcd command to specify if you want multi-track images as separate bins or not?

eddlang:
I think this is a rather important feature to have. If my input is split-bin files, I'd expect to get the same files after extraction.

MestreLion:
I was really surprised when I noticed that, not only split BIN files were not the default behavior of `extractcd`, but there was no option to make it so in `chdman` at all! Mame devs being so zealots about preservation, this was quite shocking.

Does `chdman` stores the original filenames, or does the CHD format even supports this?

rb6502:
MAME is preserving the actual CD data, not the intermediate rip format.  You can import an ISO or CDRDAO bin/toc rip and export it as bin/cue and vice-versa.  The data is what's important, not the cue file.

eddmanx:
> MAME is preserving the actual CD data, not the intermediate rip format. You can import an ISO or CDRDAO bin/toc rip and export it as bin/cue and vice-versa. The data is what's important, not the cue file.

You're correct that the data is preserved, and if you split the extracted bin into multiple ones their checksum would match the originals.

It's just that it's less user friendly, when the input was originally multiple bins.

Immersion95:
> > MAME is preserving the actual CD data, not the intermediate rip format. You can import an ISO or CDRDAO bin/toc rip and export it as bin/cue and vice-versa. The data is what's important, not the cue file.
> 
> You're correct that the data is preserved, and if you split the extracted bin into multiple ones their checksum would match the originals.
> 
> It's just that it's less user friendly, when the input was originally multiple bins.

Binmerge can do : Multiple bin/cue <-----> Single bin/cue
Oxyromon can do : chd <-----> Multiple bin/cue

eddmanx:
> Binmerge can do : Multiple bin/cue <-----> Single bin/cue Oxyromon can do : chd <-----> Multiple bin/cue

I'm well aware.

MestreLion:
[...]

PR #11727, opened 2023-11-12
refactor chdman extractcd to allow split .bin/.cue output

landfillbaby:
My partial solution to #5867 (only partial because it doesn't fix round-trip support fully, such as #10308).
Full list of things this does to `chdman`:
- Adds a `-os`/`--splitcue` option to `extractcd` to force output to `.cue` mode and output individual `.bin`s, naming them `basename (Track n).bin` if there is more than 1 track.
  - `n` will be 1 digit if there are less than 10 tracks in the CD, 2 digits otherwise.
  - `.bin` will be replaced by the file extension specified by `-ob`/`--outputbin` if applicable.
  - Its short name is `-os` in case the existing `-s`/`--size` option is used for `extractcd` in future.
- This new method also applies to `.gdi` output. Should I change this?
  - Output filenames will be different: no longer outputs audio files as `.raw`, format is as above instead of just appending the track number as 2 digits.
  - Deletes all bin files on error instead of just the last one before the error.
  - The percentage of completion no longer resets to 0 for each track, so it will actually say near 100% at completion.
- Checks output filename ends with `.cue` or `.gdi`, case insensitive, instead of just doing a case sensitive search for them anywhere in the filename.
- Removes unused `-ob`/`--outputbin` option from `extractdvd`.
- Updates its manpage, adding the missing `createdvd` and `extractdvd` documentation too.

Tell me if any changes should be made :)

### DIFF
--- a/src/tools/chdman.cpp
+++ b/src/tools/chdman.cpp
@@ -57,9 +57,18 @@ const uint32_t IDE_SECTOR_SIZE = 512;
 const uint32_t TEMP_BUFFER_SIZE = 32 * 1024 * 1024;
 
 // modes
-const int MODE_NORMAL = 0;
-const int MODE_CUEBIN = 1;
-const int MODE_GDI = 2;
+enum cd_mode
+{
+	MODE_NORMAL,
+	MODE_CUEBIN,
+	MODE_GDI,
+	MODE_CUEBIN_SPLIT,
+};
+
+static inline bool split_mode(cd_mode mode) { return mode == MODE_CUEBIN_SPLIT || mode == MODE_GDI; }
+static inline bool normal_mode(cd_mode mode) { return mode == MODE_NORMAL; }
+static inline bool cuebin_mode(cd_mode mode) { return mode == MODE_CUEBIN || mode == MODE_CUEBIN_SPLIT; }
+static inline bool gdi_mode(cd_mode mode) { return mode == MODE_GDI; }
 
 // command modifier
 #define REQUIRED "~"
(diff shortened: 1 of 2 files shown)


landfillbaby:
Just fixed another thing that was wrong with the previous `extractcd` GDI implementation: the completion percentage would reset to zero on each new track.

landfillbaby:
OK it's definitely ready for merge now! I fixed a regression I made in the 2nd commit!

rb6502:
Would it be possible to squash all the commits so it's easier to review?

landfillbaby:
done

src/tools/chdman.cpp
```diff
@@ -1448,7 +1460,7 @@ void output_track_metadata(int mode, util::core_file &file, int tracknum, const
 			file.printf("ZERO %s %s\n", modesubmode, msf_string_from_frames(info.pregap));
 
 		// all tracks but the first one have a file offset
-		if (tracknum > 0)
+		if (!split_mode(mode) && tracknum > 0)
```
landfillbaby, line 1463: This line is the one I'm not sure about.

landfillbaby:
ok actually split .toc doesn't work at all...

rb6502:
Yeah, as far as I know there's no such thing as split CDRDAO (bin/toc) format.

landfillbaby:
The option is now `--splitcue` and forces output to be in `.bin`/`.cue` mode, issuing a warning if `-o` doesn't end with `.cue`. There is no split `.toc` mode.

landfillbaby:
It is now absolutely ready for merge, unless you don't like the way I implemented the `--splitcue` forcing the output to `.cue` format.

987123879113:
re: .toc support
I'm not sure if cdrdao supports extracting .tocs with each track as a new file, but the .toc format itself does support using multiple files. Here's an example from the official cdrdao website showing it off at the bottom: https://cdrdao.sourceforge.net/example.html#secret

landfillbaby:
closing in favour of #12191 :)

Issue #10033, opened 2022-07-04
Taito F3 unemulated video effects


FredYeye:
_Elevator Action Returns_ has a visual effect that currently isn't emulated in mame: upon shooting the lamps hanging from the ceiling, the lights on that floor are supposed to go out.

I disassembled some of the code for the lamps and figured out that it's writing to 0x626200: line ram, offset 0x6200. _taito_f3_v.cpp_ labels this area as "Alpha blending control", but it seems to go unused (to me, who is unfamiliar with the code, anyway). I don't know what that is supposed to mean, but it appears to be 256 16-bit values (one u16 per line).

I looked at the effect in the saturn port to have some kind of reference as to what it's supposed to look like. Not the best as it could differ, but it seems identical. These are the values that gets written to line ram -

```
hex  | binary              | effect
-----------------------------------
BBBB | 1011 1011 1011 1011 | default brightness

BBFB | 1011 1011 1111 1011 | dark flashing
7B7B | 0111 1011 0111 1011 | bright flashing

7B8B | 0111 1011 1000 1011 | fade bright
7BAB | 0111 1011 1010 1011 | fade |
7BCB | 0111 1011 1100 1011 | fade |
7BEB | 0111 1011 1110 1011 | fade |
8BFB | 1000 1011 1111 1011 | fade v
ABFB | 1010 1011 1111 1011 | fade dark

CBFB | 1100 1011 1111 1011 | lights off
-----------------------------------

When a lamp hits the ground:
2f dark flashing
2f bright flashing
2f dark flashing
2f bright flashing
1f dark flashing
fade to dark from bright
lights off
```

Looking at the values makes me think that each nibble is some kind of control per layer. For now I'll just call it brightness control, as that is the effect they achieve. Two columns stay the same throughout this effect and two change. Layers 1 & 4 are not supposed to be affected. Layer 3 should change brightness. I don't know what layer 2 is, it's not used in stage 1. Might be something that changes brightness in later stages.

I guess my question here is, should I keep going with this? For now I've at least documented my findings, but I obviously can't know what the real PCB does with these values. I'm tempted to play around with the source code and attempt to apply the brightness control, but I'm at a loss as where to start. _taito_f3_v.cpp_ is not a crazy long file but I'm having a hard time following the code, never having looked at mame's codebase before.

Any help or comments are appreciated! If this data isn't enough try anything yet, let me know too.

-Fred

einstein95:
This footage confirms the effect is in the arcade version
[...]

PR #11728, opened 2023-11-12
Taito F3 cleanup

FredYeye:
I've looked at the taito F3 video code [before](https://github.com/mamedev/mame/issues/10033), although at the time I didn't get very far. I recently decided to look at it again. The code is a bit hard to follow, so as part of getting to understand it a bit better, I've done some cleanup / updates to the code.

There are 3 commits:

[taito_f3_v.cpp, taito_f3.h: turn code defines into functions](https://github.com/mamedev/mame/commit/87bff61c975cb3179d1b3cf8f920f02b023cda50)
`Switch + define` were being used as unrolled `for` loops, these have been turned into normal `for` loops. Some defines were turned into functions/lambdas. There are still 2 code defines (`UPDATE_PIXMAP_SP`, `UPDATE_PIXMAP_LP`) left. I didn't come up with a good replacement solution, so no point in replacing them at this time.

[taito_f3_v.cpp, taito_f3.h: use more appropriate data types for variables](https://github.com/mamedev/mame/commit/8a808e05bc9b115038587a7a97aa7bcf1717ec13)
A "narrowing pass". Maybe it's a lot of changes for one commit, though nothing has really changed. A lot of variables and function returns were just `int`, which made it hard to understand at a glance what the values could contain or mean.

[taito_f3_v.cpp, taito_f3.h: remove fixed-scale version of drawgfx](https://github.com/mamedev/mame/commit/5dac675ef0d244ab960534b26ab0add81bbe0dcb)
The only real "controversial" change. Tl;dr: delete `f3_drawgfx()`. Rename `f3_drawgfxzoom` to `f3_drawgfx`.

 `f3_drawgfx()` appears to be a fixed-scale version of `f3_drawgfxzoom`. Removing the fixed-scale version doesn't appear to affect performance negatively in any significant way. Also, as a bonus, `f3_drawgfxzoom` looks more modern. My only reservation about this is that it's hard to say `f3_drawgfxzoom` for sure replaces `f3_drawgfx()` in every case. I think it does, but `f3_drawgfx()` uses some vintage optimization so they're hard to compare.
There are also some other misc changes, like `m_add_sat` becoming `constexpr` and replacing a manual sorting loop with `std::sort`.

---

These changes are mainly for readability. The last commit especially would be a big win, reducing code duplication.
I don't expect any significant perf changes. Rudimentary benchmarking (-bench 100 for a couple of F3 games) shows a -2% change, but it varies a lot between runs.

There's a lot that can be done, but this is a good start! I'll gladly take feedback on all this.

### DIFF
--- a/src/mame/taito/taito_f3.h
+++ b/src/mame/taito/taito_f3.h
@@ -369,6 +369,8 @@ protected:
 	void dpix_1_sprite(u32 s_pix);
 	void dpix_bg(u32 bgcolor);
 	void init_alpha_blend_func();
+	void get_pixmap_pointer(const int skip_layer_num, const f3_playfield_line_inf **line_t, const int y);
+	void culc_pixmap_pointer(const int skip_layer_num);
 	void draw_scanlines(bitmap_rgb32 &bitmap, int xsize, s16 *draw_line_num, const f3_playfield_line_inf **line_t, const int *sprite, u32 orient, int skip_layer_num);
 	void visible_tile_check(f3_playfield_line_inf *line_t, int line, u32 x_index_fx, u32 y_index, const u16 *pf_data_n);
 	void calculate_clip(int y, u16 pri, u32 &clip_in, u32 &clip_ex, int &line_enable);
--- a/src/mame/taito/taito_f3.h
+++ b/src/mame/taito/taito_f3.h
@@ -231,7 +231,6 @@ protected:
 	u8 m_twidth_mask_bit = 0;
 	std::unique_ptr<u8[]> m_tile_opaque_sp;
 	std::unique_ptr<u8[]> m_tile_opaque_pf[8];
-	u8 m_add_sat[256][256];
 	int m_alpha_s_1_1 = 0;
 	int m_alpha_s_1_2 = 0;
 	int m_alpha_s_1_4 = 0;
@@ -310,7 +309,7 @@ protected:
 
 	void tile_decode();
 
-	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, const int code, const u8 color, const bool flipx, const bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
+	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, int code, u8 color, bool flipx, bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
 	void draw_sprites(bitmap_rgb32 &bitmap, const rectangle &cliprect);
 	void get_sprite_info(const u16 *spriteram16_ptr);
 	void print_debug_info(bitmap_rgb32 &bitmap);
@@ -369,8 +368,8 @@ protected:
 	void dpix_1_sprite(u32 s_pix);
 	void dpix_bg(u32 bgcolor);
 	void init_alpha_blend_func();
-	void get_pixmap_pointer(const int skip_layer_num, const f3_playfield_line_inf **line_t, const int y);
-	void culc_pixmap_pointer(const int skip_layer_num);
+	void get_pixmap_pointer(int skip_layer_num, const f3_playfield_line_inf **line_t, int y);
+	void culc_pixmap_pointer(int skip_layer_num);
 	void draw_scanlines(bitmap_rgb32 &bitmap, int xsize, s16 *draw_line_num, const f3_playfield_line_inf **line_t, const u8 *sprite, u32 orient, int skip_layer_num);
 	void visible_tile_check(f3_playfield_line_inf *line_t, int line, u32 x_index_fx, u32 y_index, const u16 *pf_data_n);
 	void calculate_clip(int y, u16 pri, u32 &clip_in, u32 &clip_ex, u8 &line_enable);
--- a/src/mame/taito/taito_f3.h
+++ b/src/mame/taito/taito_f3.h
@@ -225,13 +225,20 @@ protected:
 	u8 m_alpha_level_2bd = 0;
 	u8 m_alpha_level_3bs = 0;
 	u8 m_alpha_level_3bd = 0;
-	int m_alpha_level_last = 0;
+	u16 m_alpha_level_last = 0;
 	u16 m_width_mask = 0;
 	u8 m_twidth_mask = 0;
 	u8 m_twidth_mask_bit = 0;
 	std::unique_ptr<u8[]> m_tile_opaque_sp;
 	std::unique_ptr<u8[]> m_tile_opaque_pf[8];
-	u8 m_add_sat[256][256]{};
+	static constexpr std::array<std::array<u8, 256>, 256> m_add_sat = []
+	{
+		auto add_sat = std::array<std::array<u8, 256>, 256>{};
+		for (int i = 0; i < 256; i++)
+			for (int j = 0; j < 256; j++)
+				add_sat[i][j] = std::min(i + j, 255);
+		return add_sat;
+	}();
 	int m_alpha_s_1_1 = 0;
 	int m_alpha_s_1_2 = 0;
 	int m_alpha_s_1_4 = 0;
@@ -310,8 +317,7 @@ protected:
 
 	void tile_decode();
 
-	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, int code, u8 color, bool flipx, bool flipy, int sx, int sy, u8 pri_dst);
-	inline void f3_drawgfxzoom(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, int code, u8 color, bool flipx, bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
+	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, const int code, const u8 color, const bool flipx, const bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
 	void draw_sprites(bitmap_rgb32 &bitmap, const rectangle &cliprect);
 	void get_sprite_info(const u16 *spriteram16_ptr);
 	void print_debug_info(bitmap_rgb32 &bitmap);
--- a/src/mame/taito/taito_f3.h
+++ b/src/mame/taito/taito_f3.h
@@ -231,14 +231,7 @@ protected:
 	u8 m_twidth_mask_bit = 0;
 	std::unique_ptr<u8[]> m_tile_opaque_sp;
 	std::unique_ptr<u8[]> m_tile_opaque_pf[8];
-	static constexpr std::array<std::array<u8, 256>, 256> m_add_sat = []
-	{
-		auto add_sat = std::array<std::array<u8, 256>, 256>{};
-		for (int i = 0; i < 256; i++)
-			for (int j = 0; j < 256; j++)
-				add_sat[i][j] = std::min(i + j, 255);
-		return add_sat;
-	}();
+	u8 m_add_sat[256][256];
 	int m_alpha_s_1_1 = 0;
 	int m_alpha_s_1_2 = 0;
 	int m_alpha_s_1_4 = 0;
@@ -264,12 +257,12 @@ protected:
 	u8 m_tval = 0;
 	u8 m_pdest_2a = 0;
 	u8 m_pdest_2b = 0;
-	u8 m_tr_2a = 0;
-	u8 m_tr_2b = 0;
+	s8 m_tr_2a = 0;
+	s8 m_tr_2b = 0;
 	u8 m_pdest_3a = 0;
 	u8 m_pdest_3b = 0;
-	u8 m_tr_3a = 0;
-	u8 m_tr_3b = 0;
+	s8 m_tr_3a = 0;
+	s8 m_tr_3b = 0;
 	u16 *m_src[5]{};
 	u16 *m_src_s[5]{};
 	u16 *m_src_e[5]{};
@@ -345,8 +338,8 @@ protected:
 	inline void alpha_blend_3b_1(u32 s);
 	inline void alpha_blend_3b_2(u32 s);
 	bool dpix_1_noalpha(u32 s_pix);
-	constexpr bool dpix_ret1(u32 s_pix);
-	constexpr bool dpix_ret0(u32 s_pix);
+	inline bool dpix_ret1(u32 s_pix);
+	inline bool dpix_ret0(u32 s_pix);
 	bool dpix_1_1(u32 s_pix);
 	bool dpix_1_2(u32 s_pix);
 	bool dpix_1_4(u32 s_pix);
--- a/src/mame/taito/taito_f3.h
+++ b/src/mame/taito/taito_f3.h
@@ -169,17 +169,18 @@ protected:
 
 	struct tempsprite
 	{
-		int code = 0, color = 0;
-		int flipx = 0, flipy = 0;
+		int code = 0;
+		u8 color = 0;
+		bool flipx = 0, flipy = 0;
 		int x = 0, y = 0;
-		int zoomx = 0, zoomy = 0;
-		int pri = 0;
+		u16 zoomx = 0, zoomy = 0;
+		u8 pri = 0;
 	};
 
 	struct f3_playfield_line_inf
 	{
-		int alpha_mode[256]{};
-		int pri[256]{};
+		u8 alpha_mode[256]{};
+		u16 pri[256]{};
 
 		/* use for draw_scanlines */
 		u16 *src[256]{}, *src_s[256]{}, *src_e[256]{};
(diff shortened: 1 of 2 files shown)


src/mame/taito/taito_f3.h
```diff
@@ -337,43 +338,43 @@ class taito_f3_state : public driver_device
...
-	int dpix_3b_2(u32 s_pix);
-	int dpix_2_0(u32 s_pix);
-	int dpix_2_4(u32 s_pix);
-	int dpix_2_8(u32 s_pix);
-	int dpix_3_0(u32 s_pix);
-	int dpix_3_1(u32 s_pix);
-	int dpix_3_2(u32 s_pix);
+	bool dpix_1_noalpha(u32 s_pix);
+	constexpr bool dpix_ret1(u32 s_pix);
+	constexpr bool dpix_ret0(u32 s_pix);
```
cuavas, line 349: Having a `constexpr` function that isn’t also `const`-qualified is usually a bad sign.  Also, having a `constexpr` function that doesn’t have the body inline at declaration often makes trouble.  Considering how these are used, I don’t think marking them `constexpr` really helps.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -1358,12 +1358,12 @@ inline void taito_f3_state::draw_scanlines(
...
-	m_tr_2a =(m_alpha_level_2as == 0 && m_alpha_level_2ad == 255) ? -1 : 0;
-	m_tr_2b =(m_alpha_level_2bs == 0 && m_alpha_level_2bd == 255) ? -1 : 1;
+	m_tr_2a =(m_alpha_level_2as == 0 && m_alpha_level_2ad == 255) ? 255 : 0;
+	m_tr_2b =(m_alpha_level_2bs == 0 && m_alpha_level_2bd == 255) ? 255 : 1;
 	m_pdest_3a = m_alpha_level_3ad ? 0x40 : 0;
 	m_pdest_3b = m_alpha_level_3bd ? 0x80 : 0;
-	m_tr_3a =(m_alpha_level_3as == 0 && m_alpha_level_3ad == 255) ? -1 : 0;
-	m_tr_3b =(m_alpha_level_3bs == 0 && m_alpha_level_3bd == 255) ? -1 : 1;
+	m_tr_3a =(m_alpha_level_3as == 0 && m_alpha_level_3ad == 255) ? 255 : 0;
+	m_tr_3b =(m_alpha_level_3bs == 0 && m_alpha_level_3bd == 255) ? 255 : 1;
```
cuavas, line 1362: It’s probably clearer to change `m_tr_2a` `m_tr_2b`, `m_tr_3a` and `m_tr_3b` to `s8` and keep -1 as the “disabled” sentinel value.  It always compares these to a single bit (i.e. 0 or 1), and it’s more obvious if you use a negative value for the magic “should never compare equal” number.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2103,15 +2105,14 @@ void taito_f3_state::scanline_draw(bitmap_rgb32 &bitmap, const rectangle &clipre
...
 			{
-				int a = alpha_level;
-				const int b = (a >> 8) & 0xf;
-				const int c = (a >> 4) & 0xf;
-				const int d = (a >> 0) & 0xf;
-				a >>= 12;
+				const u8 a = alpha_level >> 12;
+				const u8 b = alpha_level >> 8 & 0xf;
+				const u8 c = alpha_level >> 4 & 0xf;
+				const u8 d = alpha_level >> 0 & 0xf;
```
cuavas, line 2105: Please leave the parentheses in where it makes things clearer (or use `BIT(alpha_level, 8, 4)` etc.).

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2764,8 +2766,8 @@ void taito_f3_state::get_sprite_info(const u16 *spriteram16_ptr)
 						x = block_x;
 					else
 					{
-						this_x = spriteram16_ptr[current_offs + 2 + 0];
-						if (this_x & 0x800) this_x = 0 - (0x800 - (this_x & 0x7ff)); else this_x &= 0x7ff;
+						this_x = spriteram16_ptr[current_offs + 2 + 0] & 0xfff;
+						if (this_x & 0x800) this_x -= 0x1000; // 12->16 sign extension
```
cuavas, line 2610: You can use `util::sext` for sign extension.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2792,8 +2795,8 @@ void taito_f3_state::get_sprite_info(const u16 *spriteram16_ptr)
 						y = block_y;
 					else
 					{
-						this_y = spriteram16_ptr[current_offs + 2 + 1] & 0xffff;
-						if (this_y & 0x800) this_y = 0 - (0x800 - (this_y & 0x7ff)); else this_y &= 0x7ff;
+						this_y = spriteram16_ptr[current_offs + 2 + 1] & 0xfff;
+						if (this_y & 0x800) this_y -= 0x1000; // 12->16 sign extension
```
cuavas, line 2639: You can use `util::sext` for sign extension.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2850,10 +2854,10 @@ void taito_f3_state::get_sprite_info(const u16 *spriteram16_ptr)
...
 
 			/* Sprite positioning */
-			this_y = spriteram16_ptr[current_offs + 2 + 1] & 0xffff;
-			this_x = spriteram16_ptr[current_offs + 2 + 0] & 0xffff;
-			if (this_y & 0x800) this_y = 0 - (0x800 - (this_y & 0x7ff)); else this_y &= 0x7ff;
-			if (this_x & 0x800) this_x = 0 - (0x800 - (this_x & 0x7ff)); else this_x &= 0x7ff;
+			this_y = spriteram16_ptr[current_offs + 2 + 1] & 0xfff;
+			this_x = spriteram16_ptr[current_offs + 2 + 0] & 0xfff;
+			if (this_y & 0x800) this_y -= 0x1000; // 12->16 sign extension
+			if (this_x & 0x800) this_x -= 0x1000; // 12->16 sign extension
```
cuavas, line 2700: You can use `util::sext` for sign extension.

src/mame/taito/taito_f3.h
```diff
@@ -369,6 +369,8 @@ class taito_f3_state : public driver_device
 	void dpix_1_sprite(u32 s_pix);
 	void dpix_bg(u32 bgcolor);
 	void init_alpha_blend_func();
+	void get_pixmap_pointer(const int skip_layer_num, const f3_playfield_line_inf **line_t, const int y);
+	void culc_pixmap_pointer(const int skip_layer_num);
```
cuavas, line 380: Please don’t use `const` parameters passed by value – compilers are inconsistent about whether the `const` becomes part of the mangled name, and it causes header churn if you need to change it for an implementation change.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -1258,33 +1262,43 @@ void taito_f3_state::init_alpha_blend_func()
...
-	m_x_zoom[pf_num] = line_tmp->x_zoom[y]; \
-	m_clip_al[pf_num] = line_tmp->clip_in[y] & 0xffff; \
-	m_clip_ar[pf_num] = line_tmp->clip_in[y] >> 16; \
-	m_clip_bl[pf_num] = line_tmp->clip_ex[y] & 0xffff; \
-	m_clip_br[pf_num] = line_tmp->clip_ex[y] >> 16; \
-	m_pal_add[pf_num] = line_tmp->pal_add[y]; \
+void taito_f3_state::get_pixmap_pointer(const int skip_layer_num, const f3_playfield_line_inf **line_t, const int y)
+{
+	for(int pf_num = skip_layer_num; pf_num < 5; ++pf_num)
+	{
```
cuavas, line 1264: Please be consistent with spacing – the rest of the file has a space between the flow control keyword and the opening parenthesis of the controlling expression.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2502,43 +2476,43 @@ inline void taito_f3_state::f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &
 						u32 *dest = dest0;
 						u8 *pri = pri0;
 
-						switch (x)
+						if (x_base > 15)
+						{
+							// pset_o
+							for(int x = x_base & 0xf; x > -1; --x)
+							{
```
cuavas: Inconsistent spacing here, too.

The `if` statement is really just checking bit 4.  It would be clearer as `if (x_base & 0x10)` (or `if (BIT(x_base, 4))`.  The count is the low four bits, so you might want to extract that to a local to make it more explicit.

Also, since you don’t actually use `x` in the loop, would it be clearer to express the loop expression as `(int = 0; x <= x_base & 0xf; ++x)`?

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2502,43 +2476,43 @@ inline void taito_f3_state::f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &
...
-							case  6: PSET_T NEXT_P [[fallthrough]];
-							case  5: PSET_T NEXT_P [[fallthrough]];
-							case  4: PSET_T NEXT_P [[fallthrough]];
-							case  3: PSET_T NEXT_P [[fallthrough]];
-							case  2: PSET_T NEXT_P [[fallthrough]];
-							case  1: PSET_T NEXT_P [[fallthrough]];
-							case  0: PSET_T
+							//pset_t
+							for(int x = x_base; x > -1; --x)
+							{
```
cuavas: Same comments as for the previous `for` loop – since you don’t actually need `x` in the loop, you can express it as an upward count and avoid the awkward comparison to -1 and dependency on signed numbers.  (Also, the inconsistent spacing.)

src/mame/taito/taito_f3.h
```diff
@@ -225,13 +225,20 @@ class taito_f3_state : public driver_device
...
 	std::unique_ptr<u8[]> m_tile_opaque_pf[8];
-	u8 m_add_sat[256][256]{};
+	static constexpr std::array<std::array<u8, 256>, 256> m_add_sat = []
+	{
+		auto add_sat = std::array<std::array<u8, 256>, 256>{};
+		for (int i = 0; i < 256; i++)
+			for (int j = 0; j < 256; j++)
+				add_sat[i][j] = std::min(i + j, 255);
+		return add_sat;
+	}();
```
cuavas, line 241: You shouldn’t use the `m_` prefix for things that behave like constants or for class statics.  But that aside, is this table even worthwhile?  Often it’s more efficient to just calculate stuff like this on-the-fly with modern CPUs.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -2265,6 +2259,8 @@ void taito_f3_state::scanline_draw(bitmap_rgb32 &bitmap, const rectangle &clipre
 		}
 
 		/* set scanline priority */
+		std::array<u8, 5> layer_tmp;
+		int count_skip_layer = 0;
```
cuavas, line 2263: Does changing `layer_tmp` to `std::array` actually help?  You can still use `std::sort` on a C array (using `std::begin` and `std::end` to get the begin/end iterators).

src/mame/taito/taito_f3.h
```diff
@@ -310,8 +317,7 @@ class taito_f3_state : public driver_device
 
 	void tile_decode();
 
-	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, int code, u8 color, bool flipx, bool flipy, int sx, int sy, u8 pri_dst);
-	inline void f3_drawgfxzoom(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, int code, u8 color, bool flipx, bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
+	inline void f3_drawgfx(bitmap_rgb32 &dest_bmp, const rectangle &clip, gfx_element *gfx, const int code, const u8 color, const bool flipx, const bool flipy, int sx, int sy, u16 scalex, u16 scaley, u8 pri_dst);
```
cuavas, line 320: What’s the performance impact of getting rid of the non-zooming version of this function?  The F3 graphics code can be pretty heavy, so having an optimised version for the 1:1 case could still be beneficial.

FredYeye:
Wow, thank you for taking your time to look through it! I'm a bit inexperienced with git / github. Would you prefer i edit the existing commits (where applicable) or make a new commit applying all changes?

src/mame/taito/taito_f3.h
FredYeye: I cloned the current master branch of mame and ran
```
taitoem -bench 12 arabianm
taitoem -bench 120 arabianm
taitoem -bench 120 elvactr
taitoem -bench 120 gekiridn
taitoem -bench 120 landmakr
taitoem -bench 120 lightbr
taitoem -bench 120 puchicar
taitoem -bench 120 quizhuhu
taitoem -bench 120 ridingf
```
twice, taking an average of all benches (minus the -bench 12 arabianm, that was the "warmup"). Then i removed `drawgfx()` and ran it again.

**average speed**
mame-master: 295.71%
fixed scale removed: 298.17%

My best guess for this result is that `drawgfxzoom` has newer style code.

angelosa:
New commits in your forked branch then push will also automatically update the PR, once merged the commit will be squash merged to master (i.e. N commits -> 1).

src/mame/taito/taito_f3.h
FredYeye: What is the preferred solution here, `const &` or simply dropping the `const`?

src/mame/taito/taito_f3.h
FredYeye: Alright! Changed these back to the previous `inline`.

src/mame/taito/taito_f3_v.cpp
FredYeye: Used BIT() here.

src/mame/taito/taito_f3_v.cpp
FredYeye: This has been changed to
```
this_x = spriteram16_ptr[current_offs + 2 + 0] & 0xfff;
this_x = util::sext(this_x, 12);
```
Question: should the `& 0xfff` stay still?

src/mame/taito/taito_f3.h
FredYeye: You're right. I thought it was neat idea but it's probably best to generate it at runtime. Reverted.

src/mame/taito/taito_f3_v.cpp
FredYeye: Ah, didn't know about std::begin & end! Changed back to C array and used those.

FredYeye:
I improved [get_line_ram_info()](https://github.com/mamedev/mame/pull/11728/commits/a0837aa4ce5d5e84e4df7f81f040327bf369ca3b), this also increased speed a bit. Still, I'd like to go back and look at my earlier commits to see what lowered perf in the first place.

src/mame/taito/taito_f3.h
```diff
@@ -337,41 +337,43 @@ class taito_f3_state : public driver_device
...
-	int dpix_3b_2(u32 s_pix);
-	int dpix_2_0(u32 s_pix);
-	int dpix_2_4(u32 s_pix);
-	int dpix_2_8(u32 s_pix);
-	int dpix_3_0(u32 s_pix);
-	int dpix_3_1(u32 s_pix);
-	int dpix_3_2(u32 s_pix);
+	bool dpix_1_noalpha(u32 s_pix);
+	inline bool dpix_ret1(u32 s_pix);
+	inline bool dpix_ret0(u32 s_pix);
```
y-ack, line 342: these seem to be specified `inline` inconsistently despite being used in the same place.
does `inline` even mean anything here when these are used to build a function table?

cuavas: It causes the compiler not to export the function symbols.  Depending on the object format and ABI, this may or may not have any effect.

src/mame/taito/taito_f3.h
cuavas: Just drop the `const` – it will be passed by value so local modification won’t leak back to the caller.

src/mame/taito/taito_f3_v.cpp
cuavas: I don’t think it’s necessary – `util::sext` is supposed to completely replace what’s in the upper bits.  Leaving it there isn’t going to hurt though.

src/mame/taito/taito_f3.h
cuavas: Have you tried removing `m_add_sat` altogether?  It’s only used in one function for three saturating additions:
```c++
inline void taito_f3_state::alpha_blend32_d(int alphas, u32 s)
{
	u8 *sc = (u8 *)&s;
	u8 *dc = (u8 *)&m_dval;
	dc[COLOR1] = m_add_sat[dc[COLOR1]][(alphas * sc[COLOR1]) >> 8];
	dc[COLOR2] = m_add_sat[dc[COLOR2]][(alphas * sc[COLOR2]) >> 8];
	dc[COLOR3] = m_add_sat[dc[COLOR3]][(alphas * sc[COLOR3]) >> 8];
}
```

You could change it to:
```c++
inline void taito_f3_state::alpha_blend32_d(int alphas, u32 s)
{
	u8 *sc = (u8 *)&s;
	u8 *dc = (u8 *)&m_dval;
	dc[COLOR1] = std::min<unsigned>(dc[COLOR1] + ((alphas * sc[COLOR1]) >> 8), 255U);
	dc[COLOR2] = std::min<unsigned>(dc[COLOR2] + ((alphas * sc[COLOR2]) >> 8), 255U);
	dc[COLOR3] = std::min<unsigned>(dc[COLOR3] + ((alphas * sc[COLOR3]) >> 8), 255U);
}
```

This will avoid keeping the 64KiB table hot.  It shouldn’t require conditional branches as the operations can all be represented as comparisons followed by conditional moves (or a vector compare/or/mask sequence if the compiler is smart enough to recognise it).  In many cases, modern CPUs perform better doing these kinds of calculations and conditional moves than fetching values from tables.

cuavas (commented):
This is looking good.  If you just drop the `const` on parameters passed by value, it can be merged.

src/mame/taito/taito_f3.h
FredYeye: Ah! Sorry, I misunderstood you. I removed m_add_sat and tried your change, seems good!

src/mame/taito/taito_f3.h
FredYeye: Alright, done! Removed const from pass-by-value params.

FredYeye:
At present, this is the -bench result (same as before, -bench 120 for 8 games, run twice and take avg):
```
no changes          (commit 632f0a0): 290.6%
all current changes (commit 5d390a2): 312.3%
```

These numbers should of course be taken with a grain of salt. But at least it doesn't seem to make it run slower!

The changes I've done here are fairly basic I guess but i'm pretty happy with this. As mentioned earlier, there's a lot that can be done (`u16 *m_src[5]{};` and friends can be pulled down to function scope instead of class scope for example). Won't make any more changes now though, unless I'm requested to do so!

cuavas:
This looks fairly safe, but I’d appreciate if people could play through a few games and see if things look OK.

FredYeye:
I ~~fed~~ played through a couple of games from start to finish, and played a few others a bit. Didn't notice any broken visuals (apart from already known issues).

cuavas:
> I ~fed~ played through a couple of games from start to finish, and played a few others a bit. Didn't notice any broken visuals (apart from already known issues).

Yeah, I’d just appreciate it if a couple more people gave it a second look.  Otherwise I’ll cop hell after the release if some issue was missed ;)

rb6502:
I played a few F3 games last night and didn't see anything out of place.

PR #11732, opened 2023-11-12
update lzma to 23.01

landfillbaby:


grant2258:
I cross compile mame with clang for the raspberry pi. Youll need to update the lua as well commit I done a while back is below for reference.

https://github.com/grant2258/mame/commit/ac80fa1cb4c707271c2f1df94e0609c552a6e639#diff-3bd5843b571999eb0c7bfcb98f31b22c5dc91c3f1eebdfc2d33dc2046e5504a0

qeeg:
Most of the time, newer versions of libraries a program uses are NOT drop-in replacements. You haven't bothered to check the code USING these libraries to see if it still works, and as shown by the fact that your build script changes only came AFTER someone told you you needed to do them, you clearly haven't even built the code like this. Test your shit, mate.

cuavas:
Thanks.  I took care of this with 79bef1e23049fdac25f1fefd132e0b6c7d2cc3a1.

PR #11735, opened 2023-11-13
heathkit/tlb.cpp: Add Imaginator I-100 for Heath H19/H89

mgarlanger:
Adds the Cleveland Codonics Imaginator I-100 option to the TLB(Terminal Logic Board) for the Heathkit H19 and H89 systems.

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -1471,19 +1471,6 @@ void heath_imaginator_tlb_device::device_start()
 	m_maincpu->space(AS_PROGRAM).install_readwrite_tap(0x8000, 0xbfff, "irq_update",
 		[this](offs_t offset, u8 &data, u8 mem_mask) { if (!machine().side_effects_disabled()) { tap_8000h(); } },
 		[this](offs_t offset, u8 &data, u8 mem_mask) { if (!machine().side_effects_disabled()) { tap_8000h(); } });
-
-	m_alphanumeric_mode_active = true;
-	m_graphics_mode_active = false;
-
-	m_allow_tlb_interrupts = true;
-	m_allow_imaginator_interrupts = false;
-	m_im2_d1_val = 1;
-
-	m_hsync_irq_raised = false;
-
-	m_mem_map = 1;
-
-	m_mem_view.select(m_mem_map);
 }
 
 void heath_imaginator_tlb_device::device_reset()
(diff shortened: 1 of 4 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -1390,6 +1417,269 @@ ioport_constructor heath_gp19_tlb_device::device_input_ports() const
...
+
+	m_allow_tlb_interrupts = true;
+	m_allow_imaginator_interrupts = false;
+	m_im2_d1_val = 1;
+
+	m_hsync_irq_raised = false;
+
+	m_mem_map = 1;
+
+	m_mem_view.select(m_mem_map);
```
cuavas, line 1486: Does this need to be done here?  It’s going to happen on reset anyway before anything starts executing.

src/mame/heathkit/tlb.cpp
```diff
@@ -1390,6 +1417,269 @@ ioport_constructor heath_gp19_tlb_device::device_input_ports() const
...
+{
+	map.unmap_value_high();
+
+	map(0x0000, 0x1fff).view(m_mem_view);
+
+	// H19 standard ROM
+	m_mem_view[0](0x0000, 0x1fff).rom().region("maincpu", 0).unmapw();
+
+	// GCP ROM mapped to 0x0000 on power-up/reset
+	m_mem_view[1](0x0000, 0x1fff).rom().region("maincpu", 0x2000).unmapw();
```
cuavas, line 1517: You don’t need to put `.unmapw()` if you aren’t trying to unmap something that’s already there.  If your intention is to silence the warning messages about unmapped writes, you can use `.nopw()`.

Also, since the only thing that’s ever mapped here is ROM and it’s never unmapped, you could use `.bankr(...)` rather than a view, and set it up in `device_start` (use a `required_memory_bank` as the object finder for it).

src/mame/heathkit/tlb.cpp
```diff
@@ -1390,6 +1417,269 @@ ioport_constructor heath_gp19_tlb_device::device_input_ports() const
...
+	heath_tlb_device::io_map(map);
+
+	// interrupt routing
+	map(0x10, 0x10).select(0x08).mirror(0x07).w(FUNC(heath_imaginator_tlb_device::config_irq_w));
+
+	// display enabling (graphics or alphanumeric)
+	map(0x30, 0x30).select(0xc0).mirror(0x0f).w(FUNC(heath_imaginator_tlb_device::display_enable_w));
+
+	map(0x50, 0x50).mirror(0x0f); // Prevent mapping to TLB when a4 is high
+	map(0xd0, 0xd0).mirror(0x0f); // Prevent mapping to TLB when a4 is high
```
cuavas, line 1587: @galibert does `map` and `mirror` with no handler make sense?  I thought this wouldn’t do anything.

cuavas:
Please consider using a memory bank for the ROM mapped at 0x0000-0x1fff to save the overhead of a memory view that isn’t really necessary.

PR #11742, opened 2023-11-13
igs/igs009.cpp: Dumped and added a new Grand Prix '98 set 

clawgrip:
New WORKING clone
------------------------
Grand Prix '98 (V100K, set 2) [Taksangs, charlie]


Osso13:
The program changes are legit, but the 2 u49 differ for only one byte (actually one bit). I believe the older dump is correct (on the right)
immagine

clawgrip:
> The program changes are legit, but the 2 u49 differ for only one byte (actually one bit). I believe the older dump is correct (on the right)

Yes, it's clearly a bad byte... We'll redump it. Thanks!!!!!

clawgrip:
Ok, we've redumped it several times, after cleaning the pins, with two different programmers, and we got the same result (with the bad byte).

Frankly, I'd use the ROM from the other set without flagging it ad BAD_DUMP, but adding a comment on the code, just in case this image surfaces again.

What do you think?

Thanks!!!!!

Osso13:
If it were up to me I wouldn't even add a comment. If romcmp says the ROMs match 99,9999% or similar, the changes should always be checked, because it's almost always just bitrot, especially in GFX ROM (in program ROMs sometimes it happens it's just a hacked copyright year / manufacturer)

clawgrip:
Agree. I've fixed it to use the U49 ROM from the other set.

Thanks again.

PR #11750, opened 2023-11-16
upd933: rework irq to use a timer

devinacker:
From my still-in-progress CZ-1 driver, this mostly just changes interrupt generation to ensure accurate timing. Also better handling of interrupt priority (probably) and a couple of other miscellaneous things for the aforementioned driver.

### DIFF
--- a/src/devices/sound/upd933.cpp
+++ b/src/devices/sound/upd933.cpp
@@ -197,9 +197,10 @@ void upd933_device::update_pending_irq()
 /**************************************************************************/
 void upd933_device::update_irq()
 {
-	if (m_irq_state != (m_cs & m_id & m_irq_pending))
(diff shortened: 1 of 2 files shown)


src/devices/sound/upd933.cpp
```diff
@@ -113,40 +123,85 @@ void upd933_device::device_clock_changed()
...
+}
+
+/**************************************************************************/
+void upd933_device::update_irq()
+{
+	if (m_irq_state != (m_cs & m_id & m_irq_pending))
+	{
+		m_irq_state ^= 1;
+		m_irq_cb(m_irq_state);
+	}
```
cuavas, line 204: If someone puts an unexpected value into `m_irq_state` (e.g. with the debugger or a Lua script, since they can access stuff registered for save states) this is going to behave *really* strangely.

PR #11754, opened 2023-11-18
seibu/dynduke.cpp: Use single-pass sprite rendering, Minor cleanups

cam900:
Rename tilemap related RAM and ROM regions
use BIT for bitfield
Use shorter/correct type values
Minor cleanup gfxdecode
Add notes
Fix spacing

### DIFF
--- a/src/mame/seibu/dynduke.cpp
+++ b/src/mame/seibu/dynduke.cpp
@@ -229,7 +229,7 @@ INPUT_PORTS_END
 static const gfx_layout charlayout =
 {
 	8,8,        /* 8*8 characters */
-	RGN_FRAC(1,2),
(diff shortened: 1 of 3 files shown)


happppp:
There is no concensus that extracting a bitmask like BIT(x, 0, 8) is cleaner than plain C code x & 0xff, or BIT(x, 4, 6) for x >> 4 & 0x3f.
So, unless you clean up a driver with a copyright holder known to prefer the 3-param BIT macro, best leave it as-is. It falls into code style that is up to the driver author. Though the same can be said about other changes such as u8 for uint8_t. Usually if I don't do a big contribution to a source file, I leave code style alone.

PR #11756, opened 2023-11-18
heathkit/tlb.cpp: Fix screensaver for superset slot device

mgarlanger:
To enable screensaver, the superset ROM sets the horizontal characters displayed (R1 of the 6845) to 0. 
The existing 6845 code would subtract 1 from this, causing the max_visible_x to be set to 0xffff. This value was considered invalid, and no further calls to update the screen were made. This caused the current text on the screen to be frozen during the screensaver, instead of cleared. If there is better way to handle this let me know. I had thought about on a valid -> invalid transition of the 6845 params, screen could be called to clear itself, but the PR change seemed more true to what the real device is probably doing.


cuavas (changes requested):
This needs more in-depth changes to the 6845 code.

When setting configuring the screen in `mc6845_device::reconfigure_parameters`, you need to:
* Check whether the configured vertical displayed (R6) or horizontal displayed (R1) is zero.
* If neither is zero, apply the -1 adjustment to the visible area before calling `screen().reconfigure(…)` and `m_reconfigure_cb` as rectangle coordinates are inclusive.
* If either the configured vertical displayed (R6) or horizontal displayed (R1) is zero, you should **attempt to maintain the previously configured visible area**.  However if the vertical or horizontal total or vertical total has reduced, you may need to reduce the visible area so it doesn’t exceed the total area.

In `mc6845_device::handle_line_timer`, you need to check whether the configured horizontal displayed (R1) is zero, and if it is, don’t assert the DE output or schedule the DE off timer.

In `mc6845_device::draw_scanline`, you need to check whether the configured horizontal displayed (R1) is zero, and if it is, always set the update row’s `de` argument to zero.

src/devices/video/mc6845.cpp
```diff
@@ -450,15 +470,15 @@ void mc6845_device::recompute_parameters(bool postload)
...
 
 	/* update only if screen parameters changed, unless we are coming here after loading the saved state */
-	if (postload ||
+	if ((postload ||
 		(horiz_pix_total != m_horiz_pix_total) || (vert_pix_total != m_vert_pix_total) ||
-		(max_visible_x != m_max_visible_x) || (max_visible_y != m_max_visible_y) ||
+		(visible_width != m_visible_width) || (visible_height != m_visible_height) ||
 		(hsync_on_pos != m_hsync_on_pos) || (vsync_on_pos != m_vsync_on_pos) ||
-		(hsync_off_pos != m_hsync_off_pos) || (vsync_off_pos != m_vsync_off_pos))
+		(hsync_off_pos != m_hsync_off_pos) || (vsync_off_pos != m_vsync_off_pos)) && !maintain_visible_area)
```
cuavas, line 477: You still want to reconfigure the screen if you’re maintaining the visible area, because the totals or sync positions could have changed.

src/devices/video/mc6845.cpp
```diff
@@ -417,8 +422,23 @@ void mc6845_device::recompute_parameters(bool postload)
...
+			visible_width = horiz_pix_total;
+			maintain_visible_area = false;
+		}
+
+		if (visible_height > vert_pix_total)
+		{
+			visible_height = vert_pix_total;
+			maintain_visible_area = false;
+		}
+	}
```
cuavas, line 441: You shouldn’t need to clear `maintain_visible_area` when you clip this.

src/devices/video/mc6845.cpp
```diff
@@ -406,6 +406,11 @@ int mc6845_device::vsync_r()
 
 void mc6845_device::recompute_parameters(bool postload)
 {
+	bool zero_horizontal_width = (m_horiz_disp == 0);
+	bool zero_vertical_height = (m_vert_disp == 0);
+	bool maintain_visible_area = zero_horizontal_width || zero_vertical_height;
+	uint8_t visual_adjustment = maintain_visible_area ? 0 : 1;
```
cuavas, line 412: You’ll still need to inset the visible area rectangle if you’re maintaining the visible area since your width/height are exclusive while `rectangle` coordinates are inclusive – you don’t need `visual_adjustment`.

src/devices/video/mc6845.cpp
```diff
@@ -473,23 +493,27 @@ void mc6845_device::recompute_parameters(bool postload)
...
-				//max_visible_y *= 2;
+				//visible_height *= 2;
 				//vert_pix_total *= 2;
 			}
 
 			if(m_show_border_area)
 				visarea.set(0, horiz_pix_total-2, 0, vert_pix_total-2);
 			else
-				visarea.set(0 + m_visarea_adjust_min_x, max_visible_x + m_visarea_adjust_max_x, 0 + m_visarea_adjust_min_y, max_visible_y + m_visarea_adjust_max_y);
+				visarea.set(0 + m_visarea_adjust_min_x, visible_width + m_visarea_adjust_max_x, 0 + m_visarea_adjust_min_y, visible_height + m_visarea_adjust_max_y);
```
cuavas, line 503: This is where you need to inset the visible area – it’s the `rectangle` right/bottom coordinates that are inclusive.

src/devices/video/mc6845.cpp
```diff
@@ -473,23 +493,27 @@ void mc6845_device::recompute_parameters(bool postload)
...
-				screen().configure(horiz_pix_total, vert_pix_total, visarea, refresh.as_attoseconds());
+			{
+				screen().configure(horiz_pix_total - visual_adjustment, vert_pix_total - visual_adjustment, visarea, refresh.as_attoseconds());
+			}
 
 			if(!m_reconfigure_cb.isnull())
-				m_reconfigure_cb(horiz_pix_total, vert_pix_total, visarea, refresh.as_attoseconds());
+			{
+				m_reconfigure_cb(horiz_pix_total - visual_adjustment, vert_pix_total - visual_adjustment, visarea, refresh.as_attoseconds());
+			}
```
cuavas, line 516: You don’t inset the total width/height – those are counts, not inclusive coordinates.

src/devices/video/mc6845.cpp
```diff
@@ -417,8 +422,23 @@ void mc6845_device::recompute_parameters(bool postload)
 	uint16_t vert_pix_total = (m_vert_char_total + 1) * video_char_height + m_vert_total_adj;
 
 	/* determine the visible area, avoid division by 0 */
-	uint16_t max_visible_x = m_horiz_disp * m_hpixels_per_column - 1;
-	uint16_t max_visible_y = m_vert_disp * video_char_height - 1;
+	uint16_t visible_width = m_horiz_disp * m_hpixels_per_column;
+	uint16_t visible_height = m_vert_disp * video_char_height;
```
cuavas, line 426: If you’re maintaining the visible area, you want to start with `m_visible_width` and `m_visible_height` here to avoid getting a zero in the first place.

cuavas (changes requested):

src/devices/video/mc6845.cpp
```diff
@@ -473,23 +490,32 @@ void mc6845_device::recompute_parameters(bool postload)
...
 
 			if(m_show_border_area)
+			{
 				visarea.set(0, horiz_pix_total-2, 0, vert_pix_total-2);
+			}
 			else
-				visarea.set(0 + m_visarea_adjust_min_x, max_visible_x + m_visarea_adjust_max_x, 0 + m_visarea_adjust_min_y, max_visible_y + m_visarea_adjust_max_y);
+			{
+				visarea.set(0 + m_visarea_adjust_min_x, visible_width - 1 + m_visarea_adjust_max_x, 0 + m_visarea_adjust_min_y,
+				 visible_height - 1 + m_visarea_adjust_max_y);
```
cuavas, line 504: If you’re going split the line, please indent the continuation by two tabs, and group lines logically, e.g. put the X and Y coordinates on their own lines, like:
```c++
			{
				visarea.set(
							0 + m_visarea_adjust_min_x, visible_width - 1 + m_visarea_adjust_max_x,
							0 + m_visarea_adjust_min_y, visible_height - 1 + m_visarea_adjust_max_y);
			}
```

src/devices/video/mc6845.cpp
```diff
@@ -417,8 +420,22 @@ void mc6845_device::recompute_parameters(bool postload)
 	uint16_t vert_pix_total = (m_vert_char_total + 1) * video_char_height + m_vert_total_adj;
 
 	/* determine the visible area, avoid division by 0 */
-	uint16_t max_visible_x = m_horiz_disp * m_hpixels_per_column - 1;
-	uint16_t max_visible_y = m_vert_disp * video_char_height - 1;
+	uint16_t visible_width = zero_horizontal_width ? m_visible_width : m_horiz_disp * m_hpixels_per_column;
+	uint16_t visible_height = zero_vertical_height ? m_visible_height : m_vert_disp * video_char_height;
```
cuavas, line 424: Please use parentheses for arguments to the conditional operator, like:
```c++
	uint16_t visible_width = zero_horizontal_width ? m_visible_width : (m_horiz_disp * m_hpixels_per_column);
	uint16_t visible_height = zero_vertical_height ? m_visible_height : (m_vert_disp * video_char_height);
```

src/devices/video/mc6845.cpp
```diff
@@ -769,7 +799,7 @@ TIMER_CALLBACK_MEMBER(mc6845_device::handle_line_timer)
 
 	/* Set VSYNC and DE signals */
 	set_vsync( new_vsync );
-	set_de( m_line_enable_ff ? true : false );
+	set_de( m_line_enable_ff && nonzero_horizontal_width ? true : false );
```
cuavas, line 802: Please use parentheses for arguments to the conditional operator, and since you’re editing the line anyway, change it to use “English” spacing around the partentheses, like:
```c++
	set_de((m_line_enable_ff && nonzero_horizontal_width) ? true : false);
```

src/mame/heathkit/tlb.cpp
```diff
@@ -1179,6 +1186,11 @@ void heath_superset_tlb_device::crtc_reg_w(offs_t reg, uint8_t val)
 	heath_tlb_device::crtc_reg_w(reg, val);
 }
 
+MC6845_BEGIN_UPDATE(heath_superset_tlb_device::crtc_begin_update)
+{
+	bitmap.fill(rgb_t::black(), cliprect);
+}
```
cuavas, line 1192: Is this actually necessary?  Shouldn’t it get a call to `heath_superset_tlb_device::crtc_update_row` with `de` set to zero that will result in the rows being cleared if the screen saver is active?

cuavas (commented):
OK, the logic is looking more-or-less sound at this point.  However, given how widespread the 6845 family are, I’d like to hold off merging it until the start of the next release cycle (about a week from now).

Can you give a few 6845-based systems a basic check to see if they still seem to work before then?

src/mame/heathkit/tlb.cpp
mgarlanger: Without this, the screen is still not being erased. I'll try to determine why the update row code is not clearing, but right now the `-log` option to mame is not providing the expected logs.

mgarlanger: @cuavas The reason that `heath_superset_tlb_device::crtc_update_row` is not clearing it, is that the x_count being passed in is 0, since it uses the `m_horiz_disp` variable, which is R1 and that is being set to zero by the firmware.

mgarlanger:
> OK, the logic is looking more-or-less sound at this point. However, given how widespread the 6845 family are, I’d like to hold off merging it until the start of the next release cycle (about a week from now).
> 
> Can you give a few 6845-based systems a basic check to see if they still seem to work before then?

Holding off until next cycle is fine. I've brought up the Super-80, ApricotPC, and Victor 9000, and they seem fine.

cuavas:
This should be harmless, but now would be a good time to test a few systems with 6845-based CRTCs.

Tafoid:
A number of machines from the following drivers are showing exception or simply drop to command prompt when run after this commit.  I've included all that I saw (there may be more in the drivers listed that I missed).  Verified against a binary from this AM (23-12-23) as well as the action binaries for this commit and one prior to it:

b128 - commodore/cbm2.cpp
b256 - commodore/cbm2.cpp
b500 - commodore/cbm2.cpp
cbm4032f - commodore/pet.cpp
cbm610 - commodore/cbm2.cpp
cbm620 - commodore/cbm2.cpp
cbm620_hu - commodore/cbm2.cpp
cbm8032 - commodore/pet.cpp
cbm8032_de - commodore/pet.cpp
cbm8032_se - commodore/pet.cpp
cbm8096 - commodore/pet.cpp
cbm8296 - commodore/pet.cpp
cbm8296d - commodore/pet.cpp
cbm8296d_de - commodore/pet.cpp
cbm8296dgv_de - commodore/pet.cpp
cbm8296ed - commodore/pet.cpp
cbm8296gd - commodore/pet.cpp
pet4032f - commodore/pet.cpp
pet8032 - commodore/pet.cpp
prof80 - conitec/prof80.cpp

mgarlanger:
Is there specific software I need to track down to repo this? I tried to repo it with a recent branch (I happened to be on commit b81647908cec20591a0ba547cc3739a18d614c80). I ran b128 with `[start empty]` and basic started up and seemed to work fine.

Screenshot 2023-12-23 at 11 53 02 PM

Robbbert:
Just tried with latest git.

C:\MAME>mame b128
OPTIONAL 901885-01.uk3 NOT FOUND (tried in c8050 b128 b500)
WARNING: the machine might not run correctly.

-----------------------------------------------------
Exception at EIP=00007ff6ed9d48c0 ((anonymous namespace)::cbm2_state::crtc_update_row(bitmap_rgb32&, rectangle const&, unsigned short, unsigned char, unsigned short, unsigned char, signed char, int, int, int)+0x01d0): ACCESS VIOLATION
While attempting to write memory at 000002cd2bb1e000
-----------------------------------------------------
RAX=0000000000000000 RBX=000002cd23946dd0 RCX=000002cd205738b0 RDX=000000000000001c
RSI=0000000000000000 RDI=00000000ffffffff RBP=000002cd20498f10 RSP=000000ce2dcf9730
 R8=000000000000022d  R9=000000000000002a R10=0000000000000106 R11=0000000000000000
R12=000000000000098a R13=000002cd2b9f9040 R14=00000000000493f0 R15=00000000ff000000
-----------------------------------------------------
Stack crawl:
  000000ce2dcf9780: 00007ff6ed9d48c0 ((anonymous namespace)::cbm2_state::crtc_update_row(bitmap_rgb32&, rectangle const&, unsigned short, unsigned char, unsigned short, unsigned char, signed char, int, int, int)+0x01d0)
  000000ce2dcf9840: 00007ff6f19198c4 (mc6845_device::draw_scanline(int, bitmap_rgb32&, rectangle const&)+0x01f4)
  000000ce2dcf9920: 00007ff6f191e425 (mc6845_device::screen_update(screen_device&, bitmap_rgb32&, rectangle const&)+0x0255)
  000000ce2dcf99c0: 00007ff6f1741060 (screen_device::update_partial(int)+0x0290)
  000000ce2dcf9a20: 00007ff6f50a3d13 (video_manager::finish_screen_updates()+0x00c3)
  000000ce2dcf9ab0: 00007ff6f50a5d88 (video_manager::frame_update(bool)+0x01f8)
  000000ce2dcf9b20: 00007ff6f173e8f3 (screen_device::vblank_begin(int)+0x0233)
  000000ce2dcf9b90: 00007ff6f1757adb (device_scheduler::timeslice()+0x014b)
  000000ce2dcf9d10: 00007ff6f1754cbf (running_machine::run(bool)+0x01bf)
  000000ce2dcff240: 00007ff6f4efe8ab (mame_machine_manager::execute()+0x024b)
  000000ce2dcff630: 00007ff6f90e0f26 (cli_frontend::start_execution(mame_machine_manager*, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)+0x03e6)
  000000ce2dcff910: 00007ff6f90e154b (cli_frontend::execute(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&)+0x007b)
  000000ce2dcff970: 00007ff6f4ef9069 (emulator_info::start_frontend(emu_options&, osd_interface&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >&)+0x0029)
  000000ce2dcffd30: 00007ff6fad5f983 (luaopen_lfs+0xc224d3)
  000000ce2dcffe00: 00007ff6ece713b1 (__tmainCRTStartup+0x0231)
  000000ce2dcffe30: 00007ff6ece714e6 (mainCRTStartup+0x0016)
  000000ce2dcffe60: 00007fffff657344 (BaseThreadInitThunk+0x0014)
  000000ce2dcffee0: 00007ff8006626b1 (RtlUserThreadStart+0x0021)

mgarlanger:
Strange, I just tried with the latest git, and not seeing any problems:

```mame % ./mame b128
OPTIONAL 901885-01.uk3 NOT FOUND (tried in c8050 b128 b500)
WARNING: the machine might not run correctly.
Average speed: 100.00% (11 seconds)
mame %```

I wonder if there is some difference between x86 and apple m1.

**Comment by Robbbert** [comment]
It errors when it tried to write line 241 on the screen (y = 240, ma = 2400). I'm not familiar with this system, and the code in the machine config is far from clear, but I assume the screen should only have 240 lines.

**Comment by mgarlanger** [comment]
> It errors when it tried to write line 241 on the screen (y = 240, ma = 2400). I'm not familiar with this system, and the code in the machine config is far from clear, but I assume the screen should only have 240 lines.

If you have a chance, could you try with changing this one line back so it initializes those variables to 0:  https://github.com/mamedev/mame/pull/11756/files#diff-bd616883e4ae4ed2edf53526db4cd2c8f93ddbe1b5d8bea9abfe0e15856e0ba3R1101 and see if it still crashes?

**Comment by Osso13** [comment]
> > It errors when it tried to write line 241 on the screen (y = 240, ma = 2400). I'm not familiar with this system, and the code in the machine config is far from clear, but I assume the screen should only have 240 lines.
> 
> If you have a chance, could you try with changing this one line back so it initializes those variables to 0: https://github.com/mamedev/mame/pull/11756/files#diff-bd616883e4ae4ed2edf53526db4cd2c8f93ddbe1b5d8bea9abfe0e15856e0ba3R1101 and see if it still crashes?

Problem still happens here.

**Comment by mgarlanger** [comment]
🤔  not sure how I'll be able to debug this since I can't repro on my mac and I don't have a wiindows dev machine.

**Comment by rb6502** [comment]
The key is the report "It errors when it tried to write line 241 on the screen (y = 240, ma = 2400)".  Just add a printf or assert or whatever to that update function if y=240, because that means you're trashing memory off the end of the frame buffer.

**Comment by mgarlanger** [comment]
> The key is the report "It errors when it tried to write line 241 on the screen (y = 240, ma = 2400)". Just add a printf or assert or whatever to that update function if y=240, because that means you're trashing memory off the end of the frame buffer.

Added some `logerror()` statements and it is going up to y=260, but the screen size is set to 768 x 312 - https://github.com/mamedev/mame/blob/8c328688fe7e2bd68b1bd52835c32ea884f511c9/src/mame/commodore/cbm2.cpp#L2521 - according to the function docs, this defines the bitmap size - https://github.com/mamedev/mame/blob/8c328688fe7e2bd68b1bd52835c32ea884f511c9/src/emu/screen.h#L276-L288.  So there shouldn't be a problem at y=240.

**Comment by angelosa** [comment]
Why I feel like this is very typical screen_device display blank setup, given that driver actually sets up 1142 x 261 later when it manages to initalize CRTC?

**Comment by mgarlanger** [comment]
The initialized CRTC is: 
```
M6845 config screen: HTOTAL: 1143  VTOTAL: 262  VIS_WIDTH: 720  VIS_HEIGHT: 200  HSYNC: 882-971  VSYNC: 224-239  Freq: 60.106990fps
```
so it line y=240 should still be fine(and it is on my Mac). I have no idea what is causing the Windows systems to crash.

**Comment by angelosa** [comment]
It's not fine lol, CRTC sets up the configuration while (very likely) the vpos beam is between 262~312, causing the crash.
Should probably use a `screen().reset_origin()` somewhere, alternatively replacing that with `set_raw` should workaround it ...

**Comment by mgarlanger** [comment]
> It's not fine lol, CRTC sets up the configuration while (very likely) the vpos beam is between 262~312, causing the crash. Should probably use a `screen().reset_origin()` somewhere, alternatively replacing that with `set_raw` should workaround it ...

It's at least fine on the Mac 🙈   So would the `screen().reset_origin()` go here: https://github.com/mamedev/mame/blob/8c328688fe7e2bd68b1bd52835c32ea884f511c9/src/devices/video/mc6845.cpp#L512-L515 ?

Do you think my PR just changed the timing to expose this existing bug?

**Comment by rb6502** [comment]
No, look again: M6845 config screen: HTOTAL: 1143  VTOTAL: 262  VIS_WIDTH: 720  VIS_HEIGHT: 200  HSYNC: 882-971  VSYNC: 224-239  Freq: 60.106990fps

The callback is being told to draw visible content 40 scanlines after the end of the visible screen.  So you are *far* into trashing memory at that point, and it's working on your system purely by accident.  Memory trashing bugs can be very OS and compiler specific as to if they crash on a given build.  Even single-driver vs full builds can be different on if what gets trashed will crash.

**Comment by mgarlanger** [comment]
> No, look again: M6845 config screen: HTOTAL: 1143 VTOTAL: 262 VIS_WIDTH: 720 VIS_HEIGHT: 200 HSYNC: 882-971 VSYNC: 224-239 Freq: 60.106990fps
> 
> The callback is being told to draw visible content 40 scanlines after the end of the visible screen. So you are _far_ into trashing memory at that point, and it's working on your system purely by accident. Memory trashing bugs can be very OS and compiler specific as to if they crash on a given build. Even single-driver vs full builds can be different on if what gets trashed will crash.

I don't see that. I had it print y and de values for the first screen drawn after values became valid and they look fine. Here is the portion of the logs:
```
[:u10] M6845 config screen: HTOTAL: 1143  VTOTAL: 262  VIS_WIDTH: 720  VIS_HEIGHT: 200  HSYNC: 882-971  VSYNC: 224-239  Freq: 60.106990fps
[:u10] M6845: Valid screen parameters - display reenabled!!!
[:u10] draw_scanline y: 0 - de: 1
[:u10] draw_scanline y: 1 - de: 1
[:u10] draw_scanline y: 2 - de: 1
[:u10] draw_scanline y: 3 - de: 1
[:u10] draw_scanline y: 4 - de: 1
[:u10] draw_scanline y: 5 - de: 1
[:u10] draw_scanline y: 6 - de: 1
[:u10] draw_scanline y: 7 - de: 1
[:u10] draw_scanline y: 8 - de: 1
[:u10] draw_scanline y: 9 - de: 1
[:u10] draw_scanline y: 10 - de: 1
[:u10] draw_scanline y: 11 - de: 1
[:u10] draw_scanline y: 12 - de: 1
[:u10] draw_scanline y: 13 - de: 1
[:u10] draw_scanline y: 14 - de: 1
[:u10] draw_scanline y: 15 - de: 1
[:u10] draw_scanline y: 16 - de: 1
[:u10] draw_scanline y: 17 - de: 1
[:u10] draw_scanline y: 18 - de: 1
[:u10] draw_scanline y: 19 - de: 1
[:u10] draw_scanline y: 20 - de: 1
[:u10] draw_scanline y: 21 - de: 1
[:u10] draw_scanline y: 22 - de: 1
[:u10] draw_scanline y: 23 - de: 1
[:u10] draw_scanline y: 24 - de: 1
[:u10] draw_scanline y: 25 - de: 1
[:u10] draw_scanline y: 26 - de: 1
[:u10] draw_scanline y: 27 - de: 1
[:u10] draw_scanline y: 28 - de: 1
[:u10] draw_scanline y: 29 - de: 1
[:u10] draw_scanline y: 30 - de: 1
[:u10] draw_scanline y: 31 - de: 1
[:u10] draw_scanline y: 32 - de: 1
[:u10] draw_scanline y: 33 - de: 1
[:u10] draw_scanline y: 34 - de: 1
[:u10] draw_scanline y: 35 - de: 1
[:u10] draw_scanline y: 36 - de: 1
[:u10] draw_scanline y: 37 - de: 1
[:u10] draw_scanline y: 38 - de: 1
[:u10] draw_scanline y: 39 - de: 1
[:u10] draw_scanline y: 40 - de: 1
[:u10] draw_scanline y: 41 - de: 1
[:u10] draw_scanline y: 42 - de: 1
[:u10] draw_scanline y: 43 - de: 1
[:u10] draw_scanline y: 44 - de: 1
[:u10] draw_scanline y: 45 - de: 1
[:u10] draw_scanline y: 46 - de: 1
[:u10] draw_scanline y: 47 - de: 1
[:u10] draw_scanline y: 48 - de: 1
[:u10] draw_scanline y: 49 - de: 1
[:u10] draw_scanline y: 50 - de: 1
[:u10] draw_scanline y: 51 - de: 1
[:u10] draw_scanline y: 52 - de: 1
[:u10] draw_scanline y: 53 - de: 1
[:u10] draw_scanline y: 54 - de: 1
[:u10] draw_scanline y: 55 - de: 1
[:u10] draw_scanline y: 56 - de: 1
[:u10] draw_scanline y: 57 - de: 1
[:u10] draw_scanline y: 58 - de: 1
[:u10] draw_scanline y: 59 - de: 1
[:u10] draw_scanline y: 60 - de: 1
[:u10] draw_scanline y: 61 - de: 1
[:u10] draw_scanline y: 62 - de: 1
[:u10] draw_scanline y: 63 - de: 1
[:u10] draw_scanline y: 64 - de: 1
[:u10] draw_scanline y: 65 - de: 1
[:u10] draw_scanline y: 66 - de: 1
[:u10] draw_scanline y: 67 - de: 1
[:u10] draw_scanline y: 68 - de: 1
[:u10] draw_scanline y: 69 - de: 1
[:u10] draw_scanline y: 70 - de: 1
[:u10] draw_scanline y: 71 - de: 1
[:u10] draw_scanline y: 72 - de: 1
[:u10] draw_scanline y: 73 - de: 1
[:u10] draw_scanline y: 74 - de: 1
[:u10] draw_scanline y: 75 - de: 1
[:u10] draw_scanline y: 76 - de: 1
[:u10] draw_scanline y: 77 - de: 1
[:u10] draw_scanline y: 78 - de: 1
[:u10] draw_scanline y: 79 - de: 1
[:u10] draw_scanline y: 80 - de: 1
[:u10] draw_scanline y: 81 - de: 1
[:u10] draw_scanline y: 82 - de: 1
[:u10] draw_scanline y: 83 - de: 1
[:u10] draw_scanline y: 84 - de: 1
[:u10] draw_scanline y: 85 - de: 1
[:u10] draw_scanline y: 86 - de: 1
[:u10] draw_scanline y: 87 - de: 1
[:u10] draw_scanline y: 88 - de: 1
[:u10] draw_scanline y: 89 - de: 1
[:u10] draw_scanline y: 90 - de: 1
[:u10] draw_scanline y: 91 - de: 1
[:u10] draw_scanline y: 92 - de: 1
[:u10] draw_scanline y: 93 - de: 1
[:u10] draw_scanline y: 94 - de: 1
[:u10] draw_scanline y: 95 - de: 1
[:u10] draw_scanline y: 96 - de: 1
[:u10] draw_scanline y: 97 - de: 1
[:u10] draw_scanline y: 98 - de: 1
[:u10] draw_scanline y: 99 - de: 1
[:u10] draw_scanline y: 100 - de: 1
[:u10] draw_scanline y: 101 - de: 1
[:u10] draw_scanline y: 102 - de: 1
[:u10] draw_scanline y: 103 - de: 1
[:u10] draw_scanline y: 104 - de: 1
[:u10] draw_scanline y: 105 - de: 1
[:u10] draw_scanline y: 106 - de: 1
[:u10] draw_scanline y: 107 - de: 1
[:u10] draw_scanline y: 108 - de: 1
[:u10] draw_scanline y: 109 - de: 1
[:u10] draw_scanline y: 110 - de: 1
[:u10] draw_scanline y: 111 - de: 1
[:u10] draw_scanline y: 112 - de: 1
[:u10] draw_scanline y: 113 - de: 1
[:u10] draw_scanline y: 114 - de: 1
[:u10] draw_scanline y: 115 - de: 1
[:u10] draw_scanline y: 116 - de: 1
[:u10] draw_scanline y: 117 - de: 1
[:u10] draw_scanline y: 118 - de: 1
[:u10] draw_scanline y: 119 - de: 1
[:u10] draw_scanline y: 120 - de: 1
[:u10] draw_scanline y: 121 - de: 1
[:u10] draw_scanline y: 122 - de: 1
[:u10] draw_scanline y: 123 - de: 1
[:u10] draw_scanline y: 124 - de: 1
[:u10] draw_scanline y: 125 - de: 1
[:u10] draw_scanline y: 126 - de: 1
[:u10] draw_scanline y: 127 - de: 1
[:u10] draw_scanline y: 128 - de: 1
[:u10] draw_scanline y: 129 - de: 1
[:u10] draw_scanline y: 130 - de: 1
[:u10] draw_scanline y: 131 - de: 1
[:u10] draw_scanline y: 132 - de: 1
[:u10] draw_scanline y: 133 - de: 1
[:u10] draw_scanline y: 134 - de: 1
[:u10] draw_scanline y: 135 - de: 1
[:u10] draw_scanline y: 136 - de: 1
[:u10] draw_scanline y: 137 - de: 1
[:u10] draw_scanline y: 138 - de: 1
[:u10] draw_scanline y: 139 - de: 1
[:u10] draw_scanline y: 140 - de: 1
[:u10] draw_scanline y: 141 - de: 1
[:u10] draw_scanline y: 142 - de: 1
[:u10] draw_scanline y: 143 - de: 1
[:u10] draw_scanline y: 144 - de: 1
[:u10] draw_scanline y: 145 - de: 1
[:u10] draw_scanline y: 146 - de: 1
[:u10] draw_scanline y: 147 - de: 1
[:u10] draw_scanline y: 148 - de: 1
[:u10] draw_scanline y: 149 - de: 1
[:u10] draw_scanline y: 150 - de: 1
[:u10] draw_scanline y: 151 - de: 1
[:u10] draw_scanline y: 152 - de: 1
[:u10] draw_scanline y: 153 - de: 1
[:u10] draw_scanline y: 154 - de: 1
[:u10] draw_scanline y: 155 - de: 1
[:u10] draw_scanline y: 156 - de: 1
[:u10] draw_scanline y: 157 - de: 1
[:u10] draw_scanline y: 158 - de: 1
[:u10] draw_scanline y: 159 - de: 1
[:u10] draw_scanline y: 160 - de: 1
[:u10] draw_scanline y: 161 - de: 1
[:u10] draw_scanline y: 162 - de: 1
[:u10] draw_scanline y: 163 - de: 1
[:u10] draw_scanline y: 164 - de: 1
[:u10] draw_scanline y: 165 - de: 1
[:u10] draw_scanline y: 166 - de: 1
[:u10] draw_scanline y: 167 - de: 1
[:u10] draw_scanline y: 168 - de: 1
[:u10] draw_scanline y: 169 - de: 1
[:u10] draw_scanline y: 170 - de: 1
[:u10] draw_scanline y: 171 - de: 1
[:u10] draw_scanline y: 172 - de: 1
[:u10] draw_scanline y: 173 - de: 1
[:u10] draw_scanline y: 174 - de: 1
[:u10] draw_scanline y: 175 - de: 1
[:u10] draw_scanline y: 176 - de: 1
[:u10] draw_scanline y: 177 - de: 1
[:u10] draw_scanline y: 178 - de: 1
[:u10] draw_scanline y: 179 - de: 1
[:u10] draw_scanline y: 180 - de: 1
[:u10] draw_scanline y: 181 - de: 1
[:u10] draw_scanline y: 182 - de: 1
[:u10] draw_scanline y: 183 - de: 1
[:u10] draw_scanline y: 184 - de: 1
[:u10] draw_scanline y: 185 - de: 1
[:u10] draw_scanline y: 186 - de: 1
[:u10] draw_scanline y: 187 - de: 1
[:u10] draw_scanline y: 188 - de: 1
[:u10] draw_scanline y: 189 - de: 1
[:u10] draw_scanline y: 190 - de: 1
[:u10] draw_scanline y: 191 - de: 1
[:u10] draw_scanline y: 192 - de: 1
[:u10] draw_scanline y: 193 - de: 1
[:u10] draw_scanline y: 194 - de: 1
[:u10] draw_scanline y: 195 - de: 1
[:u10] draw_scanline y: 196 - de: 1
[:u10] draw_scanline y: 197 - de: 1
[:u10] draw_scanline y: 198 - de: 1
[:u10] draw_scanline y: 199 - de: 1
[:u10] draw_scanline y: 200 - de: 0
[:u10] draw_scanline y: 201 - de: 0
[:u10] draw_scanline y: 202 - de: 0
[:u10] draw_scanline y: 203 - de: 0
[:u10] draw_scanline y: 204 - de: 0
[:u10] draw_scanline y: 205 - de: 0
[:u10] draw_scanline y: 206 - de: 0
[:u10] draw_scanline y: 207 - de: 0
[:u10] draw_scanline y: 208 - de: 0
[:u10] draw_scanline y: 209 - de: 0
[:u10] draw_scanline y: 210 - de: 0
[:u10] draw_scanline y: 211 - de: 0
[:u10] draw_scanline y: 212 - de: 0
[:u10] draw_scanline y: 213 - de: 0
[:u10] draw_scanline y: 214 - de: 0
[:u10] draw_scanline y: 215 - de: 0
[:u10] draw_scanline y: 216 - de: 0
[:u10] draw_scanline y: 217 - de: 0
[:u10] draw_scanline y: 218 - de: 0
[:u10] draw_scanline y: 219 - de: 0
[:u10] draw_scanline y: 220 - de: 0
[:u10] draw_scanline y: 221 - de: 0
[:u10] draw_scanline y: 222 - de: 0
[:u10] draw_scanline y: 223 - de: 0
[:u10] draw_scanline y: 224 - de: 0
[:u10] draw_scanline y: 225 - de: 0
[:u10] draw_scanline y: 226 - de: 0
[:u10] draw_scanline y: 227 - de: 0
[:u10] draw_scanline y: 228 - de: 0
[:u10] draw_scanline y: 229 - de: 0
[:u10] draw_scanline y: 230 - de: 0
[:u10] draw_scanline y: 231 - de: 0
[:u10] draw_scanline y: 232 - de: 0
[:u10] draw_scanline y: 233 - de: 0
[:u10] draw_scanline y: 234 - de: 0
[:u10] draw_scanline y: 235 - de: 0
[:u10] draw_scanline y: 236 - de: 0
[:u10] draw_scanline y: 237 - de: 0
[:u10] draw_scanline y: 238 - de: 0
[:u10] draw_scanline y: 239 - de: 0
[:u10] draw_scanline y: 240 - de: 0
[:u10] draw_scanline y: 241 - de: 0
[:u10] draw_scanline y: 242 - de: 0
[:u10] draw_scanline y: 243 - de: 0
[:u10] draw_scanline y: 244 - de: 0
[:u10] draw_scanline y: 245 - de: 0
[:u10] draw_scanline y: 246 - de: 0
[:u10] draw_scanline y: 247 - de: 0
[:u10] draw_scanline y: 248 - de: 0
[:u10] draw_scanline y: 249 - de: 0
[:u10] draw_scanline y: 250 - de: 0
[:u10] draw_scanline y: 251 - de: 0
[:u10] draw_scanline y: 252 - de: 0
[:u10] draw_scanline y: 253 - de: 0
[:u10] draw_scanline y: 254 - de: 0
[:u10] draw_scanline y: 255 - de: 0
[:u10] draw_scanline y: 256 - de: 0
[:u10] draw_scanline y: 257 - de: 0
[:u10] draw_scanline y: 258 - de: 0
[:u10] draw_scanline y: 259 - de: 0
[:u10] draw_scanline y: 260 - de: 0
[:u10] ':u13' (FE275):M6845 reg 0x11 = 0x00
```

angelosa:
You are not logging `screen().vpos()` here, which is basically what causes the crash.

Robbbert:
Some drivers have a line that says **if de == 0 return;**   The cbm2 callback doesn't have this check, so perhaps it always worked by luck before. Because DE is 0, there should be no attempt to write to the screen. Just to be sure, you should check the schematic and make sure that the DE line inhibits output of data (sync pulses will still go through).

mgarlanger:
> Some drivers have a line that says **if de == 0 return;** The cbm2 callback doesn't have this check, so perhaps it always worked by luck before. Because DE is 0, there should be no attempt to write to the screen. Just to be sure, you should check the schematic and make sure that the DE line inhibits output of data (sync pulses will still go through).

It should still be ok to write when de=0. This is the approach Vas had suggested to address fixing an issue when line 25 was turned off, basically write the background color when de is off - https://github.com/mamedev/mame/blob/master/src/mame/heathkit/tlb.cpp#L440-L466

mgarlanger:
> You are not logging `screen().vpos()` here, which is basically what causes the crash.

ok, I added screen().vpos() and I saw it consistently at 261 for all values of y, which should be ok, since the vtotal is 262 and if it is rendering the full screen at once, being at the max value seems fine. I did add `screen.set_video_attributes(VIDEO_UPDATE_SCANLINE);`  to the initialization of the system and saw the vpos matching the y value, since the scan line needs to be more accurate with that setting.

cuavas:
Did the issue with `b128` ever get fixed?

mgarlanger:
Sorry I dropped the ball on this one. But I wasn't able to reproduce it. I didn't any problems with the value of `screen().vpos()` as mentioned above. I'm not sure if I'll be able to address the issue without being able to reproduce it.

mgarlanger:
> Some drivers have a line that says **if de == 0 return;** The cbm2 callback doesn't have this check, so perhaps it always worked by luck before. Because DE is 0, there should be no attempt to write to the screen. Just to be sure, you should check the schematic and make sure that the DE line inhibits output of data (sync pulses will still go through).

Checked the schematics and it shows no connection for the DE line

Screenshot 2024-02-03 at 1 23 30 PM

PR #11758, opened 2023-11-19
seibu/dynduke.cpp: Use single-pass sprite rendering (take 2)

cam900:


cuavas:
This has been reverted as it results in missing graphics in dynduke, both in attract mode and in gameplay.  Did you test this at all?  Haven’t we already discussed that changes need to be tested *before* opening a pull request?

PR #11759, opened 2023-11-19
thunderx: simplified code of 052591 PMC emulation for clarity

jotego:
- removed unneeded loop counters
- removed unneeded multiplication/division operations
- matched variable names in both functions

Played the game for 20 minutes with no issues. Collision detection works the same.

### DIFF
--- a/src/mame/konami/thunderx.cpp
+++ b/src/mame/konami/thunderx.cpp
@@ -365,10 +365,7 @@ this is the data written to internal ram on startup:
 
 void thunderx_state::run_collisions( int s0, int e0, int s1, int e1, int cm, int hm )
 {
-	uint8_t* p0;
-	uint8_t* p1;
-
-	for ( p0 = &m_pmcram[s0]; p0 < &m_pmcram[e0]; p0 += 5)
+	for (uint8_t* p0 = &m_pmcram[s0]; p0 < &m_pmcram[e0]; p0 += 5)
 	{
 		int l0, r0, b0, t0;
 
@@ -382,7 +379,7 @@ void thunderx_state::run_collisions( int s0, int e0, int s1, int e1, int cm, int
 		t0 = p0[4] - p0[2];
 		b0 = p0[4] + p0[2];
 
-		for ( p1 = &m_pmcram[s1]; p1 < &m_pmcram[e1]; p1 += 5)
+		for (uint8_t* p1 = &m_pmcram[s1]; p1 < &m_pmcram[e1]; p1 += 5)
 		{
 			int l1,r1,b1,t1;
 
@@ -415,9 +412,7 @@ void thunderx_state::run_collisions( int s0, int e0, int s1, int e1, int cm, int
 
 void thunderx_state::calculate_collisions()
 {
-	int s0,e0;
-	int s1,e1;
-	int cm,hm;
+	int s0, s1;
 
 	// the data at 0x00 to 0x06 defines the operation
 	//
@@ -438,9 +433,8 @@ void thunderx_state::calculate_collisions()
 	// hit mask is 40 to set bit on object 0 and object 1
 	// hit mask is 20 to set bit on object 1 only
 
-	e0 = m_pmcram[0];
-	e0 = (e0 << 8) + m_pmcram[1];
-	e1 = m_pmcram[2];
+	const int e0 = (m_pmcram[0]<<8) | m_pmcram[1];
+	const int e1 = m_pmcram[2];
 
 	if (m_pmcram[5] < 16)
 	{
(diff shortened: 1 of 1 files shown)


src/mame/konami/thunderx.cpp
```diff
@@ -367,10 +367,8 @@ void thunderx_state::run_collisions( int s0, int e0, int s1, int e1, int cm, int
 {
 	uint8_t* p0;
 	uint8_t* p1;
-	int ii, jj;
 
-	p0 = &m_pmcram[16 + 5 * s0];
-	for (ii = s0; ii < e0; ii++, p0 += 5)
+	for ( p0 = &m_pmcram[s0]; p0 < &m_pmcram[e0]; p0 += 5)
```
cuavas, line 371: Please use “English” style spacing around parentheses (no space inside the parentheses) as the code was using previously.  It isn’t even consistent between the opening and closing parentheses now.

Also, you can reduce the scope of `p0` by declaring it in the `for` loop’s control expression.

src/mame/konami/thunderx.cpp
```diff
@@ -384,8 +382,7 @@ void thunderx_state::run_collisions( int s0, int e0, int s1, int e1, int cm, int
 		t0 = p0[4] - p0[2];
 		b0 = p0[4] + p0[2];
 
-		p1 = &m_pmcram[16 + 5 * s1];
-		for (jj = s1; jj < e1; jj++,p1 += 5)
+		for ( p1 = &m_pmcram[s1]; p1 < &m_pmcram[e1]; p1 += 5)
```
cuavas, line 385: Also, as above, use “English” spacing around parentheses, and you may as well reduce the scope of `p1` while you’re changing the code anyway.

src/mame/konami/thunderx.cpp
```diff
@@ -441,30 +438,28 @@ void thunderx_state::calculate_collisions()
 	// hit mask is 40 to set bit on object 0 and object 1
 	// hit mask is 20 to set bit on object 1 only
 
-	Y0 = m_pmcram[0];
-	Y0 = (Y0 << 8) + m_pmcram[1];
-	Y0 = (Y0 - 15) / 5;
-	Y1 = (m_pmcram[2] - 15) / 5;
+	e0 = m_pmcram[0];
+	e0 = (e0 << 8) + m_pmcram[1];
+	e1 = m_pmcram[2];
```
cuavas, line 443: If you’re changing this anyway, you may as well declare `e0` and `e1` at initialisation, initialise `e0` in a single statement, and make them both `const`.

src/mame/konami/thunderx.cpp
```diff
@@ -441,30 +438,28 @@ void thunderx_state::calculate_collisions()
...
-		X0 = (m_pmcram[5] - 16) / 5;
-		X1 = (m_pmcram[6] - 16) / 5;
+		s0 = m_pmcram[5];
+		s1 = m_pmcram[6];
 	}
 
-	CM = m_pmcram[3];
-	HM = m_pmcram[4];
+	cm = m_pmcram[3];
+	hm = m_pmcram[4];
```
cuavas, line 460: You may as well declare these at initialisation and make them `const` as well.

cuavas (changes requested):

jotego:
I have applied the changes requested

src/mame/konami/thunderx.cpp
```diff
@@ -441,30 +433,27 @@ void thunderx_state::calculate_collisions()
 	// hit mask is 40 to set bit on object 0 and object 1
 	// hit mask is 20 to set bit on object 1 only
 
-	Y0 = m_pmcram[0];
-	Y0 = (Y0 << 8) + m_pmcram[1];
-	Y0 = (Y0 - 15) / 5;
-	Y1 = (m_pmcram[2] - 15) / 5;
+	const int e0 = (m_pmcram[0]<<8) | m_pmcram[1];
+	const int e1 = m_pmcram[2];
```
cuavas, line 437: Note that this actually does change behaviour – previously it divided the Y1 value by 5 and then multiplied by 5 again later.  If the actual value was not a multiple of 5, the remainder would be truncated.  That’s no longer the case.  It now uses the actual value from RAM directly.

jotego: Yes, but data is accessed in groups of 5 later on. The new code makes more sense and is more parsimonius. I have not been able to find any problem in the gameplay after the change.

I do not know why the original author ended up writting those divisions and multiplications. They might be some leftover from all the guess work needed to emulate the chip or maybe he was trying to make sense of the _program_ bytes written by the CPU.

PR #11762, opened 2023-11-19
HD44780A00, KS0066_F00, SED1278_0B: fix errors in the ROM transcriptions from the datasheets. Add HD44780UA00 and HD44780UA02 variant devices. [Lord Nightmare]

Lord-Nightmare:
HD44780A00: fix errors in the ROM transcription from the 1985 datasheet. [Lord Nightmare]
Added HD44780UA00 variant device with ROM transcribed from the 1999 datasheet. [Lord Nightmare]
KS0066_F00: fix errors in the ROM transcription from the datasheet, this ends up being identical to the HD44780A00 [Lord Nightmare]
Added HD44780UA02 variant device with ROM transcribed from the 1999 datasheet. [Lord Nightmare]
SED1278_0B: fix errors in the ROM transcription from the datasheet. [Lord Nightmare]

### DIFF
--- a/src/devices/video/hd44780.cpp
+++ b/src/devices/video/hd44780.cpp
@@ -22,10 +22,12 @@
 //  DEVICE DEFINITIONS
 //**************************************************************************
 
-DEFINE_DEVICE_TYPE(HD44780,    hd44780_device,    "hd44780_a00", "Hitachi HD44780 A00 LCD Controller")
-DEFINE_DEVICE_TYPE(SED1278_0B, sed1278_0b_device, "sed1278_0b",  "Epson SED1278-0B LCD Controller") // packaged as either SED1278F0B or SED1278D0B
-DEFINE_DEVICE_TYPE(KS0066_F00, ks0066_f00_device, "ks0066_f00",  "Samsung KS0066 F00 LCD Controller")
-DEFINE_DEVICE_TYPE(KS0066_F05, ks0066_f05_device, "ks0066_f05",  "Samsung KS0066 F05 LCD Controller")
+DEFINE_DEVICE_TYPE(HD44780,      hd44780_device,      "hd44780_a00",  "Hitachi HD44780A00 LCD Controller")
+DEFINE_DEVICE_TYPE(HD44780U_A00, hd44780u_a00_device, "hd44780u_a00", "Hitachi HD44780UA00 LCD Controller")
+//DEFINE_DEVICE_TYPE(HD44780U_A02, hd44780u_a02_device, "hd44780u_a02", "Hitachi HD44780UA02 LCD Controller")
+DEFINE_DEVICE_TYPE(SED1278_0B,   sed1278_0b_device,   "sed1278_0b",   "Epson SED1278-0B LCD Controller") // packaged as either SED1278F0B or SED1278D0B
+DEFINE_DEVICE_TYPE(KS0066_F00,   ks0066_f00_device,   "ks0066_f00",   "Samsung KS0066 F00 LCD Controller")
+DEFINE_DEVICE_TYPE(KS0066_F05,   ks0066_f05_device,   "ks0066_f05",   "Samsung KS0066 F05 LCD Controller")
 
 
 //-------------------------------------------------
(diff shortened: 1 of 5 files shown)


happppp:
Could you make a decision on either removing A02 or having it in? Not a commented-out version of it.

Lord-Nightmare:
I think maybe I should just leave it out entirely. We don't (yet) have anything which uses it, but the rom is done and waiting once we do.

Lord-Nightmare:
Scratch that, I proved that the elektor/avrmax cc2schach driver does need the HD44780UA02, so I've fully added it. I also when doing a cursory check of the SED1278_0B ROM we use vs the datasheet spotted two incorrect characters, so I've fixed those as well.
(I also added a note about the weird, unusually slow clock that the Hitachi LM044L module, used by Sprow's LCD addon for the bbc micro, uses)

src/devices/video/hd44780.cpp
```diff
@@ -108,6 +147,8 @@ const tiny_rom_entry *hd44780_device::device_rom_region() const
 	switch (m_charset_type)
 	{
 		case CHARSET_HD44780_A00:   return ROM_NAME( hd44780_a00 );
+		case CHARSET_HD44780U_A00:  return ROM_NAME( hd44780u_a00 );
+		case CHARSET_HD44780U_A02:  return ROM_NAME( hd44780u_a02 );
 		case CHARSET_SED1278_0B:    return ROM_NAME( sed1278_0b );
 		case CHARSET_KS0066_F00:    return ROM_NAME( ks0066_f00 );
 		case CHARSET_KS0066_F05:    return ROM_NAME( ks0066_f05 );
```
cuavas, line 154: This is really gross.  Why aren’t you overriding `device_rom_region` in each device with a different ROM?

src/devices/video/hd44780.h
```diff
@@ -70,11 +70,13 @@ class hd44780_device : public device_t
...
 		CHARSET_SED1278_0B,
 		CHARSET_KS0066_F00,
 		CHARSET_KS0066_F05 /*,
-		CHARSET_HD44780_A01,
-		CHARSET_HD44780_A02,
+		CHARSET_HD44780U_A01,
+		CHARSET_SED1278_0A,
 		CHARSET_KS0066_F03,
 		CHARSET_KS0066_F04,
 		CHARSET_KS0066_F06,
```
cuavas, line 82: As far as I can tell, this is only used to implement `device_rom_region` the wrong way, and should go.

src/devices/video/hd44780.h
```diff
@@ -141,6 +143,24 @@ class hd44780_device : public device_t
...
+};
+
+// ======================> hd44780u_a02_device
+
+class hd44780u_a02_device :  public hd44780_device
+{
+public:
+	// construction/destruction
+	hd44780u_a02_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+};
```
cuavas, line 162: Shouldn’t at least some of these implement `parent_rom_device_type`?

happppp: These 2 new additions each have a unique char rom.

cuavas: Yeah, but one of the device seems to have an identical ROM, and making the whole group of devices use a single parent would reduce the overall file count in merged sets for people who are into that kind of thing.

Lord-Nightmare: The KS0066_F00 device already existed and already had its own ROM, which WAS different from the hd44780a00 ROM, but this turned out to be an error (the two are actually supposed to be identical). I changed it so (correctly) both devices now use the same ROM, but (to prevent user annoyance) left the two ROMs as separate files.

cuavas: But it *is* going to cause user annoyance because they’ll need to update their ks0066_f00.zip now, and it won’t be able to share the HD44780A00’s copy of that ROM because there’s no relationship between the devices.

src/devices/video/hd44780.cpp
Lord-Nightmare: I followed the same coding pattern/style that the device already had.

cuavas: Doesn’t change the fact that it’s completely nonsensical, and it should be pretty obvious there are no other devices doing this.

happppp: Isn't Sandro the correct person to direct that to, not LN?

cuavas: Well in a perfect world, we could go back and ensure it didn’t get to this state over a decade ago.  But the way it got into this state was that there was originally a single device type with the ROM selected after it was instantiated at machine configuration time.  You couldn’t set device BIOS options from machine configuration back then, so there wasn’t a proper way to select between multiple ROM choices for a single device type, and this is what we ended up with.  Of course, back then devices didn’t participate in `-validate`, `-romident`, etc. so hackery like that wasn’t as visible to end users.

Anyway, we’re carrying so many of these things that are in a bad state due to accidents of history that we really need to maintain a policy of, “If you encounter something nonsensical, don’t perpetuate it.”

PR #11763, opened 2023-11-19
nabupc/bus: Update builtin adapter to support local and remote loading

brijohn:
Adds several fixes and updates to the builtin nabu network adapter. 

Also refactors the adapter to support both loading files locally from an npz(zip) archive as well as loading the main nabu cycle over the network from cloud.nabu.ca.


src/devices/bus/nabupc/adapter.cpp
```diff
@@ -9,65 +9,246 @@
 #include "emu.h"
 #include "adapter.h"
 
+#include "client_http.hpp"
+#include "diserial.h"
 #include "emuopts.h"
+#include "hashing.h"
 #include "unzip.h"
```
cuavas, line 16: Please separate the src/emu headers from the src/lib/util headers, and sort the headers within each group.  The src/emu headers should come before the src/lib/util headers.

src/devices/bus/nabupc/adapter.cpp
```diff
@@ -352,45 +511,55 @@ void network_adapter::channel_request(uint8_t byte)
 	++m_substate;
 }
 
-void network_adapter::segment_request(uint8_t byte)
+void network_adapter_base::segment_request(uint8_t byte)
 {
+	static uint32_t segment_id = 0;
+
```
cuavas, line 517: A static function variable like this is completely unacceptable.  You should *know* we’re trying to get rid of the remaining static function variables.  Adding them is a regression.

cuavas (commented):
Besides adding a static function variable, I’m really not keen on adding a device that’s hard-coded to a specific web server somewhere on the Internet.

PR #11764, opened 2023-11-20
Revert "Revert "seibu/dynduke_v.cpp: Use single-pass sprite drawing. (#11758)""

cam900:
This PR includes fixing missing graphics, Tested and it might be working.


happppp:
Is this also broken? If no, why did you close it?

cuavas:
> Is this also broken? If no, why did you close it?

Because it was opened as a draft PR with no description, suggesting it still wasn't tested, I missed the one changed line from the previous iteration, and I'm still incredibly frustrated with the experience when this had gone through two iterations but was still obviously broken. @cam900 should know these kinds of things shouldn't be opened until they're ready.

happppp:
Ok, it's not broken, and after yesterday's merge (implying approval of source code), and only 1 line added to fix it, it can be merged

PR #11766, opened 2023-11-20
update libflac to 1.4.3

landfillbaby:
decided not to give up on #11733.  it compiles now at least


cuavas:
Thanks.  I’ve taken care of this.

PR #11767, opened 2023-11-20
nintendo/spacefb.cpp: Add a Spanish Space Firebird bootleg clone

clawgrip:
New WORKING clone
------------------------
Red Bird (bootleg of Space Firebird) [Recreativas.org]

Also fix indentation and comments capitalization.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -34405,6 +34405,7 @@ snes4sl
 snes4sln
 
 @source:nintendo/spacefb.cpp
+redbird                         // bootleg (Codematic)
 spacebrd                        // bootleg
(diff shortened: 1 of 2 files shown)


cuavas:
I believe this driver currently relies on samples to simulate the unemulated analog sound hardware.  Is this bootleg confirmed to use identical sound circuitry, or is it possibly different?  Bootlegs didn’t always reproduce analog circuitry identically.

clawgrip:
We don't have the sound board, that's why I've added a note and marked the sound ROM as BAD_DUMP, but from what the owner told us, it was a exact clone of the Nintendo PCB.

PR #11769, opened 2023-11-20
midway/williams.cpp: Add (not working) a Spanish Defender bootleg clone

clawgrip:
New NOT_WORKING clone
-----------------------------
Defender (Operamatic bootleg) [Recreativas.org]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -27552,6 +27552,7 @@ defenderb                       // (c) 1980
 defenderg                       // (c) 1980
 defenderj                       // (c) 1980
 defenderw                       // (c) 1980
+defenderom                      // bootleg (Operamatic)
 defenseb                        // bootleg
 defndjeu                        // bootleg
 galwars2                        // bootleg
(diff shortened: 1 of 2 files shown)


cuavas:
Do you have any useful information on why this isn’t working or how the hardware differs that can be added?  I’m mostly concerned for anyone who might want to try getting it working in the future.

clawgrip:
> Do you have any useful information on why this isn’t working or how the hardware differs that can be added? I’m mostly concerned for anyone who might want to try getting it working in the future.

It's exactly the same PCB as Petaco Defense.
We've finally repaired the PCB and behaves exactly as MAME, rebooting early on the rug pattern, so it's likely a bad ROM. We've redumped each ROM about ten times, getting the same results (and stable reads).

We'll try to determine at least which ones are bad and if it's worth adding it or not.

Thx!

clawgrip:
Ok, we've identified the bad ROMs. I'd say it's a set worth preserving and documenting, even with two bad dumps.

PR #11771, opened 2023-11-21
heathkit/tlb.cpp: Fix on-screen menus for superset tlbc slot option

mgarlanger:
Fixes the on-screen menu (ctrl-ESC brings it up) for the superset ROM/ tlbc slot option for the H19/H89.

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -20,7 +20,6 @@
       - In 49/50 row mode, character descenders are cut off.
       - Screen saver does not disable the screen
     - With superset slot option
-      - Screen menus not working properly
       - Screensaver freezes the screen instead of blanking the screen	
 
 ****************************************************************************/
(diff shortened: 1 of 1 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -1117,6 +1116,9 @@ void heath_superset_tlb_device::device_add_mconfig(machine_config &config)
 	m_maincpu->set_addrmap(AS_PROGRAM, &heath_superset_tlb_device::mem_map);
 	m_maincpu->set_addrmap(AS_IO, &heath_superset_tlb_device::io_map);
 
+	// per line updates are needed for onscreen menu to display properly
+	m_screen->scanline().set([](int scanline) {});
+
```
cuavas, line 1121: Wouldn’t it make more sense to do `m_screen->set_video_attributes(VIDEO_UPDATE_SCANLINE);` if what you’re trying to do is to get the screen to do partial updates for each scan line?

PR #11773, opened 2023-11-21
tvochken.xml: Add TV Ocha-Ken barcode card parts

qufb:
Added software list with barcode card artwork.

### DIFF
--- a/hash/tvochken.xml
+++ b/hash/tvochken.xml
@@ -1,0 +1,365 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="tvochken" description="TV Ocha-Ken barcode cards">
+
+	<software name="tvochken" supported="no">
+		<description>TV to Ocha-Card: Ocha-Ken 'Ho' to Seikatsu</description>
+		<year>2005</year>
+		<publisher>Sega Toys</publisher>
+		<info name="alt_title" value="テレビとお茶札 お茶犬『ほっ』と生活"/>
+		<part name="card1" interface="sega_beena_cart">
+			<feature name="part_id" value="Card 1"/>
+			<feature name="barcode" value="0x900a"/>
+			<dataarea name="card1" size="0x4f1749">
+				<rom name="0 - 0001.png" size="0x4f1749" crc="c5432bb7" sha1="375fb553245e88d37977f6a831ede72cf7c94eb8"/>
+			</dataarea>
+		</part>
+		<part name="card2" interface="sega_beena_cart">
+			<feature name="part_id" value="Card 2"/>
+			<feature name="barcode" value="0xa05a"/>
+			<dataarea name="card2" size="0x4df6f0">
+				<rom name="0 - 0002.png" size="0x4df6f0" crc="6c83c38a" sha1="616a796615423e1eaa05a2637cd606d5816c7f01"/>
+			</dataarea>
+		</part>
+		<part name="card3" interface="sega_beena_cart">
+			<feature name="part_id" value="Card 3"/>
+			<feature name="barcode" value="0xb0aa"/>
+			<dataarea name="card3" size="0x4f220b">
+				<rom name="0 - 0003.png" size="0x4f220b" crc="a14f9d50" sha1="a3bce401fe3c70bef046addb4cbc31c5e0f04f0f"/>
+			</dataarea>
+		</part>
+		<part name="card4" interface="sega_beena_cart">
+			<feature name="part_id" value="Card 4"/>
+			<feature name="barcode" value="0x90ca"/>
+			<dataarea name="card4" size="0x4f1602">
+				<rom name="0 - 0004.png" size="0x4f1602" crc="1c124ccf" sha1="840a3ee78ae0995b9b9fa9428da18d0ecd293f88"/>
+			</dataarea>
+		</part>
+		<part name="card5" interface="sega_beena_cart">
+			<feature name="part_id" value="Card 5"/>
+			<feature name="barcode" value="0x910a"/>
+			<dataarea name="card5" size="0x505a3a">
+				<rom name="0 - 0005.png" size="0x505a3a" crc="c7190bb3" sha1="e4d1307971c053fc7ad98b6b8bea19a639c88084"/>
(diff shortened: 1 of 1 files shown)


hash/tvochken.xml
```diff
@@ -0,0 +1,170 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="tvochken" description="TV Ocha-Ken ROM and barcode cards">
+
+	<software name="tvochken" supported="yes">
+		<description>TV to Ocha-Satsu Ocha-Ken Hotsu to Seikatsu</description>
```
qufb, line 9: Not very confident on this description, would appreciate if someone else could confirm it.

src/mame/layout/tvochken.lay
```diff
@@ -0,0 +1,73 @@
...
+		<rect state="49"><color alpha="0" /></rect>
+	</element>
+
+	<view name="Default View">
+		<screen index="0">
+			<bounds x="0" y="0" width="4" height="3" />
+		</screen>
+
+		<element id="card" ref="card">
+			<bounds x="0" y="3.1" width="4" height="2.66666" />
```
qufb, line 18: Actual ratio is 3/2, is there a more accurate way of writing this here?

happppp:
For a console's internal ROM(s), the preferable place is the driver file.

src/mame/layout/tvochken.lay
happppp: Looks fine to me? You could add a comment if you want
```<!-- card ratio is 3/2 -->```

qufb:
> For a console's internal ROM(s), the preferable place is the driver file.

Makes sense, reverted this change.

Just added a comment in case someone finds it odd to use a cart loader without a cart ROM.

happppp:
Here's something similar. It's not emulated yet, but if it was then the MCU ROM would be in the driver file, and the barcode data somewhere else.
https://www.youtube.com/watch?v=aqWKku8RYdY

hash/tvochken.xml
987123879113: テレビとお茶札 お茶犬『ほっ』と生活

Not sure how to properly romanize weird cases like this but it's "ほっ" not "ほつ" so at the very least "Hotsu" is not right. It'd probably be just "Ho" here.

stonedDiscord: On the box art the 札 has カード above it as furigana, so at least that character should probably be romanized as "card".
91x02IfOR+L _AC_SL1500_~2

einstein95:
> Just added a comment in case someone finds it odd to use a cart loader without a cart ROM.

I could swear there was a software list of barcodes for one device.

hash/tvochken.xml
stonedDiscord: according to https://goiryoku.com/onomatopoeia/holtu/ the 『ほっ』 seems to be the Onomatopoeia for a sigh of relief.
In this music video https://www.uta-net.com/movie/225031/ it is also pronounced as Ho.
The 犬 kanji can be read as Inu or Ken and I think in this case it is Inu, because it's a tea dog (look at the packaging).
My suggestion is "TV to Ocha-Card: Ocha-Inu 'Ho' to Seikatsu" for the romanization and "TV and tea card: Life with tea dog 'Ho'"
The machine translation agrees if you put it in like this "テレビとお茶カード: お茶犬『ほっ』と生活"

987123879113: Ocha-ken is an established character name and anything else would be wrong.

stonedDiscord: whoops
https://en.wikipedia.org/wiki/Ocha-Ken

wd40yasu:
+<softwarelist name="tvochken" description="TV Ocha-Ken ROM and barcode cards">
+
+	<software name="tvochken" supported="yes">
+		<description>TV to Ocha-Satsu Ocha-Ken Hotsu to Seikatsu</description>

https://www.amazon.co.jp/%E3%82%BB%E3%82%AC%E3%83%88%E3%82%A4%E3%82%BA-SEGA-TOYS-%E3%83%86%E3%83%AC%E3%83%93%E3%81%A8%E3%81%8A%E8%8C%B6%E6%9C%AD-%E3%81%8A%E8%8C%B6%E7%8A%AC%E3%80%8C%E3%81%BB%E3%81%A3%E3%80%8D%E3%81%A8%E7%94%9F%E6%B4%BB/dp/B0009J592Q

There is furigana 'カード(card)' on kanji '札', so the title should be
"TV to Ocha-Card Ocha-Ken Hotto Seikatsu"

cuavas:
@qufb can you fill me in on how this system works and point me at some pictures of it?  I don’t have a clear picture of it in my head, and this will affect how the stuff will get hooked up.

Some initial questions:
* Did the system come with a set of cards, or were they sold individually?
* What do the cards look like?  Can you post low-resolution versions of a couple here?
* Are the cards logically a single “deck”, or are they more standalone?

qufb:
You can find some screenshots and photos here, including of the cards: https://game.watch.impress.co.jp/docs/20050811/toy191.htm

These cards were sold with the system, and functionally are more standalone. A player can pick any of them without any particular order.

For layout interaction, it is sufficient to just cycle through cards. The currently shown card would be scanned with the corresponding input.

hash/tvochken.xml
qufb: Thanks everyone, will use the last suggestion.

cuavas:
OK, I need to add a couple more script bindings for this.  Can you please:
* Back out the changes to the C++ code and layout, I’ll deal with that part later.
* Change the software list to have a single item with one part per card.
* You can set the `name` for each part to `card1`, `card2`, etc.
* Use a `<feature name="part_id" ... />` to give the display name for each card.
* Use another `<feature ... />` to store the barcode data for each card.

I can take care of the code changes to support it later.

qufb:
Updated. Are you considering some separate interface to load these parts? I left them with the same cart loader.

cuavas:
> Updated. Are you considering some separate interface to load these parts? I left them with the same cart loader.

Yeah, it’s going to need its own implementation of `device_image_interface`.  But if the software list gets in now, it’s easy enough for me to pick up the work to get it working when I get a chance.

PR #11776, opened 2023-11-23
HD44780: Implement timing (busy/status and blink interval) based on the input clock. Add notes about device variants. [Lord Nightmare]

Lord-Nightmare:
HD44780: Implement timing (busy/status and blink interval) based on the input clock. Add notes about device variants. [Lord Nightmare]


cuavas (commented):
* Isn’t the whole point of `m_busy_factor` to allow adjustment for different clock frequencies without using the configured clock frequency?  Shouldn’t it be unnecessary if the configured clock frequency is used to calculate how long the device will be busy for?
* Should you be quantising the busy time to intervals of the clock frequency?
* Why aren’t you using `attotime::from_ticks`?

cuavas:
Please don’t force push unnecessarily when changing code for pull requests.  It just makes it unnecessarily hard for reviewers because it’s harder to see what changed each time.

Lord-Nightmare:
> Please don’t force push unnecessarily when changing code for pull requests. It just makes it unnecessarily hard for reviewers because it’s harder to see what changed each time.

This force-push was necessary because a change upstream made it un-merge-able. Absolutely no functional changes (or really any changes at all, other than fixing the merge conflict because git got confused) were made.

angelosa:
And that's doable by fixing thru `git merge <head_branch>` and fix the conflicts there (which will also report a cleaner commit in case you actively did some change there). Forcing is almost never the solution, it will break PR view.

src/devices/video/hd44780.cpp
```diff
@@ -254,12 +360,10 @@ TIMER_CALLBACK_MEMBER(hd44780_device::blink_tick)
...
 //**************************************************************************
 
-void hd44780_device::set_busy_flag(uint16_t usec)
+void hd44780_device::set_busy_flag(uint16_t cycles)
 {
 	m_busy_flag = true;
-
-	usec = float(usec) * m_busy_factor + 0.5;
-	m_busy_timer->adjust(attotime::from_usec(usec));
+	m_busy_timer->adjust(attotime::from_hz(double(clock() / cycles) * m_busy_factor));
```
happppp, line 366: This will invert the result with busy_factor?
eg. busy factor 0.5, will make the busy flag duration twice longer, or 0.0 would mean 0Hz which is attotime::never.

src/devices/video/hd44780.cpp
```diff
@@ -5,9 +5,113 @@
...
+       and if possible, remove m_busy_factor
+      The internal oscillator is known to be based on the value of the resistor
+       Rf, presumably (based on the HD44780U datasheet diagrams) with an
+       equation close to this:
+       For 5V power, Frequency in Hz = 1 / (2 * PI * Rf * 6.5pF)
+       For 3V power, Frequency in Hz = 1 / (2 * PI * Rf * 7.86pF)
+      The vast majority of devices use the internal oscillator with an Rf
+       resistor of 91KOhms, for ~270KHz clock speed.
+      However, a few Hitachi-made LCD modules use an Rf resistor of
+       200KOhms instead, for a ~122Khz clock speed, and it isn't clear
```
happppp, line 19: It's with a small k, big H, so kOhm and kHz

src/devices/video/hd44780.cpp
Lord-Nightmare: This should be fixed now, although I'm going to see if I can completely remove the whole busy factor system in favor of just adjusting the clock for the few drivers that use it, in the future

src/devices/video/hd44780.cpp
Lord-Nightmare: This should be fixed now.

Lord-Nightmare:
> * Isn’t the whole point of `m_busy_factor` to allow adjustment for different clock frequencies without using the configured clock frequency?  Shouldn’t it be unnecessary if the configured clock frequency is used to calculate how long the device will be busy for?
> 
>     * Should you be quantising the busy time to intervals of the clock frequency?
> 
>     * Why aren’t you using `attotime::from_ticks`?

I'm going to see if I can safely completely remove the busy factor from the few drivers that use it and ditch it entirely.

I've switched to attotime::from_ticks.

src/devices/video/hd44780.cpp
Lord-Nightmare: Nothing does, no.
And if I can make busy_factor go away completely, this will be a non-issue.

cuavas:
Why didn’t you wait for @Lord-Nightmare to try getting rid of the fudge factor?  Also, please look at the commit message of the commit.  Can people **please** think about release notes?

happppp:
I told LN on discord this good for a 1st step.
I removed the meaningless cleanup messages when I squashed the commits.

cuavas:
> I removed the meaningless cleanup messages when I squashed the commits.

Look at the squashed commit – the message appears twice.

happppp:
Ah I see now, yes the PR title is the 1st line of the commit msg.

PR #11777, opened 2023-11-23
misc/playcenter.cpp: Add, as not working, a new Playcenter machine

clawgrip:
New NOT_WORKING machine
Playcenter Evolution III (v14.0, 'Epox' hardware) [Cuclis, Oskarito, Ángel]

### DIFF
--- a/src/mame/misc/playcenter.cpp
+++ b/src/mame/misc/playcenter.cpp
@@ -4,22 +4,24 @@
 
     Skeleton driver for "PlayCenter" PC-based touch games
 
-    "PlayCenter Champions Tournament":
-    The machine was sold with three different hardware configurations (internally named "Epox",
-    "K6" and "Celeron"), having each one its own "official" recovery image (Norton Ghost images).
+All PlayCenter machines:
+    The machines were sold with three different hardware configurations (internally named "Epox",
+    "K6" and "Celeron"), each one with its own "official" recovery images (usually Norton Ghost).
       "Epox" version:
          -EPoX EP-MVP4F motherboard (Socket 7 with VIA Apollo MVP4 AGP, built-in Trident AGP,
           VT82C686A, built-in SoundBlaster Pro with AC'97 codec support).
-         -Unknown RAM size, AMD K6 processor (unknown speed).
-         -PCI Ethernet card (unknown chipset).
+         -128MB RAM PC133, AMD K6-2 processor (K6-2/500(100*5)).
+         -PCI Ethernet card (RTL8029AS chipset, PCI).
          -Trident Blade3D/MVP4 AGP video.
-         -Elo Touch CTR-231000 touch screen (87C51-based, undumped).
+         -56K Modem (S56MR, HAMR5603 + Si3014-KS).
+         -Elo Touch CTR-231000 touch screen (87C51-based, undumped) or custom touch I/O PCB named 
+          "Touch Presas" with unknown (and undumped) MCU.
       "K6" version:
          -Unknown AMD K6 CPU based hardware.
       "Celeron" version:
          -Unknown Intel Celeron CPU based hardware.
 
-    The security scheme is quite simple:
+About "PlayCenter Champions Tournament" (may not apply to other variants) security scheme:
     Each minigame is a VisualBasic executable that receives three numbers when
     invoked, the hard disk serial number (readed through the Windows enumerator
     file), the Ethernet MAC address, and a third one, and concatenates them
@@ -33,7 +35,7 @@
     is unique for each single machine (and these hardware parts are not
     replaceble by the user / operator).
 
-TODO (about the "Epox" version, the only one there's a BIOS dump for):
+TODO (about the "Epox" version of "PlayCenter Champions Tournament"):
 - In pcipc it shows a '1' logo then on first boot it tries to install basically
   everything it can possibly install from scratch. Once done it tries to boot the
   main program but it either do one of the following things:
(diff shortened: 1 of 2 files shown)


src/mame/misc/playcenter.cpp
```diff
@@ -86,21 +88,29 @@ INPUT_PORTS_END
...
 	// ...
 }
 
+// 'Epox' version. Windows 98 SE.
+ROM_START(plycntre3)
+	ROM_REGION32_LE(0x40000, "bios", 0) // BIOS date: 03/13/2001
+	ROM_LOAD("vp4f1313.bin", 0x00000, 0x40000, CRC(bd4b155f) SHA1(3eafe71e89bf84b72a42e933187676fe08db0492))
+
+	DISK_REGION( "ide:0:hdd" )
+	DISK_IMAGE("playcenter_e_14_0_t", 0, SHA1(b7c7fce1522e64fc5132f223d8a73555c24002a9)) // From an operator, may contain user data. Dump contains a raw image for a (c) 1998 Trident video card (C:\videorom.bin)
```
angelosa, line 104: You can generally detect what video card is by digging the `PCIR` signature inside the binary and check next 4 bytes byteswapped. i.e. `PCIR\de\10\00\01` is a nVidia `10de` GeForce 256 SDR `0100`.
Being likely a Trident Blade MVP4 should be `PCIR\23\10\80\98`

src/mame/misc/playcenter.cpp
```diff
@@ -126,6 +136,7 @@ ROM_END
 } // Anonymous namespace
 
 //   YEAR  NAME          PARENT       MACHINE     INPUT       CLASS             INIT        ROT   COMPANY                             FULLNAME                                                      FLAGS
+GAME(2004, plycntre3,    0,           playcenter, playcenter, playcenter_state, empty_init, ROT0, "Recreativos Presas / Undergaming", "Playcenter Evolution III (v14.0, 'Epox' hardware)",          MACHINE_IS_SKELETON) // E.14.0.TCT
 GAME(2000, plycntrchtr,  0,           playcenter, playcenter, playcenter_state, empty_init, ROT0, "Recreativos Presas / Undergaming", "PlayCenter Champions Tournament (v9.3, 'Epox' hardware)",    MACHINE_IS_SKELETON)
```
angelosa, line 140: Please order by year.

src/mame/misc/playcenter.cpp
clawgrip: Done

src/mame/misc/playcenter.cpp
cuavas, line 1: It would make sense to order the ROM definitions the same way as the `GAME` macros.

angelosa: At this point it isn't even confirmed if they uses the same base motherboard so the order is okay with year releases for the time being.

angelosa: Or, put in another way, if the cards/motherboard etc. are identical but the CPU diverges then it's a matter of fixing those in the machine configs instead (something we are miles off anyway).

clawgrip: placa
arranque bios y procesador y memoria

angelosa: Yeah, this is for Epox, what about the others?

clawgrip: It’s a matter of time, we’ll end up finding a working machine with a Celeron or K6-1

src/mame/misc/playcenter.cpp
```diff
@@ -123,9 +124,19 @@ ROM_START(plycntrchtrc)
 	DISK_IMAGE("playcenter_celeron_9.3_tournament", 0, SHA1(eb98f6af20a98d4bf6dc7311035431467ad56605))
 ROM_END
 
+// 'Epox' version. Windows 98 SE.
+ROM_START(plycntre3)
+	ROM_REGION32_LE(0x40000, "bios", 0) // BIOS date: 03/13/2001
+	ROM_LOAD("vp4f1313.bin", 0x00000, 0x40000, CRC(bd4b155f) SHA1(3eafe71e89bf84b72a42e933187676fe08db0492))
+
+	DISK_REGION( "ide:0:hdd" )
+	DISK_IMAGE("playcenter_e_14_0_t", 0, SHA1(b7c7fce1522e64fc5132f223d8a73555c24002a9)) // From an operator, may contain user data. Contains a BIOS for a Trident card (C:\videorom.bin), probably a Trident Blade 3D (PCIR 1023:9880)
```
angelosa, line 133: It is a Trident Blade 3d, there's no need of the probably. :)
https://admin.pci-ids.ucw.cz/read/PC/1023/9880

clawgrip: Ok!

clawgrip:
@angelosa Is anything left to do about this game?
We has to return back the machine to its owner, but we checked it and the HDD image is good, the BIOS is the right one and the hardware info is right.

Thanks

angelosa:
You can try booting the HDD with `pcipc` (which iirc panics because it needs a Pentium 3, but as long as it _boots_ it's fair enough), otherwise I don't think it's necessary to keep the machine.

angelosa (approved):

cuavas:
Are these people trying to sneak advertising into the credits?  “Valdegamers” appears to be an online merchandise store.  It seems an increasing number of people from the Iberian Peninsula seem to think MAME’s release notes are a platform for advertising their businesses.  Given the added systems appear on the MAMEdev web site, this looks like a form of SEO keyword stuffing.

clawgrip:
> Are these people trying to sneak advertising into the credits? “Valdegamers” appears to be an online merchandise store. It seems an increasing number of people from the Iberian Peninsula seem to think MAME’s release notes are a platform for advertising their businesses. Given the added systems appear on the MAMEdev web site, this looks like a form of SEO keyword stuffing.

I've asked them for alternative credits without commercial references. Thanks!

clawgrip:
> Are these people trying to sneak advertising into the credits? “Valdegamers” appears to be an online merchandise store. It seems an increasing number of people from the Iberian Peninsula seem to think MAME’s release notes are a platform for advertising their businesses. Given the added systems appear on the MAMEdev web site, this looks like a form of SEO keyword stuffing.

Ok! Removed the commercial references from the credits!!

PR #11778, opened 2023-11-23
heathkit/tlb.cpp: Use memory bank for imaginator I-100 ROM

mgarlanger:
Change from memory_view to memory bank for the Imaginator I-100 ROM.

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -1483,7 +1483,7 @@ ioport_constructor heath_gp19_tlb_device::device_input_ports() const
  */
 heath_imaginator_tlb_device::heath_imaginator_tlb_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
 	heath_tlb_device(mconfig, HEATH_IMAGINATOR, tag, owner, clock),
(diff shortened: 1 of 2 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -1581,11 +1574,7 @@ void heath_imaginator_tlb_device::mem_map(address_map &map)
 
 void heath_imaginator_tlb_device::tap_6000h()
 {
-	if (m_mem_map != 0)
-	{
-		m_mem_map = 0;
-		m_mem_view.select(m_mem_map);
-	}
+	m_mem_bank->set_entry(0);
 }
```
cuavas, line 1578: As far as I can tell, the only way for it to flip back to entry 1 is on reset, so you could make it uninstall the tap after the first time it’s hit to make it a bit more efficient.

mgarlanger: @cuavas How do you uninstall a tap? I looked for a method that would do it and didn't see anything.

cuavas: With `.remove()`: 706cc60b159b337545748c0abe0545b447037c0a

You can find a fair few uses of read/write taps around the source tree if you grep for them.

PR #11779, opened 2023-11-25
new working machines and software list items (Casio CZ-1 and MZ-1)

devinacker:
This adds the Casio CZ-1 synth and unreleased "MZ-1" rackmount version, both of which are supported by the same ROM set, and two first-party ROM cartridges.

New working machine
--------
Casio CZ-1 [Mastropiero, Devin Acker]

New working clone
--------
Casio MZ-1 (prototype) [Mastropiero, Devin Acker]

--------
cz1_cart.xml:
Orchestra (RC-10) [Edward d-tech]
Organ / Synth. Sound / Sound Effect (RC-30) [Edward d-tech]
Piano / Guitar / Percussion (RC-20) [unknown]


src/mame/casio/ra3.cpp
```diff
@@ -13,67 +13,102 @@
...
+	if (size < 0x1000 || (size & (size - 1)))
 		return std::make_pair(image_error::INVALIDLENGTH, std::string());
 
-	fseek(0, SEEK_SET);
-	const size_t ret = fread(m_ram.data(), size);
-	if (ret != size)
-		return std::make_pair(std::errc::io_error, "Error reading file");
+	// allow loading larger than the maximum size (e.g. for CZ-1 ROM carts that use oversized ROMs)
+	m_size = std::min(size, m_max_size);
+	m_mask = m_size - 1;
```
cuavas, line 61: Should this allow loading oversized RAM cartridges, or only ROM cartridges from the software list?

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+		m_led_env(*this, "led_env%u", 0U),
+		m_led_bank(*this, "led_bank%u", 0U),
+		m_led_tone(*this, "led_tone%u", 0U)
+	{ }
+
+	void mz1(machine_config &config);
+	void cz1(machine_config &config);
+
+	void cz1_palette(palette_device &palette) const;
+	HD44780_PIXEL_UPDATE(lcd_pixel_update);
```
cuavas, line 82: Why are these public?  I don’t see them used anywhere that’s necessary.

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+		m_led_bank[i] = !BIT(data, 0) && (BIT(data, 5, 3) == i);
+}
+
+/**************************************************************************/
+void cz1_state::main_to_sub_0_w(u8 data)
+{
+	machine().scheduler().synchronize();
+	m_main_to_sub[0] = data;
+	m_sync = 1;
+}
```
cuavas, line 655: This isn’t really going to work properly.  The call to `synchronize()` doesn’t take effect synchronously, so the new value of `m_main_to_sub[0]` will be visible before the other devices have caught up.  You need to give `synchronize()` a callback to do the state update.

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+		m_sync = 0;
+	return m_main_to_sub[0];
+}
+
+/**************************************************************************/
+void cz1_state::main_to_sub_1_w(u8 data)
+{
+	machine().scheduler().synchronize();
+	m_main_to_sub[1] = data;
+}
```
cuavas, line 670: Same here – using `synchronize()` like this doesn’t really do what you want.

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+{
+	return m_main_to_sub[1];
+}
+
+/**************************************************************************/
+void cz1_state::sub_to_main_w(u8 data)
+{
+	machine().scheduler().synchronize();
+	m_sub_to_main = data;
+}
```
cuavas, line 683: And this one as well.

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+{
+	return m_sync;
+}
+
+/**************************************************************************/
+void cz1_state::sync_clr_w(u8)
+{
+	machine().scheduler().synchronize();
+	m_sync = 0;
+}
```
cuavas, line 708: Yet another one.

src/mame/casio/cz1.cpp
```diff
@@ -0,0 +1,924 @@
...
+	save_item(NAME(m_sub_to_main));
+	save_item(NAME(m_sync));
+	save_item(NAME(m_sync49));
+}
+
+/**************************************************************************/
+void cz1_state::machine_reset()
+{
+	m_main_port[0] = m_main_port[1] = m_main_port[2] = 0xff;
+	m_mcu_p2 = 0xff;
```
cuavas, line 782: You shouldn’t preemptively change the MCU output values on reset – doing that makes things get out of sync.  If the MCU updates its outputs on reset, it will call you to tell you.

src/mame/casio/ra3.cpp
devinacker: Using oversized RAM carts works on real CZs too, since they all use the same connector.

devinacker:
Addressed comments and also added a third software list item converted from a sysex backup.

PR #11780, opened 2023-11-25
PXA255/39in1: Improve register mapping, fix EEPROM readback.

MooglyGuy:
-pxa255: General cleanups. [Ryan Holtz]
 * Eliminated custom header in favor of scoped enums.
 * Eliminated double-dispatch switch/case in handlers, in favor of individual register handlers.
 * Reworked GPIO handling for correctness.

-39in1: Fixed GPIO hookup, now boots on its own. [Ryan Holtz]


src/devices/machine/pxa255.cpp
```diff
@@ -539,116 +766,93 @@ TIMER_CALLBACK_MEMBER(pxa255_periphs_device::ostimer_match_tick)
...
-
-  PXA255 Interrupt registers
+u32 pxa255_periphs_device::tmr_oscr_r(offs_t offset, u32 mem_mask)
+{
+	const u32 data = m_ostimer_regs.oscr;
+	LOGMASKED(LOG_OSTIMER, "%s: tmr_oscr_r: OS Timer Count Register: %08x & %08x\n", machine().describe_context(), data, mem_mask);
+	// free-running 3.something MHz counter.  this is a complete hack.
+	m_ostimer_regs.oscr += 0x300;
+	return m_ostimer_regs.oscr;
+}
```
cuavas, line 795: This one needs a `machine().side_effects_disabled()` check to avoid issues with debugger memory views.

cuavas (commented):
This looks pretty straightforward, and makes the code take advantage of more quality-of-life stuff we get these days.

I noticed one read handler missing a check for disabled side effects.

Another thing to think about, although probably better to do separately, is adding side effect checks before logging in read handlers to avoid log spam when debugger memory windows are open.  But it isn’t critical since it isn’t a regression and logging is disabled at compile time by default.

src/devices/machine/pxa255.cpp
MooglyGuy: I've fixed this in a more elegant way by bringing over the more accurate OS-timer bookkeeping code from sa1110.cpp.

PR #11785, opened 2023-11-26
Space Battle by Hoei (galaxian.cpp) - Fix corrupt speech

Paul-Arnold:
Adjust scheduler quantum during speech board comms so handshake isn't missed


Paul-Arnold:
Ooops, I see someone already fixed this :(

The original suggest fix machine().scheduler().perfect_quantum(attotime::from_usec(50));  needs to be machine().scheduler().perfect_quantum(attotime::from_usec(100));

cuavas:
See ce6dd1b0464f706ecc1bf8695de817d1bf917de1

PR #11786, opened 2023-11-26
cpu/pps4/pps4.cpp: Fix typo on opcode comment [Arcade Hacker]

clawgrip:
The code was documenting:
  iLB -  Load B indirect (0xc*)
  iTM - Transfer and mark indirect (0xc0 en el driver)

Clearly one of them is wrong. It's iTM, but it was wrong only on the comment, the implementation is good.


clawgrip:
imagen

Evidence from the CPU manual. Note that there's also a typo on the manual, saying "11xx" instead of "11x1".

PR #10920, opened 2023-02-21
taito/taito_f3_v.cpp: line clip fixes, landmakr/quizhuhu palette fix

y-ack:
- respect all 4 clip planes (needed by landmakrj, quizhuhu)
 - remove some incorrect assumptions about clip plane 3 enable bit being a clip disable
 - found a 'invert clip inverse mode' flag. most games set it, caused problems with unset pbobble4, commandw
 - don't truncate landmakr/quizhuhu palette offsets: caused green fades in quizhuhu, incorrect win message colors in landmakrj. apparently the truncation fixed symptoms in a number of games, deserves more research.

fixes mt00950: landmakr: Winning message is missing.
fixes mt01014: quizhuhu: The question text is not displayed.
fixes mt02819: pbobble4 and clones: Victory screen not shown at end of match
fixes commandw map display
fixes quizhuhu pre-taito logo intro not visible
fixes landmakr ending slide display

PR #11788, opened 2023-11-27
taito/taito_f3_v.cpp: eliminate game-specific palette hack [ywy, 12]

y-ack:
clean up after hack from #10920
mame combines palette base and pen color with addition, but the high planes in gfx overlap with lowest palette line bits in the color address (i.e. bitwise OR). account for this discrepancy in `get_tile_info`


cuavas:
I’m reluctant to credit someone as “12” – it’s going to make it damn hard to track them down to ask a question or something in the future.

y-ack:
> I’m reluctant to credit someone as “12” – it’s going to make it damn hard to track them down to ask a question or something in the future.

hi vas,
would it be acceptable to credit them as "twelve" ?
otherwise we can use "12Me21"

PR #11789, opened 2023-11-27
Fill known Taito's PCB ID

negunma:
Whether the manufacturer description is "Taito" or "Taito Corporation" follows the source file.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -42207,14 +42207,14 @@ doapp                           // Dead Or Alive ++ (JAPAN)
 doappk                          // Dead Or Alive ++ (Korea)
 flamegun                        // Flame Gunner (USA)
 flamegunj                       // Flame Gunner (JAPAN)
-ftimpact                        // Fighters' Impact (Ver 2.02O)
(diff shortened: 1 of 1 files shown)


Osso13:
sooner or later mame.lst is hopefully going away, so I think it would be preferable to add eventual missing informations to the drivers instead than here. Just my opinion, though.

negunma:
@Osso13 I was not aware that this file looks to be abolition. Thank you for notifying me.
I will organize drivers' description from now on.

angelosa:
The trouble here is that `GAME` and friends macro is unsuited for this SKU field, and extending that is part of the deprecation.

negunma:
Were "(c)", ID, etc. originally not needed in this file?
i.e. has the information of `YEAR`, `NAME`, `COMPANY` and `FULLNAME` been all that needed in here?
If so, I will consider withdrawing this pull request.

cuavas:
Comments in mame.lst is pointless churn and just one more thing that can get out of sync with the source file.  Put additional trivia in comments above the ROM definitions or something.

PR #11790, opened 2023-11-27
konami/ksys573: Add Dance Dance Revolution (GC845 VER. JAB)

987123879113:
ROM dump thanks to tweak! and Caldenza. Track 23 of 845jab02 had 2 places where the audio dropped so Taro made a dump of his own CD to replace the old dump.

Additionally, CDDA audio volume should work properly again now. DDR will read in the CD audio control page, modify what it needs, then send it back on boot. In a refactor back in Sept the code was rewritten to use ptr instead of fixed offsets in the data buffer so all pages could be sent back when requested, but it unintentionally changed the layout of the page data so the volumes in older DDR games sounded really unbalanced when it tried to set the output gain for the CDDA channels to non-sense values.

### DIFF
--- a/src/devices/machine/t10mmc.cpp
+++ b/src/devices/machine/t10mmc.cpp
@@ -1273,19 +1273,15 @@ void t10mmc::ReadData( uint8_t *data, int dataLength )
 					data[ptr++] = 0x8e; // page E, parameter is savable
 					data[ptr++] = 0x0e; // page length
 					data[ptr++] = (1 << 2) | (m_sotc << 1); // IMMED = 1
-					ptr += 6;   // skip reserved bytes
-					// connect each audio channel to 1 output port
+					ptr += 5;   // skip reserved bytes
+					// connect each audio channel to 1 output port and indicate max volume
 					data[ptr++] = 1;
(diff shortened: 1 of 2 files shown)


cuavas:
I’ve been saying this so many times lately, but can everyone please stop doing their best to trip up scripts for scraping new/promoted systems, etc.?  Just use the format seen in whatsnew text files.  Otherwise there’s a good chance the person merging in won’t edit it.  There were numerous commits in 0.261 that wouldn’t scrape properly, and that slows me down when I’ve got three days tops to get the release notes sorted out, as well as doing some basic smoke testing for the release.

There are three things here that will mess with scripts:
* The scripts expect a line of hyphens to introduce a list.  The commit here uses equals signs.
* The scripts look for list headings starting with “New ” for lists of newly added systems or software.  This does that for a list of promotions.
* This is a promoted clone, and the list heading lacks the word “clone” or “clones” which means I need to double-check the discrepancy.

I make an effort to sort this stuff out every month, but the more stuff I have to manually reconcile, the higher my workload in what’s already a very stressful week, and the more chance I’ll make mistakes.

PR #11792, opened 2023-11-28
konami/k037122.cpp: Use device_palette_interface for palette, Remove unneeded variable, Minor cleanup

cam900:


galibert:
You're going from 64k to 8k palette entries, why?

cam900:
> You're going from 64k to 8k palette entries, why?

It's to reduce unused area, and match to palette area size in RAM. (0x8000 bytes / 4 byte per color)

PR #11794, opened 2023-11-28
begin to document the original IVL releases of Easy Karaoke (KaraokeTV Star) which use a (long gone) download service rather than carts for extra songs [Sean Riddle, David Haywood]

mamehaze:
KaraokeTV Star (US, with 50 songs) [Sean Riddle, David Haywood]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -41424,6 +41424,7 @@ e100                            //
 @source:skeleton/easy_karaoke.cpp
 easykara
 karams
+karatvst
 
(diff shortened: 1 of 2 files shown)


mamehaze:
Sean needs to build an adapter to dump the NAND inside this type, as it appears the other Flash ROM is only a bootloader.

The NAND presumably stores all the graphics, sound, built in songs, and any user downloads in this case.

Sean also identified a Japanese version of the product, which also uses a long gone download service for additional songs.

mamehaze:
NAND is now dumped

PR #11797, opened 2023-11-28
machine/8042kbdc.cpp: make aux port to be alive with PS/2 interface

angelosa:
* Fix shutms11 recognizing mouse properly instead of indefinitely hang.

### DIFF
--- a/src/devices/machine/8042kbdc.cpp
+++ b/src/devices/machine/8042kbdc.cpp
@@ -387,7 +387,14 @@ void kbdc8042_device::data_w(offs_t offset, uint8_t data)
 					m_mouse.sample_rate = 100;
 					m_mouse.from_transmit = 0;
 					m_mouse.to_transmit = 0;
-					m_mouse.reporting = false;
+					if (m_keybtype == KBDC8042_PS2)
+					{
+						m_mouse.reporting = true;
+						m_mouse.on = true;
+					}
+					else
+						m_mouse.reporting = false;
+					
 					if (m_mouse.on)
 					{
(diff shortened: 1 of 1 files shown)


angelosa:
Opening as a PR because I'm not sure if this is the right fix. It looks like that 8042kbdc is really untested for anything PS/2 based.
- pcipc just uses serial mouse
- nforcepc doesn't set double as interrupt type (will fail later I presume)
- pciagp still fails for another bug when it inits keyboard ...

cuavas:
Eugh, this thing still has the PS/2 mouse simulation in the controller, duplicating the separate PS/2 mouse simulation.

First thing, I would have thought it should be impossible to get to that point if `m_mouse.on` is false, since the mouse shouldn’t be processing commands at all if the auxiliary interface is disabled (KBC commands 0xa7/0xa8 control this).

Secondly, the mouse reset command (0xff) should set the following state:
* 100 samples/second rate
* 4 counts/millimetre resolution
* 1:1 scaling
* Reporting disabled

(For example see here: https://eaw.app/Downloads/PS2_Mouse.pdf – search for Reset Mode to find the reset state.)

After completing reset, it should do a self-test and report 0xaa (good) or 0xfc (fail), followed by its device ID 0x00.

So on the face of it, this change is wrong on both counts.  If it improves behaviour, it’s working around bigger issues elsewhere.  Given the history of this device, that shouldn’t be a surprise.

angelosa:
Has no sense leaving this open after https://github.com/mamedev/mame/commit/b77af1eb3859dabb2d620e9a79afef7a4985b01c

PR #154, opened 2015-03-29
Cosmetics build system changes

ramiropolla:

PR #11798, opened 2023-11-29
atari/toobin.cpp: Toobin' actually does have a TMS5220.

codefrog2002:
Toobin' does have a TMS5220 speech chip.
In gameplay it's most easily accessed when going over waterfalls (random trigger, not every time).
In the service mode, sound test codes #141 to #154. 
Reference: https://youtu.be/M1LW0jWvDg8 around 1:47 you can hear "whooooaah".

### DIFF
--- a/src/mame/atari/toobin.cpp
+++ b/src/mame/atari/toobin.cpp
@@ -555,7 +555,6 @@ void toobin_state::toobin(machine_config &config)
 	m_jsa->test_read_cb().set_ioport("FF9000").bit(12);
 	m_jsa->add_route(0, "lspeaker", 1.0);
 	m_jsa->add_route(1, "rspeaker", 1.0);
-	config.device_remove("jsa:tms");
 }
 
 


cuavas:
This is one of Aaron’s drivers, and as I understand it he based his Atari drivers on their hardware reference documentation.  With that in mind, I’d tend to believe he had a reason for removing the TMS5220 from the machine configuration in this case.

Have you ruled out other possibilities?  For example:
* Code to use TMS5220 was present, but the actual chip and support circuitry was cut before release to reduce costs.
* The video you showed uses a modified cabinet where the sound board has been swapped for a JSA variant with a TMS5220 present.

As supporting evidence, the operator’s manual ([available on the Internet Archive](https://archive.org/details/ArcadeGameManualToobin)) shows a PCB layout and lists components for the sound board, but makes no mention of a TMS5220 and associated support components.

Without convincing evidence to the contrary, I would agree with Aaron’s implementation.  It’s most likely the TMS5220 was planned to be included, but cut to save costs, leaving behind vestigial code that would have used it.

ajrhacker:
The schematics show some components on the main board labeled as "not stuffed" that were certainly unused: a LETA trackball interface and SLAPSTIC protection chip. The TMS5220C on the "Stand-Alone Audio PCB" is not thus labeled.

In MAME, Aaron Giles made the TMS5220 an optional component in April 1999. It doesn't seem to have been discussed at the time why it was removed from Toobin' specifically, aside from Aaron briefly mentioning "stuff where they were trying to use the YM2151 as a DAC (see Toobin' or Robot Monsters for examples)." However, Escape from the Planet of the Robot Monsters kept its TMS5220 at this time.

codefrog2002:
PXL_20231129_180111688
This is the best photo I could get of the sound board in the machine I have access to. It's in what seems like original condition. I bought it from an arcade in 1992 and later sold it to Digital Eclipse. Videos to follow of gameplay and sound test mode.

Lord-Nightmare:
That A043713 JSA-I PCB you posted a picture of doesn't have the TMS5220C populated, you can see the space where it would be just in the center of the photo. It is definitely a Toobin pcb, too, note the "TBN0642" serial number.

If the game is actually trying to play the "whoaaa!" sample via the TMS5220C when you go over the waterfall, I'm going to assume that's an accidental development leftover.

It might be worth posting about that on TCRF if you have a log of what it is sending to the speech chip, though!

angelosa:
If there's an empty space and somebody _did made it work on an actual PCB_ then this belongs to the DIY category that MAME don't have an exact infrastructure for, i.e. same league as Twin Eagle PCB having an empty ROM socket as debug overlay or PokeyMax for Atari 8-bit computers.

Lord-Nightmare:
The JSA-I (A043713) PCB is used on Toobin, Vindicators, Xybots, Blasteroids, and (and only on this version is the TMS5220C populated, and the POKEY left out) Escape from the Planet of the Robot Monsters (DEDICATED CABINETS ONLY)
The non-dedicated/conversion (JAMMA) Escape from the Planet of the Robot Monsters uses the A047069 PCB instead, which is basically a JSA-I-plus-JAMMA interface, and doesn't even have a socket for the POKEY at all. It does have a TMS5220C on it.

codefrog2002:
OK wow. The speechy sfx are coming out of the YM2151, but in the emulator (both MAME and my own) the driver doesn't play them if the TMS5220 isn't hooked up. I will check why and get back to you.

codefrog2002:
Well now I'm just embarrassed. Sorry for wasting people's time.

Lord-Nightmare:
They did do this, since TMS52xx chips were likely fairly expensive.
I have a theory that Earl Vickers(?) did actually record some of the voices, particularly on Xybots, and LPC-encoded them to monotonize the pitch, then played them back and digitized that, and then used the YM2151 on Xybots to play that audio back during gameplay. I can hear what I think are some LPC-ish artifacts in the robot overlord voices on Xybots that otherwise wouldn't be there, although the samples are such low quality (4-bits I think) that its hard to tell.

PR #11801, opened 2023-11-30
heathkit/h89.cpp: Add SigmaSoft Interactive Graphics Controller

mgarlanger:
Adds the SigmaSoft and Systems Interactive Graphics Controller (IGC) device for the H89 computer. 
Unlike the other graphics upgrades for the Heath H19 which had all the changes/connections in the H19 (or the terminal portion of the H89), this board connected to the H89's CPU board and it's the CPU board's Z80 which drives the graphics board.  Note: the IGC could also work with Heath H8 and H19, but in that configuration, the H8's CPU would still be controlling the IGC, not the TLB. 

I'm not sure this is the best way to structure the code. This IGC can be used with some of the alternate ROMs on the TLB board, such as Super19, UltraROM, and Watzman, in addition to the original Heath ROM. If there is a clean way to allow any of these ROM with the IGC, that would be preferred to this approach.

Screenshot 2023-11-29 at 10 03 03 AM
Screenshot 2023-11-29 at 7 00 29 PM


src/mame/heathkit/sigmasoft_parallel_port.cpp
```diff
@@ -0,0 +1,173 @@
...
+	m_video_mem_w(*this),
+	m_io_lo_addr(*this),
+	m_io_hi_addr(*this),
+	m_window_lo_addr(*this),
+	m_window_hi_addr(*this),
+	m_ctrl_w(*this)
+{
+}
+
+void sigmasoft_parallel_port::video_mem_w(u8 val)
```
rb6502: Are these trampolines necessary?

src/mame/heathkit/tlb.cpp
```diff
@@ -59,6 +59,22 @@
...
+#define LOG_FUNC (1U << 4)   // Function calls
+
+#define VERBOSE (LOG_REG | LOG_FUNC)
+
+#include "logmacro.h"
+
+#define LOGREG(...)        LOGMASKED(LOG_REG, __VA_ARGS__)
+#define LOGFUNC(...)       LOGMASKED(LOG_FUNC, __VA_ARGS__)
+
+#ifdef _MSC_VER
```
rb6502, line 80: This seems like something the core should provide in some way.

src/mame/heathkit/h89.cpp
```diff
@@ -598,6 +598,9 @@ static void tlb_options(device_slot_interface &device)
 static void sigma_tlb_options(device_slot_interface &device)
 {
 	device.option_add("igc", HEATH_IGC);
+	device.option_add("igc_super19", HEATH_IGC_SUPER19);
```
rb6502, line 634: If the only difference here is the ROM, this should just use BIOS options.   (Yes, slot devices can have BIOS options - for an extreme example, look at bus/isa/xtide.cpp, which has 24+ ROMs selectable).

mgarlanger: There are some additional difference that make that hard. The other ROMs all change the definitions of what the dip switch settings mean. The UltraROM has an addition 2k of RAM for off-screen memory. And both the UltraROM and Watzman ROM had different keyboard rom than the original. There are 2 versions of the Watzman ROM, and they are currently using BIOS options.

src/mame/heathkit/tlb.cpp
mgarlanger: I saw this done like this in several other files, I guess it could go in logmacro.h, but that would probably require a large number of files to be changed.

mgarlanger:
@cuavas This PR has been updated to use a similar approach as PR 11898 based on the feedback you provided there.

PR #11805, opened 2023-12-01
mame/bfm: Implement Bell Fruit 96x8 dot matrix display for Scorpion 5.

blueonesarefaster:
This implements the Bell Fruit GU96X8M-K629C3 96 * 8 VFD with led backlight.

ATM I've simply replaced the use of the BDA display which had 16 5*7 dot matrix digits with this new display driver.
If the Scorpion 5 driver is ever progressed the correct display will have to be loaded for each game.

### DIFF
--- a/src/mame/bfm/bfm_gu96x8m_k657c2.cpp
+++ b/src/mame/bfm/bfm_gu96x8m_k657c2.cpp
@@ -1,0 +1,1010 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/**********************************************************************
+
+    Bell Fruit Games 96x8 Dot matrix VFD module interface and emulation.
+
+	This was a replacement for the previous 5 x 7 x 16 display.
+	The two displays are meant to be compatible although there are some
+	differences in behaviour when sending undocumented commands or
+	commands with "don't care" bits.
+
+    TODO: Scrolling text
+		  Test sequence
+		  Background character
+		  Background colour enable/disable/protection
+		  LED backlight isn't visually correct. The area outside the 96x8
+		  matrix should be brighter than that shining through the gap
+		  between the digits.
+**********************************************************************/
+
+#include "emu.h"
+#include "bfm_gu96x8m_k657c2.h"
+
+DEFINE_DEVICE_TYPE(BFM_GU96X8M_K657C2, bfm_gu96x8m_k657c2_device, "bfm_gu96x8m_k657c2", "BFG 192x8 VFD controller")
+
+static const uint8_t charset[][6]=
+{
+	{ 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 },
+	{ 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 },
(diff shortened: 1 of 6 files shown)


mamehaze:
"If the Scorpion 5 driver is ever progressed the correct display will have to be loaded for each game"  are the custom characters not uploaded by the game?

Ideally it would be good to see all these display drivers decapped to see if they're MCUs, or at the very least contain a ROM with these character tables in them.

blueonesarefaster:
> "If the Scorpion 5 driver is ever progressed the correct display will have to be loaded for each game" are the custom characters not uploaded by the game?

Yes, the custom or UDF are uploaded by each game. What I meant was some games use this 96*8 dot matrix display and others use a 16 * 5*7 dot matrix display which are obviously visually quite different

> 
> Ideally it would be good to see all these display drivers decapped to see if they're MCUs, or at the very least contain a ROM with these character tables in them.

The display emulated here uses an ATMEGA8515, I don't know if that can be dumped ?
I don't know what the previous dot matrix display used.

src/mame/bfm/bfm_gu96x8m_k657c2.cpp
```diff
@@ -0,0 +1,1010 @@
+// license:BSD-3-Clause
+// copyright-holders:
```
rb6502, line 2: We do need this to list something for the copyright holder :-)

src/mame/bfm/bfm_gu96x8m_k657c2.cpp
```diff
@@ -0,0 +1,1010 @@
...
+					{
+						udf[4 - (i % 5)] |= BIT(m_extra_data[(i + 2) / 8], 7 - ((i + 2) % 8)) << ((i / 5) + 1);
+					}
+					m_load_extra_data = LOAD_COMPLETE;
+				}
+				break;
+
+			case LOAD_LED_COLOUR:
+				m_load_extra_data = LOAD_COMPLETE;
+
```
rb6502, line 416: This would be cleaner as a switch.

src/mame/bfm/bfm_gu96x8m_k657c2.cpp
blueonesarefaster: OK, I can put my name there.
As this file was derived from bfm_bda.cpp by James Wallace do I need to do something to acknowledge that ?

blueonesarefaster:
I think the copyright messages are probably wrong ?
Looking at bfm_bda which this was derived from the header says David Haywood and the source says James Wallace.
I've restored the messages in this PR to match the files it was derived from and added my name to the source.
Hopefully someone can correct me if I'm wrong

james-wallace-ghub:
As far as I'm concerned, you don't need to consider this as from me, you've changed enough. The copyright is because the BDA HLE is actually a derivative of my BD1 code for Scorpion 2, as we didn't have the dot matrix data in a usable form. David may well have changed it as part of the Scorpion 4 hookup, but feel free to assign yourself rather than me on this.

PR #11811, opened 2023-12-02
taito/taito_f3_v.cpp: major rewrite

y-ack:
* simplify sprite logic, eliminate game-specific kludge
 * simplify line ram access 
 * simplify draw loop
 * fix sprite bankswitch - riding fight sprites now 60fps
 * more accurate sprite scaling 
 * x mosaic effect (ring rage, riding fight, command war, moriguchi hiroko no quiz de hyuu!hyuu!)
 * screen raw parameters based on others' measurements and mosaic quirk
 * identify mosaic flags for sprites and pivot layer
 * re-derive blending rules (riding fight, **space invaders dx**, **elevator action returns**, recalhorn, gekirindan, **bubble memories**, cleopatra fortune, **puchi carat**, **land maker**)
 * identify sprite and pivot blend value select (arabian magic, grid seeker)
 * line ram alternate subsections (bubble memories)
 * sprite framebuffer unblit disable (sprite trails) (darius gaiden)
 * emulate blending priority conflict (**darius gaiden**, bubble memories, grid seeker)

MT#8697, MT#2527, MT#1922, MT#26, MT#2733, MT#3741, MT#2002, MT#1907, MT#7321, MT#2097, MT#1924, MT#1923, MT#8827, GH #10033

Systems promoted to working
---------------------------
Moriguchi Hiroko no Quiz de Hyuu!Hyuu! [ywy, 12Me21]


y-ack:
very rough draft pr for tracking, not ready for review.  may be slow until late december.
 * [ ] basic sprite drawing
 * [ ] basic tilemap drawing
 * [ ] basic pivot layer drawing
 * [ ] more accurate sprite scaling behavior
 * [ ] line alternate tilemaps
 * [ ] line column scroll
 * [ ] line clipping
 * [ ] line pivot control
 * [ ] line blend
 * [ ] line x sample/mosaic
 * [ ] line x forward blend
 * [ ] line bg palette
 * [ ] line pivot enable
 * [ ] line pivot priority
 * [ ] line sprite clipping
 * [ ] line sprite brightness
 * [ ] line sprite priority
 * [ ] line playfield scaling
 * [ ] line playfield palette addition
 * [ ] line playfield rowscroll
 * [ ] line playfield priority
 * [ ] line playfield clipping
 * [ ] layer additive blending
first pass forgoes optimizations in order to later debug blending behavior.  when closer to ready, performance regression should be evaluated and optimization passes may be necessary.
notable optimizations that have currently been **lost**:
 * [ ] priority buffer / opaque layer skip
 * [ ] line draw grouping by constant effect states

angelosa:
What's the state of this? I do believe it's already better to what's in bleeding edge so can be marked as ready?

y-ack:
status is..  not 100% correct (even blending) but code in a place where it can reasonably be worked on.
not totally opposed to putting one or two hacks in place to get this through and allow smaller iterative prs on top of it...

 - mosaic implemented (ringrage, ridingf, commandw, quizhuhu etc.)
   - promotes quizhuhu to working 
   - doesn't yet emulate the column 366 counter reset quirk
 - sprite scaling improved (accuracy and code)
   - closes [mt#8697](https://mametesters.org/view.php?id=8697) (kaiserknj)
   - not 100% hardware accurate in some cases?
 - sprite rendering improvements: gets rid of dariusg kludge, fixes ridingf sprite framerate, closes [mt#2527](https://mametesters.org/view.php?id=2527) (rayforce), [mt#1922](https://mametesters.org/view.php?id=1922) (dariusg)
 - blending improvements
   - closes [mt#26](https://mametesters.org/view.php?id=26) (gseeker), [mt#2733](https://mametesters.org/view.php?id=2733) (arabianm)
   - closes [mt#3741](https://mametesters.org/view.php?id=3741) (puchicar), [mt#2002](https://mametesters.org/view.php?id=2002) [mt#5893](https://mametesters.org/view.php?id=5893) [mt#1907](https://mametesters.org/view.php?id=1907) (elvactr), [mt#7321](https://mametesters.org/view.php?id=7321) (bubblem), [mt#2097](https://mametesters.org/view.php?id=2097) (spcinvdj)
    - eliminates game-specific palette hack for recalh, cleopatr
    - also improves landmakr, ridingf, ...
 - implemented alternate line ram subsections: closes [mt#7321](https://mametesters.org/view.php?id=7321) (bubblem)
 - implemented blending priority conflict quirk: closes [mt#7321](https://mametesters.org/view.php?id=7321) (bubblem), [mt#1924](https://mametesters.org/view.php?id=1924) (dariusg)

**currently known regressions**:
 - arabianm, after stage 3 boss, sprite block attribute locking causes corrupted scaled sprites.  understanding this edge case properly is expected to take a lot of testing.  a very simple unobtrusive game-specific hack for it is possible.
 - gseeker hangar shadow (mt#1187), the hack that put the plane over the shadow was eliminated, so the plane is now underneath the shadow layer.
 - we also disabled the 'sprite lag' behavior which should probably be re-enabled for now.

notable known remaining issues (not regressions):
 - [mt#1923](https://mametesters.org/view.php?id=1923)
   - some blending assumption is still wrong, but its effects are limited to this and gekiridn?
 - scfinals 'replay' is choppy
   - something about sprite ram write timing?  fbneo doesn't have this issue.
 - [mt#8230](https://mametesters.org/view.php?id=8230) (rayforce shadows), [mt#1187](https://mametesters.org/view.php?id=1187) (gseeker shadows)
   - my current belief is that there are bits at 7400(for sp) and 6400(for pfs) that can cause a layer to be darkened by a layer underneath it... /separately/ from the priority conflict that dariusg uses for a similar effect
 - playfield scaling accuracy (ridingf ending, scfinals name entry)
 - palette interpretation bits must be respected for the gseeker intro zoom

other mametesters issues:
[mt#1172](https://mametesters.org/view.php?id=1172) (kaiserkn shadows) - ??? they *are* green, just dark, possibly gamma issue ? but need to check the blending again just in case
[mt#956](https://mametesters.org/view.php?id=956) (trstar attract text) - don't believe in this without pcb proof
[mt#1171](https://mametesters.org/view.php?id=1171) - btanb
[mt#7776](https://mametesters.org/view.php?id=7776) - dariusg zone D - btanb??? behavior is on pcb ?


**performance regression**:
in a not-at-all-controlled-environment average-of-3 test of the windows build artifact:
`mame.exe -w -resolution 320x232 ridingf -bench 120`
this @ 929d445: 329.26%
bleeding edge: 494.04%

(unacceptable by itself, but also totally unexplored.)

angelosa:
fyi I don't mind if this will end up having a (relatively negligible) performance hit.

y-ack:
okay, this is probably ready for review now.

<details><summary>show recordings of most visible changes</summary>
note: gamma is set to 0.8 in all recordings


https://github.com/mamedev/mame/assets/12588017/465ff6f9-e902-4105-befd-8b5797a5ab2c


https://github.com/mamedev/mame/assets/12588017/ff3a0a98-7649-48ee-aa9f-18096adffb89


https://github.com/mamedev/mame/assets/12588017/38f90533-512c-4882-96ea-cf410ceb6e08


https://github.com/mamedev/mame/assets/12588017/25840463-6c5e-410c-92db-2a957a793e24


https://github.com/mamedev/mame/assets/12588017/3c74c668-b1e7-4255-a90e-75198de81f3b


https://github.com/mamedev/mame/assets/12588017/29af0f0b-c8d9-4580-ad68-47137eb7210e


https://github.com/mamedev/mame/assets/12588017/5183b615-d0c8-4b73-9d8a-de15d3f55d23


https://github.com/mamedev/mame/assets/12588017/77658f30-d9bb-4212-82f5-4cb0d58ea340


https://github.com/mamedev/mame/assets/12588017/badf5415-c4f4-477d-b287-1e4b14b4d01e


https://github.com/mamedev/mame/assets/12588017/a3779d27-28fb-43c9-b1b1-02127640a544


https://github.com/mamedev/mame/assets/12588017/13415844-298d-486c-96c7-1b4657243042


https://github.com/mamedev/mame/assets/12588017/35d6388d-ecd2-498c-8430-e9a0ada81261


https://github.com/mamedev/mame/assets/12588017/45a6c05c-e885-42e1-a81c-bf9727a564a8


https://github.com/mamedev/mame/assets/12588017/fe8f60a5-53e0-4c3c-afc9-41a6375b4022

</details>

angelosa:
I guess it's safe that you can promote quizhuhu already from this very PR. 
A commit with:
```
Systems promoted to working
---------------------------
Moriguchi Hiroko no Quiz de Hyuu!Hyuu! [<your preferred username handle>]

```
Should be enough to be catched later by MAME release script, cfr. whatsnew patterns at https://github.com/mamedev/build/blob/master/whatsnew/whatsnew_template.txt

src/mame/taito/taito_f3_v.cpp
```diff
@@ -366,94 +390,108 @@ TILE_GET_INFO_MEMBER(taito_f3_state::get_tile_info)
...
+		//get_sprite_info();
+	}
+}
+
+void taito_f3_state::set_extend(bool state) {
+	m_extend = state;
+	// TODO: we need to free these if this is called multiple times
+	// for (int i=0; i<8; i++) {
+	// 	delete m_tilemap[i];
+	// }
```
cuavas, line 457: Can you rename this to `create_tilemaps` or something to make it clearer what it does, and call the parameter `extend`?

Also, you really need to restructure the code so this *can’t* be called more than once.  Tilemaps are one of those things with session lifetime – they need to be created at start time, they get cleaned up automatically on stop, and they automatically participate in save states.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -700,2153 +686,737 @@ u16 taito_f3_state::lineram_r(offs_t offset)
...
-		if (offset == 0xb000 / 2 && data == 0x003f)
-		{
-			m_f3_skip_this_frame = 1;
-			return;
-		}
-	}
-#endif
-
 	COMBINE_DATA(&m_line_ram[offset]);
 }
```
cuavas, line 690: Since there’s no additional code here, can the line RAM be installed directly now, avoiding the trampoline?

src/mame/taito/taito_f3_v.cpp
```diff
@@ -700,2153 +686,737 @@ u16 taito_f3_state::lineram_r(offs_t offset)
...
+}
+
+/******************************************************************************/
+
+// line: [latched] line info from previous call, will modify in-place
+// y should be called 0->255 for non-flipscreen, 255->0 for flipscreen
+void taito_f3_state::read_line_ram(f3_line_inf &line, int y)
+{
+	const auto latched_addr = [=] (u8 section, u8 subsection) -> offs_t
 	{
```
cuavas, line 717: Please don’t use auto capture in lambdas – it’s better to be explicit abut what you want to capture.

src/mame/taito/taito_f3_v.cpp
```diff
@@ -700,2153 +686,737 @@ u16 taito_f3_state::lineram_r(offs_t offset)
...
+		multi = BIT(spritecont, 3);
+
+		const int tile = spr[0] | (BIT(spr[5], 0) << 16);
+		if (!tile) continue; // todo: is this the correct way to tell if a sprite exists?
+
+		const fixed8 tx = m_flipscreen ? (512<<8) - x.block_scale*16 - x.pos : x.pos;
+		const fixed8 ty = m_flipscreen ? (256<<8) - y.block_scale*16 - y.pos : y.pos;
+
+		if (tx + x.block_scale*16 <= visarea.min_x<<8 || tx > visarea.max_x<<8 || ty + y.block_scale*16 <= visarea.min_y<<8 || ty > visarea.max_y<<8)
+			continue;
```
cuavas, line 1363: You now increment `total_sprites` *before* checks that cause the sprite to be ignored, while previously it happened *after* those checks.  This changes the way the sprite limit works – previously it excluded ignored sprites, now it includes them.  Is that intentional?

cuavas (changes requested):
Before anything else, please clean up your changes to match the existing code in the files you’re working on.
* Please Allman formatting consistently (opening braces on their own lines).
* Please place right-associative qualifiers after the space (e.g. `int *a;` rather than `int* a;`).

TheCoolPup:
Regression for Elevator Action Returns, the airplane section is broken now 
Elevator Action Returns (Ver 2 2O 1995_02_20)  elvactr  - MAME 0 264 (LLP64) 4_4_2024 4_17_16 PM
Elevator Action Returns (Ver 2 2O 1995_02_20)  elvactr  - MAME 0 264 (LLP64) 4_4_2024 5_09_05 PM

src/mame/taito/taito_f3_v.cpp
12Me21: `total_sprites` is a safeguard to prevent infinite loops; under normal circumstances the limit is never reached

src/mame/taito/taito_f3_v.cpp
12Me21: yes
Should we also remove the unnecessary read handlers? (most of the `_r` functions)

y-ack:
> Regression for Elevator Action Returns, the airplane section is broken now 

hi @TheCoolPup, thanks for finding this.  it looks like playfields and sprites get disabled when set to the 'wrong' opaque mode.

y-ack:
> Before anything else, please clean up your changes to match the existing code in the files you’re working on.
>
>    Please Allman formatting consistently (opening braces on their own lines).

@cuavas we've rewritten 90% of the code already, it's easier at this point to switch to (consistently) k&r.
my understanding is that no one else has seriously worked on the driver in years, i hope it's ok?

src/mame/taito/taito_f3_v.cpp
cuavas: > `total_sprites` is a safeguard to prevent infinite loops; under normal circumstances the limit is never reached

Yes, but it will be reached faster with this change than it would have been before.

12Me21: The limit is equal to the number of entries in sprite ram (1024), so the **only** way to reach that limit is by reading the same sprite multiple times (i.e. the sprite list has a jump command which jumps backwards and creates an infinite loop)
(exiting a little sooner is fine, since the duplicate sprites would just get drawn on top of each other anyway)

If there's an infinite loop containing only nonvisible sprites, the old code gets stuck forever because total_sprites never increments

src/mame/taito/taito_f3.h
```diff
@@ -167,136 +172,220 @@ class taito_f3_state : public driver_device
...
+		int y{0};
+		int screen_y{0};
+		pri_mode pri_alp[432]{};
+		// 5000/4000
+		clip_plane_inf clip[NUM_CLIPPLANES];
+		// 6000 - pivot_control, sprite alpha
+		u16 maybe_sync_reg{0};
+		bool no_opaque_dest{false};
+		// 6200
+		u8 blend[4]{0}; // less 0 - 8 more
```
happppp, line 333: FYI initializing arrays like this, it will only initialize the 1st index, and fill the rest with 0. Best use {} like you did elsewhere.

see:
int m_array[4] { 1 };
for (int i = 0; i < 4 ; i++) printf("%d ",m_array[i]);
-> 1 0 0 0

src/mame/taito/taito_f3.cpp
```diff
@@ -9,19 +9,10 @@
     Major thanks to Aaron Giles for sound info, figuring out the 68K/ES5505
     rom interface and ES5505 emulator!
     Thanks to Acho A. Tang for Kirameki Star Road sound banking info!
-    Thank you to Shiriru for the scanline rendering (including alpha blending),
```
happppp, line 12: Let's keep it in. It's Bryan thanking Shiriru for past contributions.

happppp:
If it is as you say and 90% of the file is yours, then yes source code style choice is yours. Also, good time to include yourself to copyright holders.

src/mame/taito/taito_f3.h
```diff
@@ -12,6 +12,9 @@
 #include "emupal.h"
 #include "screen.h"
 #include "tilemap.h"
+#include <bitset>
+
+using fixed8 = int;
```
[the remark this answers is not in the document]
happppp: What variable type is this really? Just use s8/s16/s32?

src/mame/taito/taito_f3.cpp
angelosa: Legacy, can as well go in header driver file as `// thanks-to: Shiriru`, usage example: https://github.com/mamedev/mame/blob/ca7daca08f78d85deb5c0b1af974f208063265df/src/mame/nec/pc8801.cpp#L1-L3

cuavas:
I don’t want it to end up with different formatting for the different Taito F3 source files, though.  There’s taito_f3.cpp as well.

y-ack:
thanks for the comments.  i agree that having different styles across the taito f3 files should be avoided.
`taito_f3.cpp` has only 4 functions that would need reformatting.
we hadn't touched it yet because there is so much to clean up there* (and heavily mame-specific idioms that require finding modern driver examples) that it seemed like the increased scope warranted focusing on in a separate PR.  this one was supposed to address the video rendering primarily.

[*] for example:
 - the cpu clocks are wrong (as noted elsewhere in the file, it's 15.23809MHz, not 16)
 - we need to do more hardware testing to figure out how the interrupts are generated (there seems to be an hblank interrupt which not all games enable)
 - tile decoding wants refactoring?
 - repetitive init routines for almost all sets feels really bad (though maybe that's just necessary and normal for mame?)

TheCoolPup:
Forgot to take screenshots but two other bugs i noticed is that in stage 4 when you are underground when you go through a subway car it doesn’t show you in it, also when you stop the missile at the end of the final stage it’s supposed to fade to white but instead it just turns black until the credits start

y-ack:
> in stage 4 when you are underground when you go through a subway car it doesn’t show you in it, also when you stop the missile at the end of the final stage it’s supposed to fade to white but instead it just turns black until the credits start

@TheCoolPup: the first one was the same bug as before; the second is the 'background palette' effect (so probably not a regression).  both should be fixed in the latest commit.

src/mame/taito/taito_f3.h
y-ack: thanks; fixed.

src/mame/taito/taito_f3_v.cpp
y-ack: done

src/mame/taito/taito_f3_v.cpp
y-ack: immediate ask done, though tilemaps in f3 might need rethinking in general at some point.  we like to construct them for the debug viewer, but then have to just read the pixmap to apply line effects properly, and then they participate in additive blending.

src/mame/taito/taito_f3.h
```diff
@@ -309,74 +398,16 @@ class taito_f3_state : public driver_device
...
-	void get_spritealphaclip_info();
-	void get_line_ram_info(tilemap_t *tmap, int sx, int sy, int pos, const u16 *pf_data_n);
-	void get_vram_info(tilemap_t *vram_tilemap, tilemap_t *pixel_tilemap, int sx, int sy);
+	void set_extend(bool state);
+
+	inline void f3_drawgfx(const tempsprite &sprite, const rectangle &cliprect);
+	void draw_sprites(const rectangle &cliprect);
+	void get_sprite_info();
+	void read_line_ram(f3_line_inf &line, int y);
+	std::vector<clip_plane_inf> calc_clip(const clip_plane_inf (&clip)[NUM_CLIPPLANES], const mixable *line);
```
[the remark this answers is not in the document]
y-ack: ah. thanks!  done, yay

src/mame/taito/taito_f3.h
```diff
@@ -167,136 +172,220 @@ class taito_f3_state : public driver_device
...
+
+		virtual inline u16 palette_adjust(u16 pal) const { return pal; };
+		virtual inline int y_index(int y) const;
+		virtual inline int x_index(int x) const;
+		virtual bool blend_select(const u8 *line_flags, int x) const { return false; };
+
+		virtual bool used(int y) const { return true; };
+		u8 debug_index{0};
+		virtual const char* debug_name() { return "MX"; };
+	};
```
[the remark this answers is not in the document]
y-ack: should all be fixed now.

src/mame/taito/taito_f3.h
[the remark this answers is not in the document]
y-ack: oops; done.

src/mame/taito/taito_f3.h
```diff
@@ -167,136 +172,220 @@ class taito_f3_state : public driver_device
...
-		u8 *tsrc[256]{}, *tsrc_s[256]{};
-		int x_count[256]{};
-		u32 x_zoom[256]{};
-		u32 clip_in[256]{};
-		u32 clip_ex[256]{};
-		u16 pal_add[256]{};
+	static const int NUM_PLAYFIELDS = 4;
+	static const int NUM_TILEMAPS = 5;
+	static const int NUM_SPRITEGROUPS = 4; // high 2 bits of color
+	static const int NUM_CLIPPLANES = 4;
```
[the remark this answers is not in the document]
y-ack: done

src/mame/taito/taito_f3.h
```diff
@@ -167,136 +172,220 @@ class taito_f3_state : public driver_device
...
+			l = (l & 0xff) | left<<8;
+			r = (r & 0xff) | right<<8;
+			return this;
+		}
+		auto set_lower(u8 left, u8 right) {
+			l = (l & 0x100) | left;
+			r = (r & 0x100) | right;
+			return this;
+		}
 	};
```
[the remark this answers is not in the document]
y-ack: k&r'ed to effect and suggestion taken

src/mame/taito/taito_f3_v.cpp
y-ack: sorry, i'm a little confused how to do this with the 16 bit data width.

src/mame/taito/taito_f3.cpp
y-ack: done, thanks for the guidance here.

angelosa:
The alternative would be "go in full and use K&R everywhere in these files, as long as it's consistent".

12Me21:
> The alternative would be "go to the other end and use K&R everywhere in these files, as long as it's consistent".

she preemptively tentatively did that in 37ad667 and fce4bd9

src/mame/taito/taito_f3_v.cpp
angelosa: Right, those are 16-bit handlers for being implicitly reused by `2mindril`. The real fix would be to actually convert the whole block to a device, something imo out of scope with this PR.

angelosa:
imo `taito_f3_v.cpp` will ultimately require a split into device(s) so it makes sense to have that with whatever style you prefer.
As for regular `taito_f3.cpp` :
- `f3_coin_r`/`f3_control_r`/_w/`f3_analog_r` all belongs to `TC0640FIO`, which we already have a core of;
- `sound_reset_0_w`/`sound_reset_1_w` can clearly be expressed as lambda nowadays;
- Not sure about that kirameki soundbank interaction, by face value I guess it can be removed by loading sample ROMs properly (while hoping the game has a sound test that covers those);
- gfxdecode, set_raw etc. belongs to the video device, so original point stands;

All of this is non-blocking for the PR, and there are roughly two weeks for the next release.

y-ack:
i think our roadmap from here is:
 - performance work (this is already done, but want to split into another pr to handle any contentiousness separately, and if this one is "approach readability to fix blending"), could be this cycle.
 - device-ify (and timings) (this might take a little longer given relative unfamiliarity, but it's something i've had in the back of my mind the whole time)
   - maybe rework for partial screen updates at this time too
   - the existing FIO device may need work
   - not sure if we want to consider the FDP and FDA as separate units, since as far as we can tell the FDA would just be a single function (color lookup/blending) and it's not known what the inputs to that look like outside of the context of the FDP.
   - have to consider seoul symphony bootleg board in this?
   - kirameki sound banking - we don't have the cart pcb and would like to take a look at it first to make sure, but this probably really is banking.
 - miscellaneous fixes: playfield scaling/scroll correctness (ridingf/scfinals etc.), palette interpretation, secret third way to generate shadows (rayforce/gseeker), sprite ram write timing(?) (scfinals)

CPS42069:
the classic 5th boss explosion bug in Grid Seeker is still there, tested on latest nightly build.

12Me21:
> the classic 5th boss explosion bug in Grid Seeker is still there, tested on latest nightly build.
yes, this is caused by the game writing sprites past the end of the sprite list and then never clearing them.
from hardware testing a few days ago, it seems like only the first ~864 sprites are rendered, rather than 1024 like was previously assumed.
(it should be fixed in a future pr, after we find the exact limit)

PR #11814, opened 2023-12-03
konami/viper.cpp: convert EPIC to address_map & logmacro

angelosa:


src/mame/konami/viper.cpp
```diff
@@ -776,121 +775,346 @@ void viper_state::pci_config_data_w(uint64_t data)
...
+		})
+	);
+	map(0x41130, 0x41130).select(0xc0).lw8(
+		NAME([this] (offs_t offset, u8 data) {
+			int timer_num = offset >> 4;
+
+			m_epic.irq[MPC8240_GTIMER0_IRQ + timer_num].destination = data & 0x1;
+
+			if (data)
+				throw emu_fatalerror("GTDR%d in P1 mode", timer_num, data & 1);
```
angelosa, line 1020: I converted all of these because they clearly don't make sense in Konami Viper context and they are actively unused (they dispatch irq to a different processor if enabled).

PR #11819, opened 2023-12-05
Add Tandberg TDV-2115L

Frodevan:
This is the first step in a larger project of adding the first generations of Tandberg terminals/computers. This machine is the simplest in the TDV-2100 series, being only a dumb-terminal. The machine-driver for the TDV-2115L serves as a proof-of-concept for what to come for now, other products in the series will reuse the drivers added by this PR, just like the actual hardware.

The implementation of the terminal and keyboard is pretty much complete, the only thing of significance missing is the CPU interface. However, the TDV-2115 in particular did _not_ come default with a CPU module.


Frodevan (commented):
There was an overlooked typo

angelosa:
Note that you don't need to comment each commit with a conversation, we generally see the merged PR version here for evaluating.
Also I suppose you made 4 commits from web because it started 4 different CI pipelines, which I'm gonna cancel the earlier ones for saving your eligibility for using them.

angelosa:
Thanks.

Tip: you can append a `[skip ci]` at the end of a commit message to skip CI action triggers, in cases where building a full executable build is overkill like typo fixes.
https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+{
+	if(state)
+	{
+		int column = m_matrix[column_counter]->read();
+
+		for(int row_counter=0; row_counter<8; row_counter++)
+		{
+			int key_mask = 1<<row_counter;
+			int key_state = column&key_mask;
+			if(key_state != (keystate[column_counter]&key_mask))
```
cuavas, line 178: You can use the `BIT` helper to make this code tidier.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+			{
+			// This simulates the enumeration done by the internal LUT of the 20-04592-013
+			uint8_t key_nr = (~((column_counter-1)*8 + row_counter + 1))&0xFF;
+				if(key_state)
```
cuavas, line 182: I expect this should cause a misleading indentation error with compilers that support it.

src/mame/tandberg/tdv2100_kbd.h
```diff
@@ -0,0 +1,86 @@
...
+	bool shift_lock;
+	bool control;
+
+	TIMER_CALLBACK_MEMBER(key_repeat);
+	emu_timer *m_key_repeat_trigger;
+	uint8_t char_buffer;
+	uint8_t	key_nr_in_buffer;
+	int key_repeat_delay_ms;
+	int key_repeat_rate_hz;
+	bool inhibit_key_from_params;
```
cuavas, line 80: You’ve got a mixture of data members with and without the `m_` prefix.  That makes the code confusing to read.  If you’re going to use the `m_` prefix for data members, please do it for *all* data members.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+								case 7:
+									m_write_break_cb(1);
+									break;
+							}
+							break;
+
+						default:
+							new_keystroke(key_nr, (shift||shift_lock), control);
+					}
+					keystate[column_counter] |= key_mask;
```
cuavas, line 229: You’re better off just updating this to `column` after the loop rather than changing one bit at a time.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+				{
+					switch(column_counter)
+					{
+						case 0:
+							switch(row_counter)
+							{
+								case 0:
+								case 1:
+									shift = true;
+									break;
```
cuavas, line 192: On key-up, you’re expecting `shift` to contain two bit flags, but you’re treating it as a Boolean here.

src/mame/tandberg/tdv2100_kbd.h
```diff
@@ -0,0 +1,86 @@
...
+	void w_onlil(int state);
+	void w_carl(int state);
+	void w_errorl(int state);
+	void w_enql(int state);
+	void w_ackl(int state);
+	void w_nakl(int state);
+
+protected:
+	required_memory_region          m_keymap;
+	required_memory_region          m_keyparams;
```
cuavas, line 37: You should use `required_region_ptr<T>` for these.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+		{
+			// TODO: Keyboard interrupt for CPU interface
+		}
+	}
+}
+
+void tandberg_tdv2100_disp_logic_device::w_break(int state)
+{
+	break_key_held = state;
+}
```
cuavas, line 723: Can this result in the TxD output changing immediately?

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+			m_write_enql_cb(0);
+			m_write_ackl_cb(0);
+			m_write_nakl_cb(0);
+		}
+	}
+}
+
+void tandberg_tdv2100_disp_logic_device::w_linek(int state)
+{
+	if(state)
```
cuavas, line 742: You should guard against getting two calls in a row with the same value, as that isn’t a state change.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+		{
+			w_transk(1);
+		}
+		update_rs232_lamps();
+	}
+}
+
+void tandberg_tdv2100_disp_logic_device::w_transk(int state)
+{
+	if(state)
```
cuavas, line 790: Same here, you should be prepared for consecutive calls with the same value.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+			PORT_CONFSETTING(0x6, "6: 9600")
+			PORT_CONFSETTING(0x7, "7: 19200")
+
+	PORT_START("sw_rs232_settings")
+		PORT_CONFNAME(0x01, 0x01, "NO PARITY")              PORT_CONFSETTING(0x01, "YES") PORT_CONFSETTING(0x00, "NO")  PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+		PORT_CONFNAME(0x02, 0x02, "EVEN PARITY")            PORT_CONFSETTING(0x02, "YES") PORT_CONFSETTING(0x00, "NO")  PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+		PORT_CONFNAME(0x04, 0x04, "TWO STOP BITS")          PORT_CONFSETTING(0x04, "YES") PORT_CONFSETTING(0x00, "NO")  PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+		PORT_CONFNAME(0x08, 0x08, "EXT. ECHO")              PORT_CONFSETTING(0x08, "YES") PORT_CONFSETTING(0x00, "NO")
+		PORT_CONFNAME(0x10, 0x00, "AUTO RFS (RTS)")         PORT_CONFSETTING(0x10, "YES") PORT_CONFSETTING(0x00, "NO")
+		PORT_CONFNAME(0x20, 0x00, "AUTO DSR")               PORT_CONFSETTING(0x20, "YES") PORT_CONFSETTING(0x00, "NO")
```
cuavas, line 885: Please put one setting per line.  Also, can you use the default yes/no strings, and order the settings so no comes before yes, to give some kind of consistency?

How do the parity settings interact?  Turning on no parity at the same time as even parity doesn’t make a great deal of sense.  Should it be a 2-bit field so the four possible states can be labelled clearly?

What happens if you turn two stop bits off?  Would it be better to just call it stop bits, and then have the number of stop bits for the two setting values?

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+ioport_constructor tandberg_tdv2100_disp_logic_device::device_input_ports() const
+{
+	return INPUT_PORTS_NAME( tdv2115l );
+}
+
+ROM_START(tdv2115l)
+	ROM_REGION( 0x0800, "display_font_rom", ROMREGION_ERASEFF )
+	ROM_LOAD( "960558-0 1.82s141.u18", 0x0200, 0x0200, CRC(002856D2) SHA1(D86AFC7190163B3C0D7A43516ADB14046B978A4F))
+	ROM_LOAD( "960558-0 2.82s141.u19", 0x0400, 0x0200, CRC(422F1959) SHA1(AAD1B74ED194C486E53D7FE6469B15C407E08441))
+	ROM_LOAD( "960558-0 3.82s141.u20", 0x0600, 0x0200, CRC(F662DF8E) SHA1(C585731612AA1BEA8D95223A49011099953CF34E))
```
cuavas, line 956: Like other hexadecimal numbers, we use lowercase digits in CRCs and SHA1 digests.

src/mame/tandberg/tdv2115l.cpp
```diff
@@ -0,0 +1,74 @@
...
+	m_terminal->write_enql_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_enql));
+	m_terminal->write_ackl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_ackl));
+	m_terminal->write_nakl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_nakl));
+}
+
+void tdv2115l_state::machine_start()
+{}
+
+void tdv2115l_state::machine_reset()
+{}
```
cuavas, line 61: Just don’t override these if you aren’t going to do anything in them.

src/mame/tandberg/tdv2115l.cpp
```diff
@@ -0,0 +1,74 @@
...
+	m_keyboard->write_linek_callback().set(m_terminal, FUNC(tandberg_tdv2100_disp_logic_device::w_linek));
+	m_keyboard->write_transk_callback().set(m_terminal, FUNC(tandberg_tdv2100_disp_logic_device::w_transk));
+	m_keyboard->write_break_callback().set(m_terminal, FUNC(tandberg_tdv2100_disp_logic_device::w_break));
+	m_terminal->write_waitl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_waitl));
+	m_terminal->write_onlil_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_onlil));
+	m_terminal->write_carl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_carl));
+	m_terminal->write_errorl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_errorl));
+	m_terminal->write_enql_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_enql));
+	m_terminal->write_ackl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_ackl));
+	m_terminal->write_nakl_callback().set(m_keyboard, FUNC(tandberg_tdv2100_keyboard_device::w_nakl));
```
cuavas, line 54: Please group the device configuration with the device.  Callbacks support late binding – you don’t need to instantiate a device before using the object finder to configure another device.

src/mame/tandberg/tdv2100_kbd.h
Frodevan: Will fix. I thought m_ was just used when importing other drivers/modules, which is why I did not apply it to state variables.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: This is the intended behaviour, the original hardware also does this.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: I tried to be consistent with the printed labels on the back casing of the machine here, but I can change to default and more descriptive strings if that is preferred. Most of these settings were originally only described in detail through the user-manual, for instance see page 3: 

https://www.technikum29.de/de/geraete/Siemens_6610/Siemens_6610_PDFs/Siemens_6610_Wartungshandbuch_Teil_2_Einschaltung_Bruecken_und_Einstellungen.pdf

(The 6.610 is an OEM version of the TDV-2114, which uses the same terminal module as the TDV-2115. They both have most of the same switches exposed.)

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+///////////////////////////////////////////////////////////////////////////////
+//
+// General machine state
+//
+
+void tandberg_tdv2100_disp_logic_device::device_start()
+{
+	m_beep_trigger = timer_alloc(FUNC(tandberg_tdv2100_disp_logic_device::end_beep), this);
+	m_speed_ctrl = timer_alloc(FUNC(tandberg_tdv2100_disp_logic_device::expire_speed_check), this);
+}
```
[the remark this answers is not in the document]
Frodevan: I thought about it, but since this is the first driver I write from scratch I decided to add it at a later time. Can add a todo.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+	}
+}
+
+///////////////////////////////////////////////////////////////////////////////
+//
+// Graphics and character data
+//
+
+void tandberg_tdv2100_disp_logic_device::char_to_display(uint8_t byte){
+	if(cursor_y_input)
```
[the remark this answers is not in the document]
Frodevan: Will fix, missed this one.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+	{
+		// This is at the start of a char line
+		for (int char_nr_x = char_min_x; char_nr_x <= char_max_x; char_nr_x++)
+		{
+			// This is at the start of a single character
+			if(char_nr_x == 0 && char_nr_y == 0)
+			{
+				attribute = 0;
+				frame_counter++;
+			}
```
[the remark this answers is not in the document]
Frodevan: I forgot about this one, the current implementation here was made before I learned how to use the timers.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+
+			int const in_line_char_nr = char_nr_x - 4;
+			int const chr_x_pos = char_nr_x*9;
+			int const chr_y_pos = char_nr_y*14;
+			int const vram_address = in_line_char_nr + char_nr_y*80;    //NOTE: On real hardware, a pair of 4-bit PROM are used for this
+			bool const is_cursor = (cursor_x == in_line_char_nr && cursor_y == char_nr_y);
+			bool const blink_strobe = frame_counter&0x08;
+			bool const cursor_blink_strobe = blink_strobe && !speed_check;
```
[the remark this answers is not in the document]
Frodevan: Will investigate

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+					bool blank_attr_chars = !(m_dsw_u61->read()&0x040);
+					if(video_enable && (vram_data&0x60 || !blank_ctrl_chars) && !(vram_data&0x80 && blank_attr_chars))
+					{
+						data = m_font->base()[((vram_data&0x7F)<<4) + line];
+						bool extend_dot7 = ((m_dsw_u61->read()&0x003) == 1);
+						if(extend_dot7)
+						{
+							dot_nr8 = data&0x01;
+						}
+					}
```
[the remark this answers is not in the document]
Frodevan: I put these where used for better readability (a lot of these are only used once), but can move them out of the loop.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+
+					if(draw_cursor)
+					{
+						data = ~data;
+						dot_nr8 = ~dot_nr8;
+					}
+				}
+
+				bool full_inverse_video = !m_sw_invert_video->read();
+				if(full_inverse_video)
```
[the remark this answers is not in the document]
Frodevan: will fix.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,962 @@
...
+	int baud_setting = m_sw_rs232_baud->read();
+	int u46_divisor = (baud_setting == 0x0) ? 15 : 11;
+	int u22_divisor = 256>>baud_setting;
+	m_uart_clock->set_unscaled_clock(DOT_CLOCK_3 / (u46_divisor * u22_divisor));
+	m_uart->write_cs(true);
+	m_uart->write_nb1(m_dsw_u39->read()&0x200);
+	m_uart->write_nb2(true);
+	m_uart->write_np(m_sw_rs232_settings->read()&0x01);
+	m_uart->write_eps(m_sw_rs232_settings->read()&0x02);
+	m_uart->write_tsb(m_sw_rs232_settings->read()&0x04);
```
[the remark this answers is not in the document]
Frodevan: Will fix. My original idea was to try to pipe the callback state from the IO port bit directly into the write line, but could not get that to work.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Will fix

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Will fix

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Will fix

src/mame/tandberg/tdv2100_kbd.h
```diff
@@ -0,0 +1,86 @@
...
+	output_finder<>                 m_shiftlock_led;
+
+	virtual void device_add_mconfig(machine_config &config) override;
+	virtual void device_start() override;
+	virtual void device_reset() override;
+	virtual ioport_constructor device_input_ports() const override;
+	virtual const tiny_rom_entry *device_rom_region() const override;
+
+	void scan_next_column(int state);
+	void new_keystroke(uint8_t key_nr, bool shift = false, bool ctrl = false);
```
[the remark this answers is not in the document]
Frodevan: Will fix, this is a holdover from earlier

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+		PORT_BIT(0x08, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("DEL")                PORT_CODE(KEYCODE_DEL)          PORT_CHAR(UCHAR_MAMEKEY(DEL))
+		PORT_BIT(0x10, IP_ACTIVE_HIGH, IPT_UNUSED)
+		PORT_BIT(0x20, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("Z")                  PORT_CODE(KEYCODE_Z)            PORT_CHAR('z') PORT_CHAR('Z')
+		PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("F9")                 PORT_CODE(KEYCODE_F9)           PORT_CHAR(UCHAR_MAMEKEY(F9))
+		PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("LF")                 PORT_CODE(KEYCODE_ENTER)        PORT_CHAR(13)
+
+	PORT_START("X3")
+		PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("ER LINE")            PORT_CODE(KEYCODE_F11)          PORT_CHAR(UCHAR_MAMEKEY(F11))
+		PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("^ \xC2\xAC")         PORT_CODE(KEYCODE_TILDE)        PORT_CHAR('^') PORT_CHAR(0x00ac)
+		PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("\xC3\x85")           PORT_CODE(KEYCODE_OPENBRACE)    PORT_CHAR(0x00e5) PORT_CHAR(0x00c5)
```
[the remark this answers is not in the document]
Frodevan: In this case the PORT_CHAR is both cases, while the printing on the key is only upper-case. I was following an example in the Tiki-100 driver here, as that also has Norwegian characters on the keyboard.

src/mame/tandberg/tdv2115l.cpp
Frodevan: Will fix

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,555 @@
...
+	Two keyboards exist, one using switches and logic to derive keypresses,
+	the other based on an OEM design from KeyTronics, based on capacitive
+	switches and the 20-04592-013 30293E-013 chip by GI (AY-3-4592 derivative).
+	In both cases there is a translation ROM for deriving an ASCII character
+	for each keypress, and the Key-Tronics version also has a ROM for sorting
+	out key properties. For the logic-based keyboard all keys have the same
+	properties.
+
+	The logic keyboard toggles a relay in order to make a typing-sound, while
+	the Key-Tronics keyboard has a proper buzzer for this click.
```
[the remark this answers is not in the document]
Frodevan: Will fix

Frodevan:
Review comments have been addressed, here is a summary:

BREAK key:
Holding this intentionally pulls the TxD line low asynchronously to the UART clock by design. This is how it is implemented on the hardware, and also reflects how the break function was originally used in the terminals of the 60s and 70s. That is, forcing framing-errors in order to interrupt a process running on a timeshare mainframe. In turn, this was a re-use of a holdover from the days of telegraphy.

Key names:
I prefer to use the exact wording on the keycaps (if present), so the alphabetic keys still have name overrides.

Shift keys:
Added a comment to make it more clear what I am trying to do when scanning these. There is a check and intentional cast from int to bool when one of these keys are released.

Keyboard jumper-pair:
This has not been exposed, because these jumpers are tied to the keyboard layout and would otherwise be closely tied to the type of terminal module used. I can only speculate if the TDV-2115C ("Display Controller") version of the terminal, or maybe the TDV-2324, use 8-bit data for characters and the same keyboard pinout, but I can't know for sure since I have never had my hands on these. If need be, in the future, two sub-classes of the keyboard should be made: with and without the feature enabled.

Savestates:
Added. Key-states are not saved, since this is not relevant from the users perspective (won't be holding the same keys when restoring a savestate anyways), but the shift lock status is. Terminal display/cursor state along with the RS-232 flow-control state is also saved.

Screen:
Vertical trigger has been added, and finding the current active display attribute has been reworked to match this. This needed to be done at some point anyways, since the upcoming CPU interface will need a vsync interrupt.

Otherwise the rest of the comments have been addressed by pushing fixes.

PR #11822, opened 2023-12-08
emu/romload.cpp: add flag to reverse bits in each byte of a region

mgarlanger:
Adds a flag to ROM Region, to allow devices to specify reversing the bits in each byte during loading of the ROM. 

In two of the graphic add-ons for the H19/H89 systems, the pixel memory is reverse of the font data(imaginator already merged and SigmaSoft IGC currently in a PR - https://github.com/mamedev/mame/pull/11801). This requires reversing one or the other for every byte before the values can be ORed together. By allowing the font ROM to be loaded with the bits reversed, this is no longer needed. This also slightly simplifies the calculation for the pixel drawing of the font in other cases too (no longer need the `7 - b` operation to select the bit to draw).

### DIFF
--- a/src/emu/romentry.h
+++ b/src/emu/romentry.h
@@ -58,6 +58,10 @@ enum
 #define     ROMREGION_NOINVERT      0x00000000
 #define     ROMREGION_INVERT        0x00000800
 
+#define ROMREGION_REVERSEBITSMASK   0x00001000          /* reversed the bits in each byte of the region */
+#define     ROMREGION_NOREVERSEBITS 0x00000000
+#define     ROMREGION_REVERSEBITS   0x00001000
+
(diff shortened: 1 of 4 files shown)


galibert:
I have several problems with that PR, sorry:
- I don't see why it would be a region flag and not a rom load flag
- it forces the swap to be on bytes instead of taking the actual region data width into account
- it gives an inaccurate picture of the hardware, since I'm pretty sure the schematics, if any, will show the rom bits in the normal order and just have a left-shifting shifter in there.
- I doubt there's any performance difference in the end

I'm fascinated that the chargen and the graphics ram have inverted bit order in your system though...

mgarlanger:
I'll go ahead and close this PR and look at adding this to `device_start`.

PR #11823, opened 2023-12-08
upd7220: Sync command should enable/disable the DE bit

brijohn:
The SYNC command according to the upd7220 datasheet has a DE (Display Enable) on bit one of the command byte. This should be processed when receiving the command. 

This fixes LOGO professor on the Epson QX-10. That program during boot uses BCTRL to disable the display and then slightly later issues a SYNC command to re-enable the display. Without this fix the disable is never turned back on.

### DIFF
--- a/src/devices/video/upd7220.cpp
+++ b/src/devices/video/upd7220.cpp
@@ -1173,6 +1173,11 @@ void upd7220_device::process_fifo()
 		break;
 
 	case COMMAND_SYNC: /* sync format specify */
+		if (flag == FIFO_COMMAND)
+		{
+			m_de = m_cr & 1;
+		}
+
 		if (m_param_ptr == 9)
 		{
 			m_mode = m_pr[1];


rb6502:
Hi!  Have you done a sanity check with any of the other machines in MAME that use a uPD7220?  This looks entirely reasonable, but we've been bitten recently by similar changes to other devices.  As an example, if you have the software list sets, "mame dmv msdos211" should boot up MS-DOS on the NEC Decision Mate V.

brijohn:
> Hi! Have you done a sanity check with any of the other machines in MAME that use a uPD7220? This looks entirely reasonable, but we've been bitten recently by similar changes to other devices. As an example, if you have the software list sets, "mame dmv msdos211" should boot up MS-DOS on the NEC Decision Mate V. (As long as you see the text display the 7220 is working, that's all I'm asking for).

Yeah I did try some DMV software and text was being properly displayed.

cracyc:
Might be good to try the vt240 too which does an extensive test of the 7220 at start.

brijohn:
> Might be good to try the vt240 too which does an extensive test of the 7220 at start.

Tested booting up a VT240 and the screen works correctly.

PR #11824, opened 2023-12-08
cpu/xtensa.cpp, skeleton/hudson_poems.cpp: Added initial emulation for CPU core, and Poems hardware that utilizes it [David Haywood]

mamehaze:
xtensa.cpp
 - implemented an execution core with support for feature level required by hudson/konami PLAY-POEMS hardware
 - fixed a number of bugs in the disassembler

hudson_poems.cpp
 - got Marimba Tengoku to boot
 - reverse engineered many functions of the video hardware (4 tilemaps, sprites etc.) and created a preliminary implementation


rb6502:
This looks pretty reasonable at a first run-through.  Two general notes:

1) Search-and-replace the few instances of full types in the xtensa stuff (uint32_t) for the short u8/u16/u32 forms the majority of the code's using.

2) For pulling fields out of the opcode and not modifying them again (reg_s, reg_r, that kind of thing) make those variables const.  Every little bit we can help the compiler is good, especially in CPU cores.

mamehaze:
> This looks pretty reasonable at a first run-through. Two general notes:
> 
>     1. Search-and-replace the few instances of full types in the xtensa stuff (uint32_t) for the short u8/u16/u32 forms the majority of the code's using.
> 
>     2. For pulling fields out of the opcode and not modifying them again (reg_s, reg_r, that kind of thing) make those variables const.  Every little bit we can help the compiler is good, especially in CPU cores.

done and done

src/devices/cpu/xtensa/xtensa_tables.h
cuavas, line 1: You’re going to end up with two copies of all this stuff when it’s included in two files (although the compiler should be able to de-duplicate the strings themselves).  Can you make these static members of an `xtensa_cpu_utils` class or something so the contents of the constant tables and any functions that aren’t performance critical can go in a .cpp file?

src/devices/cpu/xtensa/xtensa.cpp
```diff
@@ -29,43 +46,2254 @@ std::unique_ptr<util::disasm_interface> xtensa_device::create_disassembler()
...
 {
-	return space_config_vector {
-		std::make_pair(AS_PROGRAM, &m_space_config),
+	space_config_vector spaces = {
+		std::make_pair(AS_PROGRAM, &m_space_config)
 	};
+
+	spaces.push_back(std::make_pair(AS_EXTREGS, &m_extregs_config));
+	return spaces;
+}
```
cuavas, line 55: Why not just put both the spaces in the return statement?  Like:
```c++
	return space_config_vector{
		std::make_pair(AS_PROGRAM, &m_space_config),
		std::make_pair(AS_EXTREGS, &m_extregs_config)
	};
```

src/mame/skeleton/hudson_poems.cpp
```diff
@@ -23,15 +23,27 @@
    2005/11/17   絶体絶命でんぢゃらすじーさん パーティーじゃっ!全員集合!!  (Mini-Game Collection)
    2005/11/24   ぐ〜チョコランタン スプーだいすき!プレイマット                (Kid's Floor Mat)
 
+   -------------
+
+   Marimba Tengoku has a secret test menue (unsure how to access it)
+   the ROM check code for that menu is at 0x661010 in ROM and does a full 32-bit word sum
+   of the ROM, comparing it against the value stored in the last 4 bytes (it passes)
```
cuavas, line 30: There’s a “menue” misspelling here.

src/mame/skeleton/hudson_poems.cpp
```diff
@@ -40,65 +52,666 @@ class hudson_poems_state : public driver_device
...
+		u16 spriteword2 = m_mainram[(spritebase + i * 2) + 1] & 0xffff;
+		u16 spriteword3 = m_mainram[(spritebase + i * 2) + 1] >> 16;
+
+		const int x = (spriteword3 & 0x03ff);
+		const int y = (spriteword2 & 0x03ff);
+		int tilenum = spriteword1 & 0x03ff;
+		const int pal = (spriteword2 & 0x7c00)>>10;
+
+		// is it selecting from multiple tile pages (which can have different bases?) (probably from a register somewhere)
+		tilebase = (m_spritegfxbase[(spriteword0 & 0x0300)>>8] & 0x0003ffff) / 32; // m_spritegfxbase contains a full memory address pointer to RAM
```
cuavas, line 207: Can you move the declaration of `tilebase` down to where it’s initialised, and make `spriteword*` `const` since they’re just used for extracting meaningful fields?

cuavas (commented):
This mostly looks fine.  Just a few relatively easy things and it’ll be mergeable.

src/devices/cpu/xtensa/xtensa_tables.h
mamehaze: done, although it does seem worse for performance

src/mame/skeleton/hudson_poems.cpp
mamehaze: done

src/mame/skeleton/hudson_poems.cpp
mamehaze: done

src/devices/cpu/xtensa/xtensa.cpp
mamehaze: done

src/devices/cpu/xtensa/xtensa_tables.h
cuavas: That’s primarily because you didn’t make the helper functions `static`.

PR #11825, opened 2023-12-08
nabupc: Update keyboard to use REVB ROM image

brijohn:
This updates the ROM used by the keyboard to use REVB. 

The REVB firmware fixes a bug in the keyboard's ROM that caused issues with using the second paddle on a controller port.

### DIFF
--- a/src/mame/misc/nabupc_kbd.cpp
+++ b/src/mame/misc/nabupc_kbd.cpp
@@ -57,7 +57,7 @@ namespace  {
 
 ROM_START(nabu_keyboard_rom)
 	ROM_REGION( 0x800, "mcu", 0 )
-	ROM_LOAD( "nabukeyboard-90020070-reva-2716.bin", 0x000, 0x800, CRC(eead3abc) SHA1(2f6ff63ca2f2ac90f3e03ef4f2b79883205e8a4e) )
+	ROM_LOAD( "nabukeyboard-90020070-revb-2716.bin", 0x000, 0x800, CRC(7d6ebc61) SHA1(5b0cb3da5b4afcae321c2a557e328c416b107dd7) )
 ROM_END
 
 //**************************************************************************


startaq:
You can add both ROMs using the bios system and make rev B the default. This will help people who dump their keyboard to easily identify their version.

brijohn:
> You can add both ROMs using the bios system and make rev B the default. This will help people who dump their keyboard to easily identify their version.

I had tried this originally but was unable to get the keyboard ROM to display in the BIOS selector.

If anyone dumps their keyboard ROM though it is not going to be a revision B unless they had previously flashed a new ROM since the revb is a recent patched version of the ROM I created and released to the community while working on getting paddles to work correctly on nabu hardware.

If someone can tell me why the BIOS selection does not work seem to work for the nabu keyboard I'd be happy to update the patch to support keyboard rom  selection. Below is what i tried but it like i said this does not seem to add this rom to the bios selection menu.
 
```
 ROM_START(nabu_keyboard_rom)
       ROM_REGION( 0x800, "mcu", 0 )
       ROM_DEFAULT_BIOS("revb")
       ROM_SYSTEM_BIOS( 0, "revb", "Keyboard Rev B" )
       ROMX_LOAD( "nabukeyboard-90020070-revb-2716.bin", 0x000, 0x800, CRC(7d6ebc61) SHA1(5b0cb3da5b4afcae321c2a557e328c416b107dd7), ROM_BIOS(0) )
       ROM_SYSTEM_BIOS( 1, "reva", "Keyboard Rev A" )
       ROMX_LOAD( "nabukeyboard-90020070-reva-2716.bin", 0x000, 0x800, CRC(eead3abc) SHA1(2f6ff63ca2f2ac90f3e03ef4f2b79883205e8a4e), ROM_BIOS(1) )
 ROM_END
```

startaq:
If this is something non-official then you really need to mark it that way, and you shouldn't make it look like something official. Call it `-fixed` maybe and add a comment what it fixes.

Even if the BIOS selection doesn't work the original manufacturer ROM should still be referenced to document it.

rb6502:
As far as I know the GUI BIOS selector only is for the top-level machine.  If this is a subdevice, you can change the BIOS from the command line but not in the on-screen menus.

happppp:
You mean this?:

mame.exe leonardo -exp analyst
-> Tab -> BIOS Selection -> works

rb6502:
I'd like to have this in somehow, but I'm not sure what the command line syntax would be for a non-slot device with a BIOS.  @cuavas any ideas on this one?

cuavas:
> I'd like to have this in somehow, but I'm not sure what the command line syntax would be for a non-slot device with a BIOS. @cuavas any ideas on this one?

There’s currently no way for the user to select the BIOS for a device that isn’t either the root system or a slot option.  Other devices can only have the BIOS changed from the default in machine configuration when they’re instantiated.

I agree with @startaq that the actual shipping ROM should be kept.  We don’t replace ROMs with user-created “updated” versions.  In this case, doing so would be against preservation as we’d lose the actual firmware the device shipped with, and we’d no longer be providing an accurate representation of an interesting market failure.

Calling it “revision B” when it’s just something you created yourself isn’t honest.

rb6502:
I agree that the default should be the as-shipped BIOS, but we also support modern/aftermarket upgrades for a lot of systems.  I guess in this case the best path would be a trivial clone system that just loads the other keyboard ROM?

cuavas:
May be easier to turn the keyboard port into a slot to avoid an entire clone?  I know it’s a workaround no matter which way it’s done, but I have a bunch of more pressing stuff to do than *another* options system rewrite to support these kinds of device options.

brijohn:
I can fairly easily turn the keyboard into a slot device using an RS232 port. I think that would be better then just cloning the entire system in order to allow selecting a keyboard BIOS. I'll put together an updated patch for this PR.

Issue #7402, opened 2020-10-27
CHD improved compression algorithm 


HashTang:
CHD is a streamable compression format for backup images developed by the mame project.
It's getting traction way beyond mame, and is quickly becoming the de facto compressed format for the entire emulation community.

CHD cuts the image into smaller blocks, compressed individually.
There are, as far as I know, 3 compression algorithms which are accessible : lzma (`cdlz`), zlib (`cdzl`), and flac (`cdfl`).
flac is specific to sound data, while `lzma` and `cdzl` are lossless.
`lzma` compresses better, but is also slower to decompress.

CHD utilities like `chdman` seem to highly value space, and therefore tend to lean towards `lzma`. That's fine, if one disregards decompression speed. But on many devices, notably ARM smartphones and tablets, decompression speed is a real hog. This basically leaves `zlib` as the only alternative.

More recently, a new compression algorithm, [`zstd`](https://github.com/facebook/zstd) has started to become mainstream. Granted it is especially successful in cloud center and server space, but not only. It's even documented as part of updated `zip` and `7zip` format specification, has a public [IETF RFC](https://tools.ietf.org/html/rfc8478), and can even be used to compress web traffic.

What makes `zstd` interesting with regards to CHD ?
To begin with, it has excellent decompression speed. As in ~x16 faster than `lzma`, and 3-4x to `zlib`.
One would expect to pay this speed in compression ratio, but that's effectively not the case : at its highest setting, `zstd` compresses within a few % of `lzma`, with small to negligible differences.
At the very least, compared to `zlib`, it's an all win, and a substantial one.

For these properties, `zstd` is already used in transparent compression for file systems, such as [`squashfs`](https://github.com/torvalds/linux/commit/87bf54bb43ddd385d2538b777324bf737f243042), which serve a similar use case as `CHD`.

High compression ratio, blazing fast decompression speed, could that be an interesting evolution of the `CHD` format ?

From a format perspective, it's probably not a huge deal : just an additional tag for a new format. I presume the current format already uses tags to distinguish between none, zlib, lzma and flac.
From an ecosystem perspective though, it can be trickier : it will require support from the decoder first, before any encoder can make use of the new feature.

cuavas:
It would also cause compatibility issues if people start distributing CHDs using a new compression algorithm that isn’t supported in the applications that are out there now, and add another third-party library as a hard dependency.  It’s not such a simple decision to add.

p1pkin:
[...]

PR #11827, opened 2023-12-09
Add Zstandard support for zip archives and CHDs

cuavas:
This does three things:
* Adds the Zstandard library in 3rdparty with an option to use the system-wide installation (dual BSD and GPL license).
* Adds support for zip archives members compressed with Zstandard (method 93).
* Adds support for Zstandard-compressed hunks in CHDs (`zstd` and `cdzs` codecs).

We occasionally get requests for Zstandard compression support (e.g. issue #7402).  It gives far better decompression performance than zlib or particularly LZMA while delivering better compression ratios than zlib.  This is important to people using emulators on low-end devices.  CHD decompression speed affects performance of emulated software that streams from hard disk or CD, and zip decompression performance affects initial load time for emulated systems.

By default, chdman will not use Zstandard compression.  This ensures that new CHDs will be backwards compatible if the user does not intentionally enable Zstandard compression.

You can create zip archives using Zstandard compression with WinZip, a fork of p7zip, and various other tools.


cuavas:
The main impact this has it that it bloats the source a bit.  Apart from that, it’s pretty much invisible to the user if they don’t consciously try to use it.

crashGG:
Finally, chd has added zstd support. If used properly, zstd decompression will exceed the speed of copying.

saulfabregwiivc:
This seems promising, maybe we can run these CHDs in Wii soon :)

PR #11829, opened 2023-12-09
handheld/hh_sm510.cpp: Added Nauchpribor Altair

algestam:
This PR adds a driver for a game/radiation dosimeter "Altair".

The unit is similar to the Elektronika `gnw_mmouse` clones but has a different ROM and inputs for reading dosimeter values. More info can be found here:
* https://rhbz.pw/thumbnails.php?album=223
* http://rr20.ddns.net/Item.aspx?ItemId=47363047-2f0c-4b76-8d68-202ae0634d7b&Lang=En

The dosimeter mode can be entered when the game is in time mode. Dosimeter mode is started when pressing the right up button. The port name has been changed in the MAME driver to reflect this.

Adding this as a draft as there are some things that needs to be worked out before this gets merged:
* The Altair unit is currently configured as a clone (child) of `gnw_mmouse`. Is this ok or should it be defined as separate driver perhaps?
* The "protocol" between the dosimeter reader and the game PCB are four inputs that allows incrementing the dosimeter reading with +1, +2, +4 and +8 when the input is set high for respective input. When the input is lowered, the remaining value up to 16 is incremented (i.e. if +2 was added when high, +14 gets added when lowered).

  The four dosimeter inputs are defined as separate configuration options. Could this be represented in a better way in MAME perhaps?


happppp:
It's the same game underneath so yes it can be a clone set, just like that game with the big UFO that added a different game mode.

For the sensor, use IPT_DIAL?
The only issue is that IPT_DIAL can go in reverse, and this one can't.

src/mame/handheld/hh_sm510.cpp
```diff
@@ -1626,6 +1659,11 @@ void gnw_mmouse_state::ecircus(machine_config &config)
 	kb1013vk12_common(config, 1657, 1080); // R mask option ?
 }
 
+void gnw_mmouse_state::naltair(machine_config &config)
+{
+	kb1013vk12_common(config, 1655, 1080); // R mask option ?
```
happppp, line 1664: I see it's the same SVG screen as vfutbol, then you can use the vfutbol mcfg instead of copy pasting it for naltair.

algestam:
Thanks for the feedback and suggestions @happppp :)

Have changed the dosimeter readings to use IPT_DIAL and it seems to work well. Also changed the game to reuse the naltair driver.

happppp:
Does the cheat still work, or did they remove it to make room?

algestam:
The cheat is still working.

cuavas:
That isn’t really a dosimeter – a dosimeter is a device used to measure/estimate cumulative exposure ionising radiation over an extended period of time.  They’re typically worn while working with radioactive materials and/or devices.  This thing is more like a scintillation counter.

cuavas:
Also @algestam, can you please give an indication of the system(s) added in pull request titles?

algestam:
> That isn’t really a dosimeter – a dosimeter is a device used to measure/estimate cumulative exposure ionising radiation over an extended period of time. They’re typically worn while working with radioactive materials and/or devices. This thing is more like a scintillation counter.

Agree, however the (semi-automatically translated) manual refers to the device as a "household dosimeter":
image

Let me know if you want me to change or clarify something regarding this in the driver.

cuavas:
The Russian actually does say “dosimeter”, so we should still refer to the button and mode as “dosimeter mode button”, “dosimeter mode”, etc. as that’s what it’s called.  However, in the comments, it’s probably better to be precise and say “…also included a radiation scintillation counter…” etc. when describing the hardware.

algestam:
Thanks for the feedback @cuavas , have pushed a change to the notes that clarifies that the device is a radiation scintillation counter while keeping the dosimeter mode for the inputs.

cuavas:
@happppp did you have any objections to this in its current form?

happppp:
Not really. There's still the issue that IPT_DIAL is 2-way and this thing is 1-way.
I tried, and this workaround would block the digital decrement input: PORT_CODE_DEC(INPUT_CODE_INVALID), but you can't do anything about the analog one.

cuavas:
You’d still need some way to reset it for subsequent measurements.

cuavas:
An option might be to have a paddle for setting the radiative flux, and using a timer to update the scintillation counter’s output?  To be at least somewhat realistic, the time between counts needs to be randomised though.

happppp:
It's not critical for emulation, I'm not sure it's important to simulate it accurately.
It reminds me of 'wtalker', where I just emulated the lightsensors instead of thinking of a way to input the weight.

PR #11837, opened 2023-12-13
new machines marked as NOT WORKING ( Lexibook JG7420 200-in-1 )

mamehaze:
----------------------
Lexibook JG7420 200-in-1 [David Haywood]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -44679,6 +44679,9 @@ tagalaga
 taspinv
 taturtf
 
+@source:tvgames/gpm4530a_lexibook_jg7420.cpp
+lx_jg7420
+
 @source:tvgames/magiceyes_pollux_vr3520f.cpp
 didj                            // (c) 2008 Leapfrog
 
--- a/src/mame/tvgames/gpm4530a_lexibook_jg7420.cpp
+++ b/src/mame/tvgames/gpm4530a_lexibook_jg7420.cpp
@@ -7,6 +7,7 @@
 
 #include "cpu/arm7/arm7.h"
 #include "cpu/arm7/arm7core.h"
+#include "machine/spi_sdcard.h"
 
 #include "screen.h"
 #include "speaker.h"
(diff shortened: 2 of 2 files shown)


mamehaze:
this was picked up to study the S+Core tech used by the JG7420FZ (Frozen edition) and JG7425 etc. but it turned out to be completely different hardware inside.  I haven't managed to identify the boot ROM, probably a tiny SPI chip or similar

cuavas (commented):
Could you add an `SPI_SDCARD` device to load the dump?  There are example uses in some of the enhanced Spectrum clones (tsconf, pentevo) and some Apple II and and Acorn slot devices.

mamehaze:
is that comment meant for this PR?

The SD card here is inside the device (I had to melt off the hot glue to remove it) it isn't really intended as any kind of user removable media, which that device seems designed for?

cuavas:
Yeah, it’s for this PR.  The CPU will be communicating with it over SPI, so it will need a fixed media SD card device to load it, in the same way that systems with a built-in hard disk (e.g. Killer Instict or SF Rush) have a fixed media hard disk device to load the disk image.

gatecat:
> this was picked up to study the S+Core tech used by the JG7420FZ (Frozen edition) and JG7425 etc. as I think those might have an internal ROM that needs extracting, but this turned out to be completely different hardware inside. I haven't managed to identify the boot ROM, probably a tiny SPI chip or similar

Yeah, they do, I think it lives at `0x9F000000`/`0xBF000000`. Incidentally, I've got an emulator for this system at https://github.com/gatecat/emu293 including some of the nattier parts like the descrambling built into the BLNDMA figured out and might be useful.

If the JG7420FZ is dumped, I'd love an SD card image (same for anything else SPG293 based).

---

As for this PR, in case you haven't spotted it already, an SDK for some kind of GPM4 SoC ended up on Github at https://github.com/daviskung/GPM4_MLX_0528/ including a bunch of register addresses at https://github.com/daviskung/GPM4_MLX_0528/blob/master/driver_l1/inc/drv_l1_sfr.h

The boot ROM appears to be loading `JumpMenu.bin` from the SD card to `01c00000` and the executing from the reset vector pointer at `01c00004`.

mamehaze:
There are a number of 293 systems dumped for which CHDs exist of the sd
cards in the Mame set, including the Frozen unit, Marvel unit and Zone3D. A
number of NAND based Jakks games using the tech are also dumped.  Can't
provide links here but they're easy to find

Wasn't aware of the emulator

mamehaze:
@gatecat what's the overall license on emu293 anyway? it seems to have various bits pulled from MAME in there, but I don't see an overall license which would allow me to use bits of it to further improve the emulation in MAME.

gatecat:
Oops, sorry for the delay in getting back, I've added a 3-clause BSD license for the everything that I have developed in there.

PR #11841, opened 2023-12-14
lib/formats/flacfile.cpp: Add support for cassette images in flac format.

wilbertpol:


ajrhacker:
Adding `ext_includedir("flac")` to formats.lua may fix the link error here.

ICEknigh7:
Thank you so much for implementing this!

wilbertpol:
> Adding `ext_includedir("flac")` to formats.lua may fix the link error here.

Thanks, that fixed the errors.

src/lib/formats/flacfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+	{
+		samples[channel] = std::make_unique<int16_t[]>(total_samples);
+		channel_samples[channel] = &samples[channel][0];
+	}
+	decoder.decode(channel_samples, decoder.total_samples(), false);
+	for (int channel = 0; channel < channels; channel++)
+	{
+		cassette_image::error err = cassette->put_samples(channel, 0.0,
+			(double)total_samples/decoder.sample_rate(), total_samples, 2,
+			channel_samples[channel], cassette_image::WAVEFORM_16BITLE);
```
cuavas, line 72: Does the decoder produce little Endian samples irrespective of host Endianness, or is it native word format?  I’d rather avoid introducing more Endianness bugs.

src/lib/formats/flacfile.cpp
cuavas, line 1: This code makes copies of the entire file into memory for identifying the format and decoding.  Would it be better to add an accessor to `cassette_image` to obtain a `util::random_read &` so you can seek it to zero and give it to the FLAC decoder?

src/lib/formats/flacfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include "flacfile.h"
+#include "flac.h"
+
+
+static constexpr int MAX_CHANNELS = 8;
+
+
+cassette_image::error flacfile_identify(cassette_image *cassette, cassette_image::Options *opts)
+{
```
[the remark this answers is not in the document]
wilbertpol: Made flacfile_identify static

src/lib/formats/flacfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+	std::vector<uint8_t> file_contents(file_size);
+	cassette->image_read(&file_contents[0], 0, file_size);
+	flac_decoder decoder(&file_contents[0], file_size);
+	decoder.reset();
+	const int channels = decoder.channels();
+	const int total_samples = decoder.total_samples();
+
+	std::unique_ptr<int16_t[]> samples[MAX_CHANNELS];
+	int16_t* channel_samples[MAX_CHANNELS];
+	for (int channel = 0; channel < channels; channel++)
```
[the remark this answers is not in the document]
wilbertpol: Fixed

src/lib/formats/flacfile.cpp
wilbertpol: Good point. Looking at the code in flac.cpp, the flac library code produces data in host endianness. I have changed the decode call to swap endianness when the host is big endian so the generated samples should always be little endian.

src/lib/formats/flacfile.cpp
wilbertpol: Yes, that would reduce some memory usage; updated the PR.

PR #11842, opened 2023-12-16
heathkit/tlb.cpp: Reverse bits in font roms to avoid doing it on scanout

mgarlanger:
Reversing the bits during device_start() avoids having to do it for every scan line when using the imaginator slot device, since the font bits are reverse from the pixel layout and one would be need be reversed. 
Reversing is also needed for the SigmaSoft IGC in this open PR - https://github.com/mamedev/mame/pull/11801 Depending on which PR is merged first, I will update the other PR to properly handle this.

### DIFF
--- a/src/mame/heathkit/tlb.cpp
+++ b/src/mame/heathkit/tlb.cpp
@@ -213,13 +213,10 @@ void heath_tlb_device::device_start()
 	m_key_click_timer = timer_alloc(FUNC(heath_tlb_device::key_click_off), this);
 	m_bell_timer = timer_alloc(FUNC(heath_tlb_device::bell_off), this);
 
-	// flip bits in font ROM to avoid having to flip them during scan out since some
+	// Flip bits in font ROM to avoid having to flip them during scan out since some
 	// boards with graphics has the pixel graphic reversed from the font layout. Doing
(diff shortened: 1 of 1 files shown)


src/mame/heathkit/tlb.cpp
```diff
@@ -212,6 +212,17 @@ void heath_tlb_device::device_start()
...
 	m_key_click_timer = timer_alloc(FUNC(heath_tlb_device::key_click_off), this);
 	m_bell_timer = timer_alloc(FUNC(heath_tlb_device::bell_off), this);
+
+	// flip bits in font ROM to avoid having to flip them during scan out since some
+	// boards with graphics has the pixel graphic reversed from the font layout. Doing
+	// it for all devices for consistency.
+	memory_region *region = device().memregion("chargen");
+	int font_rom_size = (region != nullptr)? region->bytes() : 0;
+
+	for(int i = 0; i < font_rom_size; i++)
```
cuavas, line 222: Since `m_p_chargen` is a `required_region_ptr`, you can’t get here if the memory region doesn’t exist.  You can also call `.length()` on it to get the length in elements, and you don’t need to call `.device()` to get `device_t` members in a `device_t` implementation.

PR #11843, opened 2023-12-16
Add new machine Tandberg TDV-2115L with accurate addressing in Display Logic driver

Frodevan:
The Tandberg TDV-2115L is a terminal based on the Tandberg Display Logic module, and is the most basic member of the TDV-2100 series of terminals and computer products from Tandberg.

Background: I recently dumped the addressing PROMs on the Tandberg TDV-2100 series Display Logic module, and the aim of this PR is to use this to add proper emulation of the vram addressing. As a result, a feature which was previously simulated (page roll), is now emulated the same way it is operating in hardware.

In addition to this, due to the more robust vram addressing, the conditional checks for the cursor have been changed to work like the real hardware.

Summary:
* Add vram address-encode PROM
* Properly emulate a roll-counter used for page up/down
* Increase accuracy of cursor positioning

The two additional ROMs can be found here:
https://github.com/Frodevan/Tandberg/tree/main/TDV-2100/ROM-Files/Loose%20cards/960310%201-19%20Display%202


cuavas:
You may as well close the older PR since all the work in it is in this PR anyway.  It will be less work to deal with it as a single unit.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+		m_cursor_col_input = false;
+	}
+	else
+	{
+		// Ordinary input
+		if((byte&0x7f) < 0x20 || (byte&0x7f) == 0x7f)
+		{
+			// Control characters
+			bool auto_roll_up = !(m_dsw_u39->read()&0x100);
+			bool clear_ascii_hs_by_cleark = ((m_dsw_u39->read()&0x003) == 0x001);
```
cuavas, line 211: You still aren’t thinking about avoiding unnecessary I/O port reads.  You’re reading the same port twice in a row here.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+	bool const cursor_blink_strobe = blink_strobe && !m_speed_check;
+
+	for(int char_row_nr = char_min_row; char_row_nr <= char_max_row; char_row_nr++)
+	{
+		// This is at the start of a char row
+		int const chr_y_pos = char_row_nr*14;
+		int attribute = get_attribute(char_row_nr, char_min_col);
+
+		for (int char_col_nr = char_min_col; char_col_nr <= char_max_col; char_col_nr++)
+		{
```
cuavas, line 494: This `for` statement has different spacing to the others.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+						data = m_font[((vram_data&0x7f)<<4) + line];
+						if(extend_dot7)
+						{
+							dot_nr8 = data&0x01;
+						}
+					}
+
+					// Pattern modifiers
+					bool underline_mode = m_dsw_u61->read()&0x008;
+					if(underline_mode)
```
cuavas, line 524: You’ve got a read of an I/O port on every line of every character here.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+
+uint32_t tandberg_tdv2100_disp_logic_device::screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect)
+{
+	rgb_t const *const palette = m_palette->palette()->entry_list_raw();
+
+	bool extend_dot7 = ((m_dsw_u61->read()&0x003) == 1);
+	bool blinking_cursor = ((m_dsw_u61->read()&0x030) == 0x010);
+	bool blank_attr_chars = !(m_dsw_u61->read()&0x040);
+	bool blank_ctrl_chars = !(m_dsw_u61->read()&0x100);
+	bool block_cursor = m_dsw_u61->read()&0x200;
```
cuavas, line 477: Please avoid redundant I/O port reads.  You can keep the value in a local.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+				}
+				if(full_inverse_video)
+				{
+					data = ~data;
+					dot_nr8 = ~dot_nr8;
+				}
+				for(int dot = 0; dot < 8; dot++)
+				{
+					bitmap.pix(chr_y_pos + line, chr_x_pos + dot) = palette[((data>>(7-dot))&0x01)<<extra_intensity];
+				}
```
cuavas, line 587: Please use the `BIT` helper for extracting individual bits – `BIT(data, 7 - dot)` is a lot easier to read than `((data>>(7-dot))&0x01)`.

src/mame/tandberg/tdv2100_disp_logic.h
```diff
@@ -0,0 +1,137 @@
...
+	auto write_nakl_callback() { return m_write_nakl_cb.bind(); }
+
+	DECLARE_INPUT_CHANGED_MEMBER(uart_changed);
+
+	// Input strobes
+	void process_keyboard_char(uint8_t key);
+	void w_cleark(int state);
+	void w_linek(int state);
+	void w_transk(int state);
+	void w_break(int state);
```
cuavas, line 44: Please follow MAME conventions – if you want to explicitly mark member functions as “write” methods, use a `_w` suffix.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+				case 0x0b:
+					if((m_cursor_row&0x18) != 24)
+					{
+						place_cursor(m_cursor_row+1, m_cursor_col);
+					}
+					break;
+
+				// LF, Move cursor to next line
+				case 0x0a:
+					if((m_cursor_row&0x18) == 24 && auto_roll_up)
```
mgarlanger, line 280: It would be clearer if you keep it in hex when doing comparisons, and most of the code has more white space which seems more readable to me. `if (((m_cursor_row & 0x18) == 0x18) && auto_roll_up)`

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+	PORT_START("sw_rs232_settings")
+		PORT_CONFNAME(0x01, 0x01, "NO PARITY  (No: Enable parity)")                 PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+			PORT_CONFSETTING(0x00, DEF_STR( No ))
+			PORT_CONFSETTING(0x01, DEF_STR( Yes ))
+		PORT_CONFNAME(0x02, 0x02, "EVEN PARITY (No: Odd parity, when enabled)")     PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+			PORT_CONFSETTING(0x00, DEF_STR( No ))
+			PORT_CONFSETTING(0x02, DEF_STR( Yes ))
+		PORT_CONFNAME(0x04, 0x04, "TWO STOP BITS (No: One stop bit)")               PORT_CHANGED_MEMBER(DEVICE_SELF, tandberg_tdv2100_disp_logic_device, uart_changed, 0)
+			PORT_CONFSETTING(0x00, DEF_STR( No ))
+			PORT_CONFSETTING(0x04, DEF_STR( Yes ))
```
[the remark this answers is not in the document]
Frodevan: This is a question about if we should be accurate to the hardware or accurate to the user-friendliness. I had the impression that Mame opted to be more on the side of being as much in line with the hardware as possible.

The names used here are literally what's printed on the switch-panel of the machine, including the "Yes/No" options. Only the content of the parenthesis is added for clarity. If I use more descriptive terms, the driver gets more user-friendly for the average user, but less preservative of the actual machine.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,563 @@
...
+	m_online_led.resolve();
+	m_carrier_led.resolve();
+	m_error_led.resolve();
+	m_enquiry_led.resolve();
+	m_ack_led.resolve();
+	m_nak_led.resolve();
+	m_wait_led.resolve();
+	m_shiftlock_led.resolve();
+
+	save_item(NAME(m_shift_lock));
```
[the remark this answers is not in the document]
Frodevan: There is nothing in this logic that is reliant upon strict timing of the scanning, as long as scanning happens, so I don't see how it is a problem. My point here is that the keyboard key-states should follow the user, and the state of the user's fingers is not saved with a savestate. Yes there may be a few milliseconds of non-deterministic delay before the current keyboard-state is reflected in the emulated machine at the instant of the load, but there is no way a user is going to notice this.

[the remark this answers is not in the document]
startaq: It's still good style to be as deterministic as possible. There might be no user visible change, but automatic regression testing might still see differences. The "price" is just a few additional `save_item` lines, so I don't see a big downside.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+	int const char_min_row = cliprect.min_y/14;
+	int const char_max_row = cliprect.max_y/14;
+	bool const blink_strobe = m_frame_counter&0x08;
+	bool const cursor_blink_strobe = blink_strobe && !m_speed_check;
+
+	for(int char_row_nr = char_min_row; char_row_nr <= char_max_row; char_row_nr++)
+	{
+		// This is at the start of a char row
+		int const chr_y_pos = char_row_nr*14;
+		int attribute = get_attribute(char_row_nr, char_min_col);
```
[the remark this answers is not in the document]
Frodevan: This is the simpler solution, but I can add and manage a starting-point for the scan if that's more desired.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Is there any performance penalty doing it this way, or is it more in terms of readability?

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Given what this logic is responsible of, I thought it would be more readable to have the it this way. 0x18 mask which lines are used for the comparison, 24 is the row-number the comparison targets. Of course the target uses all the lines of the mask by nature, but it can be useful to be able to directly read out which row is targeted without having to convert from hex.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,563 @@
...
+	m_8_bit_output(false)           // Hardwiered by PCB-trace jumpers 37-39 and 40-42, set at factory
+	                                //   true:
+	                                //     Use parameter PROM bit 2 to get which keys are inhibited
+	                                //     Replace TRANS key strobe with extra character data-bit
+	                                //   false:
+	                                //     Ignore parameter PROM bit 2
+	                                //     Use msb of char-map PROM to get which keys are inhibited
+{
+	std::fill(m_keystate, m_keystate+15, 0);
+}
```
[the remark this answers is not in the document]
Frodevan: Did not know std::begin and ::end worked with classic C-style arrays.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,563 @@
...
+		PORT_BIT(0x10, IP_ACTIVE_HIGH, IPT_UNUSED)
+		PORT_BIT(0x20, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("0 (NumPad)")         PORT_CODE(KEYCODE_0_PAD)        PORT_CHAR(UCHAR_MAMEKEY(0_PAD))
+		PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_KEYBOARD)                                    PORT_CODE(KEYCODE_F10)          PORT_CHAR(UCHAR_MAMEKEY(F10))
+		PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("CR")                 PORT_CODE(KEYCODE_BACKSPACE)    PORT_CHAR(10)
+
+	PORT_START("X5")
+		PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("(Blank 2)")
+		PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD)                                    PORT_CODE(KEYCODE_0)            PORT_CHAR('0') PORT_CHAR('_')
+		PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("O")                  PORT_CODE(KEYCODE_O)            PORT_CHAR('o') PORT_CHAR('O')
+		PORT_BIT(0x08, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("\xC3\x98")           PORT_CODE(KEYCODE_COLON)        PORT_CHAR(0x00f8) PORT_CHAR(0x00d8)     // Ø
```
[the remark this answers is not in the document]
Frodevan: UTF-8 strings is not something I have used in C++ before, but thanks for pointing me in the right direction here.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: This assumes BIT returns either 0 or 1, and will never by itself cast to bool in any foreseeable future. I was not able to find the definition used for this macro, but I can change it if we know for sure we have this under control.

mgarlanger: BIT is defined here: https://github.com/mamedev/mame/blob/master/src/lib/util/coretmpl.h#L584, it should always be a 0 or 1.

src/mame/tandberg/tdv2100_kbd.cpp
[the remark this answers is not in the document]
Frodevan: Ok, this should now have been sorted out.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1029 @@
...
+
+	bool force_on_line = m_dsw_u39->read()&0x020;
+	set_uart_state_from_switches();
+	if(force_on_line)
+	{
+		w_linek(1);
+		w_linek(0);
+		w_transk(1);
+		w_transk(0);
+	}
```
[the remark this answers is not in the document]
Frodevan: This has now been fixed.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Fixed.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Adjusted.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: This slipped under the radar, moved out of the loops now.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Ok, adjusted.

src/mame/tandberg/tdv2100_disp_logic.h
Frodevan: Fixed.

src/mame/tandberg/tdv2100_disp_logic.cpp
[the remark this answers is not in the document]
simzy39: Could you add a comment above explaining the wording is what's printed on the switch-panel so people know you didn't choose it?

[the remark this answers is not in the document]
Frodevan: That sounds like a good idea, added a note about this

Frodevan:
All review comments have now been addressed, so please mark any conversation as resolved if sufficiently fixed. Otherwise, please let me know so I can continue working on this.

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+//
+// Strobes
+//
+
+void tandberg_tdv2100_keyboard_device::key_trigger()
+{
+	m_write_kstr_cb(0x80|m_char_buffer);
+	m_write_kstr_cb(m_char_buffer);
```
cuavas, line 345: Is there potential for issues to be caused by the lack of any time between these two calls?  How long would the top bit be asserted for in reality?

src/mame/tandberg/tdv2100_kbd.cpp
```diff
@@ -0,0 +1,577 @@
...
+		PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("E")                  PORT_CODE(KEYCODE_E)            PORT_CHAR('e') PORT_CHAR('E')
+		PORT_BIT(0x08, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("F")                  PORT_CODE(KEYCODE_F)            PORT_CHAR('f') PORT_CHAR('F')
+		PORT_BIT(0x10, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("V")                  PORT_CODE(KEYCODE_V)            PORT_CHAR('v') PORT_CHAR('V')
+		PORT_BIT(0x20, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("4 (NumPad)")         PORT_CODE(KEYCODE_4_PAD)        PORT_CHAR(UCHAR_MAMEKEY(4_PAD))
+		PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_KEYBOARD)                                    PORT_CODE(KEYCODE_F6)           PORT_CHAR(UCHAR_MAMEKEY(F6))
+		PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("7 (NumPad)")         PORT_CODE(KEYCODE_7_PAD)        PORT_CHAR(UCHAR_MAMEKEY(7_PAD))
+
+	PORT_START("X12")
+		PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD)    PORT_NAME("NORM")               PORT_CODE(KEYCODE_RCONTROL)     PORT_CHAR(UCHAR_MAMEKEY(RCONTROL))
+		PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD)                                    PORT_CODE(KEYCODE_3)            PORT_CHAR('3') PORT_CHAR(0x00a3)
```
cuavas, line 531: Please put the character that’s specified by codepoint in a comment to make it clearer, since it isn’t anywhere else on the line.  (I think it’s a pound currency symbol £?  But normal people don’t memorise codepoints.)

src/mame/tandberg/tdv2100_kbd.h
```diff
@@ -0,0 +1,86 @@
...
+	output_finder<>                 m_wait_led;
+	output_finder<>                 m_shiftlock_led;
+
+	devcb_write8                    m_write_kstr_cb;
+	devcb_write_line                m_write_cleark_cb;
+	devcb_write_line                m_write_linek_cb;
+	devcb_write_line                m_write_transk_cb;
+	devcb_write_line                m_write_break_cb;
+
+	int m_column_counter;
```
cuavas, line 70: Since this is registered for save states, it should be an explicitly sized integer type for portability (e.g. `uint8_t` if it’s big enough).

src/mame/tandberg/tdv2100_disp_logic.h
```diff
@@ -0,0 +1,141 @@
...
+	devcb_write_line                    m_write_ackl_cb;
+	devcb_write_line                    m_write_nakl_cb;
+
+	// Video, Cursor & Sound
+	int m_frame_counter;
+	bool m_video_enable;
+	bool m_vblank_state;
+	int m_cursor_row;
+	int m_cursor_col;
+	int m_page_roll;
```
cuavas, line 115: Please use explicitly sized integer types for data members that need to be registered for save states for portability.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1063 @@
...
+				if(full_inverse_video)
+				{
+					data = ~data;
+					dot_nr8 = ~dot_nr8;
+				}
+				for(int dot = 0; dot < 8; dot++)
+				{
+					bitmap.pix(chr_y_pos + line, chr_x_pos + dot) = palette[BIT(data, 7 - dot)<<extra_intensity];
+				}
+				bitmap.pix(chr_y_pos + line, chr_x_pos + 8) = palette[BIT(dot_nr8, 0)<<extra_intensity];
```
cuavas, line 613: You’re better off avoiding the repeated calls to `bitmap.pix` here – the compiler may or may not optimise away the repeated line address calculation.  You can lift the invariant with something like:
```c++
				auto *pix = &bitmap.pix(chr_y_pos + line, chr_x_pos);
				for(int dot = 0; dot < 8; dot++)
				{
					*pix++ = palette[BIT(data, 7 - dot)<<extra_intensity];
				}
				*pix = palette[BIT(dot_nr8, 0)<<extra_intensity];
```

Alternatively, you could index the pixel pointer rather than incrementing it.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1063 @@
...
+		{
+			// TODO: Keyboard interrupt for CPU interface
+		}
+	}
+}
+
+void tandberg_tdv2100_disp_logic_device::break_w(int state)
+{
+	m_break_key_held = state;
+}
```
cuavas, line 872: If the break key generates a break condition on the RS232 TxD output, shouldn’t this immediately call `m_rs232->write_txd(0)` if the break input is asserted, and restore the output from the AY51013 if it’s deasserted?

With the way you’ve got it now, no break condition will be generated in response to the break key if the AY51013 doesn’t shift anything out.  Similarly, if the break key is released while a break condition is being generated, it won’t end until the AY51013 shifts something out.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1063 @@
...
+			{
+				break;
+			}
+		}
+	}
+	return attribute;
+}
+
+uint32_t tandberg_tdv2100_disp_logic_device::screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect)
+{
```
cuavas, line 490: For what it’s worth, because of the way this function always draws entire characters, it won’t work properly if something triggers a partial update on a screen line that isn’t the first row of a character.  The previous rows of the first line of characters intersecting the clipping rectangle will get redrawn.

If you *do* want to cater for cases like this, do you know how the display logic deals with attributes for each row of a line of characters?  For example does it latch the attributes when it starts drawing the first row of a line of characters, then reuse that value when it starts each subsequent row of the line?  Or does it have an line-sized attribute buffer it fills on the first row, or does it use some other approach?

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1063 @@
...
+	{
+		state = 1;
+	}
+	bool ext_echo = BIT(m_sw_rs232_settings->read(), 3);
+	if(!(ext_echo && m_data_terminal_ready))
+	{
+		state = (state && m_uart->so_r());
+	}
+	m_uart->write_si(state);
+}
```
cuavas, line 644: If `m_data_terminal_ready` or `m_rx_handshake` changes state or the external echo switch is flipped, shouldn’t that AY51013’s SI input be update immediately rather than waiting for the next state change of the external RxD input?

You’ll need to keep the input state so you can update the AY51013’s SI input when the other things that affect it change.

src/mame/tandberg/tdv2100_disp_logic.cpp
```diff
@@ -0,0 +1,1063 @@
...
+	update_rs232_lamps();
+}
+
+void tandberg_tdv2100_disp_logic_device::rs232_txd_w(int state)
+{
+	bool ext_echo = BIT(m_sw_rs232_settings->read(), 3);
+	if(!(ext_echo && m_data_terminal_ready))
+	{
+		m_uart->write_si(state);
+	}
```
cuavas, line 684: This doesn’t match with the logic in `rs232_rxd_w`.  In that function, you’re setting SI low if either the SO output is low and internal echo is enabled, or the RxD input is low.  This will set SI high if the SO output is low and internal echo is enabled even if the RxD input is low at that time.

src/mame/tandberg/tdv2100_disp_logic.cpp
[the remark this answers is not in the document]
cuavas: In general, we tend to prefer to make the setting labels as clear as possible in MAME.  You need to remember that the majority of people using MAME are not going to have the original manual at hand.  In many cases, we use standardised names for setting labels across systems in MAME rather than the wording used in original manuals or test modes (we have constants for standardised names for many settings that are common across arcade games in particular).

Here’s a somewhat extreme example: https://github.com/mamedev/mame/blob/mame0261/src/mame/bmc/bmcpokr.cpp#L834

The system displays the settings in Chinese, and attempting a literal translation would be an exercise in futility.  The actual displayed text is preserved in comments.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: The attribute is latched anywhere through the scan, and reset only on VSync (not per line). Any character on screen can store an attribute change, but with default DIP switches these will not be drawn.

So I have already accounted for this since each time screen_update is called, I scan from upper left (0,0) to get the attribute for the first character in the rectangle section before any drawing is done. The attribute variable is in other words not shared between individual calls to screen_update. In this way, overlapping draws will not mess up the attribute.

cuavas: You’re missing the point.  You’re drawing all the rows of pixels in a character in one go before moving to the next character horizontally.

It isn’t possible for the hardware to do this since the raster display scans an entire line of pixels before moving to the next line.  After drawing the first row of pixels in a character, it draws the first row of pixels in the next character to the right, and so on.  Then on the next line it draws the second row of pixels for each character in the same line.

For this to work, it needs to be able to restore the attribute state for the beginning of the line of characters each time it starts drawing a line of pixels that isn’t correspond to the first row of pixels in a line of characters.  Usually this is done either by latching the attribute state at the beginning of screen line corresponding to the first row of pixels in a line of characters, then restoring that state at the start of subsequent rows of pixels for the same line of characters.  However, other approaches are also possible, such as keeping a line width attribute buffer and filling it while drawing the first row of pixels in a line of characters.

> So I have already accounted for this since each time screen_update is called, I scan from upper left (0,0) to get the attribute for the first character in the rectangle section before any drawing is done.

Think about what happens when video RAM or registers change part way through a frame.  This is what causes visual artefacts like “tearing”.  The portion of the raster that has already been drawn will be unaffected, but the part that has yet to be drawn will use the new values (assuming the RAM/registers aren’t double-buffered).

In MAME, we deal with this using partial updates.  When something that affects video display changes, you trigger a partial update before setting the values so the parts of the screen that would be drawn before the change takes effect won’t see the change.

Consider what would happen on a partial update with your code as written now:
* The code draws all rows of any line of characters that intersects the clipping rectangle.  Supposed a partial update is triggered after three rows of a line of characters has been drawn.  With your current implementation, those three rows will be redrawn when the remaining rows of that line of characters are drawn.
* On every update, you scan attributes from the start of video RAM.  If attribute characters are changed in the part of the frame that has already been drawn in a partial update, they will still be applied in subsequent updates for the same frame.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: The RxD and SO+Echo are AND'ed together for SI, but this particular condition was not originally considered because this case is only valid if data-collision happens, and you are likely to get a framing error in both cases anyways.

But, I can add it.

src/mame/tandberg/tdv2100_kbd.cpp
Frodevan: Fixed

src/mame/tandberg/tdv2100_kbd.h
Frodevan: Fixed

src/mame/tandberg/tdv2100_disp_logic.h
Frodevan: Fixed

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Changed, although that sort of optimization does sound a little extreme to me. Then again, I am no expert about the inner workings of C++ compilers.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Good catch, fixed now.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: Some of the logic in the RS232 handling has now been streamlined, to not duplicate the same responsibility across multiple functions. This should be sorted out by this change.

src/mame/tandberg/tdv2100_disp_logic.cpp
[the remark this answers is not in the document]
Frodevan: Allright, if so be.

src/mame/tandberg/tdv2100_disp_logic.cpp
Frodevan: > For this to work, it needs to be able to restore the attribute state for the beginning of the line of characters each time it starts drawing a line of pixels that doesn’t correspond to the first row of pixels in a line of characters. Usually this is done either by latching the attribute state at the beginning of screen line corresponding to the first row of pixels in a line of characters, then restoring that state at the start of subsequent rows of pixels for the same line of characters.

Sorry for not going in too much detail, but this is exactly what the hardware does. The attribute is latched in a backup register at the end of the last line of each row, and recovered at the start of each line in the next row. See "Attribute Memory/Latch" on page 33/35:
https://www.technikum29.de/de/geraete/Siemens_6610/Siemens_6610_PDFs/Siemens_6610_Wartungskurs_Reg_I_Kap_0_Display_Logic_I_und_II.pdf

> Think about what happens when video RAM or registers change part way through a frame. This is what causes visual artefacts like “tearing”. The portion of the raster that has already been drawn will be unaffected, but the part that has yet to be drawn will use the new values (assuming the RAM/registers aren’t double-buffered).

The hardware double-buffers display RAM one row at a time, and halts any other access when the buffer is swapped out (also happens during last line of a row of characters). See "Refresh Memory", page 21/23 in the pdf linked to.

> On every update, you scan attributes from the start of video RAM. If attribute characters are changed in the part of the frame that has already been drawn in a partial update, they will still be applied in subsequent updates for the same frame.

The only way around this is to force a full redraw every time an attribute is changed.

Frodevan: Ok, I have made it so that it only scans attribute between the previous draw rectangle and the start of the current draw rectangle.

Frodevan:
@cuavas With the previous PRs I should have sorted out the RS-232 discrepancies, and altered the attribute scanning to make it consistently sequential no matter when or where a portion of the screen is drawn.

For the screen drawing algorithm, I responded with a detailed description of the hardware logic, and I would argue that the discrepancies is so minimal that it is negligible. The consequence of implementing this exactly as-is in the hardware is a lot of complications to the code, and the only differences in functionality would be at worst a 1-frame extra delay before a change would be visible. Other differences would be very minor conditional delays for certain operations, to handle collisions with the swap of the display buffer. Nothing that incorporated this module depended on that level of timing-accuracy.

cuavas (commented):
OK, I think this looks like most things are covered.  It could probably do with named constants for config port values, a few `int` changed to more appropriate types.  But I think it can be merged now.

PR #11845, opened 2023-12-17
lib/formats/flacfile.cpp: Remove unneeded calls to decoder.reset().

wilbertpol:


wilbertpol:
Oversight from my previous flac PR, reset() is already called by the invoked constructor. Calling FLAC__stream_decoder_init_stream again returned a warning status.

PR #11847, opened 2023-12-17
heathkit/h89.cpp: Implement single-step hardware

mgarlanger:
These changes implement the single-step hardware functionality in the H89. This is based the manual, and an h89 diagnostic program. It is now passing the tests:
single_step_fail
single_step_pass

### DIFF
--- a/src/mame/heathkit/h89.cpp
+++ b/src/mame/heathkit/h89.cpp
@@ -54,6 +54,14 @@
 
 namespace {
 
+// Single Step
+#define LOG_SS    (1U << 1)
+
+#define VERBOSE ( LOG_SS )
+#include "logmacro.h"
+
+#define LOGSS(...)    LOGMASKED(LOG_SS,    __VA_ARGS__)
+
 
 class h89_state : public driver_device
 {
@@ -81,7 +89,7 @@ public:
 
 private:
 
-	required_device<cpu_device> m_maincpu;
+	required_device<z80_device> m_maincpu;
 	required_memory_region m_maincpu_region;
 	memory_view m_mem_view;
 	required_device<ram_device> m_ram;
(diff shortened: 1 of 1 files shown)


src/mame/heathkit/h89.cpp
```diff
@@ -188,6 +207,45 @@ void h89_state::h89_mem(address_map &map)
 	m_mem_view[1](0x1800, 0x1fff).rom().region("maincpu", 0x1800).unmapw();
 }
 
+void h89_state::map_fetch(address_map &map)
+{
+	map(0x0000, 0xffff).lr8(NAME([this](offs_t offset)
+	{
+		return m1_r(offset);
+	}));
+}
```
cuavas, line 216: This seems overly convoluted – why can’t use use `map(0x0000, 0xffff).r(FUNC(h89_state::m1_r));` in the same way that e.g. src/mame/osborne/osborne1.cpp does?

src/mame/heathkit/h89.cpp
```diff
@@ -54,6 +54,14 @@
...
 namespace {
 
+// Single Step
+#define LOG_SS    (1U << 1)
+
+#define VERBOSE ( LOG_SS )
+#include "logmacro.h"
+
+#define LOGSS(...)    LOGMASKED(LOG_SS,    __VA_ARGS__)
+
```
cuavas, line 64: This is all macro glue – it goes outside the anonymous namespace.  Also, probably a good idea not to leave logging enabled here since it’s in a read handler and hence very hot.

src/mame/heathkit/h89.cpp
```diff
@@ -188,6 +207,45 @@ void h89_state::h89_mem(address_map &map)
...
+		return m1_r(offset);
+	}));
+}
+
+uint8_t h89_state::m1_r(offs_t offset)
+{
+	uint8_t data = m_program.read_byte(offset);
+
+	if (m_single_step_enabled && !m_556b_latch)
+	{
```
cuavas, line 223: Don’t forget to check `!machine().side_effects_disabled()` or this will cause chaos when debugger windows are open.

PR #11848, opened 2023-12-17
lib/util/flac.cpp: Add support for flac data where bits_per_sample != 16.

wilbertpol:


ICEknigh7:
Seems to work fine with the 24 bit FLACs I've been trying on the Spectrum core (even better than WAV support, actually).
image image
The only "problem" I've found is that it's only possible to play the Left channel (akin to loading from a Stereo source, rather than the recommended Monaural which would do the audio sum of both channels), but I suppose that's likely caused by something in the drivers themselves and not because of the FLAC decoding.

wilbertpol:
> The only "problem" I've found is that it's only possible to play the Left channel (akin to loading from a Stereo source, rather than the recommended Monaural which would do the audio sum of both channels), but I suppose that's likely caused by something in the drivers themselves and not because of the FLAC decoding.


Partially; the driver needs to do a set_stereo() on the cassette device (currently there seem to be no drivers using this). And it also needs this small change:
```
--- a/src/devices/imagedev/cassette.cpp
+++ b/src/devices/imagedev/cassette.cpp
@@ -452,7 +452,7 @@ void cassette_image_device::sound_stream_update(sound_stream &stream, std::vecto
 
                for (int ch = 0; ch < outputs.size(); ch++)
                {
-                       cassette->get_samples(0, time_index, duration, outputs[0].samples(), 2, &m_samples[0], cassette_image::WAVEFORM_16BIT);
+                       cassette->get_samples(ch, time_index, duration, outputs[0].samples(), 2, &m_samples[0], cassette_image::WAVEFORM_16BIT);
                        for (int sampindex = 0; sampindex < outputs[ch].samples(); sampindex++)
                                outputs[ch].put_int(sampindex, m_samples[sampindex], 32768);
                }
```
That fixed channels for flacs but gave me some weird echoing results on wavs before; it could be that I tested it wrongly though.

wilbertpol:
It also works with 8 bit flacs. For example, abcstack from the abc80_cass software list converted to flac. This never worked for me before.

src/lib/util/flac.cpp
```diff
@@ -628,11 +627,23 @@ FLAC__StreamDecoderWriteStatus flac_decoder::write_callback(const ::FLAC__Frame
...
 	}
 	return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
 }
 
+int16_t flac_decoder::convert_to_int16(const ::FLAC__Frame *frame, FLAC__int32 sample)
+{
+	if (m_uncompressed_swap)
+	{
+		sample = FLAC__int32(((uint32_t(sample) >> 24) & 0x000000ff) | ((uint32_t(sample) >> 16) & 0x0000ff00) | ((uint32_t(sample) >> 8) & 0x00ff0000) | (uint32_t(sample) & 0xff000000));
+	}
```
cuavas, line 640: You could use `swapendian_int32` from osdcomn.h

src/lib/util/flac.cpp
```diff
@@ -628,11 +627,23 @@ FLAC__StreamDecoderWriteStatus flac_decoder::write_callback(const ::FLAC__Frame
...
 			for (int chan = 0; chan < frame->header.channels; chan++)
 				if (m_uncompressed_start[chan] != nullptr)
-					m_uncompressed_start[chan][m_uncompressed_offset] = int16_t((uint16_t(buffer[chan][sampnum]) << shift) | (uint16_t(buffer[chan][sampnum]) >> shift));
+					m_uncompressed_start[chan][m_uncompressed_offset] = convert_to_int16(frame, buffer[chan][sampnum]);
 	}
 	return FLAC__STREAM_DECODER_WRITE_STATUS_CONTINUE;
 }
 
+int16_t flac_decoder::convert_to_int16(const ::FLAC__Frame *frame, FLAC__int32 sample)
+{
```
cuavas, line 636: Is there a reason this isn’t `const` qualified?  It shouldn’t change decoder state.

src/lib/util/flac.cpp
```diff
@@ -628,11 +627,23 @@ FLAC__StreamDecoderWriteStatus flac_decoder::write_callback(const ::FLAC__Frame
...
+	if (m_uncompressed_swap)
+	{
+		sample = FLAC__int32(((uint32_t(sample) >> 24) & 0x000000ff) | ((uint32_t(sample) >> 16) & 0x0000ff00) | ((uint32_t(sample) >> 8) & 0x00ff0000) | (uint32_t(sample) & 0xff000000));
+	}
+	if (frame->header.bits_per_sample >= 16)
+		return int16_t(uint32_t(sample >> (frame->header.bits_per_sample - 16)));
+	else
+		return int16_t(uint32_t(sample << (16 - frame->header.bits_per_sample)));
```
cuavas, line 644: You should round towards even when reducing resolution, and you should splat right when increasing resolution.

cuavas, line 644: This won’t work if the output format isn’t native Endianness.  You’re converting to output format *before* trying to adjust resolution.  Do this on paper and see what happens.

src/lib/util/flac.cpp
wilbertpol: No, made it const.

src/lib/util/flac.cpp
wilbertpol: Yes, I had the operations swapped.

cuavas (changes requested):
The trouble with this is it’s a massive pessimisation.  Remember this is going to be run for tens of thousands of samples at the very least.  The current code is written with pipeline performance in mind – shifts are pre-calculated before the loop so the actual format conversion in the loop doesn’t require any conditional code.

With this change, tests based on input and output formats are being done for every sample.

Please think about how you can restructure the code to lift as much conditional code out of the loop as possible, and to make the code in the loop work with conditional moves rather than requiring more expensive branches.

Also, that doesn’t round towards even – it rounds towards positive infinity.

wilbertpol:
I have moved most of the old conditional code outside the loop by using templating.
I did have to introduce another loop to generically handle scaling up to 16 bits. Assuming only 8 bits, although simpler, may be an over-simplification, in theory there could be frames with 4 or 9 bits per sample.

PR #11851, opened 2023-12-18
floppy.cpp: Update class variables to the current encouraged naming convention.

mgarlanger:
Updated all member variables to be prefixed with `m_`, the existing code had a few with the prefix but most without. This made it more difficult to follow the code.


mgarlanger:
@galibert Can you review this? The file has a mix of class members with and without the `m_` prefix. The floppy_image_device mainly has it without it, but has `m_fs`, `m_fs_managers`, `m_make_sound`, `m_sound_out`, `m_flux_per_pixel_infos`, `m_flux_per_combined_track_infos`, and `m_flux_screen`. Also most of the rest of the file is using the prefix.

PR #11855, opened 2023-12-18
sega/model3.cpp: Add "Daytona USA 2" games subtitles

clawgrip:


clawgrip:
Also... I's say that "Export" on games descriptions should be lower case ("export"). Please, tell me if it's OK to change it to lower case and I'll add it to this PR

mikehyrus:
To be honest, it’s been some time since I last helped update a few game titles but anyway, good catch there.

angelosa:
imho Export should be capitalized, it indicates a macro continent group that is "non-USA & non-Japan" (*) in Sega terms.

(*) or at least I think it is, not relevant if I'm right/wrong/if there are specific exceptions.

clawgrip:
> imho Export should be capitalized, it indicates a macro continent group that is "non-USA & non-Japan" (*) in Sega terms.

Ok, I'll left the PR as is, thx.

PR #11857, opened 2023-12-20
Nimbus joystick fix

RobinSergeant:
This pull request fixes RM Nimbus joystick emulation by using the correct port address for directional data and setting the correct bits when that address is read.  Currently the mouse data port (0xa4) is used for all joystick data, but only the buttons are actually read from this port.  The joystick port (0xa0) is used for the directional data as described in chapter 2 of of the service manual (https://www.thenimbus.co.uk/upgrades-and-maintenance/service-manual).

This can be demonstrated/tested by running the following RM Basic program which does not work without my change:

```
10 SET MODE 40
20 SET JOYSTICK
30 REPEAT
40   ASK JOYSTICK X%, Y%, Button%
50   POINTS X%, Y%
60 UNTIL Button% = 1
```

I have also removed the "Input Port 0 Device" config port as this is no longer necessary and prevented the mouse buttons from working properly when set to the wrong value (the original code read joystick data when this was to "mouse", and mouse data when it was set to "joystick").  The mouse and joystick can now co-exist quite happily with joystick buttons doubling as mouse buttons and vice versa.

The wrong bits were also being set for each direction and I have of course corrected these to match those given in the service manual.  The above RM Basic program proves that theses are now correct.

The original machine also allowed switching between two joysticks by writing to ports 0xa0 (to select joystick 0) and 0xa2 (to select joystick 1), but I have not implemented this as it seems unnecessary - I have not found any software that makes use of it and I don't think it's possible to test using RM Basic. 

Please let me know if you have any questions!

### DIFF
--- a/src/mame/rm/rmnimbus_m.cpp
+++ b/src/mame/rm/rmnimbus_m.cpp
@@ -118,13 +118,13 @@ chdman createhd -o ST125N.chd -chs 41921,1,1 -ss 512
 #define MOUSE_INT_ENABLE        0x08
 #define PC8031_INT_ENABLE       0x10
 
-#define MOUSE_NONE      0x00
-#define MOUSE_LEFT      0x01
-#define MOUSE_RIGHT     0x02
-#define MOUSE_DOWN      0x04
-#define MOUSE_UP        0x08
-#define MOUSE_LBUTTON   0x10
-#define MOUSE_RBUTTON   0x20
+#define CONTROLLER_NONE      	0x00
+#define CONTROLLER_LEFT      	0x01
+#define CONTROLLER_RIGHT     	0x02
+#define CONTROLLER_DOWN      	0x04
+#define CONTROLLER_UP        	0x08
+#define CONTROLLER_BUTTON0  	0x10
+#define CONTROLLER_BUTTON1		0x20
 
 // Frequency in Hz to poll for mouse movement.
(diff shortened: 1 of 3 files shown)


src/mame/rm/rmnimbus_m.cpp
```diff
@@ -1075,6 +1072,28 @@ void rmnimbus_state::mouse_js_reset()
...
+	{
+		// when the stick is pushed right the left bit must also be set!
+		result |= CONTROLLER_LEFT;
+	}
+
+	if (result & CONTROLLER_UP)
+	{
+		// when the stick is pushed up the down bit must also be set!
+		result |= CONTROLLER_DOWN;
+	}
```
cuavas, line 1092: This seems like a very odd design decision, and at first glance it would seem to require more logic than just mapping one direction to each bit.  Do you know why it has this quirk of always setting the left bit for horizontal directions and always setting the down bit for vertical directions?

RobinSergeant: I'm guessing it's related to the physical hardware design, but it's actually stated in the service manual. They have a table saying something like "right = left + right" :-) Maybe it's really a bit to say the X axis is effected and then a bit for positive/negative.

It threw me at first because without setting the extra bits only left and down worked. Then I noticed the cryptic table in the service manual.

All 8 directions work with my RM basic test app so fairly confident it's correct now!

PR #11859, opened 2023-12-20
taito/gladiatr.cpp: Use set_format for palette format, Minor cleanups

cam900:
Remove VIDEO_START_MEMBER and MACHINE_RESET_MEMBER, use video_start() and machine_reset() overrides instead. Reduce runtime tag lookups
Rename gfx regions for consistent.

### DIFF
--- a/src/mame/taito/gladiatr.cpp
+++ b/src/mame/taito/gladiatr.cpp
@@ -203,10 +203,10 @@ TODO:
 #include <algorithm>
 
 
-MACHINE_RESET_MEMBER(gladiatr_state,gladiator)
+void gladiatr_state::machine_reset()
 {
 	// 6809 bank memory set
-	membank("bank2")->set_entry(0);
+	m_adpcmbank->set_entry(0);
 	m_audiocpu->reset();
 }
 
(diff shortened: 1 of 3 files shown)


src/mame/taito/gladiatr.cpp
```diff
@@ -231,21 +231,21 @@ void gladiatr_state_base::ym_irq(int state)
 void gladiatr_state::gladiator_adpcm_w(u8 data)
 {
 	// bit 6 = bank offset
-	membank("bank2")->set_entry((data & 0x40) ? 1 : 0);
+	m_adpcmbank->set_entry(BIT(data, 6) ? 1 : 0);
```
cuavas, line 234: This is somewhat redundant now – if you’re using the `BIT` helper, you can just do `m_adpcmbank->set_entry(BIT(data, 6));`

src/mame/taito/gladiatr.cpp
```diff
@@ -231,21 +231,21 @@ void gladiatr_state_base::ym_irq(int state)
...
 	m_msm->reset_w(BIT(data, 5));   // bit 5
-	m_msm->vclk_w (BIT(data, 4));   // bit 4
+	m_msm->vclk_w(BIT(data, 4));    // bit 4
 }
 
 void ppking_state::ppking_adpcm_w(u8 data)
 {
 	// bit 6 = bank offset
-	//membank("bank2")->set_entry((data & 0x40) ? 1 : 0);
+	//m_adpcmbank->set_entry(BIT(data, 6) ? 1 : 0);
```
cuavas, line 244: Same here – the `BIT` helper extracts the bit so it will produce 1 or 0 anyway.

PR #11861, opened 2023-12-21
igs/goldstar.cpp: Added missing PROM and PLDs for 'hamhouse' [Taksangs]

clawgrip:
Redumped all the ROMs with a different programmer, and found out that U24 is bitrotten (replaced the current dump with the best one we got, marking it as BAD_DUMP).

Also recorded a video from the real PCB: https://youtu.be/qiSw044Twdc

### DIFF
--- a/src/mame/igs/goldstar.cpp
+++ b/src/mame/igs/goldstar.cpp
@@ -12938,6 +12938,7 @@ ROM_END
    -1 x Jfc 95101
    -3 banks of 8 DIP switches
    -12.000 MHz xtal
+  Video from the real PCB: https://youtu.be/qiSw044Twdc
 */
 ROM_START( hamhouse )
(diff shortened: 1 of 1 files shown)


Osso13:
TBH it seems to me the colors are pretty much spot on in MAME after Ioannis' fixes:

clawgrip:
You mean using MACHINE_IMPERFECT_COLORS instead or remove any COLORS flag?

Thx!

Osso13:
My attention to detail isn't really so good to make a call about this, but looking at this captures from the YouTube link you posted they seem pretty similar to the snaps I took from current MAME and posted above
immagine
immagine
I probably wouldn't add any flag, honestly.

clawgrip:
Fixed, thanks a lot!!!!

PR #11862, opened 2023-12-21
sega/sega_ferie.cpp: Add new not working system

qufb:
Basic functionality is implemented. I haven't tried most features of this system, which is why it's marked MACHINE_NOT_WORKING.

Some changes done in Z80 base class were required to support T6A84's multiple address spaces / memory pages:

* Stack reads and writes are handled by dedicated virtual functions. In this case, they are overridden to use a distinct address space;
* For interrupts, I considered using daisy-chain, but this would require even more changes to the base class. Additionally to fetching interrupt vectors, we need to execute interrupt routines in a distinct page. This seems easier to support by adding a new "return from interrupt" callback to handle switching mapped pages;

### DIFF
--- a/src/devices/cpu/z80/t6a84.cpp
+++ b/src/devices/cpu/z80/t6a84.cpp
@@ -1,0 +1,221 @@
+// license:BSD-3-Clause
+// copyright-holders:QUFB
+/***************************************************************************
+
+    Toshiba T6A84, TLCS-Z80 ASSP Family
+
+***************************************************************************/
+
+#include "emu.h"
+#include "t6a84.h"
+
+#define LOG_PAGE_R (1U << 1)
+#define LOG_PAGE_W (1U << 2)
+#define LOG_MEM    (1U << 3)
+
+//#define VERBOSE (LOG_PAGE_R | LOG_PAGE_W | LOG_MEM)
+#include "logmacro.h"
+
+
+DEFINE_DEVICE_TYPE(T6A84, t6a84_device, "t6a84", "Toshiba T6A84")
+
+void t6a84_device::internal_io_map(address_map &map) const
+{
+	map.global_mask(0xff);
+	map(0xfc, 0xfc).rw(FUNC(t6a84_device::stack_page_r), FUNC(t6a84_device::stack_page_w));
+	map(0xfd, 0xfd).rw(FUNC(t6a84_device::data_page_r), FUNC(t6a84_device::data_page_w));
+	map(0xfe, 0xfe).rw(FUNC(t6a84_device::code_page_r), FUNC(t6a84_device::code_page_w));
+	map(0xff, 0xff).rw(FUNC(t6a84_device::vector_page_r), FUNC(t6a84_device::vector_page_w));
+}
+
+t6a84_device::t6a84_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+	t6a84_device(mconfig, T6A84, tag, owner, clock, address_map_constructor(FUNC(t6a84_device::internal_io_map), this))
+{ }
+
+t6a84_device::t6a84_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, address_map_constructor io_map)
+	: z80_device(mconfig, type, tag, owner, clock)
+	, m_program_space_config("program", ENDIANNESS_LITTLE, 8, 20, 0, 16, 0)
+	, m_data_space_config("data", ENDIANNESS_LITTLE, 8, 20, 0, 16, 0)
+	, m_stack_space_config("stack", ENDIANNESS_LITTLE, 8, 20, 0, 16, 0)
+	, m_io_space_config("io", ENDIANNESS_LITTLE, 8, 16, 0, io_map)
+	, m_code_page(0)
+	, m_delay_code_page(0)
+	, m_is_delay_code_page_set(false)
+	, m_prev_code_page(0)
+	, m_data_page(8)
+	, m_stack_page(8)
+	, m_vector_page(0)
+{
+	// Interrupt vectors need to be fetched and executed from their corresponding page.
+	// For simplicity, we switch pages via callbacks, instead of using a dedicated address space.
+	irqfetch_cb().set([this](int state) {
+		LOGMASKED(LOG_PAGE_W, "IRQ FETCH %02x => %02x\n", m_code_page, m_vector_page);
+		m_prev_code_page = m_code_page;
+		m_code_page = m_vector_page;
+	});
+	reti_cb().set([this](int state) {
+		LOGMASKED(LOG_PAGE_W, "IRQ RET %02x => %02x\n", m_code_page, m_prev_code_page);
+		m_code_page = m_prev_code_page;
+	});
+	branch_cb().set([this](int state) {
+		LOGMASKED(LOG_PAGE_W, "BRANCH %02x => %02x\n", m_code_page, m_prev_code_page);
+		/*
+		    When setting a code page, it only becomes effective after jumping to a far address in that page.
+		    Any instructions fetched and executed before that jump still use the previous code page.
+		    This can be seen in Sega Ferie Kitten, when test program at page 7 gets mapped, as we are still
+		    executing on page 0, but we expect to start executing that program when jumping to RST0:
+
+		    ROM_00::1ea9 3e 07      LD   A,0x7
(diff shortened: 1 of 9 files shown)


src/devices/cpu/z80/t6a84.cpp
```diff
@@ -0,0 +1,227 @@
...
+	// For simplicity, we switch pages via callbacks, instead of using a dedicated address space.
+	irqack_cb().set([this](int state) {
+		LOGMASKED(LOG_PAGE_W, "IRQ ACK %02x => %02x\n", m_code_page, m_vector_page);
+		m_prev_code_page = m_code_page;
+		m_code_page = m_vector_page;
+	});
+	reti_cb().set([this](int state) {
+		LOGMASKED(LOG_PAGE_W, "IRQ RET %02x => %02x\n", m_code_page, m_prev_code_page);
+		m_code_page = m_prev_code_page;
+	});
```
cuavas, line 60: This isn’t really how you’re supposed to use these.  They’re intended for systems that detect interrupt acknowledge and return from interrupt cycles to use.  If you use them internally, the system can’t connect anything to them.

Does the Z80 core provide a way to customise the interrupt handling for derived classes?

src/devices/cpu/z80/t6a84.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+void t6a84_device::device_start()
+{
+	z80_device::device_start();
+
+	space(AS_PROGRAM).invalidate_caches(read_or_write::READWRITE);
+	space(AS_DATA).invalidate_caches(read_or_write::READWRITE);
```
cuavas, line 68: It doesn’t really make sense to call `invalidate_caches` here – they’re invalidated automatically when mappings change.

src/devices/cpu/z80/t6a84.cpp
```diff
@@ -0,0 +1,227 @@
...
+	    executing on page 0, but we expect to start executing that program when jumping to RST0:
+
+	    ROM_00::1ea9 3e 07      LD   A,0x7
+	    ROM_00::1eab d3 fe      OUT  (DAT_io_00fe),A
+	    ROM_00::1ead c3 00 00   JP   RST0
+	*/
+	if (!machine().side_effects_disabled() && m_is_delay_code_page_set && (address < m_delay_code_pc || address > (m_delay_code_pc + 0x20))) {
+		m_code_page = m_delay_code_page;
+		m_is_delay_code_page_set = false;
+	}
```
cuavas, line 146: Doing this on address translation seems wrong – it should happen on the jump instruction.  Does the Z80 base class allow customising handling for instructions?

src/mame/layout/ferie.lay
```diff
@@ -0,0 +1,110 @@
...
+						<line />
+					</g>
+				</svg>
+			]]></data>
+		</image>
+	</element>
+
+	<!--
+	    FIXME: Default UI font doesn't render Hiragana/Katakana.
+
```
cuavas, line 26: The default UI font is platform-dependent.  The font situation on Linux is a bit of a mess.  On macOS, due to the way MAME draws text, automatic font substitution for missing characters doesn’t happen.  However, on Windows it should be able to display kana by default.

Text rendering needs to be properly cleaned up eventually.

src/mame/sega/sega_ferie.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+	m_maincpu->space(AS_IO).install_read_handler(0xf4, 0xf4,
+			read8smo_delegate(*this, NAME([this] () { return m_rtc; })));
+
+	m_maincpu->space(AS_IO).install_read_handler(0xf5, 0xf5,
+			read8smo_delegate(*this, NAME([this] () { return m_io_buttons->read() & 0x3f; })));
+	m_maincpu->space(AS_IO).install_write_handler(0xf6, 0xf6,
+			write8smo_delegate(*this, FUNC(sega_ferie_state::tablet_ctrl_w)));
+	m_maincpu->space(AS_IO).install_read_handler(0xf7, 0xf7,
+			read8smo_delegate(*this, FUNC(sega_ferie_state::tablet_data_r)));
```
cuavas, line 269: Why do you need to install all this stuff here rather than using configured address maps?

cuavas (commented):
I’m concerned about the Z80 changes here.  Particularly in light of the ZX Spectrum Next changes, it seems like we’re hitting the limits of the Z80 code’s current architecture.  It needs to allow more flexibility for derived classes to customise instruction behaviour.

src/devices/cpu/z80/t6a84.cpp
qufb: The closest is `z80_daisy_chain_interface` which has callbacks for both interrupt ack and reti. But it also makes assumptions about the system architecture that I'm not sure about. We don't know if T6A84 has components using daisy-chaining, maybe we can add a separate callback for interrupt ack for now, just like I did here with `reti_cb`?

src/devices/cpu/z80/t6a84.cpp
qufb: I didn't see any, but your suggestion makes sense. Maybe we can add a callback for branching instructions?

src/mame/layout/ferie.lay
qufb: Ok, I'll remove this comment then.

src/mame/sega/sega_ferie.cpp
qufb: Not needed, I assumed it would override what I had defined on the processor side, but they actually get merged. I've refactored the IO space, should I try the same for the others?

qufb:
@cuavas Besides the instruction callbacks, is there anything else here that I still need to address?

simzy39:
@qufb Maybe this page is wrong, but it referrs to the two pet ones as 'Cats' and Dogs'. Maybe someone who reads Japanese can verify if it's that or 'puppy' and 'kitten':
https://segaretro.org/Ferie

qufb:
Romanization of "フェリエパピー" is "Ferie Papī", so "puppy" seems more accurate than "dogs". I haven't found any box or ads where the original model is referred as anything other than "フェリエ" or "Ferie". I mention "kitten" in the driver for the sake of distinguishing it from other models.

FoxhackDN:
I can confirm that this is likely a romanization of "Puppy." Particularly since the box... has dogs in it.

The Ferie box with cats in it doesn't have any word for cats anywhere. So that's probably the original model, and the puppy one came later?

qufb:
> The Ferie box with cats in it doesn't have any word for cats anywhere. So that's probably the original model, and the puppy one came later?

Likely the case, given these points of reference:
* Copyright date on box
    * [Kitten has 1994](https://pbs.twimg.com/media/DHa-JM6VYAAGp9t?format=jpg&name=large)
    * [Puppy has 1995](https://cdnyauction.buyee.jp/images.auctions.yahoo.co.jp/image/dr000/auc0305/users/7a724f9f0272a18c0772507e46864500e75af868/i-img1122x972-16527617910smipq216754.jpg);
* Default system date during setup:
    *  Kitten has 1994-10-01;
    * Puppy has 1995-05-01;
    * World Travel has 1995-06-01;
* Back case copyright date:
    * Kitten has 1994 (model "A"), 1994,95 (model "C");
    * Puppy has 1994,95 (model "A" and "E");
    * World Travel has 1995 (model "A" and "C");

If we consider the model naming to be chronologically consistent, then the earliest back case date for Puppy also has 1995. I was a bit surprised to see them use chip-on-board for Puppy, only to later on switch back to packages in World Travel... but they might have done it only for late production runs, independent from the actual release dates.

simzy39:
Is this ready to merge?

qufb:
All Z80 core extensions are now using dedicated callbacks. If this should still be reconsidered, we can follow up on [this discussion](https://github.com/orgs/mamedev/discussions/105).

holub:
@qufb Can you try to actuate this this current master? After new z80 core merge this one suspected to be non compilable.

qufb:
Updated. Anything else that still needs to be reviewed?

Tafoid:
@qufb It appears you did send data for this PR back in December of last year but it was never secured and the link provided no longer works.  It would be helpful if you could send the Kitten and World Travel data again as the Puppy one was secured when it was originally sent.  You can send a link or the data directly in an email as you have previously done and I'll be sure to look for it.

Email detailed here if you have forgotten:

Thank you very much for your continued contributions to the MAME project.

PR #11867, opened 2023-12-22
add support for 2nd RM Nimbus joystick

RobinSergeant:
This PR adds support for a second joystick on the RM Nimbus.  I decided to do this after discovering the RM Basic demo program that covers joystick support.  It's basically a two player paint program with two brushes (each controlled by a different joystick).  To run the demo start RM Basic and type

```
load "mousedem"
run
```

(then wait until it asks about having a joystick before pressing a key to begin).

Interestingly this demo makes it obvious that each joystick should actually only have one button!  This fits with the description of using an Atari style joystick, given in the service manual, as those joysticks only had one button.  Hence the button data returned via the mouse data port actually relates to the button status of both joysticks.

Questions and comments welcome.

### DIFF
--- a/src/mame/rm/rmnimbus.h
+++ b/src/mame/rm/rmnimbus.h
@@ -41,6 +41,7 @@
 /* Mouse / Joystick */
 
 #define JOYSTICK0_TAG           "joystick0"
+#define JOYSTICK1_TAG           "joystick1"
 #define MOUSE_BUTTON_TAG        "mousebtn"
 #define MOUSEX_TAG              "mousex"
 #define MOUSEY_TAG              "mousey"
@@ -76,6 +77,7 @@ public:
 		m_z80sio(*this, Z80SIO_TAG),
 		m_screen(*this, "screen"),
 		m_io_joystick0(*this, JOYSTICK0_TAG),
+		m_io_joystick1(*this, JOYSTICK1_TAG),
 		m_io_mouse_button(*this, MOUSE_BUTTON_TAG),
 		m_io_mousex(*this, MOUSEX_TAG),
 		m_io_mousey(*this, MOUSEY_TAG)
@@ -103,9 +105,11 @@ private:
 	required_device<z80sio_device> m_z80sio;
 	required_device<screen_device> m_screen;
 	required_ioport m_io_joystick0;
+	required_ioport m_io_joystick1;
 	required_ioport m_io_mouse_button;
 	required_ioport m_io_mousex;
 	required_ioport m_io_mousey;
+	ioport_port* m_io_selected_js;
 
 	bitmap_ind16 m_video_mem;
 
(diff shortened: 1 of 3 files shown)


src/mame/rm/rmnimbus.h
```diff
@@ -103,9 +105,11 @@ class rmnimbus_state : public driver_device
 	required_device<z80sio_device> m_z80sio;
 	required_device<screen_device> m_screen;
 	required_ioport m_io_joystick0;
+	required_ioport m_io_joystick1;
 	required_ioport m_io_mouse_button;
 	required_ioport m_io_mousex;
 	required_ioport m_io_mousey;
+	ioport_port* m_io_selected_js;
```
cuavas, line 112: Using a pointer makes it harder to implement save state support.  Please use a `required_ioport_array<2>` for the joystick ports and keep an index for the selected port.

src/mame/rm/rmnimbus.h
```diff
@@ -149,6 +153,8 @@ class rmnimbus_state : public driver_device
 	void nimbus_sound_ay8910_porta_w(uint8_t data);
 	void nimbus_sound_ay8910_portb_w(uint8_t data);
 	uint8_t nimbus_joystick_r();
+	void nimbus_select_joystick0(uint8_t data);
+	void nimbus_select_joystick1(uint8_t data);
```
cuavas, line 157: Do you actually need two of these?  Couldn’t you have a single write handler across the range, and get the selected joystick from the offset?

cuavas (changes requested):

src/mame/rm/rmnimbus.h
RobinSergeant: Thank you, I will do that.

src/mame/rm/rmnimbus.h
RobinSergeant: I will consider this while making the other change. For readability two functions might still be cleaner as there are only two.

RobinSergeant: After trying it out I think you are right so have made this change with another commit!

RobinSergeant: @cuavas, one thing I've noticed is that writing to address 0xa1 and 0xa3 now also changes the selected joystick.  Do you think that is a problem?  I could change the data type to uint8_t and then check for an even offset etc.

rb6502: You can either change to uint8_t like that or add the mem_mask parameter and use the byte lane selection tests like `if (ACCESSING_BITS_0_7)` / `if (ACCESSING_BITS_8_15)`.  The uint8_t is probably a little cleaner if ports 0xa1/0xa3 aren't actually used.

RobinSergeant: Thanks for the info.  0xa1/0xa3 are not used so I've pushed another commit with the unit8_t solution so that only even addresses are used.  I think everything should be fine now.

PR #11868, opened 2023-12-22
Improve interrupt handing in mb88xx

hackbar:
This deals with the interrupt handler a little better, and fixes Arabian bugs caused by mb88xx bugs.

I tested it with the other games that use that chip, and didn't see any regressions. Most importantly, none in bosco sound. The 54xx in bosco seem especially sensitive to interrupt races, causing the pitch of shots to be wrong.

Should fix https://mametesters.org/view.php?id=8436 & https://mametesters.org/view.php?id=3916


cuavas:
Can some people give this a spin?  This is an area that’s been fraught with issues over the years.

PR #11869, opened 2023-12-23
neogeo/neogeo_v.cpp: Make palette byte access behave the same as on real hardware

zardam:
On real hardware, a byte written in palette RAM is repeated in the whole word (CPU UDS/LDS signals seems to be ignored for palette access). This is used by some software to detect emulation.

### DIFF
--- a/src/mame/neogeo/neogeo_v.cpp
+++ b/src/mame/neogeo/neogeo_v.cpp
@@ -97,7 +97,7 @@ uint16_t neogeo_base_state::paletteram_r(offs_t offset)
 void neogeo_base_state::paletteram_w(offs_t offset, uint16_t data, uint16_t mem_mask)
 {
 	offset += m_palette_bank;
-	data = COMBINE_DATA(&m_paletteram[offset]);
+	m_paletteram[offset] = data;
 
 	int dark = data >> 15;
 	int r = ((data >> 14) & 0x1) | ((data >> 7) & 0x1e);


happppp:
Or is it like video_register_w/r where accessing the LSB only is not mapped, accessing the MSB only stores same data in MSB and LSB?

rb6502:
This is known as "byte smearing", and MAME's 68000 core already does it (byte-wide writes put the same byte on both byte lanes of the 16-bit data bus).  So this should be able to be simplified to:

		case 0xff00:
 			data = (data & 0xff00) | (data & 0x00ff);
 			break;
 		case 0x00ff:
 			data = (data & 0xff00) | (data & 0x00ff);
 			break;

zardam:
I tried writing bytes manually to palette RAM using unibios. Writing to even or odd addresses gives the same result, the value is repeated in the two bytes of the word.

I guess that for 8 bits access the CPU just repeats the byte on D0-D7 and D8-D15, and asserts UDS/LDS according to the address parity. UDS/LDS seems to be ignored for palette access, and maybe not for video register ? I did not find the information on the exact behavior of the CPU data bus in this case.

rb6502:
Actually never mind, you should be able to delete the switch statement entirely and just let the m_paletteram[offset] do the work.

zardam:
@rb6502, that is equivalent to not masking at all data. You're right !

PR #11870, opened 2023-12-23
hng64.cpp: sams64 Korean title inclusion

mikehyrus:
Samurai Shodown 64 has the Koreanized Chinese title, Pae Wang Jeon Seol 64.


mikehyrus:
Just a nice little refresh after a few months.

PR #11872, opened 2023-12-23
Holtek HT1130 CPU core, and preliminary driver for a brick game using it

mamehaze:
-----------
Brick Game 96 in 1 (E-23 Plus Mark II) [azya52, David Haywood]


mamehaze:
there is an emulator at https://github.com/ilyakurdyukov/brickgame-4bit  (with decapped ROM)
I haven't used the emulator at all as a) it's python and b) it's GPLv3

dumping research is at https://habr.com/ru/articles/773040/

disassembler implementation is based on https://pdf.dzsc.com/88889/5606.pdf CPU manual / instruction set reference.

happppp:
When the disassembler is functional, it can be added to unidasm too (src/tools/unidasm.cpp).

This is the game? https://image.konsolenkost.de/item/images/1022336/full/Specials-Hand-Held-Super-E23-Plus-Mark-2-Brick-G-a.jpg
You didn't include the actual title "Brick Game" in the title string. E-23 is not the game title, it's the model, right?
something like?: Brick Game: 96 in 1 (E-23 Plus Mark II)

mamehaze:
Yes, that thing, I'll change the title later.  There are hundreds of these things (many on different techs)

unfortunately it looks like sound isn't going to be possible unless somebody finds that on the die too, because the manual seems to indicate there are 16 sounds, with 'sound library' and the first 12 can have 32 notes, while the rest can have 64.  Sounds to me like there should be another melody ROM area like the later G&W titles.

happppp:
Yeah, sound handling isn't documented at low level in the datasheet.
If more HT1130 LCD games (there are plenty) are dumped, hh_ht1130 would be a better driver? So the common code could be shared between the games.

happppp:
Internal layout? No, this needs an SVG screen.
You could spend a few hours making a .lay file with 7segs and text labels, and a placeholder for the fruit icons. It may be useful for testing but is it worth the effort? Just buy a cheap one and request to someone experienced to scan the LCD?

mamehaze:
yes, I saw there was an SVG, but as mentioned, I'm not using anything from that emulator (as it seems to be GPLv3) so for now I'll just make the internal stuff more functional, until such a time as a fresh scan is available.

mamehaze:
Anyway I'm done with this until such a time that somebody can do an SVG and hook it up (that is beyond my ability)

It works as a proof of concept that the CPU core is running as  it should, and the basic internal layout lets you see the main pit, and control the pieces.

Hopefully there won't be too many issues preventing it from being merged

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -170,7 +170,8 @@ void ht1130_device::internal_data_map(address_map &map)
 {
 	map(0x00, 0x9f).ram().w(FUNC(ht1130_device::tempram_w)).share("tempram");
 
-	// the ht1130 is meant to have displayram from 0xe0, so this could be a newer chip?
+	// the HT1130 is meant to have displayram from 0xe0, so this could be a newer chip?
+	// HT1190 has data 00-9f and display b0-ff, possibly a closer match?
```
happppp, line 174: Sean when he read that Russian article today: "The pad locations don’t match those shown in the HT1130 datasheet.  They do match HT1190."

mamehaze: yeah, it seems logical as it doesn't seem to be using a0-af either, let me adjust things for that

mamehaze: right, I've set it to HT1190.  the CPU core folder is still called 1130, but there's a derived class for the 1190

src/mame/layout/brke23p2.lay
```diff
@@ -0,0 +1,256 @@
...
+	<element name="segment">
+		<rect state ="0">
+			<bounds x="0" y="0" width="7" height="7" />
+			<color red="1.0" green="1.0" blue="1.0" />
+		</rect>
+		<rect state ="1">
+			<bounds x="0" y="0" width="7" height="7" />
+			<color red="0.2" green="0.2" blue="0.2" />
+		</rect>
+	</element>
```
cuavas, line 15: You don’t need the `bounds` here at all – the element’s internal coordinate system is not related to the coordinate system of the view that uses it.

src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -0,0 +1,118 @@
...
+	required_ioport m_in2;
+
+	void display_offset_w(u8 data);
+	void display_data_w(u8 data);
+
+	u8 port_pm_r();
+	u8 port_ps_r();
+	u8 port_pp_r();
+
+	u8 m_displayoffset = 0;
```
cuavas, line 48: Please group data members and member functions within a section.

src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void ht11xx_brickgame_state::ht11xx_brickgame(machine_config &config)
+{
+	HT1190(config, m_maincpu, 1000000/8); // frequency?
+	m_maincpu->display_offset_out_cb().set(FUNC(ht11xx_brickgame_state::display_offset_w));
+	m_maincpu->display_data_out_cb().set(FUNC(ht11xx_brickgame_state::display_data_w));
+
+	m_maincpu->ps_in_cb().set(FUNC(ht11xx_brickgame_state::port_ps_r));
+	m_maincpu->pp_in_cb().set(FUNC(ht11xx_brickgame_state::port_pp_r));
```
cuavas, line 103: You should be able to use `set_ioport` here and avoid the trampolines.

src/devices/cpu/ht1130/ht1130d.h
```diff
@@ -0,0 +1,29 @@
...
+protected:
+	// disassembler overrides
+	virtual u32 opcode_alignment() const override;
+	virtual offs_t disassemble(std::ostream &stream, offs_t pc, const data_buffer &opcodes, const data_buffer &params) override;
+
+	const char* m_regnames[8] =
+	{
+		"R0", "R1", "R2", "R3",
+		"R4",
+	};
```
cuavas, line 25: This doesn’t need to be an instance member since it’s constant and it should be possible for the loader to resolve it.  Please throw it in an anonymous namespace in the .cpp file, and make the pointers `const` as well as what they point to.

src/devices/cpu/ht1130/ht1130d.cpp
```diff
@@ -0,0 +1,668 @@
...
+		stream << "XOR [R1R0],A";
+		return 1;
+	}
+
+	//// Opcodes with XH Immediates
+
+	case 0b01000000: // (with 4-bit immediate) : ADD A,XH : Add immediate data to the accumulator
+	{
+		u8 oprand = opcodes.r8(pc + 1);
+		if (!(oprand & 0xf0))
```
cuavas, line 328: Did you mean “operand”?

src/devices/cpu/ht1130/ht1130d.cpp
```diff
@@ -0,0 +1,668 @@
...
+		{
+			util::stream_format(stream, "<ill %02x %02x>", inst, oprand);
+		}
+		return 2;
+	}
+
+	case 0b01000010: // (with 4-bit immediate) : AND A,XH : Logical AND immediate data to accumulator
+	{
+		u8 oprand = opcodes.r8(pc + 1);
+		if (!(oprand & 0xf0))
```
cuavas, line 342: Seems to be a spelling error here as well, and the rest of the way down.

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,127 @@
...
+	devcb_write8 m_display_data_out;
+
+};
+
+class ht1190_device : public ht1130_device
+{
+public:
+	ht1190_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
+
+	void internal_data_map_ht1190(address_map &map);
```
cuavas, line 119: I’d expect this address map function should be protected or private?

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,127 @@
...
+
+	virtual std::unique_ptr<util::disasm_interface> create_disassembler() override;
+
+	virtual space_config_vector memory_space_config() const override;
+
+	void internal_map(address_map &map);
+	void internal_data_map(address_map &map);
+
+	inline void tempram_w(offs_t offset, u8 data);
+	inline void displayram_w(offs_t offset, u8 data);
```
cuavas, line 53: It’s generally not a great idea to put the `inline` qualifier on member function declarations.  It’s better to put it on the definitions.

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,127 @@
...
+	};
+
+	auto pm_in_cb() { return m_port_in_pm.bind(); }
+	auto ps_in_cb() { return m_port_in_ps.bind(); }
+	auto pp_in_cb() { return m_port_in_pp.bind(); }
+
+	auto pa_out_cb() { return m_port_out_pa.bind(); }
+
+	auto display_offset_out_cb() { return m_display_offset_out.bind(); }
+	auto display_data_out_cb() { return m_display_data_out.bind(); }
```
cuavas, line 34: Is this the best way of exposing this?  The chip stores the state of the 4*32 LCD segments internally, and provides a fixed duty cycle output to drive an LCD panel directly.  It doesn’t expose the internal RAM layout.

But even if you want to expose display RAM writes, you can use a single `devcb_write8` and call it with the offset and data.  The device callbacks all support offset and mask, even if people don’t use them a lot of the time.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+{
+	const u8 dataaddress = getr3r2();
+	return space(AS_DATA).read_byte(dataaddress);
+}
+
+void ht1130_device::setr3r2_data(u8 data)
+{
+	const u8 dataaddress = getr3r2();
+	space(AS_DATA).write_byte(dataaddress, data);
+}
```
cuavas, line 179: You really should be using a `memory_access<...>::specific` on the data space rather than going through the `space` on every access.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+{
+	map(0x00, 0x7f).ram().w(FUNC(ht1130_device::tempram_w)).share("tempram");
+	map(0xe0, 0xff).ram().w(FUNC(ht1130_device::displayram_w)).share("displayram");
+}
+
+void ht1190_device::internal_data_map_ht1190(address_map &map)
+{
+	map(0x00, 0x9f).ram().w(FUNC(ht1190_device::tempram_w)).share("tempram");
+	map(0xb0, 0xff).ram().w(FUNC(ht1190_device::displayram_w)).share("displayram");
+}
```
cuavas, line 192: You can use object finders in place of literal tags.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+
+void ht1130_device::device_start()
+{
+	space(AS_PROGRAM).cache(m_cache);
+	space(AS_PROGRAM).specific(m_space);
```
cuavas, line 197: Why do you have both a `cache` and a `specific` here?  You should choose the one that’s likely to perform best and use it.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+	m_stackcarry = 0;
+}
+
+u8 ht1130_device::fetch()
+{
+	m_icount--; // single byte opcodes take a single cycle, double take 2
+	if (m_timer_en)
+		m_timer++;
+	return m_space.read_byte(m_pc++);
+}
```
cuavas, line 251: Since you’ve got instructions that take multiple cycles and you’re making no provisions for returning to the scheduler midway through the instruction, you need to override `execute_max_cycles` and return the maximum number of cycles an instruction can take.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+		return;
+	}
+
+	//// Opcodes with XH Immediates
+
+	case 0b01000000: // (with 4-bit immediate) : ADD A,XH : Add immediate data to the accumulator
+	{
+		const u8 oprand = fetch() & 0x0f;
+		u8 acc = getacc();
+		acc += oprand;
```
cuavas, line 613: You seem to have spelled “operand” incorrectly in this file, too.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1051 @@
...
+		else
+		{
+			m_icount--;
+		}
+
+		if (m_timer & 0x100)
+		{
+			m_timer -= 0x100;
+			m_timerover = 1; // can also generate an interrupt
+		}
```
cuavas, line 1045: Is this going to work properly with multi-cycle instructions?  As far as I can tell, the timer counts clock cycles, not instructions.

cuavas (changes requested):

src/devices/cpu/ht1130/ht1130.cpp
mamehaze: My understanding of the datasheet is that any 2 byte instruction takes 2 cycles, and the timer is cycle based.  we increase timer when a byte is fetched, if it has overflowed by the time it reached here we set flag.

I believe it is correct, assuming my understanding is correct.

cuavas: Oh, this is the timer overflow check, not the timer tick.

src/devices/cpu/ht1130/ht1130.h
mamehaze: what's the reference for this? I'll change it, but I get completely opposite advice on work projects

src/devices/cpu/ht1130/ht1130.h
mamehaze: It might not be ideal, but I've simplified it.

Best reworked by somebody with a better understanding how how these usually get hooked up (I was following another driver when I did this)

mamehaze:
all feedback addressed I believe

mamehaze:
[brick.zip](https://github.com/mamedev/mame/files/13771283/brick.zip)
so the BrickEmuPy author gave us permission to use their svg.
as MAME's 'romset' SVGs don't typically include the case image (that is left for external artwork) I've stripped it back down to the above file.

I however can't figured out how to hook it up properly at all
for Game and Watch etc. the output finder is
		m_out_x(*this, "%u.%u.%u", 0U, 0U, 0U)
but looking at the game and watch SVGs the elements are just named gxxxxx, I don't see anything named/labelled with this x.x.x format in the game and watch svg files, so I'm not sure how the output finder is finding anything to work with?
@happppp / @galibert  any tips here?

happppp:
[brick_v2.zip](https://github.com/mamedev/mame/files/13771471/brick_v2.zip)

- For segment bits that belong together, group them (ctrl G in inkscape). eg. see NEXT and SPEED, and the 16 pixels at the right side, where I already did it.
- Add segment IDs to the titles (object properties -> title in Inkscape), see the NEXT seg where i filled it with "seg_id_here"

mamehaze:
[brickv4.zip](https://github.com/mamedev/mame/files/13771897/brickv4.zip)
I started from the original version because your ungrouping seemed to lose some names / groups that were already useful.  this is the file I'm using now.

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,130 @@
...
+	auto pp_in_cb() { return m_port_in_pp.bind(); }
+
+	auto pa_out_cb() { return m_port_out_pa.bind(); }
+
+	auto display_data_out_cb() { return m_display_data_out.bind(); }
+
+	void wake() // can be configured to wake from certain inputs as a mask option?
+	{
+		m_inhalt = 0;
+	}
```
cuavas, line 38: Probably better to put this through a `synchronize` barrier to avoid time travel effects on the odd chance it gets used in a system with multiple diexec implementations.

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,130 @@
...
+protected:
+	ht1130_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock, address_map_constructor data);
+
+	virtual uint32_t execute_max_cycles() const noexcept override { return 2; }
+
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+	virtual void execute_run() override;
+	virtual void execute_set_input(int inputnum, int state) override;
```
cuavas, line 49: `execute_max_cycles` should go with the rest of the `device_execute_interface` implementation.

src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -0,0 +1,118 @@
...
+void ht11xx_brickgame_state::machine_start()
+{
+	m_out_x.resolve();
+}
+
+void ht11xx_brickgame_state::machine_reset()
+{
+	for (int i=0;i<256;i++)
+		for (int j=0;j<4;j++)
+			m_out_x[i][j] = 0;
```
cuavas, line 56: Is this actually correct?  The datasheet for the CPU says, “All LCD segments are random after an initial clear.”  (See LCD driver output section on page 12.)  This suggests that a reset doesn’t affect the contents of display memory or the LCD outputs, and software shouldn’t expect the outputs to be in a given state on start.

At any rate, if you’re keeping it, please put some spaces in the controlling conditions for the loops.  (You could also use range-based `for` loops.)

src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+INPUT_CHANGED_MEMBER(ht11xx_brickgame_state::power_pressed)
+{
+	m_maincpu->wake();
+}
```
cuavas, line 62: The datasheet says that inputs configured to wake the system are “active on a low-going transition.”  You should check that the button is being pressed and not released here.  You’ll need to check that `newval` is non-zero, because MAME inverts the value for `IP_ACTIVE_LOW` before it gets here.

src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -0,0 +1,118 @@
...
+	PORT_BIT( 0x08, IP_ACTIVE_LOW, IPT_BUTTON4 ) PORT_NAME("Left")
+INPUT_PORTS_END
+
+
+void ht11xx_brickgame_state::display_data_w(offs_t offset, u8 data)
+{
+	for (int i = 0; i < 4; i++)
+	{
+		m_out_x[offset+0xb0][i] = (data >> i) & 1;
+	}
```
cuavas, line 83: What’s the fixed offset of 0xb0 here for?  Do those outputs actually exist?  Why instantiate them if they aren’t going to be used?

mamehaze: the layout from the other emulator uses the memory address, the display RAM starts at 0xb0, this made the conversion process MUCH smoother, and less prone to introducing errors.  I'd rather not have to change it right now

src/mame/handheld/hh_ht11xx.cpp
mamehaze: probably not, I was using it when converting the layout, to make sure I'd named all the segments properly.

src/mame/handheld/hh_ht11xx.cpp
cuavas: But you’re instantiating over 700 outputs that don’t exist.  That’s really ugly.  If you want the numbering to start at a different point, why don’t you use the actual output finder constructor argument that’s there for that?

src/devices/cpu/ht1130/ht1130.h
```diff
@@ -0,0 +1,131 @@
...
+
+	auto pa_out_cb() { return m_port_out_pa.bind(); }
+
+	auto display_data_out_cb() { return m_display_data_out.bind(); }
+
+	void wake() // can be configured to wake from certain inputs as a mask option?
+	{
+		machine().scheduler().synchronize();
+		m_inhalt = 0;
+	}
```
cuavas, line 39: This won’t work – `synchronize` takes effect asynchronously, so the change will still be visible before syncrhonisation takes place if you do it like this.  You need to provide a callback to `synchronize` to do the actual work.

mamehaze: I think I'll just back out the 'wake' changes instead

src/mame/handheld/hh_ht11xx.cpp
happppp: The prefered method is to match the COM and SEG pins, eg. x.y where X=0-3 for COM, y=0-31 for SEG (or however many segments this chip has).

happppp:
I did some googling, apparently the company is E-Star.
Whether or not Holtek programmed it, is speculation IMO.

src/devices/cpu/ht1130/ht1130.h
cuavas: You’ve been around long enough to know how `synchronize` works, and there are numerous examples of it in the code (e.g. Namco 06xx/50xx/51xx/52xx/54xx, the devices used for sound latches, the input merger devices, and Apple II Centronics printer cards, just off the top of my head).

mamehaze: I've backed out the changes, I'm not happy with that part anyway.

happppp: For CPU devices synchronized inputs, you can use set_input_line instead. Just like you wouldn't do irq_w(x) but instead set_input_line(x ? ASSERT_LINE : CLEAR_LINE)

mamehaze: well for now it isn't really important.  it can be added later if needs be.

rb6502: Hap's correct, if you just use an input line it's synchronized for you.

mamehaze: I've already said, I've backed out those changes, so it's irrelevant at this point.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1049 @@
...
+	m_timer = 0;
+	m_stackaddr = 0;
+	m_stackcarry = 0;
+}
+
+u8 ht1130_device::fetch()
+{
+	m_icount--; // single byte opcodes take a single cycle, double take 2
+	if (m_timer_en)
+		m_timer++;
```
happppp, line 247: Could you add a cycle() function (exact function name is up to you), move the above 3 lines there.
The four 1-byte opcodes that read from ROM are 2 cycles, so call cycle() when handling those opcodes.

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1049 @@
...
+	save_item(NAME(m_timer));
+	save_item(NAME(m_stackaddr));
+	save_item(NAME(m_stackcarry));
+}
+
+void ht1130_device::device_reset()
+{
+	m_pc = 0;
+
+	m_regs[0] = m_regs[1] = m_regs[2] = m_regs[3] = m_regs[4] = 0;
```
happppp, line 231: std::fill_n

src/devices/cpu/ht1130/ht1130.cpp
```diff
@@ -0,0 +1,1049 @@
...
+	case 0b10010000: case 0b10010001: case 0b10010010: case 0b10010011:
+	case 0b10010100: case 0b10010101: case 0b10010110: case 0b10010111:
+	case 0b10011000: case 0b10011001: case 0b10011010: case 0b10011011:
+	case 0b10011100: case 0b10011101: case 0b10011110: case 0b10011111:
+	{
+		const u8 operand = fetch();
+		const uint16_t fulladdr = ((inst & 0x07) << 8) | operand | (m_pc & 0x800);
+		const uint8_t bit = (inst & 0x18) >> 3;
+
+		if (getacc() & (1<<bit))
```
happppp, line 898: BIT(getacc(), bit)

mamehaze: I still think these macros only serve to obfuscate, but done

happppp:
According to the docs, 1 machine cycle (eg. a 1 byte opcode) takes 4 system clock cycles (from OSC pins).
The timer rate can be configured with a mask option (system clock / 2^n), n=0-13 (except 6 for some reason). So, timer rate can be faster or slower than machine cycle rate.

Add a TODO note if you want to do it later.

src/devices/cpu/ht1130/ht1130.cpp
happppp: Just add cycle() to the fetch() function :) No need to copy paste cycle() everywhere you do a fetch()

mamehaze:
given that we don't even know the frequency, or how those options are configured, I'll leave it as a note.

src/devices/cpu/ht1130/ht1130.cpp
happppp: I only use it in these kind of cases (x & 1 << y) or (x >> y & 1).

PR #11876, opened 2023-12-26
coleco.xml: Added a prototype.

FoxhackDN:
New working software list items (coleco.xml)
Masters of the Universe: The Power of He-Man (prototype, hacked) [AtariAge]

### DIFF
--- a/hash/coleco.xml
+++ b/hash/coleco.xml
@@ -2881,6 +2881,18 @@ license:CC0-1.0
 		</part>
 	</software>
 
+	<software name="powheman">
+		<!-- Title screen has been hacked to show "NOT FOR RESALE" -->
+		<description>Masters of the Universe: The Power of He-Man (prototype, hacked)</description>
+		<year>1984</year>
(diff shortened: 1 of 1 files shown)


FoxhackDN:
The game was released as a 32k binary file, but only a bit over 24k are really used, the rest of the data is just filled with FF. I wasn't sure if I should split the data into 3 or 4 8k files like the other games in the software list, so if that needs to be done, please let me know and I'll amend this submission.

I flagged this as baddump because the release has a "not for resale" watermark on the title screen.

cuavas:
Ah, damn.  Another defaced prototype ROM.

Do you know whether it actually uses the usual 4×8KiB ROM layout?  I haven’t seen a picture of the PCB for this one.

FoxhackDN:
Nope, due to all the drama surrounding the game's ownership, the program dump was made by an anonymous cart owner, and there were no photos of the PCB posted. I assume this has the standard 4x8 layout, but since it wasn't provided that way and there are no photos to confirm, I left it alone.

PR #11878, opened 2023-12-27
Epoch Cassette Vision / UPD777 - preliminary CPU core (disassembler only for now) and some basic driver structure + Software List

mamehaze:


mamehaze:
This is the most barebones possible,

I've been wanting to do this for a while, but frankly the state of the documentation of the dumps is a mess.  For all the championing of the scans that were done and put on archive.org there are significant gaps in the documentation, mismatched PCB pictures with carts, missing PCB pictures, duplicate PCB pictures etc.

I can't say for certain even which game is which.

(also I can't find the CPU documentation anymore either, it should probably be linked in the skeleton for if somebody wishes to make a start on the emulation)

mamehaze:
http://www.seanriddle.com/cassettevision/CassetteVisionPCBs1F.JPG   suggests Galaxian should be 006, there is no 006 dump.  It also has 2 game names on the 774C silkscreened cart, but we only have a single 774C dump (is it a multigame ROM depending on something external? - in that case we have more dumps than there are games though?)

This Galaxian is missing the front PCB scan
https://archive.org/download/galaxian-epochcv-hiresscans
so can't even use that to verify what was in the cart

This Battlevader shows 009 https://ia601906.us.archive.org/10/items/battlevader-epochcv-hiresscans/Battlevader%20-%20PCB%20Front.jpg
This Monster Mansion seems to have the same 009 scan
https://ia803201.us.archive.org/7/items/monster-mansion-epochcv-hiresscans/Monster%20Mansion%20-%20PCB%20Front.jpg
so one of those is incorrect for sure...

I think the only way this is gong to get sorted is if we get the ball rolling with it

happppp:
1 Yosaku = 005
2 Baseball = 778
3 Galaxian = 007
4 Big Sports 12 = 004
5 Battle Vader = 774
6 PakPak Monster = 008
7 New Baseball
8 Monster Mansion = 009
9 Astro Command = 010
10 Grand Champion
11 Monster Block = 013
12 Elevator Panic = 015

mamehaze:
maybe we're missing an earlier 006 Galaxian then?

Also it's strange that the 774 cart has "TV Vader & Yosaku" silkscreened on it.  Maybe there's a 774 version of Yosaku too?

I think New Baseball is 002 based on Sean's scans and the silkscreen.

mamehaze:
I've updated them to have names based on those numbers anyway

happppp:
Can you add a note about 10 being unreleased instead of 'not dumped'?
And reorder the game list? Either by number or by title.

mamehaze:
I've talked to other people who say it was released, just very uncommon.  I'd rather not add random 'unreleased' tags on things.

mamehaze:
As for credits, if you have a list / can get a list I'll add it (although I don't even know current policy for such things)

mamehaze:
also if you have the CPU docs links, do let me know, I can't find the links anymore (was temped to do a disassembler)

einstein95:
Probably will help
https://datasheetspdf.com/pdf-down/U/P/D/UPD7801-NEC.pdf
http://cmpslv2.starfree.jp/Scv/EnrScv.htm

happppp:
> As for credits, if you have a list / can get a list I'll add it (although I don't even know current policy for such things)

I don't have a list. I know Sean worked on these. So you could ask him.
Same for the most up to date CPU docs, best ask him.

mamehaze:
> Probably will help https://datasheetspdf.com/pdf-down/U/P/D/UPD7801-NEC.pdf http://cmpslv2.starfree.jp/Scv/EnrScv.htm

those are for the Super Cassette Vision, not the same.

mamehaze:
> > As for credits, if you have a list / can get a list I'll add it (although I don't even know current policy for such things)
> 
> I don't have a list. I know Sean worked on these. So you could ask him. Same for the most up to date CPU docs, best ask him.

I've checked to see if he has any of those things

mamehaze:
right, I think I'll call this a basic checkpoint.

it gets some important things done, such as the disassembler, which will form the essential basis of any further work.

I don't personally know how far I'll be able to take this one, as it likely needs hands from people more experienced in working with these weird MCUs, where every instruction has many meanings (most allow for the skipping of the next instruction based on specific conditions for example)

I notice the final page in every ROM is blank, but the page before it always looks nearly the same anyway (and doesn't end on a jump? I assume maybe they wrap back to the start of a page or similar?)

mamehaze:
actually according to the docs the final page simply doesn't exist physically, so I'll trim the ROMs

hash/cassvisn_cart.xml
```diff
@@ -0,0 +1,193 @@
...
+<softwarelist name="cassvisn_cart" description="Epoch Cassette Vision cartridges">
+
+	<software name="yosaku" supported="no">
+		<description>Yosaku</description>
+		<year>1981</year> <!-- 7/30/1981 -->
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="cassvisn_cart">
+			<dataarea name="prg" size="0xf00">
+				<rom name="d777c005prog.s01" size="0xf00" crc="f917d2f0" sha1="b26baf2fd9b961b5a0e0475dfac93c18bd54cc4d" loadflag="load16_word_swap"/>
+			</dataarea>
```
cuavas, line 44: The width and Endianness of these regions needs to match that of the CPU’s program space.  MAME only attempts to automatically work it out for a region with a tag matching a CPU device at the same level.  There’s no chance of working it out for a region that’s instantiated during media loading.

cuavas (changes requested):
The disassembler is written in a way that isn’t readable or maintainable.  Please rework it in terms of the bit patterns in the instruction encoding rather than using dozens or hundreds of case labels for instruction groups.

mamehaze:
Unfortunately it seems we have different ideas of readable here... 
I prefer doing things with the case statements, I can remove the commented out lines for unused instructions, otherwise I'd prefer it to stay that way.

mamehaze:
sorry, but I think due to fundamental differences of direction, I'm going to have to close this @happppp if you want to pick this up / continue it, feel free, otherwise this is dead.

happppp:
In the name of Slartibartfast, calm yourself and perhaps you'll be interested in this again soon.

cuavas:
Seriously, you need to consider other people.

It’s been Christmas/new year, and a lot of us have family commitments.  This is the first time in years I’ve put that ahead of MAME.  You’re expecting people to treat this as a full-time job with no breaks, and meet your unwritten deadlines.

If you think needing to infer a pattern from hundreds of case labels is in any way reasonable for the next person who looks at this file, I don’t know what to say.  That simply isn’t maintainable.

mamehaze:
That's because you've made everything require your approval basically.  You've become the bottleneck.

I like how it's designed, I generated all the cases, I grouped them, I put the appropriate disassembler lines with each group, I figured out the ones that weren't needed, I made sure it all aligned with the documentation.

My approach was to keep it as simple as possible, it already kinda annoys me that I reduced some of the others to if statements with bitmasks.

mamehaze:
also you're pulling me up on things that were just fine last week with the holtek core, there was no grouping of binary into blocks of 4 there.

essentially I used the same approach for the core here as I did there, which was fine there, but not fine here.

cuavas:
That is not “as simple as possible” for the next person reading the code or trying to debug it.

The Holtek core is a lot simpler with far fewer instructions.  There are no cases with hundreds of case labels there – it’s 16 at most.  That’s not too bad to read.  Are you seriously trying to argue that trying to spot a pattern in 16 values is no different to over a hundred values?

mamehaze:
personally i find it a lot easier to read than a lot of MAME code today, it's just tables which group opcodes, for a cpu used in one place.

mamehaze:
the only lines I'd consider to be ugly myself are the commented out ones, which can go anyway

the rest of the cases, are IMHO not excessive.

mamehaze:
the REASON it is developed this way is simple anyway

if I'm looking a binary, and get the bitfield of a word I can simply search that bitfield in this file, and find the relevant line that disassembles it.

if I start using if statements with masks, I lose that ability entirely.

for a CPU of this scope, my design, which I consider to have been an active decision I made, works best for the development and testing patterns I'm following.

demanding that be changed is not something I'm willing to compromise on.

mamehaze:
Ok, I've addressed the parts of the feedback I consider to be legitimate.

hash/cassvisn_cart.xml
mamehaze: this is a weird case of every word being 12-bit (in a 16-bit word in the ROM) rather than any real native endianness thing.  the hardware doesn't have a concept of bytes.  so a bit of a weird one anyway.  might need revisiting later.  Swapped in loading for now instead.

src/devices/cpu/upd777/upd777dasm.cpp
```diff
@@ -0,0 +1,721 @@
...
+		// case 0b000001001101: case 0b000001001110: case 0b000001001111: case 0b000001010000: case 0b000001010001: case 0b000001010010: case 0b000001010011:
+		// case 0b000001010101: case 0b000001010110: case 0b000001010111:
+		// case 0b000001011001: case 0b000001011010: case 0b000001011011:
+		// case 0b000001011101: case 0b000001011110: case 0b000001011111:
+		// case 0b000001100001: case 0b000001100010: case 0b000001100011: case 0b000001100100: case 0b000001100101: case 0b000001100110: case 0b000001100111: case 0b000001101000: case 0b000001101001: case 0b000001101010: case 0b000001101011: case 0b000001101100: case 0b000001101101: case 0b000001101110: case 0b000001101111:
+
+		/////////////////////////////////////////////////////////////////////////////////////////////////////
+		/////////////////////////////////////////////////////////////////////////////////////////////////////
+
+		case 0b001000000000: case 0b001000000001: case 0b001000000010: case 0b001000000011:
```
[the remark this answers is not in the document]
mamehaze: as per comments, I disagree here, I designed it this way specifically, as it has benefits to the development process.

this is the design I want.

src/devices/cpu/upd777/upd777dasm.cpp
```diff
@@ -0,0 +1,721 @@
...
+		}
+		case 0b000000110000:
+		case 0b000000110100:
+		case 0b000000111000:
+		case 0b000000111100:
+		case 0b000001110000:
+		case 0b000001110100:
+		case 0b000001111000:
+		case 0b000001111100:
+		{
```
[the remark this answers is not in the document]
mamehaze: as per comments, I disagree here, I designed it this way specifically, as it has benefits to the development process.

this is the design I want.

mamehaze:
I've put 'resolve' on the things I don't plan on changing as they are 'by design' as well as the legitimate feedback that I've addressed.

mamehaze:
casvisn
even at this early stage you can kind of identify what some of the games are from what they put in the "sprite" memory.
This is obviously Pak Pak monster for example.

mamehaze:
is even more obviously Yosaku

src/mame/epoch/cassvisn.cpp
```diff
@@ -80,12 +79,31 @@ INPUT_PORTS_END
...
-	memcpy(memregion("maincpu")->base(), m_cart->get_rom_base(), size);
+	uint8_t* prgbase = memregion("maincpu")->base();
+	memcpy(prgbase, m_cart->get_rom_base(), size);
+
+	for (int i = 0; i < size; i += 2)
+	{
+		uint8_t temp = prgbase[i + 0];
+		prgbase[i + 0] = prgbase[i + 1];
+		prgbase[i + 1] = temp;
+	}
```
happppp, line 66: Or is it just a matter of changing the cpu device endianness?

src/devices/machine/upd777_dev.h
```diff
@@ -0,0 +1,32 @@
...
+private:
+	uint32_t screen_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect);
+
+	void palette_init(palette_device &palette) const;
+
+	required_device<gfxdecode_device> m_gfxdecode;
+	required_device<palette_device> m_palette;
+};
+
+DECLARE_DEVICE_TYPE(UPD777, upd777_device)
```
happppp, line 30: Unless MCU embedded peripherals exist as separate chips (eg. Toshiba Z80 with built-in CTC), they can be implemented in the CPU device.

mamehaze: interesting, I was always told to keep video stuff out of the 'cpu' folder in the past

src/mame/epoch/cassvisn.cpp
mamehaze: is Endianess really a thing in this case? it's a single 12 bit byte?  I find this a little confusing.

happppp: Here it'd just be for letting MAME know the byte order for each ROM word. TMS0980 and HMCS40 are similar, with 9-bit and 12-bit ROM words.

src/devices/machine/upd777_dev.h
happppp: Why? Only for video? I don't see how that makes sense while at the same time having eg. the versatile serial controller for Hitachi H8 or HD6301 in the CPU device.
Or, see the LCD controller in SM51x or the HT11xx that you recently added, that is video.

mamehaze: well one of the previous rules was that things in CPU should be somewhat usable outside of MAME, with minimal effort, so using MAME specific video type stuff was considered bad.

If that has changed, fine.  I might still keep them in different files, for organization sake, but can put the _dev files in the CPU folder.

happppp: I'm not aware of such a rule. Maybe you mean it was the opinion of someone that gave feedback on an older PR of yours.

mamehaze: well given the amount of contradictory 'make it up as you go along' feedback I seem to get, that wouldn't surprise me.

src/mame/epoch/cassvisn.cpp
mamehaze: I changed everything to 'big' and the swap is still needed, I must be missing something here.

happppp: I see, that's weird. Just add a TODO note then.

PR #11882, opened 2023-12-28
x68k_flop.xml: add three dumps from original disks

b0x-Cub3d:
Be Rain [Shentok]
Akazukin Cha Cha no Chotto dake yo - Otameshi Plus Alpha ver 0.53 [Shentok] 
Prote Pruste [Tahiti Cartago]

### DIFF
--- a/hash/x68k_flop.xml
+++ b/hash/x68k_flop.xml
@@ -22956,6 +22956,7 @@ sold through Takeru vending machines -->
 		<description>Akazukin Cha Cha no Chotto dake yo - Otameshi Plus Alpha ver 0.53</description>
 		<year>1996</year>
 		<publisher>&lt;doujin&gt;</publisher>
+		<info name="alt_title" value="赤ずきんチャチャのチョットダケヨ・ヲタ飯版ＰＬＵＳα" />
 		<info name="developer" value="Psijack" />
 		<part name="flop1" interface="floppy_5_25">
 			<dataarea name="flop" size="1281968">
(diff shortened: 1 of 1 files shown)


cuavas:
@rzero9 can you give this a look?

hash/x68k_flop.xml
```diff
@@ -22928,6 +22951,18 @@ sold through Takeru vending machines -->
 			</dataarea>
 		</part>
 	</software>
+	
+	<software name="akazopa">
+		<description>Akazukin Cha Cha no Chotto dake yo - Otameshi Plus Alpha ver 0.53</description>
+		<year>1996</year>
+		<publisher>&lt;doujin&gt;</publisher>
+		<info name="developer" value="Psijack" />
```
cuavas, line 22959: Do you have the title in Japanese script?

hash/x68k_flop.xml
```diff
@@ -29649,6 +29684,20 @@ Expects to read track data from [FDC] drive 1, works if same disk loaded.
 		</part>
 	</software>
 
+	<!-- Title is confirmed in executable -->
+	<!-- Game plays in Tate mode -->
+	<software name="prote">
+		<description>Prote Pruste</description>
+		<year>1999</year>
+		<publisher>Soft Bank</publisher>
+		<info name="developer" value="ChinaSpark" />
```
cuavas, line 29693: Is “Soft Bank” in this software list different to SoftBank (e.g. in fmtowns_cd.xml)?

Also, does this game have a title in Japanese script?

rzero9:
> @rzero9 can you give this a look?

Other than adding the alt_titles if they are available, it looks fine to me. Prote Pruste was apparently released as a coverdisk with Oh!X magazine, so maybe a note about that would be nice, just to know where it came from.

I don't have access to the files, but judging from the sizes the disks look unprotected (which isn't surprising for doujin and coverdisk stuff, or for Parsley's games), so there should be no issues with that.

PR #11883, opened 2023-12-28
igs/pgm.cpp: redump martmasttw

mistydemeo:
I saw the comment that u9/u10 in martmasttw are possibly a bad dump, so I dumped them from my own cartridge. They came back as identical to martmast102c, so looks like it was a bad dump previously. The ARM ROM looks like it's the only unique ROM in this set. That dump is still bad; I'm not sure how to dump it, though if someone has the custom ROM/ROMs that are used to do that I'm happy to give it a go.

### DIFF
--- a/src/mame/igs/pgm.cpp
+++ b/src/mame/igs/pgm.cpp
@@ -3820,44 +3820,6 @@ ROM_START( martmast102c )
 ROM_END
 
 /*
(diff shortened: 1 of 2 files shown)


mamehaze:
if the set is now identical apart from the completely undumped internal ARM ROM then it should just be removed.

mistydemeo:
Makes sense. Swapped to a commit that removes it entirely.

PR #11884, opened 2023-12-28
netronics/exp85.cpp: implemented rom mirroring, fixed interrupt handling timing

fmilano:
The Netronics Explorer-85 has dedicated circuitry for mirroring the monitor ROM to  low memory (0x0000 -- 0x0800) after a system start/reset and during an interrupt. This logic was reproduced as accurately as possible, since the emulation of the intel 8085 does not have the pin outs needed for implementing the exact logic used in the original circuitry (IO/M, S0, S1 and ALE).
With the mirror switching working, it is possible to execute the MS BASIC ROM (it uses the low memory 8KB of RAM).

 Microsoft Basic

Another non-working feature was the step by step execution capability of the monitor program. This tracer uses a timer to interrupt the executing program once per instruction. In a real system the 8155 timer chip has a delay of around ~70ns (measured). This is enough for the next instruction cycle to start and for the interrupt to be acknowledged one instruction later. This effect is fundamental for the ROM monitor stepping functionality. This is solved with a MAME timer, dealying the 8155 to signal. The stepping mechanism is now working:

 Step by step execution

### DIFF
--- a/src/mame/netronics/exp85.cpp
+++ b/src/mame/netronics/exp85.cpp
@@ -5,6 +5,8 @@
     Explorer 85
 
     12/05/2009 Skeleton driver.
+	29/12/2023 Reworked driver to enable extended RAM memory,
+	           MS BASIC and disable ROM mirror after boot/interrupt.  
 
     Setting Up
     ==========
@@ -16,6 +18,17 @@
     Once started, press Space. The system will start up.
     All input must be in upper case.
 
+	Microsoft Basic is executed entering the following 2 commands 
+	in the monitor prompt:
+	.XS nnnn-F87F mmmm-C000 <CR>
+    .G <CR>
+
+	where nnnn is the previous value of the stack point, and mmmm is the previous
+	value of the stack pointer. 
+
+	Basic will request the amount of RAM memory available, 8192 (bytes) must be entered.
+
+
 ****************************************************************************/
 
 /*
@@ -23,8 +36,8 @@
     TODO:
 
     - dump of the hexadecimal keyboard monitor ROM
-    - disable ROM mirror after boot
-    - RAM expansions
+    - 64K RAM expansion
+	- Disk drive and CP/M OS support
 
 */
 
@@ -41,12 +54,18 @@
 #include "imagedev/cassette.h"
 #include "sound/spkrdev.h"
 
+#define VERBOSE 0 //(1U)
+#include "logmacro.h"
 
 namespace {
 
-#define I8085A_TAG      "u100"
-#define I8155_TAG       "u106"
-#define I8355_TAG       "u105"
+#define I8085A_TAG       "u100"
+#define I8155_TAG        "u106"
+#define I8355_TAG        "u105"
+#define EXTENDED_RAM_TAG "extended_ram"
+
+#define LOW_MEMORY_ROM_MIRROR_ENTRY 0
+#define LOW_MEMORY_RAM_ENTRY        1
 
 class exp85_state : public driver_device
 {
@@ -54,11 +73,14 @@ public:
 	exp85_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, I8085A_TAG)
+		, m_i8155(*this, I8155_TAG)
 		, m_rs232(*this, "rs232")
 		, m_cassette(*this, "cassette")
 		, m_speaker(*this, "speaker")
-		, m_rom(*this, I8085A_TAG)
+		, m_rom(*this, I8355_TAG)
+		, m_extended_ram(*this, EXTENDED_RAM_TAG)
 		, m_tape_control(0)
+		, m_timer(nullptr)
 	{ }
 
 	void exp85(machine_config &config);
@@ -68,10 +90,13 @@ public:
 
 private:
 	required_device<i8085a_cpu_device> m_maincpu;
+	required_device<i8155_device> m_i8155;
 	required_device<rs232_port_device> m_rs232;
 	required_device<cassette_image_device> m_cassette;
 	required_device<speaker_sound_device> m_speaker;
 	required_memory_region m_rom;
+	required_device<ram_device> m_extended_ram;
+
 
 	virtual void machine_start() override;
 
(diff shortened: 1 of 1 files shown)


src/mame/netronics/exp85.cpp
```diff
@@ -41,24 +54,33 @@
...
 
+#define VERBOSE 0 //(1U)
+#include "logmacro.h"
 
 namespace {
 
-#define I8085A_TAG      "u100"
-#define I8155_TAG       "u106"
-#define I8355_TAG       "u105"
+#define I8085A_TAG       "u100"
```
rb6502, line 62: A lot of the devs (myself included) have largely moved away from #defining all of the tag names because it doesn't often buy you much.  But we don't prohibit it either, so if you like it we won't stop you.

src/mame/netronics/exp85.cpp
```diff
@@ -41,24 +54,33 @@
...
+#define I8155_TAG        "u106"
+#define I8355_TAG        "u105"
+#define EXTENDED_RAM_TAG "extended_ram"
+
+#define LOW_MEMORY_ROM_MIRROR_ENTRY 0
```
rb6502, line 67: For numeric constants it's preferred now to use the C++ `static constexpr` form, so in this case `static constexpr int LOW_MEMORY_ROM_MIRROR_ENTRY = 0;`  That way the compiler can do better sanity checking.

src/mame/netronics/exp85.cpp
```diff
@@ -80,21 +105,38 @@ class exp85_state : public driver_device
 	int sid_r();
 	void sod_w(int state);
 
+	/* memory mapping during reset and interrupt mgmt */
+	void status_out(u8 status);
+	bool prep_int_call = false;
```
rb6502, line 110: Modern MAME style preference is that all member variables have an `m_` prefix to make it clear what's going on, and that member variables and member functions be grouped separately.  Usually all the functions and then all of the variables.

I realize pretty much everything I've discussed so far is inherited technical debt rather than your new work.  That's the occupational hazard of picking up a skeleton driver that's 15 years old.  Cleaning all that up now will make things simpler when the RAM and FDC are added to run CP/M later.

src/mame/netronics/exp85.cpp
```diff
@@ -201,6 +243,43 @@ void exp85_state::sod_w(int state)
...
+{
+	// In the real hardware this is monitored via the IO/M, S0, S1 
+	// and ALE output pins of the 8085.
+	// Since these pins are not emulated, then we must explicitly get the internal 
+	// interrupt acknowledge state (0x23 or 0x26) and behave as the monitor expects.
+	auto current_pc = m_maincpu->pc();
+	if (status == 0x23 || status == 0x26)
+	{
+		// When an interrupt is triggered the low memory shall be set to the ROM mirror
+		LOG("set low memory to ROM mirror (PC 0x%x, proc status 0x%x)\n", current_pc, status);
```
rb6502, line 258: Use LOGMASKED here so that CPU cycles aren't being blown logging this in normal operation.

src/mame/netronics/exp85.cpp
```diff
@@ -201,6 +243,43 @@ void exp85_state::sod_w(int state)
...
+	{
+		// Avoids setting the lower memory back to RAM until
+		// it branches to the interrupt handler.
+		prep_int_call = false;
+	} 
+	else if (!prep_int_call && (current_pc & 0xF000) == 0xF000 && membank("bank1")->entry() == LOW_MEMORY_ROM_MIRROR_ENTRY)
+	{
+		// When the interrupt handler is executing and the address is >= 0xF000
+		// the low memory is mapped to RAM
+		LOG("set low memory to RAM (PC 0x%x, status 0x%x)\n", current_pc, status);
```
rb6502, line 272: Ditto here with LOGMASKED.

src/mame/netronics/exp85.cpp
fmilano: I took your advice and removed these defines; the names "uNNN" reflect the id of the component on the circuit diagrams published by netronics.

src/mame/netronics/exp85.cpp
fmilano: Fixed

src/mame/netronics/exp85.cpp
fmilano: Removed logging, does not make much sense to leave it there at this point.

src/mame/netronics/exp85.cpp
fmilano: Fixed

src/mame/netronics/exp85.cpp
```diff
@@ -134,17 +124,17 @@ void exp85_state::exp85_mem(address_map &map)
 	// Microsoft Basic ROM
 	map(0xc000, 0xdfff).rom();
 	// monitor ROM in the 8355 chip.
-	map(0xf000, 0xf7ff).r(I8355_TAG, FUNC(i8355_device::memory_r));
+	map(0xf000, 0xf7ff).r("u105", FUNC(i8355_device::memory_r));
```
rb6502, line 127: You can use the m_ objects here instead of the names.  That avoids additional run-time string lookups, and also eliminates the need to know the name in more than one place.  So just m_rom instead of "u105", and m_i8155 instead of "u106".

rb6502:
Thanks!  Looks like one more bit of cleanup and we're good to go.

src/mame/netronics/exp85.cpp
fmilano: I agree, the code is much better this way. Anyways, I will keep an 8355 object and a rom object finder, even if the rom is in principle already contained in the 8355 device and I have a reference to it. This would be much simpler if the 8355 class implemented the device_memory_interface interface, since in this case I would be able to get a memory base pointer without the need of a memory region object finder. Do you think it would be reasonable in the future to inherit from this interface in the 8355 device class?

src/mame/netronics/exp85.cpp
```diff
@@ -67,50 +84,67 @@ class exp85_state : public driver_device
...
 
-/* Memory Maps */
 
+/* Memory Maps */
 void exp85_state::exp85_mem(address_map &map)
 {
 	map.unmap_value_high();
-	map(0x0000, 0x07ff).bankr("bank1");
+	// Extended RAM or mapped monitor ROM (only during interrupt and reset)
+	map(0x0000, 0x2000).bankrw("bank1");
```
cuavas, line 125: The problem with using `bankrw` here is that it will allow writing to the ROM, which shouldn’t be possible.  You should really use a memory view and either install the ROM in one entry and the RAM in another, or install one directly in the address space and the other in a memory view entry.

You can see a relatively simple use of a memory view in the Jaleco Psychic 5 driver (find the uses of `m_vrambank` in the file):
https://github.com/mamedev/mame/blob/mame0261/src/mame/jaleco/psychic5.cpp

You can find other uses in various places by searching for `memory_view`.

fmilano: Thanks! I have changed the bank for a memory view and I'm using it to switch between the RAM and the ROM mirror.

src/mame/netronics/exp85.cpp
```diff
@@ -201,6 +239,41 @@ void exp85_state::sod_w(int state)
...
+		m_is_preparing_interrupt_call = false;
+	} 
+	else if (!m_is_preparing_interrupt_call && (current_pc & 0xf000) == 0xf000)
+	{
+		// When the interrupt handler is executing and the address is >= 0xf000
+		// the low memory is mapped to RAM
+		m_low_memory_view.select(LOW_MEMORY_RAM_ENTRY);
+	} 
+		
+}
```
cuavas, line 270: @ajrhacker are you happy with this as a temporary solution with the current state of the 8086 CPU device, or would you suggest another way of achieving this?

cuavas (commented):
If there’s no better solution for the automatic bank switching than snooping on the 8085’s PC register, I think we can merge this.

PR #11894, opened 2024-01-03
pkspirit.cpp - fix rom loading so that bitplanes are in correct order & add palette

mamehaze:


mamehaze:
pokerspirit

angelosa:
The video hookup is known to be wrong, cfr. `accesses unmapped range $e80 in TMP68303F for a sprite DMA $300ff2 -> $b10000` note on top.

mamehaze:
yeah maybe it's missing some dma for the attributes too then or something, at least this gets the rom loading and palette format out the way anyway.

angelosa:
It's the typical Seta "World" drawing, with 4 words containing pointers to GFX structures, sizes and colors.
sprite_struct
Technically you can even fake the transfer by directly referencing the work RAM buffer instead.

mamehaze:
yeah, sounds logical, so the tilemaps aren't really tilemaps, but large floating sprites, with the colour attribute in the list.  Very weird to see a text layer done like that, but for these gambling games I guess they didn't really care.

I mostly just wanted to sort out the ROM load really anyway.

PR #11896, opened 2024-01-03
pkspirit.cpp - doesn't use tilemaps, uses a spritelist. remove tilemaps, hook up preliminary list

mamehaze:


mamehaze:
as per @angelosa comment in previous PR, the list is meant to be copied with a DMA, but for now just process it where it is (it'll be a simple case of changing a constant once it's in the correct location)

this removes the tilemaps, as they don't exist, and draws the previously shown layers with the correct colours.  no attempt has been made to draw more complex objects from the list yet.

mamehaze:
meme

mamehaze:
anyway, basically I'm done with this for now as I have other things to look at.

the x/y sprite sizes need narrowing down properly, and the purpose of the 2bpp font it uploads needs figuring out (maybe there's another fixed tilemap layer, possibly not used here?)

otherwise, it's in a much better state than it was before.

mamehaze:
pkspirit_1
pkspirit_2
pkspirit_3
pkspirit_4

of note the 'hold' sprite is currently wrong.

also there's no credit display, which is why I wonder if that 2bpp font is for a fixed layer of some kind (unless it's currently just positioned off-screen)

src/mame/taito/pkspirit.cpp
```diff
@@ -83,64 +81,111 @@ class pkspirit_state : public driver_device
...
+	u32 spriteram_start = 0x300ff2; // should be 0xb10000 but a DMA is missing
+	u32 spriteram2_start = 0xb00000;
+
+	for (int i = 0; i < 0x800; i += 8)
+	{
+		// somewhat illogical, but the first sprite (one of the large tilemap sized sprites) seems to need to be drawn last
+		// (or there is some priority control)
+		int reali = (i + 0x8) & 0x7ff;
+
+		uint16_t sp0 = m_maincpu->space().read_word(spriteram_start + reali + 0);
```
angelosa, line 116: Using `space()` like this won't work when the actual DMA support is added later. Consider using `required_shared_ptr` for all the relevant buffers instead.

src/mame/taito/pkspirit.cpp
```diff
@@ -292,12 +342,24 @@ const gfx_layout gfx_16x16x5_planar =
...
 };
 
+const gfx_layout gfx_ram =
+{
+	16,16,
+	RGN_FRAC(1,1),
+	2,
+	{ 0,1 },
+	{ 0,2,4,6,8,10,12,14, 128,130,132,134,136,138,140,142 },
+	{ STEP8(0,16), STEP8(256,16) },
```
cuavas, line 352: This would be cleared if the first line used `STEP8` as well.

src/mame/taito/pkspirit.cpp
mamehaze: it will work fine? the only thing the DMA is doing is copying the list from one place to another, change the address to the destination as per the comment and the code will still function.

angelosa: I don't mean to make the actual DMA copy but rather have some explicit pointers. Using `m_maincpu->space().read_word` N times is also inefficient when it comes to performance.

mamehaze: seems like splitting hairs, as it will need changing once the core does DMA, and performance diff here is negligible, but I've changed it to a pointer directly into mainram instead.

mamehaze:
The final changes I've made fix the HOLD display, and paytable display (draw red sprite behind numbers, not over) although there is still a small glitch on the reel for the 'with reel' double up type.  Could be there's a proper priority scheme I've missed, but with only a single game on this hardware it's difficult to draw solid conclusions.

angelosa (approved):

PR #11898, opened 2024-01-04
heathkit/h89.cpp: Add H-88-5 Cassette interface and the H88 clone of H89

mgarlanger:
Adds the H-88-5 cassette interface board and the H88 clone of the H89. 

Adding the H88 clone (which Heath Company sold with only the cassette interface board) is simpler and more historically accurate. 

MTR-88 BIOS was also change to only be valid with the H88 and the only bios supported by the H88.

### DIFF
--- a/src/mame/heathkit/h89.cpp
+++ b/src/mame/heathkit/h89.cpp
@@ -43,10 +43,10 @@
 
 #include "emu.h"
 
+#include "h_88_cass.h"
+#include "intr_cntrl.h"
 #include "tlb.h"
 #include "z37_fdc.h"
-#include "intr_cntrl.h"
-#include "h_88_cass.h"
 
 #include "cpu/z80/z80.h"
 #include "machine/ins8250.h"
@@ -65,10 +65,13 @@
 
 namespace {
 
-class h89_state : public driver_device
+/**
+ * Base Heathkit H89 functionality
+ */
+class h89_base_state : public driver_device
 {
-public:
-	h89_state(const machine_config &mconfig, device_type type, const char *tag):
+protected:
+	h89_base_state(const machine_config &mconfig, device_type type, const char *tag):
 		driver_device(mconfig, type, tag),
 		m_maincpu(*this, "maincpu"),
 		m_maincpu_region(*this, "maincpu"),
@@ -76,8 +79,6 @@ public:
 		m_ram(*this, RAM_TAG),
 		m_floppy_ram(*this, "floppyram"),
 		m_tlbc(*this, "tlbc"),
-		m_h37(*this, "h37"),
-		m_cassette(*this, "h_88_5"),
 		m_intr_socket(*this, "intr_socket"),
 		m_console(*this, "console"),
 		m_serial1(*this, "serial1"),
@@ -89,19 +90,12 @@ public:
 
 	void h89_base(machine_config &config);
 
-	void h88(machine_config &config);
-	void h89(machine_config &config);
-
-private:
-
 	required_device<z80_device> m_maincpu;
 	required_memory_region m_maincpu_region;
 	memory_view m_mem_view;
 	required_device<ram_device> m_ram;
 	required_shared_ptr<uint8_t> m_floppy_ram;
 	required_device<heath_tlb_connector> m_tlbc;
-	optional_device<heath_z37_fdc_device> m_h37;
-	optional_device<heath_h_88_cass_device> m_cassette;
 	required_device<heath_intr_socket> m_intr_socket;
 	required_device<ins8250_device> m_console;
 	required_device<ins8250_device> m_serial1;
@@ -149,14 +143,58 @@ private:
 	uint8_t m1_r(offs_t offset);
 
 	void h89_base_io(address_map &map);
-	void h89_io(address_map &map);
-	void h88_io(address_map &map);
 
 	uint8_t raise_NMI_r();
 	void raise_NMI_w(uint8_t data);
 	void console_intr(uint8_t data);
 	void reset_line(int data);
 	void reset_single_step_state();
+};
+
+/**
+ * Heathkit H88
+ *  - BIOS MTR-88
+ *  - H-88-5 Cassette Interface Board
+ *
+ */
+class h88_state : public h89_base_state
+{
+public:
+	h88_state(const machine_config &mconfig, device_type type, const char *tag):
+		h89_base_state(mconfig, type, tag),
+		m_cassette(*this, "h_88_5")
+	{
+	}
+
+	void h88(machine_config &config);
+
+protected:
+	required_device<heath_h_88_cass_device> m_cassette;
+
+	void h88_io(address_map &map);
+};
+
+
+/**
+ * Heathkit H89
+ *  - Z-89-37 Soft-sectored Floppy Controller
+ *
+ */
+class h89_state : public h89_base_state
+{
+public:
+	h89_state(const machine_config &mconfig, device_type type, const char *tag):
+		h89_base_state(mconfig, type, tag),
+		m_h37(*this, "h37")
+	{
+	}
+
+	void h89(machine_config &config);
+
+protected:
+	required_device<heath_z37_fdc_device> m_h37;
+
+	void h89_io(address_map &map);
 };
 
 /*
@@ -195,7 +233,7 @@ private:
   -------------  0k              -------------  0k
 
 */
-void h89_state::h89_mem(address_map &map)
+void h89_base_state::h89_mem(address_map &map)
 {
 	map.unmap_value_high();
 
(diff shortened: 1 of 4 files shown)


src/mame/heathkit/h89.cpp
```diff
@@ -46,6 +46,7 @@
 #include "tlb.h"
 #include "z37_fdc.h"
 #include "intr_cntrl.h"
+#include "h_88_cass.h"
```
cuavas, line 49: Please keep the `#include` statements sorted within each section – they get unmanageable otherwise.

src/mame/heathkit/h89.cpp
```diff
@@ -85,8 +87,10 @@ class h89_state : public driver_device
 	{
 	}
 
-	void h89(machine_config &config);
+	void h89_base(machine_config &config);
```
cuavas, line 90: If this is abstract, it should be `protected` or `private`.

src/mame/heathkit/h89.cpp
```diff
@@ -96,7 +100,8 @@ class h89_state : public driver_device
 	required_device<ram_device> m_ram;
 	required_shared_ptr<uint8_t> m_floppy_ram;
 	required_device<heath_tlb_connector> m_tlbc;
-	required_device<heath_z37_fdc_device> m_h37;
+	optional_device<heath_z37_fdc_device> m_h37;
+	optional_device<heath_h_88_cass_device> m_cassette;
```
cuavas, line 104: Can you use derived classes for the machines with different capabilities and avoid `optional_device`?

src/mame/heathkit/h89.cpp
```diff
@@ -258,20 +265,20 @@ uint8_t h89_state::m1_r(offs_t offset)
...
 	map.global_mask(0xff);
 
 	// Disk I/O #1 - 0170-0173 (0x78-0x7b)
 	//   Options
-	//     - Cassette I/O (only uses 0x78 - 0x79) - Requires MTR-88 ROM
 	//     - H37 5-1/4" Soft-sectored Controller - Requires MTR-90 ROM
 	//     - H47 Dual 8" Drives - Requires MTR-89 or MTR-90 ROM
 	//     - H67 8" Hard disk + 8" Floppy Drives - Requires MTR-90 ROM
-	map(0x78, 0x7b).rw(m_h37, FUNC(heath_z37_fdc_device::read), FUNC(heath_z37_fdc_device::write));
+	// map(0x78, 0x7b)
```
cuavas, line 281: This now duplicates the comment in the H89 address map – should it really be here as well?

src/mame/heathkit/h89.cpp
```diff
@@ -300,7 +307,53 @@ void h89_state::h89_io(address_map &map)
...
+
+void h89_state::h88_io(address_map &map)
+{
+	h89_base_io(map);
+
+	// Cassette I/O (uses 0xf8 - 0xf9) - Requires MTR-88 ROM
+	map(0xf8, 0xf9).rw(m_cassette, FUNC(heath_h_88_cass_device::read), FUNC(heath_h_88_cass_device::write));
+
+	// 5-1/4" Hard-sectored Controller - only other card supported MTR-88.
+	// map(0x7c, 0x7f)
```
cuavas, line 330: Please keep things in ascending address order if there isn’t a compelling reason not to.

src/mame/heathkit/h89.cpp
```diff
@@ -300,7 +307,53 @@ void h89_state::h89_io(address_map &map)
...
+}
+
 // Input ports
+static INPUT_PORTS_START( h88 )
+
+	PORT_START("SW501")
+	// MTR-88  (444-40)
+	PORT_DIPNAME( 0x1f, 0x00, DEF_STR( Unused ) )               PORT_DIPLOCATION("SW501:1,2,3,4,5")
```
cuavas, line 338: You still want to be able to toggle the unused DIP switches – this won’t allow that.

src/mame/heathkit/h89.cpp
```diff
@@ -739,6 +771,43 @@ void h89_state::h89(machine_config & config)
...
+
+	HEATH_INTR_SOCKET(config, m_intr_socket, intr_ctrl_options, "h37", true);
+	m_intr_socket->irq_line_cb().set_inputline(m_maincpu, INPUT_LINE_IRQ0);
+
+	// Z-89-37 Soft-sectored controller
+	HEATH_Z37_FDC(config, m_h37);
+	m_h37->drq_cb().set(m_intr_socket, FUNC(heath_intr_socket::set_drq));
+	m_h37->irq_cb().set(m_intr_socket, FUNC(heath_intr_socket::set_irq));
+	m_h37->block_interrupt_cb().set(m_intr_socket, FUNC(heath_intr_socket::block_interrupts));
+}
```
cuavas, line 799: You could add the interrupt controller in the base function and just use `set_default_option` to set the default interrupt controller for each concrete machine.

src/mame/heathkit/h89.cpp
```diff
@@ -778,3 +844,4 @@ ROM_END
 
 //    YEAR  NAME  PARENT  COMPAT  MACHINE  INPUT  CLASS      INIT        COMPANY          FULLNAME        FLAGS
 COMP( 1979, h89,  0,      0,      h89,     h89,   h89_state, empty_init, "Heath Company", "Heathkit H89", MACHINE_SUPPORTS_SAVE)
+COMP( 1979, h88,  h89,    0,      h88,     h88,   h89_state, empty_init, "Heath Company", "Heathkit H88", MACHINE_SUPPORTS_SAVE)
```
cuavas, line 847: Can you keep these in the same order as the ROM definitions, etc.?

src/mame/heathkit/h_88_cass.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "h_88_cass.h"
+
+#include "machine/clock.h"
+
+#include "speaker.h"
+
+#define LOG_REG   (1U << 0)
+#define LOG_LINES (1U << 1)
+#define LOG_CASS  (1U << 2)
+#define LOG_FUNC  (1U << 3)
```
cuavas, line 22: You can’t use `(1U << 0)` for one of your own categories, as it’s used for the predefined `LOG_GENERAL`.

src/mame/heathkit/h_88_cass.h
```diff
@@ -0,0 +1,53 @@
...
+****************************************************************************/
+
+#ifndef MAME_HEATHKIT_H_88_CASS_H
+#define MAME_HEATHKIT_H_88_CASS_H
+
+#pragma once
+
+#include "machine/i8251.h"
+#include "machine/timer.h"
+#include "imagedev/cassette.h"
```
cuavas, line 16: Please keep these sorted.

src/mame/heathkit/h_88_cass.h
```diff
@@ -0,0 +1,53 @@
...
+	TIMER_DEVICE_CALLBACK_MEMBER(kansas_r);
+	TIMER_DEVICE_CALLBACK_MEMBER(kansas_w);
+
+	required_device<i8251_device> m_uart;
+	required_device<cassette_image_device> m_cass_player;
+	required_device<cassette_image_device> m_cass_recorder;
+
+	u8 m_cass_data[4]{};
+	bool m_cassbit;
+	bool m_cassold;
```
cuavas, line 47: Please don’t split member initialisation between the declaration and a constructor in a separate file (`m_cass_data` is being intialised here).

rb6502:
As Vas said, you should still allow the user to set the unused DIP switches if they physically exist.  Other than that it looks like everything was addressed.

src/mame/heathkit/h89.cpp
```diff
@@ -291,16 +336,57 @@ void h89_state::h89_io(address_map &map)
...
+}
+
+
 // Input ports
+static INPUT_PORTS_START( h88 )
+
+	PORT_START("SW501")
+	// MTR-88  (444-40)
+	PORT_DIPNAME( 0x1f, 0x00, DEF_STR( Unused ) )               PORT_DIPLOCATION("SW501:1,2,3,4,5")
+	PORT_DIPSETTING( 0x00, "Undefined" )
```
cuavas, line 371: The convention is to have individual switches for the unused DIP switches.  You can use `PORT_DIPUNUSED_DIPLOC` to reduce the number of lines a bit, e.g.
```c++
	PORT_DIPUNUSED_DIPLOC(0x01, 0x00, "SW501:1")
	PORT_DIPUNUSED_DIPLOC(0x02, 0x00, "SW501:2")
	PORT_DIPUNUSED_DIPLOC(0x04, 0x00, "SW501:3")
	PORT_DIPUNUSED_DIPLOC(0x08, 0x00, "SW501:4")
	PORT_DIPUNUSED_DIPLOC(0x10, 0x00, "SW501:5")
```

mgarlanger:
> As Vas said, you should still allow the user to set the unused DIP switches if they physically exist. Other than that it looks like everything was addressed.

That was addressed. Once I added the `0x00 Undefined`option, that unlocked all 5 switches.  When changing the actual switches at the bottom, it shows `INVALID` at the top, but it still saves those settings. With 5 bits, I didn't think it was worth it to add 32 `Undefined` options, or break the 5 bits into individual bit and add a bunch more lines of code. Is there a different way to handle it?

Screenshot 2024-01-07 at 9 40 04 AM

mgarlanger:
Oh, I see Vas pointed out how to do it with minimal lines. I can do that next time, since he also made that change. Thanks.

PR #11899, opened 2024-01-04
jvc_dsk, os9_dsk.cpp: Add 3.5 inch disk formats, adjust scoring

tlindner:
The CoCo driver includes a 3.5 inch disk drive option, but the specific formats were missing from the os9 format list.
The scoring was adjust so the OS-9 format would prevail in a fight against JVC.

### DIFF
--- a/src/lib/formats/jvc_dsk.cpp
+++ b/src/lib/formats/jvc_dsk.cpp
@@ -170,9 +170,9 @@ bool jvc_format::parse_header(util::random_read &io, int &header_size, int &trac
 		break;
 	}
 
-	if (tracks>82)
+	if (tracks > 82)
 	{
-		osd_printf_verbose("jvc_format: track count of %d unsupported\n", tracks);
+		osd_printf_info("jvc_format: track count of %d unsupported\n", tracks);
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


npwoods (approved):
LGTM

src/lib/formats/os9_dsk.cpp
```diff
@@ -76,7 +76,7 @@ int os9_format::identify(util::random_read &io, uint32_t form_factor, const std:
 	int const type = find_size(io, form_factor, variants);
 
 	if (type != -1)
-		return FIFID_SIZE;
+		return FIFID_STRUCT | FIFID_SIZE;
```
galibert, line 79: Please don't lie in those returns.  You're not checking anything of the structure of the file, only its size.

If you want to give a format priority over the others in sych a context, use FIFID_HINT.

src/lib/formats/jvc_dsk.cpp
```diff
@@ -170,6 +170,12 @@ bool jvc_format::parse_header(util::random_read &io, int &header_size, int &trac
 		break;
 	}
 
+	if (tracks>82)
+	{
+		osd_printf_verbose("jvc_format: track count of %d unsupported\n", tracks);
+		return false;
+	}
```
cuavas, line 177: Please put spaces around binary operators like the rest of the code in this file.

startaq: Also it should probably use `osd_printf_info` like the other error messages.

tlindner: I switched to `info` and it appears during a failed probe. This is confusing:
`[MINGW64] C:\Users\Public\msys64\src\mame>mame coco3 -flop1 n80d.os9`
`jvc_format: track count of 160 unsupported`
I though JVC would be filtered out by the `.os9` file extension. But it doesn't appear to be so.

PR #11904, opened 2024-01-05
bus/a2bus/vista800.cpp: update dma priority and side select status

robjustice:
The schematic turned up after the original commit, so I was able to address the TODO items.
 - dma read has priority if both read and write are enabled.
 - the side select bit in the status reg is connected to the twosided signal from the drive, and not from the sso signal from the fdc.

### DIFF
--- a/src/devices/bus/a2bus/vistaa800.cpp
+++ b/src/devices/bus/a2bus/vistaa800.cpp
@@ -9,6 +9,8 @@
 	This supported up to four double sided/double density 8inch drives.
 	With DMA support for the data transfers, and booting from the first 8inch drive.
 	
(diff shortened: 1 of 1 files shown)


cuavas (changes requested):
Keeping a pointer is going to cause save states not to work properly.  You need to either add a post-load handler to restore the pointer, or implement this some other way.

simzy39:
I apologise if this is silly, but is there a way to get the year of when this card came out into the driver? The manual is from 1981.

robjustice:
The schematic is dated 19 mar 81, and the card is marked copyright 1981, so it does seem it was released in 81. I added a comment in for this, and did some other updates for the feedback.

PR #11914, opened 2024-01-06
new clones marked as NOT WORKING

mamehaze:
new clones marked as NOT WORKING
----------
Operation Tiger (Ver 2.09 J) [ethteck]


mamehaze:
this version hangs when you coin it up, otherwise runs the demo mode.
I don't think it's due to the io/sound skip patch as that basically looks the same

angelosa:
I guess it needs either sound comms working (cfr. pf2012 needing an irq latched from coin chutes) or video legacy hookup concealing a buffer swap/FIFO busy flag.

PR #11915, opened 2024-01-07
Added LD-V1000 HLE device, placeholder SSI-263A HLE device, and promoted Thayer's Quest

MooglyGuy:
Systems promoted to working
---------------------------
Thayer's Quest (set 1) [Ryan Holtz, Matt Ownby, ld-decode Team]

Clones promoted to working
--------------------------
Thayer's Quest (set 2) [Ryan Holtz, Matt Ownby, ld-decode Team]

-thayers: Added CHD and promoted to working. [Ryan Holtz]
 * Fixed periodic IRQ hookup.
 * Fixed COP421 clock divisor.
 * Switched to LD-V1000 by default and removed LD-PR7820 support for now.
 * Switched to using logmacro and shorthand data types.
 * Adjusted IRQ triggering and acknowledgement according to schematics.
 * Extracted SSI-263 skeleton into its own device using the Votrax SC-01 as a stand-in.

-ldv1000hle: Added an HLE version of the Pioneer LD-V1000 laserdisc player. [Ryan Holtz]

-laserdsc: Added a general-purpose parallel laserdisc interface, to have a common class parent for LD-PR7820. [Ryan Holtz]

-ssi263hle: Added a temporary SSI-263 device which remaps SC-02 phonemes onto the SC-01's phoneme set. [Ryan Holtz]

### DIFF
--- a/src/devices/machine/laserdsc.h
+++ b/src/devices/machine/laserdsc.h
@@ -221,7 +221,7 @@ protected:
 	// subclass helpers
 	void set_audio_squelch(bool squelchleft, bool squelchright) { m_stream->update(); m_audiosquelch = (squelchleft ? 1 : 0) | (squelchright ? 2 : 0); }
 	void set_video_squelch(bool squelch) { m_videosquelch = squelch; }
-	void set_slider_speed(int32_t tracks_per_vsync);
+	void set_slider_speed(const double tracks_per_vsync);
 	void advance_slider(int32_t numtracks);
 	slider_position get_slider_position();
 	int32_t generic_update(const vbi_metadata &vbi, int fieldnum, const attotime &curtime, player_state_info &curstate);
@@ -304,7 +304,7 @@ private:
 
 	// audio data
 	sound_stream *      m_stream;
-	std::vector<int16_t>       m_audiobuffer[2];       // buffer for audio samples
+	std::vector<int16_t> m_audiobuffer[2];      // buffer for audio samples
 	uint32_t            m_audiobufsize;         // size of buffer
 	uint32_t            m_audiobufin;           // input index
 	uint32_t            m_audiobufout;          // output index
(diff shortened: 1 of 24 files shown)


src/devices/machine/ldv1000hle.h
```diff
@@ -0,0 +1,201 @@
...
+	u32                 m_cmd_number_length;
+
+	u16                 m_user_ram[1024];
+	u32                 m_curr_register;
+	u32                 m_scan_speed;
+	u32                 m_scan_speed_accum;
+	double				m_play_speed;
+	bool                m_audio_enable[2];
+	bool                m_status_strobe;
+	bool                m_command_strobe;
```
cuavas, line 198: Can you srcclean this header to reduce some churn later?

src/devices/machine/laserdsc.h
```diff
@@ -335,6 +335,25 @@ typedef device_interface_enumerator<laserdisc_device> laserdisc_device_enumerato
...
+	virtual void enter_w(int state) {}
+	virtual int data_available_r() { return CLEAR_LINE; }
+	virtual int status_strobe_r() { return CLEAR_LINE; }
+	virtual int ready_r() { return ASSERT_LINE; }
+
+protected:
+	parallel_laserdisc_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock = 0);
+
+	virtual void player_overlay(bitmap_yuy16 &bitmap) override { }
+};
```
cuavas, line 355: I get that this is to simplify supporting multiple player types in a single driver.  But how artificial is it to make the players with parallel interfaces conform to a single interface?  (Honest question, I haven’t really looked at them.  I’m more familiar with Pioneer’s later “System Remote” interfaces for controlling players.)

The other alternative to avoid piles of `if (m_optional_player_device) ...` would be to have delegates or devcbs in the driver and hook them up in the machine configuration function when the player is added.

cuavas (commented):
Looks pretty reasonable.

In general, you’ve been a bit inconsistent with whether you put the explicit `virtual` prefix qualifier on overridden virtual member functions in addition to the `override` suffix qualifier.  Yes, `override` implies `virtual`, but in MAME we generally explicitly include the `virtual` qualifier as well.

Also, a non-dealbreaking thing – I noticed you used lowercase integer literal suffixes (e.g. `1u`).  I tend to prefer uppercase (e.g. `1U`) due to the long and long long suffixes being potentially confused for 1 digits if they’re lowercase (`1LL` is less likely to be confused with `111` than `1ll` at a glance).  By extension I use uppercase for the unsigned and single precision literal suffixes as well.

src/devices/machine/laserdsc.h
MooglyGuy: A bit artificial, but not completely. Two of the lines are mutually exclusive between the LD-V1000 and LD-PR7820. I don't feel strongly one way or another about it, but I'm leaning towards letting it ride until we can either change the player type based on machine-configuration switches, and there's working player LLE (or working LD-PR7280 HLE).

Issue #10819, opened 2023-01-10
Request to add ZX SPECTRUM NEXT models


denjhang:
This model has many interesting programs

cuavas:
Do you have relevant info that would allow people to emulate these systems?

denjhang:
https://wiki.specnext.dev/Emulators
https://github.com/mikedailly/CSpectPlugins

denjhang:
https://emutopia.com/index.php/emulators/item/364-sinclair-zx-spectrum/1895-cspect

PR #11918, opened 2024-01-08
New WORKING - sinclair/specnext.cpp: ZX Spectrum Next (known as tbblue)

holub:
System is bootable now but missing a lot of things to be completed.

Implementation intended to be "translation" (uses same var names and logical blocks even if sometimes more reasonable solution exists) from FPGA sources available: https://gitlab.com/SpectrumNext/ZX_Spectrum_Next_FPGA
Required SD image from: https://gitlab.com/thesmog358/tbblue

I'm open to any stylistic changes.

- fix: #10819


src/mame/sinclair/specnext_sprites.h
```diff
@@ -0,0 +1,87 @@
...
+        bool rotate;
+        bool xmirror;
+        bool ymirror;
+        u8 xscale : 2;
+        u8 yscale : 2;
+    };
+
+    std::vector<sprite_data> m_sprites_cache;
+    std::pair<u16, u16> m_offset = {0, 0};
+    rectangle m_clip_window = SCREEN_AREA;
```
cuavas, line 65: Don’t initialise device members in headers.

src/mame/sinclair/specnext_sprites.h
```diff
@@ -0,0 +1,87 @@
+#ifndef MAME_SINCLAIR_SPECNEXT_SPRITES_H
+#define MAME_SINCLAIR_SPECNEXT_SPRITES_H
+
+#pragma once
+
+#include "tilemap.h"
```
cuavas, line 6: This doesn’t appear to be used.

src/mame/sinclair/specnext_sprites.h
cuavas, line 1: This file is indented with spaces – MAME uses tabs.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+TODO:
+- Relative Sprite, Composite Type
+- Secondary Palette
+
+**********************************************************************/
+
+#include "emu.h"
+#include "screen.h"
+
+#include "specnext_sprites.h"
```
cuavas, line 15: The first thing you `#include` after the prefix header should be the module header.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+static GFXDECODE_START( gfx_sprites )
+    GFXDECODE_DEVICE_RAM( "pattern_ram", 0, gfx_16x16x8, 0x600, 256 )
+    GFXDECODE_DEVICE_RAM( "pattern_ram", 0, gfx_16x16x4, 0x600, 16 )
+    GFXDECODE_DEVICE_RAM( "pattern_ram", 0, gfx_16x16x8_r, 0x600, 256 )
+    GFXDECODE_DEVICE_RAM( "pattern_ram", 0, gfx_16x16x4_r, 0x600, 16 )
+GFXDECODE_END
+
+specnext_sprites_device::specnext_sprites_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
+	: device_t(mconfig, SPECNEXT_SPRITES, tag, owner, clock)
+    , device_gfx_interface(mconfig, *this, gfx_sprites, ":palette")
```
cuavas, line 62: Don’t make assumptions about other devices in the system.  Provide a means of configuration if necessary.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+    , m_sprite_attr_ram(*this, "attr_ram", 128 * 8, ENDIANNESS_LITTLE)
+{
+}
+
+void specnext_sprites_device::draw(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect)
+{
+    if (m_sprites_cache.empty()) update_sprites_cache();
+    const rectangle clipped = m_clip_window & cliprect;
+
+    for (auto i = 0; i < m_sprites_cache.size(); i++ ) {
```
cuavas, line 73: The other function uses Allman formatting for its `for` loop, but this is K&R – please be consistent.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+        const u8 x = m_zero_on_top ? i : (m_sprites_cache.size() - 1) - i;
+        const sprite_data* spr = &m_sprites_cache[x];
```
cuavas, line 75: You could use a reference rather than a pointer since you never index it.

src/mame/sinclair/specnext_sprites.h
```diff
@@ -0,0 +1,87 @@
...
+        bool h;
+        u16 x : 9;
+        u16 y : 9;
+        u8 pattern : 7;
+        u8 paloff : 4;
+        bool rotate;
+        bool xmirror;
+        bool ymirror;
+        u8 xscale : 2;
+        u8 yscale : 2;
```
cuavas, line 60: Please don’t use bitfields.  They just end up causing problems.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+        else
+            anchor_vis = is_visible;
+
+        if (is_visible)
+        {
+            u8 spr_cur_attr[5] = {};
+            if (!spr_relative) // Anchor
+            {
+                memcpy(spr_cur_attr, sprite_attr, 5);
+                anchor = &m_sprites_cache.back() + 1;
```
cuavas, line 109: Taking a pointer to a member of a vector that doesn’t exist is pretty bad form.  Since you appear to always have 128 sprites, why are you using a vector in the first place?  You could just use an array, or a dynamically allocated `std::array`, or something.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+            }
+            else
+            {
+                const u8 spr_rel_x0 = anchor->rotate ? sprite_attr[1] : sprite_attr[0];
```
cuavas, line 113: I don’t see anything to prevent `spr_relative` from being set for the first visible sprite, which will cause `anchor` to be dereferenced while uninitialised.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+void specnext_sprites_device::mirror_data_w(u8 mirror_data)
+{
+    bool mirror_num_change = 0;
+    if (m_mirror_index == 0b111)
+    {
+        m_mirror_sprite_q = mirror_data;
+        mirror_num_change = 1;
+    }
+    else if (m_mirror_inc) // ...
```
cuavas, line 234: What’s the `...` comment supposed to mean?  This doesn’t tell the person reading the code anything, but suggests there’s something they’re supposed to know.

src/mame/sinclair/specnext_sprites.cpp
```diff
@@ -0,0 +1,282 @@
...
+    save_item(NAME(m_clip_y2));
+
+    save_item(NAME(m_mirror_tie));
+    save_item(NAME(m_mirror_index));
+    save_item(NAME(m_mirror_inc));
+
+    save_item(NAME(m_attr_index));
+    save_item(NAME(m_pattern_index));
+    save_item(NAME(m_mirror_sprite_q));
+}
```
cuavas, line 268: You’ll need to ensure the sprite cache is rebuilt after a saved state is restored.

src/mame/sinclair/specnext_sprites.cpp
cuavas, line 1: This file has inconsistent indentation.

src/mame/sinclair/specnext_layer2.h
cuavas, line 1: This file has inconcistent indentation.

src/mame/sinclair/specnext_layer2.h
```diff
@@ -0,0 +1,59 @@
...
+        u8 palette_offset : 4;
+        u8 layer2_active_bank : 7;
+
+        u16 scroll_x : 9;
+        u8 scroll_y;
+        u8 clip_x1;
+        u8 clip_x2;
+        u8 clip_y1;
+        u8 clip_y2;
+    } m_ports;
```
cuavas, line 54: Please don’t implement virtual member functions in headers, initialise device members in headers, or use bitfields.

src/mame/sinclair/specnext_layer2.cpp
cuavas, line 1: This file has inconsistent indentation.

src/mame/sinclair/specnext_layer2.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include "emu.h"
+#include "screen.h"
+
+#include "specnext_layer2.h"
+
+specnext_layer2_device::specnext_layer2_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
+	: device_t(mconfig, SPECNEXT_LAYER2, tag, owner, clock)
+    , device_gfx_interface(mconfig, *this, nullptr, ":palette")
+    , m_ram(*this, "^ram")
```
cuavas, line 16: Don’t make assumptions about the other devices in the system.

src/mame/sinclair/specnext_divmmc.h
cuavas, line 1: This file has inconsistent indentation.

src/mame/sinclair/specnext_divmmc.cpp
cuavas, line 1: This file has inconsistent indentation.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+		, m_io_mouse(*this, "mouse_input%u", 1U)
+	{}
+
+	void tbblue(machine_config &config);
+
+protected:
+	virtual void machine_start() override;
+	virtual void machine_reset() override;
+	virtual void video_start() override;
+	virtual void spectrum_128_update_memory() override {};
```
cuavas, line 91: Superfluous semicolon here.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	void spi_miso_w(u8 data);
+	void palette_val_w(u16 nr_palette_value);
+	void ula_palette_w(u8 data);
+
+	void mem_update();
+	void bank_update(u8 bank);
+	void memory_change(u16 port, u8 data);
+	u16 get_layer2_active_page(u8 bank);
+	virtual u32 screen_update_spectrum(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect) override;
+	virtual void spectrum_update_screen(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect) override {}; // ULA screen just border
```
cuavas, line 113: Superfluous semicolon here.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	static const u8 G_VIDEO_INC = 0b11;
+	static const u16 UTM_FALLBACK_PEN = 0x800;
+
+	TIMER_CALLBACK_MEMBER(spi_clock);
+
+	u8 g_machine_id() { return m_io_issue->read() ? MACHINE_NEXT : MACHINE_TBBLUE; }
+	u8 g_board_issue() { return m_io_issue->read(); }
+	constexpr bool machine_type_48() { return m_regs.nr_03_machine_type == 0 || m_regs.nr_03_machine_type == 1; }
+	constexpr bool machine_type_128() { return m_regs.nr_03_machine_type == 2 || m_regs.nr_03_machine_type == 4; }
+	constexpr bool machine_type_p3() { return !machine_type_48() && !machine_type_128(); }
```
cuavas, line 132: Do `constexpr` member functions make sense at all for a device class?  You can’t have have compile time constant instances of it.  Wouldn’t it make more sense to just make them `const` qualified?

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	u8 m_port_e3_reg;
+	bool m_divmmc_active;
+	u16 m_global_transparent;
+
+	u8 m_sram_rom;
+	bool m_sram_rom3;
+	bool m_sram_alt_128_n;
+
+	struct specnext_regs
+	{
```
cuavas, line 280: Please don’t use bitfields – they cause too much trouble.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	// m_ram->set_default_size("1M").set_extra_options("2M");
+	m_ram->set_default_size("2M");
+
+	Z80N(config.replace(), m_maincpu, X1_128_SINCLAIR / 10);
+	m_maincpu->set_m1_map(&specnext_state::map_fetch);
+	m_maincpu->set_memory_map(&specnext_state::map_mem);
+	m_maincpu->set_io_map(&specnext_state::map_io);
+	m_maincpu->set_vblank_int("screen", FUNC(specnext_state::spec_interrupt)); // ...
+	m_maincpu->out_nextreg_cb().set(FUNC(specnext_state::reg_w));
+	m_maincpu->in_nextreg_cb().set(FUNC(specnext_state::reg_r));
```
cuavas, line 2698: Would it make more sense to have an additional address CPU address space for these, and use individual handler functions?  The huge switch statements in functions aren’t great form.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	m_regs.nr_02_iotrap = BIT(nr_wr_dat, 4);
+	m_regs.nr_02_generate_mf_nmi = BIT(nr_wr_dat, 3);
+	if (!m_regs.nr_02_generate_mf_nmi)
+		;
+	m_regs.nr_02_generate_divmmc_nmi = BIT(nr_wr_dat, 2);
+	if (!m_regs.nr_02_generate_divmmc_nmi)
+		;
+
+	if (BIT(nr_wr_dat, 1))
+		machine().schedule_hard_reset();
```
cuavas, line 1957: “Hard reset” tears down the emulation session and starts the machine again.  Is that really what this should do?  It doesn’t really correspond to a real life thing.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+	else if (BIT(nr_wr_dat, 0))
+	{
+		m_regs.nr_02_reset_type = (BIT(m_regs.nr_02_reset_type, 2) << 1) | BIT(m_regs.nr_02_reset_type, 1) | BIT(m_regs.nr_02_reset_type, 0);
+		machine().schedule_soft_reset();
```
cuavas, line 1961: You shouldn’t generally call `schedule_soft_reset()` – you should reset the root device and let it propagate through the device tree.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+
+		// Load snapshot after reboot from bootrom
+		snapshot_image_device *snapshot = subdevice<snapshot_image_device>("snapshot");
+		snapshot->set_delay(attotime::zero);
+		snapshot->call_load();
```
cuavas, line 1966: You really shouldn’t invoke `call_load` directly – it’s an internal part of the media image device lifecycle.

src/mame/sinclair/specnext.cpp
```diff
@@ -0,0 +1,2757 @@
...
+}
+
+void specnext_state::map_fetch(address_map &map)
+{
+	map(0x0000, 0xffff).lr8(NAME([this](offs_t offset)
+	{
+		if (m_divmmc_active /*&& !machine().side_effects_disabled()*/)
+		{
+			/* Happens after RW cycles. Ignoring side effects check is correct, because doesn't matter who reset this lines
+				and such approach gives better experience in debugger UI. */
```
cuavas, line 1994: How is the experience “better”?  Why should having a memory window open showing the M1 cause this to happen every time the window is updated?

holub:
Thank you @cuavas for the prompt review. I'll sort it out as possible.

The main concern is bitfields. They very helpful in this shape as didn't required value masking (I didn't figure out how to save them though).
As I need to remove them what would be a better approach to preserve visible size of the field? Possibly we can use names like m2_* or m_*_2. Last one will interfire with some existed, but not such a big deal.

scripts/src/cpu.lua
```diff
@@ -2936,9 +2939,15 @@ if CPUS["KC80"] then
...
-if want_disasm_z80 or want_disasm_kc80 then
+if want_disasm_z80 or want_disasm_z80n or want_disasm_kc80 then
+	if want_disasm_z80n then
+		table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/z80/z80dasm_menmonic_ed_n.cpp")
+	else
+		table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/z80/z80dasm_menmonic_ed.cpp")
+	end
 	table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/z80/z80dasm.cpp")
 	table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/z80/z80dasm.h")
 end
```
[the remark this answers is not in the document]
holub: That's why I had to split ed* tables. z80 loads old table and extended still invalid, and z80n loads extended.

src/mame/sinclair/specnext.cpp
holub: Because this condition becomes false after first call.
Alternative in debugger would be sowing next command which not going to be executed next and before execution bank view is switched to some else.

src/mame/sinclair/specnext.cpp
holub: I found it's easier to modify such form. Address space will require function for each entry which makes code much more verbose. Currently I'm using lambdas if "function" is small enough but don't really like this way because the code looks inconsistent.

src/mame/sinclair/specnext.cpp
holub: What is an alternative? The Next bootrom first which has simple loader from sdcard and does system configuration. We need to wait till this is done and system reaches "normal operation" mode and loaded roms before use the snappshot. Without this we don't even have access to the roms and palettes are not configured.

src/mame/sinclair/specnext.cpp
holub: The next has two programmable resets hard and soft. That fits into MAME's ideology and I was thinking can be used this way. Maybe I'll reconsider and handle both somehow in softreset based on a condition.

src/mame/sinclair/specnext.cpp
holub: Screenshot_2024-01-09-02-37-45-71_3aea4af51f236e4932235fdada7d1643~2

Plus FPGA implements this as a gigantic case stetment, hance it's easier to track any further updates.

src/mame/sinclair/specnext.cpp
rb6502: It fits MAME's terminology, but MAME's hard reset is by design something completely outside of the emulation, kind of like unplugging the machine entirely.  What are the soft and hard resets on the machine?  For instance, on the Apple //e the soft and hard resets both pull the CPU RESET line and the firmware does different things depending on which type it is.

src/mame/sinclair/specnext_sprites.cpp
holub: Not always. We have 128 at max, but only anchors and visible relative are stored in the cache.
With array I need to introduce extra state variable for size.

src/mame/sinclair/specnext_sprites.cpp
holub: anchor_vis init as false
relative's visibility dependent on anchors' so we'll never process invisible relative

scripts/src/cpu.lua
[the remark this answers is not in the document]
holub: Tested... you are right. I didin't find a way to have both z80 and z80n dasm happy. Please advise.

src/mame/sinclair/specnext.cpp
holub: Keeping commented for the future investigation

src/mame/sinclair/specnext_tiles.cpp
```diff
@@ -0,0 +1,111 @@
...
+        gfx(i)->set_source(m_tiles_info + ((m_tm_tile_base & 0x3f) << 8));
+        m_tilemap[i]->set_palette_offset(0x100 * BIT(m_control, 4));
+        m_tilemap[i]->set_transparent_pen(m_transp_colour);
+        m_tilemap[i]->set_scrollx(-m_offset.first - m_tm_scroll_x);
+        m_tilemap[i]->set_scrolly(-m_offset.second - m_tm_scroll_y);
+        m_tilemap[i]->mark_mapping_dirty();
+        m_tilemap[i]->mark_all_dirty();
+    }
+
+    m_tiles_info += (m_tm_map_base & 0x3f) << 8;
```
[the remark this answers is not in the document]
holub: Changed a bit. Hopefully it's clearer now.

holub:
Finished update and left couple topics opened.
dasm is critical

src/devices/cpu/z80/z80dasm.cpp
```diff
@@ -22,6 +22,11 @@ static const char *const s_mnemonic[] =
 	"retn","rl"  ,"rla" ,"rlc" ,"rlca","rld" ,"rr"  ,"rra" ,
 	"rrc" ,"rrca","rrd" ,"rst" ,"sbc" ,"scf" ,"set" ,"sla" ,
 	"sll" ,"sra" ,"srl" ,"sub" ,"xor "
+	// z80n
```
holub, line 25: As long as we have separate z80ndasm, it's possible worth move ext there.

holub:
@cuavas any chance to revisit this? PR becomes tremendous which I tried to avoid in the first place.
regarding opened requests:
* snapshot - completely removed (NextZXOS allows to load sna anyway)
* allow debugger to skip side-effect check in order to handle "before M1" - disabled
* regs switch case - I prefer to keep it as is for the reason described above, but will change to map if you insist

holub:
Thank you! I truly appreciate it!!!

holub:
As a next step I'm planning to work on graphics improvments: layers, intelligent updates, blendinding.
That's not going to be a part of this PR.

holub:
divemmc can possibly also go from this PR similar to the rest

holub:
continuation of the split in case it can be helpful: #12210

holub:
Also merged `z80-gen` branch with `next` in order to show end state:
https://github.com/holub/mame/compare/z80-gen...next-n-gen

holub:
Squashed and upgraded to the new z80 core

PR #11929, opened 2024-01-14
misc/sttechno: Implement driver for Shamisen Brothers and promote to working

987123879113:
Old driver misc/shambros.cpp has been renamed to misc/sttechno.cpp. It's unknown if/what other games run on this PCB still, but the logo "ST-Techno" is visible in 3 different places on the PCB and variations of the "ST" prefix are used in multiple places (STV01 MainPCB, STT-GA1, STT-SA1) so it seems like a better driver name.


src/devices/sound/stt_sa1.cpp
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+uint16_t stt_sa1_device::key_r()
+{
+	if (!m_enabled)
+		return 0;
+
+	m_stream->update();
+	return m_keyctrl;
```
cuavas, line 109: Is it possible to get here from a memory read handler?  Does it need a check for disabled side effects before updating the stream?

src/devices/sound/stt_sa1.cpp
```diff
@@ -0,0 +1,210 @@
...
+	if (offset >= 0x600000)
+		return m_readsample_cb(offset); // Held externally in flash chips
+
+	return m_ram[offset];
+}
+
+void stt_sa1_device::device_start()
+{
+	m_stream = stream_alloc(0, 2, clock() / 448);
+	m_ram = make_unique_clear<uint8_t[]>(0x600000); // 12 512k x 8 SRAM chips
```
cuavas, line 146: Is the RAM logically words or bytes?

Alternatively, you could store the RAM as a `std::unique_ptr<uint16_t []>`:
* No need to double the offset and use `get_u16be`/`put_u16be` when reading/writing.
* Use `util::big_endian_cast<uint8_t>(m_ram.get())[offset]` (from endianness.h) in `read_sample`.

src/devices/sound/stt_sa1.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void stt_sa1_device::sound_stream_update(sound_stream &stream, std::vector<read_stream_view> const &inputs, std::vector<write_stream_view> &outputs)
+{
+	outputs[0].fill(0);
+	outputs[1].fill(0);
+
+	for (int v = 0; v < 8; v++)
+	{
+		voice_t *voice = &m_voice[v];
```
cuavas, line 187: You can use a reference for `voice` (like `voice_t &voice = m_voice[v];`) since you shouldn’t need to index or modify the pointer.

src/devices/sound/stt_sa1.cpp
cuavas, line 1: This file has a mixture of K&R and Allman brace positioning for the `if`, `switch`, `for` etc. statements – please choose a style and stick to it for the whole file.

src/devices/sound/stt_sa1.h
```diff
@@ -0,0 +1,61 @@
...
+		uint16_t vol_l;
+		uint16_t vol_r;
+		uint16_t freq;
+		bool is_looped;
+		bool enabled;
+	};
+
+	sound_stream *m_stream;
+
+	devcb_read8 m_readsample_cb;
```
cuavas, line 48: You could derive from `device_rom_interface` (from dirom.h) rather than using a callback, which would allow the client to use an address map.

src/mame/misc/sttechno.cpp
```diff
@@ -0,0 +1,608 @@
...
+		r = m_flash[1]->read_raw((offset - 0x800000) / 2);
+
+	return BIT(r, (offset & 1) * 8, 8);
+}
+
+TIMER_DEVICE_CALLBACK_MEMBER(sttechno_state::irq6_timer)
+{
+	// IRQ 6 is responsible for incrementing the in-game song sync timer
+	// If the IRQ is too fast or too slow then it effects a lot of gameplay-related things: the measure markers, the note placements, the timing judgements
+	// The animation during the gameplay where the character on the right side jumps up and down is also based on this IRQ
```
cuavas, line 324: “…it affects a lot…”

src/mame/misc/sttechno.cpp
```diff
@@ -0,0 +1,608 @@
...
+	// This gets called at two points:
+	// - Every time before the OBJ RAM gets updated
+	// - Every time before setting FM_BYTEPROGRAM for the flash
+
+	if (m_sttga1_video_flash_write_enable)
+		return 0;
+
+	// HACK: Copy the OBJ data into the unused second half of the OBJ RAM for rendering
+	// This helps fix some small graphical glitches (sound test menu's text flicker, flickering during initial installation when programming the video flash)
+	std::copy_n(std::begin(m_sttga1_ram_obj), 0x2000, std::begin(m_sttga1_ram_obj) + 0x2000);
```
cuavas, line 443: Remember to check `!machine().side_effects_disabled()` for reads with side effects.

src/devices/bus/ata/fx5400w.cpp
```diff
@@ -0,0 +1,41 @@
...
+	m_identify_buffer[ 38 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 39 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 40 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 41 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 42 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 43 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 44 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 45 ] = (' ' << 8) | ' ';
+	m_identify_buffer[ 46 ] = (' ' << 8) | ' ';
+}
```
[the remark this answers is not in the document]
987123879113: I primarily created the new device for the `virtual bool is_ready() override { return true; }` part. Everything derived from `atapi_hle_device` defaults to `false` except for the drives that override that, which I think may only be `gdrom_device`. Shamisen Brothers won't boot without `is_ready` returning true and I don't really understand why it's defaulted to false. I'm not against throwing in another configuration method to just change it for all `atapi_cdrom_device` devices.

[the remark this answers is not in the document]
cuavas: It’s also overridden to return true in src/devices/machine/atastorage.h (I think the ATA hard disk and CF card are derived from this).

I believe it indicates whether the device is ready for a DMA transfer (really descriptive name, right).  I think it returns `false` by default on the assumption that the simplest devices don’t support DMA, and hence an implementation for a device with DMA support should override it.  As for why it isn’t overridden for CD-ROM drives, who knows.

src/mame/misc/sttechno.cpp
987123879113: Thanks, I always forget that. I found another that needs it too (uart_data_r).

src/devices/bus/ata/fx5400w.cpp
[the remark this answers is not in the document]
987123879113: I'll just make it configurable then. Might be helpful for some other stubborn game in the future that way too.

src/devices/sound/stt_sa1.cpp
987123879113: The registers are 16-bit but the RAM chips themselves are 8-bit (`U31, U32, U33, U37, U38, U39, U40, U41, U44, U45, U47, U48 (HY628400): Toshiba TC554001, 512k x 8 SRAM`) which is why I implemented the RAM as `uint8_t` in the device here just to be consistent with the real hardware.

Would you consider using `uint16_t` for the RAM the better choice here, performance and such considered? It's all implemented in the FPGA and I can't verify anything beyond how the code uses the chip so I don't have a preference here as long as it's works.

cuavas: Oh dammit.  I see what I’m missing.

It isn’t that the RAM and Flash is behind the “sound chip” per se.  It’s just that it has an external 16-bit memory bus that’s attached to the RAM and Flash, and the the FPGA also allows the CPU to access a window into this space.

The “sound chip” device should only be handling its own register reads and writes, and it should derive from `device_rom_interface` to access its sample storage.  The driver can install the RAM and Flash in its address space.

Are you comfortable looking for examples of `device_rom_interface` and doing it, or do I need to have a crack at it in the next few days?

cuavas: Oh, and just to be explicit, the 8-bit RAMs can easily be wired onto a 16-bit bus, you just wire 6 of them to the upper byte and the other 6 to the lower byte.  Lane select is easy for byte writes because you can enable one or both of the chips.

987123879113: I gave it a shot. Everything seems to be working still. Checked endianness of both the RAM and data being read from flash and everything is looking how it should still.

https://github.com/mamedev/mame/pull/11929/commits/e142229da1507ca2af5cbde9f924447684318a4e

cuavas: I’ll do a bit of cleanup and push to this PR branch so you can re-test it.

cuavas:
@987123879113 can you please pull my changes and give it a bit of a test?  I did run it and it seemed to work, but I don’t know how to play and I can’t be bothered learning right now.

Also, given this is a rhythm game, do the guessed screen parameters and periodic interrupt warrant flagging it as imperfect timing, or are we good with just imperfect sound?

987123879113:
@cuavas I fixed some small things. The endianness of the data for the sound RAM is a little tricky. One of them needs to be swapped and I think the stuff being written into m_sound_ram is easier to implement unless there's an easy way to change endianness of the `intelfsh16_device::read_raw` mappings in `sound_map`.

As for MACHINE_IMPERFECT_TIMING, sure why not. The timings are best effort guesses right now and I think it's pretty close right now but there was still a lot of guessing involved there. If someone can ever figure out the real raw screen parameters and figure out exactly how IRQ6 is supposed to actually work then I think it can be removed.

Just to be completely transparent about how I'm verifying the endianness here:
In `:pcm_sound` memory space...
At 0x751b70 is the data that corresponds to `SE/a440_11.sdx` on the CD
image

At 0x800 is the data that corresponds to the start of the BGM samples for one of the channels (assumed L, doesn't matter for verification). You can go into the song selection screen and hover over Diamonds to make it load `MUSIC/diamonds.sdx` from the CD.
image

The SE files get written to the flash during the first boot. The BGMs are always read in from the CD.

Here's how it looks in the MAME debugger with this latest change.
image

cuavas:
Oh, weird – opposite endianness on Flash and RAM from the CPU’s point of view is not something I would have expected.

PR #11931, opened 2024-01-14
videopac.xml: Metadata cleaning

ArcadeShadow:
Replaced "US" abbreviation by "USA"


cuavas:
@happppp you were the last to do major work on this, weren’t you.  Is “US” an official region code used by Philips/Magnavox, or should it be standardised to USA?

happppp:
I'm not aware. Isn't the region for most English NTSC computer hardware North America? MAME has the habit of putting US(A), poor Canadians.

Anyway, US or USA, I don't mind either way.

PR #11933, opened 2024-01-14
sinclair/sprinter.cpp: fix INT; add cdrom, dvdrom

holub:


src/mame/sinclair/sprinter.cpp
```diff
@@ -1439,7 +1441,8 @@ void sprinter_state::video_start()
 static void sprinter_ata_devices(device_slot_interface &device)
 {
 	device.option_add("hdd", IDE_HARDDISK);
-	device.option_add("cdrom", ATAPI_CDROM);
+	device.option_add("cdrom", ATAPI_FIXED_CDROM);
+	device.option_add("dvdrom", ATAPI_FIXED_DVDROM);
```
cuavas, line 1445: Why are you using the fixed media devices here?  They’re intended for systems where the optical drive is not accessible to the user and hence media can’t be changed (e.g. an arcade game that uses CD-ROM media for storage, with the drive inside the cabinet where the player can’t access it).

holub: When we tested, the base one didn't behave properly. Software in Sprinter which access CDROM only works with this one.

holub: From my notes: Base doesn't run `if (m_media_change) {}`
Which after your comment suspected to be vice versa.

PR #11934, opened 2024-01-14
Update hikara.xml - added known cartridge list

mamehaze:


hash/hikara.xml
cuavas, line 1: Please srcclean this to avoid churn.

PR #11935, opened 2024-01-14
new machines marked as NOT WORKING - Super TV PC

mamehaze:
-----------------
Super TV PC (Epoch) [Team Europe, QUFB]

new NOT WORKING software list entries
-------------
super_tv_pc_cart.xml:
-------
Tottoko Hamtaro Ham Ham Ongaku Paradichu! [Team Europe, QUFB]
Doraemon Chō Nōryoku Game Collection [Team Europe, QUFB]
Meitantei Conan: Suiri Adventure [Team Europe, QUFB]
ECC Junior: Ason de Eigo dekichatta! [Team Europe, QUFB]
Pet Mobile Link [Team Europe, QUFB]

previous Super TV PC was identified as the Hello Kitty version and is now marked as such.

### DIFF
--- a/hash/super_tv_pc_cart.xml
+++ b/hash/super_tv_pc_cart.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+
+<softwarelist name="super_tv_pc_cart" description="Epoch Super TV PC cartridges">
+
+	<!-- these appear to require the system BIOS as they access above the 4MB mark for audio during the boot screen -->
+	<software name="hamster" supported="no">
+		<description>Hamster</description>
+		<year>2004</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x400000">
+				<rom name="c-01.u1" size="0x400000" crc="0accbdcc" sha1="24fd19a87256a3335b5d1e32f11c9171f7359f32"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="doraemon" supported="no">
+		<description>Doraemon</description>
+		<year>2004</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x400000">
+				<rom name="c-02.u1" size="0x400000" crc="99fb0088" sha1="b46400c850b08abd5c8ee649f483e0694b3eb059"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<!-- C-03 -->
+	
+	<software name="conan" supported="no">
+		<description>Detective Conan</description>
+		<year>2005</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x400000">
+				<rom name="c-04.u1" size="0x400000" crc="89b6ef86" sha1="1af69047508fce08fa949a7d821255e306a329a3"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="ecc" supported="no">
+		<description>ECC</description>
+		<year>2005</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x400000">
+				<rom name="c-05.u1" size="0x400000" crc="91848208" sha1="098db879c5b77b7af031baf95a2accff13b2b215"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<!-- C-06 -->
+
+	<software name="link" supported="no">
+		<description>Link</description>
+		<year>2006</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x400000">
+				<rom name="c-07.u1" size="0x400000" crc="8aeea23c" sha1="e89921fe7714c60dcd60cc29382268350ac6f3f5"/>
+			</dataarea>			
+		</part>
+	</software>
+	
+</softwarelist>
--- a/hash/super_tv_pc_cart.xml
+++ b/hash/super_tv_pc_cart.xml
@@ -6,7 +6,6 @@ license:CC0-1.0
 
 <!-- also known to exist
 Doraemon Excite Ping-Pong
-Double Mouse Party
 -->
 
 <softwarelist name="super_tv_pc_cart" description="Epoch Super TV PC cartridges">
@@ -36,8 +35,18 @@ Double Mouse Party
 		</part>
 	</software>
 
-	<!-- C-03 -->
-	
+	<software name="dblmouse" supported="no">
+		<!-- quickly goes off the rails after not even displaying the XaviX logo correctly and making some odd accesses, is this a bad dump? -->
+		<description>Double Mouse Party</description>
+		<year>2004</year>
+		<publisher>Epoch</publisher>
+		<part name="cart" interface="super_tv_pc_cart">
+			<dataarea name="prg" size="0x200000">
+				<rom name="c-03.u1" size="0x200000" crc="94634a7b" sha1="d69ec0a8c6061e749206927ee68dbcf852fcc03e" status="baddump"/>
+			</dataarea>
+		</part>
+	</software>
+		
 	<software name="conan" supported="no">
 		<description>Detective Conan: Suiri Adventure</description>
 		<year>2005</year>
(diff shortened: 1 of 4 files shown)


mamehaze:
(I won't be working on these, this was just done by request of Team Europe / QUFB)

mamehaze:
Boxart_Front
Top_Closed
Boxart
Cart+PCB
Boxart
Cart+PCB
Boxart
Cart+PCB
Boxart
Cart+PCB
Boxart2
Cart+PCB

mamehaze:
Obviously the titles in the software list could be improved based on the images here.

ajrhacker:
The system fullnames are also questionable. Are "Dogs" and "Hello Kitty" part of the actual titles here?

mamehaze:
The Hello Kitty release has Hello Kitty theming all over the box, so I'd assume so.
Dogs is what Team Europe gave me for the one they dumped.
There are multiple different versions of the base product, with different themes, and different internal programs, so that needs to be part of the description somehow.

FoxhackDN:
Hamster:
とっとこハム太郎 ハムハムおんがくぱらだいちゅ！
Tottoko Hamtaro Ham Ham Ongaku Paradichu! (Paradise, but with "chu" added at the end because of the way they talk.)

The spine only says Ham Ham Ongaku Paradichu though.

FoxhackDN:
Also, Epoch still has the website for the system up: https://epoch.jp/gt/stvpc/

The Doraemon and Hello Kitty sets were standard ones. Then they made the Link version which I think worked with mobile internet apps. The Hamster one Haze posted isn't listed there for some reason. It also has a page for a printer: https://epoch.jp/gt/stvpc/print.html

Their game list says there are two more games: Doraemon Excite Ping-Pong and Double Mouse Party.

hash/super_tv_pc_cart.xml
```diff
@@ -0,0 +1,79 @@
...
+		</part>
+	</software>
+
+	<!-- C-06 -->
+
+	<software name="link" supported="no">
+		<description>Pet Mobile Link</description>
+		<year>2006</year>
+		<publisher>Epoch</publisher>
+		<info name="alt_title" value="ペット ケータイリ ンク"/>
```
cuavas, line 71: There are spurious spaces in the Japanese title, and “リンク” is shown as a pronunciation guide on the box.  The title is “ペットケータイ Link” and it should be transliterated as “Pet Keitai Link”.

src/mame/tvgames/xavix_2002.cpp
```diff
@@ -346,13 +346,36 @@ void xavix_i2c_jmat_state::xavix2002_i2c_jmat(machine_config &config)
...
+
+	if (size != 0x400000)
+		return std::make_pair(image_error::UNSUPPORTED, "must be 0x400000 in size");
+
+	m_cart->rom_alloc(size, GENERIC_ROM16_WIDTH, ENDIANNESS_BIG);
+	m_cart->common_load_rom(m_cart->get_rom_base(), size, "prg");
+	memcpy(memregion("bios")->base(), m_cart->get_rom_base(), size);
+
+	return std::make_pair(std::error_condition(), std::string());
+}
```
cuavas, line 364: Since this only needs a single ROM region and doesn’t use `feature` elements, is there a reason not to support loading loose files?

Also the 16-bit big endian region is somewhat suspect given the Xavix CPUs are 8-bit little endian.

hash/super_tv_pc_cart.xml
cuavas, line 1: If you’re loading these into a 16-bit region little endian, you need to mark the `prg` regions as 16-bit little endian here to match or you’ll have endianness bugs.

But is loading them into a 16-bit big endian region correct?  IIRC the Xavix CPUs are 8-bit little endian.

hash/super_tv_pc_cart.xml
```diff
@@ -0,0 +1,79 @@
...
+-->
+
+<softwarelist name="super_tv_pc_cart" description="Epoch Super TV PC cartridges">
+
+	<!-- these appear to require the system BIOS as they access above the 4MB mark for audio during the boot screen -->
+	<software name="hamatro" supported="no">
+		<description>Tottoko Hamtaro Ham Ham Ongaku Paradichu!</description>
+		<year>2004</year>
+		<publisher>Epoch</publisher>
+		<info name="alt_title" value="とっとこハム太郎 ハムハムおんがくぱらだいちゅ！"/>
```
cuavas, line 19: “Paradaichu” (not “Paradichu”).

hash/super_tv_pc_cart.xml
```diff
@@ -0,0 +1,79 @@
...
+		</part>
+	</software>
+
+	<!-- C-03 -->
+	
+	<software name="conan" supported="no">
+		<description>Meitantei Conan: Suiri Adventure</description>
+		<year>2005</year>
+		<publisher>Epoch</publisher>
+		<info name="alt_title" value="名探偵コナン 推理アドベンチャー"/>
```
cuavas, line 45: It should be transliterated as “Detective Conan: Suiri Adventure” as indicated on the box.

cuavas:
> The Hello Kitty release has Hello Kitty theming all over the box, so I'd assume so. Dogs is what Team Europe gave me for the one they dumped. The picture is above, I don't read Japanese so I don't know if it has a specific subtitle.

There doesn’t seem to be any “dogs” branding on that one.  It seems to be the generic unit with no licensed IP branding.

hash/super_tv_pc_cart.xml
FoxhackDN: There's several games on the software list that are titled Meitantei Conan, and at least one I can see has "Detective Conan" printed on the game title like on this box (mconant1 on psx). I just went with the same naming format used there.

src/mame/tvgames/xavix_2002.cpp
mamehaze: I'd rather things just be added to the softlists, I consider loose-loading to be an anti-feature in cases like this.

I think the SuperXaviX are still 8-bit, although they do use 16-bit data bus for the ROMs (which is one way you can usually tell if something is SuperXaviX, there are 16 data pins on the ROM rather than 8, but it seems transparent to the operation otherwise)

hash/super_tv_pc_cart.xml
cuavas: Yeah, but there are examples that use “Detective Conan” as well, e.g. in the e-kara software lists.  If there’s an official Latin script rendition of the title, we use it.

src/mame/tvgames/xavix_2002.cpp
cuavas: > I'd rather things just be added to the softlists, I consider loose-loading to be an anti-feature in cases like this.

Loose software loading is always useful for people developing homebrew and doing A/B testing with Flash carts on real hardware to debug emulation.

> I think the SuperXaviX are still 8-bit, although they do use 16-bit data bus for the ROMs (which is one way you can usually tell if something is SuperXaviX, there are 16 data pins on the ROM rather than 8, but it seems transparent to the operation otherwise)

The trouble is, MAME requires the region to match the CPU’s address space, or you need shim functions to translate, and the CPU cores have 8-bit address spaces in MAME.  (Kind of similar to the situation where Virtual Boy needs 32-bit regions for what’s actually 16-bit ROM, and the mess with WonderSwan which has dynamic bus sizing that MAME can’t emulate efficiently.)

hash/super_tv_pc_cart.xml
FoxhackDN: I see! Okay, that's good to know for later reference. Those games will need a double check to see if they need renaming, but that's a thing for another day.

angelosa: fwiw mconant1 also have Detective on box/title/CD printed out, very typical Jp-to-Latin manga/anime naming export.

s-l1600

src/mame/tvgames/xavix_2002.cpp
mamehaze: I've changed it to 8 anyway, even if I have no confidence either way.

hash/super_tv_pc_cart.xml
```diff
@@ -0,0 +1,79 @@
...
+
+<!-- also known to exist
+Doraemon Excite Ping-Pong
+Double Mouse Party
+-->
+
+<softwarelist name="super_tv_pc_cart" description="Epoch Super TV PC cartridges">
+
+	<!-- these appear to require the system BIOS as they access above the 4MB mark for audio during the boot screen -->
+	<software name="hamatro" supported="no">
```
FoxhackDN, line 15: Sorry, could you please rename this to 'hamtaro' so it's spelled right?

mamehaze:
It was pointed out also that there are 2 different types of TV computer units from Epoch too.

The Hamutaro TV Computer is part of an older series, with Hello Kitty, Doraemon and Thomas units available, no 'Super' in the title.  Those ones probably don't take carts.  Possibly also they run on older hardware (standard XaviX not SuperXavix?)

The Super TV Computer was released as regular (dumped), Hello Kitty (dumped) and Doraemon (not dumped)  those take these carts and run on the SuperXaviX hardware.

src/mame/tvgames/xavix_2002.cpp
cuavas: You’re still using a 16-bit big endian intermediate region here.

PR #11936, opened 2024-01-15
higemaru: Add tile x/y flip, fixes hi-score screen

jotego:
Before

After

Notice the four small red dots.

### DIFF
--- a/src/mame/capcom/higemaru.cpp
+++ b/src/mame/capcom/higemaru.cpp
@@ -168,10 +168,10 @@ void higemaru_state::c800_w(uint8_t data)
 
 TILE_GET_INFO_MEMBER(higemaru_state::get_bg_tile_info)
 {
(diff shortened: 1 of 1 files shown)


src/mame/capcom/higemaru.cpp
```diff
@@ -169,9 +169,9 @@ void higemaru_state::c800_w(uint8_t data)
 TILE_GET_INFO_MEMBER(higemaru_state::get_bg_tile_info)
 {
 	int const code = m_videoram[tile_index] + ((m_colorram[tile_index] & 0x80) << 1);
-	int const color = m_colorram[tile_index] & 0x1f;
+	int const color = m_colorram[tile_index];
 
-	tileinfo.set(0, code, color, 0);
+	tileinfo.set(0, code, color&0x1f, TILE_FLIPYX((color & 0x60) >> 5));
```
einstein95, line 174: Can you put spaces around the `&` here?

PR #11940, opened 2024-01-17
bus/pci: convert opti82c861 to a pci_slot, add basic OpenHCI values

angelosa:


angelosa:
Opening PR because I don't have access to an iMac boot image atm, as far as I can tell the behaviour is unchanged.

PR #11942, opened 2024-01-17
handheld/hh_ht11xx.cpp: Add support for Brick Game

mamehaze:
new WORKING machines
---------
Block Game & Echo Key [azya52, David Haywood]

### DIFF
--- a/src/mame/handheld/hh_ht11xx.cpp
+++ b/src/mame/handheld/hh_ht11xx.cpp
@@ -42,33 +42,71 @@ class hh_ht11xx_state : public driver_device
 public:
 	hh_ht11xx_state(const machine_config &mconfig, device_type type, const char *tag) :
 		driver_device(mconfig, type, tag),
-		m_maincpu(*this, "maincpu"),
-		m_out_x(*this, "%u.%u", 0U, 0U),
-		m_in(*this, "IN%u", 1)
+		m_maincpu(*this, "maincpu")
 	{ }
 
 	virtual DECLARE_INPUT_CHANGED_MEMBER(input_wakeup);
 
+protected:
+
+	void mcfg_svg_screen(machine_config &config, u16 width, u16 height, const char *tag = "screen");
+
+	required_device<ht1130_device> m_maincpu;
+};
+
+class hh_ht1190_state : public hh_ht11xx_state
+{
+public:
+	hh_ht1190_state(const machine_config &mconfig, device_type type, const char *tag) :
+		hh_ht11xx_state(mconfig, type, tag),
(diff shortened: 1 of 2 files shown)


src/mame/handheld/hh_ht11xx.cpp
```diff
@@ -42,33 +42,72 @@ class hh_ht11xx_state : public driver_device
 public:
 	hh_ht11xx_state(const machine_config &mconfig, device_type type, const char *tag) :
 		driver_device(mconfig, type, tag),
-		m_maincpu(*this, "maincpu"),
-		m_out_x(*this, "%u.%u", 0U, 0U),
-		m_in(*this, "IN%u", 1)
+		m_maincpu(*this, "maincpu")
 	{ }
```
cuavas, line 46: Constructors of abstract classes should be protected (this class has no public machine configuration member functions and hence can’t be used as a state class directly.

cuavas (commented):
Please give your pull requests descriptive titles.  You currently haven several with no indication of what systems, or what kinds of systems, they add.

happppp:
Do you want "ga888" to be the shortname, or something that resembles "Block Game & Echo Key" more?

mamehaze:
the names on these things tend to be so generic I don't think it really matters.  ga888 just seems more concise in this case than trying to fit a bunch of generic descriptor words there.

ICEknigh7:
Also, "GA888" seems to be displayed quite prominently on the unit.
image

mamehaze:
I do wonder how the flashing blocks are meant to work though, they seem different in each game mode, but I'm finding the behavior in the first mode to be nonsensical.  Could be a dumping issue given these are visual decaps (or a CPU bug if it makes more sense in the other emulator, but I haven't tested it there)

(or I could just be being stupid and not understanding some simple game logic)

mamehaze:
ok, azya52 pointed out that the rules for the flashing blocks are here
https://datasheet.datasheetarchive.com/originals/library/Datasheet-036/DSA0020063.pdf

so it has basically nothing to do with the individual blocks, just each shape has a different special meaning.  no wonder I couldn't figured it out.

mamehaze:
likewise I've got another MameDev member saying this looks fine to them, but is also *afraid* to merge it because they might consider it a strike against them in terms of their direct access rights if there is something minor you're unhappy with.

etabeta78:
Have you considered the fact that several contributors do no raise all such
drama, when asked to modify their submissions, sometimes quite heavily?

Also, do you plan to transform each of you pull request into a fight? 🤔
Because I do not think it helps the project in any way, nor it helps you in
proving a point against Vas

My 2c

PR #11947, opened 2024-01-18
rc2014: fix callback initialization

mmicko:
All RC2014 related drivers were broken after https://github.com/mamedev/mame/pull/11333  so this is attempt to fix it.
By architecture this is most similar to S100 but did not see any callback logic like this in implementation there.

Please suggest if there are better ways to do it.

### DIFF
--- a/src/devices/bus/rc2014/micro.cpp
+++ b/src/devices/bus/rc2014/micro.cpp
@@ -8,13 +8,15 @@
 
 #include "emu.h"
 #include "micro.h"
+
 #include "modules.h"
 
+#include "bus/ata/ataintf.h"
+#include "bus/rs232/rs232.h"
 #include "cpu/z80/z80.h"
 #include "machine/6850acia.h"
 #include "machine/clock.h"
-#include "bus/ata/ataintf.h"
-#include "bus/rs232/rs232.h"
+
 
 namespace {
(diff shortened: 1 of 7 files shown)


mmicko:
Thanks for suggestion @cuavas. Main concern from beginning was speed. Most of signals would be fine, but there are also two clock signals on board that could come from clock generator card(s) or even some variants of CPU card, and those are driving serial boards (with ACIA or SIO).  For main clock (since only that one is used to drive CPU) I have used ability to set bus speed and use derive clock (just with factor 1,1) but that is not something I could use for 2nd clock signal. 
Correct me if I am wrong, your suggestion is to use virtual methods for cards, and keep list of connected cards in bus class and loop trough all of them on devcb write ?
Anyway will experiment to see what kind of performance impact it gives compared to current code.

cuavas:
Yeah, you can keep a vector of references to cards, and then iterate over it and call the virtual member function on each one.  I think bus/intellec4 does this for various signals.

If a signal is very high frequency, devcb doesn’t give the best performance either – it does have significant overhead.

mmicko:
@cuavas Have redone bus callback as virtual member functions

mmicko:
Thanks for cleanup and merge.

PR #11954, opened 2024-01-20
misc/sttechno: Implement xflip and yflip flags

987123879113:
Implemented xflip and yflip flags that I had previously noted but didn't know where to test in-game. The big mouse no longer moonwalks on the first stage.

Small cleanup: I noticed I had swapped `tile_x` and `tile_y` when renaming them from old temp variable names during cleanup so I fixed that too.


cuavas:
No moonwalking mouse?  Are you sure?

987123879113:
Sadly.

https://github.com/mamedev/mame/assets/63495610/7405ad5e-590d-434c-83a7-e93f1b8e63c7

Just to make sure the moonwalking mouse is not forgotten, here's how it was before:

https://github.com/mamedev/mame/assets/63495610/e8e053ec-c143-44b9-b660-50009661c8ad

PR #11959, opened 2024-01-20
machine/s2350.cpp: Add new AMI S2350 USRT, prep for Heath hard-sectored support

mgarlanger:
Adds the AMI's USRT S2350 USRT needed for the Heath hard-sectored disk controller. 

This is prep work for supporting the H-88-1 hard-sectored controller for the Heath H89.

### DIFF
--- a/src/devices/machine/s2350.cpp
+++ b/src/devices/machine/s2350.cpp
@@ -222,7 +222,7 @@ u8 s2350_device::receiver_sync_search()
 {
 	set_sync_character_received(false);
 	m_sync_search_active      = true;
-	
+
 	LOGREG("%s: 0x%02x\n", FUNCNAME, m_receiver_sync_reg);
 
 	return m_receiver_sync_reg;
@@ -239,14 +239,14 @@ u8 s2350_device::receiver_output_reg_r()
 
 u8 s2350_device::status_word_r()
 {
-    u8 status = 0x00;
+	u8 status = 0x00;
 
-    status |= m_received_data_available ? 0x01 : 0x00;
-    status |= m_receiver_overrun        ? 0x02 : 0x00;
-    status |= m_receiver_parity_error   ? 0x04 : 0x00;
-    status |= m_sync_character_received ? 0x08 : 0x00;
-    status |= m_fill_char_transmitted   ? 0x40 : 0x00;
-    status |= m_transmit_buffer_empty   ? 0x80 : 0x00;
+	status |= m_received_data_available ? 0x01 : 0x00;
+	status |= m_receiver_overrun        ? 0x02 : 0x00;
+	status |= m_receiver_parity_error   ? 0x04 : 0x00;
+	status |= m_sync_character_received ? 0x08 : 0x00;
+	status |= m_fill_char_transmitted   ? 0x40 : 0x00;
+	status |= m_transmit_buffer_empty   ? 0x80 : 0x00;
 
 	set_fill_char_transmitted(false);
 	set_receiver_overrun(false);
@@ -255,7 +255,7 @@ u8 s2350_device::status_word_r()
 
 	LOGREG("%s: 0x%02x\n", FUNCNAME, status);
 
-    return status;
+	return status;
 }
 
 void s2350_device::serial_rx_w(int state)
--- a/src/devices/machine/s2350.cpp
+++ b/src/devices/machine/s2350.cpp
@@ -248,7 +248,8 @@ u8 s2350_device::status_word_r()
 	status |= m_fill_char_transmitted   ? 0x40 : 0x00;
 	status |= m_transmit_buffer_empty   ? 0x80 : 0x00;
 
-	if (!machine().side_effects_disabled()) {
+	if (!machine().side_effects_disabled())
+	{
 		set_fill_char_transmitted(false);
 		set_receiver_overrun(false);
 		set_receiver_parity_error(false);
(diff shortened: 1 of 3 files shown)


src/devices/machine/s2350.cpp
```diff
@@ -0,0 +1,411 @@
...
+	status |= m_receiver_overrun        ? 0x02 : 0x00;
+	status |= m_receiver_parity_error   ? 0x04 : 0x00;
+	status |= m_sync_character_received ? 0x08 : 0x00;
+	status |= m_fill_char_transmitted   ? 0x40 : 0x00;
+	status |= m_transmit_buffer_empty   ? 0x80 : 0x00;
+
+	set_fill_char_transmitted(false);
+	set_receiver_overrun(false);
+	set_receiver_parity_error(false);
+	set_sync_character_received(false);
```
cuavas, line 254: Remember to check for disabled side effects before doing things with side effects in read handlers.  Failing to do this makes systems undebuggable.

src/devices/machine/s2350.cpp
```diff
@@ -0,0 +1,411 @@
...
+{
+	LOGREG("%s: 0x%02x\n", FUNCNAME, data);
+
+	m_receiver_sync_reg = data;
+}
+
+u8 s2350_device::receiver_sync_search()
+{
+	set_sync_character_received(false);
+	m_sync_search_active      = true;
```
cuavas, line 224: Can this line lose the extra spaces?  There are no other similar lines in this function for it to align with.

src/devices/machine/s2350.h
```diff
@@ -0,0 +1,121 @@
...
+	auto sync_character_received_cb()    { return m_scr.bind(); }
+	auto receiver_overrun_cb()           { return m_ror.bind(); }
+	auto receiver_parity_error_cb()      { return m_rpe.bind(); }
+
+	void receiver_restart();
+
+	// transmitter clock
+	void tcp_w();
+	// receiver clock
+	void rcp_w();
```
cuavas, line 52: Please put all the configuration member functions together above the “live” functions.

src/devices/machine/s2350.h
```diff
@@ -0,0 +1,121 @@
...
+	void serial_rx_w(int state);
+	void update_receiver_shift();
+
+	void set_transmit_buffer_empty(int val);
+	void set_fill_char_transmitted(int val);
+
+	void set_received_data_available(int val);
+	void set_receiver_overrun(int val);
+	void set_receiver_parity_error(int val);
+	void set_sync_character_received(int val);
```
cuavas, line 68: Can these use `bool` parameters?  They aren’t input line handler functions, so they don’t need to use the standard input line handler function signature.

src/devices/machine/s2350.h
```diff
@@ -0,0 +1,121 @@
...
+
+	void receive_byte(uint8_t data);
+
+	TIMER_DEVICE_CALLBACK_MEMBER(rx_timer_cb);
+	TIMER_DEVICE_CALLBACK_MEMBER(tx_timer_cb);
```
cuavas, line 73: These don’t seem to be defined anywhere.  Are they a leftover?

src/devices/machine/s2350.cpp
```diff
@@ -0,0 +1,413 @@
...
+	status |= m_sync_character_received ? 0x08 : 0x00;
+	status |= m_fill_char_transmitted   ? 0x40 : 0x00;
+	status |= m_transmit_buffer_empty   ? 0x80 : 0x00;
+
+	if (!machine().side_effects_disabled()) {
+		set_fill_char_transmitted(false);
+		set_receiver_overrun(false);
+		set_receiver_parity_error(false);
+		set_sync_character_received(false);
+	}
```
cuavas, line 256: K&R formatting snuck in (rest of the file is Allman).

cuavas (commented):
Can you please make this always log (e.g. with `logerror`) when the host system uses an unimplemented, like configuring a smaller word size or enabling parity?  It helps greatly when hooking a device up to a system if you get a notification when you run into a known limitation.

mgarlanger:
> Can you please make this always log (e.g. with `logerror`) when the host system uses an unimplemented, like configuring a smaller word size or enabling parity? It helps greatly when hooking a device up to a system if you get a notification when you run into a known limitation.

The configuration for things like data size, parity on/off, even/odd, is done by setting the chip's pins high/low:

```
NDB1/NDB2        39-40       Number Data Bits
NPB                3         No Parity Bit
POE                4         Parity Odd/Even
``` 

I didn't define any methods to allow those to be changed. So it is basically hard-coded as NDB1 and NDB2 are set to high (8 bits), NPB is set to high (no parity), and POE is a don't care.  If someone wants to use this chip and needs a different configuration, they would need to implement it.

cuavas:
> > Can you please make this always log (e.g. with `logerror`) when the host system uses an unimplemented, like configuring a smaller word size or enabling parity? It helps greatly when hooking a device up to a system if you get a notification when you run into a known limitation.
> 
> The configuration for things like data size, parity on/off, even/odd, is done by setting the chip's pins high/low:
> 
> ```
> NDB1/NDB2        39-40       Number Data Bits
> NPB                3         No Parity Bit
> POE                4         Parity Odd/Even
> ```
> 
> I didn't define any methods to allow those to be changed. So it is basically hard-coded as NDB1 and NDB2 are set to high (8 bits), NPB is set to high (no parity), and POE is a don't care. If someone wants to use this chip and needs a different configuration, they would need to implement it.

Ah, so it's not controlled by register writes. That's OK then.

PR #11960, opened 2024-01-21
Improve Magic Desk CRT compatability

PugsyMAME:
Commented out ^2 check, this stops a large number of MagicDesk CRTs from working. When this check is removed they work fine. The standard MagicDesk format consists of a variable number of 8 kB Bank, common sizes are 9-11 Banks.


happppp:
This will lead to crashes further down the road when software tries to read from an invalid bank and MAME hasn't allocated memory there.

PugsyMAME:
Fair enough.

PR #11964, opened 2024-01-22
cpu/t11: minor enhancements:

shattered:
Opcode fetch ignores least significant bit of PC; some tests check this. MTPS and MFPS are byte-wide instructions.
Define instruction set variants and reject instructions that are invalid for current CPU type.

### DIFF
--- a/src/devices/cpu/t11/t11.h
+++ b/src/devices/cpu/t11/t11.h
@@ -13,11 +13,6 @@ enum
 	T11_R0=1, T11_R1, T11_R2, T11_R3, T11_R4, T11_R5, T11_SP, T11_PC, T11_PSW
 };
 
-#define T11_IRQ0        0      /* IRQ0 */
-#define T11_IRQ1        1      /* IRQ1 */
-#define T11_IRQ2        2      /* IRQ2 */
-#define T11_IRQ3        3      /* IRQ3 */
-
 
 class t11_device :  public cpu_device
 {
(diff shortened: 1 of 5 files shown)


src/devices/cpu/t11/t11ops.hxx
```diff
@@ -268,6 +268,8 @@
 #define XOR_R(d)    int sreg, dreg, source, dest, result;     GET_SREG; source = REGW(sreg); GET_DW_##d; CLR_NZV; result = dest ^ source; SETW_NZ; PUT_DW_DREG(result)
 #define XOR_M(d)    int sreg, dreg, source, dest, result, ea; GET_SREG; source = REGW(sreg); GET_DW_##d; CLR_NZV; result = dest ^ source; SETW_NZ; PUT_DW_EA(result)
 
+/* test if insn is supported by the CPU */
+#define CHECK_IS(d) if (!(c_insn_set & (d))) { illegal(op); return; }
```
cuavas, line 272: This can produce weird effects if someone does something like:
```c++
	if (foo)
		CHECK_IS(IS_LEIS)
	else
		bar();
```

To make the macro work more like a single statement, wrap it in a dummy do/while loop:
```c++
#define CHECK_IS(d) do { if (!(c_insn_set & (d))) { illegal(op); return; } while (false)
```

src/devices/cpu/t11/t11.h
```diff
@@ -13,11 +13,6 @@ enum
 	T11_R0=1, T11_R1, T11_R2, T11_R3, T11_R4, T11_R5, T11_SP, T11_PC, T11_PSW
 };
 
-#define T11_IRQ0        0      /* IRQ0 */
-#define T11_IRQ1        1      /* IRQ1 */
-#define T11_IRQ2        2      /* IRQ2 */
-#define T11_IRQ3        3      /* IRQ3 */
```
cuavas, line 19: These input line constants are used in at least src/mame/dec/pdp11.cpp and possibly other places.  The code currently doesn’t compile.  If you give them a new home (e.g. as `static inline constexpr` inside `t11_device`), please do a full build and update everything that uses them.

cuavas (changes requested):

src/devices/cpu/t11/t11.h
shattered: I'll update that to use existing enum (t11_device::CP0_LINE)

PR #11966, opened 2024-01-23
luaengine.cpp: Add output manager item_table function

goldnchild:
output.h: Add itemtable accessor

This will allow you to see the output manager item table in lua.  It can be useful to check the names and see what outputs there are.

usage:  

function printt(t) for i,j in pairs(t) do print(i,j) end end 
printt(manager.machine.output.item_table)

### DIFF
--- a/src/emu/output.h
+++ b/src/emu/output.h
@@ -194,6 +194,9 @@ public:
 	// map a unique ID back to a name
 	const char *id_to_name(u32 id);
 
+	// return the item table
+	const std::unordered_map<std::string, output_item> &itemtable() const noexcept { return m_itemtable; }
+
 private:
 	output_item *find_item(std::string_view string);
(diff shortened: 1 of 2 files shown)


cuavas:
There are multiple issues with this.

Firstly, building Lua tables for potentially huge collections when accessing properties is too inefficient, and it’s something that’s being eliminated across the Lua interface.

Secondly, the whole way outputs are handled from Lua currently is going to change completely anyway when outputs become per-device.  That’s the reason the Lua interface to the output manager is completely undocumented – it’s going to change completely, so it’s better not to give the impression that it’s “supported” pe se.

The trouble with adding something like this is people will inevitably start to write things that depend on it, making the transition even more painful when everything changes.

So, to use a quote from a famous MAME developer, I’m going to have to ask you to, “sit back and wait for cool things to happen.”

goldnchild:
Ok, no worries.

PR #11969, opened 2024-01-26
New working machines (Casio CZ-230S and SZ-1)

devinacker:
----------
Casio CZ-230S [BCM, =CO=Windler, Devin Acker]
Casio SZ-1 [BCM, Devin Acker]

This commit also:
- adds the Panasonic MN1252 LCD controller
- implements the bit in the uPD7810 MM register to enable/disable internal RAM (needed for cz230s which maps ROM in that area instead)
- fixes the channel assignment bits in the uPD934

### DIFF
--- a/src/devices/sound/upd934g.cpp
+++ b/src/devices/sound/upd934g.cpp
@@ -135,14 +135,14 @@ void upd934g_device::write(offs_t offset, uint8_t data)
 		case 1: // normal
 		case 2: // muted
 		case 3: // accented
-		{
-			const u8 ch = (data & 3) ^ 2; // effective order seems to be "2, 3, 0, 1"
-			LOG("CMD PLAY sample %x (channel %d, effect %d)\n", m_sample, ch, data >> 6);
-			m_channel[ch].pos = m_addr[m_sample];
-			m_channel[ch].playing = m_sample;
-			m_channel[ch].effect = data >> 6;
+			{
+				const u8 ch = (data & 3) ^ 2; // effective order seems to be "2, 3, 0, 1"
+				LOG("CMD PLAY sample %x (channel %d, effect %d)\n", m_sample, ch, data >> 6);
+				m_channel[ch].pos = m_addr[m_sample];
+				m_channel[ch].playing = m_sample;
+				m_channel[ch].effect = data >> 6;
+			}
 			break;
-		}
 		}
 		break;
 	case 1:
--- a/src/devices/sound/upd934g.cpp
+++ b/src/devices/sound/upd934g.cpp
@@ -9,13 +9,14 @@
     TODO:
     - Correct MUTED and ACCENTED (currently just changes volume)
     - T1 input
-    - 8 channels?
 
 ***************************************************************************/
 
 #include "emu.h"
 #include "upd934g.h"
 
+#define VERBOSE (0)
+#include "logmacro.h"
 
 //**************************************************************************
 //  DEVICE DEFINITIONS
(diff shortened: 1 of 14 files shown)


src/devices/sound/upd934g.cpp
```diff
@@ -127,25 +124,24 @@ void upd934g_device::write(offs_t offset, uint8_t data)
...
+		case 2: // muted
+		case 3: // accented
+		{
+			const u8 ch = (data & 3) ^ 2; // effective order seems to be "2, 3, 0, 1"
+			LOG("CMD PLAY sample %x (channel %d, effect %d)\n", m_sample, ch, data >> 6);
+			m_channel[ch].pos = m_addr[m_sample];
+			m_channel[ch].playing = m_sample;
+			m_channel[ch].effect = data >> 6;
+		}
 			break;
```
angelosa, line 145: I think you meant to put the break before the curly brace here, i.e.:
```
			m_channel[ch].effect = data >> 6;
			break;
		}

```

src/devices/sound/upd934g.cpp
```diff
@@ -127,25 +124,24 @@ void upd934g_device::write(offs_t offset, uint8_t data)
...
-				m_channel[m_sample >> 1].pos = m_addr[m_sample];
-				m_channel[m_sample >> 1].playing = m_sample;
-				m_channel[m_sample >> 1].volume = data & 0x03;
-				m_channel[m_sample >> 1].effect = data >> 6;
-			}
+		case 1: // normal
+		case 2: // muted
+		case 3: // accented
+		{
+			const u8 ch = (data & 3) ^ 2; // effective order seems to be "2, 3, 0, 1"
```
startaq, line 139: Why the `^ 2`? I mean, all outputs go to the same speaker anyway, so the order shouldn't matter.

devinacker: I did that mostly so that the output ordering would still be correct for the RZ-1, since that does have the individual volume control and line outputs for each sample.

src/devices/video/mn1252.h
```diff
@@ -0,0 +1,42 @@
...
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+private:
+	static const u8 OUTPUT_DIGITS[0x40];
+
+	u8 m_data;
+	u8 m_first_nibble;
+	u8 m_nibble_count;
+	int m_ce, m_std;
```
cuavas: Members that need to be included in save states should use explicitly sized integers for portability.  Please use `u8` or something for `m_ce` and `m_std`.

src/devices/video/mn1252.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+	assert(digit < 6);
+	return m_output[digit];
+}
+
+/**************************************************************************/
+void mn1252_device::data_w(u8 data)
+{
+	m_data = (data & 0xf);
+}
```
cuavas, line 64: Not very important, but the parentheses here don’t really make the expression clearer.

src/mame/casio/cz230s.cpp
```diff
@@ -0,0 +1,693 @@
...
+#include "imagedev/cassette.h"
+#include "machine/clock.h"
+#include "machine/nvram.h"
+#include "sound/beep.h"
+#include "sound/upd933.h"
+#include "sound/upd934g.h"
+#include "video/mn1252.h"
+
+#include "screen.h"
+#include "speaker.h"
```
cuavas: Can you put a blank line after `#include "ra3.h"` to visually separate local project headers from the src/devices headers?

src/mame/casio/cz230s.cpp
```diff
@@ -0,0 +1,693 @@
...
+	return m_cassette->input() > 0 ? 0 : 1;
+}
+
+/**************************************************************************/
+void cz230s_state::render_w(int state)
+{
+	if(!state)
+		return;
+
+	for (int digit = 0; digit < 6; digit++)
```
cuavas: Another minor formatting thing – this `if` statement is missing the space before the opening parenthesis of its control expression.

src/mame/casio/cz230s.cpp
```diff
@@ -0,0 +1,693 @@
...
+		MIDI_PORT(config, "mdthru", midiout_slot, "midiout");
+		mdin.rxd_handler().append("mdthru", FUNC(midi_port_device::write_txd));
+	}
+
+	NVRAM(config, "nvram", nvram_device::DEFAULT_ALL_0);
+
+	MN1252(config, m_lcdc);
+
+	auto& screen = SCREEN(config, "screen", SCREEN_TYPE_SVG);
+	screen.set_refresh_hz(60);
```
cuavas: This right-associative unary `&` is on the left of the space, unlike the other ones which are on the right of the space.

PR #11982, opened 2024-01-29
Added Vib-Ribbon (Japan, Asia) to psx.xml

monkneys:
This is my very very first time ever committing to a github project so I'm trying to be very basic in my first push and hoping I did everything right. This is a simple addition to psx.xml of the hash for Vib-Ribbon (Japan, Asia), with original dump info from redump and converted to CHD by me via chdman.

### DIFF
--- a/hash/psx.xml
+++ b/hash/psx.xml
@@ -29414,6 +29414,31 @@ For an overview of US PS1 discs that have not been dumped and added to redump's 
 		</part>
 	</software>
 
+	<software name="vribbon">
+		<!--
+		Original images (Redump)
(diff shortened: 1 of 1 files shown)


hash/psx.xml
```diff
@@ -29414,6 +29414,31 @@ For an overview of US PS1 discs that have not been dumped and added to redump's
 		</part>
 	</software>
 
+	<software name="vribbon">
```
angelosa, line 29417: Did you test this game in latest MAME?
1. Knowing that this specific game can disc swap with any Audio CD is already a `supported="partial"` in itself (i.e. without considering for now that is arguably the main selling point), 
2. nm, if it isn't playable at all `supported="no"`
NOTE: This is a rhythm game so beats syncing with music is again a liability.

angelosa (changes requested):

hash/psx.xml
monkneys: I had tested it in MAME but only the ability to load and take input; I'll go further in. Thank you! I'll withdraw and come back when it's more ready.

PR #11983, opened 2024-01-29
Add update for russian translation

Arbars:


language/Russian/strings.po
```diff
@@ -891,6 +896,8 @@ msgid ""
 "One or more ROMs/disk images for this system are incorrect. The system may "
 "not run correctly.\n"
 msgstr ""
+"Один или несколько образов ROM иои дисков для этой системы повреждены. Система может "
```
holub, line 899: или

language/Russian/strings.po
```diff
@@ -344,7 +349,7 @@ msgstr "[список ПО]"
 #: src/frontend/mame/ui/filesel.cpp:298
 #, c-format
 msgid "Error accessing %s"
-msgstr ""
+msgstr "Ошибка доступности %s"
```
holub, line 352: Maybe: ошибка доступа к ...

language/Russian/strings.po
```diff
@@ -286,43 +291,43 @@ msgstr "Поиск: %1$s_"
...
+msgstr "Устройства ввода"
 
 #: src/frontend/mame/ui/optsmenu.cpp:87
 msgid "Save Settings"
 msgstr "Сохранить настройки"
 
 #: src/frontend/mame/ui/optsmenu.cpp:183 src/frontend/mame/ui/optsmenu.cpp:225
 msgid "System Filter"
-msgstr ""
+msgstr "Фильтр систем"
```
holub, line 322: Maybe: системный фильтр?

language/Russian/strings.po
```diff
@@ -412,7 +417,7 @@ msgstr "Пропускать меню выбора BIOS"
 
 #: src/frontend/mame/ui/submenu.cpp:44
 msgid "Skip software part selection menu"
-msgstr ""
+msgstr "ПРопускать меню выбора программы"
```
holub, line 420: Lowercase Р

language/Russian/strings.po
```diff
@@ -424,7 +429,7 @@ msgstr "Скрывать из списка доступных системы б
 
 #: src/frontend/mame/ui/submenu.cpp:52
 msgid "Advanced Options"
-msgstr "Дополнительные настройки"
+msgstr "Продвинутые настройки"
```
holub, line 432: Дополнительные is better choice

language/Russian/strings.po
```diff
@@ -500,7 +505,7 @@ msgstr "Настройки оформления"
 
 #: src/frontend/mame/ui/submenu.cpp:73
 msgid "Zoom to screen area"
-msgstr ""
+msgstr "Вместить в площадь экрана"
```
holub, line 508: "Во весь экран" is more common analog

language/Russian/strings.po
```diff
@@ -512,15 +517,15 @@ msgstr "Автоматическое сохранение/загрузка"
...
 
 #: src/frontend/mame/ui/submenu.cpp:78
 msgid "Rewind capacity"
-msgstr ""
+msgstr "Объём перемотки"
 
 #: src/frontend/mame/ui/submenu.cpp:79
 msgid "Bilinear filtering for snapshots"
-msgstr ""
+msgstr "Накладывать билинейный фильтр на скриншоты"
```
holub, line 528: Снимок экрана more common term

Arbars:
'системный фильтр' is not correct, because we are sorting a systems by year, manufacturer, etc.--edit-last

holub:
Thank you @Arbars !
Whatever is marked as "tbd" are possibly can be ignored or taken into consideration - up to you.

I have mixed feelings about transliterations clone, timecode, cheats; but it hard to find better analogues. Especially because these terms are used widely enough.

Arbars:
-e

language/Russian/strings.po
```diff
@@ -2320,29 +2346,29 @@ msgstr "Изменить папку"
...
 msgctxt "gfxview"
 msgid "Zoom = %1$d"
-msgstr ""
+msgstr "Масштаб = %1$d"
 
 #: src/frontend/mame/ui/viewgfx.cpp:800
 msgctxt "gfxview"
 msgid "Expand to fit"
-msgstr ""
+msgstr "Вместить"
```
cuavas, line 2360: This translation may be misleading as it will only expand small tilemaps to fill the display – it will not shrink tilemaps that are too large to fit.  That’s why the English specifically says “Expand to fit”.

language/Russian/strings.po
```diff
@@ -2424,7 +2450,7 @@ msgstr ""
 
 #: src/frontend/mame/ui/selgame.cpp:463
 msgid "System Settings"
-msgstr ""
+msgstr "Системные настройки "
```
cuavas, line 2453: Stray space at the end of the translated message.

language/Russian/strings.po
```diff
@@ -2460,7 +2486,7 @@ msgstr "Кликние два раза или нажмите %1$s для выб
 
 #: src/frontend/mame/ui/auditmenu.cpp:60
 msgid "Audit Media"
-msgstr ""
+msgstr "ПРоверка медиа"
```
cuavas, line 2489: Why have you capitalised the second letter of “ПРоверка”?

language/Russian/strings.po
```diff
@@ -2473,6 +2499,8 @@ msgid ""
 "Auditing media for system %2$u of %3$u...\n"
 "%1$s"
 msgstr ""
+"ПРоверка медиа для %2$u систем из %3$u...\n"
+"%1$s"
```
cuavas, line 2503: Same here – is there a reason for the odd capitalisation of “ПРоверка”?

language/Russian/strings.po
```diff
@@ -2808,51 +2840,58 @@ msgid ""
 "Next device  \t\t%6$s\n"
 "Return to previous menu  \t\t%7$s"
 msgstr ""
+"Показать/скрыть настройки  \t\t%1$s\n"
+"Уменьшить значение  \t\t%2$s\n"
+"Увеличить значение  \t\t%3$s\n"
+"Вернуть значения по умолчанию \t\t%4$s\n"
+"Предыдущее устройство  \t\t%5$s\n"
+"Следующее устройство  \t\t%1\6$s\n"
+"Вернуться в предыдушее меню  \t\t%7$s\n"
```
cuavas, line 2849: The two spaces after each action are intentional to force minimum spacing – you’re missing one space after “Вернуть значения по умолчанию”.

language/Russian/strings.po
```diff
@@ -2808,51 +2840,58 @@ msgid ""
...
 msgctxt "menu-analoginput"
 msgid "Press %s to show settings"
-msgstr ""
+msgstr "Для отображения настроек нажмите %s"
 
 #: src/frontend/mame/ui/analogipt.cpp:245
 msgctxt "menu-analoginput"
 msgid "Analog Input Adjustments Help"
-msgstr ""
+msgstr "Справка по настройке"
```
cuavas, line 2865: Is there a reason for making the translated title non-specific here?

language/Russian/strings.po
```diff
@@ -2984,12 +3023,12 @@ msgstr "Удалить из Избранного"
...
 msgctxt "menu-main"
 msgid "About %1$s"
-msgstr ""
+msgstr "О %1$s"
 
 #: src/frontend/mame/ui/mainmenu.cpp:195
 msgctxt "menu-main"
 msgid "Start System"
-msgstr ""
+msgstr "Запустить системы"
```
cuavas, line 3031: This is inconsistent with the place where you translated “Start System” as “Старт системы”.

language/Russian/strings.po
```diff
@@ -3126,11 +3165,11 @@ msgstr ""
 #: src/frontend/mame/ui/selmenu.cpp:693
 #, c-format
 msgid "Software is clone of: %1$s"
-msgstr "Программа - clone от: %1$s"
+msgstr "Программа - клон от: %1$s"
 
 #: src/frontend/mame/ui/selmenu.cpp:695
 msgid "Software is parent"
-msgstr "Программа - parent"
+msgstr "Программа - оригинал"
```
cuavas, line 3172: Please don’t translate “parent” in a way that implies “original” – it’s misleading.  Parent/clone relationships are a convenience feature for organisation, and the parent is often not the “original” as such.  The parent is generally the latest version of the most widespread variant.  For example world-wide English language releases are usually the parent when the original is Japanese language software.

language/Russian/strings.po
```diff
@@ -3146,12 +3185,12 @@ msgstr "Поддерживается: Да"
 
 #: src/frontend/mame/ui/selmenu.cpp:726
 msgid "System is parent"
-msgstr "Система - parent"
+msgstr "Система - оригинал"
```
cuavas, line 3188: Same here – do not translate “parent” in a way that implies it’s the original.

language/Russian/strings.po
```diff
@@ -3220,11 +3265,11 @@ msgstr "Производитель\t%1$s\n"
 #: src/frontend/mame/ui/selmenu.cpp:2944
 #, c-format
 msgid "System is Clone of\t%1$s\n"
-msgstr "Система - Clone от\t%1$s\n"
+msgstr "Система - клон от\t%1$s\n"
 
 #: src/frontend/mame/ui/selmenu.cpp:2949
 msgid "System is Parent\t\n"
-msgstr "Система - Parent\t\n"
+msgstr "Система - оригинал\t\n"
```
cuavas, line 3272: Same again – do not translate “parent” in a way that implies it’s original.

language/Russian/strings.po
```diff
@@ -3514,7 +3561,7 @@ msgstr "Звук"
 
 #: src/frontend/mame/ui/sndmenu.cpp:149
 msgid "Compressor"
-msgstr ""
+msgstr "Метод сжатия"
```
cuavas, line 3564: This refers to dynamic range compression, not data compression.  I don’t think this translation is correct.  It’s usually called “Компрессор аудиосигнала” or just “Компрессор” where sound processing is implied from context.

language/Russian/strings.po
```diff
@@ -5750,27 +5811,27 @@ msgstr ""
...
 
 #: plugins/cheatfind/init.lua:522
 msgid "Automatically toggle pause with on-screen menus"
-msgstr ""
+msgstr "Автоматически ставить на паузу при вызове меню"
 
 #: plugins/cheatfind/init.lua:543
 msgid "All slots cleared and current state saved to Slot 1"
-msgstr ""
+msgstr "Все слоты очищены и текущее состояние сохранено в Слоте 1"
```
cuavas, line 5834: “Slot” in the cheat finder plugin is poor choice of terminology in English, and should not be propagated to translations.  In particular, it causes confusion with slot devices.  In the cheat finder plugin, “slot” refers to a snapshot of memory.  Can you think of a better translation?

language/Russian/strings.po
```diff
@@ -5874,22 +5935,23 @@ msgstr "Формат данных"
...
+msgstr "Тест/Запись указанного значения"
 
 #: plugins/cheatfind/init.lua:747
 msgid ""
 "Use this if you want to poke the Slot 1 value (eg. You started with "
 "something but lost it)"
 msgstr ""
 
+
 #: plugins/cheatfind/init.lua:749
```
cuavas, line 5955: Please don’t alter the overall formatting of the file like this – it causes trouble when the localisation sources are regenerated.

language/Russian/strings.po
```diff
@@ -5899,11 +5961,11 @@ msgstr ""
 #: plugins/cheatfind/init.lua:751
 #, lua-format
 msgid "Use this if you want to poke %s"
-msgstr ""
+msgstr "Используется для указания %s"
```
cuavas, line 5964: “Poke” here refers to replacing a value in memory.  Is this translation appropriate?

language/Russian/strings.po
```diff
@@ -5952,111 +6014,111 @@ msgstr "Поиск читов"
...
 #: plugins/portname/init.lua:155
 msgid "Failed to save input name file"
-msgstr ""
+msgstr "Не удалось сохранить файл названий порта ввода"
 
 #: plugins/portname/init.lua:167
 #, lua-format
 msgid "Input port name file saved to %s"
-msgstr ""
+msgstr "Файл названия порта ввода сохранён в  %s"
```
cuavas, line 6027: Stray double space here.

language/Russian/strings.po
```diff
@@ -5952,111 +6014,111 @@ msgstr "Поиск читов"
...
 msgctxt "plugin-autofire"
 msgid "Add autofire button"
-msgstr ""
+msgstr "Добавить кнопку авто-огня"
 
 #: plugins/autofire/autofire_menu.lua:150
 msgctxt "plugin-autofire"
 msgid "n/a"
-msgstr ""
+msgstr "н/д"
```
cuavas, line 6080: Why the inconsistency in translation of “n/a” vs preserving the Latin script version?

language/Russian/strings.po
```diff
@@ -6187,96 +6249,96 @@ msgstr "Имя"
...
 msgctxt "plugin-inputmacro"
 msgid "[not set]"
-msgstr ""
+msgstr "[не настроено]"
 
 #: plugins/inputmacro/inputmacro_menu.lua:361
 msgctxt "plugin-inputmacro"
 msgid "Activation combination"
-msgstr ""
+msgstr "КОмбинация запуска"
```
cuavas, line 6257: Is there a reason for the odd capitalisation on “КОмбинация”?

language/Russian/strings.po
```diff
@@ -6487,15 +6549,15 @@ msgstr "Отмена"
 #: plugins/commonui/init.lua:199
 msgctxt "plugin-commonui"
 msgid "Invalid combination entered"
-msgstr ""
+msgstr "ВВедена неправильная комбинация"
```
cuavas, line 6552: Is there a reason for the odd capitalisation on “ВВедена”?

Arbars:
Fixed number of "ПРо" mistypes :)
Any suggestions?

cuavas:
Please don’t force push every time like this – it makes all comments show as “outdated” even when the lines in question haven’t changed, and makes it more effort to see what’s changed on each update.

cuavas:
I’ve taken most of the uncontroversial parts of this and applied it in 34b1aef7ae8447067cf560cf3bb170bf8ee71b6f.  There are some conflicts due to correction (e.g. correcting “ID” in “Скопировать TD устройства”, and correcting “файла” in “Ошибка удаления файоа сохранения”).

Can you please merge the latest upstream master branch into your branch and resolve the conflicts so we can discuss the remaining changes?

language/Russian/strings.po
```diff
@@ -3650,12 +3664,12 @@ msgstr "* Устройств(о/а) ввода"
...
 #, c-format
 msgid "  User inputs    [%1$d inputs]\n"
-msgstr ""
+msgstr "  Ввод пользователя    [%1$d типов ввода]\n"
 
 #: src/frontend/mame/ui/devopt.cpp:304
 #, c-format
 msgid "  Mahjong inputs    [%1$d inputs]\n"
-msgstr ""
+msgstr "  Управление в Маджонге    [%1$d кнопок]\n"
```
cuavas, line 3672: As noted several times, you’re going to need to come up with a better, consistent term for referring to inputs to the emulated system.  I don’t think “типов ввода” really conveys this, either.

language/Russian/strings.po
```diff
@@ -5827,17 +5841,17 @@ msgstr "Начать новый поиск"
...
 #, lua-format
 msgid "Save current memory state to Slot %d"
-msgstr ""
+msgstr "Сохранить текущее состояние памяти в Слот %d"
 
 #: plugins/cheatfind/init.lua:614
 #, lua-format
 msgid "%d total matches found"
-msgstr ""
+msgstr "Всего найдено %d машин"
```
cuavas, line 5854: I don’t think “машин” is correct here.  It’s referring to the number of locations in memory that match the criteria.

language/Russian/strings.po
```diff
@@ -5968,7 +5987,7 @@ msgstr "Все"
 #: plugins/cheatfind/init.lua:907
 #, lua-format
 msgid "Test Cheat %08X_%02X"
-msgstr ""
+msgstr "Проветирь чит %08X_%02X"
```
cuavas, line 5990: This is the format used for the name of the cheat that was newly created for testing.  Is this translation appropriate?

language/Russian/strings.po
```diff
@@ -6124,13 +6143,13 @@ msgstr "Отмена"
...
 msgid "Select an input for autofire"
-msgstr ""
+msgstr "Выбор метода ввода лоя автоматического огня"
 
 #: plugins/timecode/init.lua:155
 #, lua-format
 msgctxt "plugin-timecode"
 msgid "TIMECODE: Intro started at %s"
-msgstr ""
+msgstr "ТАЙМКОД: Вступление с %s"
```
cuavas, line 6152: Your translations of the other “<x> started at” messages all explicitly include “начинается с” – why did you omit it for this one?

language/Russian/strings.po
```diff
@@ -6371,17 +6390,17 @@ msgstr ""
...
 msgctxt "plugin-timer"
 msgid "Emulated time"
-msgstr ""
+msgstr "Время эмуляции"
 
 #: plugins/timer/init.lua:37
 msgctxt "plugin-timer"
 msgid "Reference"
-msgstr ""
+msgstr "Ссылка"
```
cuavas, line 6403: These messages are for setting the reference clock for measuring elapsed time.  “Wall clock” refers to a conceptual stable clock for measuring elapsed real time.  “Emulated time” refers to time from the point of view of the emulated system (so it’s affected by using the “fast forward” feature, and when emulation doesn’t run at full speed).  Are these translations appropriate?

language/Russian/strings.po
```diff
@@ -6465,12 +6484,12 @@ msgstr "Команда"
...
 msgctxt "plugin-data"
 msgid "Mamescore"
-msgstr ""
+msgstr "Рекорд"
 
 #: plugins/data/data_sysinfo.lua:16
 msgctxt "plugin-data"
 msgid "Sysinfo"
-msgstr ""
+msgstr "Информация о системе"
```
cuavas, line 6492: I’m not sure these should be translated literally – they refer to specific data files that were at one time maintained by community members.

holub:
> > (It refers to the identifiers used for a software list and an item within the list, e.g. “Software list/item: megadriv:bugslife”
> 
> And that means, the line
> 
> > msgstr "Список ПО/предметов"
> 
> was translated incorrectly, and more suitable variant should be look like
> 
> > msgstr "Пункт из списка программ"
> 
> or
> 
> > msgstr "Пункт списка программ"
> 
> were "пункт" indicates the name of the program from the software list, so I will change other lines in that way.
> 
> > msgid "Wall clock"
> 
> Thats phrase reminds me an ordinary clock that hangs on the wall, but this more closer to set of clocks on the wall in the airport... Interesting :)

"Wall clock" means real time outside of emulsted system. If we run something at 200% after 1 minutes realtime clock becomes 2m, but wall still 1m.

holub:
Часы реального времени?

Arbars:
> Часы реального времени?
Hm... Maybe. But then why is not "Realtime clock"?

cuavas:
> @
> 
> > > Часы реального времени?
> > 
> > 
> > Hm... Maybe. But then why is not "Realtime clock"?
> 
> @cuavas correct me if I'm wrong. Are "wall clock" and RTC interchangable?

Not really.  “Real-time clock” or “RTC” means a device that measures elapsed time, while “wall clock” is a conceptual device that tells you the current time.  The distinction is significant in some cases.

For example in latency applications, it’s common to use a simple offset/multiply on a CPU performance counter register (e.g. invariant TSC on Intel CPUs) as a “real-time clock”, and periodically update the constants against the “wall clock” (usually the operating system clock).

In applications requiring high precision synchronisation between sites, the “real-time clock” is usually a GPS-disciplined oscillator or atomic clock, and the “wall clock” is a GNSS (GPS/GLONASS/Galileo/Beidou) or CDMA/OFDM cellular radio time signal receiver.  The real-time clock measures time locally and can give an instantaneous response, while the wall clock is an external reference for the current time available across disparate locations.

In the case of this plugin, it doesn’t actually measure elapsed time while you run the system.  It checks current time on the “wall clock” (in this case the operating system clock) at the start and end, and calculates the difference.  If you do something like adjust the operating system clock while the emulator is running, you may get an unexpected result.

holub:
In Russian such analog doesn't exist. Can it be something like "Внешние часы" (external clock) or "Время OС" (OS time)?

cuavas:
> In Russian such analog doesn't exist. Can it be something like "Внешние часы" (external clock) or "Время OС" (OS time)?

Yeah, something along those lines is probably OK.

PR #11985, opened 2024-01-29
memarray.cpp: Fix use of mismatched endianness macros

amameuser:
Fixes the little-endian functions read8_from_64le and write8_to_64le using the big-endian BYTE8_XOR_BE macro instead of the little-endian BYTE8_XOR_LE macro.


amameuser:
Fixes the little-endian functions read8_from_64le and write8_to_64le using the big-endian BYTE8_XOR_BE macro instead of the little-endian BYTE8_XOR_LE macro.

cuavas:
For what it’s worth, the `memory_array` class isn’t recommended for use in new code, but still better to have these lurking things fixed.

PR #11991, opened 2024-01-31
sega/segaai.cpp: Improve upd7759 banking.

wilbertpol:


ocornut:
Specifically this should fix the problem of incorrect voice samples being played.

PR #11992, opened 2024-02-01
Modified k052591 PMC emulation to reflect how the real programs work

furrtek:
Small changes based on silicon reverse-engineering.

spy: Confirmed projection function constants, more accurate collision check without the need for special case handling
thunderx: Simplified collision check, fixed object flags updates
hexion: Added special 16-byte VRAM clearing command

Not sure if mentionning an URL to the documentation in comments is the right thing to do.

### DIFF
--- a/src/mame/konami/hexion.cpp
+++ b/src/mame/konami/hexion.cpp
@@ -102,6 +102,7 @@ Command is 01: Set banks to 1 and fill from 0 to 0x1fff with r2.b
 #include "emupal.h"
 #include "speaker.h"
 #include "tilemap.h"
+#include "multibyte.h"
 
 
 // configurable logging
@@ -252,8 +253,9 @@ void hexion_state::bankswitch_w(uint8_t data)
 		}
 		else
 		{
-			memset(m_vram[m_pmcram[4]] + ((m_pmcram[3] << 8) + m_pmcram[2]), 0, 16);
-			m_bg_tilemap[m_pmcram[4]]->mark_all_dirty();
+			uint8_t bank = m_pmcram[4] & 1;
+			memset(m_vram[bank] + (get_u16le(&m_pmcram[2]) & 0x1fff), 0, 16);
+			m_bg_tilemap[bank]->mark_all_dirty();
 		}
 	}
 	// bit 7 = PMC-BK
--- a/src/mame/konami/hexion.cpp
+++ b/src/mame/konami/hexion.cpp
@@ -10,71 +10,81 @@ Notes:
 - The board has a 052591, which is used for protection in Thunder Cross and
   S.P.Y. In this game, however, the only thing it seems to do is clear the
   screen.
-  This is the program for the 052591:
-00: 5f 80 01 e0 08
-01: df 80 00 e0 0c
-02: df 90 02 e0 0c
-03: df a0 03 e0 0c
-04: df b0 0f e0 0c
-05: df c0 ff bf 0c
-06: 5c 02 00 33 0c
-07: 5f 80 04 80 0c
+
+This is the 052591 PMC code loaded at startup, it contains a RAM/VRAM filling program.
+See https://github.com/furrtek/SiliconRE/tree/master/Konami/052591 for details
+
+00: 5f 80 01 e0 08	Entry point, set OUT0 high
+01: df 80 00 e0 0c	r0 = 0
+02: df 90 02 e0 0c	r1 = 2
+03: df a0 03 e0 0c	r2 = 3
+04: df b0 0f e0 0c	r3 = f
+05: df c0 ff bf 0c	ExtAddr = 1fff, r4 = ffff
+06: 5c 02 00 33 0c	
+07: 5f 80 04 80 0c	Write 2 to RAM (1fff) m_bankctrl, select pmcram
 08: 5c 0e 00 2b 0c
-09: df 70 00 cb 08
-0a: 5f 80 00 80 0c
+09: df 70 00 cb 08	r7 = RAM(4)
+0a: 5f 80 00 80 0c	ExtAddr = 0
 0b: 5c 04 00 2b 0c
-0c: df 60 00 cb 08
-0d: 5c 0c 1f e9 0c
-0e: 4c 0c 2d e9 08
-0f: 5f 80 03 80 0c
+0c: df 60 00 cb 08	r6 = RAM(0) (commands 0, 1 and 30 are used)
+0d: 5c 0c 1f e9 0c	JP 1F if r6 == 0 
+0e: 4c 0c 2d e9 08	JP 2D if r6 == 1 
+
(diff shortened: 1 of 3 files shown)


src/mame/konami/hexion.cpp
```diff
@@ -231,12 +241,20 @@ void hexion_state::bankswitch_w(uint8_t data)
...
+		uint8_t command = m_pmcram[0];
+		if (command <= 1)
+		{
+			memset(m_vram[command], m_pmcram[1], 0x2000);
+			m_bg_tilemap[command]->mark_all_dirty();
+		}
+		else
+		{
+			memset(m_vram[m_pmcram[4]] + ((m_pmcram[3] << 8) + m_pmcram[2]), 0, 16);
+			m_bg_tilemap[m_pmcram[4]]->mark_all_dirty();
```
cuavas, line 256: You can use `get_u16_le` from `"multibyte.h"` to read little Endian values, e.g. `memset(m_vram[m_pmcram[4]] + get_u16_le(&m_pmcram[2]), 0, 16);` here.

src/mame/konami/spy.cpp
```diff
@@ -285,64 +299,21 @@ void spy_state::bankswitch_w(uint8_t data)
...
-					m_pmcram[i + 0xd] = 0;
-				}
-				else
-					m_pmcram[i + 0xd] = 1;
-			}
-		}
-	}
-	else if (op1 > 1)
+	int a_pos, a_size, a_min, a_max, b_pos, b_size, b_min, b_max;
+	int count, i, j, tests_failed, loopend, nearplane, op2;
```
cuavas, line 310: Please scope these down to where they’re used.  In particular:
* You should be able to declare `i` and `j` in the loop control expressions since they’re only valid during the loops.
* `loopend` and `nearplane` should only exist in one branch of the `if`/`else`.
* Most of the other things here have a logical lifetime of a single loop iteration, and several of them should be `const`.

src/mame/konami/spy.cpp
```diff
@@ -285,64 +299,21 @@ void spy_state::bankswitch_w(uint8_t data)
 	membank("bank1")->set_entry(bank);
 }
 
-void spy_state::spy_collision(  )
+void spy_state::pmc_run(  )
 {
 #define MAX_SPRITES 64
 #define DEF_NEAR_PLANE 0x6400
 #define NEAR_PLANE_ZOOM 0x0100
 #define FAR_PLANE_ZOOM 0x0000
```
cuavas, line 307: Can you replace these with `constexpr` locals while you’re working on this?  They’re easier to deal with.

src/mame/konami/spy.cpp
```diff
@@ -364,6 +335,56 @@ void spy_state::spy_collision(  )
...
+			
+			// Check all 3 dimensions
+			tests_failed = 3;
+			for (j = 0; j < 3 * 4; j += 4)
+			{
+				a_pos = (m_pmcram[j + 3] << 8) + m_pmcram[j + 4];	// Object A center position
+				a_size = (m_pmcram[j + 5] << 8) + m_pmcram[j + 6];	// Object A half size
+		
+				b_pos = (m_pmcram[i + j + 1] << 8) + m_pmcram[i + j + 2];	// Object B center position
+				b_size = (m_pmcram[i + j + 3] << 8) + m_pmcram[i + j + 4];	// Object B half size
```
cuavas, line 360: You can use `get_u16_be` from `"multibyte.h"` to read big Endian values, e.g. `a_pos = get_u16_be(&m_pmcram[j + 3]);` etc. here.

src/mame/konami/hexion.cpp
```diff
@@ -231,12 +241,20 @@ void hexion_state::bankswitch_w(uint8_t data)
...
+		}
```
cuavas, line 257: You’ve removed the mask on the video RAM bank here – that causes a potential heap smash.  You should also do a mask or bounds check on the offset when doing the 16-byte clear as there’s nothing to stop it smashing memory off the end of the area the way it is now.

src/mame/konami/hexion.cpp
```diff
@@ -231,12 +242,21 @@ void hexion_state::bankswitch_w(uint8_t data)
...
+		if (command <= 1)
+		{
+			memset(m_vram[command], m_pmcram[1], 0x2000);
+			m_bg_tilemap[command]->mark_all_dirty();
+		}
+		else
+		{
+			uint8_t bank = m_pmcram[4] & 1;
+			memset(m_vram[bank] + (get_u16le(&m_pmcram[2]) & 0x1fff), 0, 16);
+			m_bg_tilemap[bank]->mark_all_dirty();
```
cuavas, line 258: There’s still a potential heap smash here – consider what happens if `(get_u16le(&m_pmcram[2]) & 0x1fff)` is 0x1ff8.

furrtek: Fixed on my side. Should I submit a new pull request ?

cuavas: I already fixed it.

src/mame/konami/thunderx.cpp
```diff
@@ -444,10 +388,44 @@ void thunderx_state::calculate_collisions()
...
+			if (!(p1[0] & hm))
+				continue;
+			
+			if (p1[1] + p0[1] < abs(p1[3] - p0[3])) continue;
+			if (p1[2] + p0[2] < abs(p1[4] - p0[4])) continue;
 
-	run_collisions(s0, e0, s1, e1, cm, hm);
+			// set flags
+			p1[0] = (p1[0] & 0x8f) | 0x10;
+			if (p1 > (uint8_t*)0xe6)	// This address value is hardcoded in the PMC program
```
happppp, line 421: @furrtek this pointer comparison won't work since &m_pmcram[0] is not 0.
If I read your disasm correctly, it's (&p1[4] >= &m_pmcram[0xe6]), but then collisions don't work anymore.

happppp: Maybe p0[4] instead of p1[4], but then I see another issue where the address value in the US version is $36 instead of $E6,

src/mame/konami/thunderx.cpp
```diff
@@ -423,18 +368,17 @@ void thunderx_state::calculate_collisions()
...
 
-	const int e0 = (m_pmcram[0] << 8) | m_pmcram[1];
-	const int e1 = m_pmcram[2];
+	const uint16_t e0 = get_u16be(&m_pmcram[0]);
+	const uint8_t e1 = m_pmcram[2];
 
-	int s0, s1;
+	uint16_t s0, s1;
+	// Heuristic to determine version of program based on byte at 0x05 
 	if (m_pmcram[5] < 16)
```
happppp, line 378: This check won't work, you're testing a byte from data RAM, not program RAM.

happppp: Nevermind, it does intend to check data RAM.

PR #11994, opened 2024-02-01
handheld/hh_sm510.cpp: Add Tiger Double Dragon II: The Revenge

algestam:
This PR adds Double Dragon II: The Revenge by Tiger. The MCU was dumped by Sean Riddle and I have digitized the display and prepared the MAME driver.

Marking this PR as a draft since the joystick left action doesn't seem to have any effect and I haven't figured out if the configuration is incorrect or if the left joystick simply doesn't have any effect in the game.


FoxhackDN:
Someone is selling the manual on ebay, and zooming in on the poor photo they posted suggests backwards is only used in a specific mission. So it might not be broken.

algestam:
Thanks @FoxhackDN, that sounds reasonable as I'm quite certain that the configuration is correct as per the PCB routing. Haven't played the game far enough to confirm this but will change the PR to be ready for review.

FoxhackDN:
For future reference, he's a zoom in of that manual at 200%. It's blurry but you can kiiiiiiiinda see "Mission X only" there.
image

happppp:
It's from 1990 (Tiger's copyright year is on the back of the packaging)
And yes, Batman is from 1989 even though it's a higher model number.

algestam:
Thanks @happppp ! I'll update the year information 👍

PR #11995, opened 2024-02-01
cpu/t11: add basic support for K1801VM1

shattered:


src/devices/cpu/t11/t11table.hxx
```diff
@@ -25,7 +25,7 @@ modes:
 const t11_device::opcode_func t11_device::s_opcode_table[65536 >> 3] =
 {
 	/* 0x0000 */
-	OP(op_0000),    OP(halt),       OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),
+	OP(op_0000),    OP(op_0001),    OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),    OP(illegal),
```
cuavas, line 28: Won’t this change behaviour for all T11 family CPU devices now?  As far as I can tell, this changes what was previously `halt` to an illegal instruction for everything besides the K1801VM1.  Wouldn’t you need to also make `t11_device::halt(uint16_t op)` call `t11_device::halt(uint16_t op)` if the CPU is not a K1801VM1 to preserve current behaviour for existing CPU types?

shattered: T11 user's manual (page B-7) lists these opcodes as reserved and not as HALT, and I believe other processors in the pdp-11 family treat them likewise.

cuavas: Do you know why it was treating it as halt in the first place?

PR #11997, opened 2024-02-02
heath/h17_fdc.cpp: Skeleton impl for the H-88-1 Hard-Sectored Controller

mgarlanger:
Initial code for the Heath H-88-1 Hard-sectored controller for the H-89 family of systems. This supported up to 3 floppy drives. 

This initial code implements the I/O ports of the controller, includes the AMI S2350 USRT, the transmitter clock for the USRT. 

Things still needed

- MAME's core code must support hard-sectored disk images.
- Receiver clock for the USRT, based on read clock bits from the disk.
- Add support for a Heath hard-sectored disk image format.
- Add controller to the H89 systems.

### DIFF
--- a/src/mame/heathkit/h17_fdc.cpp
+++ b/src/mame/heathkit/h17_fdc.cpp
@@ -190,6 +190,7 @@ u8 heath_h17_fdc_device::read(offs_t reg)
 	{
 	case 0: // data port
 		val = m_s2350->receiver_output_reg_r();
+		break;
 	case 1: // status port
 		val = m_s2350->status_word_r();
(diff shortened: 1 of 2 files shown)


mgarlanger:
Note: my comment about "Things still needed", are things that are needed before the hard-sectored controller will fully function, but this PR as is, is ready for review/merge. I've tested this change by adding this controller to the H89, and the software is able to detect it, and detect drives attached to the controller. 
I have not included that code, since the controller will not read/write images until the other issues above are addressed, but those will be in other PRs.

Issue #11078, opened 2023-04-06
Rewrite PC8801-30 / PC Engine CD honoring the expansion bus


angelosa:
PC Engine PCB flow is:
- main unit has an expansion bus bay area, pinout is [here](https://gamesx.com/misctech/pcebp.php);
- This connects to an IFU-30 bay, which contains an μPD91061GD;
- Finally, the C-CD connector goes to the actual PC8801-30 CD drive, [part list](https://wiki.console5.com/wiki/Super_CD-ROM2).

Practically, both `pc8801mc` and `pce cdsys` have limited host SCSI interface:
- BSY, REQ, MSG, I/O, CD as input;
- SEL as output;
On boot they put `0x81 -> 0xff` on the data bus then pings the SEL line, this implies that the SCSI bus is configured to select ID devices 0 and 7. Speculation is that '91061 is a SCSI bridge mapped at 7 that is responsible for details towards device 0.

angelosa:
IFU-30 PCB:

![IFU3](https://user-images.githubusercontent.com/9251271/230451587-89ec853d-a583-445a-bf37-055b52b2f41f.jpg)

angelosa:
Some refs:
- https://x68k-pcengine.blogspot.com/2018/06/pc-engine-cd-dom2-pc-8801-mc-cdrom.html
- https://wiki.console5.com/tw/index.php?title=Turbo_Grafx_CD-ROM_Drive
> These are adapted from the specs for the NEC CDR-35, which is based on the same hardware design as the TG16 CDROM.

NEC CDR-35 being a portable alias for PC/Mac based systems, normally connects thru a BFU-35 unit. Has an internal 78c14 MCU according to following post:
https://www.reddit.com/r/TurboGrafx/comments/d73zfw/got_the_nec_cdr35d_to_work_as_a_cdrom%C2%B2turbocd

PR #11999, opened 2024-02-02
nec/pce.cpp, nec/pce_cd.cpp: Various updates

cam900:
- Use device_mixer_interface rather than using hard-coded speaker tag
- Reduce defines
- Remove outdated comments
- Remove nec/pce_m.cpp, Move pce_m.cpp codes into nec/pce.cpp


src/mame/nec/pce.cpp
```diff
@@ -69,27 +69,62 @@ Super System Card:
...
+	m_port_ctrl->sel_w(BIT(data, 0));
+	m_port_ctrl->clr_w(BIT(data, 1));
+}
+
+u8 pce_state::controller_r()
+{
+	u8 ret = (m_port_ctrl->port_r() & 0x0f) | m_io_port_options;
+#ifdef UNIFIED_PCE
+	ret &= ~0x40;
+#endif
```
angelosa, line 96: Looks like this `UNIFIED_PCE` macro is unused, it's safe to remove.

cuavas:
What’s the motivation for moving it into src/devices/machine?  Is it used for anything besides the PC-Engine?

angelosa:
File is currently a mixup of _PC8801_-30 drive, the specific (and opaque) PC Engine SCSI controller and the rest of the expansion unit bus & i/f.

https://github.com/mamedev/mame/issues/11078

afaik _PC8801_-31 SCSI isn't much different beyond relying on SEL line alone (and doing the rest internally), a preliminary core i/f (without the controller part, _maybe_ at ID 7):

https://github.com/mamedev/mame/blob/91037294f03b4abb5a38eb161f1ee7a1438d68eb/src/devices/bus/pc8801/pc8801_31.cpp

Bottom line: not convinced about the move either, it's essentially a NEC i/f that can stay in `src/mame/nec`.

angelosa (approved):

PR #12000, opened 2024-02-02
Fix compilation against musl

ypsvlq:
- `__GLIBC__` is the correct macro to check for `feenableexcept`
- `POLLIN` is the correct macro for poll, `POLL_IN` is defined in `signal.h`

it still won't compile because bx [deliberately removed support](https://github.com/bkaradzic/bx/commit/d7ecac17060ab8cf765b21e55a762aa4855dfdb1), but at least removes the need to patch mame

### DIFF
--- a/src/lib/netlist/plib/pexception.cpp
+++ b/src/lib/netlist/plib/pexception.cpp
@@ -8,7 +8,7 @@
 #include <cfloat>
 #include <iostream>
 
-#if (defined(__x86_64__) || defined(__i386__)) && defined(__linux__)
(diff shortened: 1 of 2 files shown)


cuavas:
`POLLIN` is not portable – `POLL_IN` is the standard macro, as specified in SUSv2: https://pubs.opengroup.org/onlinepubs/7908799/xsh/signal.h.html

Take it up with the musl developers.

ypsvlq:
`POLL_IN` is the standard macro for handling `SIGPOLL`, when you have included `signal.h`.
`POLLIN` is also standard and specified in the same document: https://pubs.opengroup.org/onlinepubs/7908799/xsh/poll.h.html

PR #12001, opened 2024-02-02
irem/travrusa.cpp: Add support for Traverse USA and clones different PROMs setup [Victor Fernandez (City Game), Recreativas.org]

clawgrip:
Add support for Traverse USA and clones different PROMs setup with two 256x4 character palette PROMs instead of only 256x8 one (same contents on both setups).

Confirmed to exist on Traverse USA, MotoTour and MotoRace USA.

### DIFF
--- a/src/mame/irem/travrusa.cpp
+++ b/src/mame/irem/travrusa.cpp
@@ -880,12 +880,12 @@ ROM_START( mototour )
 ROM_END
 
 /* The only difference on this set is a different PROM configuration on the sprites PCB:
-    Location mototour mototourb
+    Location mototour mototoura
           F1   74LS32      PROM
           H1   74LS74      PROM
           K2     PROM     empty
 */
-ROM_START( mototourb )
+ROM_START( mototoura )
 	ROM_REGION( 0x10000, "maincpu", 0 )
 	ROM_LOAD( "mt1-4.m3", 0x0000, 0x2000, CRC(fe643567) SHA1(2e47b6de43ff7fc1f070d34376fde697fc719b80) )
 	ROM_LOAD( "mt1-5.l3", 0x2000, 0x2000, CRC(38d9d0f5) SHA1(8b4531a28ff69df04a5eef687383dab57e0aa685) )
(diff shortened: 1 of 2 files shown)


cuavas:
Is it just a case of the organisation being different which could be addressed during ROM loading?  It would be good if someone could give this a look and see if it’s simple to get working before merging.

Osso13:
If you load them like this it will work:
	ROM_REGION( 0x0320, "proms", 0 )
	ROM_LOAD_NIB_HIGH( "prom.h1 ",  0x0000, 0x0100, CRC(2f98ddf0) SHA1(e90c3cebe3e788cbf8e23030f58a1153564207e2) ) // On tiles PCB
	ROM_LOAD_NIB_LOW( "prom.f1 ",   0x0000, 0x0100, CRC(adea1297) SHA1(8f365cc15cc3c26b388ba957d7cf3752584d5475) ) // On tiles PCB
	ROM_LOAD( "prom1.f1",           0x0200, 0x0020, CRC(a1130007) SHA1(9deb0eed75dd06e86f83c819a3393158be7c9dce) ) // On sprites PCB
	ROM_LOAD( "prom2.h2",           0x0220, 0x0100, CRC(76062638) SHA1(7378a26cf455d9d3df90929dc665870514c34b54) ) // On sprites PCB
That said, they have the exact same content.

cuavas:
> That said, they have the exact same content.

So the question is, do we keep a clone for this PROM arrangement?  It may be useful for people repairing systems if they encounter this variant.

clawgrip:
> FWIW you could use a BIOS option to load one arrangement of PROMs or the other and avoid the need for a clone.

Sounds good, will do that. Thx!!

cuavas (commented):
Should the alternate PROM arrangement be added to all the sets, or only to the ones it’s been seen on in the wild?

clawgrip:
> Should the alternate PROM arrangement be added to all the sets, or only to the ones it’s been seen on in the wild?

I've added it only to the sets I've found evidences of using both arrangements. Since it can be clearly seen on the PCB, it's easy to tell by looking at PCB pics.

Thanks

cuavas:
Have both bootleg versions really been seen with both PROM arrangements?  That would be pretty surprising.

clawgrip:
> Have both bootleg versions really been seen with both PROM arrangements? That would be pretty surprising.

Ah! You’re right, I’ve only seen this arrangement on original Irem/Tecfri/Williams.

Should it be removed from bootleg sets?

thx

cuavas:
> > Have both bootleg versions really been seen with both PROM arrangements? That would be pretty surprising.
> 
> Ah! You’re right, I’ve only seen this arrangement on original Irem/Tecfri/Williams.
> 
> Should it be removed from bootleg sets?
> 
> thx

Yes, I wouldn’t add the option to the bootleg sets if they haven’t been seen with both arrangements.

clawgrip:
> Yes, I wouldn’t add the option to the bootleg sets if they haven’t been seen with both arrangements.

Done, thanks!!

src/mame/irem/travrusa.cpp
```diff
@@ -787,9 +800,9 @@ ROM_START( travrusab2 ) // all ROMs match travrusa but the ones where differentl
 	ROM_LOAD( "10.3k3", 0x4000, 0x2000, CRC(6fcc9fdb) SHA1(88f878b9ebf07c5a16f8cb742016cac971ed3f10) )
 
 	ROM_REGION( 0x0320, "proms", 0 )
-	ROM_LOAD( "6349-2.1k2",    0x0000, 0x0200, CRC(c9724350) SHA1(1fac20cdc0a53d94e8f67b49d7dd71d1b9f1f7ef) ) // character palette - last $100 are unused
-	ROM_LOAD( "tbp18s030.3f1", 0x0200, 0x0020, CRC(a1130007) SHA1(9deb0eed75dd06e86f83c819a3393158be7c9dce) ) // sprite palette
-	ROM_LOAD( "mb7052.3h2",    0x0220, 0x0100, CRC(76062638) SHA1(7378a26cf455d9d3df90929dc665870514c34b54) ) // sprite lookup table
+	ROM_LOAD( "mmi6349.ij",   0x0000, 0x0200, CRC(c9724350) SHA1(1fac20cdc0a53d94e8f67b49d7dd71d1b9f1f7ef) ) // character palette - last $100 are unused
+	ROM_LOAD( "tbp18s.2",     0x0200, 0x0020, CRC(a1130007) SHA1(9deb0eed75dd06e86f83c819a3393158be7c9dce) ) // sprite palette
+	ROM_LOAD( "tbp24s10.3",   0x0220, 0x0100, CRC(76062638) SHA1(7378a26cf455d9d3df90929dc665870514c34b54) ) // sprite lookup table
```
cuavas, line 805: Why the change to the labels here?  This set previously had different labels to the parent and the other bootleg.  Were they found to be incorrect?

cuavas (changes requested):
Please look at the overall diff.  You’ve “restored” the PROM labels on the wrong bootleg.  Now `travrusab` and `travrusab2` have their PROM labels swapped with each other.

src/mame/irem/travrusa.cpp
clawgrip: My bad. Restored the original labels. Sorry.

clawgrip:
> Please look at the overall diff. You’ve “restored” the PROM labels on the wrong bootleg. Now `travrusab` and `travrusab2` have their PROM labels swapped with each other.

Ah! Yes, comitted before finished editing by mistake, but I saw the error, it's OK now

cuavas:
In the future, please check your changes carefully.  It’s far too easy for errors to get in like this, and noticing/fixing them after the fact is always far more difficult.

clawgrip:
> In the future, please check your changes carefully. It’s far too easy for errors to get in like this, and noticing/fixing them after the fact is always far more difficult.

Will do. Thanks a lot!!!!

PR #12003, opened 2024-02-03
apple/apple2e.cpp: Correctly reset IOU softswitches on system reset

xotmatrix:
Some IOU softswitches are not being reset when the system is reset. This can leave the display and annunciators in an incorrect state after a reset.

From a fresh start, enter the following into the monitor to enable 80-column video and then press RESET. Note the 80-column video remains enabled which is not correct.
```
*C00D:1
```

From a fresh start, enter the following into the monitor to draw MouseText characters to the screen and then press RESET. Note the MouseText characters remain enabled which is not correct.
```
*C00F:1 N 428:40 41 42 43 44 45
```

This patch corrects these issues by adding reset handling for the missing IOU softswitches:
- 80COL
- ALTCHR
- AN0
- AN1
- AN2
- AN3

In addition, reset code related to MMU, IOU, and LC softswitches has been reorganized into logical groups.

### DIFF
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -1119,16 +1119,42 @@ void apple2e_state::machine_start()
 
 void apple2e_state::machine_reset()
 {
+	// All MMU switches off (80STORE, RAMRD, RAMWRT, INTCXROM, ALTZP, SLOTC3ROM, PAGE2, HIRES, INTC8ROM)
+	// Sather, Fig 5.13
+	m_ramrd = false;
+	m_ramwrt = false;
+	m_altzp = false;
+	m_slotc3rom = false;
+	m_intc8rom = false;
+
+	// Certain IOU switches off (80STORE, 80COL, ALTCHR, PAGE2, HIRES, AN0, AN1, AN2, AN3)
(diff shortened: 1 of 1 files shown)


rb6502:
I think we have enough duplicated code now to break a lot of this out into a class member function and just call it from both the keyboard reset and machine_reset().

xotmatrix:
I'm all for a single reset routine. People have grown impatient with me for not accepting that they perform different tasks on some philosophical level.

rb6502:
It's not philosophical so much as a MAME architectural thing.

machine_reset() is intended to be a power-on everything-gets-hit reset, and there isn't a native concept of something like Control-Reset where only some components know about it.  The reset everything machine_reset() style is totally reasonable for all the stuff MAME emulates where it has no reset switch but you still would like to be able to reset it.  The A2 obviously isn't that. 

There's a *lot* of overlap so we definitely want both cases calling a common routine, but they can't be 100% identical since machine_reset() *should* put things like ROMSWITCH into a reasonable known state that a Control-Reset doesn't touch.  That's obviously a much longer list on the II/II Plus than the IIe/IIc, so hopefully for the IIe/IIc we can have most of what both paths do in the common routine.

PR #12009, opened 2024-02-04
lynx.xml: Fixed info for Raiden (CES Version).

FoxhackDN:


FoxhackDN:
Apparently the version of Raiden in MAME is an incomplete version, possibly dumped from a bootleg, which crashes and has glitches. See the following threads for more information:

https://forums.atariage.com/topic/199377-raiden-telegames-prototype-ces-v3/
https://forums.atariage.com/topic/221717-a-few-bits-from-my-atari-lynx-collection/

GDRI on Twitter posted about this a year ago, but I just heard about it today. He posted this video of someone playing their official Telegames cartridge on real hardware: https://www.youtube.com/watch?v=FAwDLh8M670

I played the game in MAME and saw a bit of different gameplay, the continue screen is also different (the current dump has a full screen continue screen, while the video shows one that lets you continue right away); the version we have also crashed hard after I got a game over. So the final version is not dumped,

PR #12012, opened 2024-02-04
Remove duplicate CD modes

### LINKED ISSUE #9084: cdrom.cpp: MODE2/2336 defined twice
In `cdrom_get_info_from_type_string`, the same key `MODE2/2336` is defined twice.

https://github.com/mamedev/mame/blob/5857dd9d2b763cd8211497a769055d73e7a503c2/src/lib/util/cdrom.cpp#L901

stonedDiscord:
Resolves #9084

### DIFF
--- a/src/lib/util/cdrom.cpp
+++ b/src/lib/util/cdrom.cpp
@@ -657,17 +657,7 @@ void cdrom_file::get_info_from_type_string(const char *typestring, uint32_t *trk
 		*trktype = CD_TRACK_MODE2_FORM_MIX;
 		*datasize = 2336;
 	}
-	else if (!strcmp(typestring, "MODE2/2336"))
-	{
-		*trktype = CD_TRACK_MODE2_FORM_MIX;
-		*datasize = 2336;
-	}
 	else if (!strcmp(typestring, "MODE2_RAW"))
-	{
-		*trktype = CD_TRACK_MODE2_RAW;
-		*datasize = 2352;
-	}
-	else if (!strcmp(typestring, "MODE2/2352"))
 	{
 		*trktype = CD_TRACK_MODE2_RAW;
 		*datasize = 2352;
--- a/src/lib/util/cdrom.cpp
+++ b/src/lib/util/cdrom.cpp
@@ -662,6 +662,11 @@ void cdrom_file::get_info_from_type_string(const char *typestring, uint32_t *trk
 		*trktype = CD_TRACK_MODE2_RAW;
 		*datasize = 2352;
 	}
+	else if (!strcmp(typestring, "MODE2/2352"))
+	{
+		*trktype = CD_TRACK_MODE2_RAW;
+		*datasize = 2352;
+	}
 	else if (!strcmp(typestring, "CDI/2352"))
 	{
 		*trktype = CD_TRACK_MODE2_RAW;


src/lib/util/cdrom.cpp
```diff
@@ -657,21 +657,11 @@ void cdrom_file::get_info_from_type_string(const char *typestring, uint32_t *trk
 		*trktype = CD_TRACK_MODE2_FORM_MIX;
 		*datasize = 2336;
 	}
-	else if (!strcmp(typestring, "MODE2/2336"))
-	{
-		*trktype = CD_TRACK_MODE2_FORM_MIX;
-		*datasize = 2336;
-	}
```
cuavas, line 664: I get that this code is unreachable so this won’t change behaviour, but what’s it supposed to be doing?  The place where it actually catches this is at line 630:
```c++
	else if (!strcmp(typestring, "MODE2/2336"))
	{
		*trktype = CD_TRACK_MODE2;
		*datasize = 2336;
	}
```

It gives `CD_TRACK_MODE2_FORM_MIX`.  Why is it trying to use `CD_TRACK_MODE2_FORM_MIX` for the same track type here?

src/lib/util/cdrom.cpp
```diff
@@ -657,21 +657,11 @@ void cdrom_file::get_info_from_type_string(const char *typestring, uint32_t *trk
...
 	else if (!strcmp(typestring, "MODE2_RAW"))
 	{
 		*trktype = CD_TRACK_MODE2_RAW;
 		*datasize = 2352;
 	}
-	else if (!strcmp(typestring, "MODE2/2352"))
-	{
-		*trktype = CD_TRACK_MODE2_RAW;
-		*datasize = 2352;
-	}
```
cuavas, line 674: There’s no other place that catches `MODE2/2352` so it isn’t a duplicate.  Why did you remove it?

stonedDiscord: oh sorry i confused it with MODE1/2352

stonedDiscord:
I've read the ISO9660 spec and I can't find a mention of a mixed form Mode2 track, can we get someone who is familiar with CDI and Dreamcast discs in here?

src/lib/util/cdrom.cpp
cuavas: Right, `CD_TRACK_MODE2` is the CD-ROM Mode 2 data track format, with 2,336 bytes of data per sector, and no additional error detection/correction data.

But do `CD_TRACK_MODE2_FORM1` and `CD_TRACK_MODE2_FORM2` really make sense?  The whole point of CD-ROM XA tracks is that they can contain a mixture of Form 1 and Form 2 sectors (with and without additional error correction data, respectively).  Or are these “artificial” formats for authoring, where the software interpreting the data is expected to generate a track containing only a single form?

And why would there be padding in a `CD_TRACK_MODE2_FORM_MIX` track?  The sector payload size is the same for both forms.  Form 1 is 8-byte subheader, 2,048-byte data, 4-byte checksum, 276-byte ECC = 2,336 bytes total.  Form 2 is 8-byte subheader, 2,324-byte data, 4-byte checksum = 2,336 bytes total.  Or is there some weirdess where the checksum and ECC bytes are treated as “padding” and replaced with the appropriate calculated content?

PR #12013, opened 2024-02-04
mattel/pixtermu.cpp: Add new not working system & software list

qufb:
New software list items marked not working (pixter_cart.xml)
-------------
Alphabet Forest [QUFB]
Clifford - The Big Red Dog [QUFB]
Cyberchase [QUFB]
Mazes, Puzzles & Games [QUFB]
SpongeBob SquarePants - Math Adventure [QUFB]

### DIFF
--- a/hash/pixter_cart.xml
+++ b/hash/pixter_cart.xml
@@ -14,8 +14,8 @@ Entries are sorted by serial number.
 		<year>2004</year>
 		<publisher>Fisher-Price</publisher>
 		<info name="serial" value="B2001"/>
-		<part name="cart" interface="pixtermu_cart">
-			<dataarea name="rom" size="0x400000">
+		<part name="cart" interface="pixter_cart">
+			<dataarea name="rom" size="0x400000" width="32" endianness="little">
 				<rom name="cs2.bin" size="0x400000" crc="0cde5364" sha1="05fa5f439204e5205fdf561d081714a626bc29b5"/>
 			</dataarea>
 		</part>
@@ -26,8 +26,8 @@ Entries are sorted by serial number.
 		<year>2004</year>
 		<publisher>Fisher-Price</publisher>
 		<info name="serial" value="B2003"/>
-		<part name="cart" interface="pixtermu_cart">
-			<dataarea name="rom" size="0x400000">
+		<part name="cart" interface="pixter_cart">
+			<dataarea name="rom" size="0x400000" width="32" endianness="little">
 				<rom name="cs2.bin" size="0x400000" crc="c4819349" sha1="65e8120ef7b3fc3c5aefe7e009d4864bafaee90a"/>
 			</dataarea>
 		</part>
@@ -38,8 +38,8 @@ Entries are sorted by serial number.
 		<year>2004</year>
 		<publisher>Fisher-Price</publisher>
 		<info name="serial" value="B2005"/>
-		<part name="cart" interface="pixtermu_cart">
-			<dataarea name="rom" size="0x400000">
+		<part name="cart" interface="pixter_cart">
+			<dataarea name="rom" size="0x400000" width="32" endianness="little">
 				<rom name="cs2.bin" size="0x400000" crc="e5f06785" sha1="b02ece2e79eabfedb8928ce41f08a6767f05f8a3"/>
 			</dataarea>
 		</part>
@@ -50,9 +50,9 @@ Entries are sorted by serial number.
 		<year>2004</year>
 		<publisher>Fisher-Price</publisher>
 		<info name="serial" value="B8515"/>
-		<part name="cart" interface="pixtermu_cart">
-			<dataarea name="rom" size="0x400000">
-				<rom name="cs2.bin" size="0x400000" crc="679d06c1" sha1="a65458d041e01d99fecf9f69527e9316c843a8ef"/>
+		<part name="cart" interface="pixter_cart">
+			<dataarea name="rom" size="0x200000" width="32" endianness="little">
+				<rom name="cs2.bin" size="0x200000" crc="679d06c1" sha1="a65458d041e01d99fecf9f69527e9316c843a8ef"/>
 			</dataarea>
 		</part>
 	</software>
@@ -62,8 +62,8 @@ Entries are sorted by serial number.
 		<year>200?</year>
 		<publisher>Fisher-Price</publisher>
 		<info name="serial" value="C7309"/>
-		<part name="cart" interface="pixtermu_cart">
-			<dataarea name="rom" size="0x400000">
+		<part name="cart" interface="pixter_cart">
+			<dataarea name="rom" size="0x400000" width="32" endianness="little">
 				<rom name="cs2.bin" size="0x400000" crc="e23ff6f3" sha1="38873874c8dd50d22789fe8fc0219daca0eda0cc"/>
 			</dataarea>
 		</part>
--- a/hash/pixter_cart.xml
+++ b/hash/pixter_cart.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+
+Partial list of released carts: https://service.mattel.com/us/results.aspx?Nty=1&Ntt=pixter&Ntx=mode%2bmatchany&N=0&No=0&Ntk=All&showall=true
+
+Entries are sorted by serial number.
+-->
+<softwarelist name="pixter_cart" description="Mattel Pixter cartridges">
+
+	<software name="alphafo" supported="no">
+		<description>Alphabet Forest</description>
+		<year>2004</year>
+		<publisher>Fisher-Price</publisher>
+		<info name="serial" value="B2001"/>
+		<part name="cart" interface="pixtermu_cart">
+			<dataarea name="rom" size="0x400000">
+				<rom name="cs2.bin" size="0x400000" crc="0cde5364" sha1="05fa5f439204e5205fdf561d081714a626bc29b5"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="clifford" supported="no">
+		<description>Clifford - The Big Red Dog</description>
+		<year>2004</year>
+		<publisher>Fisher-Price</publisher>
+		<info name="serial" value="B2003"/>
+		<part name="cart" interface="pixtermu_cart">
+			<dataarea name="rom" size="0x400000">
+				<rom name="cs2.bin" size="0x400000" crc="c4819349" sha1="65e8120ef7b3fc3c5aefe7e009d4864bafaee90a"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="spongema" supported="no">
+		<description>SpongeBob SquarePants - Math Adventure</description>
+		<year>2004</year>
+		<publisher>Fisher-Price</publisher>
+		<info name="serial" value="B2005"/>
+		<part name="cart" interface="pixtermu_cart">
+			<dataarea name="rom" size="0x400000">
+				<rom name="cs2.bin" size="0x400000" crc="e5f06785" sha1="b02ece2e79eabfedb8928ce41f08a6767f05f8a3"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="mazespg" supported="no">
+		<description>Mazes, Puzzles &amp; Games</description>
+		<year>2004</year>
+		<publisher>Fisher-Price</publisher>
+		<info name="serial" value="B8515"/>
+		<part name="cart" interface="pixtermu_cart">
+			<dataarea name="rom" size="0x400000">
+				<rom name="cs2.bin" size="0x400000" crc="679d06c1" sha1="a65458d041e01d99fecf9f69527e9316c843a8ef"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="cyberch" supported="no">
+		<description>Cyberchase</description>
+		<year>200?</year>
+		<publisher>Fisher-Price</publisher>
+		<info name="serial" value="C7309"/>
+		<part name="cart" interface="pixtermu_cart">
+			<dataarea name="rom" size="0x400000">
+				<rom name="cs2.bin" size="0x400000" crc="e23ff6f3" sha1="38873874c8dd50d22789fe8fc0219daca0eda0cc"/>
+			</dataarea>
+		</part>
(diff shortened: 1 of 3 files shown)


mamehaze:
ah, nice, I think we had one of these and a few games sent to Sean back in the day, but he didn't manage to dump the internal ROM..  looking over old emails

Fisher Price Pixter Color console
Fisher Price Pixter Color game cart: "Dino Draw"
Fisher Price Pixter Color game cart: "Mazes, Puzzles & Games"
Fisher Price Pixter Color game cart: "Creative Genius"
Fisher Price Pixter Color game cart: "Learning Starter"
Fisher Price Pixter game cart: "The Powerpuff Girls"
Fisher Price Pixter game cart: "Music Studio"
Fisher Price Pixter game cart: "Barbie Fashion Show"
Fisher Price Pixter game cart: "Learning Fun 2"
Fisher Price Pixter game cart: "Story Composer"

comes up at least, so if Sean didn't dump; those, maybe he still has them (assuming this is the same system we're talking about)

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,232 @@
...
+	memcpy(reinterpret_cast<uint32_t *>(m_bootrom.target()), memregion("bootrom")->base(), 0x2000);
+	memcpy(reinterpret_cast<uint32_t *>(m_ncs1.target()), memregion("ncs1")->base(), 0x400000);
+	if (m_cart->exists()) {
+		std::string region_tag;
+		memory_region *cart_rom = m_cart->memregion("cart:rom");
+		m_maincpu->space(AS_PROGRAM).install_rom(0x48000000, 0x483fffff, 0x400000, cart_rom->base());
+	}
+
+	m_apb = make_unique_clear<uint32_t[]>(0x27000/4);
+	save_pointer(NAME(m_apb), 0x27000/4);
```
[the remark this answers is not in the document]
qufb: "Advance Peripheral Bus", although the bridge itself exposes some registers.

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,232 @@
...
+}
+
+DEVICE_IMAGE_LOAD_MEMBER(pixter_multimedia_state::cart_load)
+{
+	uint32_t const size = m_cart->common_get_size("rom");
+
+	m_cart->rom_alloc(size, GENERIC_ROM32_WIDTH, ENDIANNESS_LITTLE);
+	m_cart->common_load_rom(m_cart->get_rom_base(), size, "rom");
+
+	return std::make_pair(std::error_condition(), std::string());
```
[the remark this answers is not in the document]
qufb: At least one cart is 0x200000, so I've changed the allocation to be dynamically adjusted to cart size.

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,232 @@
...
+	// External SRAM
+	map(0x20000000, 0x203fffff).ram().share("ndcs0");
+	// nCS0 (Unused NAND Flash?)
+	map(0x40000000, 0x403fffff).rom().share("ncs0");
+	// nCS1 (Chip-On-Board ROM)
+	map(0x44000000, 0x443fffff).rom().share("ncs1");
+	// nCS2 (Cart ROM)
+	map(0x48000000, 0x483fffff).rom().share("ncs2");
+	// nCS3 (Unused?)
+	map(0x4c000000, 0x4c3fffff).rom().share("ncs3");
```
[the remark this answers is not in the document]
qufb: Not sure what's the convention here, but I left the unmapped regions commented just to document them here.

qufb:
> ah, nice, I think we had one of these and a few games sent to Sean back in the day, but he didn't manage to dump the internal ROM.. looking over old emails
> 
> Fisher Price Pixter Color console Fisher Price Pixter Color game cart: "Dino Draw" Fisher Price Pixter Color game cart: "Mazes, Puzzles & Games" Fisher Price Pixter Color game cart: "Creative Genius" Fisher Price Pixter Color game cart: "Learning Starter" Fisher Price Pixter game cart: "The Powerpuff Girls" Fisher Price Pixter game cart: "Music Studio" Fisher Price Pixter game cart: "Barbie Fashion Show" Fisher Price Pixter game cart: "Learning Fun 2" Fisher Price Pixter game cart: "Story Composer"
> 
> come up as 'secured and sent to Sean' at least, so if Sean didn't dump; those, maybe he still has them (assuming this is the same system we're talking about)

It's likely a very similar system. Pixter Multi-Media is backwards compatible with Pixter Color carts, and probably the original Pixter as well (there was an adapter made for Pixter Color, don't know if it works with Pixter Multi-Media).

I've written some [dumping notes](https://qufb.gitlab.io/writeups/pixter) that might be helpful for dumping Pixter Color and those carts. It [also has JTAG](https://elinux.org/Pixter_Dev_Board), although the pins will have to be traced for LH75411, as it seems that wiki didn't document the pinout.

hash/pixter_cart.xml
cuavas, line 1: You need to set the width and Endianness of the ROM regions in the software list to match the CPU or it will cause portability issues (the ARM7 device you’re using is 32-bit little Endian).

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+void pixter_multimedia_state::machine_start()
+{
+	if (m_cart->exists()) {
+		std::string region_tag;
+		memory_region *cart_rom = m_cart->memregion("cart:rom");
+		uint32_t const size = m_cart->common_get_size("rom");
+		m_maincpu->space(AS_PROGRAM).install_rom(0x48000000, 0x48000000 + size - 1, size, cart_rom->base());
+	}
```
cuavas, line 115: You don’t need `common_get_size` here – you can use `cart_rom->bytes()` to get the size of the region in bytes directly.

I’m not sure the mirroring specified here is going to do exactly what you want:
* For a 0x400000 byte cartridge, the mirror bits will be 0x400000, so it will be mapped at 0x48000000-0x483fffff and 0x48800000-0x487fffff.
* For a 0x200000 byte cartridge, the mirror bits will be 0x200000, so it will be mapped at 0x48000000-0x481fffff and 0x48800000-0x483fffff.

I would expect you’d want it mirrored across the same area irrespective of the cartridge size.

The `device_generic_cart_interface::install_non_power_of_two` helper is designed to help deal with this (it works for powers of two as well).  You can use it like:
```c++
		device_generic_cart_interface::install_non_power_of_two<0>(
				cart_rom->bytes(),
				0x007f'ffff,
				0,
				0x4800'0000,
				[this, cart_rom] (offs_t begin, offs_t end, offs_t mirror, offs_t src)
				{
					m_maincpu->space(AS_PROGRAM).install_rom(begin, end, mirror, cart_rom->base() + src);
				});
```

(I haven’t tested that with this driver, I’m going by memory.)

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,205 @@
...
+void pixter_multimedia_state::pixter_multimedia(machine_config &config)
+{
+	// User's Guide - 1.3 Clock Strategy - AHB Fast CPU Clock (FCLK)
+	ARM7(config, m_maincpu, 76'205'000);
+	m_maincpu->set_addrmap(AS_PROGRAM, &pixter_multimedia_state::arm7_map);
+
+	GENERIC_CARTSLOT(config, m_cart, generic_plain_slot, "pixter_cart");
+	m_cart->set_endian(ENDIANNESS_LITTLE);
+	m_cart->set_width(GENERIC_ROM32_WIDTH);
+	m_cart->set_must_be_loaded(false);
```
cuavas, line 188: You’re still going to need `m_cart->set_device_load` to deal with loose software (i.e. not loaded from the software list).  It’s possible to just refuse to load it (in which case you need to actually check for that and return an error).  It’s better to support it though, as it makes it easier to test new dumps, develop homebrew software, etc. without needing to edit the software list for every change.

src/mame/mattel/pixtermu.cpp
cuavas: As for what I can’t remember, it’s whether the offsets need to be shifted for the address space width or only if the address space uses shifted addresses.  If that installs at the wrong addresses, it probably needs to be:
```c++
		device_generic_cart_interface::install_non_power_of_two<2>(
				cart_rom->bytes() >> 2,
				0x007f'ffff >> 2,
				0,
				0x4800'0000,
				[this, cart_rom] (offs_t begin, offs_t end, offs_t mirror, offs_t src)
				{
					m_maincpu->space(AS_PROGRAM).install_rom(begin, end, mirror, &cart_rom->as_u32() + src);
				});
```

qufb: First suggestion works. I rechecked on hardware and the mirroring covers the whole range up to the start of the next nCS address map, so I just needed to adjust the mask.

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,205 @@
...
+	m_remap_view[0](0x00000000, 0x00001fff).rom().region("bootrom", 0);
+	m_remap_view[1](0x00000000, 0x003fffff).ram().share("ndcs0");
+	m_remap_view[2](0x00000000, 0x00003fff).ram().share("internal_sram");
+	m_remap_view[3](0x00000000, 0x003fffff).share("ncs0");
+	m_remap_view[4](0x00000000, 0x003fffff).rom().region("ncs1", 0);
+
+	// External SRAM
+	map(0x20000000, 0x203fffff).ram().share("ndcs0");
+	// nCS0 (Unused NAND Flash?)
+	map(0x40000000, 0x403fffff).share("ncs0");
```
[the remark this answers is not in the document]
qufb: I left it in since the firmware remap function supports it. It's just unpopulated on this model (if any actually have a port for it).

src/mame/mattel/pixtermu.cpp
qufb: Maybe I misunderstood you previous comment. I've added a guard if it was loaded via software list, to just return without doing any allocation, was that the intention?

src/mame/mattel/pixtermu.cpp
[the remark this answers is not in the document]
cuavas: Yeah, but having the `.share` will allocate a block of RAM that isn’t being used for anything, but still gets put in save states, shows in debugger memory view source lists, etc.  If there was actually Flash there, you’d need to use a NAND Flash device or something.  It’s better to just leave a comment that the address range can be used for Flash.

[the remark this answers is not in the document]
qufb: Fair enough.

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+	if (!image.loaded_through_softlist()) {
+	   uint32_t const size = m_cart->common_get_size("rom");
+
+	   m_cart->rom_alloc(size, GENERIC_ROM32_WIDTH, ENDIANNESS_LITTLE);
+	   m_cart->common_load_rom(m_cart->get_rom_base(), size, "rom");
+	}
+
+	return std::make_pair(std::error_condition(), std::string());
+}
```
cuavas, line 144: OK, we’re almost there, but there are still a few issues:
* No error checking to ensure that the software list entry is usable (people can edit software lists, and easily screw things up)
* `rom_alloc` won’t give the same tag as the region loaded from the software list
* `common_load_rom` doesn’t report errors
* You need to remember to swap the data on big Endian hosts

Something like this should work, but I haven’t tested it, just coding from memory:
```c++
DEVICE_IMAGE_LOAD_MEMBER(pixter_multimedia_state::cart_load)
{
	uint64_t length;
	memory_region *cart_rom = nullptr;
	if (m_cart->loaded_through_softlist()) {
		cart_rom = m_cart->memregion("rom");
		if (!cart_rom)
			return std::make_pair(image_error::BADSOFTWARE, "Software list item has no 'rom' data area");
		length = cart_rom->bytes();
	} else {
		length = m_cart->length();
	}

	if (!length)
		return std::make_pair(image_error::INVALIDLENGTH, "Cartridges must not be empty");
	if (length & 3)
		return std::make_pair(image_error::INVALIDLENGTH, "Unsupported cartridge size (must be a multiple of 4 bytes)");

	if (!m_cart->loaded_through_softlist()) {
		cart_rom = machine().memory().region_alloc(m_cart->subtag("rom"), length, 4, ENDIANNESS_LITTLE);
		if (!cart_rom)
			return std::make_pair(std::errc::not_enough_memory, std::string());

		uint32_t *const base = reinterpret_cast<uint32_t *>(cart_rom->base());
		if (m_cart->fread(base, length) != length)
			return std::make_pair(std::errc::io_error, "Error reading cartridge file");

		if (ENDIANNESS_NATIVE != ENDIANNESS_LITTLE)
		{
			for (uint64_t i = 0; (length / 4) > i; ++i)
				base[i] = swapendian_int32(base[i]);
		}
	}

	return std::make_pair(std::error_condition(), std::string());
}
```

src/mame/mattel/pixtermu.cpp
```diff
@@ -0,0 +1,223 @@
...
+		m_remap_view.select(data);
+	} else {
+		logerror("Unexpected remap %d\n", data);
+	}
+}
+
+void pixter_multimedia_state::machine_start()
+{
+	if (m_cart->exists()) {
+		memory_region *cart_rom = m_cart->memregion("cart:rom");
```
cuavas, line 112: Shouldn’t this be `m_cart->memregion("rom")`?  When software list items are loaded, the data area tags are interpreted relative to the slot.  (The extra `cart:` part is only in the tag of redundant additional region that `common_load_rom` allocates.)

cuavas (commented):
This is almost there, just needs some more error checking in the image load callback.

src/mame/mattel/pixtermu.cpp
cuavas: Yes, I realise it would be useful to have a helper function to implement this pattern or something similar rather than the problematic `common_load_rom`.  Replacing the current bus/generic stuff is a long-term goal.

qufb: Works fine, thanks.

src/mame/mattel/pixtermu.cpp
qufb: I was missing the part of allocating with `m_cart->subtag("rom")` when loading manually, which makes this work, otherwise memregion wouldn't be able to find it. With the other part above it works for both cases now.

PR #12014, opened 2024-02-05
rm/rm380z.cpp: Add support for 4-line and backwards hw scrolling

RobinSergeant:
Previously only full screen scrolling was supported in one direction, and even this went a little haywire with CP/M paging, e.g. when listing the contents of a disk with more than 23 files.  This PR fixes that and adds 4-line scrolling (GRAFIX mode) and backwards scrolling.

4-line scrolling is needed for the Front Panel to work properly, otherwise rows from the bottom of the screen corrupt the display area above when scrolling!  This is why it looks so bad in this odd YouTube video:

https://www.youtube.com/watch?v=zzl37lfEymI

Notice how it gets corrupted at around 1:46 with a duplicate IO port line.  This has scrolled up from below :-)

Also notice that the 'O' in "IO" is missing.  This is a different issue that I have also fixed with this PR.  Basically 'O' can also be stored as ASCII nul, but as we don't have the real character ROM we need to manually cater for that.  Anyway, now the front panel seems to behave properly and a history of the commands you type is displayed at the bottom.

To test scrolling backwards I wrote the following bit of assembler:

```
OUTC:   equ 0x01
KBDW:   equ 0x21
WIDTH:  equ 0x34
GRAFIX: equ 0x0D

emt: macro routine
rst 0x30
defb routine
endm

org 0x100

ld a, 0x00          ; set 40 character mode (use 0x01 for 80 column mode)
emt WIDTH

emt GRAFIX          ; set 4-line scrolling mode

loop:   emt KBDW    ; read character from keyboard (ctrl+k/j to move cursor up/down)
        emt OUTC    ; and output to screen
        cp 'q'
        jr nz, loop ; repeat until 'q' pressed
```
When started from 0x100 it allows 4-line scrolling to be tested in both directions.  When started from 0x106 it does the same with full screen scrolling.  Control character are interpreted by OUTC so you can move the cursor around the screen etc.

I also tested page scrolling in CP/M and full screen scrolling at the monitor prompt etc.

I've tried to explain how hw scrolling works (to the best of my understanding at least) in the code comments.  I figured most of that out by setting watch points on the IO ports and experimenting.  Please ask if anything isn't clear.

### DIFF
--- a/src/mame/rm/rm380z_v.cpp
+++ b/src/mame/rm/rm380z_v.cpp
@@ -147,8 +147,8 @@ void rm380z_state::scroll_videoram_up(int n, bool clear)
 	if (clear)
 	{
 		// the last line is filled with spaces
-		std::memset(m_vramchars[RM380Z_SCREENROWS - 1], 0x20, RM380Z_SCREENCOLS);
-		std::memset(m_vramattribs[RM380Z_SCREENROWS - 1], 0x00, RM380Z_SCREENCOLS);
+		std::memset(m_vramchars[RM380Z_ROW_MAX], 0x20, RM380Z_SCREENCOLS);
+		std::memset(m_vramattribs[RM380Z_ROW_MAX], 0x00, RM380Z_SCREENCOLS);
 	}
 }
 
@@ -192,7 +192,6 @@ a 4-line scroll.
 Only COS 4.x suports hw scrolling and COS 3.x implemented the same calls in sw.
 
 */
-
 void rm380z_state::check_scroll_register()
 {
 	const uint8_t r[3] = { m_old_old_fbfd, m_old_fbfd, m_fbfd };
@@ -215,6 +214,7 @@ void rm380z_state::check_scroll_register()
 		address_space &space = m_maincpu->space(AS_PROGRAM);
 		uint8_t top_row = space.read_byte(0xff6a);
 		uint8_t rows_to_scroll = space.read_byte(0xff0e);
+		m_rows_to_scroll = 0;
 
 		if (((r[2] > r[1]) && (r[2] - r[1] != RM380Z_ROW_MAX)) || ((r[2] == 0) && (r[1] == RM380Z_ROW_MAX)))
 		{
--- a/src/mame/rm/rm380z_v.cpp
+++ b/src/mame/rm/rm380z_v.cpp
@@ -138,43 +138,6 @@ void rm380z_state::decode_videoram_char(int row, int col, uint8_t& chr, uint8_t 
 	}
 }
 
-/* hw scrolling notes -
-
-Scrolling can be switched between full screen and 4-line mode using the GRAFIX and SCROLL emt firmware calls.
-
-The lower 5-bits of port FBFD serve as a scroll counter, and in normal operation the counter increments
-when moving down the screen (to scroll text upwards).  It then decrements in the reverse direction, e.g.
-
-a sequence of 0, 1, 2, 3 might be seen when scrolling text upwards 3 times.
-and 3, 2, 1, 0 when scrolling text downwards 3 times.
-
-In 4-line mode only the bottom 4 rows scroll and this is used for command input in the front panel
-and also with HRG (not yet emulated) as high resolution graphics only plots over the first 20 rows. In
-this mode the counter sequence flips between two values instead, eg.
-
-a sequence of 0, 1, 0, 1, 0, 1, 0 might be seen when scrolling text upwards 3 times.
-and 1, 0, 1, 0, 1, 0, 1 when scrolling text downwards 3 times.
-
-4-line scrolling is mostly implemented in SW, but HW full screen scrolling is used to help with this which
-explains the flipped counter sequence.
-
-Only COS 4.x suports hw scrolling and COS 3.x implemented the same calls in sw.
-
-*/
-void rm380z_state::check_scroll_register()
-{
-	const uint8_t r[3] = { m_old_old_fbfd, m_old_fbfd, m_fbfd };
-
-	if (((r[2] > r[1]) && (r[2] - r[1] != RM380Z_ROW_MAX)) || ((r[2] == 0) && (r[1] == RM380Z_ROW_MAX)))
-	{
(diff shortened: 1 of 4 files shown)


RobinSergeant:
Just noticed that I accidentally reverted some formatting improvements made in the previous commit.  I will push another commit to fix this.  It must have happened during the merge.

RobinSergeant:
> Just noticed that I accidentally reverted some formatting improvements made in the previous commit. I will push another commit to fix this. It must have happened during the merge.

Formatting now corrected by second commit

src/mame/rm/rm380z_v.cpp
```diff
@@ -141,33 +138,109 @@ void rm380z_state::decode_videoram_char(int row, int col, uint8_t& chr, uint8_t
...
+	std::memmove(m_vramchars[RM380Z_SCREENROWS - n], m_vramchars[RM380Z_SCREENROWS - (n + 1)], n * RM380Z_SCREENCOLS);
+	std::memmove(m_vramattribs[RM380Z_SCREENROWS - n], m_vramattribs[RM380Z_SCREENROWS - (n + 1)], n * RM380Z_SCREENCOLS);
+
+	if (clear)
+	{
+		// the first line is filled with spaces
+		std::memset(m_vramchars[RM380Z_SCREENROWS - (n + 1)], 0x20, RM380Z_SCREENCOLS);
+		std::memset(m_vramattribs[RM380Z_SCREENROWS - (n + 1)], 0x00, RM380Z_SCREENCOLS);
+	}
+}
```
cuavas, line 167: In reality, the hardware won’t copy the contents of video memory like this – it will use the contents of the scroll registers to adjust addressing.  Rather than further bake in the flawed approach, can you please rework the code to use the scroll registers to adjust addressing when accessing video memory and when drawing?

You can see examples of this in source/mame/osborne/osborne1.cpp and src/mame/tandberg/tdv2100_disp_logic.cpp (although they only support full screen scrolling).

cuavas (changes requested):

src/mame/rm/rm380z_v.cpp
RobinSergeant: I can certainly take a look at that, but will have to do some research.  Would it then in effect mean that after scrolling future screen writes to row 23 would be stored m_vramchars[0] and the scroll registers would tell us that row 0 maps to m_vramchars[1] and row 23 maps to m_vramchars[0] etc?

By the way I've just discovered another use for FBFD.  It's used to send character data when changing the user definable characters when bit 3 of FBFC is set - another unimplemented feature that I am looking into!

RobinSergeant: I've now got a scroll register working for full screen and it does seem a lot better.  I will continue working on this.

RobinSergeant: Please review latest commit.  I was able to use a register for full screen scrolling and that was a great idea.  However, the 4-line scrolling is odd because the firmware (COS) copies one of the lines itself.  After telling the HW to scroll it copies line 23 to line 22, and then clears line 23.  For this to work the HW must be changing lines 20 and 21?  This is what I am now doing, but it's annoying.  I did have a second scroll register setup, but alas I don't think it's possible to use that.  What do you think?

RobinSergeant: Actually I think it is possible with 3 registers, but I'll need to spend a little more time on the algorithm...

RobinSergeant: After banging my head against a wall thinking it was impossible I finally realised that 4-line HW scrolling is impossible and it's actually mostly done by the firmware!  However, to help it do that it uses full screen scrolling to shift things around.  This is why we see the scroll counter increment and then decrement.  I put some trace to show this alongside vram access and it suddenly made sense :-)  This is what we now see when moving up to scroll text down (lines held "1", "2", "3", and "4"):

```
Reading row: 22, Data: 3
Writing row: 23, Data: 3
Scrolling down
Reading row: 22, Data: 2
Writing row: 23, Data: 2
Scrolling up
Reading row: 20, Data: 1
Writing row: 21, Data: 1
Writing row: 20, Data: (blank char)
```

So with just one scroll register we can do everything and it's really simple.  No need to know or care if the firmware is trying to do a 4-line scroll or not.  As long as we implement full screen scrolling everything else just works.  Thank you for the great suggestion.  I've certainly learned something anyway.

Please check the latest version.  Btw - some of the automated checks seem to fail but they are not related to my change.  I think the build must have been a bit broken when I made the branch.

RobinSergeant: One further revelation is that the value we get from port FBFD is probably a scroll register value maintained by the firmware.  I think we can use that directly, e.g.

```
	case 0xfd:      // screen line counter (?)
		//printf("%s FBFC [%2.2x] FBFDw[%2.2x] FBFE [%2.2x] writenum [%4.4x]\n",machine().describe_context().c_str(),m_port0,data,m_fbfe,writenum);

		data &= m_fbfd_mask;

		// ignore repeated values and updates while bit 4 of port 0 is set
		// (counter goes crazy when bit 4 is set so must have a different purpose then)
		if ((data != m_fbfd) && !(m_port0 & 0x10))
		{
			m_old_old_fbfd=m_old_fbfd;
			m_old_fbfd=m_fbfd;
			m_fbfd=data;

			writenum++;

			//check_scroll_register();
			m_vram.set_scroll_register(m_fbfd);
```
Just tried it quickly and it seems to work fine.  I could make this optimisation now or another time.  I was planning to look into the character generator next and that also uses the same port.

RobinSergeant: I've made another commit to remove redundant code and member variables.  Using the counter directly works really well and is of course what the real hardware did. Hopefully that should be everything unless you have any further comments.

RobinSergeant:
Please see https://github.com/mamedev/mame/pull/12031 for further work on this driver.  You might want to close this PR without delivering after review and only deliver the new PR.  Not sure what is easier.

RobinSergeant:
Closed as super-seeded by later PR.

PR #12025, opened 2024-02-11
Renamed galaxy driver and software list to galaksija

mmicko:
- galaxy: renamed galaxy driver and software list to galaksija [Miodrag Milanovic]
- galaksija: added official ROM images and char roms used for kit and commercial version [Miodrag Milanovic]

NOTE: rom we used so far is quite common among users but was not official ver 29, still left it due to being common.

### DIFF
--- a/hash/galaksija.xml
+++ b/hash/galaksija.xml
@@ -4,13 +4,13 @@
 license:CC0-1.0
 -->
 
-<softwarelist name="galaxy" description="Galaksija &amp; Galaksija Plus cassettes">
+<softwarelist name="galaksija" description="Galaksija &amp; Galaksija Plus cassettes">
 
 	<software name="asteroid">
 		<description>Asteroidi</description>
 		<year>1984</year>
 		<publisher>GAC Software</publisher>
-		<part name="cass1" interface="galaxy_cass">
+		<part name="cass1" interface="galaksija_cass">
 			<dataarea name="cass" size="1298">
 				<rom name="asteroidi.gtp" size="1298" crc="70aa80a3" sha1="1bc96cdf7602bbb24b02f5f9cb9aac709f3a7080"/>
 			</dataarea>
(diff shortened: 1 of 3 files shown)


mmicko:
Name was left from long time ago (from when it was initially added in MESS). Thing is that "galaxy" name was never used for computer (even being literal translation of name). It was actually more common that people were not aware of its existence due to it's wrong naming in MAME.

cuavas (commented):
Do you want to abbreviate the short name of the parent to keep it within eight characters?

If the only difference between the Voja Antonic and Elektronika inzenjering versions is the character generator ROM, do they need to be separate systems?  You could make the Elektronika inzenjering version a BIOS option.  The BIOS setting applies to all regions, so you’d need to make the `chargen` use `ROMX_LOAD` and `ROM_BIOS`.

mmicko:
@cuavas I have addressed "Elektronika inzenjering" version (not sure if there is way to make definitions shorter when combinations are used)
Making "galaksija" one char shorter, would just make it even more ugly/problematic than "galaxy".
Is name fine to stay as is ?

cuavas:
> Making "galaksija" one char shorter, would just make it even more ugly/problematic than "galaxy". Is name fine to stay as is ?

Well, when we allowed parent short names to exceed eight characters, it was supposed to make life easier for cases where there’s some prefix markup embedded in the short names.  For example Scorpion 5 gambling machines having the `sc5` prefix, or PlayChoice 10 games having the `pc_` prefix.  There was an agreement that it wouldn’t just lead to “short name length creep”.

Why is an abbreviated short name problematic?  The description contains the full name of the system in any case.  Short names are necessarily almost invariably abbreviated.  The number of cases where the short name *isn’t* abbreviated is relatively small.

(As for how I’d abbreviate it, the usual way is to drop a vowel somewhere that doesn’t make it unrecognisable.)

mmicko:
In this case I would rather than keep old name "galaxy" and just make this PR just adding proper bios versions . Ok with you ?

PR #12027, opened 2024-02-12
ussr/bk.cpp: K1801VP1-014 keyboard controller and MS7008 keyboard

shattered:


src/mame/ussr/1801vp014.cpp
```diff
@@ -0,0 +1,324 @@
...
+{
+	m_key_code = 0;
+	m_kbd_state = 0;
+	m_key_irq_vector = 0;
+	m_rxrdy = CLEAR_LINE;
+
+	m_write_virq(m_rxrdy);
+
+	reset_key_state();
+	start_processing(attotime::from_hz(2'400)); // FIXME real device does not use polling
```
cuavas, line 205: How does it work if it doesn’t scanning a key matrix?  Does it have a dedicated input line per key?  Does it keep all rows driven until it detects a change on a column and then scan for the row that changed?

src/mame/ussr/1801vp014.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+//-------------------------------------------------
+//  device_start - device-specific startup
+//-------------------------------------------------
+
+void k1801vp014_device::device_start()
+{
+	// save state
+	save_item(NAME(m_key_code));
+	save_item(NAME(m_kbd_state));
```
cuavas, line 187: Won’t you need to save `m_key_irq_vector` and `m_rxrdy` as well?  They seem to be live state.

src/mame/ussr/1801vp014.h
```diff
@@ -0,0 +1,77 @@
...
+	virtual void key_break(uint8_t row, uint8_t column) override;
+
+private:
+	required_ioport m_io_kbdc;
+	devcb_write_line m_write_virq;
+	devcb_write_line m_write_keydown;
+	devcb_write_line m_write_halt;
+	devcb_write8 m_write_data;
+
+	line_state m_rxrdy;
```
cuavas, line 66: `line_state` is really only for things that support the `HOLD_LINE` hack.  It also isn’t explicitly sized and will cause an issue with save states.  Just use a `uint8_t` or something.

src/mame/ussr/1801vp014.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+//**************************************************************************
+//  MACROS / CONSTANTS
+//**************************************************************************
+
+//**************************************************************************
+//  LIVE DEVICE
+//**************************************************************************
+
+INPUT_PORTS_START(ms7008)
```
cuavas, line 38: It would be nice to have Russian key cap labels for the letter keys as well.  Also, this should be `static INPUT_PORTS_START(...)`.

src/mame/ussr/1801vp014.cpp
shattered: I don't quite understand it but it looks like option 3 with extra steps (https://github.com/1801BM1/k1801/blob/master/014/ describes the process)

cuavas: Oh, interesting.  When it detects the key switch connecting the row pull-up resistor to the column line, then it actively drives the column low and senses which row line gets pulled down.  It then holds the state until the key is released.

It doesn’t need scanning logic at all, at the cost of not even attempting to deal with simultaneous key-presses besides the modifiers.

src/mame/ussr/1801vp014.cpp
shattered: Yeah -- done

PR #12031, opened 2024-02-13
rm/rm380z.cpp: Use real character ROM for COS 4.0, correct screen resolution and add support for user defined characters

RobinSergeant:
I've managed to obtain a dump of the real character ROM for COS 4.0 courtesy of the Research Machines user group, and this PR pulls this in and updates the display code to use it.

What is the correct process for updating a rom set?  This my working version that includes the new character rom:

[rm380z.zip](https://github.com/mamedev/mame/files/14268302/rm380z.zip)

The real characters are actually larger than the fake set we used before (8x10 rather then 5x9) and even include 2 extra rows to be used when the underline attribute is set.  This all makes the display handling much easier and improves the display greatly.

So with these characters 80 column mode has a display resolution of 640x240 and 40 column mode uses 320x240.

For COS 3.4 I've set the display resolution to 240x240 which I think is correct because COS 3.0 only supported the VDU40 adaptor which uses the TI 74LS262 character ROM.  These characters are 5x9, padded to 6x10 for display.  It's a shame we don't have the real ROM for this as well, but at the moment I'm concentrating on getting COS 4.0 to work properly anyway!

The new ROM only includes characters 0x00 to 0x79 because the additional teletext characters (0x80 to 0xFF) are actually generated by COS as user defined characters.  I've also added support for user defined characters so that these continue to work.

This PR includes all my changes for HW scrolling from the last PR which is still in review (https://github.com/mamedev/mame/pull/12014), but I don't want to close that until the review comment has been cleared.  I can do any further re-work under this PR.

### DIFF
--- a/src/mame/rm/rm380z.h
+++ b/src/mame/rm/rm380z.h
@@ -29,6 +29,31 @@ Research Machines RM 380Z
 //
 //
 
+template <int ROWS, int COLS>
+class rm380z_vram
+{
+public:
+	void set_char(int row, int col, uint8_t data) { m_chars[get_row(row)][col] = data; }
+	void set_attrib(int row, int col, uint8_t data) { m_attribs[get_row(row)][col] = data; }
+	void set_scroll_register(uint8_t value) { m_scroll_reg = value; }
+	void reset();
+
+	uint8_t get_char(int row, int col) const { return m_chars[get_row(row)][col]; }
+	uint8_t get_attrib(int row, int col) const { return m_attribs[get_row(row)][col]; }
+private:
+	int get_row(int row) const { return (row + m_scroll_reg) % ROWS; }
+
+	uint8_t m_chars[ROWS][COLS];
+	uint8_t m_attribs[ROWS][COLS];
+	uint8_t m_scroll_reg = 0;
+};
+
+template <int ROWS, int COLS>
+void rm380z_vram<ROWS, COLS>::reset()
+{
+	memset(m_attribs, 0, sizeof(m_attribs));
+	memset(m_chars, 0x80, sizeof(m_chars));
+}
 
 class rm380z_state : public driver_device
 {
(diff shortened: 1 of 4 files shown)


RobinSergeant:
Just fixed a logic error with my underline attribute check.  I was only swapping the last row (instead of the last two rows) when underlining.  Visually it made no difference as only the last row is normally different, but still a bug!  Everything should be ok now and this is ready for review.

src/mame/rm/rm380z.h
```diff
@@ -29,6 +29,31 @@ Research Machines RM 380Z
...
+	uint8_t m_attribs[ROWS][COLS];
+	uint8_t m_scroll_reg = 0;
+};
+
+template <int ROWS, int COLS>
+void rm380z_vram<ROWS, COLS>::reset()
+{
+	memset(m_attribs, 0, sizeof(m_attribs));
+	memset(m_chars, 0x80, sizeof(m_chars));
+}
```
cuavas, line 56: Can you nest this class inside `rm380z_state` to keep it out of the global namespace?

src/mame/rm/rm380z_m.cpp
```diff
@@ -7,7 +7,8 @@ RM 380Z machine
 
 */
 
-
+//#include <iostream>
+//#include <bitset>
```
cuavas, line 11: Are these leftovers from something?  At any rate, `#include`ing things before the prefix header is problematic.

cuavas:
This does look like a big simplification that will be nice to get in.

src/mame/rm/rm380z.h
RobinSergeant: yes, I've now done that.

src/mame/rm/rm380z_m.cpp
RobinSergeant: Yes, leftovers from my debug tracing.  Now removed.  std::bitset is quite useful for printing binary numbers in trace.

RobinSergeant:
> This does look like a big simplification that will be nice to get in.

Great, thanks for reviewing @cuavas.  I've addressed your comments so should be good to go in now.  Do I need to do anything to make the updated romset available?  It's attached to this PR.

simzy39:
Usually, the way to submit data/roms is by following this guide: https://wiki.mamedev.org/index.php/Submitting_Source_Code#Submitting_ROM_or_disk_dumps

RobinSergeant:
> Usually, the way to submit data/roms is by following this guide: https://wiki.mamedev.org/index.php/Submitting_Source_Code#Submitting_ROM_or_disk_dumps

Thanks @simzy39, I've followed that guide and emailed the rom.

PR #12034, opened 2024-02-14
sound/discrete.cpp: Use appropriate memory barriers for task synchronisation

### LINKED ISSUE #11373: src/devices/sound/discrete.cpp bad address and threading issues
### MAME version

0.255 mame0255-62-g9ff8931fa-50-dirty

### System information

odroid n2l  6 core arm ubuntu 22.04 kern 5.15 en with osd sdl alsa via hdmi audio out

### INI configuration details

```INI
all defaults except
audiodriver alsa
samplerate 48000
samples 1
waitvsync 1
syncrefresh 1
```

### Emulated system/software

galaxian,mooncrst

### Incorrect behaviour

Few moments into emulation, negative samples exception is thrown. Turning on debug, isnan exceptions also thrown. I suspect this is a threading/lock issue and use of invalid pointers (I am backed up by two fixme warnings re ptr math done before input list is actually initialized).  I am trying to fix it, but I am not an expert (yet). (For example, I've forced number of samples to always have correct pointers, but it has side effects.)

### Expected behaviour

No pointer math should result in invalid (out of bounds) access resulting in a segfault or underflow.  The glitch is "audible" on x86 linux as a chirp on emu launch. Unless someone beats me to it, I will fix this.

### Steps to reproduce

- launch mame (debug is good) with mooncrst on arm arch64 kern 5.15 smp preemp
- hear the first chirp after rom check (chirp is due to uninit data - infrequently isnan exception thrown)
- wait (usually within 5 minutes) and samples negative is thrown

### Additional details

_No response_

cuavas:
The synchronisation barriers used in the discrete sound device aren’t really thought out properly.  There are two variable where concurrent accesses matter:
* `output_buffer::ptr`, used to determine the number of samples available to the consumer.
* `discrete_task::m_threadid`, used to ensure only a single thread runs a given task at any given time.

Using the `volatile` qualifier for `output_buffer::ptr` doesn’t really help, as it doesn’t enforce ordering between writes to the buffer and writes to the pointer itself.  The pointer itself needs to be updated with release semantics, while it needs to be read with acquire semantics when determining how many samples are available to the consumer.

`discrete_task::m_threadid` needs to be updated with acquire semantics when locking and release semantics when unlocking to ensure no work gets reordered outside the critical section.  It’s a bad idea to use `compare_exchange_weak` when locking as it isn’t in a loop attempting to complete a lock-free operation.  On x86 you’ll never notice the difference, but on most other architectures it can fail on contention for a cache line, which isn’t what we want to happen here.

`discrete_task` seems to use `public`/`protected`/`private` randomly, and doesn’t need a vtable as it’s never used as a base class.

This may or may not fix #11373, but I don’t like having dodgy synchronisation code around that works by chance at best.

### DIFF
--- a/src/devices/sound/discrete.cpp
+++ b/src/devices/sound/discrete.cpp
@@ -17,7 +17,7 @@
  *
  ***********************************************************************
  *
- * Each sound primative DSS_xxxx or DST_xxxx has its own implementation
+ * Each sound primitive DSS_xxxx or DST_xxxx has its own implementation
  * file. All discrete sound primatives MUST implement the following
  * API:
  *
@@ -65,7 +65,7 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
  * Values > 500 have a slightly worse performace (too much cache misses?).
  */
 
-#define MAX_SAMPLES_PER_TASK_SLICE  (960/4)
+static constexpr int MAX_SAMPLES_PER_TASK_SLICE = 960 / 4;
 
 /*************************************
  *
@@ -91,61 +91,70 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
 
 struct output_buffer
 {
+	output_buffer() : source(nullptr), ptr(nullptr)
+	{
+	}
+
+	// required for use in vector (std::atomic<T> has deleted copy constructor)
+	output_buffer(output_buffer &&that) : node_buf(std::move(that.node_buf)), source(that.source), ptr(that.ptr.load(std::memory_order_relaxed)), node_num(that.node_num)
+	{
+	}
+
 	std::unique_ptr<double []>  node_buf;
-	const double                *source;
-	volatile double             *ptr;
+	const double *              source;
+	std::atomic<double *>       ptr;
 	int                         node_num;
 };
 
 struct input_buffer
 {
-	volatile const double       *ptr;               /* pointer into linked_outbuf.nodebuf */
-	output_buffer *             linked_outbuf;      /* what output are we connected to ? */
-	double                      buffer;             /* input[] will point here */
+	const double *              ptr;                // pointer into linked_outbuf.nodebuf
+	output_buffer *             linked_outbuf;      // what output are we connected to?
+	double                      buffer;             // input[] will point here
 };
 
 class discrete_task
 {
-	friend class discrete_device;
 public:
-	virtual ~discrete_task() { }
-
-	inline void step_nodes();
-	inline bool lock_threadid(int32_t threadid)
-	{
-		int expected = -1;
-		return m_threadid.compare_exchange_weak(expected, threadid, std::memory_order_release,std::memory_order_relaxed);
-	}
-	inline void unlock() { m_threadid = -1; }
-
-	//const linked_list_entry *list;
-	discrete_device::node_step_list_t        step_list;
-
-	/* list of source nodes */
-	std::vector<input_buffer> source_list;      /* discrete_source_node */
-
-	int task_group = 0;
-
-
-	discrete_task(discrete_device &pdev) : m_device(pdev), m_threadid(-1)
+	discrete_task(discrete_device &pdev) : m_device(pdev), m_threadid(-1), m_samples(0)
 	{
 		// FIXME: the code expects to be able to take pointers to members of elements of this vector before it's filled
 		source_list.reserve(16);
 	}
 
-protected:
-	static void *task_callback(void *param, int threadid);
-	inline bool process();
-
 	void check(discrete_task &dest_task);
 	void prepare_for_queue(int samples);
 
-	std::vector<output_buffer>      m_buffers;
-	discrete_device &                   m_device;
+	static void *task_callback(void *param, int threadid);
+
+	//const linked_list_entry *list;
+	discrete_device::node_step_list_t        step_list;
+
+	int task_group = 0;
 
 private:
-	std::atomic<int32_t>      m_threadid;
-	volatile int            m_samples = 0;
+	void step_nodes();
+	bool process();
+
+	bool lock_threadid(int32_t threadid)
+	{
+		int expected = -1;
+		return m_threadid.compare_exchange_strong(expected, threadid, std::memory_order_acquire, std::memory_order_relaxed);
+	}
+
+	void unlock()
+	{
+		m_threadid.store(-1, std::memory_order_release);
+	}
+
+	/* list of source nodes */
+	std::vector<input_buffer> source_list;      /* discrete_source_node */
+
+	std::vector<output_buffer>  m_buffers;
+	discrete_device &           m_device;
+
+	std::atomic<int32_t>        m_threadid;
+	int                         m_samples;
 };
 
 
@@ -179,17 +188,13 @@ private:
 inline void discrete_task::step_nodes()
 {
 	for (input_buffer &sn : source_list)
-	{
 		sn.buffer = *sn.ptr++;
-	}
 
 	if (EXPECTED(!m_device.profiling()))
 	{
+		// Now step the nodes
 		for (discrete_step_interface *entry : step_list)
-		{
-			/* Now step the node */
 			entry->step();
-		}
 	}
 	else
 	{
@@ -204,19 +209,22 @@ inline void discrete_task::step_nodes()
 		}
 	}
 
-	/* buffer the outputs */
+	// buffer the outputs
 	for (output_buffer &outbuf : m_buffers)
-		*outbuf.ptr++ = *outbuf.source;
+		*outbuf.ptr.load(std::memory_order_relaxed) = *outbuf.source;
+	std::atomic_thread_fence(std::memory_order_release);
+	for (output_buffer &outbuf : m_buffers)
+		outbuf.ptr.store(outbuf.ptr.load(std::memory_order_relaxed) + 1, std::memory_order_relaxed);
 }
 
 void *discrete_task::task_callback(void *param, int threadid)
 {
 	const auto &list = *reinterpret_cast<const discrete_sound_device::task_list_t *>(param);
-	do
+	while (true)
 	{
 		for (const auto &task : list)
 		{
-			/* try to lock */
+			// try to lock
 			if (task->lock_threadid(threadid))
 			{
 				if (!task->process())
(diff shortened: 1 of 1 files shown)


src/devices/sound/discrete.cpp
```diff
@@ -91,61 +91,70 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
...
+	output_buffer(output_buffer &&that) : node_buf(std::move(that.node_buf)), source(that.source), ptr(that.ptr.load(std::memory_order_relaxed)), node_num(that.node_num)
+	{
+	}
+
 	std::unique_ptr<double []>  node_buf;
-	const double                *source;
-	volatile double             *ptr;
+	const double *              source;
+	std::atomic<double *>       ptr;
 	int                         node_num;
```
cuavas, line 106: Using `volatile` here doesn’t do what it’s supposed to, as it doesn’t enforce ordering of writes to the buffer relative to writes to the pointer itself.

Making `ptr` atomic guarantees consistent loads on systems where pointer-sized writes are not inherently atomic (typically systems where the memory bus is narrower than a pointer, e.g. 32-bit CPU with 16-bit memory, or 64-bit CPU with 32-bit memory).  It also makes it easier to ensure the compiler doesn’t turn what appears to be a single access int multiple loads.

src/devices/sound/discrete.cpp
```diff
@@ -91,61 +91,70 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
...
 };
 
 struct input_buffer
 {
-	volatile const double       *ptr;               /* pointer into linked_outbuf.nodebuf */
-	output_buffer *             linked_outbuf;      /* what output are we connected to ? */
-	double                      buffer;             /* input[] will point here */
+	const double *              ptr;                // pointer into linked_outbuf.nodebuf
+	output_buffer *             linked_outbuf;      // what output are we connected to?
+	double                      buffer;             // input[] will point here
```
cuavas, line 113: There’s no point having `volatile` on `ptr` here, either.  You just need to ensure that there’s an acquire barrier between determining the number of samples available and reading the samples.

src/devices/sound/discrete.cpp
```diff
@@ -91,61 +91,70 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
...
 };
 
 class discrete_task
 {
-	friend class discrete_device;
 public:
-	virtual ~discrete_task() { }
```
cuavas, line 111: Using `public`/`protected`/`private` randomly and relying on friends is a bad design.  (There isn’t much point using `protected` at all here when there are no derived classes.)

There’s no point giving this a vtable as it has no other virtual member functions, and isn’t used as a base class for anything.

src/devices/sound/discrete.cpp
```diff
@@ -91,61 +91,70 @@ DEFINE_DEVICE_TYPE(DISCRETE, discrete_sound_device, "discrete", "Discrete Sound"
...
+		return m_threadid.compare_exchange_strong(expected, threadid, std::memory_order_acquire, std::memory_order_relaxed);
 	}
 
-protected:
-	static void *task_callback(void *param, int threadid);
-	inline bool process();
+	void unlock()
+	{
+		m_threadid.store(-1, std::memory_order_release);
+	}
```
cuavas, line 148: Locking requires acquire semantics to ensure no task work can be reordered before it happens.  Unlocking requires release semantics to ensure no task work can be reordered after it happens.

src/devices/sound/discrete.cpp
```diff
@@ -204,74 +209,78 @@ inline void discrete_task::step_nodes()
...
 	}
 
-	/* buffer the outputs */
+	// buffer the outputs
+	for (output_buffer &outbuf : m_buffers)
+		*outbuf.ptr.load(std::memory_order_relaxed) = *outbuf.source;
+	std::atomic_thread_fence(std::memory_order_release);
 	for (output_buffer &outbuf : m_buffers)
-		*outbuf.ptr++ = *outbuf.source;
+		outbuf.ptr.store(outbuf.ptr.load(std::memory_order_relaxed) + 1, std::memory_order_relaxed);
```
cuavas, line 217: This was pretty dodgy – there’s no guarantee of when the postincrement happens relative to the store.

We can use a single release barrier here between storing the samples and updating the pointers to allow some loop unrolling, etc.

src/devices/sound/discrete.cpp
```diff
@@ -204,74 +209,78 @@ inline void discrete_task::step_nodes()
...
 	for (input_buffer &sn : source_list)
 	{
-		int avail = sn.linked_outbuf->ptr - sn.ptr;
+		const int avail = sn.linked_outbuf->ptr.load(std::memory_order_relaxed) - sn.ptr;
 		if (avail < 0)
 			throw emu_fatalerror("discrete_task::process: available samples are negative");
 		if (avail < samples)
 			samples = avail;
 	}
+	std::atomic_thread_fence(std::memory_order_acquire);
```
cuavas, line 253: We only need a single acquire barrier here between determining the minimum number of samples available across all inputs and consuming the samples.

Since `output_buffer::ptr` is atomic, we get a guarantee that there’s a single load here and we won’t get an inconsistent value.

src/devices/sound/discrete.cpp
```diff
@@ -204,74 +209,78 @@ inline void discrete_task::step_nodes()
...
-		ob.ptr = ob.node_buf.get();
+		ob.ptr.store(ob.node_buf.get(), std::memory_order_relaxed);
 
-	/* initialize sources */
+	// initialize sources
 	for (input_buffer &sn : source_list)
-	{
 		sn.ptr = sn.linked_outbuf->node_buf.get();
-	}
 }
```
cuavas, line 284: We can use relaxed memory order here because there’s a barrier after preparing all the tasks.

src/devices/sound/discrete.cpp
```diff
@@ -312,16 +321,16 @@ void discrete_task::check(discrete_task &dest_task)
...
 							}
 
-						if (found<0)
+						if (found < 0)
 						{
 							output_buffer buf;
 
 							buf.node_buf = std::make_unique<double []>((task_node->sample_rate() + sound_manager::STREAMS_UPDATE_FREQUENCY) / sound_manager::STREAMS_UPDATE_FREQUENCY);
-							buf.ptr = buf.node_buf.get();
+							buf.ptr.store(buf.node_buf.get(), std::memory_order_relaxed);
```
cuavas, line 329: This only happens on start, and there’s a barrier after preparing tasks.

src/devices/sound/discrete.cpp
```diff
@@ -1017,22 +1025,18 @@ void discrete_device::process(int samples)
...
-	/* Setup tasks */
+	// Set up tasks
 	for (const auto &task : task_list)
-	{
-		/* unlock the thread */
-		task->unlock();
-
 		task->prepare_for_queue(samples);
-	}
+	std::atomic_thread_fence(std::memory_order_release);
```
cuavas, line 1031: The unlock operation has been moved inside `discrete_task::prepare_for_queue`.  The release barrier here ensures preparation work can’t be reordered after starting the tasks.

cuavas (commented):
Self-reviewing to allow in-place explanatory comments.

seleuco:
I have tested this on an ARM64 build and it seems to work fine. Games with discrete sound now do not hung when numprocessors is not one. Thank you.

joepogo:
Just chiming in here, on Xbox retroarch mame, the current core which is based on current mame, discrete games still crash back to dash and threads have to be turned off. Is there anything else here that could be causing the crash? Was hoping since this fixed issues on the pi5 it would also benefit us as well.

cuavas:
We do not support hacked up RetroArch cores.

StLouisCPhT:
> Just chiming in here, on Xbox retroarch mame, the current core which is based on current mame, discrete games still crash back to dash and threads have to be turned off. Is there anything else here that could be causing the crash? Was hoping since this fixed issues on the pi5 it would also benefit us as well.

Best way to fix the problem would be to build a standalone version of Mame instead of using RA.

joepogo:
Thanks for the reply @StLouisCPhT! Yea, the series x has really good speed on mame retroarch and we do have the option to turn off threading (as the discrete games and some chd games that use threading run just fine on xbox once you disable it). 

I think even if the series x and s had a standalone port, the discrete/chd games would still crash unless threading was disabled, so i figured id ask here if there was anything we could try in case we ever do go that route. 

Anyhow, If not though, oh well, most of the discrete games run just fine it's just when you start getting into certain chd games that use threading like san francisco rush, time crsis, etc that you start to get slowdown. That's where it'd be nice to get the speed benefit like pc does. 

Anyways, figured there was no harm in asking. Thank you again for you response! :)

StLouisCPhT:
> Thanks for the reply @StLouisCPhT! Yea, the series x has really good speed on mame retroarch and we do have the option to turn off threading (as the discrete games and some chd games that use threading run just fine on xbox once you disable it).
> 
> I think even if the series x and s had a standalone port, the discrete/chd games would still crash unless threading was disabled, so i figured id ask here if there was anything we could try in case we ever do go that route.
> 
> Anyhow, If not though, oh well, most of the discrete games run just fine it's just when you start getting into certain chd games that use threading like san francisco rush, time crsis, etc that you start to get slowdown. That's where it'd be nice to get the speed benefit like pc does.
> 
> Anyways, figured there was no harm in asking. Thank you again for you response! :)

After the fiasco with people trying to circumvent MS's retail restrictions and the lack of activity by Sir Mangler in almost a year, I gave up on any real emulation work on the Xbox series S or X and switched to a Shield TV Pro.

@cuavas Does Mame still have any of the old UWP code still in it or was it completely stripped out?

rb6502:
> This was for fixing the issue on ARM based systems iirc. Not Intel/AMD x86-64 based systems like Xbox.

The fix should be correct for all processor architectures, there is nothing ARM-specific about it.  However, Retroarch doesn't use MAME's threading system, and with their intense focus on obsolete hardware they're a little angry we use threading at all.

cuavas:
> @cuavas Does Mame still have any of the old UWP code still in it or was it completely stripped out?

The old UWP code was removed as it had rotted pretty badly and Microsoft moved to allowing a subset of Win32 for Store apps rather than requiring UWP.

PR #12036, opened 2024-02-15
cpu/tms32010.cpp: Fix TMS32016 64k-word program memory access

cam900:


cuavas:
I realise there’s a bunch of stuff in base class that doesn’t depend on the program address width, but trying to work out what does and doesn’t to separate it would be a pain because anything that needs to read immediate operands *does* depend on the program address width.

PR #12037, opened 2024-02-15
New NOT_WORKING machine (Vocalizer 1000)

devinacker:
----------
Breakaway Music Systems Vocalizer 1000 [Loscha, Devin Acker]

Uses a rebadged (or cloned?) ES5503 made by NCR. Currently "not working" due to lack of mic input + several patches that use the sync/AM mode not sounding right. Unfortunately I don't have any hardware recordings at the moment. Otherwise mostly works, though.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -32059,6 +32059,9 @@ nevada                          //
 beezerk                         //
 vlcunk                          //
 
+@source:misc/vocalizer.cpp
+vocalizer
+
 @source:misc/voyager.cpp
 voyager                         // (c) 2002 Team Play, Inc.
 voyagers                        // (c) 2002 Team Play, Inc.
--- a/src/mame/misc/vocalizer.cpp
+++ b/src/mame/misc/vocalizer.cpp
@@ -4,32 +4,32 @@
 
     Breakaway Vocalizer 1000
 
-	Voice-powered synth and MIDI controller. ES5503 "DOC"-based.
+    Voice-powered synth and MIDI controller. ES5503 "DOC"-based.
 
-	To enable MIDI input, press "Option" and then "Melody Guide".
+    To enable MIDI input, press "Option" and then "Melody Guide".
 
-	U11: 512kbit program ROM
-	U12: Sony CXK5864PM-15L (64kbit SRAM)
-	U13: Motorola MC68B09EP
-	U17: NCR E106-71 609-0381069 (ES5503 compatible/clone)
-	U18: 512kbit wave ROM
-	U19: Breakaway GA1 NCR0880995
-	Y1: 8.000 MHz
+    U11: 512kbit program ROM
+    U12: Sony CXK5864PM-15L (64kbit SRAM)
+    U13: Motorola MC68B09EP
+    U17: NCR E106-71 609-0381069 (ES5503 compatible/clone)
+    U18: 512kbit wave ROM
+    U19: Breakaway GA1 NCR0880995
+    Y1: 8.000 MHz
 
-	TODO:
-	- Microphone input. Data is read by function at $9BFD (after FIRQ_MIC)
-	  and then processed by function at $C705.
-	- Instruments that use the sync/AM bit sound weird (need hardware recordings)
-	- Link cable (uses UART 0)
-	- Song cartridges (connected to CPU at $0000-3fff)
-	- Instrument cartridges (connected to DOC at $10000+)
+    TODO:
+    - Microphone input. Data is read by function at $9BFD (after FIRQ_MIC)
+      and then processed by function at $C705.
+    - Instruments that use the sync/AM bit sound weird (need hardware recordings)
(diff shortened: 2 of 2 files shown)


src/mame/misc/vocalizer.cpp
```diff
@@ -0,0 +1,565 @@
...
+	required_device<hd44780_device> m_lcdc;
+	required_device<es5503_device> m_es5503;
+	required_device_array<vocalizer_uart_device, 2> m_uart;
+	required_memory_bank m_bank;
+	required_ioport_array<12> m_keys;
+
+	u8 m_power;
+	u8 m_config;
+	u8 m_input_sel;
+	u8 m_firq_status, m_firq_mask;
```
happppp, line 161: FYI initial variables are uncertain at MAME start if you don't initialize them (here, or at vocalizer_state(x..) or at machine_start.

devinacker: They're initialized in machine_reset, like I've normally been doing in other drivers.

happppp: Ah I missed that. It's probably not the best place when you consider that devices do device_reset before the driver does machine_reset. If a device accesses a devcb at reset that goes back to the driver, it may read an uninitialized var, or write to one that gets overwritten at machine_reset (some devices initialize output ports at reset).

src/mame/mame.lst
```diff
@@ -32059,6 +32059,9 @@ nevada                          //
 beezerk                         //
 vlcunk                          //
 
+@source:misc/vocalizer.cpp
+vocalizer
+
```
cuavas, line 32064: Can you abbreviate the short name to fit it in eight characters to avoid “short name length inflation”?

src/mame/misc/vocalizer.cpp
```diff
@@ -0,0 +1,565 @@
...
+	auto tx_cb() { return m_tx_cb.bind(); }
+	auto tx_irq_cb() { return m_tx_irq_cb.bind(); }
+	auto rx_irq_cb() { return m_rx_irq_cb.bind(); }
+
+	u8 read() { return get_received_char(); }
+	void write(u8 data) { transmit_register_setup(data); }
+
+protected:
+	virtual void device_start() override {};
+	virtual void device_reset() override;
```
cuavas, line 66: Superfluous semicolon.

src/mame/misc/vocalizer.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+	virtual void tra_callback() override { m_tx_cb(transmit_register_get_data_bit()); }
+	virtual void tra_complete() override { m_tx_irq_cb(1); }
+	virtual void rcv_complete() override { receive_register_extract(); m_rx_irq_cb(1); }
```
cuavas, line 70: You’re never deasserting the interrupt lines.  That’s bad practice.

src/mame/mame.lst
happppp: It's fine to keep is as-is, I'm ok with 9 letter shortnames when there's no nice way to abbreviate it. @devinacker your choice.

cuavas: Well, “voclizer” and “vocalizr” both read fine and fit in eight characters.

happppp: I disagree that those 2 examples are sensible to type when starting mame from cmdline.

angelosa: Can we please relax this "8 letter limit at all costs" suggestion, it's frankly silly for stuff like `megadriv`, I don't want to struggle with MAME CLI suggestions more than enough already.

cuavas: The thing is, you still need to memorise the short names for things you use frequently, and I don’t want to have to type excessively long short names at the command line all the time.  It’s already annoying enough with the length of short names for software lists when you need to type a fully qualified list:item:part identifier.

happppp: Any word with 1 letter removed is more annoying to type than the full word.
Well, except for "a" or "I" :P

cuavas: But then people will start complaining about things having a letter or two dropped to fit nine characters, and then ten characters, and so on.

You follow this to its conclusion and we end up with short names no longer being short and getting close to the 16 character limit on everything, and then you end up having trouble fitting suffixes for clones again.  Then there’s pressure to extend the limit again.

Short names are supposed to be short identifiers that can be quickly and easily typed, don’t make for excessively long URLs, etc.

happppp: Is that what you're worried about? It doesn't happen (or if it does, show me where).
When I pick shortnames, my max is 9, except for game&watch with their gnw_ prefix.

devinacker: IMO using a full nine-letter word isn't any less convenient than having to remember to spell it with one fewer vowel or what have you. I'm not really attached to any particular naming, though.

PR #12039, opened 2024-02-16
New NMK214 device: added implementation for NMK214 descrambling GFX device and hook it up to nmk16 driver

sergiopolog:
Added implementation for NMK214 as a new device. It's used for descrambling GFX data on some games in the nmk16 driver.

The implementation was made based on the reverse engineering work of the chip at silicon level: 
[https://github.com/sergiopolog/GateArray-RE/tree/master/NMK/nmk-214](https://github.com/sergiopolog/GateArray-RE/tree/master/NMK/nmk-214)

Results of the descrambling GFX process for both sprites and background tiles using this implementation match exactly with the ones from previous implementation made by @mamehaze (and Antiriad) back then.
In this case, all the internal configurations that could be selected to perform it are included, even if they are never used for any game, just for documentation purposes.

Real hardware performs the descrambling process on the fly for each word/byte fetch from GFX ROMs, but due how the drawgfx system is made on MAME, I couldn't find an option to integrate it and mimic the real hw behaviour.
I simply run the decoding of the entire GFX regions just when both devices are initialized at startup, (similar to what the previous implementation did on init handler).
If there is a way to do this on real GFX fetches from 'sprites' and 'bgtiles' regions, please, let me know.

By now, the descrambling process is only hooked up for sabotenb (and clones) set.
On a later PR, I plan to apply it to the rest of games using it on nmk16 driver and split the current "tdragon_prot_state" class into two different classes: One for NMK111 and NMK113 protection (tdragon and hachamf) and another one for NMK214-215 protection, as they are completely different processes, even if they use the same MCU model (TLCS-90)

### DIFF
--- a/src/mame/nmk/nmk16.cpp
+++ b/src/mame/nmk/nmk16.cpp
@@ -197,7 +197,6 @@ Reference of music tempo:
 #include "nmk16.h"
 
 #include "nmk004.h"
-#include "nmk214.h"
 
 #include "cpu/m68000/m68000.h"
 #include "cpu/pic16c5x/pic16c5x.h"
@@ -4918,7 +4917,7 @@ void tdragon_prot_state::mcu_port3_to_214_w(u8 data)
 		// bit 3 of the incoming value matches to the operation mode configured on each device:
 		m_nmk214[0]->set_init_config(m_init_data_nmk214);
 		m_nmk214[1]->set_init_config(m_init_data_nmk214);
-		
+
 		// Force decode gfx after setting both nmk214 init config, just for testing purposes. Real devices perform the decoding on the fly
 		// for each byte/word fetch from GFX ROMs
 		if (!m_gfx_decoded && m_nmk214[0]->is_device_initialized() && m_nmk214[1]->is_device_initialized())
@@ -4959,11 +4958,11 @@ void tdragon_prot_state::saboten_prot(machine_config &config)
 	// the 215 has these hooked up, going to the 214
 	m_protcpu->port_write<3>().set(FUNC(tdragon_prot_state::mcu_port3_to_214_w));
 	m_protcpu->port_write<7>().set(FUNC(tdragon_prot_state::mcu_port7_to_214_w));
-	
+
 	NMK214(config, m_nmk214[0], 0); // Descrambling device for sprite GFX data
 	m_nmk214[0]->set_mode(0);
 	m_nmk214[0]->set_input_address_bitswap(nmk214_sprites_address_bitswap);
-	
+
 	NMK214(config, m_nmk214[1], 0); // Descrambling device for BG GFX data
 	m_nmk214[1]->set_mode(1);
 	m_nmk214[1]->set_input_address_bitswap(nmk214_bg_address_bitswap);
@@ -5173,15 +5172,15 @@ void tdragon_prot_state::decode_sabotenb()
 		rom[A] = m_nmk214[1]->decode_byte(A, rom[A]);
 	}
 
-	// Sprites
+	// sprites
 	rom = memregion("sprites")->base();
 	len = memregion("sprites")->bytes();
-	for (int A = 0; A < len; A += 2)
+	for (int A = 0; A < (len - 1); A += 2)
 	{
-		// When the sprite ROM is loaded, it's also byte-swapped, so we keep the same order here to compound the word (A: second-byte, A+1: first-byte) (BE)
+		// sprite ROM is 16-bit big Endian
 		u16 word = get_u16be(&rom[A]);
 
-		// As Sprites ROM works in word mode, address value here is in terms of bytes and should be divided by 2 to get the proper base-16 address value:
+		// A is a byte address, divide by 2 to give word address for NMK214
 		put_u16be(&rom[A], m_nmk214[0]->decode_word(A/2, word));
 	}
 }
--- a/src/mame/nmk/nmk16.cpp
+++ b/src/mame/nmk/nmk16.cpp
@@ -4359,15 +4359,41 @@ u8 tdragon_prot_state::mcu_port6_r()
 
 void tdragon_prot_state::machine_start()
 {
+	nmk16_state::machine_start();
+
 	save_item(NAME(m_bus_status));
-	save_item(NAME(m_init_data_nmk214));
-	save_item(NAME(m_init_clock_nmk214));
-	save_item(NAME(m_gfx_decoded));
 }
 
 void tdragon_prot_state::machine_reset()
 {
+	nmk16_state::machine_reset();
+
 	m_bus_status = 0x04;
+}
+
+void tdragon_prot_214_state::device_post_load()
+{
+	tdragon_prot_state::device_post_load();
+
+	if (m_gfx_unscramble_enabled && !m_gfx_decoded)
+	{
+		// Loaded a state saved after graphics were decoded before decoding graphics
+		// Force graphics unscrambling now
+		decode_nmk214();
+		m_gfxdecode->gfx(1)->mark_all_dirty(); // background tiles
+		m_gfxdecode->gfx(2)->mark_all_dirty(); // sprites
+		m_bg_tilemap[0]->mark_all_dirty();
+		m_gfx_decoded = true;
+	}
+}
+
+void tdragon_prot_214_state::machine_start()
+{
+	tdragon_prot_state::machine_start();
+
+	save_item(NAME(m_init_data_nmk214));
+	save_item(NAME(m_init_clock_nmk214));
+	save_item(NAME(m_gfx_unscramble_enabled));
 }
 
 void tdragon_prot_state::mcu_side_shared_w(offs_t offset, u8 data)
@@ -4903,9 +4929,8 @@ void nmk16_state::bjtwin(machine_config &config)
 	nmk112.set_rom1_tag("oki2");
 }
 
-// The 215 writes minimal data here. Data is, in fact, used for selecting one of the internal configurations hardwired inside NMK214 device
-// to make the decryption logic:
-void tdragon_prot_state::mcu_port3_to_214_w(u8 data)
+// The 215 writes minimal data here to select an unscrambling configuration hardwired inside the NMK214 chip.
+void tdragon_prot_214_state::mcu_port3_to_214_w(u8 data)
 {
 	LOG("%s: mcu_port3_to_214_w (data %02x)\n", machine().describe_context(), data);
 
@@ -4920,18 +4945,24 @@ void tdragon_prot_state::mcu_port3_to_214_w(u8 data)
 
 		// Force decode gfx after setting both nmk214 init config, just for testing purposes. Real devices perform the decoding on the fly
 		// for each byte/word fetch from GFX ROMs
-		if (!m_gfx_decoded && m_nmk214[0]->is_device_initialized() && m_nmk214[1]->is_device_initialized())
+		if (!m_gfx_unscramble_enabled && m_nmk214[0]->is_device_initialized() && m_nmk214[1]->is_device_initialized())
 		{
-			decode_sabotenb();
-			m_bg_tilemap[0]->mark_all_dirty();
-			m_gfx_decoded = true;
+			m_gfx_unscramble_enabled = true;
+			if (!m_gfx_decoded)
+			{
+				decode_nmk214();
+				m_gfxdecode->gfx(1)->mark_all_dirty(); // background tiles
+				m_gfxdecode->gfx(2)->mark_all_dirty(); // sprites
+				m_bg_tilemap[0]->mark_all_dirty();
+				m_gfx_decoded = true;
+			}
 		}
 	}
 
 	m_init_clock_nmk214 = BIT(data, 2);
 }
 
-void tdragon_prot_state::mcu_port7_to_214_w(u8 data)
+void tdragon_prot_214_state::mcu_port7_to_214_w(u8 data)
 {
 	LOG("%s: mcu_port7_to_214_w (data %02x)\n", machine().describe_context(), data);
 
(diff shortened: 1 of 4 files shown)


src/mame/nmk/nmk214.h
```diff
@@ -0,0 +1,38 @@
...
+	virtual void device_reset() override;
+
+private:
+	u8 m_mode;                              // Operation mode. In practice, only LSB is used. Allowed values: 0 or 1. This value is hardwired for each NM214 out of 2 ones in the game board, each device having a different configuration than the other.
+	u8 m_init_config;                       // An init config out of 8 existing ones in NMK214 to be used while descrambling GFX. In practice, only lower 3 bits (From 0 to 2) are used, allowed values: 0 to 7. Bit 3 is used to match with operation mode and let the init config to be stored on the device or not
+	bool m_device_initialized;			    // Flag to mark the device already received the init configuration and it can perform descrambling from now on
+
+	const u8 *m_input_address_bitswap;      // Input address lines bitswap. These represents the way the 13 address lines on NMK214 (from A0 to A12) are hooked up externally related to the GFX ROMs address bus, that could be in different order/position
+
+	u8 get_bitswap_select_value(u32 addr);  // Gets the a value to select the output bitswap to apply to the input data, based on the address where the data is located
```
cuavas, line 33: These lines are excessively long.  Please wrap long comments.

src/mame/nmk/nmk214.cpp
```diff
@@ -0,0 +1,191 @@
...
+	It works in tandem with NMK215, that is a Toshiba MCU which sends initialization data to NMK214 to perform the descrambling process.
+	Every game pcb using it has two units of NMK214, one for descrambling sprite data and another one for descrambling background tiles data.
+	It can work in two different modes: word and byte. For sprites it always works in word mode. On the order side, for backgrounds it always works in byte mode.
+	There are 8 different internal configs hardwired inside the device, and the data received from NMK215 selects one of those at startup. That init data is stored in the device when bit 3 matches with the operation mode wired directly on the pcb.
+	The descrambling process is essentially a dynamic bitswap of the incoming word/byte data, doing a different bitswap based on the address of the data to be descrambled.
+	The input address bus on the device is used to determine which bitswap do for each word/byte, and it's usually hooked up in different way for sprites and bg tiles, so an 'input_address_bitswap' is included in the implementation in order to get the effective address the device will use internally for each case.
```
cuavas, line 12: Please wrap these lines at a reasonable width.

src/mame/nmk/nmk214.h
```diff
@@ -0,0 +1,38 @@
...
+protected:
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+private:
+	u8 m_mode;                              // Operation mode. In practice, only LSB is used. Allowed values: 0 or 1. This value is hardwired for each NM214 out of 2 ones in the game board, each device having a different configuration than the other.
+	u8 m_init_config;                       // An init config out of 8 existing ones in NMK214 to be used while descrambling GFX. In practice, only lower 3 bits (From 0 to 2) are used, allowed values: 0 to 7. Bit 3 is used to match with operation mode and let the init config to be stored on the device or not
+	bool m_device_initialized;			    // Flag to mark the device already received the init configuration and it can perform descrambling from now on
+
+	const u8 *m_input_address_bitswap;      // Input address lines bitswap. These represents the way the 13 address lines on NMK214 (from A0 to A12) are hooked up externally related to the GFX ROMs address bus, that could be in different order/position
```
cuavas, line 31: It’s somewhat dangerous using a pointer for this – you’ll get no array bounds checking at compile time, etc.  Can you use `std::array<u8, 13>` or C array references or something to make it safer?

src/mame/nmk/nmk214.h
```diff
@@ -0,0 +1,38 @@
...
+
+class nmk214_device : public device_t
+{
+public:
+	nmk214_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+	void set_mode(u8 mode) { m_mode = mode; }
+	void set_init_config(u8 init_config);
+	bool is_device_initialized() { return m_device_initialized; }
+	void set_input_address_bitswap(const u8 *input_address_bitswap) { m_input_address_bitswap = input_address_bitswap; }
```
cuavas, line 17: Please visually separate the setters for hard-coded things (the hard-wired mode and the address scrambling) from things that can change dynamically.

src/mame/nmk/nmk214.cpp
```diff
@@ -0,0 +1,191 @@
...
+		m_init_config = init_config & 0x7;
+		m_device_initialized = true;
+	}
+}
+
+void nmk214_device::device_start()
+{	
+	save_item(NAME(m_mode));
+	save_item(NAME(m_init_config));
+	save_item(NAME(m_device_initialized));
```
cuavas, line 184: If the mode is hard-coded, it doesn’t need to be saved.

src/mame/nmk/nmk16.cpp
```diff
@@ -4899,19 +4902,38 @@ void nmk16_state::bjtwin(machine_config &config)
...
+	// Startup only. Bit 2 on Port 3 of MCU (NMK215) acts as strobe/clock (rising edge) for storing the byte previously set on the Port 7 of MCU (NMK215) into the internal registers of both NMK214 devices
+	if (m_init_clock_nmk214 == 0 && BIT(data, 2) == 1)
+	{
+		// Value is sent to both devices at the same time. Internally, each one evaluates if the value should be stored or not based on bit 3 of the incoming value matches to the operation mode configured on each device:
+		m_nmk214[0]->set_init_config(m_init_data_nmk214);
+		m_nmk214[1]->set_init_config(m_init_data_nmk214);
+		
+		// TEMP: force decode gfx after setting both nmk214 init config, just for testing purposes. Real devices perform the decoding on the fly for each byte/word fetch from GFX ROMs
+		if (m_nmk214[0]->is_device_initialized() && m_nmk214[1]->is_device_initialized())
+			decode_sabotenb();
```
cuavas, line 4919: Please wrap the long comments.

src/mame/nmk/nmk16.cpp
```diff
@@ -4899,19 +4902,38 @@ void nmk16_state::bjtwin(machine_config &config)
...
 {
-	// startup only
-	logerror("%s: mcu_port7_to_214_w (data %02x)\n", machine().describe_context(), data);
+	LOG("%s: mcu_port7_to_214_w (data %02x)\n", machine().describe_context(), data);
+
+	// Startup only. Value written here is kept as outputs on Port 7 of MCU (NMK215) in order to be read by the NMK214 devices when the clock signal is sent (Bit 2 of Port 3)
+	m_init_data_nmk214 = data;
 }
 
+// Bitswaps that represent how the address bus of the NMK214s (13 bits) are hooked up related to the address bus of the ROMs. Every element represents the bit number in the ROM address bus that should be taken in each NMK214 address bus position, starting by the LSB
```
cuavas, line 4933: These comments should be wrapped, too.

src/mame/nmk/nmk16.cpp
```diff
@@ -5118,6 +5148,34 @@ void nmk16_state::decode_gfx()
...
+	len = memregion("sprites")->bytes();
+	for (int A = 0; A < len; A += 2)
+	{
+		// When the sprite ROM is loaded, it's also byte-swapped, so we keep the same order here to compound the word (A: second-byte, A+1: first-byte)
+		u16 word = (rom[A] << 8) | rom[A+1];
+		// As Sprites ROM works in word mode, address value here is in terms of bytes and should be divided by 2 to get the proper base-16 address value
+		u16 decoded_word = m_nmk214[0]->decode_word(A/2, word);
+		
+		rom[A] = decoded_word >> 8;
+		rom[A+1] = decoded_word & 0xff;
```
cuavas, line 5175: You can use `get_u16be` and `put_u16be` from `"multibyte.h"` to help with this.

src/mame/nmk/nmk16.cpp
```diff
@@ -4899,19 +4902,38 @@ void nmk16_state::bjtwin(machine_config &config)
...
+		m_nmk214[0]->set_init_config(m_init_data_nmk214);
+		m_nmk214[1]->set_init_config(m_init_data_nmk214);
+		
+		// TEMP: force decode gfx after setting both nmk214 init config, just for testing purposes. Real devices perform the decoding on the fly for each byte/word fetch from GFX ROMs
+		if (m_nmk214[0]->is_device_initialized() && m_nmk214[1]->is_device_initialized())
+			decode_sabotenb();
+	}
+
+	m_init_clock_nmk214 = BIT(data, 2);
 }
```
cuavas, line 4923: Do you get a problem if the user soft resets the system (pressing F3 by default)?  Will it decode the ROMs a second time?

src/mame/nmk/nmk16.cpp
```diff
@@ -5118,6 +5148,34 @@ void nmk16_state::decode_gfx()
...
+	{
+		// When the sprite ROM is loaded, it's also byte-swapped, so we keep the same order here to compound the word (A: second-byte, A+1: first-byte)
+		u16 word = (rom[A] << 8) | rom[A+1];
+		// As Sprites ROM works in word mode, address value here is in terms of bytes and should be divided by 2 to get the proper base-16 address value
+		u16 decoded_word = m_nmk214[0]->decode_word(A/2, word);
+		
+		rom[A] = decoded_word >> 8;
+		rom[A+1] = decoded_word & 0xff;
+	}
+}
```
cuavas, line 5177: Is there a potential for issues with caching in the `gfxdecode_device`?  I’d expect you’d need to use `mark_dirty` in a loop to ensure tiles are refreshed.

cuavas (commented):
If you really want decoding to happen on-the-fly, you can’t use `gfxdecode_device` – its purpose is to cache decoded tiles, so it works by reading tile graphics en masse and storing the bitmaps in an easy-to-use form.

The code looks like it would have issues with soft reset (F3 by default).  If the user resets the system, the already decoded graphics ROM content will be decoded again.  This may be masked by the caching performed by the `gfxdecode_device`, as tiles decoded before the system was reset won’t be decoded again.

Also, tiles aren’t invalidated after the NMK214 is configured, so any tiles that are decoded before that (e.g. if the user manages to pause the system immediately on start and bring up the tile viewer before the game initialises the NMK214) won’t be refreshed when.

src/mame/nmk/nmk214.cpp
Gurudumps: Just an FYI, I use a wrap width of 140 characters in my Guru-Readmes(tm)(c)(r). This allows everything to fit on a laptop screen that was common years ago and used on current low-end laptops (i.e. very common WXGA 1280x800)

src/mame/nmk/nmk16.cpp
sergiopolog: Now, I included tilemap function `mark_all_dirty()` for invalidate cached tiles, not sure if that's enough to refresh all the tiles or I need to do it in a loop as you mentioned. I didn't see any errors while testing.

Additionally, is this also needed for sprites?

src/mame/nmk/nmk16.cpp
sergiopolog: You are right. GFX data is decoded again after a soft reset. I included an additional flag to control that the decoding process is already done and avoiding doing it again.
Anyway, this is a side-effect of decoding the entire-region when devices are initialized. If the decoding could be hooked up for each GFX data fetch (byte/word), this won't happen.

src/mame/nmk/nmk16.cpp
sergiopolog: Done, thanks!

src/mame/nmk/nmk16.cpp
sergiopolog: Fixed

src/mame/nmk/nmk16.cpp
sergiopolog: Fixed

src/mame/nmk/nmk214.cpp
sergiopolog: Removed

src/mame/nmk/nmk214.h
sergiopolog: Done

src/mame/nmk/nmk214.h
sergiopolog: Now I used `std::array` for this

src/mame/nmk/nmk214.cpp
sergiopolog: Fixed

src/mame/nmk/nmk214.h
sergiopolog: Fixed

sergiopolog:
> If you really want decoding to happen on-the-fly, you can’t use `gfxdecode_device` – its purpose is to cache decoded tiles, so it works by reading tile graphics en masse and storing the bitmaps in an easy-to-use form.

Got it. Is there any other (feasible) alternative to 'gfxdecode_device' to use in decoding gfx? 

Apart from that, all other fixes/suggestions are included now in the PR

sergiopolog:
@cuavas Is any further action needed on this PR from my side?

cuavas:
I tidied up the code and pushed to your branch.

It still isn’t going to play nicely with save states because the content of the memory regions won’t stay in sync with the state of the decoder (memory regions aren’t added to save states because they’re intended to be used with ROM which is static).

The case were I’d anticipate this causing trouble is with `-state` where a save state is loaded immediately on start.  At this point, the system probably won’t have initialised the NMK214 chips, but the save state will likely have been taken after those chips have been initialised, so it will run with scrambled graphics.

Solving it properly, and allowing on-the-fly changes to the unscrambling, really requires `device_gfx_interface` to be slimmed down with the actual decoding moved to a derived class.

I can think of a couple of relatively easy ways to work around it:
* The brute force approach of adding the graphics ROM region contents to save states for systems using the NMK214 chip.
* Add an additional member not registered for save states to track whether graphics have been actually been decoded, and implement hackery in `device_post_load` to force an immediate decode if a saved state is loaded before graphics are decoded, or to suppress decoding graphics when going the other way.

cuavas:
I patched it up using the second approach, and added the code to force tiles to be decoded after unscrambling.  This will do until `device_gfx_interface` is refactored some more.

sergiopolog:
@cuavas Thanks. I will prepare another PR to include this descrambling process to all the games using it in the nmk16 driver.

`tdragon_prot_214_state` could be decoupled from `tdragon_prot_state`, as latter is only used for `tdragon` and `hachamf`, but not for the rest of games, so I will do it, too.

PR #12043, opened 2024-02-18
trs/coco12.cpp: Update deluxecoco driver with newly found ROMs and hardware information

tlindner:
Update deluxecoco driver with newly found ROMs and hardware information.


cuavas:
Where did the ROMs come from?  Who dumped them?

PR #12044, opened 2024-02-19
rm/rm380z.cpp: Add support for high resolution graphics, colour and text character dimming

RobinSergeant:
This PR adds supports for the HRG board described here:

https://vt100.net/rm/docs/380z-hrg.pdf

It was needed to run most educational software so would be nice to get this in.  As I had to change the palette I've also added the missing dimmed text attribute, and a config port to choose between a colour or monochrome monitor.

Support for both the High Resolution mode (320x192, 4 colour) and Medium Resolution (160x96, 16 colour) is included.  The two modes share video ram, and the medium mode includes two separate display pages (one page is stored at odd addresses, and the other at even addresses).  Each logical colour index maps to one of a possible 256 different colours, which was quite impressive for the time - much more colourful than many 8-bit home computers!

The HRG occupies the first 20 lines of the display, with the lower 4 lines reserved for text.  Although it is also possible to draw text over the graphics area (when the VDU is not set to graphics mode).

It is best demonstrated by using RM Basic where you can now easily draw colourful graphics like this:

image


RobinSergeant:
Not strictly related to this change, but I've just found and pushed a quick fix for the blinking cursor bug. Making the cursor blink (by pressing Ctrl+U) stopped the keyboard from responding. Only clearing the latch when the reset bit has actually been toggled fixes this, and now it can blink and be stopped with Ctrl+W.

RobinSergeant:
I came across a later version of RML Extended BASIC today (version 5.0L) and this failed to recognise the HRG card so I've made another fix for that.  Version 5.0H didn't bother to check and would just hang if the card was not installed, but the newer version monitors HRG port 0 and times out with an error message if it does not detect a change in value.  It should change when line blanking begins or ends, but the timing of this at our end is suspect.  I've extended the line blanking period to ~18% of the the scan line as Wikipedia quotes that %, but I still don't really trust the timing.  It fixes the problem as the transition is now happening frequently enough to satisfy the HRG card detection code however.

RobinSergeant:
Just discovered that on startup RML Extended BASIC v 5.0L writes character data to the HRG memory, and without this the "STPLOT" function does not work.  This is what caused the out of bounds writes I noticed before.  The HRG reference manual does not mention extra RAM for this, but it predates the "HRG Level 2" feature set (which includes "STPLOT").  The previous version of BASIC I was using does not include the level 2 features.

So another mystery solved :-)  I have increased the RAM size to account for this and can now plot text on the HRG display.  I don't actually think it uses all this memory, but it writes to pages 12 through 15 to store the data.  Without knowing exactly how the address space is mapped to a character table just allowing enough RAM for the full range of addresses seems like the best strategy.  The characters are obviously plotted in software because I've not had to change my display routines at all, but I guess it uses video RAM for the character data to free up normal RAM.  They can also be redefined using the "DEFCHAR" routine.  See the BASIC manual for details:

https://vt100.net/rm/docs/pn11006.pdf

This PR is ready for review and hopefully I won't find any other hidden features now!!!

src/mame/rm/rm380z.h
```diff
@@ -13,6 +13,7 @@ Research Machines RM 380Z
 #pragma once
 
 #include "cpu/z80/z80.h"
+#include "emupal.h"
 #include "imagedev/cassette.h"
```
cuavas, line 17: Headers are grouped by library – headers from src/emu go after headers from src/devices (most dependencies to least dependencies, to make it less likely to end up with headers that need to be included in a specific magic order).

Do it like:
```c++
#include "cpu/z80/z80.h"
#include "imagedev/cassette.h"
#include "imagedev/floppy.h"
#include "machine/keyboard.h"
#include "machine/ram.h"
#include "machine/wd_fdc.h"

#include "emupal.h"
```

src/mame/rm/rm380z.h
```diff
@@ -89,16 +95,23 @@ class rm380z_state : public driver_device
...
 	bool get_rowcol_from_offset(int &row, int &col, offs_t offset) const;
 	void put_point(int charnum, int x, int y, int col);
 	void init_graphic_chars();
 
-	void putChar_vdu80(int charnum, int attribs, int x, int y, bitmap_ind16 &bitmap);
-	void putChar_vdu40(int charnum, int x, int y, bitmap_ind16 &bitmap);
-	void decode_videoram_char(int row, int col, uint8_t &chr, uint8_t &attrib);
+	void putChar_vdu80(int charnum, int attribs, int x, int y, bitmap_ind16 &bitmap) const;
+	void putChar_vdu40(int charnum, int x, int y, bitmap_ind16 &bitmap) const;
+	void decode_videoram_char(int row, int col, uint8_t &chr, uint8_t &attrib) const;
```
cuavas, line 114: Marking member functions as `const` when they actually change state is generally frowned upon, even when you can technically do it because everything is modified via pointers.  It helps the *developer* as well as the compiler if you think about semantics.

src/mame/rm/rm380z.h
```diff
@@ -79,6 +83,8 @@ class rm380z_state : public driver_device
 		uint8_t m_scroll_reg = 0;
 	};
 
+	enum hrg_display_mode { RM380Z_HRG_MODE_NONE, RM380Z_HRG_MODE_HIGH, RM380Z_HRG_MODE_MEDIUM_0, RM380Z_HRG_MODE_MEDIUM_1 };
```
cuavas, line 86: Eventually this system should be updated to support saved states.  In preparation for that, please give all enums used for state explicit storage classes.  Also consider using a scoped enum (`enum class`) if it’s a closed set of values.

src/mame/rm/rm380z.h
```diff
@@ -121,8 +134,13 @@ class rm380z_state : public driver_device
 	DECLARE_MACHINE_RESET(rm480z);
 
 	void config_memory_map();
-	void update_screen_vdu80(bitmap_ind16 &bitmap);
-	void update_screen_vdu40(bitmap_ind16 &bitmap);
+	void palette_init(palette_device &palette) const;
+	void change_hrg_scratchpad(int index, uint8_t value, uint8_t mask);
+	void change_palette(int index, uint8_t value) const;
```
cuavas, line 139: These are also conceptually not the sort of things that should be `const`.

src/mame/rm/rm380z_m.cpp
```diff
@@ -100,6 +134,13 @@ uint8_t rm380z_state::port_read(offs_t offset)
 
 	switch (offset)
 	{
+	case 0x00:
+		// bit 0 is low during HRG frame blanking
+		// bit 1 is low duing HRG line blanking
+		// (this is the inverse of VDU port 1 shifted 6 bits to the right)
+		data = ((m_port1 & 0xf0) >> 6) ^ 0x03;
+		break;
```
cuavas, line 142: Why are you masking to isolate four bits, then shifting two of them off the end?  It’s less error-prone to mask *after* shifting, or to use the `BIT` helpers, e.g. this could be expressed as `data = BIT(~m_port1, 6, 2);` (select two bits, starting at bit 6, counting from LSB).

That aside, this isn’t going to work properly – you aren’t setting the horizontal blanking flag anywhere.  You need to configure the screen device with proper timings (preferably using `set_raw` so everything is based on the pixel clock), and then use the `hblank()` and `vblank()` member functions of `screen_device` to get the current state.  Otherwise you’re liable to get out of sync with the screen device.  The whole `static_vblank_timer` is a dreadful hack to avoid actually configuring the screen device properly and it needs to go away.

cuavas (commented):
I’m starting to get concerned that this is starting to reinforce bad/outdated patterns in this driver.  This is a case of a driver that was never in a good state to begin with being ported to C++ by the path of least resistance, and hasn’t had a great deal of attention paid to it since besides ensuring it still compiles.

* It’s better to keep just common things in a common base class, and create derived classes adding things specific to certain machines.  That way you can avoid optional object finders for the most part, and it’s a lot easier to understand what’s used where.
* Driver init functions should really only be used for stuff like decrypting or unscrambling ROMs, and other stuff that can’t practically be done another way.  Most hardware differences should be handled in machine configuration, and you should override `machine_start` etc. in your derived classes to set stuff up.
* The screen is configured with impossible parameters.  You should really be setting the screen timing up using `set_raw`, deriving everything from the master clock.
* And that brings us to `static_vblank_timer`, which is a workaround for not configuring the screen device correctly, because you can’t just ask the screen device for current vblank/hblank state if you configure it to have zero blanking area.  It needs to go away.

Getting back to the actual functional changes, is the HRG board an optional add-on?  If it is an add-on, is it conceptually a modification to the system, or a plug-in card?
* If it’s a modification to the system, you should create clone systems with the HRG board added, so the user can choose a system with or without it.  (For an example, see src/mame/osborne1.cpp where there are clones for systems with the SCREEN-PAC and Nuevo Video high-resolution video board modifications.)
* If it’s a plug-in card, it should be implemented as a slot device, so the user can choose to run the system with or without it plugged in.

src/mame/rm/rm380z.h
RobinSergeant: Yes, I agree about these so have changed the palette altering methods to be non const.

src/mame/rm/rm380z_m.cpp
RobinSergeant: I don't need to mask at all so have changed this.  All I meant to do was take the two blanking bits from the main display and convert them to the HRG format (different positions and inverse values).

It does work in so far as they change often enough to allow the software to write to the screen.  The horizontal blanking flag is being set by rm380z_state::static_vblank_timer, but as you say it's a dreadful hack.  Nonetheless it allows software to recognise the card and update the display.  Software typically just waits for the blanking period as the CPU can only access video ram when it's not being used by the display (on real hardware).  For very quick updates it can use the horizontal blanking period, but for most things it waits for the vertical blank.  The timing does effect how fast software runs as it might be spending more or less time waiting than it should of course.  Things seem to run OK on the emulator, but I've never used real hardware so cannot say if it's running noticeably faster or slower than it should!

src/mame/rm/rm380z.h
RobinSergeant: I wasn't sure what you mean by explicit storage classes, but I've made it a scoped enum with explicit values.  Is it correct now?

src/mame/rm/rm380z.h
RobinSergeant: These particular functions don't change state in my mind.  For example putChar_vdu80 draws a character onto a bitmap that is passed in by reference.  Nothing in the object itself changes, and the bitmap is part of the display rather then the machine itself.  Or am I thinking about it the wrong way?  I don't really mind if they are const, but in my day job I would naturally make them const so it's hard to change mindset.

The palette functions that you highlight in another comment are different.  I can see that these are conceptually not const as the palette is stored inside the machine.

src/mame/rm/rm380z.h
RobinSergeant: Thanks, I've done this now.

RobinSergeant:
> I’m starting to get concerned that this is starting to reinforce bad/outdated patterns in this driver. This is a case of a driver that was never in a good state to begin with being ported to C++ by the path of least resistance, and hasn’t had a great deal of attention paid to it since besides ensuring it still compiles.
> 
> * It’s better to keep just common things in a common base class, and create derived classes adding things specific to certain machines.  That way you can avoid optional object finders for the most part, and it’s a lot easier to understand what’s used where.
> * Driver init functions should really only be used for stuff like decrypting or unscrambling ROMs, and other stuff that can’t practically be done another way.  Most hardware differences should be handled in machine configuration, and you should override `machine_start` etc. in your derived classes to set stuff up.
> * The screen is configured with impossible parameters.  You should really be setting the screen timing up using `set_raw`, deriving everything from the master clock.
> * And that brings us to `static_vblank_timer`, which is a workaround for not configuring the screen device correctly, because you can’t just ask the screen device for current vblank/hblank state if you configure it to have zero blanking area.  It needs to go away.
> 

I agree in general and came to similar conclusions about the lack of inheritance.  I will look into removing the timer hack, and  can also refactor the code to make it more OO.  Would you like this to be done now, or could it wait?  I think the PR works as it stands and would make the driver better (allowing more software to run on the machine), but I can see that it needs more work.

> Getting back to the actual functional changes, is the HRG board an optional add-on? If it is an add-on, is it conceptually a modification to the system, or a plug-in card?
> 
> * If it’s a modification to the system, you should create clone systems with the HRG board added, so the user can choose a system with or without it.  (For an example, see src/mame/osborne1.cpp where there are clones for systems with the SCREEN-PAC and Nuevo Video high-resolution video board modifications.)
> * If it’s a plug-in card, it should be implemented as a slot device, so the user can choose to run the system with or without it plugged in.

The HRG board was a configurable option when buying the machine, e.g. if you only wanted to use it for word processing then you could buy a cheaper machine without the HRG card.  Things are were clearing designed with the board in mind.  For example the 4 line scrolling mode is called graphics mode and is there to display text below the HRG.  Does this sound like something that needs to be made a slot device?  Are they any good examples of this to look at?

src/mame/rm/rm380z.h
rb6502: He means that you can specify an actual base type for an enum, like:

enum foo: uint32_t
{
    blah=0x12345678
}

RobinSergeant: ah, thank you @rb6502.  I didn't even know you could do that.  I always think of enum as being an int, but good to know.  I have updated it again and am just about to push the commit.

cuavas:
> I agree in general and came to similar conclusions about the lack of inheritance. I will look into removing the timer hack, and can also refactor the code to make it more OO. Would you like this to be done now, or could it wait? I think the PR works as it stands and would make the driver better (allowing more software to run on the machine), but I can see that it needs more work.

I’d really rather the unrealistic screen timings and `static_vblank_timer` could go away as soon as possible.  Are there schematics of the machine showing the video counters, etc.?  It’s usually pretty easy to work out the parameters from that.

> The HRG board was a configurable option when buying the machine, e.g. if you only wanted to use it for word processing then you could buy a cheaper machine without the HRG card. Things were clearly designed with the board in mind. For example the 4 line scrolling mode is called graphics mode and is there to display text below the HRG. Does this sound like something that needs to be made a slot device? Are they any good examples of this to look at?

That doesn’t sound so much like a case for a slot device.  Sounds like it’s more a case for having separate machines with and without the HRG card like the three variants of the Osborne 1 (standard, with SCREEN-PAC graphics, and with Nuevo Video graphics).

RobinSergeant:
> > I agree in general and came to similar conclusions about the lack of inheritance. I will look into removing the timer hack, and can also refactor the code to make it more OO. Would you like this to be done now, or could it wait? I think the PR works as it stands and would make the driver better (allowing more software to run on the machine), but I can see that it needs more work.
> 
> I’d really rather the unrealistic screen timings and `static_vblank_timer` could go away as soon as possible. Are there schematics of the machine showing the video counters, etc.? It’s usually pretty easy to work out the parameters from that.
> 

The service manual has schematics and can be found here:

https://vt100.net/rm/docs/pn13821a.pdf

Are you able to get anything from that?  I only have a software background so cannot read schematics!  It shows both the VDU-40 and VDU-80 boards.  The VDU-80 description also talks about working with the HRG board.

The firmware guide say frame blanking occurs for approximately 4.5 milliseconds every 20 milliseconds, and that during line blanking there is only time to output one character.  That is the only clue to timing that I could find.  It can be found here:

https://vt100.net/rm/docs/pn10971.pdf

> > The HRG board was a configurable option when buying the machine, e.g. if you only wanted to use it for word processing then you could buy a cheaper machine without the HRG card. Things were clearly designed with the board in mind. For example the 4 line scrolling mode is called graphics mode and is there to display text below the HRG. Does this sound like something that needs to be made a slot device? Are they any good examples of this to look at?
> 
> That doesn’t sound so much like a case for a slot device. Sounds like it’s more a case for having separate machines with and without the HRG card like the three variants of the Osborne 1 (standard, with SCREEN-PAC graphics, and with Nuevo Video graphics).

Great, I will make it a separate machine then and try to sort out the class hierarchy while I'm at it.

galibert:
For the vdu-40 the timings are given by a pair of seemingly undumped prom, the line prom at ic31, coloured green, and the frame prom at ic36 colored yellow.  Vdu-80 seems to be.. the same actually.  Which means we won't have correct timing until one can ferret out a dump of those proms, which seem to exist but not publically.

rb6502:
Yeah, the VDU-80 seems to be the same relative timings, just a doubled pixel clock.

RobinSergeant:
> For the vdu-40 the timings are given by a pair of seemingly undumped prom, the line prom at ic31, coloured green, and the frame prom at ic36 colored yellow. Vdu-80 seems to be.. the same actually. Which means we won't have correct timing until one can ferret out a dump of those proms, which seem to exist but not publically.

Thanks guys.  I have posted a request on the RM user group just in case anybody has dumps of these.  Bit of a long shot, but worth asking...

simzy39:
How do I contact you to send you dumps or for people with the firmare to contact you?

RobinSergeant:
> How do I contact you to send you dumps or for people with the firmare to contact you? I'm talking to a person whom has everything.

Thank you, you could email robin.sergeant at Gmail.

src/mame/rm/rm380z.h
```diff
@@ -38,27 +39,121 @@ class rm380z_state : public driver_device
 		m_chargen(*this, "chargen"),
 		m_maincpu(*this, RM380Z_MAINCPU_TAG),
 		m_screen(*this, "screen"),
-		m_cassette(*this, "cassette"),
 		m_messram(*this, RAM_TAG),
 		m_fdc(*this, "wd1771"),
 		m_floppy0(*this, "wd1771:0"),
 		m_floppy1(*this, "wd1771:1")
 	{
 	}
```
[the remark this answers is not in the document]
RobinSergeant: Yes, true

src/mame/rm/rm380z.h
```diff
@@ -38,27 +39,121 @@ class rm380z_state : public driver_device
...
 protected:
-	virtual void machine_reset() override;
-	virtual void machine_start() override;
+	void machine_reset() override;
+
+	static inline constexpr int RM380Z_SCREENROWS = 24;
+	static inline constexpr int RM380Z_SCREENCOLS = 40;
+
+	bool ports_enabled_high() const { return ( m_port0 & 0x80 ); }
+	bool ports_enabled_low() const { return !( m_port0 & 0x80 ); }
```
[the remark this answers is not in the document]
RobinSergeant: done

src/mame/rm/rm380z.h
```diff
@@ -79,89 +174,100 @@ class rm380z_state : public driver_device
...
-	void disk_0_control(uint8_t data);
+	void configure(machine_config &config) override;
 
-	void keyboard_put(u8 data);
+	DECLARE_INPUT_CHANGED_MEMBER(monitor_changed);
 
-	DECLARE_MACHINE_RESET(rm480z);
+private:
+	enum class hrg_display_mode: uint8_t
+	{
```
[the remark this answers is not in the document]
RobinSergeant: I think that was a typo - fixed now.

src/mame/rm/rm380z.h
```diff
@@ -79,89 +174,100 @@ class rm380z_state : public driver_device
...
+	void draw_high_res_graphics(bitmap_ind16 &bitmap) const;
+	void draw_medium_res_graphics(bitmap_ind16 &bitmap) const;
 
-	uint8_t m_port0 = 0;
-	uint8_t m_port0_mask = 0;
-	uint8_t m_port0_kbd = 0;
-	uint8_t m_port1 = 0;
-	uint8_t m_fbfd_mask = 0;
-	uint8_t m_fbfe = 0;
+	void machine_reset() override;
```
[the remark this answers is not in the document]
RobinSergeant: I've moved the others as well.

src/mame/rm/rm380z.h
```diff
@@ -79,89 +174,100 @@ class rm380z_state : public driver_device
...
 
-	required_region_ptr<u8> m_chargen;
-	required_device<cpu_device> m_maincpu;
-	optional_device<screen_device> m_screen;
-	optional_device<cassette_image_device> m_cassette;
-	optional_device<ram_device> m_messram;
-	optional_device<fd1771_device> m_fdc;
-	optional_device<floppy_connector> m_floppy0;
-	optional_device<floppy_connector> m_floppy1;
+	DECLARE_MACHINE_RESET(rm480z);
```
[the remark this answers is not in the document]
RobinSergeant: Yes, done that now.

src/mame/rm/rm380z_m.cpp
```diff
@@ -31,25 +31,41 @@ void rm380z_state::port_write(offs_t offset, uint8_t data)
...
 		config_memory_map();
 		break;
 
-	case 0xfd:      // screen line counter (?)
-		//printf("%s FBFC [%2.2x] FBFDw[%2.2x] FBFE [%2.2x] writenum [%4.4x]\n",machine().describe_context().c_str(),m_port0,data,m_fbfe,writenum);
+	case 0xff:      // user I/O port
+		break;
+
+	default:
+		printf("unknown port [%2.2x] write of [%2.2x]\n", offset, data);
```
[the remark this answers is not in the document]
RobinSergeant: Done

src/mame/rm/rm380z_m.cpp
```diff
@@ -101,15 +157,55 @@ uint8_t rm380z_state::port_read(offs_t offset)
...
+		{
+			data |= 0x80;
+		}
+		break;
+
+	case 0xff:      // user port
+		break;
+
+	default:
+		printf("read from unknown port [%2.2x]\n", offset);
```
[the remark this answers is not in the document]
RobinSergeant: Done

src/mame/rm/rm380z.h
```diff
@@ -38,27 +39,121 @@ class rm380z_state : public driver_device
...
 	}
 
-	void rm480z(machine_config &config);
-	void rm380z(machine_config &config);
-
-	void init_rm380z();
-	void init_rm380z34d();
-	void init_rm380z34e();
-	void init_rm480z();
+	virtual void configure(machine_config &config);
```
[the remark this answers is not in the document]
cuavas: You don’t have to give it the same name in derived classes.  The thing is, making it virtual generates unnecessary additional code, and in something as big as MAME, that all adds up.

cuavas:
> Thanks @cuavas, I've just addressed your latest review comments and also squeezed in one last fix to this defect that I found last night (which I hope is ok!):
> 
> https://mametesters.org/view.php?id=6483
> 
> Basically at startup the firmware reads F600 and then tries to execute code at F601 if this comes back as zero. It does this because F600 is the location for a possible ROM extension module. As this was blank it just executed a lot of NOP instructions before eventually hitting opcode FF which triggers the built in debugger (front panel).
> 
> To fix I've just used the ROMREGION_ERASEFF flag so that unused areas of ROM read FF instead of zero. Now the COS prompt appears like it should on the COS 3.4 machines.

That fix will be OK for now.  Maybe in the future a media device can be mapped there so people can use it with user ROMs.

PR #12045, opened 2024-02-19
capcom/1942.cpp: Add a 1942 bootleg from Itisa (not working)

clawgrip:


clawgrip:
Closing this PR, will open another new one with this set, as it belongs to vulgus.cpp, not 1942.cpp

PR #12047, opened 2024-02-20
Remove some unnecessary backslash line continuation

amameuser:
* devices/*: Remove unnecessary backslash line continuation

* frontend/*: Remove unnecessary backslash line continuation

* lib/*: Remove unnecessary backslash line continuation

* mame/*: Remove unnecessary backslash line continuation

* osd/*: Remove unnecessary backslash line continuation

* cpu/m68000/m68000gen.py: Use parentheses in preference to using a backslash for line continuation

* cpu/m68000/m68000gen.py: Regenerate source files

  m68000gen.py decode m68000.lst m68000-decode.cpp
  m68000gen.py header m68000.lst m68000-head.h
  m68000gen.py sdf m68000.lst m68000-sdf.cpp
  m68000gen.py sdp m68000.lst m68000-sdp.cpp
  m68000gen.py sif m68000.lst m68000-sif.cpp
  m68000gen.py sip m68000.lst m68000-sip.cpp
  m68000gen.py headerm m68000.lst m68000mcu-head.h
  m68000gen.py sdfm m68000.lst m68000mcu-sdfm.cpp
  m68000gen.py sdpm m68000.lst m68000mcu-sdpm.cpp
  m68000gen.py sifm m68000.lst m68000mcu-sifm.cpp
  m68000gen.py sipm m68000.lst m68000mcu-sipm.cpp
  m68000gen.py header8 m68000.lst m68008-head.h
  m68000gen.py sdf8 m68000.lst m68008-sdf8.cpp
  m68000gen.py sdp8 m68000.lst m68008-sdp8.cpp
  m68000gen.py sif8 m68000.lst m68008-sif8.cpp
  m68000gen.py sip8 m68000.lst m68008-sip8.cpp


src/lib/netlist/build/makefile
```diff
@@ -151,7 +151,7 @@ OBJDIRS = $(OBJ) \
 			$(OBJ)/tests \
 			$(OBJ)/tools \
 			$(OBJ)/prg \
-			$(OBJ)/generated \
+			$(OBJ)/generated
```
cuavas, line 154: The reason for these ones is so that if an item is added or removed at the end of the list, only a single line changes.  Without the final continuation backslash, two lines need to change if an item is added or removed at the end of the list.

I don’t particularly like them though, as it’s somewhat error-prone.  Removing the blank line after the list will cause bad things to happen.

PR #12050, opened 2024-02-21
machine/z80scc.cpp: Added support for RTxC transmit/receive clock source

colinleroy:
Fixes serial emulation at 115200bps.

This leaves some things unimplemented (like TRxC and DPLL clock sources), but makes programs communicating at 115.200bps function properly.

For context, toggling this chip to 115200bps or 230400bps is done by
1) setting the clock multiplier to 16 (230400bps) or 32 (115200bps), WR4 bits 6 or 7
2) setting the clock source for TX and RX to /RTxC instead of BRG, WR11
3) stopping the baud rate generator, WR14 bit 0

I hope this is a good first PR.
Thanks!

### DIFF
--- a/src/devices/machine/z80scc.cpp
+++ b/src/devices/machine/z80scc.cpp
@@ -2092,7 +2092,7 @@ void z80scc_channel::do_sccreg_wr11(uint8_t data)
 	  /RTxC pin.*/
 	switch (data & WR11_RCVCLK_SRC_MASK)
 	{
-	case WR11_RCVCLK_SRC_RTXC: LOG("Receive clock source is RTxC - not implemented\n"); break;
+	case WR11_RCVCLK_SRC_RTXC: LOG("Receive clock source is RTxC\n"); break;
 	case WR11_RCVCLK_SRC_TRXC: LOG("Receive clock source is TRxC - not implemented\n"); break;
 	case WR11_RCVCLK_SRC_BR:   LOG("Receive clock source is Baudrate Generator\n"); break;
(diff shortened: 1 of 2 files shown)


src/devices/machine/z80scc.cpp
```diff
@@ -2893,9 +2907,17 @@ void z80scc_channel::update_serial()
...
+				&& (m_wr11 & WR11_TRACLK_SRC_MASK) == WR11_TRACLK_SRC_RTXC) {
+			m_brg_rate = get_rtxc_rate();
+			LOG("- BRG disabled, clock source RTxC (rate %d, clock %d)\n", m_brg_rate, get_clock_mode());
+			set_rcv_rate(m_brg_rate);
+			set_tra_rate(m_brg_rate);
+		} else {
+			LOG("- BRG disabled and RX/TX clock sources differ, unimplemented: stopping\n");
+			set_rcv_rate(0);
+			set_tra_rate(0);
+		}
```
cuavas, line 2920: Please follow the formatting of the file you’re editing.  The rest of this file uses Allman block formatting, don’t use K&R formatting for new code.

cuavas (changes requested):
The content of the changes looks OK, but you should follow the formatting of the file you’re editing.

cuavas (commented):
The code looks OK and matches the style of the file now.

Could someone who’s more familiar with the Z8530, and its implementation in MAME, please check to ensure this matches the documented behaviour and isn’t baking in overly specific assumptions about how the chip is hooked up in the system?

Maybe @rb6502, @pmackinlay or @MooglyGuy since you’ve worked on systems using the Z8530?

PR #12054, opened 2024-02-22
emu/save.cpp: Make illegal save state registrations always fatal

cuavas:
The only thing that causes non-fatal illegal save state registrations right now is registering an item for save states after start in a system not marked as supporting save states.

It’s high time anything doing that gets fixed.  Not supporting save states means save state support is incomplete, not that the driver has a license to do illegal things.

Opening this as a pull request to give fair warning, and so it can be merged after freeze.

### DIFF
--- a/src/emu/machine.cpp
+++ b/src/emu/machine.cpp
@@ -890,10 +890,6 @@ void running_machine::handle_saveload()
 				// handle the result
 				switch (saverr)
 				{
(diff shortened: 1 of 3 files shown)


galibert:
IIRC the only times I saw that warning is with ethernet networking.

rb6502:
+1 on the network layer having so much wrong it's not funny.  It's difficult even to trace the flow of control.

PR #12067, opened 2024-02-26
gdbstub fixes for Windows and m68000

BartmanAbyss:
- `debug_gdbstub::generate_target_xml()` contained newlines that were converted from '\n' to '\r\n' on Windows, thus breaking the gdb packet checksum (`debug_gdbstub::send_reply`), so removed newlines in `escape_packet`. GDB doesn't care about newlines in XML files. This seemed to be the least invasive fix I could think of.
- m68000: complained about not finding registers  `A7` => `SP`, `PC` => `CURPC`

I guess we should rename other instances of `PC` to `CURPC` as well, as `device_state_entry::device_state_entry()` automatically changes the name of PC to `CURPC` (see distate.cpp line 66)


src/osd/modules/debugger/debuggdbstub.cpp
```diff
@@ -682,6 +682,9 @@ static std::string escape_packet(const std::string src)
 	result.reserve(src.length());
 	for ( char ch: src )
 	{
+		if ( ch == '\n' ) // don't let socket convert line endings and messing up the checksum
+			continue;
+
```
cuavas, line 687: Wouldn’t the proper solution to this be to change the calls to `m_socket.puts(...)` to use `m_socket.write(...)` instead to avoid text conversion altogether?  This looks like working around the issue rather than actually fixing it.

BartmanAbyss: yeah, probably. wasn't familiar with the socket interface :)

PR #12073, opened 2024-02-27
New working software list items (apple2_flop_misc.xml)

Firehawke:

Anti-M (Version 1.9) [4am, Firehawke]
Anti-M (Version 2.0) [4am, Firehawke]
Anti-M (Version 2.1) [4am, Firehawke]
Anti-M (Version 2.2) [4am, Firehawke]


Firehawke:
Anti-M 1.8 had a set rename; I believe (but can't remember for sure; let's just say that it was an incredibly short-sighted naming choice on my part that I own up to) at the time I set the name, it was believed to be the last version needed. Clearly that hasn't been the case, so no more "antim", all new versions I add will be numbered period.

peterferrie:
We were trying to reach a point of "we're done" but we missed it because of
my bugs and (yesterday) one new title.  Sorry.

Firehawke:
No, the mistake is on me. No software is ever truly perfect or final.

cuavas (commented):
This isn’t validating because the parent/clone relationships are broken.  The latest version should have the short name `antim` and all the older versions should be marked as clones.  The `release` info element should also be straight YYYYMMDD with no punctuation to match other software lists, to avoid breaking front-ends that depend on it.

cuavas:
GitHub isn’t showing me updates for new commits on this PR, either.  Something’s going whacky in the backend.

PR #12055, opened 2024-02-22
ekara: add 9 cartridges [David Haywood, Team Europe]

mamehaze:
new software list entries
-----------
Kids' Song Best 40 (Japan) (EC0084-KSB) [Team Europe, David Haywood]
Morning Musume Special (Japan) (BX01-MOR) [Team Europe, David Haywood]
Enka-shū Dai Volume 1 (Japan) (EN-1) [Team Europe, David Haywood]
Enka-shū Dai Volume 2 (Japan) (EN-2) [Team Europe, David Haywood]
Kids' Song 20 (Japan) (KD-1) [Team Europe, David Haywood]
J-Pop Mix Mini Volume 1 (Japan) (MC0009-JPM) [Team Europe, David Haywood]
Artist Mini Volume 9 (Yamaguchi Momoe) (Japan) (MC0016-ATM) [Team Europe, David Haywood]
PostPet (Japan) (SC0011-PST) [Team Europe, David Haywood]
Saiten Cartridge: Nesshō vol. 5 (Japan) (SC0016-SAI) [Team Europe, David Haywood]

these titles aren't optimal, ideally this will be merged and the titles corrected by somebody who knows the language, along with the alt tags added.

new NOT WORKING machines
----------------
Sou-Kai Golf Champ (Japan) [Team Europe, David Haywood]
Nekketsu Pawapuro Champ (Japan) [Team Europe, David Haywood]
Zettai Zetsumei Dangerous Jiisan - Mini Game de Taiketsu ja! (Japan) [Team Europe, David Haywood]
Zettai Zetsumei Dangerous Jiisan Party ja! Zen-in Shuugou!! (Japan) [Team Europe, David Haywood]

PR #12074, opened 2024-02-28
Use literal romaji for Jikkyou Powerful Pro Yakyuu game titles

cuavas:
Looking to solicit opinions on this.

The 実況パワフルプロ野球 franchise is usually called “Jikkyou Powerful Pro Yakyuu” in English.  However, in Japanese media, it seems to be more frequently rendered in literal romaji as “JIKKYOU PAWAFURU PURO YAKŪ”.  You can see this in the attract sequences for the games on Konami System 573, for example (e.g. `powyakex`).

“Powerful” and “Pro” are definitely loanwords from English.  The MAME convention has always been to use standard English spellings for loanwords.  Also, if we switch to literal romaji, we’ll be the odd ones out as all the other software catalogues will still be using “Jikkyou Powerful Pro Yakyuu”.

So in this case, should “JIKKYOU PAWAFURU PURO YAKŪ” be treated as an official Latin script title and used for system and software descriptions?  Or should it be treated as a pronunciation guide so we keep the conventional spellings of the loanwords?

### DIFF
--- a/hash/dc.xml
+++ b/hash/dc.xml
@@ -17967,7 +17967,7 @@ Has [redbook] audio
 		<rom name="track02.raw" size="1735776" md5="d09be539a7074a058159505b7f5fdcb3" sha1="db680939aae4bf61d0e588414c72a47c6dd92cdd"/>
 		<rom name="track03.bin" size="1185760800" md5="b0c73d82f36dd05235c5f2b38793c15b" sha1="8b073759e6cb0a2bc5f4e77978d7a129e22c0878"/>
 		-->
-		<description>Jikkyou Powerful Pro Yakyuu - Dreamcast Edition (Japan)</description>
+		<description>Jikkyou Pawafuru Puro Yakyū - Dreamcast Edition (Japan)</description>
 		<year>2000</year>
 		<publisher>Konami</publisher>
 		<info name="serial" value="T-9507M"/>
--- a/hash/dc.xml
+++ b/hash/dc.xml
@@ -17973,6 +17973,7 @@ Has [redbook] audio
 		<info name="serial" value="T-9507M"/>
 		<info name="release" value="20000330"/>
 		<info name="alt_title" value="実況パワフルプロ野球 Dreamcast Edition"/>
+		<info name="alt_title" value="Jikkyou Powerful Pro Yakyuu - Dreamcast Edition"/> <!-- often used unofficially -->
 		<part interface="cdrom" name="cdrom">
 			<diskarea name="cdrom">
 				<disk name="jikkyou powerful pro yakyuu - dreamcast edition (jpn)" sha1="07e2e64eaa62690b431998750e7d3976d36db6a0"/>
--- a/hash/n64.xml
+++ b/hash/n64.xml
@@ -5867,7 +5867,7 @@ Black screen after Rare logo
 	</software>
 
 	<software name="powyak2k" supported="no">
-		<description>Jikkyou Powerful Pro Yakyuu 2000 (Japan, rev 1)</description>
+		<description>Jikkyou Pawafuru Puro Yakyū 2000 (Japan, rev 1)</description>
 		<year>2000</year>
 		<publisher>Konami</publisher>
 		<info name="serial" value="NUS-NPAJ-JPN (RZ034-J1)"/>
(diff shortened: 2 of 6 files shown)


etabeta78:
Personally, I think it shall be "Powerful Pro", with an alt_title field
when the literal romaji is also used.

   my 2c

ICEknigh7:
> “Powerful” and “Pro” are definitely loanwords from English. The MAME convention has always been to use standard English spellings for loanwords.

This has always worked fine towards not making Japanese game titles feel like a late 90s anime fansub.

ICEknigh7:
> “Powerful” and “Pro” are definitely loanwords from English. The MAME convention has always been to use standard English spellings for loanwords.

This has always worked fine towards not making Japanese game titles feel like a late 90s anime fansub.

> with an alt_title field when the literal romaji is also used.

Yeah, an _optional_ alt_title for that could work if this is _really_ needed.

ICEknigh7:
> “Powerful” and “Pro” are definitely loanwords from English. The MAME convention has always been to use standard English spellings for loanwords.

I feel that this has always worked fine towards not making Japanese game titles feel like a late 90s anime fansub.


> with an alt_title field when the literal romaji is also used.

Yeah, an _optional_ alt_title for that could work if this is needed.

balr0g:
How do we handle when romaji and transliterations don't necessarily match box art? Is there any official policy or practice?

https://github.com/mamedev/mame/pull/12055#issuecomment-1961730490 — "Sou-Kai Golf Champ" [box art] vs. "Soukai Golf Champ" [MAME]

https://github.com/mamedev/mame/pull/12055#issuecomment-1967302317 — "Nekketsu Pawapuro Champ" [box art] vs "Nekketsu Powerpro Champ" [MAME]

If I were to come across the boxed game e.g. in a lot or on eBay, I'd be likely searching the box art text, so I would argue that it's worth having that version in MAME as well (perhaps as an optional alt_title, noted that this is the one found on the box art for this specific game).

angelosa:
It's `Konami Sports` by definition, it's assumed that any katakana used is English loan words, that is marketed or not outside Japan doesn't fit here.
It's Jikkyou J. League World Soccer Perfect Striker not Jikkyō J Rīgu Pāfekuto Sutoraikā.

cuavas:
> It's `Konami Sports` by definition, it's assumed that any katakana used is English loan words, that is marketed or not outside Japan doesn't fit here.
> It's Jikkyou J. League World Soccer Perfect Striker not Jikkyō J Rīgu Pāfekuto Sutoraikā.

Yeah, but in that case, Japanese media always uses “J LEAGUE PERFECT STRIKER” in Latin script.  This one is weird in that they actually use literal romaji for loanwords.

angelosa:
Interesting, the Saturn version report it as you said (most if not all Jp Saturn games have a romaji translitteration in one of the cover sides)
https://www.segaretro.org/images/thumb/7/70/JPPYS_Saturn_JP_Box_Back.jpg/1524px-JPPYS_Saturn_JP_Box_Back.jpg

angelosa:
In Western market the series is known as Power Pro, but I feel like that the domestic _intention_ is to market it for what it is (an Arcade-ish not pretending to be a sim)

ICEknigh7:
This is how the Saturn box romanizes Astal, but I think it should stay "Astal":
image

angelosa:
Comparing an early release Sega game (that also has a cyan rating on front cover) with a Konami one is like comparing apples and bananas.

rb6502:
Ideally for Japanese locales we'd show the actual kanji title and the literal romaji, but as far as I know that's not possible, and there are lower-hanging things with software lists to address (save states per software list item, for instance).  So barring that I'd say make the Description the literal romaji as this PR does, and add alt_titles with the normal English spellings of the loanwords.

etabeta78:
About softlist saves, can't you use -savename "%g/%d_cart" to use savenames
based on the name of the software loaded in cart? (replace with cass,
flop1, flop2, etc. for other cases)

    f.

etabeta78:
For item you are right that there might be discrepancies, but if you save
and load in the same switch config still work

It fails when you are dealing with multipart discs, which is and was since
forever the main limitation I regretted about, but no smarter solutions
came to my mind

Overall, I think we need “multipart” save so that some can be optional
(like controller devices) and some are mandatory to be restored, and some
can be declined depending on the different software config (e.g. erroring
out if you are not mounting the same disks in the same drives)

We are not there yet, though

  Fabio

suverman:
When you arent chasing away the real devs, you are wasting other people's time with such trivial matters. Way to go killing a great project slowly..

PR #12078, opened 2024-02-28
spi_sdcard.cpp: implement SEND_CSD (SD_TYPE_V2)

holub:


src/devices/machine/spi_sdcard.cpp
```diff
@@ -259,8 +259,43 @@ void spi_sdcard_device::do_command()
 			break;
 
 		case 9: // CMD9 - SEND_CSD
-			m_data[0] = 0x00; // TODO
-			send_data(1, SD_STATE_STBY);
+			m_data[0] = 0x00;
+			m_data[1] = 0xff;
+			m_data[2] = 0xfe;
+
+			//if (m_type == SD_TYPE_V2) // CSD Version 1.0
```
rb6502, line 266: If this check isn't necessary, remove it.  (If it is, comment it back in).  Same for the else block below.

holub: It's necessary for 2.0 (oops typo there). I just don't have a case to test it but clearly it must be revisited and implemented at some point

rb6502: Ok, that's fair.  I'm a big proponent of not implementing things you don't have a test case for.

PR #12080, opened 2024-02-28
trs/mc10.cpp: Add "Multiport" cartridge and RAM expansion for the Matra & Hachette Alice

Mokona:
This addition to the Alice line (4k, 32 and 90) adds RAM and Cartridge support in a different way as the already supported Darren Atkinson's MCX-128 cartridge.


happppp:
@Mokona the new files are missing license header, see other source files to see what I mean.

Mokona:
Thanks for the review. As this was already merged, I will provide another PR soon.

Issue #12081, opened 2024-02-28
chdman 0.263 error does not compress dreamcast games


CorvuzSoli:
### MAME version

0.263

### System information

Windows 10

### Incorrect behaviour

chdman 0.263 error does not compress dreamcast games, I tried with version 0.262 and it also has that problem.

### Expected behaviour

 I tried with version 0.261 and I had no problems with it.

### Steps to reproduce

C:\Users\xxxx\Desktop\chdman> for /R %i in (*.cue *.gdi *.iso) do chdman createcd -i "%i" -o "%~ni.chd"

C:\Users\xxxx\Desktop\chdman>chdman createcd -i "C:\Users\xxxx\Desktop\chdman\Star Wars - Episode I - Racer (USA).cue" -o "Star Wars Episode I - Racer (USA).chd"
chdman - MAME Compressed Hunks of Data (CHD) manager 0.263 (mame0263)
Output CHD: Star Wars - Episode I - Racer (USA).chd
Input file: C:\Users\xxxx\Desktop\chdman\Star Wars - Episode I - Racer (USA).cue
Input tracks: 5
Input length: 122:02:00
Compression: cdlz (CD LZMA), cdzl (CD Deflate), cdf1 (CD FLAC)
Logical size: 1,344,333,888
Error reading input file (C:\Users\xxxx\Desktop\chdman\Star Wars - Episode I - Racer (USA) (Track 4).bin)'
exception occurred: std::exception
Error during compression: Input/output error

987123879113:
Looks like it broke with https://github.com/mamedev/mame/commit/b75b8d80e87c4bfb7d040385cb09b6f24eafe604. 

Someone made a comment in the related issue last month about the same problem it seems. https://github.com/mamedev/mame/issues/11903

987123879113:
The exact cause seems to be this code here. 

https://github.com/mamedev/mame/blob/1615b8551a3a73532ac234973cfb04dd8ed98ba4/src/lib/util/cdrom.cpp#L2889-L2905


```
trk 1: 603 frames @ offset 0, pad=0, split=0, area=0, phys=0, pregap=0, pgtype=0, idx0=0, idx1=0, (true 603)
trk 2: 44397 frames @ offset 0, pad=43721, split=0, area=0, phys=603, pregap=150, pgtype=7, idx0=0, idx1=150, (true 676)
[...]

PR #12087, opened 2024-03-02
util/cdrom: Refactor parse_cue and parse_gdicue

987123879113:
Closes https://github.com/mamedev/mame/issues/12081. Follow up to https://github.com/mamedev/mame/pull/11913.

After fixing the problem mentioned in the issue (the first commit) I realized there isn't that much unique code between `parse_gdicue` and `parse_cue` so I decided to merge them into one function so that there's no need to maintain two separate .cue parsers anymore.

Additionally I did some other small cleanups such as clearing out some of the line buffers and adding some additional error checks in. Sufficient size checks were already in place already so it wasn't likely to crash anything but I figured it didn't hurt while I'm at changing things around. Mainly was concerned that it looks like you could end up copying a bunch of garbage into `token` if `linebuffer` had garbage in it when `cdrom_file::tokenize` was called.

For testing I extracted some v5 CHDs I had on hand and then recreated them to make sure the hashes matched. Also tried a few Redump Dreamcast dumps to make sure the extracted .bin matched the original Redump data (compared with all individual tracks combined into a single file so I could just do a SHA1 check between the CHD dump and original data).

### DIFF
--- a/src/lib/util/cdrom.cpp
+++ b/src/lib/util/cdrom.cpp
@@ -1414,14 +1414,14 @@ void cdrom_file::ecc_clear(uint8_t *sector)
  *
  * @brief   A macro that defines tokenize.
  *
- * @param   linebuffer          The linebuffer.
- * @param   i                   Zero-based index of the.
- * @param   sizeof(linebuffer)  The sizeof(linebuffer)
- * @param   token               The token.
- * @param   sizeof(token)       The sizeof(token)
+ * @param   linebuffer             The linebuffer.
+ * @param   i                      Zero-based index of the.
+ * @param   std::size(linebuffer)  The std::size(linebuffer)
+ * @param   token                  The token.
+ * @param   std::size(token)       The std::size(token)
  */
 
-#define TOKENIZE i = tokenize( linebuffer, i, sizeof(linebuffer), token, sizeof(token) );
+#define TOKENIZE i = tokenize( linebuffer, i, std::size(linebuffer), token, std::size(token) );
 
 
 /***************************************************************************
@@ -2263,7 +2263,7 @@ std::error_condition cdrom_file::parse_cue(std::string_view tocfname, toc &outto
 				/* TODO: sessions are notated using REM commands: "REM SESSION 01" */
 
 				/* skip to actual data of REM command */
-				while (i < sizeof(linebuffer) && isspace((uint8_t)linebuffer[i]))
+				while (i < std::size(linebuffer) && isspace((uint8_t)linebuffer[i]))
 					i++;
 
 				if (is_gdrom && !strncmp(linebuffer+i, "SINGLE-DENSITY AREA", 19))
@@ -2618,7 +2618,7 @@ bool cdrom_file::is_gdicue(std::string_view tocfname)
 			if (!strcmp(token, "REM"))
 			{
 				/* skip to actual data of REM command */
-				while (i < sizeof(linebuffer) && isspace((uint8_t)linebuffer[i]))
+				while (i < std::size(linebuffer) && isspace((uint8_t)linebuffer[i]))
 					i++;
 
 				if (!strncmp(linebuffer+i, "SINGLE-DENSITY AREA", 19))


src/lib/util/cdrom.cpp
```diff
@@ -2285,30 +2320,37 @@ std::error_condition cdrom_file::parse_cue(std::string_view tocfname, toc &outto
...
 				outinfo.track[trknum].idx0offs = -1;
 				outinfo.track[trknum].idx1offs = 0;
+				outinfo.track[trknum].offset = 0;
+
+				if (wavlen != 0)
+				{
+					outtoc.tracks[trknum].frames = wavlen/2352;
+					outinfo.track[trknum].offset = wavoffs;
+					wavoffs = wavlen = 0;
+				}
```
987123879113, line 2339: This originally set `outtoc.tracks[trknum].trktype = CD_TRACK_AUDIO;` but directly below this `convert_type_string_to_track_info` gets called which will set `outtoc.tracks[trknum].trktype` based on the TRACK's typestring so I removed it. 

Maybe it would be a better idea to move the call to `convert_type_string_to_track_info` before this so the track type can be forced when a WAVE file type is selected as input?

987123879113:
I ended up effectively https://github.com/mamedev/mame/pull/11913 and rewrote how Dreamcast cue/bins are parsed. This PR breaks both the SHA1 and sometimes the Data SHA1 of any Dreamcast CHDs previously made using Redump cue/bins but it feels necessary. CHDs created using GDIs (both TOSEC and Redump GDIs) are unaffected.

I didn't test with a ton of games but I did test with all of the type varieties previously implemented. Most games are either "type I" or "type III split". Here's the breakdown using all of the Dreamcast cues from redump's database:
```
type1 796
type2 3
type3 4
type3_split 590
multisession 80
unknown 0
total 1473
```

For easy testing/reference, here are the type II and type III (non-split) games:
```
    "type2": [
        "UnderCover AD2025 Kei (Japan).cue",
        "Memories Off 2nd - Making Disc (Japan).cue",
        "Roommate Asami - Okusama wa Joshikousei - Director's Edition (Japan).cue"
    ],
    "type3": [
        "Shenmue II (Japan) (Disc 3).cue",
        "Shenmue II (Japan) (Disc 2).cue",
        "Shenmue II (Japan) (Disc 4).cue",
        "Virtua Fighter History & VF4 (Japan).cue"
    ],
```

Disclaimer: The TOSEC .raw files all seem to have the audio data start at a random offset (dump method issue?) whereas Redump's audio tracks were all uniform from my testing, so for testing purposes I manually modified the TOSEC .raw files to have the data start at offset 0 and padded the files an appropriate length to keep the same filesize.

- Create CHD from TOSEC GDI, extract to GDI
    - Working, generates same output bins+GDI as input TOSEC GDI.
- Create CHD from Redump cue/bin, extract to Redump cue/bin
    - Working but not perfect, generates invalid cue (no REM single/double commands) but the output data is the same as input. Can be fully recovered by splitting the .bin into separate tracks using a .cue from Redump.
- Create CHD from Redump cue/bin, extract to GDI
    - Because the data internally gets shifted to match TOSEC GDI, the output GDI becomes a TOSEC GDI but has the pregap data normally not present in the TOSEC rips included at the end of the track data. For example, for `Star Wars - Episode I - Racer v1.001 (US)` the TOSEC GDI says track 2 starts at 753 (so 753 * 2352 = 0x1b0630 bytes of data between track 1 and track 2, but the actual track01.bin is (753 - 150) * 2352 = 0x15a410 bytes). Personally I don't think this should cause problems because that data appears to have been just left blank in the TOSEC GDI dumps, and now it's filled in with real data that'd go in between those gaps. The actual data outside of those gaps is the same and if you remove the extra data at the end of the tracks to match the TOSEC GDI dump filesizes then the extracted track data has the same SHA-1 hashes as the original TOSEC GDI dump track data.
- Create CHD from Redump GDI, extract to GDI
    - Working, generates same output bins+GDI as input Redump GDI. Format of GDI doesn't match exactly due to text formatting but contents are the same. I would not recommend creating CHDs using the Redump GDIs simply for the fact that it'll be in a different format internally compared to the TOSEC GDI and Redump cue/bin CHDs (pregap is associated to different track compared to TOSEC layout and there's no easy way to reorder it without some hacky way to detect a Redump GDI vs TOSEC GDI).
- Create CHD from Redump GDI, extract to Redump cue/bin
    - Working but not perfect, generates same output bin as input Redump bins but the cue's INDEX times are different due to pregap being associated with a different track compared to TOSEC layout.

src/tools/chdman.cpp
```diff
@@ -2627,6 +2628,11 @@ static void do_extract_cd(parameters_map &params)
 			mode = MODE_GDI;
 		}
 
+		if (cdrom->is_gdrom() && (mode == MODE_CUEBIN))
+		{
+			util::stream_format(std::cout, "Warning: extracting GD-ROM CHDs as bin/cue is not fully supported and will result in an unusable CD-ROM cue file.\n");
+		}
+
```
987123879113, line 2635: This never worked even before the recent changes, so finally add a warning about it. Someone needs to go in and add proper support for GD-ROM cues.

src/lib/util/cdrom.cpp
```diff
@@ -1493,17 +1495,17 @@ uint64_t cdrom_file::get_file_size(std::string_view filename)
...
+	bool singlequote = false;
+	bool doublequote = false;
 
 	while ((i < linebuffersize) && isspace((uint8_t)linebuffer[i]))
 	{
 		i++;
 	}
 
-	while ((i < linebuffersize) && (j < tokensize))
+	while ((i < linebuffersize) && (j < tokensize) && (linebuffer[i] != '\0'))
```
cuavas, line 1506: How does `linebuffer[i] != '\0'` happen?  Is `linebuffersize` the size of the buffer rather than the size of its content?  The code seemed to be treating it as the size of the content.

src/lib/util/cdrom.cpp
```diff
@@ -2506,6 +2587,7 @@ std::error_condition cdrom_file::parse_cue(std::string_view tocfname, toc &outto
 
 bool cdrom_file::is_gdicue(std::string_view tocfname)
 {
+	static char token[512];
```
cuavas, line 2590: Why is this static?  It prevents the function from being called reentrantly.  I can’t immediately see a reason for it to outlive an invocation.

cuavas (commented):
I didn’t see any major issues with a quick read, but I’m pretty tired and I’m not overly familiar with this part of the code.  It really needs testing, and probably some more pairs of eyes.

I’m not very happy about the fallout from the last change to the GD-ROM handling, or the other issues in chdman that keep being found.  Issues in CHD handling don’t just affect MAME.

src/lib/util/cdrom.cpp
987123879113: `linebuffersize` is just `sizeof(linebuffer)` so it's the entire buffer size rather than its contents. So without the null byte check, the entire buffer is processed every time instead of just ending when reasonable.

https://github.com/mamedev/mame/blob/1565fb9ba6e2543558553131dc2e31dedfc41213/src/lib/util/cdrom.cpp#L1411-L1422

src/lib/util/cdrom.cpp
987123879113: `parse_cue` and `parse_toc` both handle the token buffer used by the tokenizer in that way and I remade `is_gdicue` in the style of those existing functions, so it ended up like that. As for a general why it's like that in cdrom.cpp, I don't know but probably something to do with the general lack of modernization in the file.

src/lib/util/cdrom.cpp
cuavas: Eugh, if linebuffer is a pointer, bad things will happen.  Will it work if you change it to `tokenize( linebuffer, i, std::size(linebuffer), token, std::size(token) );` so it will give an error if `linebuffer` and `token` aren’t array-like?

src/lib/util/cdrom.cpp
```diff
@@ -2519,428 +2601,37 @@ bool cdrom_file::is_gdicue(std::string_view tocfname)
...
-				int frames;
-
-				/* get index */
-				TOKENIZE
-				frames = msf_to_frames( token );
-
-				outtoc.tracks[trknum].postgap = frames;
+				/* skip to actual data of REM command */
+				while (i < sizeof(linebuffer) && isspace((uint8_t)linebuffer[i]))
+					i++;
```
cuavas, line 2622: This is another place where `std::size(linebuffer)` should be used.

src/lib/util/cdrom.cpp
987123879113: Yeah, it works from some quick testing.

cuavas:
I merged this.  It should put us in a better position than where we’ve been for the last few months, anyway.  I’d still appreciate if more people could give it a spin.

TheRealGusBus:
@cuavas Tested against a nearly complete Redump set (a few seem MIA) using oxyromon as the ROM manager. Re-validated the CHD files using the original CUE files with no obvious issues.

cuavas:
> @cuavas Tested against a nearly complete Redump set (a few seem MIA) using oxyromon as the ROM manager. Re-validated the CHD files using the original CUE files with no obvious issues.

Thanks.

johnsanc314:
OK then... specifically, this disc: [Redump ID 49774](http://redump.org/disc/49774/)

PR #12088, opened 2024-03-02
trs/mc10.cpp: add minimum rom size and block size for cartridge rom, clean global namespace

Mokona:
Resolving the comments on #12080

- Clean some mentioned code aspects
- Specify the ROM minimum size and block size
- Sort includes
- Clean the global namespace
- Add license lines

Report the comments on other previously written MC10 cart devices.

### DIFF
--- a/src/devices/bus/mc10/mc10_cart.cpp
+++ b/src/devices/bus/mc10/mc10_cart.cpp
@@ -108,6 +108,18 @@ std::pair<std::error_condition, std::string> mc10cart_slot_device::call_load()
 				image_error::INVALIDLENGTH,
 				util::string_format("Unsupported cartridge size (must be no more than %u bytes)", m_cart->max_rom_length()));
 	}
+	else if (len < m_cart->min_rom_length())
+	{
+		return std::make_pair(
+				image_error::INVALIDLENGTH,
+				util::string_format("Unsupported cartridge size (must be at least %u bytes)", m_cart->min_rom_length()));
+	}
+	else if (len % m_cart->block_rom_length() != 0)
+	{
+		return std::make_pair(
+				image_error::INVALIDLENGTH,
+				util::string_format("Unsupported cartridge size (must be a multiple of %u bytes)", m_cart->block_rom_length()));
+	}
 
 	if (!loaded_through_softlist())
 	{
@@ -176,10 +188,20 @@ void device_mc10cart_interface::interface_pre_start()
 }
 
 /*-------------------------------------------------
-    max_rom_length
+    rom size constraints
 -------------------------------------------------*/
 
 int device_mc10cart_interface::max_rom_length() const
+{
+	return 0;
+}
+
+int device_mc10cart_interface::min_rom_length() const
+{
+	return 0;
+}
+
+int device_mc10cart_interface::block_rom_length() const
 {
 	return 0;
 }
--- a/src/devices/bus/mc10/mc10_cart.cpp
+++ b/src/devices/bus/mc10/mc10_cart.cpp
@@ -108,18 +108,6 @@ std::pair<std::error_condition, std::string> mc10cart_slot_device::call_load()
 				image_error::INVALIDLENGTH,
 				util::string_format("Unsupported cartridge size (must be no more than %u bytes)", m_cart->max_rom_length()));
 	}
-	else if (len < m_cart->min_rom_length())
-	{
-		return std::make_pair(
-				image_error::INVALIDLENGTH,
-				util::string_format("Unsupported cartridge size (must be at least %u bytes)", m_cart->min_rom_length()));
-	}
-	else if (len % m_cart->block_rom_length() != 0)
-	{
-		return std::make_pair(
-				image_error::INVALIDLENGTH,
-				util::string_format("Unsupported cartridge size (must be a multiple of %u bytes)", m_cart->block_rom_length()));
-	}
 
 	if (!loaded_through_softlist())
 	{
@@ -192,16 +180,6 @@ void device_mc10cart_interface::interface_pre_start()
 -------------------------------------------------*/
 
 int device_mc10cart_interface::max_rom_length() const
-{
-	return 0;
-}
-
-int device_mc10cart_interface::min_rom_length() const
-{
-	return 0;
-}
-
-int device_mc10cart_interface::block_rom_length() const
 {
 	return 0;
 }
--- a/src/devices/bus/mc10/mc10_cart.cpp
+++ b/src/devices/bus/mc10/mc10_cart.cpp
@@ -213,7 +213,7 @@ void mc10_cart_add_basic_devices(device_slot_interface &device)
 void alice_cart_add_basic_devices(device_slot_interface &device)
 {
 	// basic devices
-	device.option_add("alice128", ALICE_PAK_MCX128);
+	device.option_add("mcx128", MC10_PAK_MCX128);
 	device.option_add("pak", MC10_PAK);
 	device.option_add("ram", MC10_PAK_RAM);
 	device.option_add("multi", ALICE_MULTIPORTS_EXT);
--- a/src/devices/bus/mc10/mc10_cart.cpp
+++ b/src/devices/bus/mc10/mc10_cart.cpp
@@ -200,7 +200,7 @@ std::pair<std::error_condition, std::string> device_mc10cart_interface::load()
 void mc10_cart_add_basic_devices(device_slot_interface &device)
 {
 	// basic devices
-	device.option_add("mcx128", MC10_PAK_MCX128);
+	device.option_add("mcx128", MC10_PAK_MCX128).default_bios("mc10");
 	device.option_add("pak", MC10_PAK);
 	device.option_add("ram", MC10_PAK_RAM);
 	device.option_add("multi", ALICE_MULTIPORTS_EXT);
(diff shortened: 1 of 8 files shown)


src/devices/bus/mc10/mc10_cart.cpp
```diff
@@ -108,6 +108,18 @@ std::pair<std::error_condition, std::string> mc10cart_slot_device::call_load()
...
+		return std::make_pair(
+				image_error::INVALIDLENGTH,
+				util::string_format("Unsupported cartridge size (must be at least %u bytes)", m_cart->min_rom_length()));
+	}
+	else if (len % m_cart->block_rom_length() != 0)
+	{
+		return std::make_pair(
+				image_error::INVALIDLENGTH,
+				util::string_format("Unsupported cartridge size (must be a multiple of %u bytes)", m_cart->block_rom_length()));
+	}
```
cuavas, line 122: Adding the minimum length and granularity to the interface over-complicates things when they can be checked in `load()` if necessary anyway.  The only reason the maximum length is checked here is to guard against attempting to read an excessively large file as loose software.

src/devices/bus/mc10/mcx128.cpp
```diff
@@ -368,4 +361,7 @@ alice_pak_mcx128_device::alice_pak_mcx128_device(const machine_config &mconfig,
 {
 }
 
+} // anonymous namespace
+
+DEFINE_DEVICE_TYPE_PRIVATE(MC10_PAK_MCX128, device_mc10cart_interface, mc10_pak_mcx128_device, "mc10_mcx128", "Darren Atkinson's MCX-128 cartridge")
 DEFINE_DEVICE_TYPE_PRIVATE(ALICE_PAK_MCX128, device_mc10cart_interface, alice_pak_mcx128_device, "alice_mcx128", "Darren Atkinson's MCX-128 cartridge (Alice ROM)")
```
cuavas, line 367: Do these actually need to be different device types at all?  Would making the ROMs BIOS options not be adequate?

cuavas (changes requested):
This complicates the interface unnecessarily.

src/devices/bus/mc10/mcx128.cpp
Mokona: I am not the author of that part, I only applied to it the comment you made on private device and namespace to the neighbour file. But anyway, as I started to change it, I can push the change further.

If I understand the suggestion, it would be to have only one device type for mcx128, but add a configuration option on it to select between both ROMs described at the beginning of the file?

src/devices/bus/mc10/mc10_cart.cpp
Mokona: Will do.

src/devices/bus/mc10/mcx128.cpp
cuavas: If you look at a card like this: https://github.com/mamedev/mame/blob/master/src/devices/bus/a2bus/grappler.cpp

Note that the Grappler+ has options for the 3.0 and 3.2 ROM revisions.  You select them like `-sl1 grapplus,bios=v30` or `-sl1 grapplus,bios=v32`.

Mokona: I finally had time to look at this. I see a difference between the grappler and the mc10 situation. From what I understand, for grappler, both ROMs are valid for the card. So this is presented as an option, with a default value.

For the MCX128 device, you must use the ROM depending on the ROM installed for the particular flavour of the computer. Both computers are (almost) the same, but they have different system ROMs. And this, if you run the MC-10, you must use the ROM for MC-10 in the MCX128 device. And the same for Alice, with its own ROM.

So that's why the author originally made two devices I guess. Changing the ROM for the device does not represent a valid configuration.

Is there a way to have the "default ROM" depending on the emulated computer?

galibert:
Yeah, look for "set_default_bios_tag".  It works well.

  OG.

src/devices/bus/mc10/mc10_cart.cpp
Mokona: Change done

src/devices/bus/mc10/mcx128.cpp
Mokona: Change done

src/devices/bus/mc10/mcx128.cpp
```diff
@@ -43,15 +43,17 @@
...
 
-ROM_START(alice_mcx128)
-	ROM_REGION(0x4000, "eprom", ROMREGION_ERASE00)
-	ROM_LOAD("alice128bas.rom", 0x0000, 0x4000, CRC(a737544a) SHA1(c8fd92705fc42deb6a0ffac6274e27fd61ecd4cc))
+	ROM_SYSTEM_BIOS(0, "mc10_mcx128", "Darren Atkinson's MCX-128 cartridge")
+	ROMX_LOAD("mcx128bas.rom", 0x0000, 0x4000, CRC(11202e4b) SHA1(36c30d0f198a1bffee88ef29d92f2401447a91f4), ROM_BIOS(0))
+
+	ROM_SYSTEM_BIOS(1, "alice_mcx128", "Darren Atkinson's MCX-128 cartridge for Alice")
+	ROMX_LOAD("alice128bas.rom", 0x0000, 0x4000, CRC(a737544a) SHA1(c8fd92705fc42deb6a0ffac6274e27fd61ecd4cc), ROM_BIOS(1))	
 ROM_END
```
cuavas, line 57: Since you know it’s an MCX-128, you can abbreviate the BIOS tags to just `ROM_SYSTEM_BIOS(0, "mc10", "Darren Atkinson's MCX-128 cartridge")` and `ROM_SYSTEM_BIOS(1, "alice", "Darren Atkinson's MCX-128 cartridge for Alice")` to make configuration more terse if you like.

src/devices/bus/mc10/mcx128.cpp
```diff
@@ -108,26 +103,30 @@ class mc10_pak_mcx128_device :
...
 	, device_mc10cart_interface(mconfig, *this)
 	, m_share(*this, "ext_ram", 1024*128, ENDIANNESS_BIG)
 	, m_view(*this, "mcx_view")
 	, m_bank(*this, "bank%u", 0U)
 {
+	const char *driver_name = mconfig.gamedrv().name;
+	if (strcmp(driver_name, "alice") == 0)
+	 	set_default_bios_tag("alice_mcx128");
+	else
+		set_default_bios_tag("mc10_mcx128");
```
cuavas, line 121: You can’t do this.  If you want to change the default BIOS in a particular system, you need to do it with `set_option_default_bios` on the slot when adding the slot to machine configuration.

See examples:
* https://github.com/mamedev/mame/blob/4183b96612eb588a76b271ca610604a607469fb7/src/devices/bus/abcbus/abc890.cpp#L78
* https://github.com/mamedev/mame/blob/4183b96612eb588a76b271ca610604a607469fb7/src/mame/pc/pc.cpp#L1115

src/devices/bus/mc10/multiports_ext.cpp
```diff
@@ -1,31 +1,34 @@
...
 
-        Only the RAM/ROM expansion is emulated here.
+		Only the RAM/ROM expansion is emulated here.
 
-    Banks are selected by writing to:
-        - $1000 to $1FFF in Cartridge mode (number of bank between 0 and 7)
-        - $C000 to $CFFF in ROM mode (number of bank between 0 and 3)
+	Banks are selected by writing to:
+		- $1000 to $1FFF in Cartridge mode (number of bank between 0 and 7)
+		- $C000 to $CFFF in ROM mode (number of bank between 0 and 3)
```
cuavas, line 23: This is going to end up getting changed back again by the `srcclean` tool – you should back it out to reduce churn (indentation inside multi-line comments only uses tabs up to the indentation level of the line containing the comment start delimiter).

src/devices/bus/mc10/multiports_ext.cpp
```diff
@@ -67,11 +69,7 @@ mc10_multiports_ext_device::mc10_multiports_ext_device(const machine_config &mco
 }
 
 mc10_multiports_ext_device::mc10_multiports_ext_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock)
-	: device_t(mconfig, type, tag, owner, clock)
-	, device_mc10cart_interface(mconfig, *this)
-	, m_bank(*this, "cart_bank")
-	, rom_bank_index(0)
-	, m_extention_ram(*this, "ext_ram", 1024 * 16, ENDIANNESS_BIG)
+	: device_t(mconfig, type, tag, owner, clock), device_mc10cart_interface(mconfig, *this), m_bank(*this, "cart_bank"), m_extention_ram(*this, "ext_ram", 1024 * 16, ENDIANNESS_BIG)
```
cuavas, line 72: Can you please put the line breaks back in?  It’s become excessively long.

cuavas (commented):
This mostly looks OK.  The one thing that *must* be fixed is using `set_option_default_bios` on the slot to set the default BIOS for slot cards per system rather than making assumptions about the system in the device.

src/devices/bus/mc10/mcx128.cpp
Mokona: I've looked at it, and it worked. Then found the default_bios() on the slot_option returned by the add_option() function. I used this.

PR #12092, opened 2024-03-02
Add cpu device PIC16x8x - Add PIC16F84 to magicle into misc/magicard.cpp

grullosgo:
Add cpu device PIC16x8x
Add PIC16F84 to magicle game into misc/magicard.cpp


src/devices/cpu/pic16x8x/pic16x8x.cpp
```diff
@@ -0,0 +1,1443 @@
...
+
+pic16x8x_device::pic16x8x_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock, int program_width, uint16_t eeprom_size, int pic_model)
+	: cpu_device(mconfig, type, tag, owner, clock)
+	, device_nvram_interface(mconfig, *this)
+	, m_region(*this, DEVICE_SELF)
+	, m_program_config("program", ENDIANNESS_LITTLE, 16, program_width, -1,
+			((program_width == 9) ? address_map_constructor(FUNC(pic16x8x_device::rom_9), this) :  address_map_constructor(FUNC(pic16x8x_device::rom_10),this)))
+	, m_data_config("data", ENDIANNESS_LITTLE, 8, 8, 0,
+			((pic_model == 0x83) ? address_map_constructor(FUNC(pic16x8x_device::ram_6), this) :
+			address_map_constructor(FUNC(pic16x8x_device::ram_7), this)))
```
ajrhacker, line 117: It's probably better just to pass the address map constructor here and eliminate the artificial `pic_model` constants.

src/devices/cpu/pic16x8x/pic16x8x.cpp
```diff
@@ -0,0 +1,1443 @@
...
+		{
+			fatalerror( "Region '%s' needs to be an 16-bit region\n", tag() );
+		}
+
+		// Get default NVRAM data from memory region
+		memcpy(m_buff, m_region->base() + eeprom_dump, m_internal_eeprom_size * 2);
+
+		// Data width conversion
+		for (u8 i=0; i < m_internal_eeprom_size; i++)
+			m_eeprom_data[i] = m_buff[i] & 0x00ff;
```
ajrhacker, line 205: This code doesn't seem endian-safe.

src/devices/cpu/pic16x8x/pic16x8x.cpp
```diff
@@ -0,0 +1,1443 @@
...
+#define T0CS_FLAG    0x20   // T0CS Timer 0 clock source select
+							// 1 = Transition on RA4/T0CKI pin
+							// 0 = Internal instruction cycle clock (CLKOUT)
+#define T0SE_FLAG    0x10   // T0SE Timer 0 clock source edge select
+							// 1 = Increment on high-to-low transition on RA4/T0CKI pin
+							// 0 = Increment on low-to-high transition on RA4/T0CKI pin
+#define PSA_FLAG     0x08   // PSA Prescaler Assignment bit
+							// 1 = Prescaler is assigned to the WDT
+							// 0 = Prescaler is assigned to the Timer0 module
+#define PS_REG       0x07   // PS Prescaler Rate select
```
ajrhacker, line 277: It's generally recommended to use `constexpr` instead of `#define` for constants in new code.

src/devices/cpu/pic16x8x/pic16x8x.cpp
```diff
@@ -0,0 +1,1443 @@
...
+	calc_zero_flag();
+}
+
+void pic16x8x_device::andwf()
+{
+	m_ALU = get_regfile(ADDR) & m_W;
+	store_result(ADDR, m_ALU);
+	calc_zero_flag();
+	if(ADDR == 0x0b)
+		logerror("ANDWF -> INTCON - data:%x\n", m_ALU);
```
ajrhacker, line 690: This seems to be a leftover from debugging. If you're running MAME with `-debug`, you can use `wpdset 0x0b, 1, w` to catch writes to this register.

src/devices/cpu/pic16x8x/pic16x8x.cpp
```diff
@@ -0,0 +1,1443 @@
...
+{
+	m_stack_pointer--;
+	u16 data = m_STACK[m_stack_pointer];
+	return data & m_program_mask;
+}
+
+void pic16x8x_device::push_stack(u16 data)
+{
+	m_STACK[m_stack_pointer] = data & m_program_mask;
+	m_stack_pointer++;
```
ajrhacker, line 417: It's better to check for stack overflow here. (I haven't double-checked the datasheet for this microcontroller, but PIC17 does recognize stack overflows, though it doesn't handle them very gracefully.)

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Since the rom load of this region is (and "must be") "ROM_REGION16_LE( 0x4280, "pic16f84", 0 )", I understand that the proposed code is correct.
The origin and destination data have been duly verified.
The memcpy() function loads u16 type data into m_buff from m_region, which in turn
have been read from the source in little_endian mode, therefore the conversion loop and its corresponding mask are correct.
Even so, if you consider that the code requires some change, I am completely willing to make it.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: working on this

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: will check the datasheet

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: As I can undertand, I guess I should split  the constructor  like pic17c4x code does. If I'm right please confirm.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: You're right. Will delete this and other two following logerrors calls.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Done the changes requested but missing the comments in the process. I'm sorry if I made some mistake.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Done. Deleted pic_model and and use program_width as selector.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Done.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Done. Added some datasheet notes and add stack overflow check.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Done. Deleted all leftovers.

src/devices/cpu/pic16x8x/pic16x8x.cpp
grullosgo: Second revision. Fixed.
 Added new base classes and instances for 16x83 and 16x84 to define properly the address maps.

grullosgo:
Hi @ajrhacker. 
I have completed all the items mentioned. Have I missed something for a second review? Should I click the "Resolve Conversation" button on each resolved comment? (There are small details about the PR's that I still don't know how to handle).

angelosa (approved):
Have seen this PR a couple times and have no direct feedback against it. Approving out of a later merging if there's no further feedback.

PR #12096, opened 2024-03-03
a2600.xml: Added 2 working items + 1 not working

ArcadeShadow:
Hobo [Atarimania]
R2D Tank [Atarimania]
Spiders [Atarimania]


cuavas:
It would really help if you listed which items were supposed to be working and which weren't.

PR #12103, opened 2024-03-04
capcom/vulgus.cpp: Add a 1942 bootleg from Itisa (not working)

clawgrip:
New not working clone
----------------------------
1942 (Itisa bootleg) [Recreativas.org]

### DIFF
--- a/src/mame/capcom/vulgus.cpp
+++ b/src/mame/capcom/vulgus.cpp
@@ -78,9 +78,10 @@ public:
 
 protected:
 	virtual void video_start() override;
+	void main_map(address_map &map);
+	required_device<cpu_device> m_maincpu;
 
 private:
-	required_device<cpu_device> m_maincpu;
 	required_device<cpu_device> m_audiocpu;
 	required_device<gfxdecode_device> m_gfxdecode;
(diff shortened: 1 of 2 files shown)


cuavas:
I haven’t had a chance to look at this, but it would be good if someone could take a look and see if the issues are something relatively simple.

@clawgrip indicated that it runs with bad graphics/colours and doesn’t start at the first level.

clawgrip:
> @clawgrip indicated that it runs with bad graphics/colours and doesn’t start at the first level.

After moving it from 1942.cpp to vulgus.cpp and some help from @Osso13 we have good colors and slightly better graphics, but still not good (and still the wrong start level).

Osso13:
Wrong start level I'm not sure, it seems to me 1942 starts with the same level.

FoxhackDN:
> Wrong start level I'm not sure, it seems to me 1942 starts with the same level.

Yep, the first stage says "Last 32 stage", then when you clear that, it says "Last 31 stage". The numbers should be correct.

clawgrip:
Yes, you're both right! Saw too many planes on screen, but was just the corrupted graphics :D

Issue #391, opened 2015-10-15
Convert drivers using PCI to latest MAME PCI interface.


andrew-gardner:
Seattle, Vegas, Model3, and more drivers hook their PCI devices together using a legacy method.  Convert all drivers with PCI to the hookup style of drivers/iteagle.c.

MoochMcGee:
This interface is incomplete for the purposes of at586, as there are devices required by at586 which need to be able to hook up custom registers in the configuration space. The new interface can't do this as far as I can tell.

galibert:
Huh?  The configuration memory map is virtual for that reason.  Look at
i82875p for instance, it adds a bunch of registers.

  OG.

angelosa:
Got an hard time understanding what's "Legacy" PCI, nowadays, my best guess is anything prefixed by L (lpci) therefore any class in [lpci.h](https://github.com/mamedev/mame/blob/master/src/devices/bus/lpci/pci.h) should be marked with the` [[deprecated("use non-legacy PCI device")]` marker?

angelosa:
Break down OP message with tasks, marked drivers as resolved where `#include "machine/lpci.h"` is not found with current tree.

angelosa:
Errata: there are THREE PCI bus implementations, the proper one is actually in [src/devices/machine/pci.cpp](https://github.com/mamedev/mame/blob/master/src/devices/machine/pci.cpp)

angelosa:
Added model3.cpp: does its own PCI bus implementation, which for all purposes fits here.

Hydreigon223:
Mind adding any drivers using the k033906.cpp PCI bridge device. The three drivers that use this are gticlub.cpp, nwk-tr.cpp and hornet.cpp.

angelosa:
Good call, added. Also I'll change directory entry over the new format later on.

angelosa:
Converted misc\comebaby.cpp to base PCI driver in PR #10830 , it wasn't doing anything useful to warrant using the legacy hookup. I'll deal with full conversion later since being i440bx.

angelosa:
Updated OP message with easy to track base chipsets for x86 based systems.

angelosa:
Updated OP with the (mostly) i440bx additions from [PR#11037](https://github.com/mamedev/mame/pull/11037).
As for the others:
- queen.cpp can be converted to stub PCI driver (as per voyager.cpp), unless anybody wants to tackle VIA architecture. That will be needed by calchase.cpp anyway;
[...]

PR #12105, opened 2024-03-05
cobra.cpp: attempted QA

Hydreigon223:
no comment

### DIFF
--- a/src/mame/konami/cobra.cpp
+++ b/src/mame/konami/cobra.cpp
@@ -68,7 +68,6 @@
     - proper PCI conversions
      * hook up MPC105 from main and gfx boards (to be added in mpc106.cpp)
     - hook up TMS57002
-    - figure out TMP47P241 usage and hook up if necessary
     - remove hacks
     - separate state classes between the two games
     - racjamdx: hook up main I/O and adc1038
--- a/src/mame/konami/cobra.cpp
+++ b/src/mame/konami/cobra.cpp
@@ -55,6 +55,24 @@
         Xilinx XC5204 FPGA
         Konami K001604 (2D tilemaps + 2x ROZ)
         Bt121KPC80 RAMDAC
+
+    TODO:
+    - timings are way off
+     * games run at incorrect speeds, bujutsu runs slow ingame and fast on character select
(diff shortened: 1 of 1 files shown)


src/mame/konami/cobra.cpp
```diff
@@ -55,6 +55,24 @@
...
+     * all games have 15 frames of input lag!
+     * bujutsu: ingame coin counter doesn't decrease when starting a game
+     * bgms have a tendency to skip parts occasionally
+     * occasional performance hits
+    - bujutsu: sometimes crashes
+    - racjamdx: segfaults on boot, error message mentions the 'render_texture_scan' command, CPU core bug?
+    - proper PCI conversions
+     * hook up MPC105 from main and gfx boards (to be added in mpc106.cpp)
+    - hook up TMS57002
+    - figure out TMP47P241 usage and hook up if necessary
```
ajrhacker, line 71: I don't think this can be hooked up now, since it's a 4-bit single-chip MCU with 2 KB of undumped internal OTP.

Hydreigon223: I'm going remove that. On a gfx board pic I have, the MCU is near a two-digit 7-seg LED so it likely drives that. This can easily be simulated and doesn't impact anything in terms of game logic (but quite helpful for diagnosing error codes).

angelosa:
Why this is a thing?
It's a PowerPC PCI HW, it's been listed in #391 alongside the other Konami HWs since forever, it needs a full refactor from the ground up, you can tell it's "not working" by the white rain renderer pulls off in one of the stages, I don't see any value here?

Hydreigon223:
I don't think there's any more point in keeping this open if everyone else seems to be too busy. Sorry if this stressed you out Kale.

angelosa:
Well you should've stopped at "MPC105 / MPC106" really. MPC105 is Bebox, MPC106 is Sega Model 3 (+ Power Macintosh), this goes 3rd place as (very) bare minimum.

rb6502:
There is a new-PCI MPC106 used by imacg3.cpp and powermacg3.cpp.  MPC105 and rewriting BeBox is on my list.

PR #12108, opened 2024-03-05
megasys1.cpp: added MCU dump for cybattlr and hook it up [kuze]

sergiopolog:
Added MCU dump for cybattlr. [kuze]
Also added and hooked it up for chimerab set which uses the same MCU, but marked as BAD_DUMP, as it only was dumped by kuze from a Cybattler PCB


happppp:
If the MCU has the same maskrom serial code, then you don't have to mark it bad dump.

sergiopolog:
> If the MCU has the same maskrom serial code, then you don't have to mark it bad dump.

Silkscreen on these Jaleco MCUs are always sanded and they have a sticker with a reference over it.
I never saw an original Chimera Beast pcb (it's a prototype), so I cannot ensure if the MCU on that board has the same sticker with same reference.

PR #12109, opened 2024-03-06
igs/goldstar.cpp: Added a 4th Animal House set and PLDs dumps for 'animalhsb'

clawgrip:
New not working clone
Animal House (set 4) [Taksangs, charlie]

Aso added PLDs dumps for 'animalhsb' [Taksangs, charlie]


cuavas:
It would be better if you commented explaining why this is a draft pull request.

clawgrip:
> It would be better if you commented explaining why this is a draft pull request.

Yes! We're currently dumping more Animal House sets, I'd rather compare them to add them all together.

PR #12112, opened 2024-03-08
netlist: fix missing timings and 74ls368

stonedDiscord:
You're in fact not allowed to omit them even if they're the same for all devices.
Ran into this now that I almost got the Monaco GP netlist working.
`../devices/nlid_truthtable.cpp:589: timing count not matching`


src/lib/netlist/macro/nlm_ttl74xx_lib.cpp
```diff
@@ -3470,26 +3474,12 @@ static TRUTH_TABLE(TTL_74279B, 4, 1, "")
...
-	TT_LINE("0,1,1,1,0|0,0,0,1|12")
-	TT_LINE("0,1,1,1,1|0,0,0,0|12")
+static TRUTH_TABLE(TTL_74368_GATE, 2, 1, "")
+{
+	TT_HEAD("OE,A|Y")
+	TT_LINE("1,X|0|12")
+	TT_LINE("0,1|0|12")
+	TT_LINE("0,0|1|12")
 	TT_FAMILY("74XX")
 }
```
cuavas, line 3484: Is this even correct?  Doesn’t the ’LS368 have 3-state outputs?  When the G input is high, the output should be high-Z.

stonedDiscord: Yes, it should be high-Z but the truthtable doesn't support high-Z and that's how the other chips handle it

cuavas: Well anything doing that needs a massive warning, because it’s going to cause no end of problems.  Also, treating high-Z as low is more likely to cause problems than treating it as high in TTL logic (which may be a good thing so people catch it sooner).

This chip in particular is likely to cause problems if it doesn’t implement high-Z outputs as the whole point of a tri-state buffer is so you can put it into a high-Z state.  If truthtables don’t support that, the chip isn’t a good candidate to implement that way.

Which “other chips” are treating high-Z as low?

stonedDiscord: 7425, 74125, 74126, 74368(this one),

ajrhacker: I also see this in src/lib/netlist/devices/nld_74365.cpp (which is for a close relative of 74368):

>  Note: Currently the netlist system does not support proper tristate output, so this
>        is not a "real" bus driver, it simply outputs 0 if the chip is not enabled.

PR #12113, opened 2024-03-08
machine/t10mmc: Fix MSF to LBA conversion for T10MMC_CMD_PLAY_AUDIO_MSF

987123879113:
This fixes audio being noticeably out of sync in pumpit1. The current dump of pumpit1 was thought to be a bad dump previously due to this bug, but T10MMC_CMD_PLAY_AUDIO_MSF had its own MSF to LBA conversion code which did not adjust the LBA value as shown in the formula below which was causing all songs played via the MSF playback command to be 2 seconds off.

ref: MMC-3 draft
image


src/devices/machine/t10mmc.cpp
```diff
@@ -961,9 +971,7 @@ void t10mmc::ReadData( uint8_t *data, int dataLength )
 						m_device->logerror("T10MMC: header data is not available for track type %d, inserting fake header data\n", track_type);
 
 						uint32_t msf = to_msf(m_lba);
-						data[data_idx] = BIT(msf, 16, 8);
-						data[data_idx+1] = BIT(msf, 8, 8);
-						data[data_idx+2] = BIT(msf, 0, 8);
+						put_u24be(&data[data_idx], msf);
```
987123879113, line 974: While auditing all usages of `to_msf` in t10mmc when I noticed this was the only one that didn't do lba+150. I checked a mode 2 CD image I had on hand and the first sector's timestamp starts at 00:02:00, so not having lba+150 here was a bug. The new `to_msf` adds the +150 so this should be fine now.

PR #12118, opened 2024-03-08
formats/dmk_dsk.cpp: Add structure checks in identify.

wilbertpol:
And be more lenient towards images where the amount of stored tracks does not match the amount of tracks set in the header.

### DIFF
--- a/src/lib/formats/dmk_dsk.cpp
+++ b/src/lib/formats/dmk_dsk.cpp
@@ -29,7 +29,7 @@ uint32_t wide_fm(uint16_t val)
 {
 	uint32_t res = 0;
 	for (int i = 15; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


ajrhacker:
I've thought about this subject since the discussion on #12062, and I've come to the conclusion that performing such "structural" checks on stored data is simply not warranted, especially due to the high priority of `FIFID_STRUCT`.

ajrhacker:
@galibert commented on #12062 that "Pure sector formats can't have a structural or a signature check in any case." I blame him for not making that clear when he added `FIFID_STRUCT`.

wilbertpol:
Right, then both FIFID_STRUCT and FIFID_SIGN are too strong for this format. I changed it to use FIFID_HINT.

PR #12122, opened 2024-03-10
bus/a2bus: - Add SNES MAX Super Nintendo Controller serial interface card.

ksherlock:
"The SNES MAX is a serial interface card that allows an Apple II to communicate with SNES Controllers."

https://lukazi.blogspot.com/2021/06/game-controller-snes-max-snes.html

The linked example code/schematics (https://docs.google.com/open?id=1MZvEiNi_7AO1DtoYO4dpHIPd7kykrG1q) includes some test programs and patched games (although tetris refuses to work for me).

```
mame apple2ee -sl3 snesmax -flop1 Robotron_SNESMAX.dsk
```

### DIFF
--- a/src/devices/bus/a2bus/snesmax.cpp
+++ b/src/devices/bus/a2bus/snesmax.cpp
@@ -1,0 +1,116 @@
+// license:BSD-3-Clause
+// copyright-holders:Kelvin Sherlock
+/*********************************************************************
+
+    snesmax.cpp
+
+    SNES controller serial interface card for the Apple II by Lukazi
+    https://lukazi.blogspot.com/2021/06/game-controller-snes-max-snes.html
+
+*********************************************************************/
+
+#include "emu.h"
+#include "snesmax.h"
+#include "bus/snes_ctrl/ctrl.h"
+
+namespace {
+
+//**************************************************************************
+//  TYPE DEFINITIONS
+//**************************************************************************
+
+class a2bus_snes_max_device:
+	public device_t,
+	public device_a2bus_card_interface
+{
+public:
+	// construction/destruction
+	a2bus_snes_max_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+protected:
+	a2bus_snes_max_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock);
+
+	virtual void device_start() override;
+	virtual void device_add_mconfig(machine_config &config) override;
+
+	virtual uint8_t read_c0nx(uint8_t offset) override;
+	virtual void write_c0nx(uint8_t offset, uint8_t data) override;
+
+	required_device<snes_control_port_device> m_ctrl1;
+	required_device<snes_control_port_device> m_ctrl2;
+
+	uint32_t m_latch1;
+	uint32_t m_latch2;
+};
+
+a2bus_snes_max_device::a2bus_snes_max_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+		a2bus_snes_max_device(mconfig, A2BUS_SNES_MAX, tag, owner, clock)
+{
+}
+
+a2bus_snes_max_device::a2bus_snes_max_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock) :
+		device_t(mconfig, type, tag, owner, clock),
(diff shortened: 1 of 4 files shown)


src/devices/bus/a2bus/snesmax.cpp
```diff
@@ -0,0 +1,116 @@
...
+	switch(offset & 0x01)
+	{
+	case 0:
+		/* latch */
+		m_ctrl1->port_poll();
+		m_ctrl2->port_poll();
+
+		/* bit 16 indicates if the controller is plugged in */
+		m_latch1 = m_ctrl1->get_card_device() ?  0x10000 : 0;
+		m_latch2 = m_ctrl2->get_card_device() ?  0x10000 : 0;
```
cuavas, line 91: How does this actually work?  The SNES controller doesn’t have a sense line as far as I know.

src/devices/bus/a2bus/snesmax.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+		for (int i = 0; i < 16; ++i)
+		{
+			m_latch1 |= (m_ctrl1->read_pin4() & 1) << i;
+			m_latch2 |= (m_ctrl2->read_pin4() & 1) << i;
+		}
```
cuavas, line 97: This won’t work.  You’re just reading the state of pin 4 repeatedly.  It won’t change if you don’t clock the shift register.

src/devices/bus/a2bus/snesmax.cpp
```diff
@@ -0,0 +1,116 @@
...
+		m_latch1 = ~m_latch1;
+		m_latch2 = ~m_latch2;
+		break;
+	case 1:
+		/* clock */
+		m_latch1 >>= 1;
+		m_latch2 >>= 1;
+		break;
```
cuavas, line 105: I suspect this is where it actually clocks the shift register, but how does it give the controller time to respond before reading the state of pin 4?

cuavas (changes requested):
This can’t possibly work.  The NES/SNES/A'Can controller interface looks like a shift register to the host.  There are outputs to latch the inputs and shift the output.  You aren’t doing either.  You’re just reading the state of the shift register output repeatedly without shifting it.  You need to generate the latch and shift outputs correctly and give the controller time to respond.

cuavas:
Looking at that schematic and reading the blog post, the guy doesn’t have a great understanding of electronics.

Essentially what it’s supposed to do:
* IC2A activates Y0 for C0nx write and Y1 for C0nx read.
* IC2B activates Y0 for C0nx write with A0 low and Y1 with A0 high.
* IC1A is being used as an over-complicated inverter.
* IC1B is also being used as a buffers/inverters.
* Controller 1 pin 4 output should appear on D7 for C0nx read.
* Controller 2 pin 4 output should appear on D0 for C0nx read.
* D5-D0 will read high.

There’s no “latch” on the card (the controller contains a latch) and nothing should happen in a loop.  Basically what you need to do:
* Generate a 1-cycle positive latch pulse for both controllers on C0nx writes with A0 clear (you’ll need a timer).
* Generate a 1-cycle negative shift pulse for both controllers on C0nx writes with A0 set (you’ll need another timer).
* Read the current pin 4 outputs for both controller on C0nx reads and put them in the most significant bits, inverted if necessary.  Make the other bits read high.

cuavas:
OK, after looking at the code, bus/snes_ctrl is in a dreadful state.  It should work like bus/nes_ctrl with the additional signals to support lightguns, but of course it doesn’t, it has line reads with side effects and other nonsense.

rb6502:
Yes, the real card borders on working by accident, and the blog post is nonsense.  US developers commonly used imported Super Famicom controllers on US and European SNES consoles because the much shorter cord takes up less desk space.

cuavas:
I added it in c53b36b892470df03616d3175245d4e465c9a35b – this is about the best that can be done given the current state of bus/snes_ctrl.

Bad things will happen if you try to use the Barcode Battler or Miracle Piano peripherals as they make assumptions about the host system.  Other peripherals may have issues as well.

cuavas:
> The linked example code/schematics (https://docs.google.com/open?id=1MZvEiNi_7AO1DtoYO4dpHIPd7kykrG1q) includes some test programs and patched games (although tetris refuses to work for me).

Tetris didn’t work for me, either.  Mario Bros and the test program do work.

ksherlock:
Thank you!

PR #12126, opened 2024-03-11
taito_h.cpp: some updates

sergiopolog:
- added notes about real functon of DSW2-7 and `0x430000` address on the memory map.
- updated default values for DSW2 - 5 to 7 as per shown in operator manual. Default player lives at '2' and DSW2-7 to 'ON' (TC0160ROM addressing mode as 1MB Mask ROM DIP28 pinout): [https://i.imgur.com/hEC4hMd.jpg](https://i.imgur.com/hEC4hMd.jpg)
- added note about flip-screen doesn't work on backgrounds.
- added flag to `syvalion` and clones marking they don't work fine in cocktail mode.

### DIFF
--- a/src/mame/taito/taito_h.cpp
+++ b/src/mame/taito/taito_h.cpp
@@ -29,8 +29,8 @@ System specs :
  Chips : TC0070RGB (RGB/Video Mixer)
          TC0220IOC (Input)
          TC0140SYT (Sound communication)
-         TC0130LNB (???)
-         TC0160ROM (???)
+         TC0130LNB (Graphics Data and palette addressing related)
+         TC0160ROM (Graphics ROM addressing related)
          TC0080VCO (Tilemap & Motion Object Gen)
 
  name             irq    resolution   tx-layer   tc0220ioc
@@ -52,6 +52,7 @@ Known issues :
  - Sprite zoom is a bit wrong.
  - Title screen of dynamite league is wrong a bit, which is mentioned by
    Yasuhiro Ogawa.
+ - Flip screen doesn't work fine on background layers
 
 
 Stephh's notes (based on the game M68000 code and some tests) :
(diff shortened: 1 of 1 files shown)


src/mame/taito/taito_h.cpp
```diff
@@ -52,6 +52,7 @@ Known issues :
  - Sprite zoom is a bit wrong.
  - Title screen of dynamite league is wrong a bit, which is mentioned by
    Yasuhiro Ogawa.
+ - Flip screen doesn't work fine on background layers
```
cuavas, line 55: What do you mean by “doesn’t work fine”?  It doesn’t work?  It doesn’t work at a fine (as opposed to coarse) level?

arcadez2003: You know perfectly well the users first language is not English hence sometimes there can be translation issues
dont be so elitest, it's attitudes like that which turn people away from wanting to helpout this project

wilbertpol: There is nothing elitist about asking for some more clarification when something is not clear.

arcadez2003: Well it's clearly a typo or a translation issue i suppose it should have read as such.......

- Flip screen doesn't work on background layers

There are ways to ask for clarification which dont come across as being possibly condesending

cuavas: It could mean:
* The layout of background tilemaps is flipped (works at a “coarse”) level but the tiles themselves are not flipped (doesn’t work at a “fine” level).
* It “works” but doesn’t “work fine” – the background layers are flipped but there’s some other issue (e.g. layer alignment).
* Background layers aren’t flipped at all.

I *don’t* know what it means – I haven’t tried running these games in cocktail cabinet mode.

It’s better to make the comments explicit and unambiguous, i.e. to describe the incorrect behaviour, both to help people unfamiliar with the system and to make it easier to tell whether the bug was fixed and the comment wasn’t updated or it’s a subtle bug you aren’t noticing immediately.

src/mame/taito/taito_h.cpp
```diff
@@ -354,12 +364,12 @@ static INPUT_PORTS_START( syvalionp )
...
 	PORT_DIPSETTING(    0x04, "2000k" )
 	PORT_DIPSETTING(    0x00, DEF_STR( None ) )
-	PORT_DIPNAME( 0x30, 0x30, DEF_STR( Lives ) )        PORT_DIPLOCATION("SW2:5,6")
+	PORT_DIPNAME( 0x30, 0x00, DEF_STR( Lives ) )        PORT_DIPLOCATION("SW2:5,6")
 	PORT_DIPSETTING(    0x00, "2" )
 	PORT_DIPSETTING(    0x30, "3" )
 	PORT_DIPSETTING(    0x20, "4" )
 	PORT_DIPSETTING(    0x10, "5" )
-	PORT_DIPNAME( 0x40, 0x40, DEF_STR( Unknown ) )      PORT_DIPLOCATION("SW2:7") /* code at 0x002af8 - see notes */
+	PORT_DIPNAME( 0x40, 0x00, DEF_STR( Unknown ) )      PORT_DIPLOCATION("SW2:7") /* code at 0x002af8 - see notes */
```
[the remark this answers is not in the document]
sergiopolog: Used `Graphics ROM Addressing Mode` as name and `JEDEC` and `Non-JEDEC` (default value) as option names.
This switch has no real effect in emulation

src/mame/taito/taito_h.cpp
sergiopolog: Updated notes to add more details  on flip screen issue for each game in the driver

PR #12127, opened 2024-03-11
Adding sound support to Model Racing Dribbling driver.

aovestdipaperino:
I followed the schematics component by component.
There might be some changes needed to fix the overall mixing but the netlist is done, done, done.


cuavas:
This won’t pass review.  Just from a quick look:
* You appear to have either pushed an old version of dribbling.h and dribbling.cpp over the top of the current versions, or split the code in a regressive way.  Please apply your changes to the latest version of the source and follow current practices.
* Just at a glance, the log messages you’ve added are not meaningful without reading the code, and can’t be enabled conditionally.  Please use the helpers from logmacro.h.
* The code to calculate times with global variables is not acceptable.  Why are you using host time anyway?  If you want meaningful timestamps for debugging you should be using emulated time from the scheduler.
* Adding empty comments to the ends of lines is unhelpful.
* nl_dribbling.h lacks `#include` guards.
* You want submodule netlists to be `static` so they aren’t globally visible.  Things that aren’t globally visible shouldn’t be declared `extern` in headers.
* There’s mixed space and tab indentation.

aovestdipaperino:
I fixed all the outstanding stuff, I can't see yet tabs vs spaces. Suggestions?
I commented one of the sounds (PARATA) as it doesn't perform well in real time.

src/mame/mr/dribling.cpp
```diff
@@ -34,32 +32,20 @@
...
-
-#define LOGMISC(...)     LOGMASKED(LOG_MISC,     __VA_ARGS__)
-#define LOGSOUND(...)    LOGMASKED(LOG_SOUND,    __VA_ARGS__)
-#define LOGPB(...)       LOGMASKED(LOG_PB,       __VA_ARGS__)
-
+#include "nl_dribling.h"
 
 namespace {
-
 class dribling_state : public driver_device
```
cuavas, line 49: You’re including two headers twice, you’ve lost the blank lines between header groups, you’ve lost the double blank line between separating the headers and logging helpers from the code, and you’ve lost the blank line between the start of the anonymous namespace and the class declaration.

src/mame/mr/dribling.cpp
```diff
@@ -94,6 +98,26 @@ class dribling_state : public driver_device
...
 
+	// sound
+	required_device<netlist_mame_logic_input_device>    m_i_pb0;
+	required_device<netlist_mame_logic_input_device>    m_i_pb1;
+	required_device<netlist_mame_logic_input_device>    m_i_pb2;
+	required_device<netlist_mame_logic_input_device>    m_i_pb3;
+	required_device<netlist_mame_logic_input_device>    m_i_pb4;
+	required_device<netlist_mame_logic_input_device>    m_i_pb5;
+	required_device<netlist_mame_logic_input_device>    m_i_pb6;
+	required_device<netlist_mame_logic_input_device>    m_i_pb7;
```
cuavas, line 109: It would be more concise if you used a `required_device_array` for these.

src/mame/mr/dribling.cpp
```diff
@@ -231,7 +255,6 @@ uint8_t dribling_state::input_mux0_r()
 }
 
 
-
 /*************************************
```
cuavas, line 258: You’ve unnecessarily removed a blank line here – it was using three blank lines to separate major sections.

src/mame/mr/dribling.cpp
```diff
@@ -246,46 +269,56 @@ void dribling_state::misc_w(uint8_t data)
...
 	// bit 2 = (9) = PC2
 	// bit 1 = (10) = PC1
 	// bit 0 = (32) = PC0
 	m_input_mux = data & 7;
-	LOGMISC("%s:misc_w(%02X)\n", machine().describe_context(), data);
-}
 
+}
 
 void dribling_state::sound_w(uint8_t data)
```
cuavas, line 291: Grouping is all messed up – there was one blank line between blocks in the function, and two blank lines between functions.

src/mame/mr/dribling.cpp
```diff
@@ -246,46 +269,56 @@ void dribling_state::misc_w(uint8_t data)
...
-	// bit 0 = folla b (crowd b)
-	LOGSOUND("%s:sound_w(%02X)\n", machine().describe_context(), data);
+	m_i_stop_palla->write_line(BIT(data, 7));
+	m_i_contrasto->write_line(BIT(data, 6));
+	m_i_calcio_a->write_line(BIT(data, 5));
+	m_i_fischio->write_line(BIT(data, 4));
+	m_i_calcio_b->write_line(BIT(data, 3));
+	m_i_folla_a->write_line(BIT(data, 2));
+	m_i_folla_m->write_line(BIT(data, 1));
+	m_i_folla_b->write_line(BIT(data, 0));
```
cuavas, line 300: This has lost the English translations from the comments.

src/mame/mr/dribling.cpp
```diff
@@ -246,46 +269,56 @@ void dribling_state::misc_w(uint8_t data)
...
+	m_i_pb7->write_line(BIT(data, 7));
 }
 
 
 void dribling_state::shr_w(uint8_t data)
 {
 	// bit 3 = watchdog
-	m_watchdog->reset_line_w(BIT(~data, 3));
+	if (data & 0x08)
+		m_watchdog->watchdog_reset();
```
cuavas, line 321: This is a change in behaviour.  Previously it only reset the watchdog when the state of bit 3 changed, now it resets it on any write with bit 3 set.  That’s less robust – are you sure this change is correct?

src/mame/mr/dribling.cpp
```diff
@@ -451,11 +484,35 @@ void dribling_state::dribling(machine_config &config)
 	screen.set_palette("palette");
 
 	PALETTE(config, "palette", FUNC(dribling_state::palette), 256);
-
 	// sound hardware
-}
+	SPEAKER(config, "mono").front_center();
```
cuavas, line 488: You’ve lost the blank line between blocks here.

src/mame/mr/dribling.cpp
```diff
@@ -451,11 +484,35 @@ void dribling_state::dribling(machine_config &config)
...
+	NETLIST_LOGIC_INPUT(config, "snd_nl:folla_m", "I_FOLLA_M.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:folla_a", "I_FOLLA_A.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:calcio_a","I_CALCIO_A.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:fischio",  "I_FISCHIO.IN",0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:calcio_b",  "I_CALCIO_B.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:contrasto", "I_CONTRASTO.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:stop_palla","I_STOP_PALLA.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:parata","I_PARATA.IN", 0);
+	NETLIST_LOGIC_INPUT(config, "snd_nl:enable",  "ENABLE_SOUND.IN",0);
+	NETLIST_STREAM_OUTPUT(config, "snd_nl:cout0", 0, "OUTPUT").set_mult_offset(1.0, 0);
```
cuavas, line 512: You should use object finders in machine configuration functions rather than explicit tags where possible

src/mame/mr/dribling.cpp
```diff
@@ -451,11 +484,35 @@ void dribling_state::dribling(machine_config &config)
...
 
 
+}
 
 /*************************************
```
cuavas, line 517: You should keep the three blank lines separating major sections and not add the extra blank lines between the last line in the function and the closing brace.

src/mame/mr/dribling.cpp
```diff
@@ -544,14 +601,15 @@ ROM_END
 
 } // anonymous namespace
 
-
 /*************************************
```
cuavas, line 604: This blank line doesn’t need to be removed.

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,494 @@
...
+	NET_C(R57.1, GND)
+	NET_C(C32.1, R58.2)
+	NET_C(R58.1, GND)
+}
+
+// STOP_PALLA AND PARATA are very similar. PARATA requires JFETs and 
+// MOSFET EMULATION doesn't work. So using the same circuit with different
+// values to emulate the sound.
+NETLIST_START(stop_palla_parata) 
+{
```
cuavas, line 130: This isn’t used directly outside the file – it should be `static`.

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,494 @@
...
+	NET_C(C24.1, GND)
+
+    // CONTRASTO
+	SUBMODEL(NE556_DIP, IC_Q9)		// VCO
+	NET_C(IC_Q9.7, GND)
+	NET_C(IC_Q9.14, I_V5)
+
+    RES(R38, RES_M(1))
+	RES(R39, RES_K(100))
+	CAP(C23, CAP_U(0.1))
```
cuavas, line 422: There’s a line indented with spaces here.

src/mame/mr/nl_dribling.h
```diff
@@ -0,0 +1,12 @@
+// license:BSD-3-Clause
+// copyright-holders: Enzo Lombardi
+
+#ifndef _NL_DRIBBLING_H
+#define _NL_DRIBBLING_H
```
cuavas, line 5: Please use the conventional form for `#include` guards.  Also, any name beginning with an underscore followed by an uppercase letter is reserved in both C and C++.

src/mame/mr/nl_dribling.h
```diff
@@ -0,0 +1,12 @@
...
+#include "netlist/nl_setup.h"
+
+NETLIST_EXTERNAL(LM339)
+NETLIST_EXTERNAL(calcio)
+NETLIST_EXTERNAL(stop_palla_parata)
+NETLIST_EXTERNAL(dribling)
```
cuavas, line 11: You don’t declare things `extern` when they are static.

src/mame/mr/dribling.cpp
```diff
@@ -544,14 +601,15 @@ ROM_END
...
-GAME( 1983, driblingbr, dribling, dribling, dribling, dribling_state, empty_init, ROT0, "bootleg (Videomac)",             "Dribbling (bootleg, Brazil)", MACHINE_NO_SOUND | MACHINE_SUPPORTS_SAVE )
+GAME( 1983, dribling,   0,        dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing",                   "Dribbling (set 1)",           MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE )
+GAME( 1983, driblinga,  dribling, dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing",                   "Dribbling (set 2)",           MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE )
+GAME( 1983, driblingo,  dribling, dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing (Olympia license)", "Dribbling (Olympia)",         MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE )
+GAME( 1983, driblingbr, dribling, dribling, dribling, dribling_state, empty_init, ROT0, "bootleg (Videomac)",             "Dribbling (bootleg, Brazil)", MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE )
+
+
```
cuavas, line 615: You’ve added two trailing blank lines here.

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,494 @@
...
+	RES(R_OUT, RES_K(2.2))
+	RES(R_PU_OUT, RES_K(10))
+	NET_C(R_PU_OUT.1, I_V5)
+	NET_C(ENABLE_SOUND, R_OUT.1, R_PU_OUT.2)
+	NET_C(Q_OUT.B, R_OUT.2)
+	NET_C(Q_OUT.C, R_TOS.2, R_FISCHIO.2, R_FOLLA.2, R_CALCIO_A.2, R_CALCIO_B.2, R_STOP_PALLA.2, R_PARATA.2)
+	NET_C(Q_OUT.E, GND)
+
+	ALIAS(OUTPUT, Q_OUT.C)
+}
```
cuavas, line 494: This file is missing the ending newline.

src/mame/mr/nl_dribling.h
```diff
@@ -0,0 +1,12 @@
...
+#endif
```
cuavas, line 12: This file is missing the ending newline.

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,494 @@
+// license:BSD-3-Clause
+// copyright-holders: Enzo Lombardi
```
cuavas, line 2: Is there a reason for using the BSD license?  The content of a netlist is not eligible for copyright protection (it doesn’t qualify as “creative”), so the only parts that are eligible for copyright protection are things like the formatting.  It’s usually less trouble to just use CC0-1.0 for things with minimal “creative” content.

cuavas (changes requested):
Not all the issues listed have been resolved, and there are still a lot of editing errors.

src/mame/mr/dribling.cpp
aovestdipaperino: I see "screen.h" being included twice. But that's the only one. I am sorting them alphabetically to make sure (if sort order is irrelevant).

src/mame/mr/dribling.cpp
aovestdipaperino: Fixed

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/nl_dribling.h
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/nl_dribling.cpp
aovestdipaperino: done

src/mame/mr/nl_dribling.cpp
aovestdipaperino: Done

aovestdipaperino:
I should have addressed everything except for the second double include in dribling.cpp which I couldn't find.
Thanks for your patience, looking for more feedback or getting this in.
Still planning to work on the remaining sound (PARATA) after I am able to build this on my Mac.

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: I will take a look at the next round, I've used existing code as guide.

src/mame/mr/nl_dribling.cpp
aovestdipaperino: If I declare them static I get this:
```
In file included from ../../../../../src/mame/mr/nl_dribling.cpp:3:
../../../../../src/lib/netlist/nl_setup.h:61:28: error: 'void netlist_LM339(netlist::nlparse_t&)' was declared 'extern' and later 'static' [-fpermissive]
   61 | #define NETLIST_NAME(name) netlist##_##name
```
even if I have removed them the NETLIST_EXTERNAL references.

src/mame/mr/nl_dribling.h
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: done

src/mame/mr/dribling.cpp
aovestdipaperino: Done

src/mame/mr/dribling.cpp
aovestdipaperino: Done

aovestdipaperino:
This was the last fix on the netlist as all the sounds work.

src/mame/mr/nl_dribling.cpp
aovestdipaperino: ah, found it. Fixed.

src/mame/mr/dribling.cpp
aovestdipaperino: Double checked, no duplicates.

aovestdipaperino:
(squashed all commits into a single meaningful one)

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,524 @@
...
+	LM324_DIP(IC)
+	ALIAS(1, IC.3)
+	ALIAS(2, IC.2)
+	ALIAS(3, IC.4)
+	ALIAS(4, IC.11)
+	ALIAS(5, IC.1)
+	ALIAS(GND, IC.11)
+	NET_C(GND, IC.5, IC.6, IC.10, IC.9, IC.12, IC.13)
+}
+#else
```
cuavas, line 44: What’s the intention of replacing the LM339 with an LM324?

The LM339 is a quad comparator – it’s designed to implement simple voltage comparison, and has single-ended open collector output (i.e. it can only sink current, not source current).

The LM324 is a quad operational amplifier – it’s designed for linear amplification and has push/pull outputs (i.e. it can source as well as sink current).

The two chips aren’t equivalent at all, and will behave very differently.

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,524 @@
...
+	NET_C(NAND.8, AND.1)
+	NET_C(IC_C7.5, AND.2)
+	NET_C(GND.Q, AND.4, AND.5, AND.9, AND.10, AND.12, AND.13)
+	NET_C(GND, AND.7)
+	NET_C(I_V5, AND.14)
+	ALIAS(TOS, AND.3)
+#else
+	// Use original flip-flop output as from schematics.
+	ALIAS(TOS, IC_C7.5)
+#endif
```
cuavas, line 280: It’s pretty common for these analog sound systems to produce high-frequency idle sound that won’t make it through the audio outputs (e.g. listen to Cheeky Mouse in MAME, `cheekyms`).  Is it really a good idea to add fake circuitry like this?

src/mame/mr/nl_dribling.cpp
```diff
@@ -0,0 +1,524 @@
...
+	NET_C(R57.1, GND)
+	NET_C(C32.1, R58.2)
+	NET_C(R58.1, GND)
+}
+
+// STOP_PALLA AND PARATA are very similar. PARATA requires JFETs and 
+// MOSFET EMULATION doesn't work. So using the same circuit with different
+// values to emulate the sound.
+static NETLIST_START(stop_palla_parata) 
+{
```
cuavas, line 150: What are the JFETs being used for?  What kind of MOSFETs did you use to try simulating them?

src/mame/mr/dribling.cpp
cuavas: No it isn’t.  You’ve still left extra blank lines *inside* the function while removing blank lines

src/mame/mr/nl_dribling.cpp
balr0g: I would argue that such bandpass filtering, if desired, should be handled outside the netlist subsystem (e.g. in the audio output pipeline, after any/all effects have been processed), and have an on/off setting toggle. It definitely does not belong in the netlist, unless it was present on original hardware.

aovestdipaperino: Yeah, I've tried that leaving the sound on but it's unbearable since the output pipeline doesn't help here. The square wave of the 40Khz clock leaves too many harmonics out.
Consider also that the schematic uses a TDA2003 for final amplification (and pass-band I guess) that's not available. 
If there are better alternatives, I more than willing to pursue them.

src/mame/mr/nl_dribling.cpp
aovestdipaperino: That's a good question and the circuit simulator seems to show that the effect is to clip below 2.5V to output, meaning let only the portion above it pass through.
I've used the suggestion in the netlist FAQ used all over JFETs are required but it would not produce results (a bunch of NaN).
The current solution though comes pretty close given the schematics being identical except for the output.

src/mame/mr/nl_dribling.cpp
aovestdipaperino: The purpose of the LM339 is to output a square wave. I explained how the circuit works [here](https://medium.com/@enzo-lombardi/sounding-dribbling-episode-two-769fed9a7d8f). So for the purpose of squaring the wave, the LM324 is a great fit. I simulated both circuits and the difference are negligible.
Considering that a more accurate LM339 emulation is not an option for performance reason, I would say that taking the freedom to replace it with LM324 is ok since the OpAmp does go into saturation. Emulation in analog devices seems pretty common in netlist base code.

src/mame/mr/dribling.cpp
```diff
@@ -34,29 +32,17 @@
...
-#define LOG_SOUND    (1U << 2)
-#define LOG_PB       (1U << 3)
-
-//#define VERBOSE (LOG_GENERAL | LOG_MISC | LOG_SOUND | LOG_PB)
-
-#include "logmacro.h"
-
-#define LOGMISC(...)     LOGMASKED(LOG_MISC,     __VA_ARGS__)
-#define LOGSOUND(...)    LOGMASKED(LOG_SOUND,    __VA_ARGS__)
-#define LOGPB(...)       LOGMASKED(LOG_PB,       __VA_ARGS__)
```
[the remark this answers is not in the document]
aovestdipaperino: I removed the logging because the purpose was likely to log the events in order to implement the netlist. And I used that logging a lot to generate input files for ``nltool`` to run later. But I genuinely thought that they were put there only to help implement the netlist.

src/mame/mr/dribling.cpp
```diff
@@ -34,29 +32,17 @@
...
+#include "machine/netlist.h"
+#include "netlist/devices/net_lib.h"
 #include "cpu/z80/z80.h"
 #include "machine/i8255.h"
-#include "machine/watchdog.h"
-
 #include "emupal.h"
+#include "speaker.h"
+#include "machine/watchdog.h"
 #include "screen.h"
```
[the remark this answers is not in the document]
aovestdipaperino: will do. From the diffs the only one moved seems to be the ``watchdog`` one.

src/mame/mr/dribling.cpp
```diff
@@ -246,39 +257,56 @@ void dribling_state::misc_w(uint8_t data)
...
 }
 
 
 void dribling_state::pb_w(uint8_t data)
 {
-	// write PB0-7
-	LOGPB("%s:pb_w(%02X)\n", machine().describe_context(), data);
+	for (int i=0; i < 8; i++) {
+		m_i_pb.at(i)->write_line(BIT(data, i));
+	}
```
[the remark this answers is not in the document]
aovestdipaperino: will do.

aovestdipaperino:
FIxed the fixable outstanding issues.
Open to suggestions for more.

aovestdipaperino:
(one more squash of all CR commits)

aovestdipaperino:
Video demo of all the sounds: https://www.youtube.com/watch?v=t3chYXHCzG0

simzy39:
In your video, the players aren't making sounds when they move.
This is a video of real hardware: https://www.youtube.com/watch?v=BnkXawBCuvE
Watch from 1:22, and you'll seee that the players make little pacman-esque sounds when moving

aovestdipaperino:
> In your video, the players aren't making sounds when they move. This is a video of real hardware: https://www.youtube.com/watch?v=BnkXawBCuvE Watch from 1:22, and you'll seee that the players make little pacman-esque sounds when moving

Thanks for the catch.
My theory is that the available schematics are missing this or it's described in a different section from all the sounds. I've added logging statements to the code and there's no signal to any of the known netlists to switch either direction.
I suspected it was a different ROM but I've tried all available ones and they all behave the same: no sound signals.

I have access to a cabinet and double check, my memory tells me that that sound does come out.

aovestdipaperino:
> > In your video, the players aren't making sounds when they move. This is a video of real hardware: https://www.youtube.com/watch?v=BnkXawBCuvE Watch from 1:22, and you'll seee that the players make little pacman-esque sounds when moving
> 
> Thanks for the catch. My theory is that the available schematics are missing this or it's described in a different section from all the sounds. I've added logging statements to the code and there's no signal to any of the known netlists to switch either direction. I suspected it was a different ROM but I've tried all available ones and they all behave the same: no sound signals.
> 
> I have access to a cabinet and double check, my memory tells me that that sound does come out.

This[ is a video of my friends playing the game on real hardware,](https://youtu.be/RwsIru-HWuI) no sound while players move around. I am suspecting that the other video might have been a special board.

src/mame/mr/dribling.cpp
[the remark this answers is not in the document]
cuavas: The netlist isn’t fully implemented at this point, and configurable logging is harmless to leave in if it may help someone debugging in the future.

src/mame/mr/nl_dribling.cpp
cuavas: No, an LM324 is *not* a great fit.  It will not work the same way at all.

Looking at the “STOP PALLA” circuit:
* With an LM339, the timing capacitor will charge via the 2.2kΩ and 10kΩ resistors in series an the LM339 has no high driver output.
* With an LM324, the timing capacitor will charge via the 10kΩ resistor only, the 2.2kΩ resistor will be effectively short-circuited by the high driver output of the LM324.

This will affect the frequency and symmetry of waveform.

An LM324 is also not a suitable replacement in the “PARATA” circuit, for more complex reasons.

If the LM339 model you used is too slow, you need a more efficient

cuavas: If the LM339 model you used has performance issues, the solution is to make a simplified LM339 model.  Note that the op-amps don’t use transistor-level models, they use high-level models found in src/lib/netlist/analog/nld_opamps.cpp

The key thing is that the LM339 **is not an op-amp**.  It’s a voltage comparator with a single-ended output.  Replacing it with an op-amp is not going to work.

src/mame/mr/nl_dribling.cpp
cuavas: OK, I looked at the schematic.  It definitely isn’t “identical except for the output stage.”  The first JFET is part of the function generator.

The first JFET has its source and gate tied together on the LM339 output:
* When the LM339 is pulling its output low, the JFET will operate in constant current mode.  The timing capacitor will discharge via the JFET and the LM339 output.  The discharge curve will be a linear ramp.
* When the LM339 is not pulling its output low, the gate-drain region of the JFET will be forward-biased, so it will present a low impedance in series with the 2.2kΩ resistor.  The capacitor will charge as though it’s effectively connected to the 2.2kΩ resistor directly.  The charge curve will be a normal RC charge curve.

The second JFET is connected in source follower configuration.  It presents a high-impedance input and low-impedance output.  This prevents the mixing stage from interfering with the function generator.

The circuit is supposed to generate a kind of sawtooth waveform, rising with a relatively quick RC charge curve, and falling with a linear ramp.  You’ll need to use *something* to substitute for the JFETs.

Once again, this won’t work with an op-amp.  With an op-amp, the 2.2kΩ resistor will be effectively shorted when the op-amp is driving its output high.  This will mean the capacitor will charge far quicker than it should.

Please answer this question:

>  What kind of MOSFETs did you use to try simulating them?

Did you substitute depletion mode MOSFETs with appropriate Vgs=0 current?  Did you try using a depletion mode MOSFET with a diode connected across the drain and source for the first JFET?

cuavas:
Please stop marking conversations as resolved when they aren’t just because you don’t like the answers.

simzy39:
Please open the PR again. You've done a great job, and we do want this game to have sound.
You just might need to do a little more work to get it to Mame's standards. Also, sometimes it takes a while to get a PR ready or reviewed too. The people reviewing your PR are using their free time to do it.

rb6502:
I think the issue was performance when using the correct parts in the circuit, but it's unclear what that means in numbers.  If it's too slow for the submitter's system but in range for higher-end hardware that's usable.  If it's "no system will ever run this with sound that way" then we might want to #ifdef the correct implementation and tag it IMPERFECT_SOUND.

cuavas:
I’m pretty sure simplified models with adequate performance are viable – it doesn’t push the LM339 or JFETs to their limits.  It’s just a matter of deciding exactly how to do that.  But it’s hard to do that if conversations keep getting marked resolved and hidden.

cuavas:
> I think the issue was performance when using the correct parts in the circuit, but it's unclear what that means in numbers. If it's too slow for the submitter's system but in range for higher-end hardware that's usable.

The transistor-level model of an LM339 is probably too slow even for a high-end system.  There’s a reason we don’t use transistor-level models of op-amps, either.

It’s probably good enough to just hack up a model with a few resistors representing the input impedance, and a voltage-controlled current source representing the output transistor.  It’s not like the circuit will get anywhere near the LM339’s frequency response limits, and the output transistor will be in cutoff or saturation all the time anyway.

aovestdipaperino:
I am opening up the PR (I will try or open a new one and link to here) but please provide guidance on:
**LM339:**

1. the best option would be to create a comparator model in C++, but that's way beyond my understanding of how the netlist code works. If anyone can help, from a layman perspective it seems a much simpler circuit than the Schmitt Trigger, I will more than happy to help. **I don't have the skills to help here**
2. second best option is to use a type 1 OpAmp (see [here](https://github.com/mamedev/mame/blob/44744e93cc1267131410e8a76aa18f054564aa4b/src/lib/netlist/analog/nld_opamps.cpp#L51); idealized OpAmps are close to comparators. **I can help with this.**
3. third best, is to use LM324. I can provide MicroCap charts and wavs showing how close these are; at least for the time being it could be acceptable until a 1. is done; **this is what is already done.**
4. any other solution, **I don't have the skills to do this**

>  It’s not like the circuit will get anywhere near the LM339’s frequency response limits, and the output transistor will be in cutoff or saturation all the time anyway.

Exactly. That is why an OpAmp (idealized?) is a good approximation.

**JFET:**
as I mentioned above, I used the exact suggestion presented in the FAQ (linking it again [here](https://github.com/mamedev/mame/blob/44744e93cc1267131410e8a76aa18f054564aa4b/src/lib/netlist/FAQ.md?plain=1#L81) ). It suggests to use ``MOSFET(Q21, "NMOS(VTO=-1.0)")``

> Did you substitute depletion mode MOSFETs with appropriate Vgs=0 current? Did you try using a depletion mode MOSFET with a diode connected across the drain and source for the first JFET?

Nope, I can try.
As for a stopgap in case, the current netlist is pretty close from a sound perspective (compare the videos provided to see what I mean).

So can I get help and guidance? **this is close to my skill border**

**Using named objects:** I don't understand what it means, a link with an example would suffice. Keeping things as is it's an option as well since it's only 9 sounds total.

**High pitch removal additional circuitry**: let me know what you prefer, I can only comment that either we keep it (with a #define for now) or we add an OpAmp 1x amplifier to act as cut-off filter. My quick attempts at the latter haven't produced results as clean as the logic circuitry.

Logs: I agree to disagree but I am totally fine in putting them back as they don't hurt performance anyway.

Sorry if I closed comments that seemed to be just questions on why had I taken some choices, I believe the github threaded model doesn't help with this specific type of discussion.

aovestdipaperino:
> The transistor-level model of an LM339 is probably too slow even for a high-end system. There’s a reason we don’t use transistor-level models of op-amps, either.
Exactly. I consider a Mac running an M2 quite high-end, and of course too many transistors for expecting real time.

One more thing: the STOP_PALLA (and PARATA I guess) require an higher resolution since the code switches the circuitry off too fast for the netlist code to realized that there was an event. They can be moved to a different netlist, but it won't help anyway since they require the LM339.

aovestdipaperino:
I don't seem to be able to re-open this pull-request.
Please let's move the discussion over there.

PR #12132, opened 2024-03-12
igs/pgm2.cpp: Add the program ROM of "Bu Bu Car"

clawgrip:
New not working machine
-----------------------------
Bu Bu Car (English) [Victor Fernandez (City Game), Recreativas.org]

We haven't desoldered and dumped the surface mounted flash ROMs because we'd better wait for the program ROM to be decrypted (by anyone able of doing it) so we can then run a trojan and extract the CPU internal ROM.

### DIFF
--- a/src/mame/igs/pgm2.cpp
+++ b/src/mame/igs/pgm2.cpp
@@ -750,7 +750,7 @@ GFXDECODE_END
 
 void pgm2_state::pgm2(machine_config &config)
 {
-	/* basic machine hardware */
+	// basic machine hardware
 	IGS036(config, m_maincpu, 100000000); // Unknown clock / divider
 	m_maincpu->set_addrmap(AS_PROGRAM, &pgm2_state::pgm2_rom_map);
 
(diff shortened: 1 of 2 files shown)


mamehaze:
as far as I know only Andreas was able to decrypt these

clawgrip:
> it'd be better if all the ROMs could be dumped, anyway, You never know when someone comes around that can work on this and it'd be better if they had everything ready if / when the time comes.

Yes, but soldering them back to run a trojan (SMD) is not an easy task, and without the CPU internal ROM any progress is very limited.

mamehaze:
nobody is going to run a trojan that way on these things.

the other ROMs were extracted using an FPGA based solution acting as a ROM emulator, changing the code in realtime, and using memory read offsets to 'send' back data to the FPGA.

that's the only realistic way to dump the ROMs, and requires Peter / Morten level of skill, although as noted, first the program ROM needs to be decrypted.  None of those people are involved with MAME anymore for various reasons.

clawgrip:
Ooook, We'll dump the missing flash chips ASAP!

PR #12136, opened 2024-03-14
util/cdrom.cpp: Replace stdio usage with core_file.

### LINKED ISSUE #12095: chdman createcd doesn't work with input path containing accents on Windows
### MAME version

0.263

### System information

Windows 11 in spanish

### INI configuration details

```INI
N/A
```

### Emulated system/software

N/A

### Incorrect behaviour

Create a folder with accents and put the ".iso" or ".bin/.cue" inside, for example:
D:\Té

And use the path in the "-i" argument:
```
>chdman createcd -i "D:\Té\Something.iso" -o "Something.chd"
chdman - MAME Compressed Hunks of Data (CHD) manager 0.263 (mame0263)
Error parsing input file (D:\T├®\Something.iso: No such file or directory)
```

### Expected behaviour

It should read "D:\Té\Something.iso"

### Steps to reproduce

1. Create a directory with accents, like "Emulación"
2. Put an iso/bin+cue inside this folder
3. Try to use createcd adding the folder to the "-i" argument 

### Additional details

The -o argument works well with accents.

"createdvd" shows strange symbols to the console, but it works (reads and generates the output):
```
> chdman.exe createdvd -i D:\Té\Something.iso -o D:\Té\Test.chd
chdman - MAME Compressed Hunks of Data (CHD) manager 0.263 (mame0263)
Output CHD:   D:\T├®\Test.chd
Input file:   D:\T├®\Something.iso
Compression:  lzma (LZMA), zlib (Deflate), huff (Huffman), flac (FLAC)
Logical size: 211,288,064
Compressing, 24.5% complete... (ratio=56.4%)
```

invertego:
Fixes #12095

This is mostly 1:1 replacement, but having CUE file parsing go through `core_text_file::getc` may expose some differences in behavior vs. the user's standard C library.

### DIFF
--- a/src/lib/util/cdrom.cpp
+++ b/src/lib/util/cdrom.cpp
@@ -18,6 +18,7 @@
 
 #include "cdrom.h"
 
+#include "corefile.h"
 #include "corestr.h"
 #include "multibyte.h"
 #include "osdfile.h"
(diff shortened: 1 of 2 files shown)


cuavas:
I’m going to have to say no on this for now, for two reasons:
* There are people doing ongoing work on util/cdrom.cpp, and this would interfere with it.
* The “text mode” features of `util::core_file` are in a pretty dreadful state to begin with.

I realise it’s a real issue, but people will have to work around it for a while longer.  Also note that all the Lua scripts that use LFS (Lua File System) and/or standard Lua file objects have issues when paths can’t be represented in the current code page.  There are a few other things that use `std::ifstream` and `std::ofstream` in the codebase as well.

I was planning to get rid of the stuff using `std::ofstream` sooner by writing a stream buffer implementation that uses the stuff from ioprocs.h as a backend.  That would also allow the text mode writing features to be removed from `util::core_file`.  Dealing with the stuff that needs to *read* text files is more complicated.

PR #12137, opened 2024-03-14
casio/cfx9850.cpp: Various fixes

qufb:
- Fix CFX9850GB `display_ram` mapping, this one doesn't write to segment 0x60;
- Debugger edits to register values were all being incorrectly applied to the first register, now they should go to the corresponding set of 4 registers;
- Implement TIME register;
- Add more instructions tested on hardware;
- Fix swap flags, they are always cleared regardless of values;
- Fix timer so that it expires on KO enabled key input;
- Add an infinite timer variant, which seems to only be used along with the low power variant of `timer_wait`;
- Prefer more accurate palette taken from an [online Casio manual](https://support.casio.com/storage/en/manual/pdf/EN/004/fx_plus_ch_intro_EN.pdf), although it isn't clear what contrast levels they used for those screenshots;

### DIFF
--- a/src/devices/cpu/hcd62121/hcd62121.cpp
+++ b/src/devices/cpu/hcd62121/hcd62121.cpp
@@ -39,7 +39,7 @@ enum
 	HCD62121_R70, HCD62121_R74, HCD62121_R78, HCD62121_R7C
 };
 
-
+// TODO - Max value stored with "movb reg,f" is 0x3f, is bit 5 set in other instructions?
 constexpr u8 FLAG_CL = 0x10;
 constexpr u8 FLAG_Z = 0x08;
 constexpr u8 FLAG_C = 0x04;
@@ -62,11 +62,15 @@ hcd62121_cpu_device::hcd62121_cpu_device(const machine_config &mconfig, const ch
 	, m_sseg(0)
 	, m_f(0)
 	, m_time(0)
+	, m_time_op(0)
+	, m_is_timer_started(false)
+	, m_is_infinite_timeout(false)
 	, m_lar(0)
 	, m_opt(0)
 	, m_port(0)
 	, m_program(nullptr)
 	, m_icount(0)
+	, m_cycles_until_timeout(0)
 	, m_kol_cb(*this)
 	, m_koh_cb(*this)
 	, m_port_cb(*this)
@@ -81,6 +85,14 @@ device_memory_interface::space_config_vector hcd62121_cpu_device::memory_space_c
 	return space_config_vector {
 		std::make_pair(AS_PROGRAM, &m_program_config)
 	};
+}
+
+TIMER_CALLBACK_MEMBER(hcd62121_cpu_device::timer_tick)
+{
+	// TODO - Only stores seconds? How can it stop/reset?
+	if (m_is_timer_started) {
+		m_time = (m_time + 1) % 60;
+	}
 }
 
 u8 hcd62121_cpu_device::read_op()
@@ -136,6 +148,33 @@ void hcd62121_cpu_device::write_reg(int size, u8 op1)
 			m_reg[(op1 + i) & 0x7f] = m_temp1[i];
 	}
 }
+
+
+void hcd62121_cpu_device::read_ireg(int size, u8 op1)
+{
+	u16 ad = m_reg[(0x40 | op1) & 0x7f ] | (m_reg[(0x40 | (op1 + 1)) & 0x7f] << 8);
+
+	for (int i = 0; i < size; i++)
+	{
+		m_temp1[i] = m_program->read_byte((m_dseg << 16) | ad);
+		ad += (op1 & 0x40) ? -1 : 1;
+	}
+	m_lar = ad;
+}
+
+
+void hcd62121_cpu_device::write_ireg(int size, u8 op1)
+{
+	u16 ad = m_reg[(0x40 | op1) & 0x7f] | (m_reg[(0x40 | (op1 + 1)) & 0x7f] << 8);
+
+	for (int i = 0; i < size; i++)
+	{
+		m_program->write_byte((m_dseg << 16) | ad, m_temp1[i]);
+		ad += (op1 & 0x40) ? -1 : 1;
+	}
+	m_lar = ad;
+}
+
 
 void hcd62121_cpu_device::read_regreg(int size, u8 op1, u8 op2, bool copy_extend_immediate)
 {
@@ -307,6 +346,9 @@ void hcd62121_cpu_device::device_start()
 {
 	m_program = &space(AS_PROGRAM);
 
+	m_timer = timer_alloc(FUNC(hcd62121_cpu_device::timer_tick), this);
+	m_timer->adjust(attotime::from_seconds(1), 0, attotime::from_seconds(1));
+
 	save_item(NAME(m_prev_pc));
 	save_item(NAME(m_sp));
 	save_item(NAME(m_ip));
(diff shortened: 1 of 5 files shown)


src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -810,6 +857,20 @@ void hcd62121_cpu_device::execute_run()
...
+		} else if (m_is_infinite_timeout) {
+			m_icount = 0;
+		} else if (m_cycles_until_timeout > 0) {
+			int cycles_to_consume = std::min(m_cycles_until_timeout, m_icount);
+			m_cycles_until_timeout -= cycles_to_consume;
+			m_icount -= cycles_to_consume;
+		}
+		if (m_icount <= 0) {
+			break;
+		}
```
cuavas, line 872: This should also use Allman formatting.

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -1152,6 +1213,87 @@ void hcd62121_cpu_device::execute_run()
...
+				}
+				else
+				{
+					read_ireg(size, reg1);
+					for (int i = size-1; i > 0; i--) {
+						m_temp1[i] = m_temp1[i - 1];
+					}
+					m_temp1[0] = 0;
+					write_ireg(size, reg1);
+				}
```
cuavas, line 1241: Same here, please use Allman formatting for consistency with the rest of the file.

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -1761,18 +2002,19 @@ void hcd62121_cpu_device::execute_run()
...
 
 		case 0xFC:      /* unk_FC - disable interrupts/stop timer?? */
-		case 0xFD:      /* unk_FD */
 			logerror("%02x:%04x: unimplemented instruction %02x encountered\n", m_cseg, m_ip-1, op);
 			break;
 
+		case 0xFD:      /* timer_wait_low (no X1 clock, address or data bus activity) */
 		case 0xFE:      /* timer_wait */
-			if (m_time & 0x01) {
+			if (m_time_op & 0x01) {
```
cuavas, line 2017: While you’re working on it, can you reformat this to Allman style as well?  I think you added it as K&R last time.

src/devices/cpu/hcd62121/hcd62121.h
```diff
@@ -92,6 +95,10 @@ class hcd62121_cpu_device :  public cpu_device
 	u8 m_sseg;
 	u8 m_f;
 	u8 m_time;
+	u8 m_time_op;
+	bool m_is_timer_started;
+	bool m_is_infinite_timeout;
+	emu_timer *m_timer;
```
cuavas, line 101: You haven’t added the new data members to save states.

src/devices/cpu/hcd62121/hcd62121.h
```diff
@@ -108,6 +115,7 @@ class hcd62121_cpu_device :  public cpu_device
 	address_space *m_program;
 
 	int m_icount;
+	int m_cycles_until_timeout;
```
cuavas, line 118: This needs to be an explicitly sized type because it will need to be included in save states.  Also, don’t keep it with `m_icount` which is managed by the scheduler – put it with the other data members related to managing this.

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -307,6 +346,9 @@ void hcd62121_cpu_device::device_start()
 {
 	m_program = &space(AS_PROGRAM);
 
+	m_timer = timer_alloc(FUNC(hcd62121_cpu_device::timer_tick), this);
+	m_timer->adjust(attotime::from_seconds(1), 0, attotime::from_seconds(1));
```
cuavas, line 350: Where does this one-second tick come from?  Does the CPU have a secondary clock input?  Is it an external 1-second tick input, or is it pre-scaled on the chip?  (Seiko Epson PDA CPUs have a secondary oscillator that’s usually connected to a 32.768kHz watch crystal that has configurable dividers for generating the clock tick signal.)

qufb: There's a secondary crystal oscillator that can be connected to the CPU, but it seems to be unrelated, since it's optional: it's present on Picky Talk, but on CFX9850GB both ends are connected with a resistor.

The 1-second tick feature is present on both systems. The way I measured it was by sampling the TIME register value after various timer waits, and it was being incremented roughly every second, including the cutoff at 60. It isn't very clear to me yet how the CPU is configuring this timer.

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -336,38 +378,38 @@ void hcd62121_cpu_device::device_start()
...
+	state_add(HCD62121_R58, "R58", *(u32*)&m_reg[0x58]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R5C, "R5C", *(u32*)&m_reg[0x5C]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R60, "R60", *(u32*)&m_reg[0x60]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R64, "R64", *(u32*)&m_reg[0x64]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R68, "R68", *(u32*)&m_reg[0x68]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R6C, "R6C", *(u32*)&m_reg[0x6C]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R70, "R70", *(u32*)&m_reg[0x70]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R74, "R74", *(u32*)&m_reg[0x74]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R78, "R78", *(u32*)&m_reg[0x78]).callimport().callexport().formatstr("%8s");
+	state_add(HCD62121_R7C, "R7C", *(u32*)&m_reg[0x7C]).callimport().callexport().formatstr("%8s");
```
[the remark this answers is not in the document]
qufb: It ended up being a bit verbose, let me know if it can be simplified.

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -336,42 +383,240 @@ void hcd62121_cpu_device::device_start()
...
+			m_reg[0x01] = (m_debugger_temp >> 16) & 0xff;
+			m_reg[0x02] = (m_debugger_temp >> 8) & 0xff;
+			m_reg[0x03] = m_debugger_temp & 0xff;
+			break;
+		case HCD62121_R04:
+			m_reg[0x04] = (m_debugger_temp >> 24) & 0xff;
+			m_reg[0x05] = (m_debugger_temp >> 16) & 0xff;
+			m_reg[0x06] = (m_debugger_temp >> 8) & 0xff;
+			m_reg[0x07] = m_debugger_temp & 0xff;
+			break;
```
cuavas, line 437: Since you know these indices are sequential, and you know your register storage is linear, can’t you do something like:
```c++
...
#include "multibyte.h"
...
void hcd62121_cpu_device::state_import(const device_state_entry &entry)
{
	if ((entry.index() >= HCD62121_R00) && (entry.index() <= HCD62121_R7C))
	{
		put_u32be(&m_reg[(entry.index() - HCD62121_R00) * 4], m_debugger_temp);
	}
	else switch (entry.index())
	{
...
```

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -381,6 +626,102 @@ void hcd62121_cpu_device::state_export(const device_state_entry &entry)
 		case STATE_GENPCBASE:
 			m_rtemp = (m_cseg << 16) | m_ip;
 			break;
+		case HCD62121_R00:
+			m_debugger_temp = (m_reg[0x00] << 24) | (m_reg[0x01] << 16) | (m_reg[0x02] << 8) | m_reg[0x03];
+			break;
+		case HCD62121_R04:
+			m_debugger_temp = (m_reg[0x04] << 24) | (m_reg[0x05] << 16) | (m_reg[0x06] << 8) | m_reg[0x07];
+			break;
```
cuavas, line 634: Similarly, for this could you do:
Since you know these indices are sequential, and you know your register storage is linear, can’t you do something like:
```c++
...
#include "multibyte.h"
...
void hcd62121_cpu_device::state_export(const device_state_entry &entry)
{
	if ((entry.index() >= HCD62121_R00) && (entry.index() <= HCD62121_R7C))
	{
		m_debugger_temp = get_u32be(&m_reg[(entry.index() - HCD62121_R00) * 4]);
	}
	else switch (entry.index())
	{
...
```

src/devices/cpu/hcd62121/hcd62121.cpp
```diff
@@ -522,6 +868,8 @@ void hcd62121_cpu_device::device_reset()
 	{
 		elem = 0;
 	}
+
+	m_debugger_temp = 0;
```
cuavas, line 872: There’s no reason to set this on reset – it’s only used temporarily anyway.

PR #12151, opened 2024-03-18
cpu/t11: Improve trace trap processing; add basic interrupt processing to the K1801VM1.

shattered:
…g to the K1801VM1.

### DIFF
--- a/src/devices/cpu/t11/t11.cpp
+++ b/src/devices/cpu/t11/t11.cpp
@@ -163,6 +163,8 @@ int t11_device::POP()
 #define GET_V (PSW & VFLAG)
 #define GET_Z (PSW & ZFLAG)
 #define GET_N (PSW & NFLAG)
+#define GET_T (PSW & (1 << 4))
+#define GET_I (PSW & (1 << 7))
 
 /* clears flags */
 #define CLR_C (PSW &= ~CFLAG)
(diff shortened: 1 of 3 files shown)


src/devices/cpu/t11/t11.cpp
```diff
@@ -210,6 +212,123 @@ static const struct irq_table_entry irq_table[] =
...
+	{
+		m_mcir = MCIR_IRQ;
+		m_vsel = VM1_IRQ3;
+	}
+
+	// 12. nVIRQ pin; PSW7, PSW10
+	else if (m_vec_active && !GET_I)
+	{
+		int vec = m_in_iack_func(0);
+		if (vec == -1 || vec == 0)
```
cuavas, line 282: Please don’t leave blank lines between the end of an `if` block and a following `else` – it greatly increases the risk of confusion when reading the code.

src/devices/cpu/t11/t11.h
```diff
@@ -106,6 +117,8 @@ class t11_device :  public cpu_device
 	PAIR                m_psw;
 	uint16_t            m_initial_pc;
 	uint8_t             m_wait_state;
+	int                 m_mcir;
+	uint16_t            m_vsel;
```
cuavas, line 121: It appears that `m_mcir` and `m_vsel` are supposed to remain valid when returning to the scheduler.  Should they be included in save states?

Also, things included in save states need to use explicitly sized types for portability (`int` is not explicitly sized).

PR #12153, opened 2024-03-18
Add a skeleton for a new PC-based game named 'Pull The Trigger'

clawgrip:
New not working machine
-----------------------------
Pull The Trigger [Arcade Planet, Recreativas.org]

### DIFF
--- a/src/mame/skeleton/pullttrigger.cpp
+++ b/src/mame/skeleton/pullttrigger.cpp
@@ -53,8 +53,8 @@ private:
 void pullttrig_state::pullttrig_map(address_map &map)
 {
 	map(0x00000000, 0x0009ffff).ram();
-	map(0x000e0000, 0x000fffff).rom().region("bios", 0);
-	map(0xfffe0000, 0xffffffff).rom().region("bios", 0);
+	map(0x000e0000, 0x000fffff).rom().region("bios", 0x20000);
+	map(0xfffc0000, 0xffffffff).rom().region("bios", 0);
 }
 
 void pullttrig_state::pullttrig_io(address_map &map)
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/pullttrigger.cpp
```diff
@@ -0,0 +1,94 @@
...
+	void pullttrig_io(address_map &map);
+	void pullttrig_map(address_map &map);
+};
+
+
+void pullttrig_state::pullttrig_map(address_map &map)
+{
+	map(0x00000000, 0x0009ffff).ram();
+	map(0x000e0000, 0x000fffff).rom().region("bios", 0);
+	map(0xfffe0000, 0xffffffff).rom().region("bios", 0);
```
angelosa, line 57: BIOS is 0x40000 in size, should work with this arrangement:
```
	map(0x000e0000, 0x000fffff).rom().region("bios", 0x20000);
	map(0xfffc0000, 0xffffffff).rom().region("bios", 0);
```

angelosa (commented):
I'm pretty convinced that covielsa needs its own folder, something that can be done later. It's otherwise a SiS651, which I'm planning support for.
The crash may be the same one as https://github.com/mamedev/mame/blob/master/src/mame/misc/gfamily.cpp , that fails at the end of Windows 2k bootstrap punting for a page fault with trap opcode (and yes, i386 throwing exception with a uint64 needs to be massaged too).

src/mame/skeleton/pullttrigger.cpp
clawgrip: Done, thanks!!!

PR #12159, opened 2024-03-20
Gp32.xml: Metadata cleanups

ArcadeShadow:
Replaced countries' abbreviations by their full name
Moved test comments to `notes` tag

### DIFF
--- a/hash/gp32.xml
+++ b/hash/gp32.xml
@@ -23,11 +23,13 @@ plus reload after save needs to be counterchecked for all games unless explicitl
 -->
 <softwarelist name="gp32" description="GamePark GP32 SmartMedia cards">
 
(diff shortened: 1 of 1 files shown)


rb6502:
I don't think any of the "boot OK" comments or notes are necessary.  If something is supported="yes" or "partial" that means that it boots, and for "partial" you can list just the problems (graphics glitches or save problems or whatever) as a note.

angelosa:
Yeah, `boot OK` can be safely removed from everything, it's a worthless anchor/adds nothing of value particularly since we consolidated that `supported=<undefined>` is a better alias for stuff that isn't tested at least once.

ArcadeShadow:
done it.

PR #12161, opened 2024-03-21
Add console support for KIM-1.

jefftranter:
Adds console support including emulation of hardware character echo.
Serial monitor commands work and I am able to load and run programs like Microsoft Basic.


src/mame/commodore/kim1.cpp
```diff
@@ -266,6 +302,13 @@ void kim1_state::sync_map(address_map &map)
 	map(0x0000, 0xffff).r(FUNC(kim1_state::sync_r));
 }
 
+// Called when serial data comes in from console.
+void kim1_state::tty_callback(uint8_t data)
+{
+	// Send data back to terminal to simulate the KIM-1 hardware
+	// echo.
+	m_rs232->write_txd(data);
+}
```
cuavas, line 311: You’re going to need to AND this with bit 0 of the port B output to avoid issues when they’re both changing at the same time.  Also, as this is a write line handler, the parameter type should conventionally be `int`.

src/mame/commodore/kim1.cpp
```diff
@@ -226,6 +259,9 @@ void kim1_state::u2_write_b(uint8_t data)
 	// Cassette write/speaker update
 	if (data & 0x20)
 		m_cass->output((data & 0x80) ? -1.0 : 1.0);
+
+	// Write bit 0 to serial console
+	m_rs232->write_txd(data & 0x01);
```
cuavas, line 264: You need to keep the state of line in `tty_callback` and AND it with bit 0 here to avoid issues when they’re both changing at once.

src/mame/commodore/kim1.cpp
```diff
@@ -113,11 +130,13 @@ class kim1_state : public driver_device
 	required_device_array<mos6530_device, 2> m_miot;
 	required_device<pwm_display_device> m_digit_pwm;
 	required_device<cassette_image_device> m_cass;
-	required_ioport_array<3> m_row;
+	required_ioport_array<4> m_row;
 	required_ioport m_special;
+	required_device<rs232_port_device> m_rs232;
```
cuavas, line 135: Please keep the devices together rather than mixing them with the I/O ports.

src/mame/commodore/kim1.cpp
```diff
@@ -113,11 +130,13 @@ class kim1_state : public driver_device
...
 
 	int m_sync_state = 0;
 	bool m_k7 = false;
+	uint8_t m_u2_port_a = 0;
```
cuavas, line 139: Why did you add this data member?  It doesn’t appear to be used for anything.

src/mame/commodore/kim1.cpp
```diff
@@ -80,7 +97,6 @@ Paste test:
 
 #include "kim1.lh"
 
-
 namespace {
```
cuavas, line 101: Please restore the two blank lines between sections.

src/mame/commodore/kim1.cpp
```diff
@@ -152,6 +173,13 @@ void kim1_state::machine_reset()
...
 
+static DEVICE_INPUT_DEFAULTS_START(terminal)
+	DEVICE_INPUT_DEFAULTS("RS232_RXBAUD", 0xff, RS232_BAUD_2400)
+	DEVICE_INPUT_DEFAULTS("RS232_TXBAUD", 0xff, RS232_BAUD_2400)
+	DEVICE_INPUT_DEFAULTS("RS232_DATABITS", 0xff, RS232_DATABITS_8)
+	DEVICE_INPUT_DEFAULTS("RS232_PARITY", 0xff, RS232_PARITY_NONE)
+	DEVICE_INPUT_DEFAULTS("RS232_STOPBITS", 0xff, RS232_STOPBITS_2)
+DEVICE_INPUT_DEFAULTS_END
 
 //**************************************************************************
```
cuavas, line 184: Please restore the two blank lines between sections here, too.

src/mame/commodore/kim1.cpp
```diff
@@ -266,6 +302,13 @@ void kim1_state::sync_map(address_map &map)
...
 
+// Called when serial data comes in from console.
+void kim1_state::tty_callback(uint8_t data)
+{
+	// Send data back to terminal to simulate the KIM-1 hardware
+	// echo.
+	m_rs232->write_txd(data);
+}
 
 //**************************************************************************
```
cuavas, line 313: Please restore the two blank lines between sections here, too.

src/mame/commodore/kim1.cpp
```diff
@@ -58,15 +73,17 @@ Paste test:
 
 
 TODO:
+- Get console working on screen (works with pty)
 - LEDs should be dark at startup (RS key to activate)
-- add TTY support
+- Make console toggle a DIP switch?
```
cuavas, line 78: What is it physically?  If it isn’t a DIP switch, it should not be a DIP switch in MAME.  Did you choose `IPT_KEYBOARD` just so it could easily be enabled/disabled in the Keyboard Selection menu when you want to use the terminal keyboard?

Speaking of which, is the reason the terminal isn’t “working” because you have multiple devices with keyboard inputs in the system and the terminal keyboard isn’t enabled?  MAME only enables the first keyboard in the system by default so you aren’t typing on all the keyboards at once. (go to main menu, Input Settings, Keyboard Selection to enable/disable keyboards).

cuavas (changes requested):

jefftranter:
The MR was updated to address the review comments.

x48x4b:
> Adds console support including emulation of hardware character echo. Serial monitor commands work and I am able to load and run programs like Microsoft Basic.

Hi & thank you!

- I started KIM-1 on my Linux (Fedora 39 on a Intel core i5 machine) box with
./mame kim1  -rs232 pty -resolution 800x600 -window
 
- In the "Machine Configuration" I set the serial paramters to "300 baud no parity 1 stop bit".

- I connected with 
  minicom -p /dev/pts/0 -b 300 -8

So far everything is fine and I can connect to the KIM-1 console as designed.

However, loading a papertape file with "L" and sending the *.pap file with minicom is not stable. If I send a short file this is okay, if I send a long file it does not work properly, even when setting character and line feed delays in minicom (20ms/200ms).

Best regards

x48x4b:
> > Adds console support including emulation of hardware character echo. Serial monitor commands work and I am able to load and run programs like Microsoft Basic.
> 
> Hi & thank you!
> 
> * I started KIM-1 on my Linux (Fedora 39 on a Intel core i5 machine) box with
>   ./mame kim1  -rs232 pty -resolution 800x600 -window
> * In the "Machine Configuration" I set the serial paramters to "300 baud no parity 1 stop bit".
> * I connected with
>   minicom -p /dev/pts/0 -b 300 -8
> 
> So far everything is fine and I can connect to the KIM-1 console as designed.
> 
> However, loading a papertape file with "L" and sending the *.pap file with minicom is not stable. If I send a short file this is okay, if I send a long file it does not work properly, even when setting character and line feed delays in minicom (20ms/200ms).
> 
> Best regards

With the following setup I was able to load the Basic interpreter papertape file and run it:
./mame kim1 -sl1 mtuk1016 -rs232 pty
Therefore, I conclude that yesterday, I did not load sufficient memory.

PR #12162, opened 2024-03-21
mr/dribling.cpp: Add a Dribbling bootleg from Automave (partial dump)

clawgrip:
New working clone
---------------------
Dribbling (Automave bootleg, Spain) [David Bosque, Recreativas.org]

Also renamed 'driblingbr' to 'driblingvm' and change game description.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -32281,8 +32281,9 @@ uchroma68                       // 1980 Motorola evaluation board for the MC6847
 @source:mr/dribling.cpp
 dribling                        // (c) 1983 Model Racing
 driblinga                       // (c) 1983 Model Racing
-driblingbr                      // bootleg
+driblingam                      // bootleg (Automave)
 driblingo                       // (c) 1983 Olympia
+driblingvm                      // bootleg (Videomac)
 
 @source:mr/sshot.cpp
 gunchamps                       // (c) 1980 Model Racing (sshot.cpp)
--- a/src/mame/mr/dribling.cpp
+++ b/src/mame/mr/dribling.cpp
@@ -644,8 +644,8 @@ ROM_END
  *
  *************************************/
 
-GAME( 1983, dribling,   0,        dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing",                   "Dribbling (set 1)",                    MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE)
-GAME( 1983, driblinga,  dribling, dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing",                   "Dribbling (set 2)",                    MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE)
-GAME( 1983, driblingo,  dribling, dribling, dribling, dribling_state, empty_init, ROT0, "Model Racing (Olympia license)", "Dribbling (Olympia)",                  MACHINE_IMPERFECT_SOUND | MACHINE_SUPPORTS_SAVE)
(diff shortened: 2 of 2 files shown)


src/mame/mr/dribling.cpp
```diff
@@ -598,7 +598,26 @@ ROM_START( driblingo )
 	ROM_LOAD( "tbp24s10.2d",  0x0500, 0x0100, CRC(a17d6956) SHA1(81724daf2e2d319f55cc34cc881b6a9a4e64e7ac) )
 ROM_END
 
-ROM_START( driblingbr )
+// Original Model Racing PCB (EF00284)
+ROM_START( driblingam )
```
cuavas, line 602: Hang on, if it’s an original Model Racing PCB, How is it a bootleg?  If they needed to obtain original Model Racing PCBs to make it, how would the bootleg be any cheaper than the original?  That makes it sound more like a hack or a version licensed for Spanish distribution than a bootleg.

Are you sure it isn’t a licensed version of Spanish distribution, or an importer changing the ROMs for the local market?

Also, do you know whether this used original sound hardware, or the bootleggers changed it?

cuavas: Considering it’s apparently built with an original board, I’d go with either “(hack?)” or assuming it was licensed.  Using an original board to make a bootleg of the same game (as opposed to a bootleg conversion to a different game) just doesn’t make sense.

clawgrip: Then I'd say it's a license, it does not make sense for a hack to present the machine on Italy on a trade fair referencing Model Racing (an Italian company). Neither make sense for a bootleg, indeed.
I'll change it tomorrow.
Thanks!

aovestdipaperino: Do we know if there is a different ROM?
Could this explain this [comment](https://github.com/mamedev/mame/pull/12127#issuecomment-1999902285)?

@cuavas OT: I had an idea about the PARATA JFET. How can I share/discuss it? Basically using the same AFUNC trick switching between current limiter and Ohmic region mode.

src/mame/mame.lst
```diff
@@ -32281,8 +32281,9 @@ uchroma68                       // 1980 Motorola evaluation board for the MC6847
 @source:mr/dribling.cpp
 dribling                        // (c) 1983 Model Racing
 driblinga                       // (c) 1983 Model Racing
-driblingbr                      // bootleg
+driblingam                      // bootleg (Automave)
 driblingo                       // (c) 1983 Olympia
+driblingvm                      // bootleg (Videomac)
```
aovestdipaperino, line 32287: Is it necessary to rename the file breaking existing setups? Legitimate question, speaking from an app-compat?

clawgrip: > Is it necessary to rename the file breaking existing setups? Legitimate question, speaking from an app-compat perspective

The bootlegs are usually defined by the bootlegger, and not by the bootlegger nationality.
IMHO, "Brazil" says nothing about this particular game itself (tagging it as "Brazilian Portuguese" would be different, tho).

cuavas: Short names change from time to time.  That’s just life.

src/mame/mr/dribling.cpp
cuavas: > @cuavas OT: I had an idea about the PARATA JFET. How can I share/discuss it? Basically using the same AFUNC trick switching between current limiter and Ohmic region mode.

You could start a discussion over here, in the General or Ideas section: https://github.com/orgs/mamedev/discussions

aovestdipaperino: @cuavas  Thanks. I've actually made good progress with the first JFET behaving correctly. The sound comes out right but it's attenuated. Opening a PR hoping your help can get this done.

PR #12163, opened 2024-03-21
xavix.cpp : gave tomplc & tcarnavi inputs (they seem to use digital pedals/steering etc.) and promoted both to working [David Haywood]

mamehaze:


src/mame/tvgames/xavix.cpp
```diff
@@ -564,6 +564,29 @@ static INPUT_PORTS_START( tomcpin )
...
+	PORT_BIT( 0x08, IP_ACTIVE_HIGH, IPT_BUTTON1 ) PORT_NAME("Horn / Select")
+
+	// should this be a 3 position shifter? (off or either state) - this is the up/down handle
+	PORT_BIT( 0x10, IP_ACTIVE_HIGH, IPT_BUTTON6 ) PORT_NAME("Throttle Low")
+	PORT_BIT( 0x20, IP_ACTIVE_HIGH, IPT_BUTTON7 ) PORT_NAME("Throttle High")
+
+	PORT_MODIFY("IN1")
+	// should this be a 3 position shifter? (off or either state) - this is the rotational handle
+	PORT_BIT( 0x01, IP_ACTIVE_HIGH, IPT_BUTTON8 ) PORT_NAME("Brake Low")
+	PORT_BIT( 0x02, IP_ACTIVE_HIGH, IPT_BUTTON9 ) PORT_NAME("Brake High")
```
cuavas, line 583: `IPT_POSITIONAL` or `IPT_POSITIONAL_V` might be a better choice for these – they both have off/1/2 notches, and holding buttons for them isn’t very realistic.

mamehaze: yeah, I'm not really sure what the best solution is 
plarail

navicar

are what these things look like anyway.

cuavas: Alternatively, I guess they could be treated as very low-resolution `IPT_PADDLE`/`IPT_PADDLE_V`.  But MAME will map those to auto-centring joystick axes, which will make them snap to the “1” position.  That’s not very helpful for a train simulator.

mamehaze: I mean if you have a preference, I don't really mind you adjusting it, I just wanted to identify the inputs.

These plug and plays tend to use all sorts of weird inputs, either simplified versions of things you might expect to be analog (as we see here) or ridiculously complex camera based where they're looking at an entire grid of analog values to work out exactly where you're holding something and at what angle in 3D space.

cuavas: > I mean if you have a preference, I don't really mind you adjusting it, I just wanted to identify the inputs.

Can you go with `IPT_POSITIONAL` and `IPT_POSITIONAL_V` for the train controller for now?

> simplified versions of things you might expect to be analog (as we see here) 

FWIW, train controls are often digital, just with more notches than the three you see here.  The old UK “standard” electropneumatic brake system has five notches (off, three levels, plus emergency), old electric trains often had four power notches, the standard AAR diesel loco controls have nine notches on the power control.  Digital (“notched”) brake controls tend to be less common than “notched” power controls, but they aren’t unheard-of.

cuavas: I thought you didn’t need the remap table if the values were sequential?  I didn’t use it in Sega BEENA.

Does something like this work?
```
	PORT_BIT( 0x30, 0x00, IPT_POSITIONAL_V ) PORT_POSITIONS(3) PORT_SENSITIVITY(10) PORT_KEYDELTA(1) PORT_FULL_TURN_COUNT(3) PORT_NAME("Power")

	PORT_MODIFY("IN1")
	PORT_BIT( 0x03, 0x00, IPT_POSITIONAL ) PORT_POSITIONS(3) PORT_SENSITIVITY(10) PORT_KEYDELTA(1) PORT_FULL_TURN_COUNT(3) PORT_NAME("Brake")
```

cuavas: That should give increment/decrement buttons to assign to step through the positions.

mamehaze: it kinda works, but it doesn't lock in place, it keeps trying to go back to 0

cuavas: Just to be annoying, positional controls default to auto-centre if they don’t wrap.  Easy enough to fix.

PR #12166, opened 2024-03-22
rm/rm380z.cpp: Fix 8" disk clock frequency and minor VDU-80 display improvements

RobinSergeant:
The RM 8" disk format is actually the IBM-3740 standard which means that generic CP/M disks in this format are compatible.  By testing these I've discovered that the FD-1771 square wave clock frequency needs to be changed for 8" disks.  This PR does that and allows such disks to load.  For example, I've just played the Infocom text adventure Zork from an 8" disk image :-)

I have also removed the slot device "fixed" restriction to allow switching between MDS and FDS drives.  It is actually possible to use FDS drives with MDS firmware if you patch the CP/M disk parameter tables, i.e boot from MDS disk in drive A:, then patch the disk parameter tables for drive B: before using with FDS disks!  I have written an assembler program to do just that.  The extra disk space allows larger CP/M programs to load which is actually very useful.  Apparently RM originally used something similar to transfer data between the two systems.

Display improvements include supporting the VID INHIB port flag which is used to control whether text is displayed with HRG or not.  There is an escape sequence for switching between these modes, which of course now works.

### DIFF
--- a/src/mame/rm/rm380z_m.cpp
+++ b/src/mame/rm/rm380z_m.cpp
@@ -74,14 +74,18 @@ void rm380z_state_cos40::port_write(offs_t offset, uint8_t data)
 				m_user_defined_chars[(m_character % 128) * 16 + m_character_row] = data;
 			}
 		}
-		// ignore updates while bit 4 of port 0 is set
-		// (counter is not used to set the scroll register in this case, maybe used for smooth scrolling?)
-		else if (!(m_port0 & 0x10))
+		else
 		{
-			// set scroll register (used to verticaly scroll the screen and effect vram addressing)
-			m_vram.set_scroll_register(data & 0x1f);
+			// ignore scroll updates while bit 4 (CGMUX) of port 0 is set
+			// (counter is then used for smooth scrolling which is not currently implemented)
+			if (!(m_port0 & 0x10))
+			{
+				// set scroll register (used to verticaly scroll the screen and effect vram addressing)
+				m_vram.set_scroll_register(data & 0x1f);
+			}
 			// bit 6 drives the speaker
 			m_speaker->level_w(BIT(data, 6));
+			m_fbfd = data;
 		}
 		break;
 
@@ -348,6 +352,8 @@ void rm380z_state_cos34::machine_reset()
 void rm380z_state_cos40::machine_reset()
 {
 	rm380z_state::machine_reset();
+
+	m_fbfd = 0x00;
 
 	m_vram.reset();
 	memset(m_user_defined_chars, 0, sizeof(m_user_defined_chars));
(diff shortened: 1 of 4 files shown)


src/mame/rm/rm380z_m.cpp
```diff
@@ -321,6 +325,16 @@ void rm380z_state::disk_0_control(uint8_t data)
...
 	{
+		// FDS and MDS drives use a differnt square wave clock frequency
+		if (floppy->get_form_factor() == floppy_image::FF_8)
+		{
+			m_fdc->set_unscaled_clock(2_MHz_XTAL);
+		}
+		else
+		{
+			m_fdc->set_unscaled_clock(1_MHz_XTAL);
+		}
```
cuavas, line 336: Is the system actually capable of detecting the drive type and changing the floppy controller clock when selecting a drive like this?  Because if it isn’t, we shouldn’t be adding magical abilities like this in MAME.

cuavas (commented):
Automatically changing the FDC clock speed to suit the drive when selecting a drive strikes me as unlikely.  As nice as it is to be able to use an 8" drive and a 5.25" drive on the same controller, I doubt this would have worked on the real machine.

I expect the real machine just had jumpers or soldered links to select the 2 MHz or 1 MHz FDC clock, and required them to be changed manually if you wanted to switch between 8" and 5.25" drives.

Unless you can show evidence that the system could switch the FDC clock frequency on-the-fly like this, I’d rather we don’t add the capability to the system in MAME.

The rest of the change looks OK, though.

src/mame/rm/rm380z_m.cpp
RobinSergeant: Fair comment, but Mame does already do this in at least one other driver.  Please see xerox/xerox820.cpp (line 267).  This is where I got the idea!  Does that create a precedent or would you still prefer it be fixed based upon machine type (2 Mhz for FDS, and 1 Mhz for MDS)?

With real hardware you needed a special machine with two FDC boards and modified firmware.  I believe these were available for organisations that needed to transfer data between MDS and FDS systems.  Although there is no dump of that firmware.

cuavas: There’s a lot of bad code in MAME.  That doesn’t mean we want more of it.  Just set the FDC clock in machine configuration.

Also, does it actually have a 1MHz or 2MHz crystal for the FDC, or does it get the FDC clock frequency by dividing some other clock?  You aren’t supposed to add crystals that aren’t actually present in the system, e.g. if it gets the FDC clock frequency by dividing the main 16MHz clock, you should use `16_MHz_XTAL / 16` or `16_MHz_XTAL / 8` to get the 1MHz or 2MHz frequency.

RobinSergeant: Done!  There is no crystal so I've divided as you suggested.  I did wonder why clocks were divided like that throughout the code, but now I understand.

PR #12167, opened 2024-03-22
misc/diamondking.cpp: Added "Caribe" slot machine from Segasa Sonic

clawgrip:
New not working machine
-----------------------------
Caribe [Arcade Planet, Recreativas.org]

### DIFF
--- a/src/mame/misc/diamondking.cpp
+++ b/src/mame/misc/diamondking.cpp
@@ -164,6 +164,8 @@ void ss1b550base_state::ss1b550base(machine_config &config)
 {
 	M68340(config, m_maincpu, 3.68640_MHz_XTAL);
 
+	SPEAKER(config, "mono").front_center();
+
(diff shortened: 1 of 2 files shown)


clawgrip:
This machine uses as main CPU board the same PCB that 'Diamond King' uses as I/O.

src/mame/misc/diamondking.cpp
```diff
@@ -123,6 +160,14 @@ void ss1b202base_state::ss1b202base(machine_config &config)
 	ay8910.add_route(ALL_OUTPUTS, "mono", 1.0); // Guess
 }
 
+void ss1b550base_state::ss1b550base(machine_config &config)
+{
+	M68340(config, m_maincpu, 3.68640_MHz_XTAL);
+
+	msm6585_device &msm6585(MSM6585(config, "msm6585", 640'000)); // 640 Hz resonator
```
cuavas, line 167: 640 Hz?

clawgrip: Fixed! (I think so ;) )

PR #12168, opened 2024-03-22
Fix the PARATA netlist in nl_dribbling.cpp

aovestdipaperino:
Hey @cuavas the first JFET seems to behave correctly, see chart below.
If you can help me figure out why adding a second JFET messes things up instead of making them better, than this whole netlist will be complete. I tried to amplify the sound by cheating in different ways without success.

output


aovestdipaperino:
Little bit of progress empirically
If I set the R8 resistor to 1M to compensate for not setting the K parameter on the JFET I get a proper wave form
output
This matches very closely what I get with MicroCap 12.
I cannot hear any sound and that makes me suspicious of something happening afterwards, so still investigating.

aovestdipaperino:
Put the sound recorded over the phone [here](https://jumpshare.com/s/GLexUpjra0PhSerFhkJP)

aovestdipaperino:
With all the simulations improvements, it still feels the fake netlist is much more accurate than the real one

aovestdipaperino:
Accidentally close while fixing my repos.
Started the discussion here: https://github.com/orgs/mamedev/discussions/100

PR #12174, opened 2024-03-23
cpu/i386: saner fatal error handling

angelosa:


angelosa:
throwing an uint64 to emu layer will cause MAME to just return unhandled caught exception, I don't think it's intended behaviour but need some direction here.

cracyc:
Exceptions there are supposed to land at https://github.com/mamedev/mame/blob/master/src/devices/cpu/i386/i386.cpp#L2857.  The only reason it's caught there is to fix esp.  OS/2, at least, requires that because it lazily allocates the stack.

PR #12176, opened 2024-03-23
mr/dribling.cpp: Add the missing ROMs from the Automave version of Dribbling [David Bosque, Recreativas.org]

clawgrip:


cuavas:
So only two ROMs are different to the Model Racing version?

clawgrip:
Yes!

PR #12179, opened 2024-03-24
sgi_mips: new software list additions

davide125:
- NetBSD 4.0
- NetBSD 4.0.1
- NetBSD 5.0
- NetBSD 5.0.1
- NetBSD 5.0.2
- NetBSD 5.1
- NetBSD 5.1.1
- NetBSD 5.1.2
- NetBSD 5.1.3
- NetBSD 5.1.4
- NetBSD 5.1.5
- NetBSD 5.2
- NetBSD 5.2.1
- NetBSD 5.2.2
- NetBSD 5.2.3
- NetBSD 6.0
- NetBSD 6.0.1
- NetBSD 6.0.2
- NetBSD 6.0.3

### DIFF
--- a/hash/sgi_mips.xml
+++ b/hash/sgi_mips.xml
@@ -5846,7 +5846,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="netbsd_2_0">
+	<software name="netbsd_2_0" cloneof="netbsd_2_0_2">
 		<description>NetBSD 2.0</description>
 		<year>2004</year> <!-- 12/04 -->
 		<publisher>The NetBSD Project</publisher>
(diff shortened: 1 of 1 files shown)


rb6502:
Do all of these boot and install?  It would be good to have supported="yes" or supported="no" tags on all of the items.

rb6502:
@davide125 Any updates on this?  If you could at least parent/clone the sub-releases that would be helpful.

davide125:
Sorry, I forgot about this one. Updated now with the parent/clone relationships as requested. As for whether these work... it's complicated. They all boot to some extent, but I never managed to actually install any of them because the hard disk is never recognized by the installer (and I didn't have time to debug it further than that when I put this together). Some versions also fail to initialize graphics correctly and are only usable over serial.

PR #12182, opened 2024-03-26
video/s3virge: refactor

angelosa:
(TBD)


angelosa:
Opening as PR for checking where I fumbled during the conversion (timer drift?).

angelosa:
I don't yet have a well defined goal here, shortening the FIFO queue is challenging, namely because the docs don't exactly define what's a "slot" vs. the actual BitBLT command length. This is important for win2k (which hangs as soon as you start accessing stuff) and for a couple corruption bugs still happening.
May consider just doing a bare `std::function` conversion in selected BitBLT places + other misc cleanups and call it for this branch, postponing localized fixes at later date unless something really obvious pops up.

angelosa:
Closing this, applied with https://github.com/mamedev/mame/commit/3b5e60b420b8c59d4373828dd4a2cf548523d810 , no use having this in WIP and as a feature branch.

PR #12185, opened 2024-03-26
snes.xml: Added a prototype.

FoxhackDN:
New working software list items (snes.xml)
Rockman's Soccer (Japan, sample) [C. Davis]

### DIFF
--- a/hash/snes.xml
+++ b/hash/snes.xml
@@ -22732,6 +22732,19 @@ more investigation needed...
 		</part>
 	</software>
 
+	<software name="rockmansp" cloneof="megamans">
+		<description>Rockman's Soccer (Japan, sample)</description>
+		<year>1994</year>
+		<publisher>Capcom</publisher>
+		<info name="alt_title" value="ロックマンズサッカー" />
+		<part name="cart" interface="snes_cart">
+			<feature name="slot" value="lorom" />
+			<dataarea name="rom" size="1572864">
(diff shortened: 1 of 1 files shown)


hash/snes.xml
```diff
@@ -22732,6 +22732,19 @@ more investigation needed...
...
 
+	<software name="rockmansp" cloneof="megamans">
+		<description>Rockman's Soccer (Japan, sample)</description>
+		<year>1994</year>
+		<publisher>Capcom</publisher>
+		<info name="alt_title" value="ロックマンズサッカー" />
+		<part name="cart" interface="snes_cart">
+			<feature name="slot" value="lorom" />
+			<dataarea name="rom" size="1572864">
+				<rom name="rockman's soccer (prototype).sfc" size="1572864"  crc="B999B148" sha1="28E030ACD8A41DE9CA8F6AED5144D3DD85DE6D11" offset="0x000000" />
```
cuavas, line 22743: Please use lowercase hex digits, and keep the size and offset together when adding ROMs to make things more readable.

cuavas (changes requested):

hash/snes.xml
FoxhackDN: Sorry about the uppercase hashes, fixed now. I think I moved the offset to where you want but if not just lemme know.

PR #12186, opened 2024-03-26
Add -listconfigs and -config command line options.

wilbertpol:
The -listconfigs verb lists configuration and dipswitch ports.

Example:
```
mame gridlee -listconfigs
SYSTEM           CONFIG NAME      CONFIG VALUE     CONFIG DESCRIPTION
---------------- ---------------- ---------------- ----------------------------
gridlee          :DSW:0x03                         Bonus Life
                                  0x00             8000 points
                                  0x01             10000 points (default)
                                  0x02             12000 points
                                  0x03             14000 points
                 :DSW:0x0c                         Lives
                                  0x00             2
                                  0x04             3 (default)
                                  0x08             4
                                  0x0c             5
                 :DSW:0x10                         Free Play
                                  0x00             Off (default)
                                  0x10             On
                 :DSW:0x20                         Cabinet
                                  0x00             Upright (default)
                                  0x20             Cocktail
                 :DSW:0x40                         Reset Hall of Fame
                                  0x00             No (default)
                                  0x40             Yes
                 :DSW:0x80                         Reset Game Data
                                  0x00             No (default)
                                  0x80             Yes
                 :IN1:0x30                         Coinage
                                  0x20             2 Coins/1 Credit
                                  0x00             1 Coin/1 Credit (default)
                                  0x10             1 Coin/2 Credits
                 :IN2:0x20                         Service Mode
                                  0x20             Off (default)
                                  0x00             On
```

 And -config can be used to set options for configuration and dipswitch ports.

For instance: `mame gridlee -config :DSW:0x0c=0x0c,:IN1:0x30=0x20` will start the game with 5 lives and 2 coins/1 credit coinage.

### DIFF
--- a/src/emu/config.cpp
+++ b/src/emu/config.cpp
@@ -102,11 +102,12 @@ bool configuration_manager::load_settings()
 
 void configuration_manager::apply_command_line_configs()
 {
-	if (machine().options().conf() != nullptr)
+	if (machine().options().config() != nullptr)
 	{
 		const char config_separator = ',';
 		const char setting_separator = '=';
-		const std::string passed_configs = machine().options().conf();
+		const std::string port_mask_hex_prefix = ":0x";
+		const std::string passed_configs = machine().options().config();
 		std::istringstream configs_stream(passed_configs);
 		for (std::string config_setting; std::getline(configs_stream, config_setting, config_separator);)
 		{
@@ -137,6 +138,7 @@ void configuration_manager::apply_command_line_configs()
 			bool found = false;
 			for (auto &port : machine().ioport().ports())
 			{
+				// Apply the provided value to all fields in the port
 				if (port.first == key)
 				{
 					found = true;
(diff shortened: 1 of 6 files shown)


ajrhacker:
I have very little confidence that adding a `-config` option would be an acceptable solution to a real problem.

First of all, MAME's option processing system already has to manage a plethora of system-dependent option types, including slot options, image options, `-bios` and `-ram`, and these don't always work as they should, especially when switching systems through the UI. Adding `-config` on top of that would only complicate matters further.

Second, MAME already stores DIP and configuration switch settings in .cfg files, and I'm a bit worried about potential conflicts between these and settings in both command lines and INI files.

Third, I personally doubt that "start the game with 5 lives and 2 coins/1 credit coinage" is an important use case for option configuration. The real problem with configuration switches, I believe, is in systems that expect different values depending on the slot options, BIOSes, and software images. This, however, is a really hard problem with no simple and universal solution.

As for `-listconfigs`, it doesn't seem worth adding by itself; it's fairy simple to extract all information presented here from the `-listxml` output.

wilbertpol:
> I have very little confidence that adding a `-config` option would be an acceptable solution to a real problem.
> 
> First of all, MAME's option processing system already has to manage a plethora of system-dependent option types, including slot options, image options, `-bios` and `-ram`, and these don't always work as they should, especially when switching systems through the UI. Adding `-config` on top of that would only complicate matters further.
> 
> Second, MAME already stores DIP and configuration switch settings in .cfg files, and I'm a bit worried about potential conflicts between these and settings in both command lines and INI files.
> 

The command line option would override whatever what was set in .cfg files. Afaik, a similar approach is used elsewhere in mame.

> Third, I personally doubt that "start the game with 5 lives and 2 coins/1 credit coinage" is an important use case for option configuration. The real problem with configuration switches, I believe, is in systems that expect different values depending on the slot options, BIOSes, and software images. This, however, is a really hard problem with no simple and universal solution.
>

For such simple things, indeed, not much value. 
I figured it might be useful in setting up a more complicated system where the instructions could then just contain a config option to pass.

PR #12187, opened 2024-03-27
igs/igs036crypt.cpp: added preliminary decryption for bubucar [XingXing]

HappyASR:


cuavas:
Hang on, this isn’t hooked up to anything – it’s just an unreferenced `extern` array in a source file that isn’t even declared in the corresponding header.  This needs corresponding changes in igs036crypt.h and igs_m036.cpp.  Also, the comment about the reliability of the keys in igs036crypt.cpp needs to be updated if this key is only “preliminary”.

HappyASR:
Sorry for confusion. 

Those IGS036/IGS027 keys are only for informational purposes. They should be included in comments or separate binary files, as someone might need them to decrypt the ROM and trojan the internal code.

The game won't boot without the internal code, and the internal code is responsible for uploading those keys to the decryption hardware, so those keys become useless for the final emulation.

Regarding reliability, I'm pretty sure this table will produce 100% correct results for this game.

mamehaze:
right, to clarify, for the games where the internal ROM is dumped, the tables get uploaded (by the internal ROM, from the internal ROM), and MAME handles it that way.

all the emulated games using an 036 CPU already have the internal ROM dumped at this point, so the offline decryption functions were removed at some point.

however, as HappyASR says, if somebody wants to trojan the game, they'll need this information to decrypt the ROM in order to modify it, so the information here is useful.

PR #12188, opened 2024-03-27
Add Robotron K7071 (text-only video card for A7100 series)

shattered:


src/devices/bus/multibus/robotron_k7071.cpp
```diff
@@ -0,0 +1,221 @@
...
+void robotron_k7071_device::device_add_mconfig(machine_config &config)
+{
+	Z80(config, m_cpu, XTAL(16'000'000) / 6);
+	m_cpu->set_addrmap(AS_PROGRAM, &robotron_k7071_device::cpu_mem);
+	m_cpu->set_addrmap(AS_IO, &robotron_k7071_device::cpu_pio);
+
+	I8257(config, m_dma, XTAL(16'000'000) / 6);
+	m_dma->out_hrq_cb().set(FUNC(robotron_k7071_device::hrq_w));
+	m_dma->in_memr_cb().set([this] (offs_t offset) { return m_cpu->space(AS_PROGRAM).read_byte(offset); });
+	m_dma->out_iow_cb<0>().set([this] (u8 data) { m_crtc->dack_w(data); });
```
cuavas, line 82: Can you please add a `memory_access<16, 0, 0, ENDIANNESS_LITTLE>::specific` member for accessing CPU memory?  It’s a lot faster than accessing the address space like this.

Also, do you really need the trampoline for IOW -> DACK?  Can’t you just do:
```c++
	m_dma->out_iow_cb<0>().set(m_crtc, FUNC(&i8275_device::dack_w));
```

Issue #11950, opened 2024-01-19
megadriv.xml: The rom ferias.zip from the software list is a chinese hack rom


arthurthekidboy:
### MAME version

0.261

### System information

Windows 10
Radeon Vega 3
Athlon 3000G
8GB Ram

### Emulated system/software

megadriv

### Incorrect behaviour

The rom of the software list of genesis have a chinese hack rom from the game ferias frustadas do pica pau (ferias.zip).the rom of the software list have a glitched intro.
The original rom (that has dumped from no intro database in 2021) have a original sega logo, tectoy logo (with a character woody woodpecker appears in the logo) and the trademark in the intro.
This rom was discovered by a Brazilian YouTube channel called ucollecionador who dumped the rom and he tells the story of this rom

### Expected behaviour

redump the rom to the original from 2021

### Steps to reproduce

1. Start the MAME
2. Open the Genesis rom
3. In The Software List runs ferias.zip

arthurthekidboy:
Theres a comparison from the roms

The Chinese Hacked Rom

![image](https://github.com/mamedev/mame/assets/102826358/d08ae26b-8bd3-4943-8dac-1207f24c2cab)

ICEknigh7:
Just replaced that bad dump for the good one, thanks for reporting it.

mamehaze:
The old dump should be marked as a bootleg / pirate rather than removed, it clearly came from a cart, just not the original Brazilian cart.
(I haven't seen your PR for it yet, so maybe that's what you did)

ICEknigh7:
Right, hadn't thought of that. It's now back in the softlist as a clone.

bootmii:
Can we close this issue? It looks resolved to me.

ICEknigh7:
I wouldn't mark it as resolved until the Pull Request is actually merged, since the wrong version will still be showing up until then.

ICEknigh7:
PR has been merged, so this can be closed.

arthurthekidboy:
Thanks a lot

PR #12190, opened 2024-03-27
megadriv.xml: Added 2 working items, replaced bad dump

ICEknigh7:
Bad Dump replaced with verified one:
- Férias Frustradas do Pica-Pau (Brazil) (fixes GitHub Issue https://github.com/mamedev/mame/issues/11950) [MARCSLASH, arthurthekidboy]

Clones added:
- Sonic The Hedgehog (World, JP2, ripped from Sonic Mega Collection) [Sonic Cult, SonicBlur, Just Me, Haku Ronin]
- Sonic The Hedgehog 2 (prototype 199209xx, pirate) [drx, Hidden Palace]

### DIFF
--- a/hash/megadriv.xml
+++ b/hash/megadriv.xml
@@ -17163,7 +17163,18 @@ https://segaretro.org/Feng_Shen_Ying_Jie_Chuan
 		<publisher>Tec Toy</publisher>
 		<part name="cart" interface="megadriv_cart">
 			<dataarea name="rom" width="16" endianness="big" size="1048576">
-				<rom name="ferias frustradas do pica-pau (bra).bin" size="1048576" crc="7b2e416d" sha1="ec546d1c00c88554b5e9135f097cb1c388cfd68f"/>
+				<rom name="ferias frustradas do pica-pau (bra).bin" size="1048576" crc="b03507c4" sha1="295e3fe1eaec3e5181e7dcb46cec6507d393b82d"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="feriasp" cloneof="ferias">
+		<description>Férias Frustradas do Pica-Pau (Brazil, pirate)</description>
+		<year>1995</year>
+		<publisher>Tec Toy</publisher>
+		<part name="cart" interface="megadriv_cart">
+			<dataarea name="rom" width="16" endianness="big" size="1048576">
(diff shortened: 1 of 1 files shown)


hash/megadriv.xml
```diff
@@ -27866,6 +27891,22 @@ Black screen, writes in ROM area then jumps to PC=1a2000 (encryption?)
 		</part>
 	</software>
 
+	<software name="sonic2p12" cloneof="sonic2">
+		<description>Sonic The Hedgehog 2 (prototype 199209xx, pirate)</description>
+		<year>1992</year>
+		<publisher>Sega</publisher>
+		<notes><![CDATA[
+Earlier than the "Beta 4" prototype. No equivalent build has been found in the Sega QA Archive.
+Original scene release which was altered to include a cracktro and bypass the copy protection. Dump originally in smd format, converted to raw binary due to this softlist not supporting smd files.
```
angelosa, line 27900: It's not `softlist` blame, it's because .smd are intended to be used by https://www.segaretro.org/Super_Magic_Drive which needs a separate softlist altogether.

mamehaze: smd is just a different data ordering, you can convert freely between them.

a lot of the prototypes etc. that were dumped years ago were in smd format.

I don't really think there's any point in making a list containing smd dumps, converting them to .bin is the correct approach.

ICEknigh7: "Softlist blame" removed.

angelosa: Gotta love your assertive way of putting whatever you think it should be an universal approach to do things.
https://old.reddit.com/r/everdrive/comments/s8rd8l/smd_roms_not_loading_mega_everdrive_pro/
> EDIT: To be clear, you will need to use UCON64 or something similar to convert the interleaved smd file to a bin/md file to compare the hashes.

angelosa (approved):

ICEknigh7:
Ready to go then, finally?

hash/megadriv.xml
mamehaze: yes, there are tools to convert between them, there always have been.  It's 16kb sections interleaved, rather than byteswwapped, and a header, but either way, it's a reproducable transformation, there's really no need to have a list full of copier format files when they can be made 100% from the binary files.

https://www.raphnet.net/electronique/genesis_cart/smdform.txt

Issue #12154, opened 2024-03-19
dc.xml: complete SW list


angelosa:
Dreamcast software list goes up to 'N' only (roughly, not all #-N by description are in), which puts QA in a state where most of the games described in (the rather old) [Psyman's list](https://github.com/flyinghead/flycast/blob/master/docs/Notable%20game%20bugs.md) cannot be tested.

cuavas:
I’m not sure the current content of the list is in great shape anyway.  The content already there was added when chdman still discarded all track gap information for GD-ROM images.

I realise there’s an opinion in some sectors that the gap information “doesn’t matter”, but throwing away data that “doesn’t matter” never ends well.  Too many times it’s turned out that discarded data *does* matter.

The current code on master supposedly preserves gap data properly for GD-ROM images.  However, it seems current versions of certain other emulators can’t boot Dreamcast CHDs if gap data is present.

Has anyone tried booting a Dreamcast CHD with gap data present in MAME to see what happens?

Do we want to start over with the Dreamcast software list when we’re sure GD-ROM handling in CHD format has stabilised more?

PR #12191, opened 2024-03-29
chdman: Add support for exporting Dreamcast cues and outputting one bin per track

987123879113:
The output split bins code took inspiration from from @landfillbaby's PR (https://github.com/mamedev/mame/pull/11727). I included some of their changes and did some additional refactoring, and also implemented the start of a formatting system for output filenames after some general design feedback from cuavas.

An example of the formatting system:
`chdman extractcd -i test.chd -o test.cue -ob "test %03t.bin"`
will result in the output bin filenames in the format of "test 001.bin", "test 002.bin", "test 003.bin", etc. 
And `%t` will give the track number without any additional formatting.

The output split bins is a prerequisite for implementing exporting of Dreamcast cues so the two changes have been included together. The commits are fully independent changes so I recommend looking at the specific commits separately instead of the full PR diff for reviewing.

Throwing this out as a draft PR for now to just get some feedback and additional testing.

### DIFF
--- a/src/tools/chdman.cpp
+++ b/src/tools/chdman.cpp
@@ -97,7 +97,7 @@ constexpr int MODE_GDI = 2;
 #define OPTION_INPUT "input"
 #define OPTION_OUTPUT "output"
 #define OPTION_OUTPUT_BIN "outputbin"
-#define OPTION_OUTPUT_SPLIT "split"
+#define OPTION_OUTPUT_SPLITBIN "splitbin"
 #define OPTION_OUTPUT_FORCE "force"
 #define OPTION_INPUT_START_BYTE "inputstartbyte"
 #define OPTION_INPUT_START_HUNK "inputstarthunk"
@@ -630,7 +630,7 @@ static const option_description s_options[] =
 	{ OPTION_INPUT_PARENT,          "ip",   true, " <filename>: parent file name for input CHD" },
 	{ OPTION_OUTPUT,                "o",    true, " <filename>: output file name" },
 	{ OPTION_OUTPUT_BIN,            "ob",   true, " <filename>: output file name for binary data" },
-	{ OPTION_OUTPUT_SPLIT,          "sb",   false, ": output one binary file per track" },
+	{ OPTION_OUTPUT_SPLITBIN,       "sb",   false, ": output one binary file per track" },
 	{ OPTION_OUTPUT_FORCE,          "f",    false, ": force overwriting an existing file" },
 	{ OPTION_OUTPUT_PARENT,         "op",   true, " <filename>: parent file name for output CHD" },
 	{ OPTION_INPUT_START_BYTE,      "isb",  true, " <offset>: starting byte offset within the input" },
@@ -785,7 +785,7 @@ static const command_description s_commands[] =
 		{
 			REQUIRED OPTION_OUTPUT,
 			OPTION_OUTPUT_BIN,
-			OPTION_OUTPUT_SPLIT,
+			OPTION_OUTPUT_SPLITBIN,
 			OPTION_OUTPUT_FORCE,
 			REQUIRED OPTION_INPUT,
 			OPTION_INPUT_PARENT,
@@ -2568,11 +2568,11 @@ static void do_extract_cd(parameters_map &params)
 		default_name.erase(chop, default_name.size());
 
 	// GDIs will always output as split bin
-	bool is_splitbin = mode == MODE_GDI || params.find(OPTION_OUTPUT_SPLIT) != params.end();
+	bool is_splitbin = mode == MODE_GDI || params.find(OPTION_OUTPUT_SPLITBIN) != params.end();
 	if (!is_splitbin && cdrom->is_gdrom() && mode == MODE_CUEBIN)
 	{
 		// GD-ROM cue/bin is in Redump format which should always be split by tracks
-		util::stream_format(std::cout, "Warning: --%s is required for this specific combination of input disc type and output format, enabling automatically\n", OPTION_OUTPUT_SPLIT);
+		util::stream_format(std::cout, "Warning: --%s is required for this specific combination of input disc type and output format, enabling automatically\n", OPTION_OUTPUT_SPLITBIN);
 		is_splitbin = true;
 	}
 
@@ -2599,15 +2599,15 @@ static void do_extract_cd(parameters_map &params)
 	{
 		output_bin_file_str = output_bin_file_fnd->second;
 
-		chop = (*output_bin_file_str).find_last_of('.');
+		chop = output_bin_file_str->find_last_of('.');
 		if (chop != std::string::npos)
 		{
-			output_bin_file_ext = (*output_bin_file_str).substr(chop, (*output_bin_file_str).size() - chop);
-			(*output_bin_file_str).erase(chop, (*output_bin_file_str).size());
+			output_bin_file_ext = output_bin_file_str->substr(chop, output_bin_file_str->size() - chop);
+			output_bin_file_str->erase(chop, output_bin_file_str->size());
 		}
 	}
 
-	if ((*output_bin_file_str).find('"') != std::string::npos || output_bin_file_ext.find('"') != std::string::npos)
+	if (output_bin_file_str->find('"') != std::string::npos || output_bin_file_ext.find('"') != std::string::npos)
 		report_error(1, "Output bin filename (%s%s) must not contain quotation marks", *output_bin_file_str, output_bin_file_ext);
 
 	// print some info
@@ -2690,7 +2690,7 @@ static void do_extract_cd(parameters_map &params)
 
 			if (is_splitbin && !found_track_variable)
 			{
-				report_error(1, "A track number variable (%%t) must be specified in the output bin filename when --%s is enabled\n", OPTION_OUTPUT_SPLIT);
+				report_error(1, "A track number variable (%%t) must be specified in the output bin filename when --%s is enabled\n", OPTION_OUTPUT_SPLITBIN);
 			}
 
 			// verify output BIN file doesn't exist
@@ -2913,7 +2913,7 @@ static void do_extract_cd(parameters_map &params)
 		// delete the output files
 		output_bin_file.reset();
 		output_toc_file.reset();
-		for (auto output_bin_filename : output_bin_filenames)
+		for (auto const &output_bin_filename : output_bin_filenames)
 			osd_file::remove(output_bin_filename);
 		osd_file::remove(*output_file_str->second);
 		throw;
--- a/src/tools/chdman.cpp
+++ b/src/tools/chdman.cpp
@@ -97,7 +97,7 @@ constexpr int MODE_GDI = 2;
 #define OPTION_INPUT "input"
 #define OPTION_OUTPUT "output"
 #define OPTION_OUTPUT_BIN "outputbin"
-#define OPTION_OUTPUT_SPLIT "outputsplit"
+#define OPTION_OUTPUT_SPLIT "split"
 #define OPTION_OUTPUT_FORCE "force"
 #define OPTION_INPUT_START_BYTE "inputstartbyte"
 #define OPTION_INPUT_START_HUNK "inputstarthunk"
@@ -630,7 +630,7 @@ static const option_description s_options[] =
 	{ OPTION_INPUT_PARENT,          "ip",   true, " <filename>: parent file name for input CHD" },
 	{ OPTION_OUTPUT,                "o",    true, " <filename>: output file name" },
 	{ OPTION_OUTPUT_BIN,            "ob",   true, " <filename>: output file name for binary data" },
-	{ OPTION_OUTPUT_SPLIT,          "os",   false, ": output one binary file per track" },
+	{ OPTION_OUTPUT_SPLIT,          "sb",   false, ": output one binary file per track" },
 	{ OPTION_OUTPUT_FORCE,          "f",    false, ": force overwriting an existing file" },
 	{ OPTION_OUTPUT_PARENT,         "op",   true, " <filename>: parent file name for output CHD" },
 	{ OPTION_INPUT_START_BYTE,      "isb",  true, " <offset>: starting byte offset within the input" },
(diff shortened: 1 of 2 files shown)


landfillbaby:
if you can figure out what i broke about split .toc in 70e8418b4f91135f6295a46842376d41f1866eaa too this would completely obsolete my semi-abandoned PR :)

987123879113:
@landfillbaby Is this an updated version of the code that wasn't in the PR? I didn't realize you worked on a version with TOC support.

I gave a quick comparison between the output this PR and the linked commit:
- GDI: same output (both .gdi and bins), although filenames are forced to a different format than is conventional for gdi
- Split TOC: same output (both .toc and bins)
- Non-split TOC: the first track has the proper filename in the `DATAFILE` command but every other track becomes `DATAFILE ""`. Output bins are the same.
- Split cue/bin: the INDEX times aren't starting at 00:00:00 for tracks starting in a new file. The indexes are relative to the loaded file so this would be a big bug.
- Non-split cue/bin: same output (both .cue and bins)

The issue with non-split TOCs seems to be because you're passing `trackbin_name` to `output_track_metadata` but `trackbin_name` is only set when a new file is loaded, but it's inside the for loop so always starts as a blank string. Would be an easy fix to just move the `std::string trackbin_name;` out of the loop. The INDEX issue could be fixed by clearing `discoffs` every time a file is loaded but only if it's not GDI (GDI keeps the running total between each track but TOC and CUE work relative to the loaded file).

landfillbaby:
oh you're right! oops!

src/tools/chdman.cpp
```diff
@@ -1462,62 +1466,19 @@ static void compress_common(chd_file_compressor &chd)
...
-		file.printf("%d %d %d %d %s%s%s %d\n", tracknum+1, frameoffs, mode, size, quotestr, filename, quotestr, discoffs);
+		file.printf("%d %d %d %d %s%s%s %d\n", tracknum+1, frameoffs, tracktype, info.datasize, quotestr, filename, quotestr, outputoffs);
 	}
 	else if (mode == MODE_CUEBIN)
 	{
-		// first track specifies the file
-		if (tracknum == 0)
+		// specify a new file when writing to the beginning of a file
+		if (outputoffs == 0)
 			file.printf("FILE \"%s\" BINARY\n", filename);
```
cuavas, line 1482: I know the code never attempted to deal with it, but how are double quote characters supposed to be escaped in filenames in cuesheets?  Does it use CSV-style `""` to represent one `"` character inside a quoted string, or C-style backslash escapes, or something else?

src/tools/chdman.cpp
```diff
@@ -2591,66 +2551,151 @@ static void do_extract_cd(parameters_map &params)
...
+					if (format_type == "t")
+					{
+						// track number
+						replacement = util::string_format("%" + format_part + "d", tracknum+1);
+
+						if (!is_splitbin)
+						{
+							util::stream_format(std::cout, "Warning: --%s not specified but track format string detected, enabling automatically\n", OPTION_OUTPUT_SPLIT);
+							is_splitbin = true;
+						}
```
cuavas, line 2670: I wouldn’t do this – I’d just treat it as a literal `%t` if split binary isn’t enabled.

Also, are you checking to ensure there’s at least one `%t` if split binary *is* enabled?  If you don’t, it will overwrite the same file repeatedly, which isn’t very useful.

src/tools/chdman.cpp
balr0g: Generally unsupported — https://hydrogenaud.io/index.php/topic,87257.msg747275.html#msg747275

src/tools/chdman.cpp
```diff
@@ -96,6 +97,7 @@ constexpr int MODE_GDI = 2;
 #define OPTION_INPUT "input"
 #define OPTION_OUTPUT "output"
 #define OPTION_OUTPUT_BIN "outputbin"
+#define OPTION_OUTPUT_SPLIT "outputsplit"
```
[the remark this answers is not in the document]
987123879113: How do you feel about the shorthand name for the command? Do you think it should be `-s` or is `-os` ok?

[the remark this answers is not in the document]
cuavas: `-s` or `-sb` maybe?

src/tools/chdman.cpp
987123879113: >If you don’t, it will overwrite the same file repeatedly, which isn’t very useful.

Yeah that shouldn't happen, I forgot to test that specific case. I changed the code a little bit to be more saner in that regard (https://github.com/mamedev/mame/pull/12191/commits/8335ada0d17a8af3df5dcb1d3ae1681c007fc12c). Now without a track specifier in the filename it'll act the same as chdman currently does instead.

987123879113: Ok, the last few commits should address this now.

```
> ./chdman extractcd -i test.chd  -o test_split.cue -f -ob "test.bin" -sb
chdman - MAME Compressed Hunks of Data (CHD) manager 0.264 (mame0252-4086-g578558cd79a)
Input CHD:    test.chd
Output TOC:   test_split.cue
A track number variable (%t) must be specified in the output bin filename when --split is enabled

Fatal error occurred: 1
```


```
> ./chdman extractcd -i test.chd  -o test_split.cue -f -ob "test %03t.bin" -sb
chdman - MAME Compressed Hunks of Data (CHD) manager 0.264 (mame0252-4086-g578558cd79a)
Input CHD:    test.chd
Output TOC:   test_split.cue
Output Data:  test 001.bin
Output Data:  test 002.bin
Output Data:  test 003.bin
Output Data:  test 004.bin
Output Data:  test 005.bin
Output Data:  test 006.bin
Output Data:  test 007.bin
Output Data:  test 008.bin
Output Data:  test 009.bin
Output Data:  test 010.bin
Output Data:  test 011.bin
Output Data:  test 012.bin
Extraction complete
```

```
> ./chdman extractcd -i test.chd  -o test.cue -f -ob "test %03t.bin"
chdman - MAME Compressed Hunks of Data (CHD) manager 0.264 (mame0252-4086-g578558cd79a)
Input CHD:    test.chd
Output TOC:   test.cue
Output Data:  test %03t.bin
Extraction complete
```

src/tools/chdman.cpp
987123879113: I can't see there being a lot of demand for quotation marks in the exported filenames but it definitely could cause problems depending on implementation it seems. I wouldn't be against disallowing quotation marks in output bin filenames.

src/tools/chdman.cpp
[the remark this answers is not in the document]
987123879113: Went with `-sb` but willing to go shorter with `-s` if anyone has a stronger opinion.

landfillbaby:
i think i'll close my PR in favour of this one now :)

src/tools/chdman.cpp
```diff
@@ -2818,7 +2910,8 @@ static void do_extract_cd(parameters_map &params)
 		// delete the output files
 		output_bin_file.reset();
 		output_toc_file.reset();
-		osd_file::remove(*output_bin_file_str);
+		for (auto output_bin_filename : output_bin_filenames)
+			osd_file::remove(output_bin_filename);
```
cuavas, line 2917: Can you use `for (auto const &output_bin_filename : output_bin_filenames)` here to avoid unnecessary copies?

src/tools/chdman.cpp
```diff
@@ -2591,66 +2551,155 @@ static void do_extract_cd(parameters_map &params)
...
 	if (output_bin_file_fnd == params.end())
+	{
 		output_bin_file_str = &default_name;
+	}
 	else
+	{
 		output_bin_file_str = output_bin_file_fnd->second;
 
-	check_existing_output_file(params, *output_bin_file_str);
+		chop = (*output_bin_file_str).find_last_of('.');
```
cuavas, line 2602: You can use the `->` operator to avoid some ugliness – `chop = output_bin_file_str->find_last_of('.');` etc.  It’s present for pointers, iterators, and other similar things.

src/tools/chdman.cpp
987123879113: Added an error for quotation marks in output bin filename.
https://github.com/mamedev/mame/pull/12191/commits/9b6d4bd5b22976c7788f86d0b2807ed0904bbfa5

```
> ./chdman extractcd -i test.chd  -o test_split.cue -f -sb -ob 'test".bin'
chdman - MAME Compressed Hunks of Data (CHD) manager 0.264 (mame0252-4087-gbed1cd5180a-dirty)
Output bin filename (test") must not contain quotation marks
Fatal error occurred: 1
```

src/tools/chdman.cpp
```diff
@@ -96,6 +97,7 @@ constexpr int MODE_GDI = 2;
 #define OPTION_INPUT "input"
 #define OPTION_OUTPUT "output"
 #define OPTION_OUTPUT_BIN "outputbin"
+#define OPTION_OUTPUT_SPLIT "split"
```
cuavas, line 100: I’d rather `--splitbin` and `-sb` or `--split` and `-s` so there’s more obvious correspondence between them.

Personally I’d lean towards `--splitbin` to hint that it affects the `--outputbin` file(s).

cuavas (commented):
I’m OK with this PR in principle.  Of course it would be nice to have a few people try it over the next few weeks before release when it gets merged.

src/tools/chdman.cpp
987123879113: Me completely forgetting how pointers work in C++. Fixed.

src/tools/chdman.cpp
987123879113: Makes sense. Fixed: https://github.com/mamedev/mame/pull/12191/commits/118b7b2083b11b3d2d33488be26004ff73c6062a

cuavas:
The last thing this is missing is to add the new option to the “Additional options” list for `extractcd` in this file: https://github.com/mamedev/mame/blob/master/docs/source/tools/chdman.rst

Yes, the majority of the options in that file aren’t explained.  You don’t have to explain the new option – just add it after `--outputbin`.

cuavas:
This will address #5867.

987123879113:
Went through and tested all of the GD-ROM "types" when extracting as cue using data from both TOSEC and Redump sources. Found some bugs addressed in https://github.com/mamedev/mame/pull/12191/commits/0ed5dbcdf13e14f44b8529d19ff46b1bfd8ac363. Wrote a previous comment [here](https://github.com/mamedev/mame/pull/12087#issuecomment-1975098822) for some info about the breakdown of the "types".

Also tested various cue/bin (split tracks), cue/bin (single bin), toc/bin (single bin) sources and extracting back out to cue/bin (split tracks), cue/bin (single bin), toc/bin (single bin), and toc/bin (split tracks) and didn't find any particular issues with those.

cuavas:
Let’s go!  It isn’t even April yet, so we should have plenty of time to fix it if an issue surfaces.

johnsanc314:
@987123879113 - Was this tested for Dreamcast? I tried this with a sample of every DC disc type with chdman from mame 0.265 and every single one failed to match the source tracks.

```
Test                Result   Notes
----------------------------------------------------------------------------------
Type 1 Redump CUE   Fail     Track 3 correct, track 1 and 2 incorrect
Type 1 Redump GDI   Fail     Tracks 1 and 3 correct, track 2 incorrect
Type 1 TOSEC GDI    Fail     Tracks 1 and 3 correct, track 2 incorrect

Type 2 Redump CUE   Fail     All tracks incorrect
Type 2 Redump GDI   Fail     Tracks 1 and 3 correct, tracks 2, 4, and 5 incorrect
Type 2 TOSEC GDI    Fail     Tracks 1 and 3 correct, tracks 2, 4, and 5 incorrect

Type 3 Redump CUE   Fail     All tracks incorrect
Type 3 Redump GDI   Fail     Tracks 1, 3, and 5 correct, tracks 2 and 4 incorrect
Type 3 TOSEC GDI    Fail     Tracks 1, 3, and 5 correct, tracks 2 and 4 incorrect


Sources used:
----------------------------------------------------------------------------------
Type 1 Redump: Virtua Athlete 2000 (USA) (En,Fr,De,Es)
Type 1 TOSEC:  Virtua Athlete 2000 v1.002 (2000)(Agetec)(US)(M4)[!]

Type 2 Redump: Memories Off 2nd - Making Disc (Japan)
Type 2 TOSEC:  Memories Off 2nd Making Disc v1.002 (2001)(KID)(JP)[!][Memories Off 2nd, T-19708M]

Type 3 Redump: Namco Museum (USA)
Type 3 TOSEC:  Namco Museum v1.010 (2000)(Namco)(US)[!][2S][compilation]
```

987123879113:
@johnsanc314 I gave it another test just now with latest master and everything is working as expected. What's expected is that you should be able to round trip the data (convert a Redump cue to CHD and extract to cue again, or convert a TOSEC GDI to CHD and extract to GDI again) and the output data should be the same. You'll never get data that matches one of the original dumps if you try converting between Redump and TOSEC formats because they're just fundamentally different, so that's expected too.

You'd probably be interested in this comment for more details. All of the comments about invalid cues were fixed with this PR. https://github.com/mamedev/mame/pull/12087#issuecomment-1975098822

johnsanc314:
@987123879113 - I could be wrong, but some discs may not be stored "correctly" in CHD format even though the roundtrip works ok. For example "MoHo (Europe) (En,Fr,De)" seems to have some data encoded as audio when it shouldn't be. Try comparing an uncompressed CHD with track 5 of that disc.

angelosa:
Gotta love people still using filenames as a way to identify discs instead of using checksums like already mentioned in the chain of events that lead me to open the dc.xml incomplete SW list: https://github.com/mamedev/mame/issues/12154

johnsanc314:
More specifically, its [Redump ID 49774 ](http://redump.org/disc/49774/)

987123879113:
@johnsanc314 That's just an unfortunate side effect of having to convert to TOSEC layout internally. If it's not done then either the data is irrecoverably discarded, which is bad, or it breaks compatibility with pretty much everything that uses Dreamcast CHDs, which is also bad.

PR #12192, opened 2024-03-29
strtdriv: fixed incorrect brake pedal port mapping in harddriv.cpp

Madcadden:
Fixed wrongly mapped brake pedal input from 8ADC:6 to 8ADC:2 and swapped the Voice Mic function which is unused from 8ADC:2 to 8ADC:6 to avoid any conflict/loss of functionality. Steering input still needs to be fixed (if you turn to the far left of right the centre point of the steering wheel will shift and constantly pull the wheel in the opposite direction) Hard Drivin's Airborne has the same problem. Will try to fix that if I can and update the code here on GitHub.

### DIFF
--- a/src/mame/atari/harddriv.cpp
+++ b/src/mame/atari/harddriv.cpp
@@ -1324,6 +1324,7 @@ static INPUT_PORTS_START( strtdriv )
 	PORT_START("mainpcb:8BADC.5")        /* b00000 - 8 bit ADC 5 - canopy */
 	PORT_BIT( 0xff, 0x80, IPT_UNUSED )
 
+	/* Placeholder ADC port for voice mic, NOT tested or verified! */
 	PORT_START("mainpcb:8BADC.6")        /* b00000 - 8 bit ADC 6 - voice mic */
 	PORT_BIT( 0xff, IP_ACTIVE_LOW, IPT_UNUSED )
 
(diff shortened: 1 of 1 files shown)


cuavas:
Do you know it’s supposed to be on that ADC channel?  Is there a schematic showing the connections, or can you see the code accessing ADC channel 6?

Madcadden:
> Do you know it’s supposed to be on that ADC channel? Is there a schematic showing the connections, or can you see the code accessing ADC channel 6?

The brake was originally mapped in in MAME to be on ADC channel 6 and it wasn't working because it was supposed to be on ADC channel 2 but that channel was already assigned to the "voice mic" in the code so i just swapped them around so that the brake would work but i have no clue what ADC channel the voice mic should be connected to.

cuavas:
Better to make the comment less confident and say it’s possibly the voice mic input if there’s no other supporting evidence.

Madcadden:
> Better to make the comment less confident and say it’s possibly the voice mic input if there’s no other supporting evidence.

Sorry if i sound silly but how would I write that to the code? Could i add this on the end of the line of code or above it? 

"/* Placeholder ADC port for voice mic, not tested or verified */"

I'm learning as i go along it's why I'm not sure what is the best way to add that information to the code!

cuavas:
Just update the comment on the `PORT_START` line as @rb6502 suggested, rather than adding an extra comment line.

Madcadden:
> Just update the comment on the `PORT_START` line as @rb6502 suggested, rather than adding an extra comment line.

Sorry i already added the extra line can i just edit it again so that it's the way you both suggested it to be?

cuavas:
> > Just update the comment on the `PORT_START` line as @rb6502 suggested, rather than adding an extra comment line.
> 
> Sorry i already added the extra line can i just edit it again so that it's the way you both suggested it to be?

Yes, just edit it again.  It gets squashed to a single commit on merge anyway.

PR #12193, opened 2024-03-29
sms.xml: Added 2 working items

ArcadeShadow:
Gemitas [SMS Power!]
Sydney Hunter and the Sacred Tribe [No-Intro]

Renamed set [jewelstn] to [jeweljon] and renamed description to "Jewel Jones"

### DIFF
--- a/hash/sms.xml
+++ b/hash/sms.xml
@@ -3132,6 +3132,16 @@ license:CC0-1.0
 		</part>
 	</software>
 
+	<software name="gemitas">
+		<description>Gemitas</description>
+		<year>2018</year>
+		<publisher>Cyttorak</publisher>
+		<part name="cart" interface="sms_cart">
+			<dataarea name="rom" size="65536">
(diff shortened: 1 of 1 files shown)


cuavas (commented):
When adding items, please keep the sizes and offsets together – it makes it much easier to see how multiple ROMs are loaded.

Also, you only seem to have included the offset for one of these.  There are at least a few Master System games with multiple ROM chips, and at least the vast majority of items in the software list seem to include an offset even if there’s only one ROM.  Maybe include the offset for both for consistency?

ArcadeShadow:
Added the offset in both set. I kept the `size` and `offset` together, but I think for consistency it should be done the same treatment for the existing sets.

cuavas:
> Added the offset in both set. I kept the `size` and `offset` together, but I think for consistency it should be done the same treatment for the existing sets.

Yeah, but that would be a huge change, and there’s also the issue of mixing decimal sizes with hex offsets to think about.

cuavas (commented):
Please use hex offsets – all the existing items in this software list that have non-zero offsets use hex offset, and it makes more sense in binary addressing schemes.

ArcadeShadow:
done it.

PR #12206, opened 2024-04-02
igs/pgm.cpp: fix chinese spelling in multiple descriptions

barbudreadmon:
A chinese friend was telling me "chuan" is incorrect for those descriptions, it should be "zhuan".

### DIFF
--- a/src/mame/igs/pgm.cpp
+++ b/src/mame/igs/pgm.cpp
@@ -5423,18 +5423,18 @@ GAME( 1997, pgm,          0,         pgm,                   pgm,       pgm_state
    Working (at least one set of the game is fully working)
    -----------------------------------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


barbudreadmon:
further explanations : 

传 (傳), a Chinese character at the first level, pronounced as chuán or zhuàn, a morpheme first found in the oracle bone inscriptions. The original meaning is to pass on, to transmit. Derived from this meaning, it means to pass on, to keep and pass on. When read as zhuàn, it has the meanings of a post station, a post house, a text recording historical events or personal deeds, a literary work or a work explaining scriptures.

Due to the context of those games (historical events), it should be zhuan.

taoenwen:
Among mame's reviewers, this issue is not so conspicuous, but it is so simple that if ten minutes is not enough, I believe that more time is not necessarily effective, and at the moment when you are sure of the finalization of the draft, do you want to leave some time for you to leave down the omission.

PR #12207, opened 2024-04-02
machine/icd2061a: Fix out of bounds array access

987123879113:
Just to be sure I checked and all other usages of `m_reg_clocks` are using constants instead of variables. `truckk` and `ujlnow` are both still booting so code-wise it's working as expected (they rely on the feature clock working properly to even boot).

cc: @happppp

### DIFF
--- a/src/devices/machine/icd2061a.cpp
+++ b/src/devices/machine/icd2061a.cpp
@@ -24,10 +24,14 @@
 
 #define LOG_PINS (1 << 1)
 #define LOG_STATE (1 << 2)
+#define LOG_TODO (1 << 3)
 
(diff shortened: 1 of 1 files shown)


happppp:
Should it keep the clock unchanged if select = VCLKOUT_FORCED_HIGH or VCLKOUT_HIGH_Z?

987123879113:
There's a TODO for those clock settings in the header of the file. I never found anything that used those so they're left unimplemented since I didn't want to implement it if I couldn't test it.

987123879113:
Added in a logger for TODOs and made it print out a message if it encounters one of the unimplemented clock settings instead of silently ignoring it entirely.

PR #12210, opened 2024-04-03
sinclair: adding devices for ZX Spectrum Next. [holub]

holub:
sinclair/specnext_divmmc.cpp: Adding ZX Spectrum Next devices.

cpu/z80n.cpp: z80 extension implemented in ZX Spectrum Next

### DIFF
--- a/src/devices/machine/ie15.cpp
+++ b/src/devices/machine/ie15.cpp
@@ -494,7 +494,7 @@ void ie15_device::device_reset()
 {
 	update_serial(0);
 
-	memset(&m_video, 0, sizeof(m_video));
+	m_video = decltype(m_video)();
 	m_kb_ruslat = m_long_beep = m_kb_control = m_kb_data = m_kb_flag0 = 0;
 	m_kb_flag = IE_TRUE;
 	m_kbd_sdv = false;
--- a/src/mame/midway/midyunit_v.cpp
+++ b/src/mame/midway/midyunit_v.cpp
@@ -58,7 +58,7 @@ VIDEO_START_MEMBER(midyunit_state,common)
 
 	/* reset DMA state */
 	memset(m_dma_register, 0, sizeof(m_dma_register));
-	memset(&m_dma_state, 0, sizeof(m_dma_state));
+	m_dma_state = dma_state_t();
 
 	/* register for state saving */
 	save_item(NAME(m_autoerase_enable));
@@ -272,14 +272,13 @@ void midyunit_state::midyunit_paletteram_w(offs_t offset, uint16_t data, uint16_
 
 void midyunit_state::dma_draw(uint16_t command)
 {
-	struct dma_state_t &dma_state = m_dma_state;
 	int dx = (command & 0x10) ? -1 : 1;
-	int height = dma_state.height;
-	int width = dma_state.width;
+	int height = m_dma_state.height;
+	int width = m_dma_state.width;
 	uint8_t *base = m_gfx_rom;
-	uint32_t offset = dma_state.offset >> 3;
-	uint16_t pal = dma_state.palette;
-	uint16_t color = pal | dma_state.color;
+	uint32_t offset = m_dma_state.offset >> 3;
+	uint16_t pal = m_dma_state.palette;
+	uint16_t color = pal | m_dma_state.color;
 	int x, y;
 
 	/* we only need the low 4 bits of the command */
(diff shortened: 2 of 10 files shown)


src/mame/sinclair/specnext_divmmc.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+	if (m_automap_reset || m_retn_seen)
+		m_button_nmi = 0;
+	else if (m_divmmc_button)
+		m_button_nmi = 1;
+	else if (automap_held())
+		m_button_nmi = 0;
+
+	return m_button_nmi;
+}
```
cuavas, line 87: This is called from `disable_nmi_r`, which is named like a read handler.  Can it be called (directly or indirectly) from an address space read handler?  If it can, you need to check `machine().side_effects_disabled()` before changing the state.

src/mame/sinclair/specnext_divmmc.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+bool specnext_divmmc_device::automap_held()
+{
+	if (m_automap_reset || m_retn_seen)
+		m_automap_held = 0;
+	else if (m_cpu_mreq_n)
+		m_automap_held = automap_hold();
+
+	return m_automap_held;
+}
```
cuavas, line 119: This is called from `automap_held_r``, which is named like a read handler.  Can it be called (directly or indirectly) from an address space read handler?  If it can, you need to check machine().side_effects_disabled() before changing the state.

src/mame/sinclair/specnext_divmmc.h
```diff
@@ -0,0 +1,90 @@
+#ifndef MAME_SINCLAIR_SPECNEXT_DIVMMC_H
+#define MAME_SINCLAIR_SPECNEXT_DIVMMC_H
```
cuavas, line 2: This file is still missing the license header comment.

holub: Let me re-apply couple changes from main PR.

src/mame/sinclair/specnext_divmmc.cpp
holub: We don't need these two `*_r()` public/exists at all

src/mame/sinclair/specnext_divmmc.cpp
cuavas: This can still be called from `divmmc_rom_en_r()` and `divmmc_ram_en_r()` (via `rom_en()` and `ram_en()`, respectively, which both call `automap()` which then calls `automap_held()`, which are named like read handlers.

I briefly looked at the code in the other PR, and it seems they’re only called from `specnext_state::bank_update`, which shouldn’t be called where side effects aren’t expected.

Should `divmmc_rom_en_r()` and `divmmc_ram_en_r()` be renamed to make it clearer they that they aren’t just getting signal states and have side effects?  Can they be called `try_rom_en()` and `try_ram_en()`, or `request_rom_en()` and `request_ram_en()`, or something?

holub: Yeah. Let me reflect on this.
If that's the only thing you unhappy about here, fill free to merge it. I'll catch up on this in the main PR.

holub: Must be cleaner this way

src/mame/sinclair/specnext_divmmc.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+bool specnext_divmmc_device::rom_en()
+{
+	return page0() && (conmem() || automap()) && !mapram();
+}
+
+bool specnext_divmmc_device::ram_en()
+{
+	return (page0() && (conmem() || automap()) && mapram()) || (page1() && (conmem() || automap()));
+}
```
cuavas, line 44: The other thing about these is that it’s not immediately clear whether the thing with side effects will happen, because it’s deep in a Boolean expression with short-circuit operators.

```
bool specnext_divmmc_device::rom_en()
{
	return page0() && (conmem() || automap()) && !mapram();
}

bool specnext_divmmc_device::ram_en()
{
	return (page0() && (conmem() || automap()) && mapram()) || (page1() && (conmem() || automap()));
}
```

cuavas (commented):
I’ll grab specnext_divmmc.{h,cpp} and tidy them a little to get one more device in and keep this moving a bit.

src/mame/sinclair/specnext_divmmc.cpp
cuavas: Oh never mind, the functions no longer has side effects.  That’s easier to understand.

cuavas:
I added the DivMMC device here: 71b4db167835a12b771400804998d7709395ffcd

The only real changes are:
* `const` qualifiers to hint to the developer that things don’t change state.
* `noexcept` qualifiers for most of the out-of-line member functions that just do simple operations on member variables so the compiler can assume no exceptions will be thrown across them.
* Prefix on the device short name.

holub:
> I added the DivMMC device here: 71b4db167835a12b771400804998d7709395ffcd
> 
> The only real changes are:
> * `const` qualifiers to hint to the developer that things don’t change state.
> * `noexcept` qualifiers for most of the out-of-line member functions that just do simple operations on member variables so the compiler can assume no exceptions will be thrown across them.
> * Prefix on the device short name.

Thanks a lot!

PR #12220, opened 2024-04-04
midway/astrocde.cpp: Add a Irecsa licensed Spanish set of Gorf 

clawgrip:
New working clone
---------------------
Gorf (Irecsa license, Spanish) [Recreativas.org]


cuavas:
For clarification, does the game play in Spanish, or is it licensed for distribution in Spain, or both?

PR #12224, opened 2024-04-08
docs: remove unnecessary repetition in othertools.rst

angelosa:
Header claims "Developer-focused" tools, repeating "Most users will not need to work with this." is an unnecessary hyperbole considering all these tools expect specific domain knowledge already.


cuavas:
The page isn’t particularly useful at the moment anyway.  The repetition doesn’t really help.

PR #12226, opened 2024-04-08
casio/pickytlk.cpp: Add tablet support

qufb:
Most screens can now be interacted as expected. Some additional changes include:
* Add a new flag register bit, which is used for detecting when the pen is pressed and then hold down;
* Remove the last column from the LCD display, which now matches the visible area in hardware;

### DIFF
--- a/src/mame/layout/pickytlk.lay
+++ b/src/mame/layout/pickytlk.lay
@@ -1,0 +1,116 @@
+<?xml version="1.0"?>
+<!--
+license:CC0-1.0
+-->
+<mamelayout version="2">
+	<element name="pencursor">
+		<image state="1">
+			<data><![CDATA[
+				<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="1" height="1" viewBox="0 0 1 1">
+					<g fill="none" stroke="#6666ff" stroke-width="0.06" stroke-opacity="1">
+						<circle cx="0.5" cy="0.5" r="0.1" />
+						<circle cx="0.5" cy="0.5" r="0.47" />
+						<line x1="0.03" y1="0.5" x2="0.4" y2="0.5" />
+						<line x1="0.6" y1="0.5" x2="0.97" y2="0.5" />
+						<line x1="0.5" y1="0.03" x2="0.5" y2="0.4" />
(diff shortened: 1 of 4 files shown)


src/mame/layout/pickytlk.lay
```diff
@@ -0,0 +1,116 @@
...
+	</element>
+
+	<!--
+	    Panel labels:
+
+	    |  ワリア  |   ▲   |   △   |メニユー|
+	               | ◀ | ▶ |
+	    |   修正   |   ▼   |   ▽   |  ＯＫ  |
+
+	    Panel labels (translated):
```
qufb, line 31: These might not be very accurate, so any input is appreciated. "Reset" is used for functions like clearing the screen after drawing, while "Edit" is used in the main screen for settings, including redoing the pen callibration / constrast config functions.

ajrhacker: I'd guess ワリア should actually be クリア (which literally means "clear").

src/mame/layout/pickytlk.lay
```diff
@@ -0,0 +1,116 @@
...
+					</g>
+				</svg>
+			]]></data>
+		</image>
+	</element>
+
+	<!--
+	    Panel labels:
+
+	    |  ワリア  |   ▲   |   △   |メニユー|
```
FoxhackDN, line 27: クリア = Clear
メニュー	= Menu (note the smaller yu particle)

FoxhackDN: Here's an image of the LCD menu for future reference:
image

qufb: Thanks, I'll amend these.

qufb:
@cuavas Do you have any concerns with the layout stuff done here?

In the meantime I scanned the touchpad overlay for this system, and would likely be preferable to use that instead of the discrete shapes being drawn here for buttons. Probably can be included in software list like Beena carts.

Note that the overlay sits on top of the LCD screen, so all the covered area can also be interacted. There's program code logic to match coordinates with the visually distinct buttons, so it shouldn't be implemented with discrete inputs like other overlay layouts.

PR #12228, opened 2024-04-09
asmjit: sync with upstream

pmackinlay:


pmackinlay:
> For reference, which upstream revision is this based on?

Upstream version was 1.13.0 reflected [here](https://github.com/asmjit/asmjit/blob/master/src/asmjit/core/api-config.h#L19), commit [e5d7c0b](https://github.com/asmjit/asmjit/commit/e5d7c0bd5d9aec44d68830187138149e6a8c4e32).

PR #12229, opened 2024-04-09
konami/konamigv: Implement sensors and printer check for Tokimeki Memorial Oshiete Your Heart

987123879113:
- Implemented heart beat sensor and GSR sensor for Tokimeki Memorial Oshiete Your Heart to make the games playable
- Implemented basics of the printer for Tokimeki Memorial Oshiete Your Heart. Only enough to trigger simple errors via dip switches and also to allow the game to not error out whenever the printer is used
- Added hand crafted EEPROM for `tmosh` to allow the game boot
- Swapped the left and right channel audio in the SPU for konamigv. The game `btchamp` has a sound test that specifically says the words "left", "right", and "center" to clear up doubts.

Not promoted to working due to the printer not being fully functional, but otherwise the games seem to be playable from limited testing. I also noticed some weirdness with the graphics that get shown when something is printed but I don't know if it's bugged or not.


src/mame/konami/konamigv.cpp
```diff
@@ -680,19 +779,115 @@ void konamigv_state::tokimeki_serial_w(offs_t offset, uint16_t data, uint16_t me
...
+	PORT_START("GSR")
+	PORT_BIT( 0xff, 0x00, IPT_POSITIONAL ) PORT_SENSITIVITY(100) PORT_KEYDELTA(2) PORT_CENTERDELTA(0) PORT_MINMAX(0x00, 0x80) PORT_NAME("GSR Sensor")
+
+	PORT_START("PRINTER_META")
+	PORT_DIPNAME( 0x01, 0x01, "Printer Paper Empty" )
+	PORT_DIPSETTING( 0x01, DEF_STR( Off ) )
+	PORT_DIPSETTING( 0x00, DEF_STR( On ) )
+	PORT_DIPNAME( 0x02, 0x02, "Printer Malfunction" )
+	PORT_DIPSETTING( 0x02, DEF_STR( Off ) )
+	PORT_DIPSETTING( 0x00, DEF_STR( On ) )
```
cuavas, line 887: Please use either `IPT_OTHER` with `PORT_TOGGLE` or `PORT_CONFNAME` etc. for these as they aren’t real DIP switches.

987123879113: Changed it to `PORT_CONFNAME` because I don't really want them showing up as keybinds like what happens with `IPT_OTHER`.

PR #12230, opened 2024-04-10
seta/seta2.cpp: Fixed year for Guardians / Denjin Makai II (bootleg)

barbudreadmon:
As seen on title screen, this bootleg is from 1998 : 

grdiansbl-240410-075554


Osso13:
Thanks, my bad

PR #12231, opened 2024-04-10
itech/itech32.cpp: Various cleanups and updates:

cam900:
Recuce runtime tag macros
Fix metadatas (ex: shufshot and clones are hasn't strata related stuff (logo, etc)) Use machine().side_effects_disabled() for debug stuffs Use correct, constant typename values
Fix naming for ROM regions
Use logmacro for logging features
Use C++ style comments for single line comments
Reduce defines
Use array for reduce duplicates
Move drivedge specific stuffs into drivedge_state

### DIFF
--- a/src/mame/itech/itech32.cpp
+++ b/src/mame/itech/itech32.cpp
@@ -404,14 +404,26 @@ Notes:
 #include "speaker.h"
 
 
-#define FULL_LOGGING                0
 #define LOG_DRIVEDGE_UNINIT_RAM     0
 
+#define LOG_SCREEN  (1U << 1)
+#define LOG_PIA     (1U << 2)
+#define LOG_DSP     (1U << 3)
+
+#define LOG_ALL     (LOG_SCREEN | LOG_PIA | LOG_DSP)
+
+#define VERBOSE (0)
+#include "logmacro.h"
+
+#define LOGSCREEN(...)  LOGMASKED(LOG_SCREEN, __VA_ARGS__)
+#define LOGPIA(...)     LOGMASKED(LOG_PIA, __VA_ARGS__)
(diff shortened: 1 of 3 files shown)


src/mame/itech/itech32.cpp
```diff
@@ -1869,7 +1834,9 @@ void itech32_state::bloodstm(machine_config &config)
 
 void drivedge_state::drivedge(machine_config &config)
 {
-	/* basic machine hardware */
+	static constexpr XTAL TMS_CLOCK     = XTAL(40'000'000);          // TMS320C31 clocks on drivedge
+
+	// basic machine hardware
```
cuavas, line 1839: Please don’t make this `static` – a local `constexpr` is better (avoids global data bloat).

src/mame/itech/itech32.h
```diff
@@ -20,12 +20,7 @@
 #include "emupal.h"
 #include "screen.h"
 
-#define VIDEO_CLOCK     XTAL(8'000'000)           // video (pixel) clock
-#define CPU_CLOCK       XTAL(12'000'000)          // clock for 68000-based systems
-#define CPU020_CLOCK    XTAL(25'000'000)          // clock for 68EC020-based systems
-#define SOUND_CLOCK     XTAL(16'000'000)          // clock for sound board
-#define TMS_CLOCK       XTAL(40'000'000)          // TMS320C31 clocks on drivedge
-
+#define LOG_DRIVEDGE_UNINIT_RAM     0
```
cuavas, line 23: Is there a specific reason you moved this macro to the header?  It only seems to be used in one file.  I’d understand if you were going to use it to avoid creating the `m_written` member at all if it’s disabled or something.  But it doesn’t seem to be used in the header at all.

cam900: That's for avoid to create RAM logging related variables and functions.

PR #12232, opened 2024-04-11
lib/formats/dmk_dsk.cpp: Header byte 4 is not reserved. Fixes MT 08851.

wilbertpol:


Robbbert:
#8851 indicates issue or PR number 8851 (you can see that github has added a link). For a mametesters report you need to use MT 08851.

PR #12233, opened 2024-04-12
konami/konamigv: Add printer support for Tokimeki Memorial Oshiete Your Heart

987123879113:
HLE implementation of the printer Sony UP-1200 analog video printer used by Tokimeki Memorial Oshiete Your Heart. The printer receives an S-Video signal from the subboard PCB and separately a communication line to control the machine. `tmosh` and `tmoshs` only take and print one picture, but `tmoshsp` takes 16 different pictures and uses the 16 image layout setting of the printer.

The printer output is shown as a second screen as can be seen in the example pictures below.

tmosh (1 large image):

tmoshs (Konami logo position changes, text and icon in the very center of the picture):

tmoshsp (Konami logo and Tokimeki Memorial logo are always in the same spot):

Here's a picture of multiple prints from both the `tmosh` version and the `tmoshsp` version for layout reference.
image

This PR also promotes the games to working, except for the `tmoshspa` clone which has a note about the CD being damaged.
```
Tokimeki Memorial Oshiete Your Heart (GQ673 JAA)
Tokimeki Memorial Oshiete Your Heart Seal Version (GE755 JAA)
Tokimeki Memorial Oshiete Your Heart Seal Version Plus (GE756 JAB)
```

### DIFF
--- a/src/mame/konami/konamigv.cpp
+++ b/src/mame/konami/konamigv.cpp
@@ -180,10 +180,10 @@ Notes: (all main parts shown)
        This PCB is plugged into the Tokimeki Memorial Oshiete Your Heart main board into CN4
        It provides additional functionality for the printer and sensor(s) and possibly other things.
 
-       CN10 & CN11 - TCS7927-54 4-pin mini DIN connectors
+       CN10 & CN11 - TCS7927-54 4-pin mini DIN connectors, S-Video in and out for printer
                CN9 - 6 pin connector
                CN8 - 5 pin connector
-               CN7 - 7 pin connector
+               CN7 - 7 pin connector, printer communication
                CN6 - 3 pin connector
                CN5 - 2 pin connector
                CN3 - 4 pin power connector. Joins to CN6 on mainboard via a Y-splitter cable. The other end of the
@@ -213,6 +213,13 @@ Notes:
       uPC817 - Sharp PC817 Optocoupler
          CN1 - 4 pin power connector joining to GQ673 PCB CN3 and CN6 on mainboard via a Y-splitter cable.
      CN1/CN2 - 2 pin connector
+
+
+Notes:
+The Tokimeki Memorial Oshiete Your Heart printer appears to be a model based on the Sony UP-1200 which is
+a color video printer that takes S-Video as input.
+https://www.ykuns-mechanical-club.com/tokimemo.html
+https://www2.biglobe.ne.jp/~tell/keihin/tokimeki/tyheart/tyheart.html
 
 ***************************************************************************/
 
@@ -245,6 +252,7 @@ class konamigv_state : public driver_device
 public:
 	konamigv_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
+		, m_screen(*this, "screen")
 		, m_ncr53cf96(*this, "scsi:7:ncr53cf96")
 		, m_btc_trackball(*this, "upd%u", 1)
 		, m_maincpu(*this, "maincpu")
@@ -271,6 +279,8 @@ protected:
 	void kdeadeye_map(address_map &map);
 
 	TIMER_CALLBACK_MEMBER(scsi_dma_transfer);
+
+	required_device<screen_device> m_screen;
 
 	required_device<ncr53cf96_device> m_ncr53cf96;
 	optional_device_array<upd4701_device, 2> m_btc_trackball;
@@ -319,6 +329,7 @@ public:
 		, m_gsr(*this, "GSR")
 		, m_printer_meta(*this, "PRINTER_META")
 		, m_device_val_start_state(0)
+		, m_printer_is_manual_layout(false)
 	{
 	}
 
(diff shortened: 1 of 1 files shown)


src/mame/konami/konamigv.cpp
```diff
@@ -486,8 +519,19 @@ void tokimeki_state::machine_start()
...
+	save_item(NAME(m_printer_is_printing));
+	save_item(NAME(m_printer_current_image));
+	save_item(NAME(m_printer_data));
+	save_item(NAME(m_page_bitmap));
+
 	m_heartbeat_timer = timer_alloc(FUNC(tokimeki_state::heartbeat_timer_tick), this);
 	m_heartbeat_timer->adjust(attotime::zero);
+
+	m_printer_printing_status_timeout = timer_alloc(FUNC(tokimeki_state::printing_status_timeout), this);
+	m_printer_printing_status_timeout->adjust(attotime::never);
```
cuavas, line 534: You don’t need to cancel the timer on creating it.  Did you mean to do that on reset?

src/mame/konami/konamigv.cpp
```diff
@@ -500,6 +544,17 @@ void tokimeki_state::machine_reset()
...
+	m_printer_bit = 0;
+	m_printer_curbit = 0;
+	m_printer_curbyte = 0;
+	m_printer_is_printing = 0;
+	m_printer_current_image = 0;
+
+	std::fill(std::begin(m_printer_data), std::end(m_printer_data), 0);
+
+	m_page_bitmap.allocate(PRINTER_PAGE_WIDTH, PRINTER_PAGE_HEIGHT);
+	m_page_bitmap.fill(0xffffffff);
```
cuavas, line 557: Is it necessary to reallocate the bitmap here?  Could you just use `resize` rather than forcing the bitmap’s memory to be released and reallocated with `realloc`?

src/mame/konami/konamigv.cpp
```diff
@@ -772,36 +832,201 @@ uint16_t tokimeki_state::tokimeki_serial_r()
...
+			if (m_printer_data[0] != m_printer_data[2] || m_printer_data[0] != m_printer_data[4] || m_printer_data[1] != m_printer_data[3] || m_printer_data[1] != m_printer_data[5])
+			{
+				logerror("printer command not accepted, found different bytes [%02x %02x %02x] [%02x %02x %02x]", m_printer_data[0], m_printer_data[2], m_printer_data[4], m_printer_data[1], m_printer_data[3], m_printer_data[5]);
+			}
+			else if (m_printer_data[1] != 0xf9)
+			{
+				logerror("printer command not accepted, found second byte that wasn't 0xf9 [%02x %02x %02x] [%02x %02x %02x]", m_printer_data[0], m_printer_data[2], m_printer_data[4], m_printer_data[1], m_printer_data[3], m_printer_data[5]);
+			}
+			else if (m_printer_data[0] == 0x10)
+			{
```
cuavas, line 931: My gut feel is that this should wait for vertical blanking to ensure it gets a properly rendered frame.  Can you start a timer on receiving this message to wait until some arbitrary point early in the vertical blanking interval to ensure you grab a frame after the final screen update for the frame happens?

In real life, the printer is going to have to see a complete frame after receiving this command (i.e. wait for current frame to complete, then start capturing at the top of the next frame).

src/mame/konami/konamigv.cpp
```diff
@@ -834,13 +1059,28 @@ void tokimeki_state::tokimeki_serial_w(offs_t offset, uint16_t data, uint16_t me
...
+}
+
+uint32_t tokimeki_state::printer_screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect)
+{
+	if (!m_printer_is_printing)
+		return 0;
+
+	copybitmap(bitmap, m_page_bitmap, 0, 0, 0, 0, cliprect);
+	return 0;
 }
```
cuavas, line 1084: You may be able to get a slight performance win by:
* Set a flag on reset and when you actually update the printer bitmap.
* If the flag is not set, just `return UPDATE_HAS_NOT_CHANGED;` here.
* If the flag is set, update the bitmap, clear the flag, and return zero here.

src/mame/konami/konamigv.cpp
987123879113: Wouldn't even need it on reset since the timer is only setting m_printer_is_printing to 0 which is already happening in machine_reset. I'll remove it.

src/mame/konami/konamigv.cpp
987123879113: I'm assuming you meant specifically the allocate line and not including the fill, right? The allocate can get moved to machine_start no problem.

src/mame/konami/konamigv.cpp
987123879113: Done. https://github.com/mamedev/mame/pull/12233/commits/0614bd38a9c0fe08aa02877324319174a37d434c

src/mame/konami/konamigv.cpp
cuavas: Yeah, the `allocate` line will cause it to delete the storage and reallocate it.  If you use `resize` instead, it won’t reallocate if it doesn’t need to.

(The `fill` isn’t a problem.  I selected both lines because you can’t select a single line for an in-place review comment, and if you don’t select two or more lines, you get some arbitrary number of lines quoted that will include stuff not relevant to the comment.)

src/mame/konami/konamigv.cpp
987123879113: Does this seem sane for what you're suggesting? https://github.com/mamedev/mame/pull/12233/commits/eab9298c6f0e8a835085ab0997cbdf683bfe2f98

src/mame/konami/konamigv.cpp
```diff
@@ -355,7 +377,22 @@ class tokimeki_state : public konamigv_state
...
+	uint8_t m_printer_bit;
+	uint8_t m_printer_data[6];
+	uint8_t m_printer_curbit;
+	uint8_t m_printer_curbyte;
+	attotime m_printer_pulse_starttime;
+	uint8_t m_printer_is_printing;
+	uint32_t m_printer_current_image;
+	bool m_printer_is_manual_layout;
+	bool m_printer_page_is_dirty;
+	int m_printer_video_last_vblank_state;
```
cuavas, line 389: `m_printer_video_last_vblank_state` is registered for save states, so it should be an explicitly sized integer or `bool`.

src/mame/konami/konamigv.cpp
cuavas: Yeah, looks about right.

PR #12235, opened 2024-04-12
itech/itech8.cpp: Various cleanups

cam900:
Reduce unnecessary lines
Reduce preprocessing defines
Constantize variables
Use correct typename values
Use C++ style comment for single line comments
Use palette_device for grmatch color behaviour
Remove unpopulated device
Fix filename comments
Use input_merger for FIRQ with VIA
Use machine().side_effect_disabled() for avoid debugger issue
Replace pointers to references
Fix spacing in metadatas

### DIFF
--- a/src/mame/itech/itech8.cpp
+++ b/src/mame/itech/itech8.cpp
@@ -1851,7 +1851,7 @@ void grmatch_state::grmatch(machine_config &config)
 	m_screen->set_visarea(0, 399, 0, 239);
 	m_screen->set_screen_update(FUNC(grmatch_state::screen_update));
 
-	config.device_remove("tlc34076"); // replaced to TMS34070s
+	config.device_remove("tlc34076"); // replaced with TMS34070s
 
 	for (auto &elem : m_palette)
 		PALETTE(config, elem).set_entries(16);
--- a/src/mame/itech/itech8.cpp
+++ b/src/mame/itech/itech8.cpp
@@ -505,6 +505,7 @@
 #include "cpu/z80/z80.h"
 #include "machine/6522via.h"
 #include "machine/6821pia.h"
+#include "machine/input_merger.h"
 #include "sound/okim6295.h"
 #include "sound/ymopn.h"
 #include "sound/ymopl.h"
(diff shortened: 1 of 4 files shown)


src/mame/itech/itech8_m.cpp
```diff
@@ -306,53 +300,53 @@ void slikshot_state::vels_to_inters(u8 x, u8 vx, u8 vy,
...
-			*word1 = inter2 - word2mod;
-			*word2 = ((uint64_t)word2mod << 16) / 0x16553;
+			// compute the other values from that
+			word1 = inter2 - word2mod;
+			word2 = ((u64)word2mod << 16) / 0x16553;
 		}
 		else
 			LOGSENSOR("inters_to_words: unable to convert %04x %04x %04x %02x\n",
-					(u32)inter1, (u32)inter2, (u32)inter3, (u32)*beams);
+					(u32)inter1, (u32)inter2, (u32)inter3, (u32)beams);
```
cuavas, line 329: Why the casts to `u32` here?  The logging functions work fine with any width integer types.

src/mame/itech/itech8_m.cpp
```diff
@@ -306,53 +300,53 @@ void slikshot_state::vels_to_inters(u8 x, u8 vx, u8 vy,
...
-			*word1 = inter1 - word2mod;
-			*word2 = ((uint64_t)word2mod << 16) / 0x16553;
+			// compute the other values from that
+			word1 = inter1 - word2mod;
+			word2 = ((u64)word2mod << 16) / 0x16553;
 		}
 		else
 			LOGSENSOR("inters_to_words: unable to convert %04x %04x %04x %02x\n",
-					(u32)inter1, (u32)inter2, (u32)inter3, (u32)*beams);
+					(u32)inter1, (u32)inter2, (u32)inter3, (u32)beams);
```
cuavas, line 349: Same here – the casts look unnecessary.

src/mame/itech/itech8_m.cpp
```diff
@@ -558,40 +569,38 @@ void slikshot_state::machine_start()
...
-	/* if the shoot button is pressed, fire away */
+	// if the shoot button is pressed, fire away
 	if (totaldy < m_last_ytotal && m_last_ytotal > 50 && m_crosshair_vis)
 	{
-		/* compute the updated values */
+		int temp;
+		// compute the updated values
 		temp = totaldx;
 		if (temp <= -0x80) temp = -0x7f;
 		if (temp >=  0x80) temp =  0x7f;
```
cuavas, line 606: Can you simplify this, and avoid the need for `temp` using `std::clamp`?  Something like:
```c++
		m_curvx = std::clamp<int>(totaldx, -0x7f, 0x7f);
		m_curvy = std::clamp<int>(m_last_ytotal - 50, 0x10, 0x7f);
		m_curx = std::clamp<int>(0x60 + (m_curxpos * 0x30 / 0x80), 0x30, 0x90);
```

The expression for `m_curx` looks like it could be further simplified to something like:
```c++
		m_curx = std::clamp<int>(0x60 + ((m_curxpos * 3) >> 3), 0x30, 0x90);
```

src/mame/itech/itech8.cpp
```diff
@@ -1847,6 +1850,11 @@ void grmatch_state::grmatch(machine_config &config)
 
 	m_screen->set_visarea(0, 399, 0, 239);
 	m_screen->set_screen_update(FUNC(grmatch_state::screen_update));
+
+	config.device_remove("tlc34076"); // replaced to TMS34070s
```
cuavas, line 1854: Should be “replaced with TMS34070s”.

PR #12239, opened 2024-04-12
msx/msx.cpp: Small MSX2+ related updates

wilbertpol:
- Move MSX2+ machines to msx/msx2p.cpp.
- Add support for Kanji level 2 I/O ports.
- Add support for MSX2+ boot flags register.
- Hook up msx2p_cart and msxr_cart software lists.

### DIFF
--- a/src/devices/bus/msx/slot/music.cpp
+++ b/src/devices/bus/msx/slot/music.cpp
@@ -9,6 +9,7 @@ DEFINE_DEVICE_TYPE(MSX_SLOT_MUSIC, msx_slot_music_device, "msx_slot_music", "MSX
 
 msx_slot_music_device::msx_slot_music_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
 	: msx_slot_rom_device(mconfig, MSX_SLOT_MUSIC, tag, owner, clock)
+	, m_ym2413_base(nullptr)
 	, m_ym2413(nullptr)
 	, m_ym2413_tag(nullptr)
 {
@@ -23,7 +24,7 @@ void msx_slot_music_device::device_start()
 		fatalerror("msx_slot_music_device: no YM2413 tag specified\n");
 	}
 
-	m_ym2413 = owner()->subdevice<ym2413_device>(m_ym2413_tag);
+	m_ym2413 = m_ym2413_base->subdevice<ym2413_device>(m_ym2413_tag);
 
 	if (m_ym2413 == nullptr)
 	{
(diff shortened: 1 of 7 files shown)


src/devices/bus/msx/slot/music.h
```diff
@@ -19,12 +19,17 @@ class msx_slot_music_device : public msx_slot_rom_device
...
+		m_ym2413_tag = finder.finder_target().second;
+	}
 
 protected:
 	virtual void device_start() override;
 
 private:
+	device_t *m_ym2413_base;
 	ym2413_device *m_ym2413;
 	const char *m_ym2413_tag;
```
cuavas, line 34: Is there a reason it can’t just use a `required_device<ym2413_device>` rather than this hackery?  It looks like leftovers from when someone didn’t understand object finders.

src/mame/msx/msx.cpp
```diff
@@ -453,6 +455,35 @@ void msx_state::kanji_w(offs_t offset, u8 data)
...
+		u32 latch = m_kanji_fsa1fx ? bitswap<17>(m_kanji_latch, 4, 3, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 2, 1, 0) : m_kanji_latch;
+		result = m_region_kanji->as_u8(0x20000 | latch);
+
+		if (!machine().side_effects_disabled())
+		{
+			m_kanji_latch = (m_kanji_latch & ~0x1f) | ((m_kanji_latch + 1) & 0x1f);
+		}
+	}
+	return result;
+}
```
cuavas, line 474: Can you turn `m_kanji_region` into an `optional_region_ptr<u8>` and index it directly (plus get some more validation) rather than needing `as_u8`?

Should you take the `address_space &` parameter and return `space.unmap()` if the kanji ROM is not present or too small, or is it guaranteed to read high?

Or alternatively, should these handlers only be installed if the kanji ROM is present and large enough (e.g. install them on start after doing the check)?

src/devices/bus/msx/slot/music.h
wilbertpol: Using required_device does indeed make it a lot simpler.

src/mame/msx/msx.cpp
wilbertpol: The handlers will now only be installed when a kanji region is present.

src/devices/bus/msx/slot/music.cpp
```diff
@@ -9,27 +9,14 @@ DEFINE_DEVICE_TYPE(MSX_SLOT_MUSIC, msx_slot_music_device, "msx_slot_music", "MSX
 
 msx_slot_music_device::msx_slot_music_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
 	: msx_slot_rom_device(mconfig, MSX_SLOT_MUSIC, tag, owner, clock)
-	, m_ym2413(nullptr)
-	, m_ym2413_tag(nullptr)
+	, m_ym2413(*this, "^ym2413")
```
cuavas, line 12: Please initialise this with `m_ym2413(*this, finder_base::DUMMY_TAG)` and let the user configure it.  Making assumptions about the host system never ends well – we’re trying to eliminate devices that construct object finders with root or parent references in tags.

PR #12242, opened 2024-04-13
trs/coco.cpp: Fix light gun offset

tlindner:
Changed offset to align MAME's light gun cursor with what the games expect. I believe this fell out of adjustment with the recent MC6847 changes.


tlindner:
I can guarantee this one character change will not cause the build to fail.

happppp:
It was a build failure on master branch with the coleco exp slot.

PR #12246, opened 2024-04-14
midway/williams.cpp, wmg.cpp: Various cleanups

cam900:
Use memory_view for ROM/Palette overlay
Reduce unnecessary devices, variables
Reduce ROM region size, rearrange ROM offsets
Fix naming for tile ROM region
Reduce runtime tag lookups
suppress machine().side_effects_disabled() for debugger reading PIA constantize variables


cuavas:
This causes `defender` and other games with the same configuration to crash.  I fixed it in 65b5718e0c0c9d37d83864e169dd67734f6ee7ee, but you should really test all the machine configurations you’re changing.

PR #12247, opened 2024-04-14
atari/atetris.cpp: Add a new Atari Tetris bootleg (with an extra OKI for sound)

clawgrip:
New working clone
---------------------
Tetris (bootleg set 7, with OKI M5205) [Taksangs]

### DIFF
--- a/src/mame/atari/atetris.cpp
+++ b/src/mame/atari/atetris.cpp
@@ -107,6 +107,7 @@ Notes:
 #include "machine/eeprompar.h"
 #include "machine/gen_latch.h"
 #include "machine/watchdog.h"
+#include "sound/msm5205.h"
 #include "sound/pokey.h"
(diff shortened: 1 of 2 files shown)


cuavas:
I ran it for a while, and there are no unmapped memory accesses for the MSM5205.

Given the additional sound module connects to a ROM socket, perhaps reading certain ROM locations selects the MSM5205, and it sees the data read from ROM as the byte to be written?  If it does that, it may require an address space tap.

At any rate, working it out will require debugging the code.  I’ll leave it open for a bit in case anyone wants to have a crack at it.

ajrhacker:
MC14040 is a CMOS ripple counter, and counter ICs are a typical part of the complex supporting logic required to stream samples directly from ROMs into a MSM5205 (since the MSM5205 has no memory interface of its own).

PR #12250, opened 2024-04-14
rx78_cart.xml: New working item added, metadata cleanup

FoxhackDN:
* Fixed names for Hamburger Shop and Tatakae! Ultraman
* Added serial and alt_title metadata where available
* Updated list of available titles

New working software list items (rx78_cart.xml)
Cannon Ball [Gaming Alexandria]

### DIFF
--- a/hash/rx78_cart.xml
+++ b/hash/rx78_cart.xml
@@ -113,10 +113,13 @@ The monitor will expect you to load a program from cassette.
 		</part>
 	</software>
 
-	<software name="cannonbl">
+	<software name="cannonbl" supported="partial">
 		<description>Cannon Ball</description>
 		<year>1983</year>
 		<publisher>Bandai</publisher>
+		<notes><![CDATA[
+Controls during gameplay do not work correctly; pressing A to move left also fires a shot. Gamepad input works fine.
+]]></notes>
 		<info name="serial" value="0201028"/>
 		<info name="alt_title" value="キャノンボール"/>
 		<info name="usage" value="Press A for single player or D for two players. A or D to move, S to Shoot" /> <!-- Keyboard controls are glitchy and don't respond correctly, gamepad works fine. -->
--- a/hash/rx78_cart.xml
+++ b/hash/rx78_cart.xml
@@ -113,16 +113,16 @@ The monitor will expect you to load a program from cassette.
 		</part>
 	</software>
 
-	<software name="cannonbl" supported="partial">
+	<software name="cannonbl">
 		<description>Cannon Ball</description>
 		<year>1983</year>
 		<publisher>Bandai</publisher>
 		<notes><![CDATA[
-Controls during gameplay do not work correctly; pressing A to move left also fires a shot. Gamepad input works fine.
+Keyboard controls do not work correctly because of conflicts with Joystick 2's default keys. Unmapping them fixes the issue.
 ]]></notes>
 		<info name="serial" value="0201028"/>
 		<info name="alt_title" value="キャノンボール"/>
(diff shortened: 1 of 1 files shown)


hash/rx78_cart.xml
```diff
@@ -109,14 +113,16 @@ The monitor will expect you to load a program from cassette.
...
-	<software name="burger">
-		<description>Burger Shop</description>
+	<software name="cannonbl">
+		<description>Cannon Ball</description>
 		<year>1983</year>
 		<publisher>Bandai</publisher>
-		<info name="usage" value="Z and X to start and move; Space to pick up and put down" />
+		<info name="serial" value="0201028"/>
+		<info name="alt_title" value="キャノンボール"/>
+		<info name="usage" value="Press A for single player or D for two players. A or D to move, S to Shoot" /> <!-- Keyboard controls are glitchy and don't respond correctly, gamepad works fine. -->
```
angelosa, line 122: Legit sounds a `supported="partial"` +comment should be in notes field.

FoxhackDN: I agree. I did some additional testing by running the game with -nojoy, and the game still had the same input problems. I'll amend this in a bit.

happppp:
Are you certain it's not a conflict of default inputs? -nojoy won't disable the emulated joystick, that cmd line option is for the host joystick.

Press TAB, go to the input assignments menu for this system. See that the joystick 2 default keys are: R F D G A S. Press DEL once on each to unmap them and try again?

FoxhackDN:
> Are you certain it's not a conflict of default inputs? -nojoy won't disable the emulated joystick, that cmd line option is for the host joystick.

Confirmed... And it uses the same input keys for both players (A/D/Space), too. I'll rewrite the note to mention this, and remove the support=partial tag later today.

happppp:
For drivers with joysticks + keyboard, a conflict between default inputs is just the way it is, and I don't think it needs per-game notes. Maybe a note in the driver itself, but even then, every driver with kb+joy has this.

FoxhackDN:
> For drivers with joysticks + keyboard, a conflict between default inputs is just the way it is, and I don't think it needs per-game notes. Maybe a note in the driver itself, but even then, every driver with kb+joy has this.

This was the only game that seems to have this issue, though - so I figured a note was fine, people are gonna think it's a bug when it's not. A normal user wouldn't know about these conflicts. :)

cuavas:
A normal user who uses any of the home computers learns pretty quickly.

FoxhackDN:
So you want me to remove the note?

angelosa:
If it resolved as a `(btanb)` just add it in the note.

hash/rx78_cart.xml
```diff
@@ -109,14 +113,19 @@ The monitor will expect you to load a program from cassette.
...
-	<software name="burger">
-		<description>Burger Shop</description>
+	<software name="cannonbl">
+		<description>Cannon Ball</description>
 		<year>1983</year>
 		<publisher>Bandai</publisher>
-		<info name="usage" value="Z and X to start and move; Space to pick up and put down" />
+		<notes><![CDATA[
+Keyboard controls do not work correctly until you unmap Joystick 2's default keys. (btanb)
+]]></notes>
```
cuavas, line 122: The wording really isn’t good if you’re going to leave the comment in.

* Firstly, it isn’t a bug, it’s just that default game control assignments overlap with default keyboard assignments.
* Secondly, you don’t need to clear the assignments for the second controller, just assign it to something that doesn’t conflict with the keyboard.
* Thirdly, as this affects all emulated systems with keyboards, anyone who actually emulates home computers regularly will likely have set this up in general input assignments already and won’t encounter an issue at all.

Issue #12181, opened 2024-03-25
Documentation search is currently broken


sairuk:
### MAME version

0.263

### System information

docs.mamedev.org

Accessed through Firefox 120, although this is an upstream issue with the rtd project.

### Incorrect behaviour

Searching documentation on docs.mamedev.org does not complete instead reporting errors to the js console in developer tools

```
Uncaught ReferenceError: jQuery is not defined
    <anonymous> https://docs.mamedev.org/search.html?q=test&check_keywords=yes&area=default:129
search.html:129:7
```
```
Uncaught ReferenceError: jQuery is not defined
    <anonymous> https://docs.mamedev.org/search.html?q=test&check_keywords=yes&area=default:134
search.html:134:5
```

### Expected behaviour

Users are able to successfully search documentation

### Additional details

This appears to be an upstream issue with the sphinx_rtd_theme (half) removing jquery support https://github.com/readthedocs/sphinx_rtd_theme/issues/1546 other projects appear to have implemented workarounds, raising this for visibility on both projects.

Firehawke:
Assigning this to myself and monitoring the RTD issue to see how they intend to handle this.

Firehawke:
It doesn't look like RTD is going to fix their botched 2.0.0 release any time soon. There are a number of other smaller things I'd noticed that were concerning in 2.0.0, but the broken search was the primary problem. I've replaced the JQuery call on our local RTD version with a standard script tag, which is what I expect that the final upstream solution will be.

Search is functional again. I'm going to need to make a note to keep an eye out for future RTD/Sphinx updates to TRY to avoid breakages like this from happening again.

PR #12253, opened 2024-04-15
Update RTD to 2.0.0 and fix search

Firehawke:
**DO NOT MERGE THIS WITHOUT REVIEW**

*Important: Due to the way that the RTD theme imports the search script (it's different between localhost and online due to ajax handling) I was unable to completely test this. Be ready to back this back out immediately if it breaks in production. It SHOULD work, but I can't 100% guarantee it and it's not something I can easily test..*

As for what changed:

Three issues needed to be resolved:

First, Sphinx no longer inherently carries JQuery, so any RTD calls to JQuery broke (e.g. search).
Secondly, RTD 2.0.0 came out back in November, and that's going to ne needed going forward for future Sphinx compatibility. 
Unfortunately, RTD 2.0.0 still tries to call out to JQuery for search, as noted in https://github.com/mamedev/mame/issues/12181

The best workaround as things stand is to update our RTD to V2.0.0, make necessary adjustments to the theme and download template I created back in 2021, and hold things in this place until the RTD team actually fixes their own stuff upstream. Considering it's been three months since the last commit on their theme repo and no response to the issue filed on the search issue, it could be a while before any kind of fix happens.

I'm not entirely thrilled with the results, but it's better than having no search at all (and potentially a completely broken docs site with future Sphinx versions). I'll take a look at refactoring the download panel template at some point in the near future; the changes I had to make were related to the old version of the template completely breaking with the new theme version.


docs/themes/sphinx_rtd_theme/__init__.py
```diff
@@ -31,33 +31,47 @@ def config_initiated(app, config):
...
-        if not app.config.html_experimental_html5_writer:
-            logger.warning("'html4_writer' is deprecated with sphinx_rtd_theme")
-    else:
-        if app.config.html4_writer:
-            logger.warning("'html4_writer' is deprecated with sphinx_rtd_theme")
+        logger.error("Python 2 is not supported with sphinx_rtd_theme, update to Python 3.")
+
+    app.require_sphinx('5.0')
+    if app.config.html4_writer:
+        logger.error("'html4_writer' is not supported with sphinx_rtd_theme.")
```
angelosa, line 47: 1. the URL is outdated, resolves to 404;
2. https://github.com/readthedocs/sphinx_rtd_theme diverges greatly, including giving back Dockerfile option for local compiling. So it's not really using `__version__ = '2.0.1alpha1'`

Firehawke: This is a direct pull off their git repo for RTD 2.0.0, no other changes were made.

PR #12254, opened 2024-04-15
uknc.cpp: hook up cassette and floppy

shattered:


src/devices/bus/qbus/uknc_kmd.h
```diff
@@ -0,0 +1,46 @@
...
+***************************************************************************/
+
+#ifndef MAME_BUS_QBUS_UKNC_KMD_H
+#define MAME_BUS_QBUS_UKNC_KMD_H
+
+#pragma once
+
+#include "machine/1801vp128.h"
+
+#include "qbus.h"
```
cuavas, line 16: Local headers should come first.  It helps avoid headers that don’t `#include` everything they need.

src/devices/bus/qbus/uknc_kmd.cpp
```diff
@@ -0,0 +1,81 @@
...
+    UKNC floppy controller (device driver MZ.SYS)
+
+***************************************************************************/
+
+#include "emu.h"
+
+#include "machine/pdp11.h"
+#include "formats/bk0010_dsk.h"
+
+#include "uknc_kmd.h"
```
cuavas, line 14: The translation unit’s header should be the first thing after the prefix header.

PR #12255, opened 2024-04-15
bmc/koftball: decode bitmap layer

angelosa:


src/mame/bmc/koftball.cpp
```diff
@@ -238,8 +212,10 @@ uint32_t koftball_state::screen_update(screen_device &screen, bitmap_ind16 &bitm
 	girl select after coin up   prev screen prev screen over        under       0x13        0x3a
 	*/
 
+	draw_pixlayer(*m_pixbitmap, cliprect);
+
 	if (BIT(m_priority, 3))
-		copyscrollbitmap_trans(bitmap, *m_pixbitmap, 0, 0, 0, 0, cliprect, 0);
+		copyscrollbitmap(bitmap, *m_pixbitmap, 0, 0, 0, 0, cliprect);
 
 	if (BIT(m_gfx_ctrl, 5)) // TODO: or bit 1?
```
angelosa, line 220: Note: I didn't put a great effort in checking how games behave, but this is concealed by the lack of transparent pens in tilemaps 1 & 3.

src/mame/bmc/koftball.cpp
```diff
@@ -143,58 +142,13 @@ TILE_GET_INFO_MEMBER(koftball_state::get_tile_info)
...
-	{
-		pen = (data >> 12) & 0xf; m_pixbitmap->pix(y, x + 0) = pen ? pixpal + pen : 0;
-		pen = (data >>  8) & 0xf; m_pixbitmap->pix(y, x + 1) = pen ? pixpal + pen : 0;
-	}
-	if (ACCESSING_BITS_0_7)
-	{
-		pen = (data >>  4) & 0xf; m_pixbitmap->pix(y, x + 2) = pen ? pixpal + pen : 0;
-		pen = (data >>  0) & 0xf; m_pixbitmap->pix(y, x + 3) = pen ? pixpal + pen : 0;
-	}
-}
```
angelosa, line 168: I got rid of this because it's hardly adaptable, the performance penalty for drawing 512x256 in screen_update rather than the handler isn't all that noticeable in a stock 68k game.
It's also imho the wrong way of doing it, if anybody really bothers the better approach is to just use a `is_dirty` boolean flag attached to the handlers.

Osso13:
Thanks, now jxzh test mode and bookkeeping screens look way better.
However, if you let the attract run you'll see this causes a black screen after the 'demon' screen and until the attract cycle restarts (so the blacked out screens are 'slot with road signs', 'Chinese lanterns', 'slot with numbers', 'pirates' and 'tile race'). Guess it may be due to me failing at some priority / transparency stuff.

angelosa:
> Thanks, now jxzh test mode and bookkeeping screens look way better. However, if you let the attract run you'll see this causes a black screen after the 'demon' screen and until the attract cycle restarts (so the blacked out screens are 'slot with road signs', 'Chinese lanterns', 'slot with numbers', 'pirates' and 'tile race'). Guess it may be due to me failing at some priority / transparency stuff.

Fixed in a bit.

src/mame/bmc/koftball.cpp
```diff
@@ -207,7 +166,31 @@ void koftball_state::video_start()
...
+	{
+		const u16 pitch = y * 0x80;
+		for (int x = cliprect.min_x; x <= cliprect.max_x >> 2; x++)
+		{
+			const u16 tile_data = m_pixram[(pitch + x) & 0xffff];
+			for (int xi = 0; xi < 4; xi ++)
+			{
+				const u8 nibble = (tile_data >> ((3 - xi) * 4)) & 0xf;
+				if (nibble)
+					bitmap.pix(y, (x << 2) + xi) = pix_bank | nibble;
```
angelosa, line 190: Note: needs this switch because `copyscrollbitmap_trans` on its own is unable to break on a pal base. It's likely doable with `prio*` / `primask*` methods but I've forgot/don't know the incantation off the bat ...

Osso13 (approved):
Awesome, the screens now look just like the original references. Thanks!

src/mame/bmc/koftball.cpp
```diff
@@ -207,7 +158,31 @@ void koftball_state::video_start()
 	m_tilemap[0]->set_transparent_pen(0);
 	m_tilemap[2]->set_transparent_pen(0);
 
-	m_pixbitmap = std::make_unique<bitmap_ind16>(0x400, 0x200);
+	m_tilemap[1]->set_transparent_pen(0);
+	m_tilemap[3]->set_transparent_pen(0);
+
+	m_pixbitmap = std::make_unique<bitmap_ind16>(512, 256);
```
cuavas, line 164: Is there a reason for dynamically allocating the bitmap object itself?  The bitmap classes don’t allocate storage until you call `resize` or `allocate` – this adds an extra layer of indirection.

angelosa: I just reused what was already available in the driver for convenience, and didn't made great efforts in testing this beyond a couple service mode items (i.e. if there's a video register that controls pitch width or if it uses this as priority mask).

PR #12264, opened 2024-04-17
subsino/subsino.cpp: Add a new bootleg of 'Treasure Bonus' named 'Lucky Seven'

clawgrip:
New not working clone
--------------------------
Lucky Seven [Taksangs, charlie]


cuavas:
If I understand correctly:
* In the video, the tiles graphics themselves are displaying correctly, but the layout appears wrong and the background around the crown looks like it shouldn’t be opaque black.
* This hardware has ROM-based tiles, but RAM-based tilemaps.  The layout is written to RAM by the game program.
* If a tile ROM was bad, it would cause corrupted tile graphics, which is not the case here.

Someone else can tell me if I’m missing something, but this doesn’t look like a bad tile ROM to me.  It looks more like mismatched program and tile ROMs, a poor effort at a hack, or a buggy development version.

PR #12265, opened 2024-04-17
Add a skeleton for a Spanish touch games machine from Comatel

clawgrip:
New not working machine
-----------------------------
Onliner Touch Party [Cuclis, Oskarito]


src/mame/skeleton/conliner.cpp
angelosa, line 1: Belongs to [src/mame/pc folder](https://github.com/mamedev/mame/blob/master/src/mame/pc/README.md)

clawgrip: Done, thanks!

PR #12269, opened 2024-04-17
sinclair/sprinter.cpp: update

holub:
- accelerator writes through page fd only in COPY mode
- io access 1f (z84 int) available through replacing to 0f
- hold IRQ for 32 clocks (unscaled)

### DIFF
--- a/src/mame/sinclair/sprinter.cpp
+++ b/src/mame/sinclair/sprinter.cpp
@@ -123,6 +123,7 @@ protected:
 	void update_cpu();
 
 	virtual TIMER_CALLBACK_MEMBER(irq_on) override;
+	virtual TIMER_CALLBACK_MEMBER(irq_off) override;
 	TIMER_CALLBACK_MEMBER(cbl_tick);
 
 	u32 screen_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect);
@@ -258,6 +259,7 @@ private:
 	u8 m_isa_addr_ext;
 	std::pair<s8, s8> m_hold;
 	u8 m_kbd_data_cnt;
+	bool m_in_out_cmd;
 
 	bool m_ata_selected; // 0-primary, 1-secondary
 	u8 m_ata_data_latch;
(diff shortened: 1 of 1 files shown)


src/mame/sinclair/sprinter.cpp
```diff
@@ -1506,10 +1518,19 @@ void sprinter_state::do_cpu_wait(bool is_io)
...
+		m_irq_off_timer->adjust(attotime::from_ticks(32, m_maincpu->unscaled_clock()));
+	}
 	update_int(false);
 }
 
+TIMER_CALLBACK_MEMBER(sprinter_state::irq_off)
+{
+	m_maincpu->set_input_line(INPUT_LINE_IRQ0, CLEAR_LINE);
+}
+
```
cuavas, line 1533: Using `HOLD_LINE` to stop the interrupt pulse from `irq_on` clearing the interrupt from `cbl_tick` seems like a bit of a hack.  Can you keep the state of the two sources and combine them to assert/clear the interrupt line to the CPU?

holub: done

src/mame/sinclair/sprinter.cpp
```diff
@@ -1506,10 +1521,20 @@ void sprinter_state::do_cpu_wait(bool is_io)
...
 TIMER_CALLBACK_MEMBER(sprinter_state::irq_on)
 {
-	m_maincpu->pulse_input_line(INPUT_LINE_IRQ0, attotime::from_ticks(26, m_maincpu->clock()));
+	if (!m_hold_irq)
+	{
+		m_maincpu->set_input_line(INPUT_LINE_IRQ0, ASSERT_LINE);
+		m_irq_off_timer->adjust(attotime::from_ticks(32, m_maincpu->unscaled_clock()));
+	}
 	update_int(false);
 }
```
cuavas, line 1530: Should this interrupt continue to be held for the full 32 cycles even if there’s an interrupt acknowledge cycle, or should an interrupt acknowledge cycle clear this interrupt as well as the other one?  (The implementation currently does the latter – clears both kinds of interrupt on an interrupt acknowledge cycle.)

src/mame/sinclair/sprinter.cpp
```diff
@@ -1506,10 +1521,20 @@ void sprinter_state::do_cpu_wait(bool is_io)
...
 
+TIMER_CALLBACK_MEMBER(sprinter_state::irq_off)
+{
+	m_hold_irq = 0;
+	m_maincpu->set_input_line(INPUT_LINE_IRQ0, CLEAR_LINE);
+}
```
cuavas, line 1536: This can be called two ways now – from `m_irq_off_timer` expiring, and from an interrupt acknowledge cycle on the main CPU.  If it’s called from an interrupt acknowledge cycle, should it cancel the timer?  If it doesn’t, this sequence of events can happen:

* `irq_on` asserts interrupt and starts `m_irq_off_timer`
* Interrupt acknowledge cycle causes `irq_off` to be called, clearing interrupt
* `m_irq_off_timer` fires, causing `irq_off` to be called again

It’s probably harmless given what it’s doing, but it will go through another timer update and then go through the scheduler again to clear the IRQ line.

holub:
Please hold on merging this. I need to double check this IRQ behaviour

holub:
False alarm. It's good to go.

PR #12273, opened 2024-04-18
pinball/decodmd*.cpp: Add missing item to save state, Cleanups

cam900:
Initialize member variables
Reduce unnecessary bank finders, Reduce unnecessary DUMMY_TAG
Reduce literal tag usage
Constantize variables
Suppress side effects for debugger reads


cuavas:
pinball/de_3.cpp should use real PIA devices so the DMDs can be hooked up to them in machine configuration and the memory handlers checking for each kind of DMD can be eliminated.

PR #12275, opened 2024-04-19
heathkit/tlb.cpp: Various file cleanup

mgarlanger:
General cleanup including
- Removed method declarations that were not used/implemented.
- Changed to use u8/u16/u32.
- Add consistent horizontal spacing for readability.
- Removed some duplicate comments.


src/mame/heathkit/tlb.cpp
```diff
@@ -141,26 +148,26 @@ device_heath_tlb_card_interface::device_heath_tlb_card_interface(const machine_c
...
+	m_crtc(*this,         "crtc"),
+	m_p_videoram(*this,   "videoram"),
+	m_p_chargen(*this,    "chargen"),
+	m_config(*this,       "CONFIG"),
+	m_ace(*this,          "ins8250"),
+	m_beep(*this,         "beeper"),
+	m_mm5740(*this,       "mm5740"),
+	m_kbdrom(*this,       "keyboard"),
+	m_kbspecial(*this,    "MODIFIERS"),
 	m_repeat_clock(*this, "repeatclock")
```
cuavas, line 171: The trouble with tabulating everything like this is that if you introduce a data member with a longer name or remove the data member with the longest name, every line in the block needs to change.

Also, in this case you’ve tabulted the tags, but not the `*this` references.

src/mame/heathkit/tlb.h
```diff
@@ -33,19 +33,19 @@ class device_heath_tlb_card_interface : public device_interface
...
-	virtual void sigma_window_lo_addr_w(uint8_t val) {}
-	virtual void sigma_window_hi_addr_w(uint8_t val) {}
+	virtual void sigma_ctrl_w(u8 data)          {}
+	virtual u8   sigma_ctrl_r()                 { return 0x00; }
+	virtual u8   sigma_video_mem_r()            { return 0x00; }
+	virtual void sigma_video_mem_w(u8 val)      {}
+	virtual void sigma_io_lo_addr_w(u8 val)     {}
+	virtual void sigma_io_hi_addr_w(u8 val)     {}
+	virtual void sigma_window_lo_addr_w(u8 val) {}
+	virtual void sigma_window_hi_addr_w(u8 val) {}
```
cuavas, line 48: It’s the same with these – it’s prone to either ending up looking ragged again, or requiring all the lines in the block to change for a small functional change.

src/mame/heathkit/tlb.h
```diff
@@ -209,14 +215,14 @@ class heath_watz_tlb_device : public heath_tlb_device
...
-	heath_ultra_tlb_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock = 0);
+	heath_ultra_tlb_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock = 0);
 
 protected:
 	virtual const tiny_rom_entry *device_rom_region() const override;
 	virtual ioport_constructor device_input_ports() const override;
 	virtual void device_add_mconfig(machine_config &config) override;
 
-	void mem_map(address_map &map);
+	void         mem_map(address_map &map);
```
cuavas, line 225: See here’s an example where you’ve arbitrarily lined up `device_add_mconfig` and `mem_map`, but you haven’t lined up `device_rom_region` and `device_input_ports` with them.  It’s a bit concerning if even the initial tabulation is inconsistent.

cuavas (commented):
Personally, I’m not a fan of tabulating initialisations, assignments and inline functions where it isn’t naturally tabular data involved.  It either ends up ragged again quickly, or requires large numbers of lines to change for relatively small real changes.

I’m fine with tabulating stuff that’s naturally somewhat tabular, like I/O ports, ROM load commands, device type declarations and large blocks of variable declarations.  It’s just that tabulating other things tends to lead to churn for minimal improvements in readability at best.

I’ll let other people take a look and chime in on this.

src/mame/heathkit/tlb.cpp
mgarlanger: Ok, I undid these changes and the other ones you pointed out.

PR #12285, opened 2024-04-20
heathkit/h89.cpp and h19.cpp: Temporary remove gp19 option for tlbc slot

mgarlanger:
Currently gp19 option is not working. I narrowed it down to commit 1af40e3a1c828e8394d51b8f3b611d39a2b4bb5e, but it is not obvious what causing the system from not coming up. Temporarily remove the option until issue can be resolved.

### DIFF
--- a/src/mame/heathkit/h19.cpp
+++ b/src/mame/heathkit/h19.cpp
@@ -38,7 +38,8 @@ private:
 static void tlb_options(device_slot_interface &device)
 {
 	device.option_add("heath", HEATH_TLB);
-	device.option_add("gp19", HEATH_GP19);
+	// temporary disable GP19 due not initializing properly after commit 1af40e3a1c828e8394d51b8f3b611d39a2b4bb5e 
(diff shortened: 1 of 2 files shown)


rb6502:
In general I don't think disabling it is worthwhile as it just adds friction to fixing it.  I'd suggest instead having the gp19 device overriding imperfect_features() and returning feature::GRAPHICS since it's a broken video add-on.

mgarlanger:
> In general I don't think disabling it is worthwhile as it just adds friction to fixing it. I'd suggest instead having the gp19 device overriding imperfect_features() and returning feature::GRAPHICS since it's a broken video add-on.

GP19 already has that flag set: https://github.com/mamedev/mame/blob/master/src/mame/heathkit/tlb.h#L230 due interlace not working which is required for 50-line mode, and a few other issues with the text display. But this regression causes everything including the keyboard to not work. I was looking to see if I could add a `NOT_WORKING` tag, but couldn't find any way to do that.
If you think that the imperfect graphics is enough, we can close this PR.

PR #12286, opened 2024-04-21
atlus/cave.cpp: Added Jumbo Godzilla.

Mikejmoffitt:
New Working Software: Jumbo Godzilla [Toho / Namco]

Jumbo Godzilla runs on hardware nearly identical to Pac-Slot, with the board moniker N-42 EM VIDEO PCB stamped on it.

Notable differences compared to N-44:

* The second OKI is used and has its ROM socket populated
* Instead of a GAL16V8 at U1 controlling the memory map, a simple 74LS138 is installed (albeit socketed).

All ROM names are based directly upon the labels on the ROMs on the PCB. All GALs were found to be locked and thus remain undumped.

I do not have access to an original cabinet, so I can not 100% verify the accuracy of the lamp and motor control outputs. My PCB has damage, possibly from overcurrent on one of these controls, so at this time it is not something I can use for further testing. Furthermore, I do not have a wiring harness for it, so no clues may be offered there either.

This game has both an operator menu as well as a self-check screen, even though the PCB itself has no DIP switches.


src/mame/atlus/cave.cpp
```diff
@@ -21,6 +21,7 @@ Other        :  93C46 EEPROM
 Year + Game               License       PCB         Tilemaps        Sprites         Other
 -----------------------------------------------------------------------------------------
 94 Mazinger Z             Banpresto     BP943A      038 9335EX706   013 9341E7009   Z80
+95 Jumbo Godzilla         Toho          N-42 EM     038 9345E7008   013 9442WX002
```
Osso13, line 24: The rest of the entries in this table are ordered  by year, so please put this at least below Power Instinct 2

src/mame/atlus/cave.h
```diff
@@ -71,6 +71,7 @@ class cave_state : public driver_device
 	void paccarn(machine_config &config);
 	void paceight(machine_config &config);
 	void pacslot(machine_config &config);
+	void jumbogod(machine_config &config);
```
Osso13, line 74: The machine_config methods in this section are in alphabetical order, so please move the jumbogod one accordingly

src/mame/atlus/cave.h
```diff
@@ -247,6 +249,7 @@ class cave_state : public driver_device
 	void paccarn_map(address_map &map);
 	void paceight_map(address_map &map);
 	void pacslot_map(address_map &map);
+	void jumbogod_map(address_map &map);
```
Osso13, line 252: The methods in this section are in alphabetical order, too.

src/mame/mame.lst
```diff
@@ -2204,6 +2204,7 @@ nmaster                         // (c) 1995 Banpresto / Pandorabox
 paccarn                         // (c) 1996 Namco
 paceight                        // (c) 1996 Namco
 pacslot                         // (c) 1996 Namco
+jumbogod                        // (c) 1995 Namco / Toho
```
Osso13, line 2207: the lists in mame.lst are supposed to be in alphabetical order (some exception has slipped in, but should be avoided).
Can you please put jumbogod in the right order?

Osso13 (approved):
Very nice, thanks!
I left some notes that may seem or even be a bit pedantic, but trying to keep consistency, when possible, is IMHO a good thing.
Another pet peeve of mine are IMHO superfluous comments like
	map(0x100000, 0x10ffff).ram();                                                                                // RAM
but I do realize that the whole driver is like that, so I don't think you should remove them.

src/mame/atlus/cave.cpp
```diff
@@ -1229,6 +1230,72 @@ void cave_state::paccarn_map(address_map &map)
...
+void cave_state::jumbogod_leds_w(u8 data)
+{
+	// Numbering of attack button light outputs is guessed based upon the
+	// ordering of the button inputs in the test menu; in-game they all
+	// appear to be activated at once, so disambiguation is challenging.
+	m_led_outputs[0] = data & 0x0001;  // tai atari / body slam
+	m_led_outputs[1] = data & 0x0002;  // nissen kougeki / atomic breath
+	m_led_outputs[2] = data & 0x0004;  // shippo kougeki / tail attack
+	m_led_outputs[3] = data & 0x0008;  // nissen kougeki / atomic breath
+	m_led_outputs[4] = data & 0x0010;  // motor & spinning lamps
```
cuavas, line 1277: These should use 1/0 values - please use the `BIT` helper for this.

src/mame/atlus/cave.cpp
```diff
@@ -1229,6 +1230,72 @@ void cave_state::paccarn_map(address_map &map)
...
+	map(0x400068, 0x400069).w("watchdog", FUNC(watchdog_timer_device::reset16_w));                   // Watchdog
+	map(0x500000, 0x500005).w(m_tilemap[0], FUNC(tilemap038_device::vregs_w));                       // Layer 0 Control
+	map(0x600000, 0x60ffff).ram().w(m_palette[0], FUNC(palette_device::write16)).share("palette.0"); // Palette
+	map(0x700000, 0x700001).portr("IN0");                                                            // Inputs + EEPROM
+	map(0x700002, 0x700003).portr("IN1");                                                            // Inputs
+	map(0x800001, 0x800001).rw("oki1", FUNC(okim6295_device::read), FUNC(okim6295_device::write));   // M6295
+	map(0x900001, 0x900001).rw("oki2", FUNC(okim6295_device::read), FUNC(okim6295_device::write));   // M6295
+	map(0xc00001, 0xc00001).w(FUNC(cave_state::jumbogod_leds_w));                                    // Leds
+	map(0xe00001, 0xe00001).w(FUNC(cave_state::tjumpman_eeprom_w));                                  // EEPROM
+}
```
cuavas, line 1297: This is almost identical to `pacslot_map` – please call `pacslot_map` and then just install the two different handlers you need.

src/mame/atlus/cave.cpp
```diff
@@ -1756,6 +1823,27 @@ static INPUT_PORTS_START( paccarn )
...
+        PORT_START("IN1")
+        PORT_SERVICE( 0x01, IP_ACTIVE_LOW ) // must stay on during service mode
+        PORT_BIT( 0x06, IP_ACTIVE_LOW, IPT_UNKNOWN )
+        PORT_CONFNAME( 0x08, 0x08, "Self Test" )
+        PORT_CONFSETTING(    0x08, DEF_STR( Off ) )
+        PORT_CONFSETTING(    0x00, DEF_STR( On ) )
+        PORT_BIT( 0x10, IP_ACTIVE_LOW, IPT_BUTTON2   ) PORT_NAME( "Nissen Kougeki (Atomic Breath)" )
+        PORT_BIT( 0x20, IP_ACTIVE_LOW, IPT_BUTTON4   ) PORT_NAME( "Nissen Kougeki (Atomic Breath)" )
+        PORT_BIT( 0xC0, IP_ACTIVE_LOW, IPT_UNKNOWN )
+INPUT_PORTS_END
```
cuavas, line 1845: It’s going to confuse users if there are two identically named buttons.

Also, you should use `PORT_SERVICE` for the test mode switch.

cuavas (changes requested):

src/mame/atlus/cave.cpp
Mikejmoffitt: In this case, PORT_SERVICE is already being used for the operator test menu in line with how it seems to be used for many titles. The game has a separate diagnostic self test mode, operated by another switch. Do you mean that they should be switched? In the current configuration, striking F2 brings up the operator menu with settings, which is the behavior I expected to see.

src/mame/atlus/cave.cpp
Mikejmoffitt: Any objection to adding `oki2` to pacslot_map and consolidating the two? I checked and pacslot seems to initialize the second (unused) OKI during startup at the same address. I'd then just have to install the alternate handler for the lamp I/O.

angelosa: Do you know if pacslot has the extra OKI on board?

cuavas: If the second OKI isn’t populated on `pacslot`, it shouldn’t be added.  Many games do run code code to initialise unpopulated peripherals.

src/mame/atlus/cave.cpp
cuavas: Oh, sorry, I missed that `PORT_SERVICE` is already being used.  Can you use another one of the service controls?  Is `IPT_SERVICE1` available?  For example something like:
```
        PORT_BIT( 0x08, IP_ACTIVE_LOW, IPT_SERVICE1 ) PORT_TOGGLE PORT_NAME( "Self Test" )
```

src/mame/atlus/cave.cpp
Mikejmoffitt: The peripheral itself is present, it just has no ROM data attached. So, it wouldn't make noise properly if commanded, but it would respond to register reads, etc. It is physically on the PCB on the PCB though.

src/mame/atlus/cave.cpp
```diff
@@ -21,7 +21,9 @@ Other        :  93C46 EEPROM
 Year + Game               License       PCB         Tilemaps        Sprites         Other
 -----------------------------------------------------------------------------------------
 94 Mazinger Z             Banpresto     BP943A      038 9335EX706   013 9341E7009   Z80
+95 Jumbo Godzilla         Toho          N-42 EM     038 9345E7008   013 9442WX002
 94 Power Instinct 2       Atlus         AT047G2-B   038 9429WX709   013 9341E7009   Z80 NMK 112
+95 Jumbo Godzilla         Toho          N-42 EM     038 9345E7008   013 9442WX002
 95 Gogetsuji Legends      Atlus         AT047G2-B   038 9429WX709   013 9341E7009   Z80 NMK 112
```
cuavas, line 27: Jumbo Godzilla seems to be duplicated here now.

PR #12290, opened 2024-04-21
t11: reuse z80_daisy_chain for QBus vectored interrupts; forward INIT

shattered:
signal to all cards on QBus.


shattered:
@ajrhacker -- could you review this?

rb6502:
We might want to separate the daisy chain mechanism from the Z80 if it turns out to be useful for other CPUs like this, but the changes here themselves look unobjectionable.

PR #12295, opened 2024-04-22
misc/coolpool.cpp: Cleanups

cam900:
Split driver state to reduce optional finders
Use palette_device for amerdart palette
Suppress side effects for debugger read
Constantize variables
Use logmacro.h for logging functions
Use C++ style comments for single line comments
Use std::swap for swap function

### DIFF
--- a/src/mame/misc/coolpool.cpp
+++ b/src/mame/misc/coolpool.cpp
@@ -41,7 +41,7 @@
 #include <algorithm>
 
 #define LOG_PORT    (1U << 1)
-#define LOG_INPUT   (1U << 1)
+#define LOG_INPUT   (1U << 2)
 
 #define LOG_ALL     (LOG_PORT | LOG_INPUT)
 
@@ -164,6 +164,7 @@ void amerdart_state::machine_start()
 {
 	coolpool_base_state::machine_start();
 
+	m_lastresult = 0xffff;
 	save_item(NAME(m_old_cmd));
 }
 
@@ -1103,11 +1104,6 @@ ROM_END
  *
  *************************************/
 
-void amerdart_state::init_amerdart()
-{
-	m_lastresult = 0xffff;
-}
-
 
 void _9ballsht_state::init_9ballsht()
 {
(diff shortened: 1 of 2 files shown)


src/mame/misc/coolpool.h
```diff
@@ -10,14 +10,15 @@
...
 {
 public:
-	coolpool_state(const machine_config &mconfig, device_type type, const char *tag)
+	coolpool_base_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
 		, m_dsp(*this, "dsp")
-		, m_tlc34076(*this, "tlc34076")
 		, m_main2dsp(*this, "main2dsp")
 		, m_dsp2main(*this, "dsp2main")
```
cuavas, line 23: If this is an abstract class, the contructor should be `protected`.

src/mame/misc/coolpool.cpp
```diff
@@ -38,7 +38,18 @@
...
+#define LOG_PORT    (1U << 1)
+#define LOG_INPUT   (1U << 1)
+
+#define LOG_ALL     (LOG_PORT | LOG_INPUT)
+
+#define VERBOSE (0)
+#include "logmacro.h"
+
+#define LOGPORT(...)  LOGMASKED(LOG_PORT,  __VA_ARGS__)
+#define LOGINPUT(...) LOGMASKED(LOG_INPUT, __VA_ARGS__)
```
cuavas, line 52: You’ve duplicated the same value for two log categories.

src/mame/misc/coolpool.cpp
```diff
@@ -1039,36 +1103,21 @@ ROM_END
...
-	save_item(NAME(m_iop_romaddr));
-}
-
-
-
-void coolpool_state::init_amerdart()
+void amerdart_state::init_amerdart()
 {
 	m_lastresult = 0xffff;
 }
```
cuavas, line 1109: Does this really need to be a driver initialisation function?  AmeriDarts is the only machine configuration in `amerdart_state`, so you could simply override `machine_start` and set the initial value.

cuavas (changes requested):

PR #12297, opened 2024-04-22
midway/midyunit.cpp: Further cleanups

cam900:
Reduce unnecessary MACHINE_RESET_MEMBER and VIDEO_START_MEMBER Split machine config to reduce duplicates
Split driver state class to reduce optional finders Add missing items in save state

### DIFF
--- a/src/mame/midway/midyunit.h
+++ b/src/mame/midway/midyunit.h
@@ -168,7 +168,7 @@ public:
 protected:
 	virtual void machine_reset() override;
 
-	optional_device<williams_cvsd_sound_device> m_cvsd_sound;
+	required_device<williams_cvsd_sound_device> m_cvsd_sound;
 
 	uint8_t *m_cvsd_protection_base = nullptr;
 
--- a/src/mame/midway/midyunit.h
+++ b/src/mame/midway/midyunit.h
@@ -21,16 +21,13 @@
 #include "emupal.h"
 
 
-class midyunit_state : public driver_device
+class midyunit_base_state : public driver_device
 {
-public:
-	midyunit_state(const machine_config &mconfig, device_type type, const char *tag)
+protected:
+	midyunit_base_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
 		, m_palette(*this, "palette")
-		, m_narc_sound(*this, "narcsnd")
-		, m_cvsd_sound(*this, "cvsd")
-		, m_adpcm_sound(*this, "adpcm")
 		, m_nvram(*this, "nvram")
 		, m_paletteram(*this, "paletteram")
 		, m_gfx_rom(*this, "gfx_rom", 0x800000, ENDIANNESS_BIG)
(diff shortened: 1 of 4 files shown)


src/mame/midway/midyunit_m.cpp
```diff
@@ -583,32 +562,54 @@ void midyunit_state::init_totcarn()
...
+
+	m_left_flash.resolve();
+	m_right_flash.resolve();
+	m_left_gun_recoil.resolve();
+	m_right_gun_recoil.resolve();
+	m_left_gun_green_led.resolve();
+	m_left_gun_red_led.resolve();
+	m_right_gun_green_led.resolve();
+	m_right_gun_red_led.resolve();
 }
```
cuavas, line 594: Can you have this call its immediate direct base `midyunit_adpcm_state::machine_start()`?

src/mame/midway/midyunit_m.cpp
```diff
@@ -583,32 +562,54 @@ void midyunit_state::init_totcarn()
...
+{
+	m_cvsd_sound->reset_write(1);
+	m_cvsd_sound->reset_write(0);
+}
+
+void midyunit_adpcm_state::machine_reset()
+{
+	m_adpcm_sound->reset_write(1);
+	m_adpcm_sound->reset_write(0);
+}
```
cuavas, line 612: Can you have these call `midyunit_base_state::machine_reset()` in case it has something important added?

src/mame/midway/midyunit_m.cpp
```diff
@@ -628,23 +629,41 @@ void midyunit_state::midyunit_sound_w(offs_t offset, uint16_t data, uint16_t mem
...
+	if (offset)
+	{
+		logerror("%08X:Unexpected write to sound (hi) = %04X\n", m_maincpu->pc(), data);
+		return;
+	}
+
+	// call through based on the sound type
+	if (ACCESSING_BITS_0_7 && ACCESSING_BITS_8_15)
+	{
+		m_cvsd_sound->reset_write((~data & 0x100) >> 8);
```
cuavas, line 647: You could use `BIT(~data, 8)` here to avoid needing two constants that have to line up.

src/mame/midway/midyunit_m.cpp
```diff
@@ -628,23 +629,41 @@ void midyunit_state::midyunit_sound_w(offs_t offset, uint16_t data, uint16_t mem
...
+	if (offset)
+	{
+		logerror("%08X:Unexpected write to sound (hi) = %04X\n", m_maincpu->pc(), data);
+		return;
+	}
+
+	// call through based on the sound type
+	if (ACCESSING_BITS_0_7 && ACCESSING_BITS_8_15)
+	{
+		m_adpcm_sound->reset_write((~data & 0x100) >> 8);
```
cuavas, line 664: Same here – you could use `BIT(~data, 8)`.

src/mame/midway/midyunit.h
```diff
@@ -115,44 +87,140 @@ class midyunit_state : public driver_device
...
+	void init_hiimpact();
+	void init_shimpact();
+	void init_smashtv();
+	void init_strkforc();
+	void init_trog();
+
+protected:
+	virtual void machine_reset() override;
+
+	optional_device<williams_cvsd_sound_device> m_cvsd_sound;
```
cuavas, line 171: You use this in `machine_reset` without checking that it’s present.  Should it be `required_device`?

PR #12300, opened 2024-04-23
sound/swp00.cpp: Improved DPCM sample decompression

TaleTN:
Through reverse engineering the ROMs of the MU50 and related sound modules I think I've figured out that the magic number that the MU50 stores along with the sample format isn't the fractional part of the loop point, but instead holds an offset and scale for DPCM decompression (which I guess makes it ADPCM).

I guess some of the related sound devices (like maybe the SWP30?) cloud also use this patch.

### DIFF
--- a/src/devices/sound/swp00.h
+++ b/src/devices/sound/swp00.h
@@ -43,6 +43,7 @@ private:
 	static const std::array<s32, 0x20> decay_linear_step;
 	static const std::array<s32, 16> panmap;
 	std::array<s32, 0x80> m_global_step;
+	static const std::array<u8, 4> m_dpcm_offset;
 	std::array<s16, 0x100> m_dpcm;
 
 	static const std::array<u32, 4> lfo_shape_centered_saw;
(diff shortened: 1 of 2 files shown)


src/devices/sound/swp00.h
```diff
@@ -43,6 +43,7 @@ class swp00_device : public device_t, public device_sound_interface, public devi
 	static const std::array<s32, 0x20> decay_linear_step;
 	static const std::array<s32, 16> panmap;
 	std::array<s32, 0x80> m_global_step;
+	static const std::array<u8, 4> m_dpcm_offset;
```
cuavas, line 46: You shouldn’t use `m_` prefixes for static data members, only instance members.  It’s also clearer to group static data members and instance data members rather than intermixing them.

TaleTN: @cuavas Ah right... Will fix (but maybe I'll wait a bit, in case there are other issues). Thanks!

cuavas: @galibert is probably the best person to review the functional changes here.

galibert:
Yeah, looks nice.  And I need to add that to the swp30 and the swx00...

TaleTN:
@galibert I could also (try to) apply the fix to the swp30 and swx00, if that helps.

BTW, do you want me to squash the current two commits into one commit?

galibert:
It's ok, github allows to squash when merging.  If you have to time to add that to the other two (swp20 does not have the audio implemented yet) I'd love that.

PR #12301, opened 2024-04-23
galaxian/galaxian.cpp: Add another Video Dens Pac-Man bootleg

clawgrip:
New working clone
---------------------
Pac-Man (Video Dens, Spanish bootleg on Galaxian hardware, set 2) [Arcadedealer aka Edgar Solé aka Dios, Recreativas.org]


clawgrip:
ROM 1 only differs on just one byte from the existing set (pacmanblv). Can anyone please check it this seems like a legit change and is not a bit rot?

Osso13:
The only difference is byte 0x14d which is 0x40 in the old dump and 0x00 here. Looking at the disassembly, it seems to be doing a series of accesses to RAM which starts at 0x4000.
immagine
It would be strange that the bootleggers intentionally changed only one of those accesses to point to ROM area, so I'd say the old dump is correct.

clawgrip:
Thanks a lot!

I've modified it to use ROM 1 from "pacmanblv" and marked it as BAD_DUMP.

PR #12305, opened 2024-04-24
sound/swp30.cpp, sound/swx00.cpp: Improved DPCM sample decompression

TaleTN:
Just like commit 48161ce9ca, but this time for SWP30 and SWX00.

@galibert As promised! :-)


cuavas:
Looks sane enough.  Plenty of time before the next release for @galibert to yell at me if he doesn’t like it.

galibert:
I like it, so there's no yelling to be expected.

  OG.

PR #12312, opened 2024-04-26
taito/taito_f3_v.cpp: regain performance after major rewrite

y-ack:
addresses my own concerns with #11811 speed regression against previous implementation.

 * switch AoS z buffers and per-pix blend info to SoA
 * allow vectorization of line blending operation
 * regains empty line optimization by tracking tilemap row usage
 * consolidate sprite framebuffers (we still pull from it multiple times for each sprite priority group)
 * other minor wins from safe logic reorderings

### DIFF
--- a/src/mame/taito/taito_f3_v.cpp
+++ b/src/mame/taito/taito_f3_v.cpp
@@ -688,7 +688,7 @@ void taito_f3_state::read_line_ram(f3_line_inf &line, int y)
 		if (const offs_t where = latched_addr(0, i)) {
 			const u16 colscroll = m_line_ram[where];
 			line.pf[i].colscroll   = colscroll & 0x1ff;
-			line.pf[i].alt_tilemap = colscroll & 0x200;
+			line.pf[i].alt_tilemap = !m_extend && colscroll & 0x200;
 			line.clip[2*(i-2) + 0].set_upper(BIT(colscroll, 12), BIT(colscroll, 13));
 			line.clip[2*(i-2) + 1].set_upper(BIT(colscroll, 14), BIT(colscroll, 15));
 		}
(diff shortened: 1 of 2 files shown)


y-ack:
<reserved to substantiate performance benefit>

src/mame/taito/taito_f3_v.cpp
```diff
@@ -958,68 +995,74 @@ inline int taito_f3_state::pivot_inf::y_index(int y) const
...
+		if (gfx.blend_mode == pri.src_blendmode[x])
 			continue; // note that layers cannot blend against the same blend mode
 
 		const int real_x = gfx.x_sample_enable ? mosaic(x, line.x_sample) : x;
 		const int gfx_x = gfx.x_index(real_x);
+
+		if constexpr (std::is_same_v<Mix, sprite_inf>) {
+			if (BIT(src[gfx_x], 10, 2) != gfx.index)
+				continue;
+		}
```
cuavas, line 1014: Can you avoid using an `if constexpr` checking the type of the layer helper here?  It goes against the spirit of object oriented programming.

For example add to `struct mixable`:
```c++
	bool skip_pixel(u16 p) const { return false; }
```

Add to `struct sprite_inf`:
```c++
	bool skip_pixel(u16 p) const { return BIT(pix, 10, 2) != index; }
```

Then here you can avoid the messiness of comparing types:
```c++
		if (gfx.skip_pixel(src[gfx_x]))
			continue;
```

(Member function and parameter names are just for example, and there may be better choices.)

src/mame/taito/taito_f3_v.cpp
```diff
@@ -1030,44 +1073,83 @@ bool taito_f3_state::mix_line(Mix &gfx, mix_pix *z, pri_mode *pri, const f3_line
...
+		r1 += r2;
+		g1 += g2;
+		b1 += b2;
+
+		r1 >>= 3;
+		g1 >>= 3;
+		b1 >>= 3;
+		r1 = std::min(r1, static_cast<u16>(255));
+		g1 = std::min(g1, static_cast<u16>(255));
+		b1 = std::min(b1, static_cast<u16>(255));
```
cuavas, line 1115: You don’t need `static_cast` – it’s more readable to use function style casts or explicit template arguments:

Like this:
```c++
		r1 = std::min(r1, u16(255));
		g1 = std::min(g1, u16(255));
		b1 = std::min(b1, u16(255));
```

Or this:
```c++
		r1 = std::min<u16>(r1, 255);
		g1 = std::min<u16>(g1, 255);
		b1 = std::min<u16>(b1, 255);
```

y-ack: _oh_

PR #12313, opened 2024-04-26
Add a skeleton for a new Recreativos Franco slot game and move all RF drivers to their own folder

clawgrip:
New not working machine
-----------------------------
Jail Break (Recreativos Franco slot machine) [Hammy, ClawGrip]


src/mame/recfranco/rfjailbrk.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+namespace {
+
+class rfjailbrk_state : public driver_device
+{
+public:
+	rfjailbrk_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_maincpu(*this, "maincpu")
+		, m_okim6376(*this, "oki")
```
angelosa, line 61: I dislike this naming. I don't need to remember this being 6376 in code, can as well be short-handed to `m_oki` as also the device tag hints.

clawgrip: You mean changing _m_okim6376(*this, "oki")_ to _m_oki(*this, "okim6376")_ ??

Thanks!!!!

angelosa: `m_oki(*this, "oki")`

clawgrip: Done, thanks!!!!

PR #12314, opened 2024-04-27
galaxian/galaxian.cpp: 'bongoa' dip switches are memory mapped

devinacker:
Compared to the parent set, `bongoa` ignores the AY input ports and reads dip switches from a memory-mapped port.

This set also hacks out the code which initializes the AY port direction bits (compare code at $2288 in both sets), which means the 'noise enable' bits are never initialized either, leaving the noise generator enabled on all three channels. Is that actually what's supposed to happen or should the noise output be disabled by default instead?

### DIFF
--- a/src/mame/galaxian/galaxian.cpp
+++ b/src/mame/galaxian/galaxian.cpp
@@ -1913,6 +1913,12 @@ void galaxian_state::bongo_map(address_map &map)
 	map(0xb800, 0xb800).mirror(0x7ff).nopw(); // written once at start
 }
 
+void galaxian_state::bongoa_map(address_map &map)
+{
+	bongo_map(map);
+	map(0xb000, 0xb000).mirror(0x07ff).portr("DSW");
+}
+
 void galaxian_state::bongo_io_map(address_map &map)
 {
 	map.global_mask(0xff);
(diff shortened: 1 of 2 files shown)


cuavas:
@clawgrip can you please consider this?  Did this Bongo set have any PCB mods?

clawgrip:
Cannot tell... The PCB was modified on many ways.

RUID2d593c87cf744106a2dddc4480b1b657
RUIDa2441f8dfbdf47ca8cd1738ba0030674
RUIDbff4c3210639469d9a964f0345a0d017

devinacker:
I don't have the redumped ROM 5 either, but that's also where all the AY8910 code lives in the original set, so there's a good chance that's what the actual difference is in this set.

clawgrip:
Hi!
  If colors are wrong, it's not a matter of changing the PROM, but decoding it properly, as we dumped the PROM from the actual PCB.

Thx.
________________________________
De: 2k2 ***@***.***>
Enviado: lunes, 6 de mayo de 2024 12:36
Para: mamedev/mame ***@***.***>
Cc: ClawGrip ***@***.***>; Mention ***@***.***>
Asunto: Re: [mamedev/mame] galaxian/galaxian.cpp: 'bongoa' dip switches are memory mapped (PR #12314)


For what it's worth, a board of this set actually exists and this is what the it may apparently should look like:
The palette looks different: same as 'bongo', but the main character is green, and Bongo and the monsters are blue! Is there any PROM that matches?
https://www.youtube.com/watch?v=ANLFKoMLPRI

PR #12317, opened 2024-04-27
mz700_cass.xml: Metadata updates

ArcadeShadow:
- Flagged non supported items, based on the MT08811 report
- Replaced countries' abbreviations by their full names.
- Lowercase on descriptive words
- Added loading instructions for the Sharp MZ-700 system

### DIFF
--- a/hash/mz700_cass.xml
+++ b/hash/mz700_cass.xml
@@ -52,7 +52,7 @@ Easidata (English Version)
 
 <!-- This is a straight recording of the tape. It might be preferrable to convert to some cooked format. Is MZF considered adequate for preservation? -->
 	<software name="1z013bswe" cloneof="1z013b">
-		<description>BASIC Interpreter 1Z-013B V1.0A (Swe)</description>
+		<description>BASIC Interpreter 1Z-013B V1.0A (Sweden)</description>
 		<year>1983</year>
 		<publisher>Sharp Corporation</publisher>
 		<part name="cass1" interface="mz_cass">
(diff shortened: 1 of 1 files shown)


hash/mz700_cass.xml
```diff
@@ -17,6 +17,21 @@ Symbolic Debugger 5Z-011D
...
 -->
 
+<!--
+	Loading instructions for games that require BASIC
+	1) Initialize the system with the BASIC Interpreter tape
+	2) Type L or LOAD then press enter
+	3) Play tape until BASIC is loaded
+	4) Change to the game's tape
+	5) Type LOAD then press enter
+	6) Play tape until the game is loaded
```
angelosa, line 27: I checked a bit: it's also RUN once return to BASIC is provided.

angelosa:
At face value I gather that anything that has PCG-700 supported/required should be marked as `supported="partial"` or `supported="no"`, respectively, given that I don't see support anywhere unless it's `m_cgram` inside `mz800_state` (which in turn doesn't allow mounting MZ-700 SW items anyway).

hash/mz700_cass.xml
```diff
@@ -52,7 +67,7 @@ Easidata (English Version)
 
 <!-- This is a straight recording of the tape. It might be preferrable to convert to some cooked format. Is MZF considered adequate for preservation? -->
```
angelosa, line 68: This note is outdated (and repeated below a couple extra times). Mind if you remove those? It's outdated opinion, and migration of these .wav to .flac is pretty self explainatory by just grep the hash folder.

hash/mz700_cass.xml
ArcadeShadow: Added RUN command at the end

hash/mz700_cass.xml
ArcadeShadow: - Updated `supported` flag regarding the PCG-700 supported or required
- Removed outdated comments

PR #12319, opened 2024-04-28
namco/quickncrash.cpp: Add a new 'Quick & Crash' set

clawgrip:
New not working clone
--------------------------
Quick & Crash (set 2) [Museo del Recreativo, ClawGrip]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -33260,7 +33260,8 @@ topracera                       // bootleg
 topracern                       // bootleg
 
 @source:namco/quickncrash.cpp
-qncrash
+qncrash                         // (c) 1999 Namco
+qncrasha                        // (c) 1999 Namco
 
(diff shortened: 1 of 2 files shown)


clawgrip:
@ajrhacker How can I tell the version on this new set? (to promote it to parent if it's newer).
Thanks!

ajrhacker:
The original set has "NAMCOV2.200" at $A975 in the main program ROM. The offset could be different here.

clawgrip:
> The original set has "NAMCOV2.200" at $A975 in the main program ROM.

There are no strings on this program ROM... I'll leave it just as "set 2" then. Thx.

Osso13:
Given the program ROM in MAME is labeled QC1 and this one QC3 it may be the common Namco scheme where 1 usually means Japan and 3 usually means US, but it's just speculation so I think leaving set 2 is fine.

PR #12321, opened 2024-04-28
Add a skeleton driver for Cubieboard4 (CC A-80) single board computer

clawgrip:
New not working machines
Monkey Jump [Museo del Recreativo, Recreativas.org]

New not working software list additions
cubieboard4.xml:
  Android 4.1.20161119 for Cubieboard4 (v4.4, internal eMMC) [ClawGrip]
  Android 4.3.20170717 for Cubieboard4 (v4.4, internal eMMC) [ClawGrip]
  Debian server for Cubieboard4 (v1.0, internal eMMC) [ClawGrip]
  Debian server for Cubieboard4 (v1.0, MicroSD card) [ClawGrip]
  Linaro desktop for Cubieboard4 (v1.1, HDMI, internal eMMC) [ClawGrip]
  Linaro desktop for Cubieboard4 (v1.0, HDMI, MicroSD card) [ClawGrip]
  Linaro server for Cubieboard4 (v2.0, HDMI, MicroSD card) [ClawGrip]
  Linaro server for Cubieboard4 (v2.0, VGA, MicroSD card) [ClawGrip]

### DIFF
--- a/hash/cubieboard4.xml
+++ b/hash/cubieboard4.xml
@@ -23,7 +23,7 @@ Software list for Cubieboard4 (CC A-20) external SD card slot.
 		</part>
 	</software>
 
-	<software name="cb4_linarodstcrd">
+	<software name="cb4_linardstcrd">
 		<description>Linaro desktop for Cubieboard4 (v1.0)</description>
 		<year>20??</year>
 		<publisher>cubieboard.org</publisher>
@@ -35,7 +35,7 @@ Software list for Cubieboard4 (CC A-20) external SD card slot.
 		</part>
 	</software>
 
-	<software name="cb4_linarosvrcrdh">
+	<software name="cb4_linarsvrcrdh">
 		<description>Linaro server for Cubieboard4 (v2.0, HDMI)</description>
 		<year>20??</year>
 		<publisher>cubieboard.org</publisher>
@@ -47,7 +47,7 @@ Software list for Cubieboard4 (CC A-20) external SD card slot.
 		</part>
 	</software>
 
-	<software name="cb4_linarosvrcrdv">
+	<software name="cb4_linarsvrcrdv">
 		<description>Linaro server for Cubieboard4 (v2.0, VGA)</description>
 		<year>20??</year>
 		<publisher>cubieboard.org</publisher>
--- a/hash/cubieboard4.xml
+++ b/hash/cubieboard4.xml
@@ -1,0 +1,62 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+
+Software list for Cubieboard4 (CC A-20) external SD card slot.
+
+-->
+
+<softwarelist name="cubieboard4" description="Cubieboard4 software (for external SD card)">
+
+	<!-- "Official" images from cubieboard.org -->
+
+	<software name="cb4_debiansvrcrd">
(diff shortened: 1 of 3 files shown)


src/mame/skeleton/cubieboard4.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+ROM_START( cb4_debiansvr )
+	DISK_REGION( "nand" )
+	DISK_IMAGE( "cb4-debian-server-hdmi-emmc-v1.0",  0, SHA1(31db5866016cfd6f1be331c70859cdb27022a7aa) )
+ROM_END
+
+ROM_START( cb4_linarodst )
+	DISK_REGION( "nand" )
+	DISK_IMAGE( "linaro-desktop-cb4-emmc-hdmi-v1.1", 0, SHA1(249095e6faaee11e330eaf59285c2e68d0b5a30f) )
+ROM_END
```
angelosa, line 111: Not sure I gather this: what are those romset wrt what's available in SW list?

clawgrip: There are two kind of images, ones to be restored on a SD and then inserted on the MicroSD slot (on the software list) and others to be transferred to the internal NAND Flash (on the driver).
The internal NAND has a eMMC interface, but is not removable media (flash chips directly soldered to the PCB).

Should we move eMMC images to the software list also, creating a new interface (something like "cubieboard4_emmc"), and leaving on the driver just the arcade games?

Thanks!!!!

angelosa: Inclined to think that they both belongs to the same SW items, as different interface loading.

hash/cubieboard4.xml
```diff
@@ -0,0 +1,62 @@
...
+
+<softwarelist name="cubieboard4" description="Cubieboard4 software (for external SD card)">
+
+	<!-- "Official" images from cubieboard.org -->
+
+	<software name="cb4_debiansvrcrd">
+		<description>Debian server for Cubieboard4 (v1.0)</description>
+		<year>20??</year>
+		<publisher>cubieboard.org</publisher>
+		<part name="image" interface="cubieboard4_sdcard">
```
angelosa, line 18: interface="sdcard"

src/mame/skeleton/cubieboard4.cpp
```diff
@@ -0,0 +1,127 @@
...
+	// Basic machine hardware
+	ARM9(config, m_maincpu, 200'000'000); // Actually an Allwinner A80 2 GHz
+
+	// Video hardware
+	//SCREEN(...)
+
+	// Audio hardware
+	//SPEAKER(...)
+
+	GENERIC_CARTSLOT(config, m_cart, generic_plain_slot, "cubieboard4_sdcard");
```
angelosa, line 67: As above, needs to be shorthanded to `	GENERIC_CARTSLOT(config, m_cart_ext, generic_plain_slot, "sdcard");` which also hints being the wrong device type here (question arises about how `SPI_SDCARD` device is supposed to load from a SW interface so I don't mind for that atm).
Another one needs to be added above here, `GENERIC_CARTSLOT(config, m_cart_int, generic_plain_slot, "emmc");`, so that you can move files below as `interface="emmc"`.

clawgrip: Done!!
Non-arcade software moving to the software list, having two interfaces, "sdcard" and "emmc".

Thanks!!!!!

PR #12339, opened 2024-05-04
new NOT WORKING machines (TCH Power Ball)

mamehaze:
new NOT WORKING machines
----
Power Ball (prototype) [OldComputers ES]
Kong Ball (early prototype) [OldComputers ES]
Radical Enduro (early prototype) [OldComputers ES]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -44003,6 +44003,7 @@ topdrive                        // (c) 1995
 
 @source:tch/wheelfir.cpp
 wheelfir                        // (c) 199? TCH
+pwball
 
 @source:teamconcepts/comquest.cpp
 comquest                        // Comquest Plus German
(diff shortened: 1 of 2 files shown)


mamehaze:
This runs on the same board as Wheels and Fire, see https://www.oldcomputers.es/tch-power-ball/  vs http://citylan.it/index.php/Wheels_%26_Fire

It doesn't like the zoom logic and a few other things about the Wheels and Fire driver, so there's a kludge for it to run at all right now.  It does boot with functional controls in this state however.

I've also demoted Wheels and Fire, because frankly it doesn't meet any reasonable working criteria due to the heavy glitching and vastly incomplete emulation, it was promoted prematurely.  I've also removed myself from driver credits, as I don't recognize this driver as my own; I did the skeleton, but nothing really beyond that.  The whole thing could do with a rewrite anyway.

PR #12345, opened 2024-05-06
taito/taito_f3.cpp: split graphics into TC0630FDP device

12Me21:
<wip, not ready for review>

* move FDP (and FDA) functionality into a separate device (tc0630fdp.cpp)
* rearrange ROM layouts: merge `tiles_hi`/`sprites_hi` into `tiles`/`sprites`
* simplify gfxdecode slightly
* add second palette device for 12-bit palette mode (both can be used at the same time, per-scanline)
* implement programmable timer interrupt


12Me21:
need to rewrite 2mindril to use fdp device instead of inheriting from taito f3 class

12Me21:
wow we've been ignoring 3 bits of color this entire time (in the few games that use this color mode).. we thought the format was `RRRRGGGGBBBBxxxx` but it's actually `RRRRGGGGBBBBRGBx` !  
before/after comparison on ringrage: (look at the floor texture)
pjgds

y-ack:
kaiser knuckle monitor calibration screen is using sprites for color bars and displays wrong graphics.
sprite id 0x1ffff (out of range)

existing behavior is `sprite id % length`
highest rom chip is too small, probably effectively mirrored by address mapping pld.
this is awkward to fix, ... not fdp's responsibility ... should happen at gfxdecode but then implies lying about rom size and loading 4 times?

angelosa:
> this is awkward to fix, ... not fdp's responsibility ... should happen at gfxdecode but then implies lying about rom size and loading 4 times?

Not sure why _lying_, sounds pretty normal ROM mirroring to me.

mamehaze:
yeah using ROM_RELOAD or having a larger region size and filling it with 00/ff is a common way to handle this kind of thing - empty sockets often pull high or low, or in some cases there will be an addressing PAL/GAL of some kind causing mirroring.

a comment with the ROM loading explaining why it's being done is usually helpful to prevent anybody from changing it back thinking it's a mistake.

PR #12350, opened 2024-05-07
Add a new Jump Bug bootleg on Galaxian hardware

clawgrip:
New working clone
-----------------------
Jump Bug (bootleg, set 2) [Arcadedealer aka Edgar Solé aka Dios, Recreativas.org]


src/mame/galaxian/galaxian.cpp
```diff
@@ -2537,6 +2537,27 @@ void galaxian_state::jumpbug_map(address_map &map)
 	map(0xb000, 0xbfff).r(FUNC(galaxian_state::jumpbug_protection_r));
 }
 
+// Same as 'jumpbug_map' but without the AY-8910
```
Osso13, line 2540: If it's just that, then you might be able to update the jumpbug_map to call this one and just have the AY methods in there.

something like:
void galaxian_state::jumpbug_map(address_map &map)
{
     jumpbugbrf_map(map);

     AY methods here
}

src/mame/galaxian/galaxian.cpp
```diff
@@ -7882,17 +7902,25 @@ void guttangt_state::guttangts3(machine_config &config)
 
 
 
-void galaxian_state::jumpbug(machine_config &config)
+void galaxian_state::jumpbugbrf(machine_config &config)
 {
-	galaxian_base(config);
+    galaxian_base(config);
```
Osso13, line 7907: What's changing here and below? Could you run srcclean on this?

Osso13 (commented):
Just a couple suggestions

src/mame/galaxian/galaxian.cpp
clawgrip: Done!! Thanks!

src/mame/galaxian/galaxian.cpp
clawgrip: Fixed!

PR #12355, opened 2024-05-09
Fixed tracks in hdrivairp and brake in harddrivc/racedrivc test menu …

Madcadden:
…harddriv.cpp

Made the tracks playable in hdrivairp by fixing the collision detection file format so you are not offroad all the time and fixed the brake calibration error in the test menu for Hard Drivin' Compact and Race Drivin' Compact.


happppp:
hdrivairp rom load was a bug, I applied your fix here: https://github.com/mamedev/mame/commit/9bfc455ad680c6ed07658a68099f034e9b6afd20 (FYI no need for a 'this was changed because ..'. comment in src)

The change to harddriv_m.cpp looks like a hack, and I didn't try your brake pedal change (all the other games are fine?).

Madcadden:
@happppp Yeah all the other games are fine it is the same as one of the brake pedal patches that i did for "strtdriv'" last year, it allows you to complete the brake pedal calibration in the test menu in the compact versions of the game.

There is also a steering fix that fixes the steering wheel drifting off center in all the compact cabinets that has to be applied in harddriv_m.cpp, i use it on my own personal build of MAME but i didn't write that code it was done by a guy called "Geecab " who i was talking to on the arcadecontrols forum last year. I think he hasn't made a commit for it yet on mamedev/master branch but i can share his work here for you to take a look at, he is very knowlegable about how the steering wheel works in mame and the real hardware.

https://github.com/mamedev/mame/compare/master...geecab:mamefork:HARDDRIV2

https://forums.bannister.org/ubbthreads.php?ubb=showflat&Number=123436&page=1

PR #12356, opened 2024-05-09
Add a skeleton for a electromechanical arcade by Shoken

clawgrip:
New not working machine
-----------------------------
Ganbare Momotarou Oni Taiji [Museo del Recreativo]


angelosa:
https://github.com/mamedev/mame/blob/d693b3f8d93d2dd45ff026d10241139b9ddbde10/src/mame/dynax/ddenlovr.cpp#L13361

https://github.com/mamedev/mame/blob/d693b3f8d93d2dd45ff026d10241139b9ddbde10/hash/pce.xml#L2257

By consistency with other entries should be Momotarou.
Omake: https://en.wikipedia.org/wiki/Momotar%C5%8D

PR #12358, opened 2024-05-09
casio/fp1100.cpp: Add FP1000

clawgrip:
New not working machine
-----------------------------
FP-1000 [OldComputers ES]

### DIFF
--- a/src/mame/casio/fp1100.cpp
+++ b/src/mame/casio/fp1100.cpp
@@ -25,7 +25,7 @@
     TODO:
     - unimplemented instruction PER triggered (can be ignored)
     - Display can be interlaced or non-interlaced. Interlaced not emulated.
-    - Cassette Load is quite complex, using 6 pins of the sub-cpu. Not done.
+    - Cassette Load is quite complex, using 6 pins of the sub-CPU. Not done.
     - subcpu is supposed to be in WAIT except in horizontal blanking period.
       WAIT is not emulated in our cpu.
     - Keyboard not working.
@@ -377,7 +377,7 @@ void fp1100_state::handle_int_to_main()
 	}
 }
 
-/* Input ports */
+// Input ports
 static INPUT_PORTS_START( fp1100 )
 	PORT_START("KEY.0")
 	PORT_BIT(0xff, IP_ACTIVE_LOW, IPT_UNUSED)
@@ -648,7 +648,7 @@ void fp1100_state::machine_reset()
 
 void fp1100_state::fp1100(machine_config &config)
 {
-	/* basic machine hardware */
+	// basic machine hardware
 	Z80(config, m_maincpu, MAIN_CLOCK/4);
 	m_maincpu->set_addrmap(AS_PROGRAM, &fp1100_state::main_map);
 	m_maincpu->set_addrmap(AS_IO, &fp1100_state::io_map);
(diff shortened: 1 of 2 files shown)


src/mame/casio/fp1100.cpp
```diff
@@ -713,10 +714,22 @@ ROM_START( fp1100 )
...
+ROM_END
+
 } // anonymous namespace
 
 
-/* Driver */
+// Drivers
 
-/*    YEAR  NAME    PARENT  COMPAT  MACHINE  INPUT   CLASS         INIT        COMPANY  FULLNAME   FLAGS */
+//    YEAR  NAME    PARENT  COMPAT  MACHINE  INPUT   CLASS         INIT        COMPANY  FULLNAME   FLAGS
```
angelosa, line 733: I dislike this cargo cult, it's outdated and it throws bad habits for people using this as a net safety for something they should look at where the macro is actually located.
Could you please get rid of it?

clawgrip: Done!

src/mame/casio/fp1100.cpp
```diff
@@ -713,10 +714,21 @@ ROM_START( fp1100 )
...
+
 } // anonymous namespace
 
 
-/* Driver */
+// Drivers
 
-/*    YEAR  NAME    PARENT  COMPAT  MACHINE  INPUT   CLASS         INIT        COMPANY  FULLNAME   FLAGS */
 COMP( 1983, fp1100, 0,      0,      fp1100,  fp1100, fp1100_state, empty_init, "Casio", "FP-1100", MACHINE_NOT_WORKING)
+COMP( 1982, fp1000, 0,      0,      fp1100,  fp1100, fp1100_state, empty_init, "Casio", "FP-1000", MACHINE_NOT_WORKING)
```
cuavas, line 734: These should either be parent/clone, or one marked as “compatible” with the other (the fourth parameter to the `COMP` macro).

clawgrip: Done, thank!

src/mame/casio/fp1100.cpp
```diff
@@ -714,7 +714,8 @@ ROM_START( fp1100 )
 	ROM_LOAD( "sub3.rom", 0x2000, 0xf80, BAD_DUMP CRC(fb2b577a) SHA1(a9ae6b03e06ea2f5db30dfd51ebf5aede01d9672))
 ROM_END
 
-// FP1000 has less memory than FP1100, using the same PCB, but with unpopulated sockets from RAM9 to RAM24 (only RAM1 to RAM8 are populated).
+/* FP1000 has less memory than FP1100, using the same PCB, but with unpopulated sockets from RAM9 to RAM24 (only RAM1 to RAM8 are populated).
+   The dumped PCB has the uPD7801G socket labeled as "uPD7801G-101", but the chip seems like a newer revision, silkscreened as "D7108G-118". */
```
cuavas, line 718: Do you have any evidence that systems ever shipped with a chip with mask code 101?  All the photos I’ve seen of FP1100 and FP1000 boards show a µPD7801G with mask code 118 while the PCB has “-101” on the silk screen.  The µPD7801G is sometimes socketed and sometimes soldered.

cuavas:
This needs its own machine configuration with reduced video RAM, etc.

There are more revisions out there.  One has main CPU ROM “3E A10 KK” with SHA1 fca8b0d19460e3be5f51db0bfcfe94a7657b4bfd and µPD7801 program ROM “3E A1S GTA” with SHA1 b69026a1efe704f972510190a6cd1d8af19148fc.

angelosa:
It also doesn't boot as things stands: lacks the actual IPL ROM as also noted by the FP-1000 French manual
image

PR #12360, opened 2024-05-11
galaxian/galaxian.cpp: hook up Galaxian discrete sound for 'bongoa', rename set to indicate the different hardware

devinacker:
This is a followup to https://github.com/mamedev/mame/pull/12314, using [this video](https://www.youtube.com/watch?v=ANLFKoMLPRI) as a reference. (I haven't attempted to figure out which color PROM the board in that video is using, though)

I was inclined to mark this set as a bootleg since the ROM differences are very clearly hacks on top of the parent set, but I'll wait for a second opinion on that.


mamehaze:
I kinda agree, it seems like an unofficial hack / conversion to run on different hardware, the code changes are patched in and not 'clean'

PR #12362, opened 2024-05-11
heathkit & zenith: Merge Heath and Zenith, since same company

mgarlanger:
Heathkit (Heath Company) and Zenith (ZDS - Zenith Data Systems) were the same company. Heath Company originally developed the systems, but the company was bought out by Zenith in 1979. The Zenith Data Systems division was created  and many of their computers were dual branded with both Heath and Zenith/ZDS branding.

Added a few of the duplicate models, including Z-19 terminal and Z-89 and Z-90 computers.


src/mame/heathzenith/h19.cpp
```diff
@@ -66,7 +67,11 @@ void h19_state::h19(machine_config &config)
...
+ROM_START( z19 )
+ROM_END
+
 } // anonymous namespace
 
-//    year  name     parent  compat  machine       input   class      init        company          fullname                         flags
-COMP( 1979, h19,     0,      0,      h19,          0,      h19_state, empty_init, "Heath Company", "Heathkit H-19",                 MACHINE_SUPPORTS_SAVE )
+//    year  name  parent  compat  machine input  class       init         company                fullname          flags
+COMP( 1979, h19,  0,      0,      h19,    0,     h19_state,  empty_init,  "Heath Company",       "H-19 Terminal",  MACHINE_SUPPORTS_SAVE )
+COMP( 1980, z19,  h19,    0,      h19,    0,     h19_state,  empty_init,  "Zenith Data Systems", "Z-19 Terminal",  MACHINE_SUPPORTS_SAVE )
```
cuavas, line 77: I don’t think duplicated machines are really useful if there are no differences at all.

src/mame/heathzenith/h89.cpp
```diff
@@ -998,10 +996,94 @@ ROM_START( h89_sigmasoft )
...
+//    year  name           parent compat machine        input class                init        company                fullname                   flags
+COMP( 1979, h88,           h89,   0,     h88,           h88,  h88_state,           empty_init, "Heath Company",       "H-88",                    MACHINE_SUPPORTS_SAVE)
+COMP( 1979, h89,           0,     0,     h89,           h89,  h89_state,           empty_init, "Heath Company",       "H-89",                    MACHINE_SUPPORTS_SAVE)
+COMP( 1980, z89,           h89,   0,     h89,           h89,  h89_state,           empty_init, "Zenith Data Systems", "Z-89",                    MACHINE_SUPPORTS_SAVE)
+COMP( 1981, z90,           h89,   0,     h89,           h89,  h89_state,           empty_init, "Zenith Data Systems", "Z-90",                    MACHINE_SUPPORTS_SAVE)
+COMP( 1984, h89_sigmasoft, h89,   0,     h89_sigmasoft, h89,  h89_sigmasoft_state, empty_init, "Heath Company",       "H-89 with SigmaSoft IGC", MACHINE_SUPPORTS_SAVE)
+COMP( 1984, z90_sigmasoft, h89,   0,     h89_sigmasoft, h89,  h89_sigmasoft_state, empty_init, "Zenith Data Systems", "Z-90 with SigmaSoft IGC", MACHINE_SUPPORTS_SAVE)
```
cuavas, line 1089: Same with this – it adds clutter to the supported system list and so on, and requires dual-maintenance of the copy/pasted ROM definitions.

src/mame/heathzenith/h19.cpp
mgarlanger: @rb6502 had mentioned there would be value in documenting the systems like this, but I will remove them for now so this PR can be merged (it would be trivial to add back in a later PR).


As for the copy/paste of the rom definitions, I was thinking doing something like this would be better: 
```
static const auto rom_z89 = rom_h89;
```
or have a macro in `romentry.h`
like
```
#define ROM_DUP(new_sys, existing_system)  static const auto rom##new_sys rom##existing_system;
```
So if anything changes in the rom code, it obvious that this macro needs updating.

src/mame/heathzenith/README.md
```diff
@@ -0,0 +1,53 @@
+# Heath Company / Zenith Data Systems
```
rb6502, line 1: I don't know that we've done anything like this before.  I'm not totally against it, but my initial feeling is that this is more like something that would either be in source comments or the wiki: https://wiki.mamedev.org/index.php/System-Specific_Setup_and_Information

@cuavas @galibert Do you guys have any feelings about this?

src/mame/heathzenith/h89.cpp
```diff
@@ -996,7 +996,36 @@ ROM_START( h89_sigmasoft )
...
+	ROM_SYSTEM_BIOS(2, "mms84b", "MMS 444-84B")
+	ROMX_LOAD("2732_444_84b_mms.u518",    0x0000, 0x1000, CRC(7e75d6f4) SHA1(baf34e036388d1a191197e31f8a93209f04fc58b), ROM_BIOS(2))
+
+	ROM_SYSTEM_BIOS(3, "kmr-100", "Kres KMR-100 V3.a.02")
+	ROMX_LOAD("2732_kmr100_v3_a_02.u518", 0x0000, 0x1000, CRC(fd491592) SHA1(3d5803f95c38b237b07cd230353cd9ddc9858c13), ROM_BIOS(3))
+
+	ROM_SYSTEM_BIOS(4, "mtrhex_4k", "Ultimeth 4k ROM")
+	ROMX_LOAD("2732_mtrhex_4k.u518",      0x0000, 0x1000, CRC(e26b29a9) SHA1(ba13d6c9deef682a9a8262bc910d46b577929a13), ROM_BIOS(4))
+
+	ROM_SYSTEM_BIOS(5, "mtr90-84", "Heath's MTR-90 (444-84 - Superseded by 444-142)")
```
rb6502: Was this actually called "Heath's MTR-90"?  These systems are hard to find info on so I don't actually know, but it seems unlikely even for the 80s :-)

mgarlanger: I just looked at the manual, it shows "Zenith Data Systems MONITOR MTR-90" 
Screenshot 2024-05-24 at 10 10 44 PM

rb6502: My objection is to the "Heath's" part, because I assume it wouldn't be named like grape jelly.  If the ROM is just the Heath version of what is otherwise badged as a Zenith Data Systems MTR-90, I'd call that BIOS "MTR-90, Heathkit (444-84 - Superseded by 444-142)".

mgarlanger: The MTR-90 model numbers were actually the same between ZDS and Heath. I had just added the `Heath's` part, since the other ROMs had the company's name. Adding the `'s` wasn't intentional. I could just use `Zenith Data Systems` for the two MTR-90s, and `Heath` for the MTR-89 to make it consistent.

mgarlanger: Since there are questions on whether or not this PR will get merged, I'll make the description changes in a followup PR.

rb6502: Ok.  Given the 's wasn't intentional, I think this is OK if you fix that.

src/mame/heathzenith/h19.cpp
mgarlanger: @rb6502 what do you think about having things like h19/z19 and h89/z89 declared even if the branding is the only difference between the systems? And a define like I mentioned above could address the copy/pasted ROM definitions Vas mentioned in the other comment.

src/mame/heathzenith/README.md
cuavas: It doesn’t belong here.  It’s a combination of trivia that belongs on Wikipedia, usage information that belongs on the MAMEdev wiki, and notes that belong in source files.

PR #10816, opened 2023-01-09
coleco/coleco.cpp: Added Super Game Module emulation.

0kmg:
- Added a base class and restructured a bit. Only the official ColecoVisions (NTSC/PAL) get the SGM, since it plugs into their specific expansion port.

This adds 32K RAM and an AY to the base unit. Mostly it seems this has been used to convert MSX titles—of which quite a load of cartridges have been released. I've set the machine config default to have the unit installed (On). I don't think it is known to interfere with any traditional ColecoVision games.

Ideally someone can properly emulate the expansion port and devicify the SGM. The Adam has the same port and the SGM can be used there too.

PR #12368, opened 2024-05-13
Add DRAC-1, an AIM 65 clone from Comelta

clawgrip:
New working clones
------------------
DRAC-1 [OldComputers ES]

### DIFF
--- a/src/mame/rockwell/aim65.cpp
+++ b/src/mame/rockwell/aim65.cpp
@@ -395,8 +395,8 @@ ROM_START( drac1 )
 	ROM_REGION(0x0800, "cr120d", 0) // CR-120D expansion board: EPROM Programmer Module
 	ROM_LOAD("cr120_crosaim_v-b_120_ve-c_7000.bin",     0x0000, 0x0800, CRC(09840652) SHA1(ad9ae4cbdcb86e749d8a385ec6334e97d37ef0d4))
 
-	ROM_REGION(0x0800, "cr125", 0) // CR-120D expansion board: Floppy Control Card
-	ROM_LOAD("cr125_v94_800.bin",                       0x0000, 0x0800, CRC(09840652) SHA1(ad9ae4cbdcb86e749d8a385ec6334e97d37ef0d4))
+	ROM_REGION(0x0800, "cr125", 0) // CR-125 expansion board: Floppy Control Card
+	ROM_LOAD("cr125_v94_800.bin",                       0x0000, 0x0800, CRC(f0fddc5a) SHA1(d7e927c5f177977a7032b8dddeb17a217fcb3fdd))
 ROM_END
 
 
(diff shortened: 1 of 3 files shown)


src/mame/rockwell/aim65.cpp
```diff
@@ -300,28 +299,111 @@ void aim65_state::aim65(machine_config &config)
...
+	 -1 x 4 dip switches bank
+	 -1 x EPROM
+	 -1 x floppy connector
+	https://www.oldcomputers.es/wp-content/uploads/CR-125.pdf
+	*/
+	MOS6522(config, "via1_cr125", AIM65_CLOCK); // R6522P, unknown clock
+	MOS6522(config, "via2_cr125", AIM65_CLOCK); // R6522P, unknown clock
+	FD1791(config, "fdc_cr125", AIM65_CLOCK); // FD1791A, unknown clock
+	//FLOPPY_CONNECTOR(...)
+	M6503(config, "cpu_cr125", AIM65_CLOCK);
```
cuavas, line 353: If these are expansions and not fixed parts of the system, they should be in card devices rather than the machine itself.  From what I understand, this system uses a standard AIM 65 logic board, and the expansion box connects to the J3 expansion connector, which is present on all AIM 65 systems, and just not emulated.

From looking at the AIM 65 manual, the expansion connector J3 seems to expose the 6502 bus directly, with the same pinout as the KIM-1 expansion connector B.

From looking at the schematic of the CR 113 board in the manual, it seems to use a different pinout to the J3 connector on the logic board, so I guess the card connectors in the DRAC-1 expansion box have the signals from the J3 expansion connector rearranged.

cuavas (commented):
The DRAC-1 expansion box and cards connect to J3 on the AIM 65 logic board, which has the same pinout as the KIM-1 expansion connector B.  These cards belong in src/devices/bus/kim1.  The manuals do include schematics – they aren’t the best quality, but it should be good enough to work out how stuff connects.

The AIM 65 needs to have KIM-1 bus slots added with whatever changes made to ensure there’s nothing specific to the KIM-1 itself mixed up in the slot code.

src/mame/rockwell/aim65.cpp
clawgrip: I'll just add then the DRAC-1 as another BIOS, and I'd appreciate if someone else will preserve somehow the other dumps as devices. Take into account that DRAC-1 uses a backplane so multiple expansion cards may be connected simultaneously, don't know if this feature is supported for the AIM 65 devices...
I'll have a look at the DRAC-1 BASIC, as it seems different than the regular Extended BASIC.

Thanks!

cuavas: Can you leave this open as a reminder?  Eventually someone will pick it up, like the Super Game Module from #10816 was done as a proper slot device in ad7fdda12b4cade7032a0e34c3f6ae6dd3964cfa.

clawgrip: Yes! Opened a separate PR, #12387

PR #12371, opened 2024-05-13
Librie

clawgrip:


src/mame/sony/librie.cpp
```diff
@@ -0,0 +1,123 @@
...
+	ROM_LOAD( "c_29lv400.ic1611", 0x00000, 0x80000, NO_DUMP )
+
+	ROM_REGION( 0x4000, "mcu", 0 ) // H8/3802 HD6473802FP
+	ROM_LOAD( "v22.ic801", 0x0000, 0x4000, NO_DUMP ) // 16 kbytes ROM
+ROM_END
+
+} // anonymous namespace
+
+
+SYST( 2004, librie, 0, 0, ebx5003, ebx5003, librie_state, empty_init, "Sony", "Librie EBR 1000", MACHINE_IS_SKELETON )
```
angelosa, line 123: No EOF marker, should always close with an empty line.

clawgrip: Fixed, thanks. I'll try to check it next time.

PR #12375, opened 2024-05-15
konami/konamigv, konami/konamigq, konami/twinkle: Adjust DMA timings

987123879113:
The 10usec delay per bytes is too slow and causing issues with games, so this should bring load speeds back closer to what it was with the old code. Should also fix https://mametesters.org/view.php?id=8860.

Previous related commits:
https://github.com/mamedev/mame/commit/a6a81a692955a8abfd41db832ec33cd64fe9323e
https://github.com/mamedev/mame/commit/cb8b90cd389ca928ffade1d8384893fdde87d4d1

### DIFF
--- a/src/mame/konami/konamigq.cpp
+++ b/src/mame/konami/konamigq.cpp
@@ -287,7 +287,7 @@ void konamigq_state::scsi_dma_read( uint32_t *p_n_psxram, uint32_t n_address, in
 	m_dma_offset = n_address;
 	m_dma_size = n_size * 4;
 	m_dma_is_write = false;
-	m_dma_timer->adjust(attotime::from_usec(10));
+	m_dma_timer->adjust(attotime::zero);
 }
 
 void konamigq_state::scsi_dma_write( uint32_t *p_n_psxram, uint32_t n_address, int32_t n_size )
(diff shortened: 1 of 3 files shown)


src/mame/konami/konamigq.cpp
```diff
@@ -313,13 +313,13 @@ TIMER_CALLBACK_MEMBER(konamigq_state::scsi_dma_transfer)
 	}
 
 	if (m_dma_requested && m_dma_size > 0)
-		m_dma_timer->adjust(attotime::from_usec(10));
+		m_dma_timer->adjust(attotime::zero);
```
cuavas, line 316: Since this is in a timer callback, using a zero delay here will result in the scheduler calling the callback again before executing another timeslice.  You could get more-or-less the same effect with a loop inside this function.

Do you want to leave all the timer adjust calls in and add notes about the timing being unknown, or do you want to change the callbacks to use loops (and maybe still add notes about it)?

FWIW I did try to find information about how DMA timings are generated and got nowhere.

mamehaze:
slightly unrelated, but btchamp could do with a default / initialized NVRAM or it will hang on the high score screen as it doesn't have valid data to display.

src/mame/konami/konamigq.cpp
987123879113: I removed the timer adjust from `scsi_dma_transfer` and added a note about finding the proper DMA timings. I think the original situation I was trying to avoid that I have accidentally ran into in the past was a kind of loop situation where scsi_drq would start the DMA transfer, which would then kick off another call to scsi_drq, etc. So the timer calls were used to ensure that the functions would get called after scsi_drq returned so it wouldn't end up in recursion hell. I tested with various konamigv, konamigq, and twinkle games and just using a while loop here works fine.

PR #12376, opened 2024-05-15
namco/sweetland4.cpp: Rename 'tairyouda' as 'tairyoda', as per game name

clawgrip:


cuavas:
It looks weird if you abbreviate it like that.  Why not “tairyodk”?  That fits neatly in eight characters and is a better mnemonic.

cuavas:
See a49035cc72ca1cc4977988509b887fdb33b510c9.

PR #12379, opened 2024-05-17
shared/decospr.cpp: Use device_gfx_interface for gfx decode functions, Cleanups:

cam900:
- Reduce unnecessary lines
- Constantize variables
- Use Bit macros for bitfields
- Use C++ style comment for single line comments
- Fix spacing
- Move sprite configuration related stuff into machine config
- Fix file directory, Sort to Alphabetical order

dataeast/minor cleanups for various drivers
- Constantize values
- Reduce literal tags and runtime tag lookups
- Fix ROM region namings
- Simplify deco146/104 related function namings
- Reduce duplicates
- Reduce unnecessary lines
- Fix sound routing if chip and/or PCB supports mono sound only
- Fix palette entries

dataeast/backfire.cpp, deco156.cpp, simpl156.cpp:
- Use memory_share_creator instead unique_ptr/array

dataeast/boogwing.cpp:
- Reduce preprocessor defines

dataeast/pktgaldx.cpp:
- Simplify gfxdecode layout
- Fix metadata for bootleg set (title screen says (c)1993 Data West)

f32/silvmil.cpp:
- Fix ROM region namings
- Reduce literal tag usage

misc/esd16.cpp:
- Constantize variables

misc/gotcha.cpp:
- Simplify gfx decode layout

yunsung/nmg5.cpp:
- Fix ROM region naming
- Simplify gfx decode layout

### DIFF
--- a/src/mame/dataeast/simpl156.h
+++ b/src/mame/dataeast/simpl156.h
@@ -15,17 +15,17 @@
 class simpl156_state : public driver_device
 {
 public:
-	simpl156_state(const machine_config &mconfig, device_type type, const char *tag)
-		: driver_device(mconfig, type, tag),
+	simpl156_state(const machine_config &mconfig, device_type type, const char *tag) :
+		driver_device(mconfig, type, tag),
 		m_maincpu(*this, "maincpu"),
 		m_deco_tilegen(*this, "tilegen"),
 		m_eeprom(*this, "eeprom"),
 		m_okimusic(*this, "okimusic"),
+		m_sprgen(*this, "spritegen"),
+		m_palette(*this, "palette"),
 		m_rowscroll(*this, "rowscroll_%u", 1U, 0x1000U, ENDIANNESS_LITTLE),
 		m_mainram(*this, "mainram", 0x4000U, ENDIANNESS_LITTLE),
 		m_systemram(*this, "systemram"),
-		m_sprgen(*this, "spritegen"),
-		m_palette(*this, "palette"),
 		m_spriteram(*this, "spriteram", 0x1000U, ENDIANNESS_LITTLE) { }
 
 	void joemacr(machine_config &config);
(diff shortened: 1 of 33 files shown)


src/mame/dataeast/simpl156.h
```diff
@@ -15,16 +15,19 @@
...
-		m_mainram(*this, "mainram"),
-		m_systemram(*this, "systemram"),
+		m_okimusic(*this, "okimusic"),
 		m_sprgen(*this, "spritegen"),
-		m_palette(*this, "palette") { }
+		m_palette(*this, "palette"),
+		m_rowscroll(*this, "rowscroll_%u", 1U, 0x800U, ENDIANNESS_LITTLE),
+		m_mainram(*this, "mainram", 0x4000U, ENDIANNESS_LITTLE),
+		m_systemram(*this, "systemram"),
+		m_spriteram(*this, "spriteram", 0x1000U, ENDIANNESS_LITTLE)
```
cuavas, line 29: I’m not sure what size `m_rowscroll` is supposed to be.  Before this PR, it was allocating 0x800/2 16-bit words, i.e. 0x400 16-bit words or 0x800 bytes.  I’ve made it not reduce that.

However, if you look at dataeast/simpl156.cpp, it maps the handlers across a range of 0x2000 byte addresses, or 0x800 32-bit word addresses.  There’s one 16-bit row scroll word at each 32-bit CPU word address, so to avoid buffer overruns, it needs to be *0x800 16-bit words* rather than bytes.

Currently, accessing the second half of either row scroll area causes a buffer overrun (e.g. 0x161000-0x161fff and 0x165000-0x165fff in Joe and Mac Returns, or 0x3e1000-0x3e1fff and 0x3e5000-0x3e5fff in Magical Drop).

The solution is probably one of:
* Double the row scroll RAM size
* Mirror the second half of each row scroll area
* Halve the length of each row scroll area

PR #12381, opened 2024-05-17
Move V.Smile Pro driver to Vtech folder

clawgrip:


cuavas:
Moving files between projects is a bit disruptive as it requires people to regenerate projects and leaves stale object files around, so it’s better to batch them up than to do them one at a time – see 4f6df8379a74e3c1607ecd91b08674d3a9c89ed3.

PR #12385, opened 2024-05-18
ninjakd2.cpp: Updates and cleanups:

cam900:
- Fix robokid, omegaf palette format
- Fix function namings
- Fix spacings
- Reduce preprocessor defines
- Suppress side effect for debugger read
- Use C++ style comments for single line comments
- Use correct typename values
- Fix ROM region namings
- Constantize variables
- Reduce literal tag usage

### DIFF
--- a/src/mame/upl/ninjakd2_v.cpp
+++ b/src/mame/upl/ninjakd2_v.cpp
@@ -61,13 +61,13 @@ TILE_GET_INFO_MEMBER(mnight_state::mnight_get_bg_tile_info)
 
 TILEMAP_MAPPER_MEMBER(robokid_state::robokid_bg_scan)
 {
-	/* logical (col,row) -> memory offset */
+	// logical (col,row) -> memory offset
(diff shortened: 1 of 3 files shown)


src/mame/upl/ninjakd2_v.cpp
```diff
@@ -339,13 +339,11 @@ static bool stencil_omegaf(   uint16_t pal ) { return( true ); }
...
-void ninjakd2_state::erase_sprites( bitmap_ind16 &bitmap)
+void ninjakd2_state::erase_sprites(bitmap_ind16 &bitmap)
 {
 	// if sprite overdraw is disabled, clear the sprite framebuffer
 	if (!m_next_sprite_overdraw_enabled)
-	{
 		m_sprites_bitmap.fill(0xf);
-	}
 	else
 	{
```
cuavas, line 348: Please don’t remove the braces from one path where all other paths have braces.  It’s a common recommendation to use braced compound statements for all paths if at least one path contains multiple lines.

PR #12392, opened 2024-05-20
galaxian/galaxian.cpp: Dumped PROM for "catacomb" [Arcadedealer aka Edgar Solé aka Dios, Recreativas.org]

clawgrip:


Osso13:
Thanks, should the  MACHINE_WRONG_COLORS flag be removed or arte the colors still wrong?

clawgrip:
Cannot tell, the PCB is broken... Maybe yes. Should I remove it?

Osso13:
Maybe leave it and add a note about needing hw references to verify colors?

clawgrip:
> Maybe leave it and add a note about needing hw references to verify colors?

Done, thanks!

PR #12396, opened 2024-05-21
tch\wheelfir.cpp - make pwball look better, first step of code clean-ups

mamehaze:
this reworks the wheelfir driver to be more easily maintainable / worked on in the future, while improving visuals in pwball

### DIFF
--- a/src/mame/tch/wheelfir.cpp
+++ b/src/mame/tch/wheelfir.cpp
@@ -529,7 +529,7 @@ void wheelfir_state::coin_cnt_w(uint16_t data)
 void wheelfir_state::wheelfir_main(address_map &map)
 {
 	map(0x000000, 0x0fffff).rom();
-	map(0x200000, 0x20ffff).ram();
+	map(0x200000, 0x20ffff).ram().mirror(0x010000); // kongball either needs the mirror (or has more ram?)
 
 	map(0x700000, 0x70001f).w(FUNC(wheelfir_state::wheelfir_blit_w));
 	map(0x720001, 0x720001).w("ramdac", FUNC(ramdac_device::index_w));
--- a/src/mame/tch/wheelfir.cpp
+++ b/src/mame/tch/wheelfir.cpp
@@ -251,6 +251,7 @@ private:
 		return 0;
 	}
 	void wheelfir_scanline_cnt_w(offs_t offset, uint16_t data, uint16_t mem_mask = ~0);
+	void do_direct_write(uint8_t sixdat);
 	void do_blit();
 	void wheelfir_blit_w(offs_t offset, uint16_t data, uint16_t mem_mask = ~0);
 	void wheelfir_7c0000_w(offs_t offset, uint16_t data, uint16_t mem_mask = ~0);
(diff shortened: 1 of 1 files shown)


mamehaze:
Honestly, this hardware just seems a case of them setting random bits in the blit registers and hoping they got the results they wanted.  everything is just a mess, from the zoom to the raster, and the words spoken from devs who worked on these games seem to tell a similar story.

I'm focusing here on making the driver a bit more maintainable in the future, and while Power Ball now works much better, Wheels & Fire does not.   Kong Ball also runs now, albeit with visual issues.

You can see even from the graphic pages, how every 'tile' in Power Ball has blank space around it that the hardware was weird when it came to deciding which pixels to draw.  The code that makes those render properly still leaves garbage around 'sprites' in Kong Ball, which seems to expect source pixels to be skipped.  I can skip some source pixels and make Kong Ball look better, but there are still sprites that don't work like that.  Radical Enduro looks worse than the PCB regardless of what I do, but that isn't a playable game even on real hardware.

Wheels and Fire is the only one to use zoom, raster effects, and the direct framebuffer writes, most of which feel like more hardware hackery than anything else.


```

we use u/v to refer to the x/y co-ordinates in the source page

for most things see comment at end of like
Z vals in table are currently being used for horizontal source zoom in wheelfir (other games unhappy with it)
z vals in table are currently being used for vertical source zoom ^^

00 uuuu uuuu xxxx xxxx  (u = horizontal src page co-ordinate, x = horizontal destination start)
01 ---- ---- XXXX XXXX  (X = horizontal destination end)
02 vvvv vvvv yyyy yyyy  (v = vertical src page co-ordinate, y = vertical destination start)
03 ---- ---- YYYY YYYY  (Y = vertical destination end)
04 ---- ---- ---- ----
05 ---- ---- ---- ----
06 pppp ppvu DDDD DDDD  (p = source page, u = horizontal src page co-ordinate high bit, v = vertical src page co-ordinate high bit, D = direct framebuffer write port)
07 ---- ---- yx-- --ff  (x = horizontal destination start high bit, y = vertical destination start high bit, ff = flipy;flipx)
08 FFzZ zZzZ ---- ----  (FF = vu source flip)
09 ---- ---- ---- YXzZ  (X = horizontal destination end high bit, Y = vertical destination end high bit)
0a -z-Z ZZZZ ---- ----
0b zz-Z ZZZZ ---- ----
0c zz-z zzzz ---- ----
0d ---- ---- ---- ----
0e ---- ---- ---- ----
0f SSSS SSSS SSSS SSSS  S = start blit trigger (write 0xffff, also writes of 0x0000 happen)



```

mamehaze:
ok wheelfir isn't really much worse now than before, if at all.

the vertical scroll register behavior seems to be weird, the only way I can see the rasters working is if the vertical scanline position is taken into account at the time the register is written and causes an offset to the register value, some kind of weird latching behavior?

nothing about how this hardware works is conventional.

rb6502:
I committed a slightly cleaned up version of this at https://github.com/mamedev/mame/commit/5cfb1cc07607259f4c6ef3a4c4012f24ecb85dcd.

mamehaze:
> I committed a slightly cleaned up version of this at [5cfb1cc](https://github.com/mamedev/mame/commit/5cfb1cc07607259f4c6ef3a4c4012f24ecb85dcd).

you seem to have dropped all the actual improvements?

mamehaze:
the only changes you've merged are mechanical ones, that fail to untangle anything, and aren't even worth accreditation.

PR #12397, opened 2024-05-23
gba.xml: Added five prototypes

FoxhackDN:
Crouching Tiger, Hidden Dragon (prototype, 20020220) [Anonymous, CodeBound]
Fuzz & Rocket (prototype, 20050118) [Anonymous, CodeBound]
Hamtaro - Rainbow Rescue (prototype, 20030729) [Zoda-Y13, CodeBound]
Tom Clancy's Rainbow Six - Rogue Spear (prototype, version 23, 20011025) [Anonymous, CodeBound]

New non-working software list items
I-Ninja (Prototype) [Hard4Games, CodeBound]

### DIFF
--- a/hash/gba.xml
+++ b/hash/gba.xml
@@ -6410,6 +6410,17 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
 		</part>
 	</software>
 
+	<software name="cthdp" cloneof="cthd">
+		<description>Crouching Tiger, Hidden Dragon (prototype, 20020220)</description>
+		<year>2001</year>
+		<publisher>Ubi Soft</publisher>
+		<part name="cart" interface="gba_cart">
+			<dataarea name="rom" size="3353604">
+				<rom name="cthd.bin" size="3353604" crc="a1d2d006" sha1="3d554e4e33563211865b9c5481444b0b8b4e6127"/>
+			</dataarea>
+		</part>
+	</software>
+
 	<software name="cthdup" cloneof="cthd">
 		<description>Crouching Tiger, Hidden Dragon (USA, prototype)</description>
 		<year>2001</year>
@@ -12321,6 +12332,17 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
 		</part>
 	</software>
 
+	<software name="fuzzrock" cloneof="cthd">
+		<description>Fuzz &amp; Rocket (prototype, 20050118)</description>
+		<year>2005</year>
+		<publisher>Halfbrick Studios</publisher>
+		<part name="cart" interface="gba_cart">
+			<dataarea name="rom" size="4124984">
+				<rom name="fuzz050118.bin" size="4124984" crc="fd239437" sha1="e6dab28aa3e9a5b1417bbf3929c45ff7260803b0"/>
+			</dataarea>
+		</part>
+	</software>
+
 	<software name="gachaddb">
 		<description>Gachasute! Dino Device - Blue (Japan)</description>
 		<year>2002</year>
@@ -14559,6 +14581,19 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
 		</part>
 	</software>
 
+	<software name="hamtarorp" cloneof="hamtaror">
+		<description>Hamtaro - Rainbow Rescue (prototype, 20030729)</description>
+		<year>2003</year>
+		<publisher>Nintendo</publisher>
+		<info name="serial" value="A84E"/>
+		<part name="cart" interface="gba_cart">
+			<feature name="slot" value="gba_sram" />
+			<dataarea name="rom" size="33554432">
+				<rom name="hamutaro4-a84e.gba" size="33554432" crc="e9442e7e" sha1="255432a07d0ea54bc9c5be204a207f203a44ad64"/>
+			</dataarea>
+		</part>
+	</software>
+
 	<software name="hanabi">
 		<description>Hanabi Hyakkei Advance (Japan)</description>
 		<year>2004</year>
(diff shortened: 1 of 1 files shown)


FoxhackDN:
The I-Ninja dump crashes MAME immediately, but works fine in VisualBoyAdvance, so I flagged it as supported="no". Hopefully someone can figure out why it crashes.

ArcadeShadow:
Fix those metadata on the "I-Ninja" set, please:
- Put "Prototype" word in lowercase
- ROM name cannot have the ! character

hash/gba.xml
```diff
@@ -35846,6 +35893,18 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
...
 
+	<software name="rainb6rsp" cloneof="rainb6rs">
+		<description>Tom Clancy's Rainbow Six - Rogue Spear (prototype, version 23, 20011025)</description>
+		<year>2002</year>
+		<publisher>Ubi Soft</publisher>
+		<part name="cart" interface="gba_cart">
+			<feature name="slot" value="gba_eeprom_4k" />
+			<dataarea name="rom" size="7963152">
+				<rom name="roguespear_v23.bin" size="7963152" crc="666e31b5" sha1="d8426022537149670871817ff9336261b31172b5"/>
+			</dataarea>
```
cuavas, line 35904: This ROM size looks rather suspicious.  What was this actually dumped from?  Is it “trimmed”?  The GBA expects cartridges to behave like “burst ROM”, either with actual burs ROM chips, or using a counter and conventional ROM.  It’s usually a power-of-two size.

hash/gba.xml
```diff
@@ -15794,6 +15829,18 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
...
 
+	<!-- Game crashes MAME immediately -->
+	<software name="ininjap" supported="no">
+		<description>I-Ninja (prototype)</description>
+		<year>2003</year>
+		<publisher>Namco</publisher>
+		<part name="cart" interface="gba_cart">
+			<dataarea name="rom" size="52428800">
+				<rom name="i-ninja (unlabeled prototype) (gameboy advance, unreleased).gba" size="52428800" crc="e0cf48d9" sha1="431f9cd0c2b5976bd3362844369e6377dfda77a8"/>
+			</dataarea>
```
cuavas, line 15840: This is an odd size as well – what chips did the cartridge actually have?

hash/gba.xml
```diff
@@ -12321,6 +12332,17 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
...
 	</software>
 
+	<software name="fuzzrock" cloneof="cthd">
+		<description>Fuzz &amp; Rocket (prototype, 20050118)</description>
+		<year>2005</year>
+		<publisher>Halfbrick Studios</publisher>
+		<part name="cart" interface="gba_cart">
+			<dataarea name="rom" size="4124984">
+				<rom name="fuzz050118.bin" size="4124984" crc="fd239437" sha1="e6dab28aa3e9a5b1417bbf3929c45ff7260803b0"/>
+			</dataarea>
```
cuavas, line 12342: This is also a suspicious size.

hash/gba.xml
```diff
@@ -6410,6 +6410,17 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
...
 	</software>
 
+	<software name="cthdp" cloneof="cthd">
+		<description>Crouching Tiger, Hidden Dragon (prototype, 20020220)</description>
+		<year>2001</year>
+		<publisher>Ubi Soft</publisher>
+		<part name="cart" interface="gba_cart">
+			<dataarea name="rom" size="3353604">
+				<rom name="cthd.bin" size="3353604" crc="a1d2d006" sha1="3d554e4e33563211865b9c5481444b0b8b4e6127"/>
+			</dataarea>
```
cuavas, line 6420: This one as well.

cuavas (commented):
Where did these come from?  The sizes look rather suspicious.  I haven’t looked at the code, but it’s possible the crash is caused by MAME assuming the ROM size is a power of 2.

FoxhackDN:
> Where did these come from? The sizes look rather suspicious. I haven’t looked at the code, but it’s possible the crash is caused by MAME assuming the ROM size is a power of 2.

Crouching Tiger, Fuzz & Rocket, Hamtaro and Rainbow Six were released through Hidden Palace; I-Ninja was released through the Internet Archive (and via Youtube.) The files are untouched, I just grabbed the files, tested them, and marked as working / non-working where needed.

The I-Ninja dump has a lot of empty data, I tried trimming it to a more sane size and it still crashed. The game is all sorts of weird, so I assumed the crash was due to it trying to do something that isn't emulated correctly, or even at all.

hash/gba.xml
FoxhackDN: There were no photos of the board posted, just the dev cart, so I couldn't add that info to the XML, sorry.

cuavas:
The dumps are all suspect due to the weird sizes, though.  It looks like someone already tried to trim them.  Almost all GBA ROMs are power of two sizes.  The weird sizes suggest they’re not complete dumps.

FoxhackDN:
Well, it's your call. If you think the dumps are too weird and may just cause issues then feel free to close this pull request. Neither of us can do anything about dumps that are released like this. :(

cuavas:
Yeah, it’s annoying.  We often don’t have channels to ask the dumpers what’s going on.

hash/gba.xml
einstein95: This was released alongside a copy of the final game (`ABPZJ.SRLうみにんfinal`) and was apparently a leak by a former employee with a last-modified date of Jan 18, 2005. As for the file size, this is common for ROMs built from source but haven't been flashed onto a dev cart.

hash/gba.xml
einstein95: Overdump, actual size seems to be 33554432 (0x2000000)

FoxhackDN: Can confirm that trimming this makes it work.

mamehaze:
99% of the time dev builds won't be a power of 2, they're programmed into the chips that way, with the programming tools usually only writing the blocks that are needed.

MAME typically should just round up to the biggest size that would fit in such cases.

PR #12399, opened 2024-05-23
skeleton/pullttrigger.cpp: "Pull The Trigger" is a GAME, not a COMPuter

clawgrip:


cuavas:
The macros are functionally identical – there is no longer a “system type” field.  There’s no need to obsess over them.

PR #12400, opened 2024-05-23
Add a skeleton driver for ICE's Cut The Rope (PC-based)

clawgrip:
Cut The Rope [Museo del Recreativo]

### DIFF
--- a/src/mame/pc/cuttherope.cpp
+++ b/src/mame/pc/cuttherope.cpp
@@ -1,0 +1,78 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/***********************************************************************************************************
+
+Skeleton driver for "Cut The Rope", by Innovative Concepts in Entertainment, Inc. (ICE).
+
+Standard PC with Intel DH61CR motherboard (IT8892E PCIe to PCI Bridge, Nuvoton W83677HG-i Super I/O, etc.):
+-Unknown RAM, unknown exact CPU type.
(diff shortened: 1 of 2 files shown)


src/mame/pc/cuttherope.cpp
```diff
@@ -0,0 +1,78 @@
...
+	ROM_LOAD("mcf51jm32vlk", 0x00000, 0x40000, NO_DUMP) // On the I/O PCB
+
+	DISK_REGION("ide:0:hdd")
+	DISK_IMAGE("cuttherope", 0, SHA1(25be8285ca75d759962f4b59b313a97d9077747e))
+ROM_END
+
+} // anonymous namespace
+
+
+GAME(2003, cuttherope, 0, cuttherope, cuttherope, cuttherope_state, empty_init, ROT90, "ICE", "Cut The Rope", MACHINE_IS_SKELETON )
```
cuavas, line 78: Is the year here correct?  From a quick look around, there seem to be two completely different games called Cut The Rope from ICE: a 2013 crane game (uses a suction cup rather than a claw), and 2016 touchscreen game using a vertical flat panel display and paying out with tickets.  Is this one of those, or is it a different game altogether?

The original game the redemption games are based on was released in 2010, so a 2003 year should logically be impossible.

clawgrip: You're right. I've opened a PR (#12516) to fix it. Thx!!!!

PR #12401, opened 2024-05-23
Create skeleton device for Sega Megalo 50 DASS and hook it on supported Sega C2 games [jordigahan, Recreativas.org]

clawgrip:


philipjbennett:
This should be implemented as a device, like segabill.cpp for example.

clawgrip:
> This should be implemented as a device, like segabill.cpp for example.

Done, thanks!

clawgrip:
@cuavas , can you please review this PR? Thanks!

PR #12407, opened 2024-05-25
vsystem/fromanc2.cpp: Cleanups

cam900:
- Split driver state class for reduce optional finders
- Correct type for variables
- Use C++ style comments for single line comments
- Reduce literal tag usage and runtime tag lookups
- Constantize variables
- Fix spacing

### DIFF
--- a/src/mame/vsystem/fromanc2.cpp
+++ b/src/mame/vsystem/fromanc2.cpp
@@ -52,20 +52,12 @@ void fromanc2_base_state::portselect_w(uint16_t data)
 
 uint16_t fromanc2_base_state::keymatrix_r()
 {
-	uint16_t ret;
+	uint16_t ret = 0xffff;
 
-	switch (m_portselect)
-	{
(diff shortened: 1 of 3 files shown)


src/mame/vsystem/fromanc2.cpp
```diff
@@ -45,23 +45,25 @@ void fromanc2_state::sndcmd_w(uint16_t data)
...
+		case 0x01:  ret = m_in_key[0]->read(); break;
+		case 0x02:  ret = m_in_key[1]->read(); break;
+		case 0x04:  ret = m_in_key[2]->read(); break;
+		case 0x08:  ret = m_in_key[3]->read(); break;
+		default:
+			ret = 0xffff;
+			if (!machine().side_effects_disabled())
+				logerror("PC:%08X unknown %02X\n", m_maincpu->pc(), m_portselect);
 			break;
 	}
```
cuavas: This is a key matrix that the author failed to recognise.  It’s better to emulate them properly.

PR #12408, opened 2024-05-25
HP9816A added

fulivi:
Hi,
with this PR I'm adding the support for HP9816A system. ROM images will follow at the usual mail address.
I got the idea to write a driver for 9816 because I stumbled on my name listed as copyright holder of `hp9k.cpp`. Here this same system was implemented but only at skeleton level. I think someone mistakenly put my name there because I never contributed a single line to that file. Do you think we should remove the old implementation?
Thanks.
--F.Ulivi


Robbbert:
@fulivi The copyright was added here: https://github.com/mamedev/mame/commit/af5c8113b302c81b19e6e2448807dbaa761188bf

PR #12413, opened 2024-05-26
heathkit/h89.cpp: Add Magnolia Microsystems DD controller and h89-mms system

mgarlanger:
- Implement MMS 77316 DD controller. 
- Add new clone of H89 'h89-mms' for Magnolia MicroSystems hardware.


rb6502:
I'm going to merge this because I'm the poster child for giant-pile-of-code drivers that are made right once I understand how everything fits together, but we definitely need to add slot bus support for this system sooner rather than later.  I appreciate that the FDC here is a separate file and device, which will help significantly with that process.

PR #12418, opened 2024-05-29
nitchibutsu/shettle.cpp - Change shettle palette PROM decode to match wiping

mamehaze:


mamehaze:
this doesn't match the reference shots provided, but it's very easy to incorrectly hook up r/g/b on a PCB anyway.

mamehaze:
With the changes it looks like this

The fire effect on the shuttle is red/yellow (logical colours)
The handle of the axe is brown, the blade is grey (logical colours)
The torchlight is yellow (again a more logical colour)
Skin of character is a shade of pink
You can much more easily read the numbers on the flags when you collect the items

mamehaze:
The existing PROM decode gives the following, where none of those colour choices seem logical

mamehaze:
while neither match the PCB output shown at https://www.mameworld.info/ubbthreads/showflat.php?Cat=&Number=398233 I strongly suspect it's just hooked up incorrectly there.

mamehaze:
alternatively just using the wiping decode looks a lot better too, giving a more realistic skin tone.  given how close the rest of the driver is to wiping I might just be inclined to use that.

PR #12421, opened 2024-05-30
machine/watchdog.cpp: relax log message

angelosa:


src/devices/machine/watchdog.cpp
```diff
@@ -144,14 +144,14 @@ void watchdog_timer_device::watchdog_enable(int state)
...
-	logerror("Reset caused by the watchdog!!!\n");
+	logerror("watchdog_reset issued\n");
 
 	bool verbose = machine().options().verbose();
 #ifdef MAME_DEBUG
 	verbose = true;
 #endif
 	if (verbose)
-		popmessage("Reset caused by the watchdog!!!\n");
+		popmessage("watchdog_reset issued");
```
angelosa, line 154: PR because I'm legit not sure why or how even we need this `popmessage` printed out if verbose is on. 
`popmessage` is for exceptional cases, and by quick glance a watchdog reset isn't (most NeoGeo games, airbustr ...)

cuavas:
I think getting rid of the `popmessage` is fine, but why change the wording of the `logerror` messages?  The multiple exclamation marks are silly, but plain English messages are better than weird snake case that doesn’t correspond to an actual symbol.

angelosa:
> but plain English messages are better than weird snake case that doesn’t correspond to an actual symbol.

Didn't put a great effort about rewording, what about `<funcition_name>: reset issued.`? The intention was effectively to remove exclamation marks to begin with.

PR #12426, opened 2024-06-01
pc9801: new software list for original disks

rzero9:
This is the first step in adding the PC-98 floppy dumps that krugman has been doing for a few months. At this point there are already more than 600 of them, so this is more or less a "proof of concept" PR with just a few ones, as I don't want to commit too much of my time to this until I'm sure I'm doing it properly.

The new softlist entries include the proper interface attributes (floppy_5_25 or floppy_3_5) matching the real disks they were dumped from. This will affect compatibility with the different emulated PC-98 models, unless the user manually swaps the drive type, but it's more accurate to how they would work on real hardware (and it's how the IBM-compatible drivers do it, for example).

The sector structure of the disks can actually vary quite a bit, even without taking protections into account. The most common ones are 77/8/2/1024 DOS-formatted HD disks, but there are also some DD disks (usually from very old pre-VM software), BASIC-formatted disks with 256-byte sectors, and even some custom ones. The HxC MFM image format can cover all of that and also a good number of protected disks, so I'm using it.

Some things that might need clarification:

- pc98.xml has been renamed to pc98_flop_misc.xml to avoid confusion.
- Software that shipped with protection has been added to pc98_flop_orig.xml, but the existing cracked versions are still in pc98_flop_misc.xml, marked as such.
- Software that shipped without protection has been handled in a case-by-case basis, by comparing the existing images with the newly-dumped ones and choosing the ones that have the least amount of "defects" (added/modified/deleted files, alterations caused by Windows 9x like the infamous "IHC" OEM ID, and such)
- In some cases, the new dumps are a different software revision compared to the existing images. In that case, both of them have been kept.
- Existing comments about compatibility issues have been turned into a "notes" tag so they go through validation and the MAME UI can potentially use them at some point.
- Brandish Renewal is a re-conversion of my own dump. I must have done something wrong while converting it to MFM last time and it didn't actually pass the protection check. This time I have made sure that it works properly.

### DIFF
--- a/hash/pc98_flop_misc.xml
+++ b/hash/pc98_flop_misc.xml
@@ -3,7 +3,7 @@
 <!--
 license:CC0-1.0
 -->
-<softwarelist name="pc98" description="NEC PC-9801 disk images">
+<softwarelist name="pc98_flop_misc" description="NEC PC-9801 miscellaneous disk images">
 <!--
 
   This software list collects disks from many different sources, including the 13GB pack with ~90 rar files floating around since 2011
@@ -2430,8 +2430,8 @@ TODO:
 		</part>
 	</software>
 
-	<software name="2069ad" supported="yes">
-		<description>2069AD</description>
+	<software name="2069adc" supported="yes">
+		<description>2069AD (cracked)</description>
 		<year>1991</year>
 		<publisher>ホームデータ (Home Data)</publisher>
 		<info name="release" value="19910803" />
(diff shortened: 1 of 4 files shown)


ICEknigh7:
For these cases, would it be possible to have MAME keep the parent/clone relationships between different softlists?

Something like `<software name="anniversc" cloneof="pc98_flop_orig/annivers">`, perhaps?

hash/pc98_flop_misc.xml
```diff
@@ -3,7 +3,7 @@
 <!--
 license:CC0-1.0
 -->
-<softwarelist name="pc98" description="NEC PC-9801 disk images">
+<softwarelist name="pc98_flop_misc" description="NEC PC-9801 miscellaneous disk images">
```
angelosa, line 6: Not a fan of using `misc`, in data context it has absolutely no value. `Unverified` / `Legacy` or even just `Cracked` would be better options, while filename can just be `pc98_flop.xml` (given that we won't get rid of this soon anyway).

rzero9: Done. I went with "unverified" since that's basically what it contains, images that haven't been compared with original disks, and not all of them are necessarily cracked.

rzero9:
> For these cases, would it be possible to have MAME keep the parent/clone relationships between different softlists?
> 
> Something like `<software name="anniversc" cloneof="pc98_flop_orig/annivers">`, perhaps?

I don't think that's possible right now. As far as I know MAME only tracks parent/clone relationships within the same list.

cuavas:
I think it’s better to limit the “_orig” lists to dumps believed to come from unmodified disks.  For example I wouldn’t add the dump with an apparently corrupted boot sector to the “_orig” list.

rzero9:
Sorry for not responding in a long time. Real life issues got in the way, and after that, a hard drive crash pretty much wiped out all the progress I had made with converting files and such. It's all recoverable with enough time, but for now I'm going to close this PR, and I'll get it started again when I'm ready, since I will have to take into account Kale's changes to pc98.xml anyway.

hash/pc98_flop_orig.xml
```diff
@@ -0,0 +1,1151 @@
...
+				<rom name="agalta2.mfm" size="3415833" crc="0056b967" sha1="b225ff12328da976c86c52ddfc7e4683fae37c9d" status="baddump"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="aggres" supported="no">
+		<description>Aggres</description>
+		<year>1986</year>
+		<publisher>リバーヒルソフト (Riverhill Soft)</publisher>
+		<notes>Hangs while booting the BASIC interpreter, could be related to the copy protection or the disks being 640 KB double density.</notes>
```
angelosa, line 144: Disks are really in 2D format, they requires the PC80S31 interface I've recently hooked up in the driver (and only for vanilla class).
Also I'm commenting here for a catch: Krugman's .mfi for this game fails with Disk B, does your .mfm conversion works fine instead?

rzero9: From what I remember, the .mfm file at least worked in the sense that MAME accepted it as valid. I think Krugman converts the files with the HxC software, and the MFI support in it is fairly recent, so maybe there are still some bugs in the MFI export code.

ICEknigh7: @rzero9 Try this when converting:
image

PR #12428, opened 2024-06-01
tomy/tomy_princ.cpp: Promoted princ and princnt to working

qufb:
To simplify the changeset, I've split instruction set changes in another PR, which should be merged first: https://github.com/mamedev/mame/pull/12427

* cpu/f2mc16/mb9061x.cpp: Added most ports/registers accessed by Prin-C;
* hash/princ.xml: Marked software items as supported, and adjusted some dates to match ICs;
* tomy/tomy_princ.cpp: Implemented graphics, keyboard and pen inputs; 
* tomy/tomy_princ.cpp: Adjusted machine names to match official names given in manuals;
* tomy/tomy_princ.cpp: Added new ROM revision 3 for Prin-C non touch-pad model;

### DIFF
--- a/hash/princ.xml
+++ b/hash/princ.xml
@@ -4,22 +4,30 @@
 license:CC0-1.0
 
 Known undumped titles:
-New Kazoku de Nengajou
+- New Kazoku de Nengajou (https://www.youtube.com/watch?v=1eZ72yNZrqc)
+
+Most (all?) carts share PCB markings: Prin-C E100-T002-11.
+
+PCBs have surface mount pads for two chips, but only one of them is populated:
+- Mask ROM carts use U1;
+- Memory Cassette uses U2 with flash ROM (Fujitsu 29F002T-90);
 -->
 
 <softwarelist name="princ" description="Tomy Prin-C cartridges">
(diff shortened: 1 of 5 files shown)


mamehaze:
I couldn't figure out how to use the cartridge, but maybe I was doing it wrong

anyway, in Games, if you select the first option, then during that 'game' (which I don't know if works correctly) you can manage to trigger
Fatal error: Unknown F2MC EA72 opcode 20 (PC=fefb38)

qufb:
@mamehaze 

> I couldn't figure out how to use the cartridge, but maybe I was doing it wrong

If you load any of those `ROM DATA` carts in princnt/princnt3, an icon shows up next to the main cat sprite. They just provide a different set of stamps you can apply on the postcards. The only exception is Doraemon's `ROM PRG` cart, which customizes more options. On princ, the stamps are available under the dog icon.

> Fatal error: Unknown F2MC EA72 opcode 20 (PC=fefb38)

Thanks for catching that, it seems to be related with inactivity. I was able to finish more than one level without it happening, but this time just waiting for a while I also got the error.

Let me know if you hit any more missing instructions, I'll get a fix pushed this week.

PR #12435, opened 2024-06-02
sony/news_r3k.cpp: Add support for NEWS-OS 4.1R

briceonk:
cxd1185.cpp: Add support for SPHI feature
dmac_0448.cpp: Add support for autopad feature

NEWS-OS 5 is unaffected by this change. It worked before, and still works. Adding in these features also allows NEWS-OS 4.1R to install from magneto-optical media and boot.

NEWS-OS 4 login screen:
image

Some applications running:
image

NEWS-OS 5 still running with these changes:
image

Networking also works, however, I will open up a PR later with the changes needed for that, since it is independent of these.


briceonk:
cc @pmackinlay: These are the non-networking changes needed for NEWS-OS 4, let me know if there is anything else you'd like for me to test.

briceonk:
SPHI snippet from datasheet:
image

PR #12439, opened 2024-06-03
imagedev/floppy: fix a bug with reading the first 1-bit on a track

SpecLad:
`floppy_image_device::find_index` uses binary search to find the index for which `buf[spos] <= position < buf[spos + 1]`. However, the algorithm behaves incorrectly when `position < buf[0]`. In this case, the algorithm returns 0, as if `position` was between `buf[0]` and `buf[1]`.

The effect of this is that if `get_next_transition` is called with a timestamp that is between the start of the revolution and the first transition, then instead of returning the timestamp of that transition, it returns the timestamp of the second transition instead. Essentially, the first 1-bit on the track gets flipped to a 0.

I have encountered this in Apple II emulation, where this bug manifests as sporadic I/O errors.

Fix it by doing two things:

1. Replace `find_index` with a call to `upper_bound` from the standard library, which behaves correctly in edge cases.

2. If `upper_bound` signals that `position < buf[0]`, then adjust `base` and `index` to point to the last transition of the previous revolution.

### DIFF
--- a/src/devices/imagedev/floppy.cpp
+++ b/src/devices/imagedev/floppy.cpp
@@ -33,6 +33,8 @@
 #include "util/ioprocsfilter.h"
 #include "util/zippath.h"
 
+#include <algorithm>
+
 /*
     Debugging flags. Set to 0 or 1.
 */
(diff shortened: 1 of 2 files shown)


rb6502:
Tried this on a small assortment of systems including Apple II and Amiga 500 and it doesn't seem to have any ill effects.   I'd still like @galibert to have a look as a sanity check though.

mgarlanger:
Interesting, I wonder if this could be the cause of the glitch, when we tried to have hard-sectored tracks keep the index hole as the start of the track. This caused intermittent glitches on the last sector of the track which has the index hole in the middle of the sector.

Lord-Nightmare:
I would not be at all surprised if this was the cause of that issue with track splices with hard sectored disks.

mgarlanger:
Confirmed this fixes the hard-sectored glitches at the index hole. I added this change to my hard-sectored PR - https://github.com/mamedev/mame/pull/11952 and added back the half sector rotate, and things are working fine. The Vector 4 booted without problem, able to run programs off the disk, copied the program, erase the file, and format a new disk. All without any sign of disk  errors.

galibert:
Much thanks for finding that :-)

PR #12442, opened 2024-06-04
CI: use more CPUs for building

SpecLad:
The standard GitHub runners have 3 to 4 CPUs, depending on the OS: <https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories>. Increase the number of Make jobs to utilize those CPUs better.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -58,7 +58,7 @@ jobs:
         ARCHOPTS: ${{ matrix.archopts }}
         SUBTARGET: ${{ matrix.subtarget }}
         TOOLS: 1
-      run: make -j2
+      run: make -j4
     - name: Validate
       run: ./${{ matrix.executable }} -validate
     - name: Reconcile driver list
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -33,7 +33,7 @@ jobs:
       env:
         USE_LIBSDL: 1
         TOOLS: 1
-      run: make -j2
+      run: make -j3
     - name: Validate
       run: ./mame -validate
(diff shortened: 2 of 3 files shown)


rb6502:
The change itself is reasonable, but my understanding is that there are GitHub-imposed resource limits that we potentially could run into and that's why the settings are low right now.

SpecLad:
I'm not aware of any additional limits. You get the runner that you get, and it's up to you how to use it.

> and that's why the settings are low right now.

I assumed that they are low because they were configured before GitHub started using 4-core runners (which is this year: <https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/>).

rb6502:
https://docs.mamedev.org/contributing/index.html says we've had contributors banned from GitHub for using "too much" CPU and GitHub would not define what "too much" is.  I don't remember the specific details (@cuavas might) but apparently it's a thing.

plaes:
Proper way to do that would be `make -j $(nproc)`.

cuavas:
I’d be somewhat concerned about that with limited memory.  A `-j5` build on a fairly minimal Linux system with 6GB RAM hits swap pretty hard at two points whether you use GCC or clang.  The macOS runners only have 7GB RAM, so I wouldn’t want to sail too close to the wind.

cuavas:
> https://docs.mamedev.org/contributing/index.html says we've had contributors banned from GitHub for using "too much" CPU and GitHub would not define what "too much" is. I don't remember the specific details (@cuavas might) but apparently it's a thing.

Correct, @0kmg was banned from GitHub actions.

PR #12445, opened 2024-06-05
cpu/z80/z80.cpp Update core generator to support r800 [holub, Wilbert Pol]

holub:
Update for z80 generation to support case required in #12433
Now core avoid partial generation and always build full commands list which looks cleaner based on the z80n and r800 requirements

### DIFF
--- a/src/devices/cpu/z80/z80.cpp
+++ b/src/devices/cpu/z80/z80.cpp
@@ -206,14 +206,14 @@
 #define TDAT8    m_shared_data.b.l // Typically represents values from D0..8 pins. 8bit input or output in steps.
 
 static bool tables_initialised = false;
-static u8 SZ[256];       // zero and sign flags
-static u8 SZ_BIT[256];   // zero, sign and parity/overflow (=zero) flags for BIT opcode
-static u8 SZP[256];      // zero, sign and parity flags
-static u8 SZHV_inc[256]; // zero, sign, half carry and overflow flags INC r8
-static u8 SZHV_dec[256]; // zero, sign, half carry and overflow flags DEC r8
+std::unique_ptr<u8[]> z80_device::SZ = std::make_unique<u8[]>(0x100);       // zero and sign flags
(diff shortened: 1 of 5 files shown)


src/devices/cpu/z80/z80.cpp
```diff
@@ -206,14 +206,14 @@
...
+std::unique_ptr<u8[]> z80_device::SZ = std::make_unique<u8[]>(0x100);       // zero and sign flags
+std::unique_ptr<u8[]> z80_device::SZ_BIT = std::make_unique<u8[]>(0x100);   // zero, sign and parity/overflow (=zero) flags for BIT opcode
+std::unique_ptr<u8[]> z80_device::SZP = std::make_unique<u8[]>(0x100);      // zero, sign and parity flags
+std::unique_ptr<u8[]> z80_device::SZHV_inc = std::make_unique<u8[]>(0x100); // zero, sign, half carry and overflow flags INC r8
+std::unique_ptr<u8[]> z80_device::SZHV_dec = std::make_unique<u8[]>(0x100); // zero, sign, half carry and overflow flags DEC r8
 
-static u8 SZHVC_add[2*256*256];
-static u8 SZHVC_sub[2*256*256];
+std::unique_ptr<u8[]> z80_device::SZHVC_add = std::make_unique<u8[]>(2 * 0x100 * 0x100);
+std::unique_ptr<u8[]> z80_device::SZHVC_sub = std::make_unique<u8[]>(2 * 0x100 * 0x100);
```
cuavas, line 216: Using dynamic allocations at static initialisation time is something we really want to avoid.  It hurts startup time and memory usage.  It would be better to wrap this stuff up in a nested class so you can use a function static and call the function to get it the first time a Z80 CPU is constructed.

holub: > Using dynamic allocations at static initialisation time is something we really want to avoid. It hurts startup time and memory usage. It would be better to wrap this stuff up in a nested class so you can use a function static and call the function to get it the first time a Z80 CPU is constructed.

#12433 is taking care of it

PR #12449, opened 2024-06-07
taito/taito_f2.cpp, seta/ssv.cpp: change some games to 2-way joystick

greg-kennedy:
Reviewing pinouts on arcade-history, and test mode in these games, I changed 8-way to 2-way for the following Taito F2 games:
* driftout (Drift Out)
* ssi (Super Space Invaders 91)
* liquidk (Liquid Kids)

Similarly, service mode for drifto94 (Drift Out '94) in the seta/ssv.cpp driver supports only one player, 2-way joystick and 2 buttons - not the generic 8-way + 3 button + 2 player provided by the driver


greg-kennedy:
Don Doko Don also appears to be a 2-way 2-button game: U/D do work to change initials on the high score screen, but they're aliased to L/R, and don't work anywhere else (aside from service mode test).  Without a pinout from the manual to check I didn't want to change that one though.

cuavas:
Several of these games do have full eight-way joystick input hardware and were typically used with standard control panels even if gameplay only uses two directions.  In cases like this, we expose all the inputs in MAME.

Maybe the change to SWAT is correct.  Is there an operator manual for this game available?

happppp:
Swat flyer says 4-way
https://flyers.arcade-museum.com/videogames/show/1102

They really do mean 4-way. See some other Sega conversion kits distributed by Bally
Up N Down says 8-way: https://flyers.arcade-museum.com/videogames/show/1226
Mr Viking says 8-way: https://flyers.arcade-museum.com/videogames/show/646
Flicky says 2-way: https://flyers.arcade-museum.com/videogames/show/373

greg-kennedy:
> Several of these games do have full eight-way joystick input hardware and were typically used with standard control panels even if gameplay only uses two directions. In cases like this, we expose all the inputs in MAME.
> 
> Maybe the change to SWAT is correct. Is there an operator manual for this game available?

I remember looking into this and finding that there was a change some time ago for games that have buttons but don't use them - the addition of a `reqbuttons` attribute in the `<control>` XML entry.  For example, this snippet from a game that has 2 buttons, but the second is used only for an Animation Test when enabled in DIP switches:

```xml
     <input players="2" coins="2" tilt="yes">
          <control type="joy" player="1" buttons="2" reqbuttons="1" ways="4"/>
          <control type="joy" player="2" buttons="2" reqbuttons="1" ways="4"/>
     </input>
```

Is there any desire for a similar `reqtype` attribute, for games like the few Taito F2 ones above?

cuavas:
To be honest, I think the optional flag should just be removed from I/O port fields.  It’s rarely used, it would be too much trouble to go through every game and try to work out which inputs are “required” considering DIP switches and soft settings, and it would cause a lot of bloat with excessive `PORT_MODIFY` for platforms with a common control setup where some games don’t use all the buttons.

For example imagine trying to go through all the Dynax mahjong and hanafuda games to work out which ones can optionally use F.F. for don den (with the option to use Start), and which can optionally use the double-up game controls.  You could also argue that the double-up game controls are optional for poker and slots games where the double-up game can be disabled.

PR #12455, opened 2024-06-08
msxr_cart.xml: Added 4 not working items.

wilbertpol:
Losaben Akel (demo) [file-hunter]
MSX View (Japan, v1.21) [file-hunter]
MSX View (Japan, v1.00) [file-hunter]
Zone Terra (demo) [file-hunter]

### DIFF
--- a/hash/msxr_cart.xml
+++ b/hash/msxr_cart.xml
@@ -1,0 +1,106 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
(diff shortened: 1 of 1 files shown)


cuavas:
Are these all cartridges that are only supported on Turbo-R?  How do you plan to deal with cartridges that support previous MSX generations but take advantage of Turbo-R functionality if available (assuming such things exist)?

wilbertpol:
i would put such cartridges in the cartridge list of the previous generation.

These are the only Turbo-R specific cartridges I came across so far. I don't think there will be much, at that time most software was being released on floppies.

PR #12456, opened 2024-06-08
msx softlists: Rename gtin info entries to barcode.

wilbertpol:
msx1_cart.xml, msx1_cass.xml, msx2_cart.xml, msx2_flop.xml: Rename gtin info entries to barcode.

### DIFF
--- a/hash/msx1_cart.xml
+++ b/hash/msx1_cart.xml
@@ -317,7 +317,7 @@ Ultra Chess - [K085] Al Alamiah
 		<year>1986</year>
 		<publisher>Koei</publisher>
 		<info name="alt_title" value="蒼き狼と白き牝鹿 ジンギスカン" />
-		<info name="gtin" value="04988615000225"/>
+		<info name="barcode" value="4988615000225"/>
 		<info name="usage" value="Requires a Japanese system."/>
 		<part name="cart" interface="msx_cart">
 			<feature name="pcb" value="TA-621 KAGA" />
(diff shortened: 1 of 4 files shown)


cuavas:
Does changing it to be less specific help?  If these are actually GTINs as opposed to some other barcode scheme, I would expect using a more specific `gtin` info item would be more useful.  If they’re a specific subtype of GTIN (e.g. ISBN, ISSN, EAN), they can use a even more specific info item.

wilbertpol:
I changed them to barcode because that is a standard defined info item as described here: https://docs.mamedev.org/contributing/softlist.html

The barcodes on the Japanese follow the same rules as EANs, just with a different country prefix of course.

cuavas:
It gets more complicated when things have multiple barcodes.  We’ve already got things with both a GTIN and an ISBN in msx2_flop.xml.  Taking Chaos Angel as an example, I’m not sure what the ISBN even belongs to – the GTIN is a Japan Article Number found on the case.

Then there are things that *only* have an ISBN like sample code disks included with textbooks (e.g. `ibm5170:apg_svga` – it has the ISBN of the book the disk was included with).

wilbertpol:
Ah, yes. Thanks for pointing out that isbn code in msx2_flop, I have to check those out as well. Regular isbn (prefixed with 978) is a regular gtin subtype.

wilbertpol:
I do not know where the isbn code on Chaos Angel came from. It is not on the packaging or the manual; It is also an invalid isbn-10 code so I have removed it. Which fixes all the entries with both barcode and isbn.
One existing isbn code was not valid, it has been marked as such. All the other isbn codes in these software lists are valid codes.

PR #12465, opened 2024-06-10
neogeo.xml: Added 1 working item

ArcadeShadow:
New working software list addition
GladMort (demo 2) [PixelHeart]

### DIFF
--- a/hash/neogeo.xml
+++ b/hash/neogeo.xml
@@ -11622,7 +11622,7 @@ license:CC0-1.0
 	</software>
 
 	<software name="gladmort">
-		<description>GladMort (demo 2)</description>
+		<description>GladMort (demo²)</description>
 		<year>2024</year>
 		<publisher>PixelHeart</publisher>
 		<sharedfeat name="release" value="MVS,AES" />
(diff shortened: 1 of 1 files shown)


hash/neogeo.xml
```diff
@@ -11622,7 +11622,36 @@ license:CC0-1.0
 	</software>
 
 	<software name="gladmort">
-		<description>Gladmort (demo)</description>
+		<description>GladMort (demo 2)</description>
```
angelosa, line 11625: What `demo 2` is supposed to mean, versioning?

ArcadeShadow: It's a new available demo, which title contains "demo 2"

angelosa: It's supposed to be `(demo²)` then.

ArcadeShadow: > What `demo 2` is supposed to mean, versioning?

OK. Fair enough.
This means that the Amiga CD32 should also be written with superscript style?
image

happppp: No. Unless maybe the MAME Amiga driver author(s) insists on it. Likewise, we don't have Odyssey².

angelosa: It's fine for me, yes, assuming it's changed in `amiga/amiga.cpp` too.

PR #12470, opened 2024-06-12
recfranco/rf51w14cg.cpp: Add new slot games on similar hardware

clawgrip:
Santa Fe Mine (set 1) [Sergio y Vicente de electronicaVyS, Recreativas.org]

New clones marked not working
-----------------------------
Santa Fe Mine (set 2) [Sergio y Vicente de electronicaVyS, Recreativas.org]

### DIFF
--- a/src/mame/recfranco/rf51w14cg.cpp
+++ b/src/mame/recfranco/rf51w14cg.cpp
@@ -2,7 +2,7 @@
 // copyright-holders:
 /*******************************************************************************************
 
-    Skeleton driver for Recreativos Franco "RF51W14-CG" and other games on smiliar hardware.
+    Skeleton driver for Recreativos Franco "RF51W14-CG" and other games on similar hardware.
     "RF51W14-CG" is the legal registry name, the game commercial name is unknown.
 
     TODO:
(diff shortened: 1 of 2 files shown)


src/mame/recfranco/rf51w14cg.cpp
```diff
@@ -109,15 +112,65 @@ static INPUT_PORTS_START( rf51w14cg )
...
+	I8255A(config, "pia2");
+	I8255A(config, "pia3");
+
+	// PIA on PIAs PCB
+	I8255A(config, "pia4");
+
+	SPEAKER(config, "mono").front_center();
+
+	// Sound PCB
+	OKIM6376(config, m_okim6376, XTAL(9'830'400)/64).add_route(ALL_OUTPUTS, "mono", 1.0); // Guessed frequency and divisor, the Sound PCB has no xtal.
```
happppp, line 165: It's an RC osc. according to schematics, so don't add an imaginary XTAL.
If you want the frequency to be the same as the other driver: 153600 // RC OSC, guessed frequency

src/mame/recfranco/rf51w14cg.cpp
```diff
@@ -1,8 +1,8 @@
 // license:BSD-3-Clause
 // copyright-holders:
-/******************************************************************************
+/*******************************************************************************************
 
-    Skeleton driver for Recreativos Franco "RF51W14-CG".
+    Skeleton driver for Recreativos Franco "RF51W14-CG" and other games on smiliar hardware.
```
happppp, line 5: typo: similar*

src/mame/recfranco/rf51w14cg.cpp
```diff
@@ -109,15 +112,65 @@ static INPUT_PORTS_START( rf51w14cg )
...
 }
 
+void rf51w14cg_state::rfsantafem(machine_config &config)
+{
+	I80188(config, m_maincpu, 32_MHz_XTAL / 2); // N80C188XL20
+
+	// PIAs on CPU PCB
+	I8255A(config, "pia1");
+	I8255A(config, "pia2");
+	I8255A(config, "pia3");
```
happppp, line 157: Do the schematics call them "PIA"? If so: OK. If not: 8255 is PPI.
Same for above where you added 6 of them, and one more below.

clawgrip: Fixed, thanks!

src/mame/recfranco/rf51w14cg.cpp
clawgrip: Fixed, thanks!

PR #12473, opened 2024-06-12
sony/news_r3k.cpp, sony/news_r4k.cpp, sony/news_68k.cpp, sony/news_38xx.cpp: Add software list for Sony NEWS systems

briceonk:
Because both the r3k and r4k have some level of binary compatibility, I added these into the same software list even though the current entries only work for OS installs on one or the other, so I can change this if having separate ones would be preferred.

### DIFF
--- a/hash/sony_news.xml
+++ b/hash/sony_news.xml
@@ -3,14 +3,14 @@
 <!--
 license:CC0-1.0
 -->
-<softwarelist name="sony_news_risc" description="Sony NEWS RISC (MIPS) software">
+<softwarelist name="sony_news" description="Sony NEWS software">
 
 	<software name="nwf_672rb" supported="partial">
 		<description>NEWS-OS Release 4.1R Version Up Kit</description>
 		<year>1991</year>
 		<publisher>Sony</publisher>
 		<notes>The floppies can boot on both CISC and RISC systems due to some bootloader magic, but the MO image is only for the first wave of RISC NEWS machines (NWS-3000 series)</notes>
-		
+		<sharedfeat name="compatibility" value="RISC" />
 		<!--
 		The installation floppy and MO disk work, but the format disk hangs after printing a few messages on RISC systems. 
 		Therefore, use of this kit requires use of a pre-formatted hard disk image for now.
@@ -44,6 +44,7 @@ license:CC0-1.0
 		<year>1993</year>
 		<publisher>Sony</publisher>
 		<notes>Usable only on NWS-5000 series computers</notes>
+		<sharedfeat name="compatibility" value="RISC" />
 
 		<!-- Source: WinWorld -->
 		<!-- The "STAMP" file has the build date of 05/05/93 -->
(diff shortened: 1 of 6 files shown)


rb6502:
I have a similar situation with the Mac drivers.  In the software list, I tag items with things like this for M68K only:
`<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040" />`

or this for M68K and PowerPC:
`<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040,PPC601" />`

And then in the driver,
`SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("MC68030,MC68040");`
to indicate only software that's compatible with a 68030 or 68040 should match for this machine

or
`SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("PPC601");`

For most machines you don't need to get that specific, "CISC" and "RISC" would probably work for NeWS, but it's nice that you can get that fine-grained

briceonk:
> I have a similar situation with the Mac drivers.  In the software list, I tag items with things like this for M68K only:
> 
> `<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040" />`
> 
> 
> 
> or this for M68K and PowerPC:
> 
> `<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040,PPC601" />`
> 
> 
> 
> And then in the driver,
> 
> `SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("MC68030,MC68040");`
> 
> to indicate only software that's compatible with a 68030 or 68040 should match for this machine
> 
> 
> 
> or
> 
> `SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("PPC601");`
> 
> 
> 
> For most machines you don't need to get that specific, "CISC" and "RISC" would probably work for NeWS, but it's nice that you can get that fine-grained

Cool! I'll update this. I won't have access to my computer for a few days but I'll update it when I get back!

briceonk:
> I have a similar situation with the Mac drivers. In the software list, I tag items with things like this for M68K only: `<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040" />`
> 
> or this for M68K and PowerPC: `<sharedfeat name="compatibility" value="MC68000,MC68020,MC68030,MC68040,PPC601" />`
> 
> And then in the driver, `SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("MC68030,MC68040");` to indicate only software that's compatible with a 68030 or 68040 should match for this machine
> 
> or `SOFTWARE_LIST(config, "cd_list").set_original("mac_cdrom").set_filter("PPC601");`
> 
> For most machines you don't need to get that specific, "CISC" and "RISC" would probably work for NeWS, but it's nice that you can get that fine-grained

Updated! Let me know if there is anything else you'd like me to change :)

cuavas:
Where do these dumps come from?  Are they from your CDs, or did you get them from somewhere else?

briceonk:
The 4.2.1aR is from WinWorld, the 4.1R floppy and Magento-optical dumps are from my upgrade set.

PR #12485, opened 2024-06-18
cops.cpp: Promote Revelations to working

james-wallace-ghub:
-cops.cpp: Added CHD for Revelations [James Wallace, AintBigAintClever]

-ldp1450hle.cpp: Create an HLE implementation of the LDP1450 player [James Wallace]

Systems promoted to working
Revelations [James Wallace, AintBigAintClever]


rb6502:
@MooglyGuy can you take a look at this since it's LD?

MooglyGuy:
James has been keeping me in the loop, it's fine by me. I vaguely recall it being a more traditional capture-card run rather than an ld-decode run, but given the finicky process and rarity of discs, I'm happy to see work being done on bringing up the surrounding system in the hope that some ld-decode captures show up.

src/mame/atari/cops.cpp
```diff
@@ -722,15 +414,15 @@ void cops_state::io2_w(offs_t offset, uint8_t data)
 		case 0x02:
 			m_flash_red_lamp = BIT(data, 0);
 			m_flash_blue_lamp = BIT(data, 7);
-			if ( data & ~0x91 ) logerror("Unknown io2_w, offset = %02x, data = %02x\n", offset, data);
+//			if ( data & ~0x91 ) logerror("Unknown io2_w, offset = %02x, data = %02x\n", offset, data);
```
rb6502, line 424: These should be LOGMASKED() rather than commented out if they're of interest.  If you're done with them, remove them entirely.

james-wallace-ghub:
> James has been keeping me in the loop, it's fine by me. I vaguely recall it being a more traditional capture-card run rather than an ld-decode run, but given the finicky process of getting an ld-decode capture and rarity of discs, I'm just happy to see work being done on bringing up the surrounding system. Hopefully some ld-decode captures show up in the future.

That is indeed the case, I'm not particularly happy about the quality of the player it was run on either, but I can't raise the original provider since it's been so long. I've labelled it BAD_DUMP as I would like to see something better, or at least confirmatory as to how bad the picture quality is on this. I can believe it was always this washed out, and it certainly looks like the Platoon/COPS discs made by the same people, but it's not really top quality IMHO.

src/mame/atari/cops.cpp
```diff
@@ -56,15 +57,20 @@ namespace {
 #define AUX_REGISTER 1
 
 #define MAIN_CLOCK XTAL(4'000'000)
+#define DACIA_CLOCK XTAL(3'686'400)
```
rb6502, line 68: You can use 3.6864_MHz_XTAL for this and not have to have a #define since it's a common value with ACIAs.

james-wallace-ghub: There's a few crystals I haven't done that with, there's even a small one on the RTC I missed so I'll deal with those.

As for the 6552 - there's just enough different that it is an issue to merge with 6551 (the addition of comparison registers, the reworking of some of the status pins to do different things. It's effectively a double up of a slightly different design). However, I have at least put this in a more 'modern' framework which might facilitate a merge if others feel it suitable.

The most recent commit addresses the crystals, uses the more modern shorthand and structure for the devices (not the driver as that's preexisting code) and adds the 'Ltd' to fit the company trading name. Hope this addresses the concerns and that, if nothing else, the 1450 might be useful for something people might like ;).

MooglyGuy: Beyond the concerns that were already brought up, I believe this PR should go in without any fuss regarding the 6552 vs. 6551 distinction.

Given the current implementation of the 6551 device using the read/write paradigm, it'd be difficult to override any specific register behavior for the sake of the 6552 device.

I'd much rather see these changes go in to improve the overall driver, and then have a separate pull request for improving the 6551/6552 split (since it seems like @james-wallace-ghub is planning to do that), than to stand on ceremony over a tangential peripheral chip.

james-wallace-ghub: Yes, I think my next plan would be to try to unify based on the 'new' structure rather than the old, but I want to try to get my head around that first. In this situation, the drivers using the 6551 are unaffected, and then I can merge and test at relative leisure.

rb6502: Yeah, I'm fine with the 6552 as it was originally submitted, I just wanted the crystal constant used and the commented-out logerrors resolved.  Please resist the temptation to pile on more work when I'm just asking for a few simple changes, it makes things easier for me and for you.

james-wallace-ghub: If leaving it at this is ok, I'm not intending to go further. I just thought a map with the 6552 would help for any future work outside of this scope.

src/devices/machine/r65c52.cpp
```diff
@@ -0,0 +1,987 @@
...
+{
+	bool irq = false;
+	LOG("R65C52: %x  IER  %x ISR %x\n", idx + 1, m_ier[idx], m_isr[idx]);
+	for (int i = 0; i < 8; i++)
+	{
+		if ((m_ier[idx] & (1 >> i)) && ((m_isr[idx] & (1 >> i))))
+		{
+			irq = true;
+		}
+	}
```
cuavas, line 268: How is this supposed to work?  `(1 >> i)` will be zero whenever `i` is not zero, so the loop is pointless – it only checks a bit on the first iteration.

james-wallace-ghub: That's meant to be a bitwise comparison (i.e is the same bit set in the relevant slice of m_ier as m_isr). However, it only does that once, as you say. I have a fix, that I will submit separately, as this has already been merged.

cuavas: You know you don’t need a loop at all to test if bits on corresponding positions are set…

james-wallace-ghub: Undoubtedly, but I'm still finding my feet with bitwise operations of this nature, so I thought I'd keep it relatively simple in the first phase rather than try an XOR. An optimisation might be useful here though, since this triggers a lot, but I'm not sure what best practice is. For now, I've been using the BIT macro to do it

cuavas: An XOR is wrong.  You want the intersection, not the disjunction.  And the loop is definitely not “simple” as this shows.

PR #12487, opened 2024-06-19
Benesse work / new NOT WORKING Machine (Challenge Gear)

mamehaze:
New NOT WORKING Machine
----------
Challenge Gear [TeamEurope, SSJ, Youkal3]

New NOT WORKING Software List Entries
---------
challenge_gear_cart.xml:
Challenge 6 Years [TeamEurope, SSJ, Youkal3]

- moved pockchal.cpp and the new driver to a benesse folder, as benesse had many more products

- created a separate machine for the Pocket Challenge v2, the BIOS isn't dumped, but carts aren't physically compatible with a Wonderswan even if the software is.  Makes more sense to have a driver with the borrowed Wonderswan BIOS marked as BAD_DUMP than hiding everything behind the Wonderswan itself.

### DIFF
--- a/hash/challenge_gear_cart.xml
+++ b/hash/challenge_gear_cart.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+
+<!--
+
+Cartridge:
+FRONT:                BACK:
++~~~~~~~~~~~~~~~~~+   +~~~~~~~~~~~~~~~~~+
+|                 |   |                 |
+|    +~~~~~~~+    |   |                 |
+|    |TSOP 32|    |   |                 |
+|    +~~~~~~~+    |   |                 |
++~~+20       01+~~+   +~~+21       40+~~+
+   |___________|         |___________|
+
+
+01 | GND        21 | GND
+02 | D01        22 | D02
+03 | D00        23 | D03
+04 | OE         24 | D04
+05 | NC         25 | GND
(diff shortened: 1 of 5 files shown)


mamehaze:
this ties to the work @ajrhacker is doing on the new CPU core.  As there's not even a stub of an execution core to hook it up to yet you can't do a -debug with the new driver and get the dasm at this point.

mamehaze:
Console no.1

mamehaze:
Console no.2

mamehaze:
Cartridge

asiekierka:
> created a separate machine for the Pocket Challenge v2, the BIOS isn't dumped, but carts aren't physically compatible with a Wonderswan even if the software is. 

That's not entirely accurate.

* The Pocket Challenge V2's BIOS is bit-identical to the mono WonderSwan BIOS, as both consoles use identical SoCs. See [this tweet from Adam Jaremko](https://x.com/adamjaremko/status/1163284248884891650) - the difference in behaviour is achieved by a difference in chip wiring, but not in chip firmware.
* While it is true that Pocket Challenge V2 cartridges are not physically compatible, this is merely due to a difference in shape; once disassembled, the PCBs themselves are electrically compatible with the WonderSwan, as evidenced by [being able to put a PCv2 cartridge in a WonderSwan](https://x.com/adamjaremko/status/1163485972631347200) and [vice versa](https://x.com/kyenprod/status/1532418259504472064).

mamehaze:
well better to further the documentation then, and find what controls it skipping the logo so it can be implemented that way.

from the point of view of an end user they are incompatible systems though.

PR #12499, opened 2024-06-21
unico/goori.cpp: Use generic gfx decode layout, Cleanups

cam900:
- Fix ROM region naming
- Correct type for variables
- Make variables constant
- Reduce literal tag usages
- Fix spacing


angelosa:
I'm on this.

PR #12504, opened 2024-06-22
cpu/xa: New Philips XA disassembler 

mamehaze:
- added a disassembler for the XA architecture from Philips which is sometimes referred to as  80c51XA (it isn't binary compatible with an 8051 in any way, just designed to have a similar enough instruction set and memory layout that code can be translated easily over using software Philips provided)

- hooked this up to igs_fear.cpp (fearless, superkds) and igs_m027.cpp (haunthig, crzybugs)


mamehaze:
The output here is comparable to IDA Pro, and I've used that as a template for the general formatting.  A few bugs were found in the IDA Pro output (wrong register sizes used)

The code in general could do with a bit of beautification, but is functional, IMHO easy to follow, and will help unlock further progress on the IGS titles (specifically superkds for now, as we have a complete dump of that one)

@ajrhacker could you give this a quick lookover?

src/tools/unidasm.cpp
```diff
@@ -729,6 +730,7 @@ static const dasm_table_entry dasm_table[] =
 	{ "xavix2",          le,  0, []() -> util::disasm_interface * { return new xavix2_disassembler; } },
 	{ "xerox530",        be, -1, []() -> util::disasm_interface * { return new xerox530_disassembler; } },
 	{ "xtensa",          le,  0, []() -> util::disasm_interface * { return new xtensa_disassembler; } },
+	{ "xa",              le,  0, []() -> util::disasm_interface * { return new xa_dasm; } },
```
rb6502, line 733: These are in alphabetical order, so this should be just before Xavix.

PR #12507, opened 2024-06-24
atastorage.h: increase ATA cylinders to u32

holub:
LBA cylinders number is not limited to 0xffff.
image


holub:
Possibly we shouldn't extend cylinders over 0xffff but prevent chdman from creating such images.

PR #12513, opened 2024-06-26
rendersw: Generate cosine table at compile time

FlykeSpice:


rb6502:
This looks reasonable to me, but compile-time execution is a bit beyond my daily-driver C++ subset so I'm going to defer to whichever of @cuavas or @galibert for the final call.

galibert:
I love it, didn't even realize that was possible that readibly.  But it needs to be tested for all the minimal versions of compilers with support, including the visual studio stuff...

rb6502:
@pmackinlay I think you use MSVC?  Can you check that this compiles there?

MooglyGuy:
MSVC 2022 says no:
```Severity	Code	Description	Project	File	Line	Suppression State	Details
Error	C3892	's_cosine_table': you cannot assign to a variable that is const	emu	C:\mingw\mame\src\emu\rendersw.hxx	454		
Error	C3892	's_cosine_table': you cannot assign to a variable that is const	osd_windows	C:\mingw\mame\src\emu\rendersw.hxx	454		```

MooglyGuy:
Please disregard my previous comment which I have now deleted, as it was in error; the diff didn't quite apply correctly against my local changes.

This compiles just fine on MSVC 2022.

PR #12514, opened 2024-06-26
hp_dio: add HP98624 GPIB interface card

svenschnelle:


svenschnelle:
This adds support for the HP98624 GPIB card. It is described here: http://www.hpmuseum.net/display_item.php?hw=508

src/devices/bus/hp_dio/hp98624.h
```diff
@@ -0,0 +1,57 @@
+// license:BSD-3-Clause
+// copyright-holders:Sven Schnelle
+
+#ifndef MAME_BUS_HPDIO_98624_H
```
rb6502, line 4: For slot cards we're moving towards a new structure where basically nothing is in the header and the .cpp is encapsulated in an anonymous namespace and then you use DEFINE_DEVICE_TYPE_PRIVATE() to register it.  This makes the card completely self-contained and reduces work for the archiving and linking stages when building MAME.

A good example of a brand-new card device written that way is nubus/8lc.cpp and .h.

I'm not going to reject this if you don't want to do that yet, but it can get you a head start on figuring out how you want to evolve the DIO bus.

rb6502 (approved):

src/devices/bus/hp_dio/hp98624.h
svenschnelle: The DIO bus lives in namespace bus::hp_dio, so that would mean i have to convert all the devices to the anonymous namespace?

rb6502: You can inherit from a namespaced bus and still have the card implementation in an anonymous namespace.  bus/intellec is a great example from Vas (I'm looking at intellec/prommemory.h).  The only real difference is using DECLARE_DEVICE_TYPE_NS() in the .h file.

PR #12518, opened 2024-06-27
gaelco/gaelcopc.cpp: Add an alt BIOS for Tuning Race [Museo del Recreativo, Recreativas.org]

clawgrip:
Also add a note about Tuning Race I/O board and clean up comments


clawgrip:
Nah! Was the same BIOS with different ACFG

PR #12522, opened 2024-06-28
heathzenith/h89.cpp: Fix shadow memory writes

mgarlanger:
Fixes shadow writes when in ROM mode. 

- This now properly writes to the floppy RAM which fixes a boot issue with HDOS 2.0.

### DIFF
--- a/src/mame/heathzenith/h89.cpp
+++ b/src/mame/heathzenith/h89.cpp
@@ -725,8 +725,16 @@ void h89_base_state::machine_start()
 		m_maincpu->space(AS_PROGRAM).install_ram(0x2000, 0xffff, m_ram_ptr);
 
 		// install shadow writing to RAM when in ROM mode
-		m_mem_view[0].install_writeonly(0x0000, 0x1fff, m_ram_ptr + 0xe000);
-		m_mem_view[1].install_writeonly(0x0000, 0x1fff, m_ram_ptr + 0xe000);
+		m_mem_view[0].install_write_tap(0x0000, 0x1fff, "shadow_w", [this](offs_t offset, u8 &data, u8 mem_mask)
+		{
+			u8 *to = m_ram->pointer() + 0xe000;
+			to[offset] = data;
+		});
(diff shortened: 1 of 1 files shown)


src/mame/heathzenith/h89.cpp
```diff
@@ -725,8 +725,16 @@ void h89_base_state::machine_start()
 		m_maincpu->space(AS_PROGRAM).install_ram(0x2000, 0xffff, m_ram_ptr);
 
 		// install shadow writing to RAM when in ROM mode
-		m_mem_view[0].install_writeonly(0x0000, 0x1fff, m_ram_ptr + 0xe000);
-		m_mem_view[1].install_writeonly(0x0000, 0x1fff, m_ram_ptr + 0xe000);
+		m_mem_view[0].install_write_tap(0x0000, 0x1fff, "shadow_w", [this](offs_t offset, u8 &data, u8 mem_mask)
+		{
+			u8 *to = m_ram->pointer() + 0xe000;
```
rb6502, line 730: A few things.

1. The m_ prefix indicates a class *m*ember, which m_ram_ptr isn't.
2. I don't understand how this is functionally any different from what it was doing before.  In both cases writes to the 0x0000-0x1fff range go to 0xe000-0xffff in RAM, it just has more overhead now.

rb6502 (changes requested):

src/mame/heathzenith/h89.cpp
mgarlanger: I'll update the m_ram_ptr variable name.

With the previous code, the writes to floppy ram - defined here: https://github.com/mamedev/mame/pull/12522/files#diff-c18528b917f86f6bc2a8b58c889ac0919a61b93edcba82997c6cfe70bfe1e561R313-R315 were not occurring, which prevented HDOS 2.0 from booting.  Actually since m_mem_view[0] is the write protected, this only needs to be defined for m_mem_view[1] where the floppy ram is write-enabled.

PR #12524, opened 2024-06-28
machine/ncr5385.cpp: raise IRQ for INT_INVALID_CMD

Elektraglide:
- Tek4404 issues cmd=0xff which expects an IRQ even though its an INT_INVALID_CMD
- minor readability; replaced magic hex value with symbolic names

### DIFF
--- a/src/devices/machine/ncr5385.cpp
+++ b/src/devices/machine/ncr5385.cpp
@@ -23,7 +23,7 @@
 #define LOG_DMA      (1U << 4)
 #define LOG_COMMAND  (1U << 5)
 
-//#define VERBOSE (LOG_GENERAL|LOG_REGW|LOG_REGR|LOG_STATE|LOG_DMA|LOG_COMMAND)
+#define VERBOSE (LOG_GENERAL|LOG_REGW|LOG_REGR|LOG_STATE|LOG_DMA|LOG_COMMAND)
 #include "logmacro.h"
 
 DEFINE_DEVICE_TYPE(NCR5385, ncr5385_device, "ncr5385", "NCR 5385 SCSI Protocol Controller")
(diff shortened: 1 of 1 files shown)


src/devices/machine/ncr5385.cpp
```diff
@@ -23,7 +23,7 @@
 #define LOG_DMA      (1U << 4)
 #define LOG_COMMAND  (1U << 5)
 
-//#define VERBOSE (LOG_GENERAL|LOG_REGW|LOG_REGR|LOG_STATE|LOG_DMA|LOG_COMMAND)
+#define VERBOSE (LOG_GENERAL|LOG_REGW|LOG_REGR|LOG_STATE|LOG_DMA|LOG_COMMAND)
```
rb6502: Please turn this back off so it's not spamming the logs :-)

src/devices/machine/ncr5385.cpp
```diff
@@ -816,7 +817,9 @@ void ncr5385_device::set_dreq(bool dreq)
 
 void ncr5385_device::update_int()
 {
-	bool const int_state = m_int_status & 0x5f;
+	bool const int_state = m_int_status & (INT_FUNC_COMPLETE | INT_BUS_SERVICE |
```
rb6502, line 820: Excellent clarification!

cuavas:
Also, **please give the pull request a meaningful title**.  What does it change?  What does it fix?

Elektraglide:
> Also, **please give the pull request a meaningful title**. What does it change? What does it fix?

Title amended.  In large code base like this,  I've generally used PR titles to give 'area of interest' for reviewers and used following lines to explain the details.   But if having a detailed title is the MAME thang, happy to go with it. :-)

PR #12525, opened 2024-06-28
spg2xx: Switch ADPCM algorithm to IMA ADPCM

sp1187:
Based on some investigations I made of V.Smile sound samples and the waveforms generated by different algorithms, it seems very likely to me that the ADPCM algorithm used in SPG2xx is actually IMA rather than OKI.

This PR adds a new class for IMA ADPCM adapted from the existing OKI ADPCM class and updates the SPG2xx machine to use this instead.

### DIFF
--- a/scripts/src/sound.lua
+++ b/scripts/src/sound.lua
@@ -700,6 +700,20 @@ end
 
 
 ---------------------------------------------------
+-- IMA ADPCM sample player
+--@src/devices/sound/imaadpcm.h,SOUNDS["IMAADPCM"] = true
+---------------------------------------------------
+
+if (SOUNDS["IMAADPCM"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/sound/imaadpcm.cpp",
+		MAME_DIR .. "src/devices/sound/imaadpcm.h",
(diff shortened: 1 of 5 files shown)


mamehaze:
Interesting, I'll give this a test.  One of the main games I was having issues with was 'Tiger Rescue' which is fine on some of the units, but has really bad sound in places on others - I suspect they recompressed the audio for those units, maybe as ADPCM.

mamehaze:
fwiw Tiger Rescue in lexizeus still sounds awful with this (once the first larger plane appears and you start shooting it)
This happens in some other units too in MAME, and definitely doesn't happen on hardware (this was dumped from my unit)

it does not happen in "lx_jg7415"  (I think it's about game 45 in the minigames) or the other larger multigames.

so they definitely changed something, and this doesn't fix it (maybe lexizeus uses a different SoC with more / less compression modes, and either it's picking one not available on that chip and incorrectly using it, or it's simply one that isn't implemented?)

mamehaze:
anyway maybe an interesting test case for something if you want to dig around a bit more with SPG audio @sp1187

PR #12526, opened 2024-06-29
heathzenith/z29.cpp: Add some documentation based on ROM listing

mgarlanger:
Add port definitions and memory map based on ROM listing for the terminal.

### DIFF
--- a/src/mame/heathzenith/z29.cpp
+++ b/src/mame/heathzenith/z29.cpp
@@ -74,11 +74,30 @@ void z29_state::keyin_w(int state)
 	m_keyin = state;
 }
 
+/**
+ * Port 1 (based on ROM listing)
+ * 
+ * bit 0 - KB output line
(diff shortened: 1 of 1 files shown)


src/mame/heathzenith/z29.cpp
```diff
@@ -124,12 +143,23 @@ void z29_state::prg_map(address_map &map)
 	map(0x6000, 0x67ff).mirror(0x800).r(FUNC(z29_state::bs_24k_r));
 }
 
+/**
```
rb6502, line 146: I don't think this comment is a value add over the actual ext_map() below it, other than the 0x6000 range.  Assuming the ext_map() has the correct addresses for the rest, I'd just add a comment between the entries for 0x5000 and 0x7000 stating that 0x6000 is DMA memory (and anything else the listing might tell about it).

PR #12539, opened 2024-07-02
scripts/emu.lua: Remove redundant "precompile" project

FlykeSpice:
I dunno why is it there but it doesn't seem to participate in anything besides being a stub.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1123,7 +1123,7 @@ windows_x64: generate $(PROJECTDIR)/$(MAKETYPE)-mingw64-gcc/Makefile
 ifndef MINGW64
 	$(error MINGW64 is not set)
 endif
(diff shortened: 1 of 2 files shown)


cuavas:
The precompile target ensures the precompiled prefix header is compiled before the main target is built (you can see the make file for it at e.g. build/projects/windows/mame/gmake-mingw64-gcc/precompile.make for a MinGW GCC build).

FlykeSpice:
> The precompile target ensures the precompiled prefix header is compiled before the main target is built (you can see the make file for it at e.g. build/projects/windows/mame/gmake-mingw64-gcc/precompile.make for a MinGW GCC build).

Actually, this is automatically ensured by GENie on the generated makefile that the precompiled header must be compiled before sourcefiles that includes/depends on it, otherwise the build system would be broken (Just look at build/projects/.../emu.make which uses precompiled header)

besides the precompiled header must be compiled before the sourcefile that uses it not before the linking of target.

PR #12545, opened 2024-07-04
nmk/nmk004.cpp: redumped internal ROM of NMK004 sound CPU

sergiopolog:
It was dumped from two different devices taken from `acrobatm` and `tdragon` boards.

[NMK004 dump](https://imgur.com/pi6B6jc)

The dump was exactly the same from those 2 devices, so I consider it a good and reliable dump. It was achieved using a custom tool to read internal ROMs from TLCS90 MCUs: [https://github.com/sergiopolog/tlcs90-rom-reader](https://github.com/sergiopolog/tlcs90-rom-reader)

Previously existing dump made by 'trap15' only differs in 3 bytes starting at 0xFBD address. That difference could be caused by the method used to extract the ROM back then.

Actually, I talk to him about it and he stated that area around 0xFBF address was marked as unsure when he carried out the process.


cuavas:
That’s a pretty cool exploit, and it should work for any device in this family.

PR #12546, opened 2024-07-05
viper.cpp: implement basic LAN controller usage

Hydreigon223:
Fixes a lot of problems related to the LAN controller and hooked it to MPC8240 IRQ 1.

Added lightgun inputs for wcombatu as this specific version uses optical lightguns.

Marked CF dumps for thrild2ac and thrild2c as bad.

### DIFF
--- a/src/mame/konami/kviper_lanc.cpp
+++ b/src/mame/konami/kviper_lanc.cpp
@@ -1,0 +1,138 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/***************************************************************************
+
+Konami Viper LAN Controller
+
+TODO: nearly everything
+
+***************************************************************************/
+
+#include "emu.h"
+
+#include "kviper_lanc.h"
+
+#define LOG_READS   (1U << 1)
+#define LOG_WRITES  (1U << 2)
+#define LOG_RAM_READS   (1U << 3)
+#define LOG_RAM_WRITES  (1U << 4)
+#define LOG_ALL (LOG_READS | LOG_WRITES | LOG_RAM_READS | LOG_RAM_WRITES | LOG_UNKNOWNS)
+
+#define VERBOSE (0)
+#include "logmacro.h"
+
+DEFINE_DEVICE_TYPE(KVIPER_LANC, kviper_lanc_device, "kviper_lanc", "Konami Viper LAN Controller")
+
+
+kviper_lanc_device::kviper_lanc_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
+	: device_t(mconfig, KVIPER_LANC, tag, owner, clock)
+	, m_irq_cb(*this)
+{
+}
+
+void kviper_lanc_device::device_start()
+{
+	save_item(NAME(m_irq_state));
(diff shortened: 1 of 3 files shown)


MooglyGuy:
Isn't this just a duplicate of the existing LANC skeleton? It doesn't seem to really be "implementing" anything. I would split the lightgun and bad-dump changes into a separate pull request.

Hydreigon223:
That's what I thought at first too. On the surface some registers are similar to 056230 but after some behavior analysis on the various games that use this (code1d, xtrial, gticlub2, thrilld, wcombat and ppp2nd), I noticed slight differences:

- Buffer ram is always fixed to have eight 0x400 ram pages in Viper where 056230 has two additional bits configure the number of ram pages (two, four or eight).
- Three bits determine the size buffer memory capacity per node instead of two.
- Network ID written is determined by a single bit instead of a three bit combination (i.e. bit 0 - ID1, bit 1 - ID2, bit 2 - ID3...and so on).
- The CRC error register, while same, is read at an offset of 0x03 instead of 0x01.
- The wake/sleep bit is written at an offset at 0x05 at bit 0 instead of offset 0x01 bit 5.

I will document most of these changes later on today.

MooglyGuy:
None of these are particularly significant differences. You should be subclassing the existing LANC device if that's the case, not reinventing the wheel.

cuavas:
You might be better off opening a discussion to talk about the differences you’ve discovered and how you discovered them.

Hydreigon223:
Been thinking about this for a while. Mog's suggestion isn't a bad idea. I would move code from the old 056230 device to here (or so). Furthermore, I may as well implement this for sscope2 as its LAN behavior has similarities with 056230 and Viper making it the missing link between the two.

As for how I discovered these, it's with debugger creating a watchpoint for the LAN related memory areas they utilize (e.g. how bits 4-6 in Viper's control_w correspond with the area of buffer ram the system reads/writes to).

Gurudumps:
Both a41a02_alt and a41a02_alt2 can just be deleted.  They are using the same aaa nvram. The proper CF a41a02 is already dumped. The different versions coming from that are done through the nvram. There is only one a41a02 cf and it's good so the other alt bad versions can just be removed.

As for Warzaid using guns light or IR guns, as far as I know all the Warzaid versions can use either type of gun. The Viper board has support for light guns built in, simply plug light guns into the extra 28 way connector as documented in my readme. For IR guns there's an extra IO board and IR sensors around the edge of the screen. Those versions are big twin cabs and use 50" rear-projection screens so the cabinet can't support light guns. The Viper board can use either type of gun, only the IR guns need additional hardware but those game versions should work with light guns if they are connected correctly.

If any of that PR gets any of those games advancing further into the game then it should be accepted. Otherwise if there's no in-game improvements then that info should just be added as a note somewhere in the driver.

angelosa:
I applied & cleaned the LANC portions in https://github.com/mamedev/mame/commit/93060d50e33870a4c71c7c3166f5314450385927
which leaves (beyond the conflict) the wcombat changes from here.

Hydreigon223:
Closing this PR as I moved the wcombat changes to #12591

PR #12547, opened 2024-07-05
atari/atetris.cpp: Add a new Tetris Korean bootleg and redumped atetb5205 OKI ROMs

clawgrip:
New working clones
-------------------
Link (Korean bootleg of Atari Tetris) [Taksangs]


clawgrip:
About "Link", the only different ROM was damaged, but being a set never seen before, we think it's important to have it documented anyway.
About "atetb5205", we found two more PCBs in much better shape, so we could finally get and verify good dumps.

PR #12550, opened 2024-07-07
Correct the erroneous msgid and Chinese characters

agugu2000:
just fix some mistakes in chinese string po file.

### DIFF
--- a/language/Chinese_Simplified/strings.po
+++ b/language/Chinese_Simplified/strings.po
@@ -930,11 +930,11 @@ msgstr "未完美模拟功能: "
 
 #: src/frontend/mame/ui/info.cpp:133
 msgid "Screen flipping in cocktail mode is not supported.\n"
-msgstr "尚未支援台面型筐体模式的画面翻转。"
+msgstr "尚未支援台面型筐体模式的画面翻转。\n"
 
 #: src/frontend/mame/ui/info.cpp:135
 msgid "This system requires external artwork files.\n"
-msgstr "此系统需要额外的装饰图档"
+msgstr "此系统需要额外的装饰图档\n"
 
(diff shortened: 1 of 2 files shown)


agugu2000:
Pause the game when the main menu is activated
Resume the game when the main menu is deactivated

happppp:
Could you put that pause feature in a separate PR?
Though I can tell you beforehand: enforcing MAME to pause on Tab menu will be rejected. If it was optional, maybe.

agugu2000:
> Could you put that pause feature in a separate PR? Though I can tell you beforehand: enforcing MAME to pause on Tab menu will be rejected. If it was optional, maybe.

Yep,back to the string commit..
I will do your suggest later.Make an option will be better.
Errrr~~~While trying to enable invincible cheat before dead,you r already dead,hahhaha. 
An extra step to pause is just too inconvenient.

rb6502:
With the pause feature separated out, the actual fixes look completely reasonable (adding \n to the Chinese where the English also has it, and fixing the "Cheat Options" key).  @cuavas any comments on the actual Chinese-language changes here?

cuavas:
I’ll crib the corrections to the translations.  I don’t want to merge the changes to the string keys at this point, as having the translation files out-of-sync with each other is a pain and it’s better to regenerate them all at once.

There are two issues with regenerating them all at the moment:
* I need to get through more of the open translation PRs so they don’t get out of sync too badly.
* Some new menus have been added in a way that’s unfriendly for localisation, and it’s better to sort that out before regenerating the translation source files.

cuavas:
I applied the corrections in 84a5fe16185e375fe069e79aa8d9ff4b3c98758e.

PR #12555, opened 2024-07-07
Fixes/ROMs for Roland stuff

giulioz:
- Renames `rolandpcm` into `roland_lp` to make it distinct from other PCM chips
- Adds a skeleton for the GP (TC6116) chip
- Fixes the T6963C LCD controller by adding ram read support and correct text/graphics enable register handling. This makes the Roland D-70 behave correctly with message popups
   image

- Fixes the allocation of waveroms in the Roland D-70
- Adds the first ever dump of the internal mask ROM of the Roland D-50 CPU
- Fixes the memory map of the JV80 and wires up the PCM chip skeleton
- Adds a skeleton and ROM dumps for the RD-500

### DIFF
--- a/src/mame/roland/roland_jv80.cpp
+++ b/src/mame/roland/roland_jv80.cpp
@@ -37,7 +37,7 @@ private:
 
 void roland_jv80_state::jv880_mem_map(address_map &map)
 {
-	map(0x00000, 0x08000).rom().region("maincpu", 0);
+	map(0x00000, 0x07fff).rom().region("maincpu", 0);
 	map(0x08000, 0x0dfff).ram().mirror(0xa0000);
 	map(0x0f000, 0x0f3ff).rw(m_pcm, FUNC(tc6116_device::read), FUNC(tc6116_device::write));
 	map(0x10000, 0x3ffff).rom().region("progrom", 0x10000);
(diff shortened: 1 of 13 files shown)


src/mame/roland/roland_jv80.cpp
```diff
@@ -19,25 +21,30 @@ class roland_jv80_state : public driver_device
...
 	required_device<h8532_device> m_maincpu;
+	required_device<tc6116_device> m_pcm;
 };
 
 
-void roland_jv80_state::mem_map(address_map &map)
+void roland_jv80_state::jv880_mem_map(address_map &map)
 {
-	map(0x08000, 0x09fff).ram();
+	map(0x00000, 0x08000).rom().region("maincpu", 0);
```
galibert, line 40: That should be 0x7fff and not 0x8000.  Recommend you try a mame -valid before doing a submission, it doesn't take long and avoids duh-moments.

giulioz: Hi, just pushed a fix, thanks for the tip!
I'm making my first contributions and I'm still trying to learn how all the internal works, sorry.

giulioz:
@galibert I just checked the available rom zip files for the new mame version and the ones for the d50 got entirely replaced with my file only, instead of just adding it to the existing ones

PR #12561, opened 2024-07-09
Add HP9133 disk drive

svenschnelle:
This adds the HP9133 disk drive, which was common with HP9000 200 and 300 systems. It also adds a sync acknowledge callback to the 6809, and multi sector read/write to wd2010.

### DIFF
--- a/src/devices/bus/ieee488/hp9133.cpp
+++ b/src/devices/bus/ieee488/hp9133.cpp
@@ -1,0 +1,877 @@
+// license:BSD-3-Clause
+// copyright-holders: Sven Schnelle
+/*********************************************************************
+
+    hp9133.cpp
+
+    HP9133 hard disk/floppy drive
+
+*********************************************************************/
+
+#include "emu.h"
+#include "hp9133.h"
+#include "cpu/m6809/m6809.h"
+#include "machine/i8291a.h"
+#include "machine/wd2010.h"
+#include "machine/wd_fdc.h"
+#include "imagedev/floppy.h"
+#include "imagedev/harddriv.h"
+#include "ieee488.h"
+
+//#define VERBOSE 1
+#include "logmacro.h"
+
+namespace {
+
+class hp9133_device : public device_t,
+		      public device_ieee488_interface
+{
+public:
+	// construction/destruction
+	hp9133_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+protected:
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+	// device-level overrides
+	virtual ioport_constructor device_input_ports() const override;
+	virtual const tiny_rom_entry *device_rom_region() const override;
+	virtual void device_add_mconfig(machine_config &config) override;
+
+	// device_ieee488_interface overrides
+	virtual void ieee488_eoi(int state) override;
+	virtual void ieee488_dav(int state) override;
+	virtual void ieee488_nrfd(int state) override;
+	virtual void ieee488_ndac(int state) override;
+	virtual void ieee488_ifc(int state) override;
+	virtual void ieee488_srq(int state) override;
+	virtual void ieee488_atn(int state) override;
+	virtual void ieee488_ren(int state) override;
+
+private:
+
+	static constexpr int IO1_R_BCS_N = 0x10;
+	static constexpr int IO1_R_BDRQ = 0x20;
+	static constexpr int IO1_R_HDINT = 0x40;
+	static constexpr int IO1_R_HDIND = 0x80;
+
+	static constexpr int IO2_R_FDCINT = 0x10;
+	static constexpr int IO2_R_FDKDCH_N = 0x20;
+	static constexpr int IO2_R_WRTFLTERR = 0x80;
+
+	static constexpr int IO3_R_FASTDON = 0x20;
+	static constexpr int IO3_R_SELTST = 0x40;
+	static constexpr int IO3_R_TEST = 0x80;
+
+	static constexpr int IO1_W_FLT_N = 0x01;
+	static constexpr int IO1_W_DS1 = 0x02;
+	static constexpr int IO1_W_BRDY = 0x40;
+	static constexpr int IO1_W_WDRESET_N = 0x80;
+
+	static constexpr int IO2_W_FDCRES_N = 0x01;
+	static constexpr int IO2_W_CLRIND_N = 0x04;
+	static constexpr int IO2_W_DKCHRES_N = 0x08;
+	static constexpr int IO2_W_FDHDLD_N = 0x10;
+	static constexpr int IO2_W_FDMOTON_N = 0x20;
+	static constexpr int IO2_W_FDHDSEL_N = 0x40;
+
+	static constexpr int IO3_W_INTSEL_MASK = 0x03;
+	static constexpr int IO3_W_INTENBL_N = 0x04;
+	static constexpr int IO3_W_FAST = 0x10;
+	static constexpr int IO3_W_DMAR_W_N = 0x20;
+	static constexpr int IO3_W_HPB_FDC_N = 0x40;
+	static constexpr int IO3_W_DMA = 0x80;
+
+	required_device<mc6809_device> m_cpu;
+	required_device<i8291a_device> m_gpib;
+	required_device<wd2010_device> m_hdc;
+	required_device<harddisk_image_device> m_harddisk;
+	required_device<wd2793_device> m_fdc;
+	required_device<floppy_connector> m_floppy;
+	required_ioport m_hpib_addr;
+	required_ioport m_selftest;
+	required_ioport m_selftesten;
+	required_ioport m_volcfg;
+	required_ioport m_model;
+
+	TIMER_CALLBACK_MEMBER(floppy_motor_timeout);
+	TIMER_CALLBACK_MEMBER(index_timer);
+	TIMER_CALLBACK_MEMBER(fast_timer);
+
+	void i8291a_eoi_w(int state);
+	void i8291a_dav_w(int state);
+	void i8291a_nrfd_w(int state);
+	void i8291a_ndac_w(int state);
+	void i8291a_ifc_w(int state);
+	void i8291a_srq_w(int state);
+	void i8291a_atn_w(int state);
+	void i8291a_ren_w(int state);
+	void i8291a_dio_w(uint8_t data);
+	void i8291a_int_w(int state);
+	void i8291a_drq_w(int state);
+	void fdc_intrq_w(int state);
+	void fdc_drq_w(int state);
+	void hdc_reg_w(offs_t offset, uint8_t data);
+	uint8_t hdc_reg_r(offs_t offset);
+	uint8_t i8291a_dio_r();
+	void io1_w(uint8_t data);
+	void io2_w(uint8_t data);
+	void io3_w(uint8_t data);
+	uint8_t io1_r();
+	uint8_t io2_r();
+	uint8_t io3_r();
+
+	void dma_ram_w(uint8_t data);
+	uint8_t dma_ram_r(offs_t offset);
+	void cpu_map(address_map &map);
+	void floppy_index_cb(floppy_image_device *floppy, int state);
+	void hdc_bcs_cb(int state);
+	void hdc_bcr_cb(int state);
+	void hdc_bdrq_cb(int state);
+	void hdc_intrq_cb(int state);
+	void hdc_wg_cb(int state);
+	void hdc_readwrite_sector(bool write);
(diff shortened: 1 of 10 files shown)


src/devices/cpu/m6809/m6809.h
```diff
@@ -270,6 +272,7 @@ class m6809_base_device : public cpu_device
 	bool is_ea_addressing_mode() { return m_addressing_mode == ADDRESSING_MODE_EA; }
 	uint16_t get_pending_interrupt();
 	void log_illegal();
+	devcb_write_line m_syncack_write_func;
```
cuavas, line 275: Please keep data members together.  Don’t add them in random places.

src/devices/machine/wd2010.cpp
```diff
@@ -57,7 +57,7 @@ UNIMPLEMENTED FEATURES :
 // #define CYLINDER_HIGH_MASK 0x03
 
 // --------------------------------------------------------
-#define MAX_MFM_SECTORS 17      // STANDARD MFM SECTORS/TRACK
+#define MAX_MFM_SECTORS 32      // STANDARD MFM SECTORS/TRACK
```
cuavas, line 60: Is the comment correct?  “Standard” PC 3600 RPM drives had 17 sectors of 512 bytes per track, didn’t they?

Are the drives you need to support 32 sectors of 256 bytes each?

src/devices/machine/wd2010.cpp
```diff
@@ -580,20 +581,14 @@ void wd2010_device::read_sector(uint8_t data)
...
-
 			m_out_bcs_cb(0); // deactivate BCS (!)
 
 			m_out_bcr_cb(0); // strobe BCR
 			m_out_bcr_cb(1);
 
 			// set BDRQ (NOTE: DRQ status bit 3 reflects state of BDRQ)
 			m_status |= STATUS_DRQ;
-			m_out_bdrq_cb(1);
+			m_out_bdrq_cb(0);
```
cuavas, line 591: Is this going to affect anything else using this device?

src/devices/machine/wd2010.cpp
```diff
@@ -703,7 +702,13 @@ void wd2010_device::complete_write_sector(uint8_t data)
...
 
 	// 'complete_cmd' ON THE FALLING EDGE OF _BUFFER_READY_ ( set by WRITE_SECTOR ) !
-	m_deassert_write_timer->adjust(attotime::from_usec(1), newstatus);
+	if ((m_task_file[TASK_FILE_COMMAND] & 4) && m_task_file[TASK_FILE_SECTOR_COUNT] > 1) {
+		m_task_file[TASK_FILE_SECTOR_COUNT]--;
+		m_task_file[TASK_FILE_SECTOR_NUMBER]++;
+		m_multsect_write_timer->adjust(attotime::from_msec(1), data);
+	} else {
+		m_deassert_write_timer->adjust(attotime::from_usec(1), newstatus);
+	}
```
cuavas, line 711: Please use Allman formatting like the existing code in this file.

src/devices/machine/wd2010.cpp
```diff
@@ -896,8 +912,13 @@ TIMER_CALLBACK_MEMBER(wd2010_device::deassert_read)
...
 
-		m_out_bdrq_cb(0);
-		complete_immediate(param);
+		if ((m_task_file[TASK_FILE_COMMAND] & 4) && m_task_file[TASK_FILE_SECTOR_COUNT] > 1) {
+			m_task_file[TASK_FILE_SECTOR_COUNT]--;
+			m_task_file[TASK_FILE_SECTOR_NUMBER]++;
+			read_sector(m_task_file[TASK_FILE_COMMAND]);
+		} else {
+			complete_immediate(param);
+		}
```
cuavas, line 921: Please use Allman formatting like the existing code in this file.

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+	void cpu_map(address_map &map);
+	void floppy_index_cb(floppy_image_device *floppy, int state);
+	void hdc_bcs_cb(int state);
+	void hdc_bcr_cb(int state);
+	void hdc_bdrq_cb(int state);
+	void hdc_intrq_cb(int state);
+	void hdc_wg_cb(int state);
+	void hdc_readwrite_sector(bool write);
+	void update_intsel(void);
+	void cpu_sync_ack(int state);
```
cuavas, line 136: Convention in MAME is not to put the `void` in the parentheses for functions with no arguments (use empty parentheses).

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+			dma_ram_w(data);
+		}
+		m_cpu->set_input_line(M6809_IRQ_LINE, CLEAR_LINE);
+		LOG("%s: sync data: %02x\n", __func__, data);
+		if (m_dma_addr > 256) {
+			LOG("DMA: all bytes transferred\n");
+			m_dmaenable = false;
+		}
+		break;
+	}
```
cuavas, line 464: You’ll only see 1/0 values in something attached to a `devcb_write_line` – you can just use `if (state) { ...`.

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+	uint8_t ret;
+
+	if (machine().side_effects_disabled())
+		return 0;
+	if (m_dma_addr >= sizeof(m_dma_ram))
+		return 0;
+	ret = m_dma_ram[m_dma_addr++];
+	LOG("%s: %03x = %02x\n", __func__, m_dma_addr - 1, ret);
+	return ret;
+}
```
cuavas, line 689: You still want to return the appropriate value here for the debugger – you only want to suppress the address increment.  Something like:
```
	if (m_dma_addr >= sizeof(m_dma_ram))
		return 0;
	uint8_tconst ret = m_dma_ram[m_dma_addr];
	{
		LOG("%s: %03x = %02x\n", __func__, m_dma_addr, ret);
		m_dma_addr++;
	}
	return ret;
```

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+		ret |= IO1_R_HDIND;
+	if (m_hdc_intrq)
+		ret |= IO1_R_HDINT;
+	if (m_hdc_bdrq)
+		ret |= IO1_R_BDRQ;
+	if (m_hdc_bcs)
+		ret |= IO1_R_BCS_N;
+	LOG("%s: %02x\n", __func__, ret);
+	return ret;
+}
```
cuavas, line 623: It’s generally polite to suppress log messages when side effects are disabled to avoid spamming the log when a debugger window is open.

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+	} else {
+		ret = m_hpib_addr->read() ^ 0xf;
+	}
+	if (m_fdc_irq)
+		ret |= IO2_R_FDCINT;
+	if (!m_floppy->get_device()->dskchg_r())
+		ret |= IO2_R_FDKDCH_N;
+	LOG("%s: %02x\n", __func__, ret);
+	return ret;
+}
```
cuavas, line 647: You still want to return the expected value here when side effects are disabled.  If any of the things it’s calling aren’t suppressing side effects, they need to be fixed.

src/devices/bus/ieee488/hp9133.cpp
```diff
@@ -0,0 +1,878 @@
...
+		break;
+	case 7:
+		m_hdc_cmd = data;
+		break;
+	}
+}
+
+void hp9133_device::cpu_map(address_map &map)
+{
+	map(0x0000, 0x0007).m("i8291a", FUNC(i8291a_device::map));
```
cuavas, line 813: You should be able to use `m_gpib` here rather than the literal tag.

src/devices/machine/wd2010.cpp
svenschnelle: Yes, HP used 32 sectors with 256 bytes each. The comment is indeed not correct anymore as this is not "Standard". However, as MAX_MFM_SECTORS is the maximum number of sector supported, i wonder whether the comment should be just removed. It should be changed to at least something along "Maximum number of sectors this driver supports", but that would be documenting the obvious.

src/devices/machine/wd2010.cpp
svenschnelle: The only other wd2010 user i found is the DEC rainbow. I couldn't test this as i have no images and didn't found any image online.

src/devices/machine/wd2010.cpp
cuavas: I’m fine with just removing the comment – the name is fairly obvious.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Indeed.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Ok.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Ok.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Ok.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Ok, i found the switch easier to read than a large if/else, but doesn't really matter.

src/devices/cpu/m6809/m6809.h
svenschnelle: Ok.

src/devices/machine/wd2010.cpp
svenschnelle: Ok.

src/devices/machine/wd2010.cpp
svenschnelle: Ok.

src/devices/bus/ieee488/hp9133.cpp
svenschnelle: Ok.

src/devices/machine/wd2010.cpp
svenschnelle: After finding all the various disk images required for DEC rainbow, i tried it, and it looks like DEC rainbow is no longer working with the wd2010 changes. So better drop this PR.

cuavas:
What was the reason for closing this?

cuavas:
> After finding all the various disk images required for DEC rainbow, i tried it, and it looks like DEC rainbow is no longer working with the wd2010 changes. So better drop this PR.

Sorry, I missed this comment.

Does this PR actually cause the Rainbow to regress?  It has always required some effort to get it to boot as the hardware isn’t understood properly.  If it’s no more broken that it was before, that’s no big deal.

If the WD2010 do cause the Rainbow to regress, can we get the other parts of the PR isolated so the HP9133 card can be added without hard disk support until the WD2010 can be updated to pacify both it and the Rainbow?

svenschnelle:
> > After finding all the various disk images required for DEC rainbow, i tried it, and it looks like DEC rainbow is no longer working with the wd2010 changes. So better drop this PR.
> 
> Sorry, I missed this comment.
> 
> Does this PR actually cause the Rainbow to regress? It has always required some effort to get it to boot as the hardware isn’t understood properly. If it’s no more broken that it was before, that’s no big deal.

Well, without my changes i was at least able to low-level format the winchester disk in the rainbow, partition it and copy files to it. I wasn't able to from the winchester disk, but that maybe just user error. The DEC rainbow is quite different compared to normal XT's/PC's. 

With my changes i'm not even able to LL format the disk - it immediately reports bad sectors on each track. I'm assuming the read/write sector callbacks are now called when they are not expected to get called.

> If the WD2010 do cause the Rainbow to regress, can we get the other parts of the PR isolated so the HP9133 card can be added without hard disk support until the WD2010 can be updated to pacify both it and the Rainbow?

I think without a working WD2010/HDD the 9133 enter selftest fail mode, and you cannot even use the floppy drive in it. I'm away over the weekend, but let me look into the rainbow problem next week, and see whether i can make it work. I'll reopen the PR if i find a way to make both (HP9133 + rainbow) work.

svenschnelle:
I implemented multi sector read/writes now differently and verified that the DEC rainbow is still working with the changes.

PR #12563, opened 2024-07-10
igs/igs_fear.cpp: made game playable without sound [XingXing]

HappyASR:
- added timer and interrupt
- added rtc
- add input and dipswitch
- add fipx in sprite drawing
- correct game version of superkids

### DIFF
--- a/src/mame/igs/igs_fear.cpp
+++ b/src/mame/igs/igs_fear.cpp
@@ -44,16 +44,6 @@ protected:
 	virtual void video_start() override;
 
 private:
-	required_device<cpu_device> m_maincpu;
-	required_device<xa_cpu_device> m_xa;
-	required_shared_ptr<uint32_t> m_videoram;
-	required_device<palette_device> m_palette;
-	required_region_ptr<uint8_t> m_gfxrom;
-
-	required_device<ticket_dispenser_device> m_ticket;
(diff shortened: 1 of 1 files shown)


src/mame/igs/igs_fear.cpp
```diff
@@ -45,42 +50,75 @@ class igs_fear_state : public driver_device
...
 
 	void sound_irq(int state);
 	void vblank_irq(int state);
 
-	void draw_sprite(bitmap_ind16 &bitmap, const rectangle &cliprect, int xpos, int ypos, int height, int width, int palette, int romoffset);
+	void draw_sprite(bitmap_ind16 &bitmap, const rectangle &cliprect, int xpos, int ypos, int height, int width, int palette, int flipx, int romoffset);
 
 	uint32_t screen_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect);
+
+	uint32_t m_gpio_o;
```
rb6502, line 66: We prefer to keep function signatures and variables separate.  I usually do all the function stuff and then all the variables, but the other way is fine too if that works better for you.  I used to write stuff all mixed together like this and it turns out it wasn't great for maintaining it later.

rb6502:
This looks quite good aside from the one thing I pointed out.  Get that rearranged and it should be good to go.

PR #12567, opened 2024-07-11
dectalk: fixed distortion in DTC07 DAC emulation

datajake1999:


datajake1999:
I just implemented a hack to fix the broken DSP output.

cracyc:
The only oscillator on the board is 20mhz.  I'm okay with a hack since there are already other in it but there should be a comment that it's wrong.

datajake1999:
Done.

cracyc:
The board dac is an ad7541 https://www.analog.com/media/en/technical-documentation/data-sheets/AD7541A.pdf .

datajake1999:
I just reduced the DAC bit depth back down to 12, as it sounds more accurate when compared to actual hardware (most notably when the synth occasionally generates faint electronic beeps).

datajake1999:
I just updated the comment that says which DAC the hardware uses, as it is now known.

cuavas:
It’s definitely an unrealistic frequency – the TMS320C15 was rated for 25 MHz maximum.

PR #12570, opened 2024-07-12
casio/pv1000.cpp: Three tiny changes

lidnariq:
* ioFF has a toggle that just draws border color instead of tiles, so that the CPU runs faster.
* The pixel clock is slow, so the pixel aspect is noticeably wide - about 4:3
* ioFF supports drawing tile data from all eight 8KiB windows out of the 64KiB address space; add support for the easy bottom four

### DIFF
--- a/src/mame/casio/pv1000.cpp
+++ b/src/mame/casio/pv1000.cpp
@@ -485,6 +485,13 @@ void pv1000_state::pv1000(machine_config &config)
 	/* D65010G031 - Video & sound chip */
 	SCREEN(config, m_screen, SCREEN_TYPE_RASTER);
 	m_screen->set_raw(17897725/4, 288, 0, 224, 262, 0, 244);
+	// Pixel aspect is 48/35.
+	// Display aspect is MAME's 4:3 default.
+
+	// Note that this value is overridden by the user's pv1000.cfg, if present.
+	// 206px x 48/35(PAR) / 4/3(DAR) = 212sl
+	m_screen->set_default_position(216/206.0,0, //216 px in storage aspect; cropped to 206 px
+	                               244/212.0,0); //244 sl in storage aspect; cropped to 212 sl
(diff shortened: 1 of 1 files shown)


rb6502:
I don't think the aspect part of this can be right.  The real PV-1000 was used on 4:3 NTSC TVs, so the final output image should be 4:3 and you shouldn't care about the individual pixel shape.  We don't for extreme cases like 640x240 on the PlayStation and it works out fine.

lidnariq:
What's the desired behavior for things that are extremely overscanned?

The PV-1000 uses a pixel clock that is NTSC x 5/4, 4.47MHz. Per rec.701 for digitizing SDTV, that works out to an exact PAR of 48:35. The PV-1000 generates a raw image that is 224 pixels by 244 scanlines, which when scaled by that 48:35 works out to roughly 307 x 244 square pixels. The TV then crops that down to somewhere around 290 x 215 square pixels: video of real hardware show the left-most and right-most tiles in the tilemap are often partially lost to overscan.

For comparison, a real NTSC NES generates a 284 x 242 image, much of which is also border. Early emulators pretended it's 256x240, but real TVs show somewhere around 250 x 215 ish.

cuavas:
I think all you can do is set the default position/scaling so that it will be cropped the way it would be on a TV – see `screen_device::set_default_position`.

src/mame/casio/pv1000.cpp
```diff
@@ -262,8 +263,9 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
 //  case 0x06 VRAM + PCG location, always fixed at 0xb8xx
 	case 0x07:
 		/* ---- -xxx unknown, border color? */
-		m_pcg_bank = (data & 0x20) >> 5;
+		m_pcg_bank = (data & 0xE0) >> 5;
```
angelosa, line 266: For consistency in the driver hex values are in lowercase.

angelosa (approved):

lidnariq:
is this now ok?

PR #12572, opened 2024-07-13
spg2xx: Fix behavior of audio channel enable/stop flags

sp1187:
Summary of new behavior:
* Automatic stops sets channel stop flag while keeping channel enable flag unchanged (unlike current emulation where it's cleared). 
* Channel is only active (channel status flag true) when enable flag is true and stop flag is false.

This was verified in real hardware by running audio test roms on V.Smile. From what testing I have done so far, the changes don't really affect the games much if at all, but it should be technically more accurate.

### DIFF
--- a/src/devices/machine/spg2xx_audio.cpp
+++ b/src/devices/machine/spg2xx_audio.cpp
@@ -449,7 +449,9 @@ void spg2xx_audio_device::audio_ctrl_w(offs_t offset, uint16_t data)
 	case AUDIO_CHANNEL_ENABLE:
 	{
 		LOGMASKED(LOG_SPU_WRITES, "audio_ctrl_w: Channel Enable: %04x\n", data);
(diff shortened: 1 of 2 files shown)


MooglyGuy:
All the changes look good to me. I remember basing the start/stop behavior off of how it was (poorly) described in a manual, and I filled in some gaps with guesswork.

With how many games and systems used the SPG2xx, it's probably only a matter of time before something would have come up that this fixes. Thanks for making the SPG emulation better!

src/devices/machine/spg2xx_audio.cpp
```diff
@@ -981,9 +982,33 @@ void spg2xx_audio_device::sound_stream_update(sound_stream &stream, std::vector<
 	}
 }
 
+inline void spg2xx_audio_device::start_channel(const uint32_t channel) {
```
cuavas, line 985: Don’t forget to move the opening brace to a new line.

PR #12574, opened 2024-07-14
Tek4404 work_in_progress.

Elektraglide:
This is a WIP getting Tek4404 to run.
It now passes all the h/w tests apart from mouse. ( RAM, MMU, Serial, Timers, Sound, Video, Storage Devices)
A boot manages to load the OS from "system.boot", starts up and tries to write some swap data.  After writing 512 bytes, it never receives a STATUS ACK and just spins forever running the idle job and checking whether the write has completed.

I do not have the knowledge around scsi and new_scsi_bus to understand what the emulated devcies are not doing that Tek4404 expects, so this is where I have got to.


rb6502:
Can you fix the conflicts with the already-merged ncr5385 changes?

src/devices/machine/ncr5385.cpp
```diff
@@ -245,6 +246,9 @@ u8 ncr5385_device::aux_status_r()
 
 	if (!m_int_status)
 	{
+		// AB mask out any bits
+		data &= ~(AUX_STATUS_MSG | AUX_STATUS_CD | AUX_STATUS_IO);
```
pmackinlay, line 250: I don't think this should be necessary; resetting and latching of these bits should be properly controlled by changes to the interrupt status. Did you find a bug in the logic?

src/devices/machine/ncr5385.cpp
```diff
@@ -759,9 +767,29 @@ int ncr5385_device::state_step()
 			else
 				m_sbx = false;
 
+#if 0
 			// clear data and ACK
 			scsi_bus->data_w(scsi_refid, 0);
 			scsi_bus->ctrl_w(scsi_refid, 0, S_ACK);
+#else
+			// clear ACK except after last byte of message output phase
```
pmackinlay, line 775: I don't think this is correct; this logic only applies for the message input phase, to allow the initiator the chance to raise ATN to reject a message from the target if necessary. The same thing does not apply during message output.

pmackinlay (commented):
There are a few stray blank lines, some possible accidental whitespace changes, and some instances of inconsistent formatting that should be cleaned up prior to merging, but since this is a WIP asking for assistance, I haven't tried to nitpick those here.

src/mame/tektronix/tek440x.cpp
```diff
@@ -57,17 +58,295 @@
...
+	if (((address>>12) & 0x7ff) < 20)
+		LOG("PTEwrite: %08x  <= %04x paddr(%08x) PID(%d) dirty(%d) write_enable(%d)\n",
+		(address>>12) & 0x7ff, data,
+		(BIT(data, 0, 12)<<12), BIT(data, 11, 3), data & 0x8000 ? 1 : 0, data & 0x4000 ? 1 : 0);
+
+	if (BIT(*m_map_control, MAP_SYS_WR_ENABLE))
+	{
+		m_map[(address>>12) & 0x7ff] = data;
+	}
+}
```
cuavas, line 212: Indentation is off here:
* These functions are at the left margin while the other inline member function bodies are indented.
* You should use one level of indentation per nested scope level.
* Please use braces for any multi-line `if`/`else` body even if it’s technically only a single statement.
* Please indent statement continuations by two tabs.

src/devices/machine/am9513.cpp
cuavas, line 1: I already explained on one of your previous pull requests why these logging changes are undesirable.

PR #12575, opened 2024-07-14
speglsht.cpp: Add notes

cam900:
Transparent color is wrong at title screen animation, reference: https://youtu.be/cK7mGnt6YfY?si=Ah5hC_iFS2Li7Xwl


angelosa:
There are three bugs here by just looking the initial attract sequence:
1. bird's eye view should be circular (sprite?) masked

2. example shot shows a grid that is missing

3. title screen effectively don't blend the background emblem properly


I feel like this should be turned into a task based issue, 2 in particular looks gameplay breaking and imo worth for a demotion.

mamehaze:
the ball isn't even visible in a lot of situations either.

I kinda agree the issues here are big enough for this to simply be MNW, you wouldn't ship a commercial product with this quality of emulation.

PR #12576, opened 2024-07-14
rm/rm380z.cpp: Split RM480Z out to its own driver with MQ2 floppy drive support

RobinSergeant:
This PR adds a new driver to replace the non working 480Z fragments in the 380Z driver.  As the hardware is so different I think it really belongs in a separate driver.  Everything from the keyboard interface to video RAM access is different and the 480Z is generally much more complex.

I have emulated all the basic features, plus HRG and floppy drives.  There are still more things to add in the future (such as cassette support), but I think it's in a good state for release now.  Everything I've tested seems to work including the CP/M ram disk on drive M: (which is a good test of the bank switching).

It largely uses the existing ROM files, but I have added a new BIOS version (ROS 1.2D) for completeness.

Some screenshots (welcome disk title screen, Trains game, CP/M boot screen, and ROS/BIR):

Screenshot from 2024-07-14 10-58-37

Screenshot from 2024-07-14 11-00-30

Screenshot from 2024-07-14 10-53-03

Screenshot from 2024-07-14 11-02-53


src/mame/rm/rm_mq2.cpp
```diff
@@ -135,9 +136,13 @@ void rmMQ2_device::input_rts(int state)
 {
 	if (started())
 	{
-	m_sio->dcdb_w(state);
-	// NMI generated to reset baud rate to 9600
-	m_maincpu->set_input_line(INPUT_LINE_NMI, state ? ASSERT_LINE : CLEAR_LINE);
+		m_sio->dcdb_w(state);
+		// give firmware ~50ms to finish initialisation before allowing first NMI
+		if (m_maincpu->total_cycles() > 200'000)
```
rb6502, line 164: On hardware, is there some kind of handshake or something that makes this work, or is it more of a magical timing thing?

RobinSergeant: It seems to be a magical timing thing, but I suppose with real hardware it would very unusual to start both CPUs at exactly the same time - there would naturally be at least a few seconds gap between pressing one power switch and then the other!

RobinSergeant:
Conflicts with recent ATTR_COLD changes resolved!  Could somebody please merge this pull request if there are no further comments?   I've tested the driver extensively and have even ported CP/M Plus to the 480Z using it (see https://github.com/RobinSergeant/cpmplus480z) :-)

As there are no other 480Z emulators out there I think it would be nice to get this into MAME so that more people can use the machine.

PR #12578, opened 2024-07-14
atari/asteroid.cpp: Cleaned up the driver and split the audio hardware into devices.

jackson2k2:
- Reduced runtime tag lookups by using finders.
- Mostly consolidated driver into a single file.
- Split Lunar Lander, Asteroids, and Asteroids Deluxe into their own class.
- Moved the Lunar Lander audio circuit to asteroids_a.cpp instead of in the driver.
- Reworked the memory maps and machine configurations. 'asterock' and 'asteroidb' no longer use their own init.
- Reduced duplicated code by using templated functions.
- Cleaned up function names.
- Fixed spacing and formatting.

No functional change intended, and I'm taking cues from Vas touching my work. If there are any accidents, let me know.
Splitting the audio hardware will make way for a Lunar Lander netlist in the future and hopefully Asteroids will follow after. _(Lunar Lander and Asteroids are already below 50% on a Core 2 Duo with MSVC building as Debug, the discrete code should be eventually legacy.)_


happppp:
Why scatter the notes that were at the top of the driver file? If the part of the driver notes are about game xyz, you don't have to move them to the rom defs of game xyz.

PR #12579, opened 2024-07-15
Add an option to enable auto-pausing when tab menu is popped up in the game.

agugu2000:


angelosa:
I don't see any ill effect about this, and actually can be useful in some scenarios. @cuavas ?

happppp:
- Remove "game" from the option name and description, a better option name is perhaps "autopause".
- We already have auto-pause on F4 (the palette/tile viewer). If this option is added, it might as well be enabled for that.
- What happens if you manually pause -> enter menu -> exit menu?

angelosa:
And yeah, for something running at very slow speeds (say, xbox) bringing auto pause for tab menu can help.

cuavas:
I know it’s over a year late, but 3dfb72cb6245da3cbdcafc02afd42315512d1557 implements this properly in the UI manager, and d9b6e5f41af8092cbc7b733db4cff5b04a428e2d patches up some fallout from the rest of the change.  The option is in the “Miscellaneous Options” menu, also accessible by editing ui.ini.

PR #12580, opened 2024-07-15
spg2xx: Auto-increment wave address register when channel is playing

sp1187:
Also discovered from testing in real hardware (V.Smile). Probably not likely to noticeably affect games, but should still be technically more accurate.

### DIFF
--- a/src/devices/machine/spg2xx_audio.cpp
+++ b/src/devices/machine/spg2xx_audio.cpp
@@ -89,7 +89,6 @@ void spg2xx_audio_device::device_start()
 
 	save_item(NAME(m_sample_shift));
 	save_item(NAME(m_sample_count));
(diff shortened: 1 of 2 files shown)


sp1187:
Somehow this seems to have solved the Tiger Rescue bug mentioned in #12525. @mamehaze Could you check again as you seem to be more familiar with this game?

mamehaze:
I'll take a look when I'm at my PC

mamehaze:
it definitely changes the behavior to be less obnoxious at least, however I think it might just be because it's chopping off samplse really early now.

note the main game select menu sounds in the lexizeus set, they're broken with these changes, they don't play out in full when you move the cursor, you simply get a random fraction of the star of the sound, so I'm not convinced this changeset is making things better overall

PR #12581, opened 2024-07-15
microsoft/xbox.cpp: hookup HDD SW list

angelosa:
OEM Dashboard [archive.org]

### DIFF
--- a/hash/xbox_hdd.xml
+++ b/hash/xbox_hdd.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="xbox_hdd" description="XBOX hard disk images">
+    <software name="oemdash" supported="partial">
+        <!-- TODO: add version -->
+		<description>OEM Dashboard</description>
+		<year>2001</year>
+		<publisher>Microsoft</publisher>
+    	<notes><![CDATA[
+Original Dashboard
+Converted to .raw thru qemu-img then back into .chd format.
+At the time of this writing chdman v5 fumbles with -chs (tries to convert it to 32k cylinders) assume standard Seagate -chs 16383,16,63 was used instead, cfr. https://xboxdevwiki.net/Xbox_Hard_Disk_Technical_Details
+Expects to be writable straight off the bat
+-debug has custom options, particularly the "xbox nv2a_combiners", cfr. "xbox help" for details.
+]]></notes>
+		<part name="hdd" interface="ide_hdd">
+			<diskarea name="harddriv">
+				<disk name="xbox_hdd" sha1="0408bc71a66e633856c622efe7d2562c26e99886" writeable="yes"/>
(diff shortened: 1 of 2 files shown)


angelosa:
https://mametesters.org/view.php?id=8617 should also be marked as fixed, roughly either 
https://github.com/mamedev/mame/commit/23ab5bba61e21514d21ca83b599aba56b0055b2b or
https://github.com/mamedev/mame/commit/71b9d2f072f6c4d7406030590a3edc5ae24c3216

angelosa:
Drafting, need to checkout the full installation loop from recovery disks.

mamehaze:
shouldn't the machine have an 'as you would have got it from the shop' HDD by default, rather than a softlisted one?

angelosa:
That would be impractical, there are regular dashboard images, dev dashboard images, a TBA number of unofficial dashboards, Linux, ReactOS ...

mamehaze:
It was literally a retail product though, at the very least mame needs a
driver representative of that.

mamehaze:
Having a pc style one is fine too, but that wasn't the primary use case or
intended market for a games console

angelosa:
> It was literally a retail product though, at the very least mame needs a driver representative of that.

There's likely a method for defaulting a SW list to a given system, which I'm not gonna dig in the current state of the driver, won't load any DVD disc assuming you even reach that (for missing xbox page select at very least), and won't enforce for **practical** reasons.

rb6502:
Haze is correct in that if the driver ever becomes actually working it should have a default non-softlist HDD, but then you also need to split the driver by region.  And I'm not sure you can have a device with a pre-defined image get overridden by a software list right now. But since it's a pretty long distance from working right now that's kind of irrelevant.  The software list will make it easier for people to work on the driver if they want, which is good.

yz70s:
> https://mametesters.org/view.php?id=8617 should also be marked as fixed, roughly either [23ab5bb](https://github.com/mamedev/mame/commit/23ab5bba61e21514d21ca83b599aba56b0055b2b) or [71b9d2f](https://github.com/mamedev/mame/commit/71b9d2f072f6c4d7406030590a3edc5ae24c3216)

[23ab5bb](https://github.com/mamedev/mame/commit/23ab5bba61e21514d21ca83b599aba56b0055b2b) is the one at which the problem is really solved, the other one does not give any visible effects.

Gurudumps:
Yes a default factory xbox Seagate 10GB HDD needs to be part of the default roms loaded at boot time. The solution to the missing HDD is to make a default HDD image (for each region xbox if needed) that loads with the roms and as per other arcade games simply make it use the newest version dash available. 
Latest dash version seems to be 5960 (https://archive.org/details/XboxDashboard5960)
The older versions can be provided elsewhere but for xbox there's no need to have older versions as the latest was likely to be the best one to have, unless there is documented proof that a specific version was the most compatible.

angelosa:
This will probably be merged as-is, the "we need a default option" in driver is irrelevant with current driver state (i.e. it's food for an actual "allow softlisted default in drivers" issue ticket).

PR #12585, opened 2024-07-16
upd765: Drive busy bits should not be cleared until a SIS is executed

brijohn:
The upd765 does not clear the drive busy bits until you issue a sense interrupt status command. Without this fix CP/M Plus on the QX-10 gives disk I/O errors because it expects the busy bit to stay set until it has a chance to issue the SIS command.

I have verified this is the correct behavior on a real QX-10, which uses a NEC D765AC as its floppy controller. Regarding the cpc6128, this fix does not seem to cause an issues with accessing a disk, I can run cat from basic and get a directory as well load files from the disk. Same thing with the specpl3e, loading programs from disk appears to work fine.

### DIFF
--- a/src/devices/machine/upd765.cpp
+++ b/src/devices/machine/upd765.cpp
@@ -219,6 +219,7 @@ void ps2_fdc_device::set_mode(mode_t _mode)
 void upd765_family_device::device_start()
 {
 	save_item(NAME(selected_drive));
+	save_item(NAME(drive_busy));
 
 	for(int i=0; i != 4; i++) {
 		char name[2];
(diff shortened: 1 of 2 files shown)


angelosa:
This needs to be tested with pc88 & pc98 line of computers, which are known to be picky to changes in 765 core.

rb6502:
He said the behavior is verified on hardware, so I wouldn't expect it to cause problems with pc88/pc98.  But I would also like to see a few more systems tested before applying it.

cuavas:
Yeah, code looks OK in principle, but let’s wait for some more testing.

brijohn:
Ok so I found a minor issue with the PC98 computers, its easily fixed by moving the following code back to be i82072 specific. This was preventing the pc98 from handling multiple SIS commands and causing boot failure. I'm not 100% sure it makes sense for the i82072 controller either, but since it was there originally I'm leaving it there. With this fix QX-10, cpc6128,specpl3e,and pc98 computers seem to work correctly. I tried the pc88 line as well, but i could not get it booting the d88 disk images i tried even without this patch installed.
```
       for(floppy_info &fi : flopi) {
                fi.st0_filled = false;
       }
```

brijohn:
Ok so i found a combination of working disk image and oc88 machine that i could verify worked before this patch and it still loads correctly afterwards as well. So the current state is that this fixes CP/M plus on the QX-10, and the cpc6128, specpl3e, pc88, and pc98 still load disks afterwards as well.

PR #12586, opened 2024-07-16
remove u8 at fullname

flama12333:
shouldn't be there at fullname


flama12333:
closed down as outdated

PR #12587, opened 2024-07-17
dectalk: set sample rate of DAC output stream to 10 kHz

datajake1999:
On the actual hardware, The DSP generates audio at a 10 kHz sample rate, so it makes sense to run the audio stream at this rate.

### DIFF
--- a/src/devices/bus/isa/dectalk.cpp
+++ b/src/devices/bus/isa/dectalk.cpp
@@ -226,4 +226,5 @@ void dectalk_isa_device::device_reset()
 	m_ctl = 0;
 	m_vol = 63;
 	m_bio = ASSERT_LINE;
+	m_dac->set_sample_rate(10000); // the DSP generates audio with a 10 kHz sample rate
 }
--- a/src/devices/sound/dac.h
+++ b/src/devices/sound/dac.h
@@ -108,6 +108,12 @@ public:
 	}
 	dac_device_base &set_output_range(stream_buffer::sample_t vref) { return set_output_range(-vref, vref); }
 
+	// set the sample rate
+	void set_sample_rate(u32 rate)
+	{
+		m_stream->set_sample_rate(rate);
+	}
+
 private:
 	// internal state
 	sound_stream *m_stream;


happppp:
A DAC doesn't have a sample rate. The reason mame sets up one here anyway: ```m_stream = stream_alloc(inputs, 1, 48000 * 4);``` is because MAME sound.cpp resampler doesn't support a variable sample rate.

Is this PR a workaround for audio glitches? I'd think 10,000 -> 192,000 -> [MAME's output sample rate] should work fine?

happppp:
Even though I don't know if the reason for this PR was for resampler quality, overall DAC quality should be improved here: https://github.com/mamedev/mame/commit/2695a185f97c29cc00e67f2b95debed676413264

datajake1999:
I understand a DAC doesn't have a sample rate. The main motivation for this PR is the poor resampling quality. Even if the resampler is improved, running the DAC emulation at 192 kHz doesn't make sense for devices that have a known low output sample rate, as generating unnecessary samples causes aliasing. In other words, devices that implement analog anti aliasing filters such as the DTC07 should run their audio stream at a low sample rate if it is known, then be resampled to the mixer sample rate with a decent algorithm. This PR partially addresses this problem.

happppp:
It's guaranteed that no matter which software runs on it, that it updates the DAC at exactly 10kHz? Even if so, I'm not sure this setter is a good idea.
Better to improve the DAC resampling in general (what I did), at the cost of some performance. Unfortunately, worse if a MAME driver has many DACs such as offroad. But accuracy over performance I'd say.

datajake1999:
The DTC07 has a rarely used mode that allows playing digital audio, and the sample rate can be changed in this mode. I am all for accuracy as well, however I also want to make the experience bearable for people who will be listening to this for extended periods of time (myself included). Scenarios where this is the case include using a DOS screen reader, and reading long passages of text. I have shown the recent DTC07 emulation to several other people, and they all said they don't like the aliasing in the output. I will build your changes and give them a shot. I am totally open for another way of solving this problem without sacrificing accuracy.

datajake1999:
I just tested this latest code, and this is working great! It has the intended effect of my pull request without the drawbacks. As such, I am closing with this comment as this PR is now irrelevant.

PR #12588, opened 2024-07-17
rs232: Add Mouse Systems serial mouse to default rs232 list.

brijohn:
This adds the support for using the Mouse Systems serial mouse in the QX-10 emulation. Valdocs supported this mouse when using programs like valpaint and valdraw.

### DIFF
--- a/src/devices/bus/rs232/rs232.cpp
+++ b/src/devices/bus/rs232/rs232.cpp
@@ -155,6 +155,7 @@ device_rs232_port_interface::~device_rs232_port_interface()
 
 
 #include "ie15.h"
+#include "hlemouse.h"
 #include "keyboard.h"
 #include "loopback.h"
 #include "mboardd.h"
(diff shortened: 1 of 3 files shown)


cuavas:
Is there a good reason this mouse isn’t included in the default RS-232 devices?  Mouse Systems protocol was pretty widely supported.

brijohn:
> Is there a good reason this mouse isn’t included in the default RS-232 devices? Mouse Systems protocol was pretty widely supported.

I'm not sure, I based this off the DEC rainbow code. The other systems that do currently support the Mouse Systems protocol aren't using the the default rs232 device list and instead provide their own list of supported serial devices.

cuavas:
To be honest, I think Mouse Systems-compatible mouses were widespread enough to add it to the “standard” RS-232 peripherals.  Even if a system doesn’t have software that uses it, someone might want to develop homebrew in MAME, right?

rb6502:
Agreed 100% with just making the Mouse Systems mouse a standard serial peripheral.

brijohn:
Ok updated this to include mouse systems mouse as part of the default RS232 list of devices.

PR #12053, opened 2024-02-21
Add front panel keypad to Heathkit H8. Added support for serial console.

jefftranter:
Adds a graphical front panel keypad to the Heathkit H8.

PR #12590, opened 2024-07-17
Heath/Zenith H8: Add Front Panel Keypad and Support for Serial Console

jefftranter:
Add front panel keypad and support for serial console to Heathkit H8.

### DIFF
--- a/src/mame/heathzenith/h8.cpp
+++ b/src/mame/heathzenith/h8.cpp
@@ -10,6 +10,8 @@
 
 STATUS:
         It runs, keyboard works, you can enter data.
+        Serial console works. You can make it visible by setting Video
+        Options in settings.
 
 Meaning of LEDs:
         PWR = power is turned on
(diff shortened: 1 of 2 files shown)


rb6502:
The actual code looks fine to me.  @cuavas is there a better/less verbose way to accomplish what the layout file is doing?

cuavas:
This has the same issues as #12053 and more.

cuavas:
Also, given the UART is optional, shouldn’t there be an option to run without it?

mgarlanger:
The optional H-8-5 board was the first serial board Heath offered. It contains two 8251 UARTs, one for the cassette port and one for the console port. Since the existing code doesn't have the cassette 8251 as optional, this change doesn't really make the H8 code any worse. Later, Heath offered the H-8-4 serial board, which provided four 8250 UARTs.  

Ideally, the H8 would have slots for its bus, and both of these serial boards would be implemented as virtual cards which could then be selected by the user.

Actually, the H8 bus is a passive bus, and everything including the CPU board should be on a card (and selectable) since Heath (and other companies) offered Z80 replacement CPU cards.

I've been working with @rb6502 to add slots to the H89, and once that is working, I planned to see what it would take to get slots added to the H8.

PR #12594, opened 2024-07-18
machine/at28c64b.cpp: Fix read overflow initializing nvram

ksherlock:
actual rom size may be less than AT28C64B_DATA_BYTES. a2bus/booti,  as a motivating example, has a ROM size of 0x2000.

MT 08524


cuavas:
Should this be checking that it doesn’t read more than `AT28C64B_DATA_BYTES` as well as checking that it doesn’t run of the end of the region?

PR #12599, opened 2024-07-20
hp/hp98x6.cpp: Support 3 new machines

fulivi:
Hi,
this PR adds the emulation of hp9826a, 9836a & 9836c machines (of the hp98x6 family).
I'll e-mail the relevant ROM images shortly.
Thanks.
--F.Ulivi


MooglyGuy:
Had a quick scroll through it. Looks pretty good to me, just a few feedback points that popped into my head, but I don't feel strongly enough about them that devs with commit access should hold things up about it:
- Consider turning verbose logging off. Helps with performance.
- Some of the multi-line pieces of code that have been added seem like they're being word-wrapped strangely early. Line 440 of hp98x6_upi.cpp is a good example, the word-wrap point isn't even close to the midpoint of my monitor. That said, I don't know what your own editor setup is like, so I don't feel strongly about it.
- I notice that the ID PROM is an optional device, but you now have a class structure where there is a base class. Would it be feasible to make it a required device, and have it in only the derived classes that need it?

src/mame/hp/hp98x6.cpp
```diff
@@ -73,13 +109,25 @@ template<typename T> void BIT_SET(T& w , unsigned n)
...
+//         ↑                ↑                  ↑
+// +--------------+  +--------------+  +---------------+
+// | hp9826_state |  | hp9836_state |  | hp9836c_state |
+// +--------------+  +--------------+  +---------------+
+class hp98x6_base_state : public driver_device
 {
 public:
-	hp9816_state(const machine_config &mconfig, device_type type, const char *tag)
+	hp98x6_base_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
```
cuavas, line 131: If this is now an abstract base class, the constructor should be `protected`.

src/mame/hp/hp98x6.cpp
```diff
@@ -696,6 +807,1236 @@ ROM_START(hp9816a)
...
+{
+	m_floppy_ctrl = 0;
+	m_fdc->mr_w(0);
+	floppy_update_sel();
+}
+
+void hp9826_36_state::floppy_update_sel()
+{
+	floppy_image_device *drives[2] = { get_drive(0), get_drive(1) };
+	floppy_image_device *new_floppy = drives[get_sel_floppy()];
```
cuavas, line 1019: Wouldn’t it be simpler to just do:
```c++
	floppy_image_device *const new_floppy = get_drive(get_sel_floppy());
```

src/mame/hp/hp98x6.cpp
```diff
@@ -696,6 +807,1236 @@ ROM_START(hp9816a)
...
+	if (new_floppy != m_curr_floppy) {
+		m_curr_floppy = new_floppy;
+		m_fdc->set_floppy(m_curr_floppy);
+	}
+	if (m_curr_floppy) {
+		m_curr_floppy->setup_index_pulse_cb(floppy_image_device::index_pulse_cb(&hp9826_36_state::floppy_index_cb, this));
+		m_curr_floppy->ss_w(BIT(m_floppy_ctrl, FC_SIDE_SEL_BIT));
+	}
+	floppy_update_mon_hlt();
+}
```
cuavas, line 1030: It looks like you’re going to need to implement `device_post_load()` to correct `m_curr_floppy` after loading a saved state, or it could get out of sync with the rest of the state.

src/mame/hp/hp98x6_upi.h
```diff
@@ -24,6 +24,8 @@ class hp98x6_upi_device : public device_t
 	// construction/destruction
 	hp98x6_upi_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
 
+	void set_id_prom(const uint8_t *ptr) { m_idprom = ptr; }
+
```
cuavas, line 28: Rather than doing this, it’s better to have an `optional_region_ptr` to find the IDPROM.  You can either have it start with the guaranteed not to be found dummy tag and let the system instantiating it configure the tag of the ROM region, or use a tag relative to the UPI device (e.g. initialise with `m_idprom(*this, "idprom")` and use the region tag `"upi:idprom"` in the ROM definitions for the parent system).

src/mame/hp/hp98x6.cpp
fulivi: absolutely, that was a relic from an early implementation

src/mame/hp/hp98x6_upi.h
fulivi: Ok, I think I addressed the issue you saw. Keep in mind that access to IDPROM through the UPI only happened in 9816, other systems had the PROM directly mapped on 68k bus.

fulivi: Also, to reply to @MooglyGuy, IDPROM region was marked as optional because I initially thought to replicate in emulation the behavior of real systems that could also run without ID PROMs. But in the end, ID PROM image is always present so why bother.

PR #12607, opened 2024-07-22
namco/namcos10: Add additional notes for Hajime no Ippo 2

987123879113:
The real hardware checksums were verified thanks to buffi.

MAME:
image

Real hardware: https://imgur.com/a/1q1atDB

### DIFF
--- a/src/mame/namco/namcos10.cpp
+++ b/src/mame/namco/namcos10.cpp
@@ -3044,6 +3044,7 @@ void namcos10_memio_state::machine_reset()
 void namcos10_memio_state::namcos10_memio_map_inner(address_map &map)
 {
 	map(0xf468000, 0xf468001).w(FUNC(namcos10_memio_state::nand_dataxor_w));
+	// map(0xf478000, 0xf478001).w(); // bit 0 = Triggers solenoid when set. This is used to make the punch pad stand up, triggering the speed sensors along the way
 }
 
 void namcos10_memio_state::namcos10_memio_map(address_map &map)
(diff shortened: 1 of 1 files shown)


src/mame/namco/namcos10.cpp
```diff
@@ -3457,20 +3457,29 @@ static INPUT_PORTS_START( ippo2 )
...
+	PORT_BIT( 0x00001000, IP_ACTIVE_HIGH, IPT_BUTTON4 ) PORT_NAME("Human Sensor 1")
+	PORT_BIT( 0x00002000, IP_ACTIVE_HIGH, IPT_BUTTON5 ) PORT_NAME("Human Sensor 2")
+	PORT_BIT( 0x00004000, IP_ACTIVE_HIGH, IPT_BUTTON6 ) PORT_NAME("Human Sensor 3")
+
+	PORT_BIT( 0x00008000, IP_ACTIVE_LOW, IPT_SELECT )
+
+	PORT_BIT( 0x08000000, IP_ACTIVE_LOW, IPT_TILT )
+
+	PORT_BIT( 0x10000000, IP_ACTIVE_LOW, IPT_COIN2 ) PORT_IMPULSE(1)
+	PORT_BIT( 0x20000000, IP_ACTIVE_LOW, IPT_COIN1 ) PORT_IMPULSE(1)
```
987123879113, line 3481: This attempts to fix a coin error that would occasionally happened when you coined in.

mamehaze: most Taito games have this whereby if you hold a coin for too long they'll throw a coin error.  A lot of the time we don't bother to limit it though (I think some secret test mode codes require the input to be held in certain cases)

987123879113: Well in this case it actually hinders things so it needed to be fixed to even continue working on it in my WIP branch.

PR #12613, opened 2024-07-24
new NOT WORKING Software List entry

mamehaze:
new NOT WORKING Software List entry
----------
megaduck.xml: Bilder Lexikon [PeT, TeamEurope, SSJ, Youkal3]

"The Bilder Lexikon contains 512kbyte via 32kbyte Bankswitching at $0, but the switching address is $1 like for the 16kbyte bankswitching."

### DIFF
--- a/hash/megaduck.xml
+++ b/hash/megaduck.xml
@@ -14,7 +14,7 @@ license:CC0-1.0
 		<part name="cart" interface="megaduck_cart">
 			<feature name="fixedbank" value="yes" />
 			<dataarea name="rom" size="0x20000">
-				<rom name="028.bin" size="0x20000" crc="43272949" sha1="b2b54faa87996d614941630c7f098431bf699b58" offset="00000" />
+				<rom name="028.bin" size="0x20000" crc="43272949" sha1="b2b54faa87996d614941630c7f098431bf699b58" />
 			</dataarea>
 		</part>
 	</software>
@@ -41,7 +41,7 @@ license:CC0-1.0
 
 		<part name="cart" interface="megaduck_cart">
 			<dataarea name="rom" size="0x8000">
-				<rom name="006.bin" size="0x8000" crc="f88f2d25" sha1="4f3022126f33964e0fb5505fa0caeaa13b995309" offset="00000" />
+				<rom name="006.bin" size="0x8000" crc="f88f2d25" sha1="4f3022126f33964e0fb5505fa0caeaa13b995309" />
 			</dataarea>
 		</part>
 	</software>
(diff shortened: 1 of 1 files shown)


mamehaze:
this isn't a driver I'm familiar with, so I don't really have the interest in adding the bankswitching here.  Everything I find seems to suggest it's meant to be used with the computer version of the system (we have one version of that dumped as mduckspa) but that also doesn't work in MAME.

It's basically a GameBoy, so if anybody familiar with the GameBoy wants to do these feel free, I'm just adding what has been sent to me.

wilbertpol:
This title is also not for the Megaduck/Cougarboy handheld, but for the megaduck "computer".

It apparently uses the same cpu but nothing is known about how the rest of the hardware like the keyboard, mouse, or printer are hooked up. This title seems to check the serial a lot before failing to start up properly.

mamehaze:
yeah, as mentioned there's that single 'mduckspa' system set, which I assume it should work with if the peripherals are hooked up there, but it isn't clear if that's for one region or multiple (there's more than one language in the ROM IIRC)

I was just going through my backlog of things I'd been sent but haven't added yet, and this had been sitting there for a couple of months.

bbbbbr:
Hope it's ok to jump in here :)

Some links in case any of this info is useful for people wanting to work on support for the computer model:

I've done a partial disassembly of the Mega Duck Super Quique computer System ROM (the `mduckspa` mentioned above, which is the Spanish version) mainly focusing on how it communicates with the serial peripherals. (There is a separate non-dumped German version, or... there are some partial dumps I've seen but they're corrupted)
https://github.com/bbbbbr/megaduck-quique-disasm

Lately I've also patching support for the computer's serial peripherals (keyboard, rtc) into a fork of the SameBoy emulator. It's all fairly rough right now but is able to start up the System ROM and type some things in. It's the `sameduck_laptop` fork which is currently the default.
https://github.com/bbbbbr/SameDuck

And there's some source for how I added computer keyboard support into gb-wordyl which can run on the MegaDuck handheld and computer model. Basically a C version of initializing the serial peripheral controller and then polling it for keyboard data:
https://github.com/bbbbbr/gb-wordyl/tree/killjoy/src/megaduck_laptop

bbbbbr:
Got the ROM running and working with the keyboard in my MegaDuck Computer fork of SameBoy mentioned above, which might be used as a reference. Video below.

Some notes:
- The Bank switching register for this ROM **is** at `0x1000` as in the MegaDuck Computer System ROM.
- The SB and SC registers are addressed in the opposite order as GB guidelines indicate (they trigger SC before SB, when it's normally the opposite). This may cause emulation problems via a dropped serial bits/misaligned bytes if it either isn't patched or compensated for.
- As with the System ROM it does a CALL and RET before initializing the stack pointer (SP). If SP defaults to `0x0000` as in many GB emulators, then the ROM will crash part of the way through init when it gets to that first RET. The workaround is to change the SP power-on default to something like `0xFFFE` for the MegaDuck Laptop model.

https://github.com/user-attachments/assets/a75c2809-ea82-41fc-97af-af23152a7267

wilbertpol:
Oh nice. Thank you for the additional information.

wilbertpol:
By the way, using 0xffff as startup value for SP works as well. That writes the first return address to fffd and fffe.

bbbbbr:
Right, good point! :+1:

PR #12615, opened 2024-07-25
a800_flop.xml; a2600.xml: Metadata cleanings

ArcadeShadow:
Lowercase on descriptive words (Prototype, Light Pen, Extensions, etc.)

### DIFF
--- a/hash/a2600.xml
+++ b/hash/a2600.xml
@@ -14787,7 +14787,7 @@ MOS Atari-made game NTSC ROMs had a CO16xxx number and PAL ROMs had CO17xxx numb
 	</software>
 
 	<software name="spacecave" cloneof="spacecav">
-		<description>Space Cavern - Les Guerriers de l'Espace (PAL)</description>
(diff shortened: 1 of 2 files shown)


hash/a2600.xml
```diff
@@ -14787,7 +14787,7 @@ MOS Atari-made game NTSC ROMs had a CO16xxx number and PAL ROMs had CO17xxx numb
 	</software>
 
 	<software name="spacecave" cloneof="spacecav">
-		<description>Space Cavern - Les guerriers de l'espace (PAL)</description>
+		<description>Space Cavern - Les Guerriers de l'Espace (PAL)</description>
 		<year>1981</year>
```
cuavas, line 14791: I think the capitalisation of the subtitle here intentionally uses sentence case to match the box.  See images of front and back here: https://www.atarimania.com/pgesoft.awp?version=12591

ArcadeShadow: You're correct. I undid this change.

PR #12617, opened 2024-07-25
cpu/xa: CPU core work for fearless/superkds - adds sound, promotes to working

mamehaze:
(squashed version of the PR, as requested)

this adds basic execution to XA core.  It emulates enough of the XA to give fearless and superkds working sound.

some considerations
- only the exact forms of the opcodes used so far have been implemented
- no optimizations have been done, use of const, inline use, templates etc. are planned for a future update; code is still primed for debugging and development
- overall structure, code style are not 100% final (see above) and will be adjusted as the CPU is better understood
- peripherals, interrupts etc. are not yet fully implemented, nor is anything outside of the page zero mode used here due to lack of test cases

Machines Promoted to WORKING
Super Kids (S019CN) [David Haywood, XingXing]
Fearless Pinocchio (V101US) [David Haywood, XingXing, Peter Wilhelmsen, rtw]

### DIFF
--- a/scripts/src/cpu.lua
+++ b/scripts/src/cpu.lua
@@ -1209,6 +1209,7 @@ end
 if CPUS["XA"] then
 	files {
 		MAME_DIR .. "src/devices/cpu/xa/xa.cpp",
+		MAME_DIR .. "src/devices/cpu/xa/xa_ops.cpp",
 		MAME_DIR .. "src/devices/cpu/xa/xa.h",
 	}
 end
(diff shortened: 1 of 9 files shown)


rb6502:
This is generally pretty good.  I have one request: search-and-replace to get rid of XA_EXECUTE_PARAMS and XA_EXECUTE_CALL_PARAMS.  It's unnecessary obfuscation and it's a lot more typing than just "u8 op" and "op".  I can accept the disassembler versions where there's a ton of boilerplate parameters going everywhere, but in the execution it's just making it worse for no gain.

mamehaze:
> This is generally pretty good. I have one request: search-and-replace to get rid of XA_EXECUTE_PARAMS and XA_EXECUTE_CALL_PARAMS. It's unnecessary obfuscation and it's a lot more typing than just "u8 op" and "op". I can accept the disassembler versions where there's a ton of boilerplate parameters going everywhere, but in the execution it's just making it worse for no gain.

done

PR #12562, opened 2024-07-10
nmk/nmk16.cpp: upgraded interrupts trigger system based on PROM contents

sergiopolog:
* Added a new system for triggering interrupts based on PROM contents:  `set_interrupt_timing()` and `nmk16_scanline`.
* Bootleg sets, Afega and Comad games, and other sets that PROMs are still undumped (or not exist), on which the new interrupt triggering system cannot be applied, remain using the old `set_hacky_interrupt_timing()` and `nmk16_hacky_scanline` handlers.
* Added info about screen resolutions and timings.
* Removed "hacky" denomination to `set_screen_xxxxxx()` functions as the screen sizes are confirmed for each available resolution
* PROM dumps from `hachamfp` are added to `hachamf` and other clones, and marked as `BAD_DUMP` until they are properly dumped
* PROM dumps from `bjtwin` are added to `nouryoku` and clones, and marked as `BAD_DUMP` until they are properly dumped
* `strahl` and clones could benefit from the new interrupt system, but PROMs for them are undumped. They still use the old system until they are properly dumped
* `ssmissin` and `airattck` "Comad" games could also benefit from the new interrupt system, but the V-Timing PROM on them is a 82S147 (512x8bit) in which only half of space is used and A5 address line is tied to ground. The `nmk16_scanline` code could be tweaked to support them, but more research is needed to apply that. They still use the old system.
* Merged `powerins.cpp` into `nmk16.cpp` driver to make use the new interrupt system, too. New `set_screen_midres()` function is added to be used by `powerins` sets.
* Sprite DMA process might be triggered using PROM contents aswell, but that needs more testing and it could be a future enhancement.

PR #12620, opened 2024-07-27
nmk/ddealer.cpp: upgraded interrupts trigger system based on PROM contents

sergiopolog:
Same work done on this PR for **nmk/nmk16.cpp** driver:

https://github.com/mamedev/mame/pull/12562

is now applied in **nmk/ddealer.cpp**

Double Dealer uses the same "hires" screen resolution than several games in the nmk16 driver, and the same interrupt system based on PROM contents, so the same config is applied to it.

Pixel Clock was confirmed on real hw, too.

### DIFF
--- a/src/mame/nmk/ddealer.cpp
+++ b/src/mame/nmk/ddealer.cpp
@@ -57,6 +57,7 @@ public:
 		m_work_ram(*this, "work_ram"),
 		m_mcu_shared_ram(*this, "mcu_shared_ram"),
 		m_in0_io(*this, "IN0"),
+		m_vtiming_prom(*this, "vtiming"),
 		m_maincpu(*this, "maincpu"),
 		m_protcpu(*this, "protcpu"),
 		m_screen(*this, "screen"),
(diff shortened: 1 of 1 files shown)


cam900:
nmk/quizpani.cpp uses similar hardware, so it's possibly applied a PROM based interrupt system too?

sergiopolog:
> nmk/quizpani.cpp uses similar hardware, so it's possibly applied a PROM based interrupt system too?

I wasn’t aware that `quizpani` has similar hw. By a quick look at the code, it apparently uses the same screen resolution (“hi-res”) and same interrupts: IRQ1 for music and IRQ4 each VBlank.
Also PROMs are dumped, so I could definitely check it and apply if possible.

Thanks for pointing it out!

sergiopolog:
I think this could be merged as is, because `quizpani` board doesn't use NMK902 chip, in which reg-eng work the previous PR is based, so further investigation needs to be done for that case, and that requires to check actual pcb to see how the counters are implemented in this board without NMK902.

PR #12623, opened 2024-07-27
casio/pv1000.cpp: Toggle BUSREQ like hardware does

lidnariq:
This slows down games to realistic speeds.

Please let me know if there's a better way to do any of this.

### DIFF
--- a/src/mame/casio/pv1000.cpp
+++ b/src/mame/casio/pv1000.cpp
@@ -196,6 +196,8 @@ private:
 
 	emu_timer *m_irq_on_timer = nullptr;
 	emu_timer *m_irq_off_timer = nullptr;
+	emu_timer *m_busrq_on_timer = nullptr;
+	emu_timer *m_busrq_off_timer = nullptr;
 	uint8_t m_pcg_bank = 0;
 	uint8_t m_force_pattern = 0;
 	uint8_t m_fd_buffer_flag = 0;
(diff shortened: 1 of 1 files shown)


src/mame/casio/pv1000.cpp
```diff
@@ -266,6 +270,24 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
 		m_pcg_bank = (data & 0xe0) >> 5;
 		m_force_pattern = ((data & 0x10) >> 4); /* Dig Dug relies on this */
 		m_render_disable = ((data & 0x08) >> 3);
+		if (m_render_disable == 0) // If we're enabling rendering mid-scanline...
+		{
+			int hpos = m_screen->hpos();
+			int vpos = m_screen->vpos();
+			hpos += 32; // adjust from screen coordinates to something that's easier to math
```
angelosa, line 277: MAME historically starting at origin 0,0 is just for a generic stub convenience.
This kind of math is unnecessary if you acknowledge overscan origins in `screen_update` and `set_raw` setups.

rb6502: To further clarify, if you do set_raw with the real hcount/vcount values you also have to change where you draw in screen_update to coincide with the specified visible area.  I get *why* MAME does that, but it really does violate the principle of least surprise.

lidnariq:
Is this acceptable now?

PR #12641, opened 2024-08-05
konami/xmen.cpp: Add service switch inputs

jotego:
Connects service ports, following our [schematic extraction for a GX065 board](https://github.com/jotego/jtbin/blob/master/sch/xmen.pdf) (xmen 2p)

image

Note that I am reviewing the schematics and the PDF may change in the coming weeks with corrections.

### DIFF
--- a/src/mame/konami/xmen.cpp
+++ b/src/mame/konami/xmen.cpp
@@ -20,7 +20,7 @@ The board only has one of each gfx chip, the only additional chip not found
 on the 2/4p board is 053253. This chip is also on Run n Gun which is
 likewise a 2 screen game.
 
-Reference schematics: https://github.com/jotego/jtbin/blob/master/sch/xmen.pdf
+Reverse-engineered schematics: https://github.com/jotego/jtbin/blob/master/sch/xmen.pdf
 
 ***************************************************************************/
 
--- a/src/mame/konami/xmen.cpp
+++ b/src/mame/konami/xmen.cpp
@@ -20,6 +20,8 @@ The board only has one of each gfx chip, the only additional chip not found
 on the 2/4p board is 053253. This chip is also on Run n Gun which is
 likewise a 2 screen game.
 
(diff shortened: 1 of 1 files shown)


jotego:
This connection can be verified through the game service menu (F2 key)

rb6502:
The schematic shows the service inputs on bits 1-4, while the patch has them on 0-3.  Which is correct?

cuavas:
> The schematic shows the service inputs on bits 1-4, while the patch has them on 0-3. Which is correct?

You’re misreading it – he’s labelled the lines on the other side of the hex buffer as B1-4 and B7-8, which correspond to D0-3 and D6-7.

cuavas:
> Ok, then the EEPROM connection to B0 is wrong then.

It matches the MAME source though – EEPOROM DO is on B7, which corresponds to D6, i.e. 0x40 (1 << 6) in the source, and READY is on B8, which corresponds to D7, i.e. 0x80 (1 << 7) in the source.

rb6502:
That's an 8?  It looks like a slashed zero.

jotego:
I've changed to "reverse-engineered schematics" for clarity. 

The bus connection in the schematics is a bit confusing indeed. I've changed to this:

image

This should be more readable. I am still reviewing it though. We extracted the schematics last year but didn't get to review them till last week.

happppp:
I did a small correction but you reverted it? Due to the force push it was erased, you can see the revert here (IPT_UNUSED vs IPT_UNKNOWN): https://github.com/mamedev/mame/compare/b7b075f072196377aa2a1343ac0c4932d3d9f8cf..083eb0c58f2fe6020c3914c63f8cce0e0a075d7b

jotego:
Ok! I didn't understand why it rejected my push. I thought it was a problem in my fork. Sorry I didn't notice.

jotego:
But the two bits are unused indeed. They go to nowhere in the schematics. Not even a pull up.

PR #12644, opened 2024-08-08
namco/quickncrash.cpp: Add a third romset, and some little changes

man-of-eel:
Clarifications for the clarifications:

- Cool Gunman namedrop comes from a vague comment [here](https://vixamusements.com/2016/01/02/cool-gunman-information/). IIRC the Quick & Crash main PCB says something like "TAMURA EM GAME PCB" on the silkscreen with no mention of either game, and some clusters of unpopulated silkscreen markings, so it's likely they share that board too.
- I saw some MAME developer say that's how you should implement DMDs, don't think I'll be able to find that now. But to support this, the old pinball machine drivers (that probably haven't been touched in years because of PinMAME) output to video, but some of the fruit machine drivers with a DMD use lamps + a layout. (`sc2ofool` is one)
- The game has two speakers, one in the gun cabinet to play announcer voices and another in the target cabinet to play the crashing and smashing sounds. Since they are both in front of you setting them both to `front_center` seems like the best option

Also i did this in the github editor so if it doesn't compile that's why. Sorry

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -33372,6 +33372,7 @@ topracern                       // bootleg
 @source:namco/quickncrash.cpp
 qncrash                         // (c) 1999 Namco
 qncrasha                        // (c) 1999 Namco
+qncrashb                        // (c) 1999 Namco
 
 @source:namco/rallyx.cpp
 commsega                        // (c) 1983 Sega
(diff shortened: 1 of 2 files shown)


man-of-eel:
Layout i forgot i made if someone wants to build off it. not sure if i want to put it in the pr
[qcrash.lay.txt](https://github.com/user-attachments/files/16589009/qcrash.lay.txt)

Osso13:
Thanks and sorry for the delay getting at this, which caused some merge conflicts. I added the ROM and most of the comments here 650b7e1d711ca13d3e4b51d8f5fa32bc9fa60173

PR #12646, opened 2024-08-08
cpu/t11, ussr/bk: Improve interrupt handling on 1801VM1, misc fixes.

shattered:
Allows bus error handlers on BK to deal with interrupt from STOP key (which causes a HALT-mode IRQ and triggers access to unmapped region at 177674).


shattered:
@AJR ?

holub:
@shattered I'm currently prototyping pk32 which is 1839. Whould you like to participate?

PR #12647, opened 2024-08-08
sinclair/tsconf.cpp: fix layers priority, screen updates

holub:


src/mame/sinclair/tsconf_m.cpp
```diff
@@ -451,6 +478,8 @@ void tsconf_state::cram_write16(offs_t offset, u16 data)
 
 void tsconf_state::sfile_write16(offs_t offset, u16 data)
 {
+	m_sprites_cache.clear();
```
cuavas, line 481: Should this also be cleared in `device_post_load`?  Loading a saved state will cause the contents of memory to change under it.

PR #12651, opened 2024-08-11
konami/k052109.cpp: Remove the color kludge for xmen/gradius3

jotego:
- adds missing information about internal registers
- correctly implements color bit selection (one kludge removed)

Tested on xmen and gradius3 (set the selection bit high) and several of the other games (tmnt,tmnt2,aliens,mia,scontra,thunderx,ssriders) which do not set it.

### DIFF
--- a/src/mame/konami/k052109.cpp
+++ b/src/mame/konami/k052109.cpp
@@ -36,12 +36,10 @@ address lines), and then reading it from the 051962.
 - misc interface stuff
 - ROM bank selector (CAB1-CAB2)
 - character "code" (VC0-VC10)
-- character "color" (COL0-COL7); used foc color but also bank switching and tile
-  flipping. Exact meaning depends on externl connections. All evidence indicates
-  that COL2 and COL3 select the tile bank, and are replaced with the low 2 bits
-  from the bank register. The top 2 bits of the register go to CAB1-CAB2.
-  However, this DOES NOT WORK with Gradius III. "color" seems to pass through
-  unaltered.
+- character "color" (COL0-COL7); used for color but also bank switching and tile
+  flipping. Exact meaning depends on external connections. COL2 and COL3 select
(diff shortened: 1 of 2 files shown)


happppp:
When adding a new variable, don't forget to clear it (here it's done in k052109_device::k052109_device in the cpp file). If it's cleared when the chip is reset, initialize it in device_reset too.
Same goes for adding it to the save_item list in device_start.

jotego:
> When adding a class new variable, don't forget to clear it (here it's done in k052109_device::k052109_device in the cpp file). If it's cleared when the chip is reset, initialize it in device_reset too. Same goes for adding it to the save_item list in device_start.

Thank you. I have added it to the init and save functions.

cuavas:
There’s still a kludge for X-Men at line 358.  It’s pretty gross that it’s setting the `m_has_extra_video_ram` flag based on whether it’s seen a write to tilemap RAM above 0x4000.

What *really* doesn’t make sense is that X-Men sets the `m_k052109_cb` tile mapping callback, but this presumably won’t get used because `m_has_extra_video_ram` will be set so it will just adjust the `code` itself.

jotego:
I agree. The `m_has_extra_video_ram` flag should be set on the K052109 object instantiation from my point of view, as it is fixed on the board and, of course, known before hand.

The whole memory decoding inside the K052109 class is also a per-game kludge. The chip uses the register for which I added some succint information in this PR to map input addresses to each one of the RAM chips. 

But fix the K052109 class to do the decoding like the original would require to change the memory maps of most game drivers. Because the drivers are built assuming the final programmed value, rather than the start up one. Most games start up with one decoding scheme, then the CPU writes to the configuration register in the K052109 plus another bit in a discrete flip flop, whose output is part of the memory decoding logic.

happppp:
On xmen6p, the right side screen colors are broken now, see attract mode.

PR #12659, opened 2024-08-13
casio/pv1000.cpp: implement IRQs (and acknowledgement) like hardware does

lidnariq:
Hardware documentation is at https://obscure.nesdev.org/wiki/Casio_PV-1000/ASIC_registers#Interrupt_enable_($FC)_%3E_write

### DIFF
--- a/src/mame/casio/pv1000.cpp
+++ b/src/mame/casio/pv1000.cpp
@@ -178,7 +178,8 @@ public:
 		m_p_videoram(*this, "videoram"),
 		m_gfxdecode(*this, "gfxdecode"),
 		m_screen(*this, "screen"),
-		m_palette(*this, "palette")
+		m_palette(*this, "palette"),
+		m_joysticks(*this, "IN%u", 0)
 	{ }
 
 	void pv1000(machine_config &config);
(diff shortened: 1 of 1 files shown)


src/mame/casio/pv1000.cpp
```diff
@@ -261,8 +259,23 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
 		m_sound->voice_w(offset, data);
 	break;
 
+	case 0x04:
+		/* Bit 1 = Matrix IRQ enabled */
```
rb6502, line 263: MAME preference is C++ style // for single-line comments and /* */ for multi-line.  You're mostly following that already, just a few exceptions.

rb6502:
The size here is fine, it's just everyone was very busy this week.  This looks good aside from a few instances of comment style mismatch.

src/mame/casio/pv1000.cpp
```diff
@@ -291,39 +304,50 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
 
 uint8_t pv1000_state::io_r(offs_t offset)
 {
+	/* Ports 0xF8-0xFB, 0xFE, and 0xFF are undriven, and pulled high by the
+	   resistors next to the Z80
+
+	   We can remove this bit when we adjust the range */
+
+	if (offset < 4 || offset > 5)
+		return 0xFF;
```
cuavas, line 313: We prefer lowercase digits in hexadecimal literals.

src/mame/casio/pv1000.cpp
```diff
@@ -291,39 +304,50 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
...
+		if (m_io_regs[5] & 1 << i)
 		{
-			if (m_io_regs[5] & 1 << i)
-			{
-				data |= ioport(joynames[i])->read();
-				m_fd_data &= ~(1 << i);
-			}
+			data |= ioport(joynames[i])->read();
 		}
+	}
```
cuavas, line 328: On-the-fly tag lookups are error-prone and hurt performance.  Please use a `required_ioport_array<4>` data member.

src/mame/casio/pv1000.cpp
```diff
@@ -291,39 +304,50 @@ void pv1000_state::io_w(offs_t offset, uint8_t data)
...
 
-		//m_fd_data = 0;
-		break;
+	switch (offset)
+	{
+		case 4: // port $FC returns player 2 joystick and interrupt status
+			data = (data & 0xC) | (m_irq_active & 3);
+			/* Bit 1 = Matrix IRQ asserted */
+			/* Bit 0 = Prerender IRQ asserted */
+			break;
```
cuavas, line 336: Another hex literal that should be lowercase.

PR #12666, opened 2024-08-17
a2bus: Add support for Wico Trackball, Wico Command Control Joystick Adapter, and 4 paddles connected to the Sirius JoyPort to the Apple ][/][+

as-tb-dev:
Adds support for these Apple ][ / ][+ controls:
- Wico Trackball
- Wico Command Control Joystick Adapter
- 4 paddles connected to the Sirius JoyPort

To test the Wico Command Control Adapter with 2 Atari-style joysticks:
mame apple2p -gameio wicojoy 1on1

Play Mode: Two Player
Play Ball!: New Game
Ensure that both players can play using the 1st and 2nd joysticks

To test the Wico Trackball control:
mame apple2p -sl2 wicotrackball apple2_flop_misc:spaceark
Press Control-N
Change the input device from Keyboard to Wico Trackball using the space-bar.
Press Return, and select SLOT -2
Press Esc to return to the game

mame apple2p -sl7 wicotrackball apple2_flop_orig:matezone
Press Control-I
Select: WICO TRACKBALL: 7 and WICO FIREBUTTON: 1
Verify that the trackball and first button works
Press Control-I
Select: WICO TRACKBALL: 7 and WICO FIREBUTTON: 2
Verify that the trackball and second button works

To test the Sirius JoyPort with 4 paddles:
mame apple2p -gameio joyport_paddles foosball
When prompted for the player count, press 4 to select 4 player mode
Verify that each of the four paddles can control a distinct set of player rows


rb6502:
On a first read-through this is generally OK.  My major note is that the actual slot card isn't using the modern pattern with the entire class defined in the .CPP file and DEFINE_DEVICE_TYPE_PRIVATE().  The a2gameio devices in this PR *do* follow that and so do a majority of existing a2bus cards, so you were just unlucky picking an example.  I definitely need to finish modernizing the remaining cards and write some documentation so there's better guidance on this stuff.

as-tb-dev:
Agreed.  I committed a change to move the class definition to the .CPP file and to use DEFINE_DEVICE_TYPE_PRIVATE().

src/devices/bus/a2bus/a2wico_trackball.cpp
```diff
@@ -0,0 +1,248 @@
...
+			temp_axis = 0xff;
+		} else {
+			temp_axis = m_axis[axis] + diff_pos / speed_scale;
+		}
+	}
+	m_last_pos[axis] = cur_pos;
+	m_axis[axis] = temp_axis;
+
+	return m_axis[axis];
+}
```
cuavas, line 183: This is called from a read handler and has side effects, which means it will wreak havoc when debugger memory windows are open.

src/devices/bus/a2bus/a2wico_trackball.cpp
```diff
@@ -0,0 +1,248 @@
...
+			data = read_position(0);
+			break;
+		case 0x1: /* read Y-position */
+			data = read_position(1);
+			break;
+		case 0x2: /* set Bounded/Wraparound Soft Switch to Bounded */
+			m_wraparound = false;
+			break;
+		case 0x3: /* set Bounded/Wraparound Soft Switch to Wraparound */
+			m_wraparound = true;
```
cuavas, line 204: More read side effects that need to be suppressed for the debugger.

PR #12668, opened 2024-08-18
zx81_cass.xml: Added 18 working items + 2 not working

ArcadeShadow:
Accounts [zx81stuff]
Adventure (Gavin Baker) [zx81stuff]
Bridgehead [Team Zeddy, Planeta Sinclair]
Gamespack [Team Zeddy, Planeta Sinclair]
Grand Prix [Planeta Sinclair]
Inheritance [Team Zeddy, Planeta Sinclair]
Le Chateau Hanté [Team Zeddy, Planeta Sinclair]
Purchases [zx81stuff]
Racehorse Trainer [zx81stuff]
Sabotagem (Timex Portugal) [Planeta Sinclair]
Serpents Tomb [Team Zeddy, Planeta Sinclair]
Simulador de Voo (Timex Portugal) [Planeta Sinclair]
Star Trek + 3D O's and X's [zx81stuff]
The Nowotnik Puzzle & Other Diversions [zx81stuff]
Tutor (French) [zx81stuff]
Xadrez (Timex Portugal) [Planeta Sinclair]
Zedman + Spacers [Team Zeddy, Planeta Sinclair]
Zor [zx81stuff]

New NOT working software list additions
Deflexx [Team Zeddy, Planeta Sinclair]
Lemonade [Team Zeddy, Planeta Sinclair]

### DIFF
--- a/hash/zx81_cass.xml
+++ b/hash/zx81_cass.xml
@@ -7,7 +7,7 @@ Sinclair ZX81 cassettes
 
 Loading instructions:
 
-At the prompt, press the following sequence to execute the LOAD "" command: J, LShift+P, LShift+P, Enter
+At the prompt, press the following sequence to execute the LOAD "" command: J, Shift+P, Shift+P, Enter
 Enable UI controls (Scroll Lock) and start the tape (F2).
 If the program doesn't start automatically after loading, execute the RUN command by pressing: R, Enter
 
@@ -262,8 +262,8 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 		<publisher>Hestacrest</publisher>
 		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
-			<dataarea name="cass" size="15859">
-				<rom name="accounts.p" size="15859" crc="407879b7" sha1="0b104f0168ed456f21434e7aee820d24b2f748cd"/>
+			<dataarea name="cass" size="16091">
+				<rom name="accounts.tzx" size="16091" crc="2de97707" sha1="1b6b4a8743548be28680bddae616b39155909837"/>
 			</dataarea>
 		</part>
 	</software>
@@ -313,56 +313,13 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 	</software>
 
 	<software name="adventgb">
-		<description>Adventure (Gavin Baker) (TZX tape image)</description>
+		<description>Adventure (Gavin Baker)</description>
 		<year>1984</year>
 		<publisher>Gavin Baker</publisher>
 		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="48948">
 				<rom name="Adventure(GBarker).tzx" size="48948" crc="61fea9c8" sha1="f09791d6c0d836302ec3d42568e92d9620163bd6"/>
-			</dataarea>
-		</part>
-	</software>
-
-	<software name="adventgb_a" cloneof="adventgb">
-		<description>Adventure (Gavin Baker) (splitted P tape image)</description>
-		<year>1984</year>
-		<publisher>Gavin Baker</publisher>
-		<info name="usage" value="16K" />
-		<part name="cass1" interface="zx81_cass">
-			<feature name="part_id" value="Intro 1"/>
-			<dataarea name="cass" size="2419">
-				<rom name="intro1.p" size="2419" crc="971b6fcd" sha1="7c874c224553f6de610a4288d7d5cc8c75b5ef0d"/>
-			</dataarea>
-		</part>
-		<part name="cass2" interface="zx81_cass">
-			<feature name="part_id" value="The Challenge"/>
-			<dataarea name="cass" size="14629">
-				<rom name="thechallenge.p" size="14629" crc="589ab55a" sha1="867118c9cdfa711ed8ef0db805e9efc8b544c3a8"/>
-			</dataarea>
-		</part>
-		<part name="cass3" interface="zx81_cass">
-			<feature name="part_id" value="Intro 2"/>
-			<dataarea name="cass" size="2100">
-				<rom name="intro2.p" size="2100" crc="a9cf345e" sha1="3c997702c718f2949bedf3ff81b8348a16691d75"/>
-			</dataarea>
-		</part>
-		<part name="cass4" interface="zx81_cass">
-			<feature name="part_id" value="Haunted House"/>
-			<dataarea name="cass" size="13443">
-				<rom name="hauntedhouse.p" size="13443" crc="a11163d1" sha1="5280c2b3b49c396fe961c0d71e7a4180bf16f4bc"/>
-			</dataarea>
-		</part>
-		<part name="cass5" interface="zx81_cass">
-			<feature name="part_id" value="Intro 3"/>
-			<dataarea name="cass" size="2560">
-				<rom name="intro3.p" size="2560" crc="d5ad980f" sha1="7c4b355d8b7b7e3bf3afc46c417742ee01d80525"/>
-			</dataarea>
-		</part>
-		<part name="cass6" interface="zx81_cass">
-			<feature name="part_id" value="In Search of the Black Staff"/>
-			<dataarea name="cass" size="12942">
-				<rom name="insearchoftheblackstaff.p" size="12942" crc="b47e3aab" sha1="2119f059c68f11790fdf836ad26371b09e7f155b"/>
 			</dataarea>
 		</part>
 	</software>
@@ -1002,6 +959,18 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 		</part>
 	</software>
 
+	<software name="bridgehed">
+		<description>Bridgehead</description>
+		<year>1984</year>
+		<publisher>Gavin Barker</publisher>
+		<info name="usage" value="16K" />
+		<part name="cass1" interface="zx81_cass">
+			<dataarea name="cass" size="17462">
+				<rom name="Bridgehead.tzx" size="17462" crc="ba08e53a" sha1="dd59f5192f001f3c48fe64b79a1b71900d18ffd6"/>
+			</dataarea>
+		</part>
+	</software>
+
 	<software name="bubblbug">
 		<description>Bubble Bugs</description>
 		<year>1983</year>
@@ -1290,7 +1259,7 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 	</software>
 
 	<software name="centipedeg" cloneof="centipede">
-		<description>Centipede (dK'tronics - green cover)</description>
+		<description>Centipede (dK'tronics, green cover)</description>
 		<year>198?</year>
 		<publisher>dK'tronics</publisher>
 		<part name="cass1" interface="zx81_cass">
@@ -1334,10 +1303,9 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 	</software>
 
 	<software name="chateauh">
-		<description>Le Chateau Hante</description>
+		<description>Le Chateau Hanté</description>
 		<year>1984</year>
 		<publisher>Sprites</publisher>
-		<info name="alt_title" value="Le Chateau Hanté" /> <!-- in game title -->
 		<info name="language" value="French" />
 		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
@@ -1817,19 +1785,6 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="16379">
 				<rom name="Dallas(French).tzx" size="16379" crc="4a65562e" sha1="499bee247f3bdc37eca67e95eee1624de11ec5d6"/>
-			</dataarea>
-		</part>
-	</software>
-
-	<software name="dallasf_a" cloneof="dallasf">
-		<description>Dallas (French) (P tape image)</description>
-		<year>1983</year>
-		<publisher>Cases Computer Simulations</publisher>
-		<info name="language" value="French" />
-		<info name="usage" value="16K" />
-		<part name="cass1" interface="zx81_cass">
-			<dataarea name="cass" size="16151">
-				<rom name="Dallas(French).p" size="16151" crc="6fc5f7ec" sha1="1d1cd2f2ce3cdfc7a8c18e8bdd1cd0cdc8f2797e"/>
 			</dataarea>
 		</part>
 	</software>
@@ -5063,7 +5018,7 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 	</software>
 
 	<software name="nowotnik">
-		<description>The Nowotnik Puzzle &amp; Other Diversions (TZX tape image)</description>
+		<description>The Nowotnik Puzzle &amp; Other Diversions</description>
 		<year>1982</year>
 		<publisher>Phipps Associates</publisher>
 		<info name="usage" value="16K" />
@@ -5071,31 +5026,6 @@ TZX files for ZX81 are just the .P dumps with an additional header on top, so th
 			<feature name="part_id" value="1.Puzzle 2.Demolition 3.Tenpin"/>
 			<dataarea name="cass" size="20711">
 				<rom name="NowotnikPuzzle.tzx" size="20711" crc="fd11e293" sha1="bafa4197d4ba285fd31f9147ccd47eed9cfc514a"/>
-			</dataarea>
-		</part>
-	</software>
-
-	<software name="nowotnik_a" cloneof="nowotnik">
-		<description>The Nowotnik Puzzle &amp; Other Diversions (splitted P tape image)</description>
-		<year>1982</year>
-		<publisher>Phipps Associates</publisher>
-		<info name="usage" value="16K" />
-		<part name="cass1" interface="zx81_cass">
-			<feature name="part_id" value="Puzzle"/>
-			<dataarea name="cass" size="7385">
-				<rom name="puzzle.p" size="7385" crc="0fcd568c" sha1="c397878e951cfebeb5cd55a7bd0a9c1cf011960a"/>
-			</dataarea>
-		</part>
-		<part name="cass2" interface="zx81_cass">
-			<feature name="part_id" value="Demolition"/>
-			<dataarea name="cass" size="3565">
-				<rom name="demolition.p" size="3565" crc="9c5bc7d3" sha1="0097c3d7b20c0163b315bcef68dabb2e1e190e92"/>
-			</dataarea>
-		</part>
-		<part name="cass3" interface="zx81_cass">
-			<feature name="part_id" value="Tenpin"/>
-			<dataarea name="cass" size="9283">
-				<rom name="tenpin.p" size="9283" crc="a14ba685" sha1="84459743b188ec326890c759589aee37392bcfe8"/>
 			</dataarea>
 		</part>
 	</software>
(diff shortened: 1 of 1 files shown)


Pernod70:
Why are you adding P files for Accounts, Dallas (French), The Nowotnik Puzzle, Purchases, Racehorse Trainer, etc. when TZX files exist? In some cases you're adding both formats which contain exactly the same data, seems like unnecessary duplication. The TZX versions are easier to use as it's an image of the whole tape, whereas P files have to be mounted individually as a program loads. I'm genuinely interested in knowing why you're preferring P files.

ArcadeShadow:
On the zx81_cass.xml hash file, it is mentioned that P files are cleaner than TZX:

_Due to the way the ZX81 saves to tape, different recordings of identical data (such as those in both sides of a cassette)
might be stored along with a few different bytes in them. Additional dumps of the same releases should match those differences.
TZX files for ZX81 are just the .P dumps with an additional header on top, so they may not be the cleanest format for archival._

That's why I've added this format on some sets.

Pernod70:
I don't agree with that comment, if it's misleading then it should be removed. My understanding is that a TZX contains more information from the original tape as it also includes the gaps between multiple files. You can easily extract P files from a TZX but to create a TZX from P files you need additional information such as file order and gap sizes between the files, therefore TZX should be preferred of the two formats. 
Personally, if you have both formats for a software item then just include the TZX, otherwise it's just adding needless duplication.

ArcadeShadow:
I can keep only the TZX set, I have no problem about that.
Let's see if anyone else give their 2 cents about that.

ICEknigh7:
I suppose that paragraph only applies to games with one block (which is the vast majority of existing dumps), in which case there's no pauses to be accounted for and their TZXs are indeed P dumps with an additional header (just checked with existing 3D Monster Maze dumps).

hash/zx81_cass.xml
```diff
@@ -310,6 +312,61 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="usage" value="16K" />
+		<part name="cass1" interface="zx81_cass">
+			<dataarea name="cass" size="48948">
+				<rom name="Adventure(GBarker).tzx" size="48948" crc="61fea9c8" sha1="f09791d6c0d836302ec3d42568e92d9620163bd6"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="adventgb_a" cloneof="adventgb">
+		<description>Adventure (Gavin Baker) (splitted P tape image)</description>
```
cuavas, line 328: Is there a reason for using the obsolete past tense “splitted”?

hash/zx81_cass.xml
```diff
@@ -1276,6 +1333,20 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
+	<software name="chateauh">
+		<description>Le Chateau Hante</description>
+		<year>1984</year>
+		<publisher>Sprites</publisher>
+		<info name="alt_title" value="Le Chateau Hanté" /> <!-- in game title -->
+		<info name="language" value="French" />
```
cuavas, line 1341: Is there a good reason to have “Hante” without the acute accent in the description?  It looks like it’s just a result of the printing system they used lacking support for accented uppercase letters, not an actual stylistic choice.

hash/zx81_cass.xml
```diff
@@ -4890,6 +5062,44 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<part name="cass1" interface="zx81_cass">
+			<feature name="part_id" value="1.Puzzle 2.Demolition 3.Tenpin"/>
+			<dataarea name="cass" size="20711">
+				<rom name="NowotnikPuzzle.tzx" size="20711" crc="fd11e293" sha1="bafa4197d4ba285fd31f9147ccd47eed9cfc514a"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="nowotnik_a" cloneof="nowotnik">
+		<description>The Nowotnik Puzzle &amp; Other Diversions (splitted P tape image)</description>
```
cuavas, line 5079: Once again, why the obsolete “splitted”?

hash/zx81_cass.xml
```diff
@@ -5634,6 +5856,38 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<part name="cass1" interface="zx81_cass">
+			<feature name="part_id" value="1.Puzzle 2.Demolition 3.Tenpin"/>
+			<dataarea name="cass" size="17083">
+				<rom name="RacehorseTrainer.tzx" size="17083" crc="c25b4909" sha1="7b970f528e90da9faf7093204bf9bf940124c7d0"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="racehors_a" cloneof="racehors">
+		<description>Racehorse Trainer (splitted P tape image)</description>
```
cuavas, line 5873: Here as well.

hash/zx81_cass.xml
```diff
@@ -7886,6 +8211,37 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="usage" value="16K" />
+		<part name="cass1" interface="zx81_cass">
+			<dataarea name="cass" size="17561">
+				<rom name="Zor.tzx" size="17561" crc="981a342e" sha1="9ca9f0ea3ff878174bbdbeec1864759e2f142a6a"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="zor_a" cloneof="zor">
+		<description>Zor (splitted P tape image)</description>
```
cuavas, line 8227: And here.

hash/zx81_cass.xml
```diff
@@ -1233,7 +1290,7 @@ Known dumps not yet added (as of 2023-06-14):
 	</software>
 
 	<software name="centipedeg" cloneof="centipede">
-		<description>Centipede (Green Cover)</description>
+		<description>Centipede (dK'tronics - green cover)</description>
```
cuavas, line 1293: Why the dash here in “dK'tronics - green cover” when you’ve used a comma in other similar situations (e.g. “Mikro-Gen, orange cover” for `chess_b`)?

cuavas:
I agree with @Pernod70 about not adding “cooked” P images containing the same data as better formats.

ArcadeShadow:
> I agree with @Pernod70 about not adding “cooked” P images containing the same data as better formats.

OK. Fixed.

hash/zx81_cass.xml
ArcadeShadow: fixed

hash/zx81_cass.xml
ArcadeShadow: removed P image entries

hash/zx81_cass.xml
ArcadeShadow: I changed the description, but what is the rule? I thought it would always be the name that would be in the box/inlay, regardless of not having the accents.
Because in this case there's a lot of games (ZX Spectrum and Oric 1) that the accented letters are missing.

hash/zx81_cass.xml
```diff
@@ -1305,10 +1345,11 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
-	<software name="chessmg" cloneof="chess">
-		<description>Chess (MikroGen)</description>
+	<software name="chess_a" cloneof="chess">
```
Pernod70, line 1348: Why are you changing these shortnames? The 'mg' for MikroGen was more descriptive than your generic '_a'.

hash/zx81_cass.xml
```diff
@@ -1323,10 +1364,11 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
-	<software name="chessp" cloneof="chess">
-		<description>Chess (Psion)</description>
+	<software name="chess_b" cloneof="chess">
```
Pernod70, line 1367: chessmgo would be more descriptive.

hash/zx81_cass.xml
```diff
@@ -1341,7 +1383,7 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
-	<software name="chesst" cloneof="chess">
+	<software name="chess_c" cloneof="chess">
```
Pernod70, line 1386: 't' for Timex is better than '_c', please revert.

hash/zx81_cass.xml
```diff
@@ -1352,8 +1394,28 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
-	<software name="chessty" cloneof="chess">
-		<description>Chess (Typed)</description>
+	<software name="chess_d" cloneof="chess">
```
Pernod70, line 1397: Please give it it's actual name 'xadrez'.

hash/zx81_cass.xml
```diff
@@ -1715,9 +1777,11 @@ Known dumps not yet added (as of 2023-06-14):
 	</software>
 
 	<software name="dallasf">
-		<description>Dallas (French)</description>
+		<description>Dallas (French) (TZX tape image)</description>
```
Pernod70, line 1780: Remove the TZX label.

hash/zx81_cass.xml
```diff
@@ -2231,35 +2306,51 @@ Known dumps not yet added (as of 2023-06-14):
...
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="13855">
 				<rom name="FlightSimulation.tzx" size="13855" crc="945da3c1" sha1="de020d7531c0904a8a152193e7b3d5203e5c1017"/>
 			</dataarea>
 		</part>
 	</software>
 
-	<software name="flsimult" cloneof="flsimul">
-		<description>The Flight Simulator</description>
+	<software name="flsimul_a" cloneof="flsimul">
```
Pernod70, line 2317: Please revert this shortname.

hash/zx81_cass.xml
```diff
@@ -2231,35 +2306,51 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="13855">
 				<rom name="FlightSimulation.tzx" size="13855" crc="945da3c1" sha1="de020d7531c0904a8a152193e7b3d5203e5c1017"/>
 			</dataarea>
 		</part>
 	</software>
 
-	<software name="flugsim" cloneof="flsimul">
+	<software name="flsimul_b" cloneof="flsimul">
```
Pernod70, line 2329: 'flugsim' was perfectly reasonable, please revert.

hash/zx81_cass.xml
```diff
@@ -2231,35 +2306,51 @@ Known dumps not yet added (as of 2023-06-14):
...
 		<publisher>Sinclair Research</publisher>
+		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="13855">
 				<rom name="FlightSimulation.tzx" size="13855" crc="945da3c1" sha1="de020d7531c0904a8a152193e7b3d5203e5c1017"/>
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="flsimul_c" cloneof="flsimul">
```
Pernod70, line 2341: 'simdevoo' would be more descriptive.

hash/zx81_cass.xml
```diff
@@ -5833,13 +5997,28 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="language" value="English" />
+		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="9070">
 				<rom name="Sabotage.tzx" size="9070" crc="0120fe03" sha1="4d123b512a757fe2e0bfd7ddee111322d8238d9f"/>
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="sabotage_a" cloneof="sabotage">
```
Pernod70, line 6009: 'sabotagm' would be more descriptive.

hash/zx81_cass.xml
ArcadeShadow: fixed

hash/zx81_cass.xml
cuavas: Yeah, even if you were going to go for “parent plus suffix”, something mnemonic like `chess_pt` would be a lot more descriptive than a non-descript letter suffix.

hash/zx81_cass.xml
cuavas: Well you converted it to title case, which isn’t a verbatim copy of what’s shown on the box anyway.  The only reason it has no accent on the box is because it’s in all uppercase and their printing system doesn’t appear to support accented uppercase.  When it isn’t forced into uppercase without accents it’s correctly written as “Hanté”.

ArcadeShadow:
Personally I prefer to start the clone's name with the prefix composed by the parent's name plus the underscore character. Whatever that, I've changed the clones sets' name to a more understandable name, hoping it pleases the most people.

hash/zx81_cass.xml
```diff
@@ -2253,13 +2330,27 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="13855">
 				<rom name="FlightSimulation.tzx" size="13855" crc="945da3c1" sha1="de020d7531c0904a8a152193e7b3d5203e5c1017"/>
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="flsimul_txp" cloneof="flsimul">
+		<description>Simulador de Voo (Timex Portugal)</description>
```
Pernod70, line 2342: Again, no need to translate the title for the shortname, something like 'simdevoo' is suitable.

hash/zx81_cass.xml
```diff
@@ -5833,13 +6010,28 @@ Known dumps not yet added (as of 2023-06-14):
...
+		<info name="usage" value="16K" />
 		<part name="cass1" interface="zx81_cass">
 			<dataarea name="cass" size="9070">
 				<rom name="Sabotage.tzx" size="9070" crc="0120fe03" sha1="4d123b512a757fe2e0bfd7ddee111322d8238d9f"/>
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="sabotage_txp" cloneof="sabotage">
+		<description>Sabotagem (Timex Portugal)</description>
```
Pernod70, line 6023: And again you ignored my previous advice, base the shortname on the actual title so 'sabotagm'.

hash/zx81_cass.xml
```diff
@@ -6351,8 +6556,27 @@ Known dumps not yet added (as of 2023-06-14):
 		</part>
 	</software>
 
+	<software name="startrox">
+		<description>Star Trek + 3D O's and X's</description>
+		<year>1981</year>
+		<publisher>Macronics</publisher>
+		<info name="usage" value="16K" />
+		<part name="cass1" interface="zx81_cass">
+			<feature name="part_id" value="Side A"/>
```
Pernod70, line 6565: Useful to specify that this side is 'Star Trek'.

hash/zx81_cass.xml
```diff
@@ -6351,8 +6556,27 @@ Known dumps not yet added (as of 2023-06-14):
...
+			<dataarea name="cass" size="14586">
+				<rom name="StarTrekAnd3DOsAndXs.A.tzx" size="14586" crc="b9a844e7" sha1="7094c5142b1cea501d8c2c1ba556436c9794e2d3"/>
+			</dataarea>
+		</part>
+		<part name="cass2" interface="zx81_cass">
+			<feature name="part_id" value="Side B"/>
```
Pernod70, line 6571: And this side is "3D O's and X's".

Pernod70 (commented):
Personally I prefer using the actual name of a game for shortnames rather than using the parents name. I'm surprised you didn't give 'Le Chateau Hante' the shortname 'haunted' or similar. So in a few months time when someone else adds something to this list are they free change everything to their personal preference too? Additions are always welcome but if everyone that changes a file also applies their personal preferences it would be chaos, always adhere to the style of the file you're working on.

ArcadeShadow:
I'm trying the changes using a rule. 
Clone records are prefixed with the parent's name, followed by an underscore and at the end a suffix that distinguishes them.
This way, files associated with a set of parents/clones are easily identified.
I have been applying this rule and carrying out this work on several software lists.
If this is not the rule, then define or say which rule should be applied across software lists.

If the changes I made aren't appropriate. feel free to close this PR.

happppp:
I don't think swlists should be different from MAME romsets. Parent shortname + a/b/c for other revisions, or for example j/u for Japan/USA. Best avoid underscores, it makes it harder to type from commandline. If a clone has a different title altogether, use that.

ICEknigh7:
I'd merge it as it is and let whoever wants to change them to open their own PR.

Personally, I think any shortnames should be fine as long as they convey what the description tag says.

ICEknigh7:
I'd just fix the shortnames that don't match the description (translated titles) and merge, then let whoever wants to change certain style aspects to open their own PR for it.

Personally, I think any shortnames should be fine as long as they convey what the full description tag says. Maybe add a "no underscores" rule if it's an incommodity.

ICEknigh7:
Personally, I think any shortnames should be fine as long as they convey what the full `description` tag says. Maybe add a "no underscores" rule if it's an incommodity.

Oh and yeah, I guess the paragraph about .P and .TZX dumps should be removed since it doesn't apply to all dumps, though I thought that TZXs were considered by some as something to be avoided.

Pernod70:
This PR is about adding numerous titles, no problem. There's no need to change any of the shortnames which were all perfectly descriptive before this PR.

I've now taken a look at spectrum_cass.xml which has lots of Jet Set Willy clones, they are currently jetsetwi_(i-z)(aa-az)(ba-bz)(ca-cz)(da-dl), these all had descriptive shortnames when originally added. This is vandalism and has to stop!

ArcadeShadow:
About `spectrum_cass` SW, someone imported thousands of sets around 0.211 MAME version, which was a complete mess with wrong parent/clone relationships and without information, and already with those names. Which I'm trying to fix all that confusion over the last years.
But for me, I'll stop that work. I don't want to feed more this PR with this subject.

ICEknigh7:
I do think that WAV/FLAC files are something that should also be added (once the known issues with reversed polarity have been fixed where needed), preferably in separate `_audio` softlists with shortnames that correspond to their equivalent entries in the softlist for the "cooked" (is this really a technical term?) formats.

But yes, it's also a must to keep preserving formats such as TZX and its siblings CDT and TSX (somebody please allow its support as there's already more than 1600 verified MSX dumps waiting to be added), since they'd act as an OCR of sorts for the audio recordings.

mamehaze:
> I do think that WAV/FLAC files are something that should also be added (once the known issues with reversed polarity have been fixed where needed), preferably in separate `_audio` softlists with shortnames that correspond to their equivalent entries in the softlist for the "cooked" (is this really a technical term?) formats.
> 
> But yes, it's also a must to keep preserving formats such as TZX and its siblings CDT and TSX (somebody please allow its support as there's already more than 1600 verified MSX dumps waiting to be added), since they'd act as an OCR of sorts for the audio recordings.

yeah, alck of TSX support is the main reason I'm seeing people say to avoid MAME's MSX like the plague these days, even if our emulation actually seems on par with the best of them.

happppp:
This is getting off-topic, but I haven't heard anyone say to avoid MAME for MSX specifically because of lack of TSX (on msx.org for example).
openMSX is the go-to MSX emulator and AFAIK it doesn't officially support TSX either.

Anyway, back when TSX was new, my opinion was so-so, these days I'm ok with it due to the ongoing MSX cassette software TSX preservation efforts.

ArcadeShadow:
Fixed the names.
@cuavas now I think it is ready to merge.

Pernod70 (approved):

holub:
TSX: #13317

PR #12671, opened 2024-08-19
igs_m027.cpp - add 4 more internal IGS027A dumps [Guru, XingXing]

mamehaze:


Gurudumps:
I checked several of my boards and the 027A clock is 22MHz on these too so it's probably safe to say they all run at 22MHz, especially if there is a 22MHz OSC on the board.
For future reference and board checking CLK_IN is pin 53 on the actual chip or pin 52 on the unpopulated header/connector near the 027A.

PR #12674, opened 2024-08-20
igs017.cpp - fix a regression with starzan

mamehaze:


src/mame/igs/igs_m027.cpp
```diff
@@ -323,6 +369,59 @@ TIMER_DEVICE_CALLBACK_MEMBER(igs_m027_state::interrupt)
...
+	switch (m_dsw_io_select & 0x7) // 0x10 is set in test mode
+	{
+	case 0x03: return m_dsw3->read();
+	case 0x05: return m_dsw2->read();
+	case 0x06: return m_dsw1->read();
+
+	default: // slqz3 does't write the io_select, maybe different PCB?
+		logerror("(unhandled)\n");
+		return 0xff;
+	}
```
cuavas, line 385: This is a switch matrix – bits 0, 1 and 2 are the row drive for DIP switch banks 1, 2 and 3, respectively.  Please do it properly by testing bits and ANDing the port values rather than using a switch statement.

mamehaze: I'm not sure what anything is for now, I'm learning / experimenting and making notes, nitpicking is not appreciated

cuavas: If you’re experimenting and don’t want your code reviewed, please mark the pull request as a draft and let us know you don’t want it reviewed until you mark it ready for review.

mamehaze: I still don't think you get MAME at all, everything is an iterative process.  I'm just deleting the code you don't want and leaving it with less functionality, you can add it to your own preference then.

mamehaze:
ready for review, but I'm abandoning this now.

src/mame/igs/igs_m027.cpp
cuavas: You don’t seem to understand working with people at all.

If you create an open pull request, it will get reviewed.  If you’re experimenting and don’t want feedback, don’t create an open pull request.  Since some people open draft pull requests for the purpose of soliciting feedback, it’s best to say you don’t want feedback on a draft pull request if that’s the case.

I keep finding obvious cases where a switch matrix is misidentified as a multiplexer or priority encoder.  The code is simpler reflects the hardware better if it’s implemented as a switch matrix.

Calling any feedback “nitpicking” and pushing commits with passive-aggressive or profanity-laced and abusive messages is not helpful.  It’s also incredibly hypocritical for someone who accuses others of “rudeness” for being direct with feedback.

src/mame/igs/igs017_igs031.cpp
```diff
@@ -570,9 +570,9 @@ void igs017_igs031_device::tjsb_decrypt_sprites()
...
 }
 
-void igs017_igs031_device::tarzan_decrypt_sprites(size_t max_size)
+void igs017_igs031_device::tarzan_decrypt_sprites(size_t max_size, size_t flip_size)
 {
-	mgcs_flip_sprites(max_size);
+	mgcs_flip_sprites(flip_size);
 
 	const int rom_size = max_size ? max_size : memregion("sprites")->bytes();
 	u8 *rom = memregion("sprites")->base();
```
cuavas, line 578: This should throw a fatal error if `max_size` is greater than the size of the ROM region.  With the code as-is, it will read and then write off the end of the ROM region if someone inadvertently passes a `max_size` that’s too large.

mamehaze: again you're nitpicking over me following the style of the logic already in the code.

src/mame/igs/igs_m027.cpp
mamehaze: well, you can implement it then, if you're so sure that's what it is.

I was still looking around trying to make notes on what I was seeing.

src/mame/igs/igs017_igs031.cpp
cuavas: You’re trying to use the justification that because code already in MAME has a particular issue, it’s fine to add more code with the same issue.  That’s how issues propagate and technical debt builds up.

mamehaze: and when people don't stick to the style already in MAME you mean that they're not doing that.

anyway, as I said, I've abandoned this now.  I was going above and beyond by even trying to get it to do something, when really the only interest I have in this is making sure the dumps we were getting were good.

my effort now drops to absolute minimum for this.

Osso13:
I applied the changes myself, adding the fatal_error

PR #12675, opened 2024-08-20
machine/intelfsh.cpp: Add ST M29W640GB

wilbertpol:


src/devices/machine/intelfsh.cpp
```diff
@@ -912,6 +940,38 @@ void intelfsh_device::write_full(uint32_t address, uint32_t data)
...
+			m_data[address] = data;
+			break;
+		case 16:  // 16bit mode not tested
+			m_data[address*2] = data >> 8;
+			m_data[address*2+1] = data;
+			break;
+		default:
+			logerror("FM_WORDPROGRAM not supported when m_bits == %d (address %08x data %04x)\n", m_bits, address, data);
+			break;
+		}
```
cuavas, line 972: Should these `logerror` in 16-bit mode if it isn’t fully trusted?

src/devices/machine/intelfsh.cpp
```diff
@@ -463,6 +471,18 @@ uint32_t intelfsh_device::read_full(uint32_t address)
...
+		else if (m_maker_id == MFG_ST)
+		{
+			switch (address & 0xff)
+			{
+				case 0: data = m_maker_id; break;
+				case 2: data = m_device_id; break;
+				case 4: data = 0; break;
+				// TODO 6 - extended memory block verify code
+				case 0x1c: data = m_device_id2; break;
+				case 0x1e: data = m_device_id3; break;
```
cuavas, line 483: You should at least comment that 0x04 is unimplemented block protection status.

cuavas (commented):
All the logic for choosing the correct logic for each supported device is getting very complicated.  Does it lend itself to simplification using a feature bitset or something?

wilbertpol:
> All the logic for choosing the correct logic for each supported device is getting very complicated. Does it lend itself to simplification using a feature bitset or something?

I agree that it is getting quite complicated, I thought it was just me not knowing enough about flash chips. I think there is also a growing risk of accidentily supporting commands that some chips never supported.

PR #12679, opened 2024-08-21
Fix MAME Debugger GUI issues: horizontal disassembly scrolling and maxwidth

dave-br:
@tlindner : FYI

This fixes https://github.com/mamedev/mame/issues/12678

### PROBLEMS FIXED
1. Horizontally scrolling disassembly pane causes garbage to appear.
2. The parent debugger window refuses to be widened or horizontally scrolled even when disassembly contents are very wide (e.g., due to comments).

NOTE: I have observed the above problems on the MAME Windows Debugger under Windows.  I briefly tried the QT Debugger under Windows: Problem 1 occurs, but Problem 2 seems ok.  I briefly tried the imgui debugger, but it crashes every time I click on it.  I do not have access to MacOS or Linux.

### SCENARIOS FIXED

Anyone using the disassembly window (either embedded in the main window or free-floating) might hit problem (1) if the window is sized narrowly enough.  But the real issue is for folks (like me) who extensively use the comments feature.  In particular, I use a CoCo-based tool which allows the original assembly source you're debugging to appear as a comment for each machine code instruction (complete with labels and symbols).  This can make the information in the disassembly view quite wide (each comment alone could be about 100 characters).  Inability to widen the window or scroll the right portion of the comments into view is a pain.

### LIST OF CODE CHANGES

- debug_view_disasm::redraw: Fixed calls to print that were passing a "relative" end parameter (a column number based on where the horizontal scrollbar was situated) so they instead pass the expected "absolute" end parameter (a column number independent of what is horizontally scrolled into view).  This fixes the garbage (problem 1).

- debug_view_disasm: Replace hard-coded value of 50 for comment length with a dynamically calculated value based on the comments retrieved for the current buffer of disassembly to be shown (debug_view_disasm::complete_information).

- consolewin_info: Factor code that computes minwidth/maxwidth based on child windows so it may also be called from consolewin_info::recompute_children.

- debugwin_info::window_proc: Added call to recompute_children() at top of case WM_GETMINMAXINFO.  This ensures that when the system queries the parent window's min/max width, the parent first consults its children (specifically the disassembly window) so it calculates the values properly.

- Some minor changes to replace use of "int" in neighboring code with s32/u32.

### DIFF
--- a/src/emu/debug/dvdisasm.cpp
+++ b/src/emu/debug/dvdisasm.cpp
@@ -352,8 +352,25 @@ void debug_view_disasm::generate_dasm(debug_disasm_buffer &buffer, offs_t pc)
 	generate_from_address(buffer, pc);
 }
 
-void debug_view_disasm::complete_information(const debug_view_disasm_source &source, debug_disasm_buffer &buffer, offs_t pc)
+//-------------------------------------------------
+//  complete_information - helper used by
+//  view_update() to populate m_dasm just before
+//  calling redraw().  This includes three
+//  chunks of text: address, op codes + params,
+// 	and user-selected option (raw op codes /
+//  encrypted op codes / comments).
+//
+//  Returns the maximum number of characters
+//  needed for that third chunk: maximum comment
+//  length, but always at least 50 to accomodate
+//  other otions
+//-------------------------------------------------
+
+u32 debug_view_disasm::complete_information(const debug_view_disasm_source &source, debug_disasm_buffer &buffer, offs_t pc)
 {
+	// Always allow at least 50 characters for third chunk of text on each line
+	u32 max_comment_length = 50;
+
 	for(auto &dasm : m_dasm) {
 		offs_t adr = dasm.m_address;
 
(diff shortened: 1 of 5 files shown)


cuavas (commented):
This change is fundamentally incompatible with the Cocoa debugger.  The Cocoa debugger needs to know the logical size of the virtual canvas for the whole scrollable area.  It won’t work if scrolling can cause that to change.

cuavas:
The issue with garbage when horizontally scrolling disassembly views seems to stem from the disassembly view incorrectly calculating the rightmost visible column incorrectly.  Is it always using the visible width rather than visible width plus leftmost visible column?  Just fixing that issue should be straightforward and won’t risk breaking debugger implementations.

cuavas:
I think what happens with the Qt debugger is that you can’t scroll a comment that’s longer than 50 characters into view if the window is too narrow to show it at once, but the maximum window size isn’t restricted, so you can make the window as wide as you like and see longer comments as long as they fit on the screen.

IIRC the Cocoa debugger doesn’t restrict maximum window sizes, either.

Is it simple enough to simply make remove the maximum window size restriction for the windows debugger for now?  That should at least bring it on par with the other debuggers.

dave-br:
I've updated the PR the way I think you're suggesting.  It's a bit odd that the scrollbar doesn't let you scroll to the end when comments are long enough, but you can resize the window to view them all.  Though with this fix, the window can't be resized wider than the screen--even if the left portion of the window is moved off-screen to make room for stretching the right side--which means that extremely long comments still can't be viewed entirely.  This is likely not a problem for my scenarios, though.

It does behave similarly to QT, so at least it's on par as you suggest.

If you're amenable, I'm still interested in trying out an initialization-time-only setting of max width, to allow the horz scrollbar to be used for arbitrarily long comments.  This would still keep dynamic changes out of the picture, while making for a more natural use of the scrollbar (assuming it works).

PR #12680, opened 2024-08-22
igs_m027.cpp - add 2 more 027A internal dumps (and 2 in igs_m027_link too) [Guru, XingXing]

mamehaze:
igs_m027.cpp - add 2 more 027A internal dumps [Guru, XingXing]
igs_m027_link.cpp - add 2 more 027A internal dumps from the link 'HOST' boards [Guru, XingXing]


mamehaze:
as per comment, there are 4x025 chips in here that really need their labels too

PR #12681, opened 2024-08-22
msx1_cart.xml: Promoted 1 item to working.

wilbertpol:
X & Y (Arab)


cuavas (commented):
The `<notes>` element is effectively a comment that isn’t a comment syntactically, and hence doesn’t show in UIs.  Can you provide a `usage` info element giving a hint as to which systems it may work with?

wilbertpol:
I have moved the comment to a usage element and changed the text to include some systems that do work.

PR #12682, opened 2024-08-23
Add a Skeleton driver For Lai pink skilltester

flama12333:
its a Crane machine.
Dumped by Ramclaw.
sadly internal rom are not dumped.
No progress until the internal rom is dumped
pdf https://drive.google.com/file/d/1AKCjoVtmLqWX0Y-Lp6BBS6pE5jokx0Lh/view?usp=drivesdk
idk is ok to add without internal rom?.

I'm going to check and change the CPU clock.
Then the DIP switch.nothing else.


flama12333:
Closing since without internal code. cant do anything

PR #12684, opened 2024-08-23
Improvements to Technics KN5000

felipesanches:
This is ready for review and replaces PR #11835 but does not include the kn1500 driver, which I'll leave for a future PR.


felipesanches:
After @ajrhacker made this comment (https://github.com/mamedev/mame/pull/11835#issuecomment-2307930039) I investigated the problem a bit more and I was able to get rid of the segfault on the extension interface for the Technics KN5000 (which is used to install the HDAE5000 board). 

I've just force-pushed the amended fix.

src/mame/matsushita/kn5000.cpp
```diff
@@ -591,6 +618,16 @@ void kn5000_state::machine_start()
 	save_item(NAME(m_mstat));
 	save_item(NAME(m_sstat));
 
+	if(m_extension)
+	{
```
cuavas, line 622: The existing code in this file (as well as the majority of MAME) uses a space between flow control keywords and the opening parenthesis of the controlling expression, like `if (whatever)`.

src/mame/matsushita/kn5000.cpp
```diff
@@ -591,6 +618,16 @@ void kn5000_state::machine_start()
...
+		m_extension->rom_map(m_extension_view[1], 0x280000, 0x2fffff);
+		m_extension_view.select(1);
+	}
+	else
+	{
+		m_extension_view.select(0);
+	}
```
cuavas, line 629: This looks like you’re using views inappropriately.  The whole point of using a view is so you don’t have to install handlers on-the-fly.  If you’re installing handlers when memory mapping changes, you’re doing it wrong.  You should make the extension device set up whatever it needs to once at start.

src/mame/matsushita/kn5000.cpp
```diff
@@ -746,6 +784,16 @@ void kn5000_state::kn5000(machine_config &config)
 
 	FLOPPY_CONNECTOR(config, "fdc:0", kn5000_floppies, "35dd", floppy_image_device::default_mfm_floppy_formats).enable_sound(true);
 
+    /* extension port */
+	KN5000_EXTENSION(config, m_extension, 16_MHz_XTAL / 16); //FIXME: which clock signal should be passed to the extension port?
+//	m_extension->firq_callback().set("mainfirq", FUNC(input_merger_device::in_w<3>)); // FIXME: do we need something like this?
+//	m_extension->irq_callback().set("mainirq", FUNC(input_merger_device::in_w<3>)); // FIXME: do we need something like this?
+	m_extension->option_add("hdae5000", HDAE5000);
```
cuavas, line 791: Please `srcclean` this to reduce churn.

scripts/src/bus.lua
```diff
@@ -2577,6 +2577,21 @@ if (BUSES["SS50"]~=null) then
...
+---------------------------------------------------
+
+if (BUSES["TECHNICS"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/bus/technics/kn5000_extension.cpp",
+		MAME_DIR .. "src/devices/bus/technics/kn5000_extension.h",
+		MAME_DIR .. "src/devices/bus/technics/hdae5000.cpp",
+		MAME_DIR .. "src/devices/bus/technics/hdae5000.h",
+	}
+end
```
cuavas, line 2592: You want to trigger this on inclusion of the slot header, not one specific card.

src/devices/bus/technics/hdae5000.cpp
```diff
@@ -0,0 +1,110 @@
...
+// HD-AE5000, Hard Disk & Audio Extension for Technics KN5000 emulation
+//
+// The HD-AE5000 was an extension board for the Technics KN5000 musical keyboard.
+// It provided a hard-disk, additional audio outputs and a serial port to interface
+// with a computer to transfer files to/from the hard-drive.
+
+#include "emu.h"
+#include "bus/ata/hdd.h"
+#include "hdae5000.h"
+#include "machine/i8255.h"
```
cuavas, line 13: The modules own header should come immediately after the prefix header, before headers for other things it uses.

src/devices/bus/technics/hdae5000.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+void hdae5000_device::io_map(address_map &map)
+{
+	//map(0x130000, 0x13ffff).m(m_hddc, FUNC(?_device::?)); // Hard-drive Controller (model?) IC? on HD-AE5000 board
+	//map(0x160000, 0x16ffff) ... Optional parallel port interface (NEC uPD71055) IC9
+	map(0x160000, 0x160000).lrw8([this](offs_t a) { return m_ppi->read(0); }, "ppi_r0", [this](offs_t a, u8 data) { m_ppi->write(0, data); }, "ppi_w0");
+	map(0x160002, 0x160002).lrw8([this](offs_t a) { return m_ppi->read(1); }, "ppi_r1", [this](offs_t a, u8 data) { m_ppi->write(1, data); }, "ppi_w1");
+	map(0x160004, 0x160004).lrw8([this](offs_t a) { return m_ppi->read(2); }, "ppi_r2", [this](offs_t a, u8 data) { m_ppi->write(2, data); }, "ppi_w2");
+	map(0x160006, 0x160006).lrw8([this](offs_t a) { return m_ppi->read(3); }, "ppi_r3", [this](offs_t a, u8 data) { m_ppi->write(3, data); }, "ppi_w3");
```
cuavas, line 75: This would be clearer and less error-prone if you make helpers like:
```c++
	template <unsigned Offset> u8 ppi_r() { return m_ppi->read(Offset); }
	template <unsigned Offset> void ppi_w(u8 data) { m_ppi->write(Offset, data); }
```

Then installed it like:
```c++
	map(0x160000, 0x160000).rw(FUNC(hdae5000_device::ppi_r<0>), FUNC(hdae5000_device::ppi_w<0>));
	map(0x160002, 0x160002).rw(FUNC(hdae5000_device::ppi_r<1>), FUNC(hdae5000_device::ppi_w<1>));
	map(0x160004, 0x160004).rw(FUNC(hdae5000_device::ppi_r<2>), FUNC(hdae5000_device::ppi_w<2>));
	map(0x160006, 0x160006).rw(FUNC(hdae5000_device::ppi_r<3>), FUNC(hdae5000_device::ppi_w<3>));
```

But taking a step back, is this a case of installing an 8-bit device on half the width of a 16-bt bus?  If it is, that’s what `umask` is for:
```c++
	map(0x160000, 0x160007).umask(0x00ff).rw(m_ppi, FUNC(i8255_device::read), FUNC(i8255_device::write));
```

src/tools/unidasm.cpp
```diff
@@ -637,10 +637,13 @@ static const dasm_table_entry dasm_table[] =
...
 	{ "t11",             le,  0, []() -> util::disasm_interface * { return new t11_disassembler; } },
 	{ "tlcs870",         le,  0, []() -> util::disasm_interface * { return new tlcs870_disassembler; } },
-	{ "tlcs900",         le,  0, []() -> util::disasm_interface * { return new tlcs900_disassembler; } },
 	{ "tmp90c051",       le,  0, []() -> util::disasm_interface * { return new tmp90c051_disassembler; } },
 	{ "tmp90840",        le,  0, []() -> util::disasm_interface * { return new tmp90840_disassembler; } },
 	{ "tmp90844",        le,  0, []() -> util::disasm_interface * { return new tmp90844_disassembler; } },
+	{ "tmp94c241",       le,  0, []() -> util::disasm_interface * { return new tmp94c241_disassembler; } },
+	{ "tmp95c061",       le,  0, []() -> util::disasm_interface * { return new tmp95c061_disassembler; } },
+	{ "tmp95c063",       le,  0, []() -> util::disasm_interface * { return new tmp95c063_disassembler; } },
+	{ "tmp96c141",       le,  0, []() -> util::disasm_interface * { return new tmp96c141_disassembler; } },
```
cuavas, line 646: I think there’s still potential value in having a generic TLCS900 core disassembler without specifying a specific device for the purpose of investigating code when the exact device it targets is unknown or unsupported.

src/devices/cpu/tlcs900/dasm900.cpp
cuavas, line 1: As I said last time, the approach you’ve taken to this is error-prone.
* Needing to fill unused spaces with the actual hex is liable to lead to copy/paste errors.
* Needing to supply the number of SFRs separately rather than letting the code infer it is error-prone.
* You don’t need a separate class for each of these disassemblers – it bloat the code unnecessarily.  The SFR definitions can be supplied on construction.

You need to:
* Just allow the disassembler to receive configuration on construction.  See the SPARC or M6805 disassemblers for real-world examples.
* Have some way of allowing the number of SFR definitions to be inferred (template on array length, initialiser list, or something).
* Either allow `nullptr` in the SFR definitions for unused locations, or switch to using (address, name) pairs like the M6805 disassembler does to avoid the need to have unused locations in the definitions.

src/devices/cpu/tlcs900/dasm900.cpp
```diff
@@ -1369,12 +1374,260 @@ const char *const tlcs900_disassembler::s_cond[16] =
...
+	case 0x4e: util::stream_format(stream, " DMAM3"); break;
+	case 0x52: util::stream_format(stream, " DMAM4"); break;
+	case 0x56: util::stream_format(stream, " DMAM5"); break;
+	case 0x5a: util::stream_format(stream, " DMAM6"); break;
+	case 0x5e: util::stream_format(stream, " DMAM7"); break;
+	default:
+		util::stream_format(stream, " unknown (encoding is 0x%02x)", imm);
+		break;
+	}
+}
```
cuavas, line 1551: The control register definitions should be supplied as configuration on construction as well to avoid the need for these virtual member functions.  You can use (imm, name) pairs like the M6805 disassembler does for SFRs.

Also, the disassembly would probably be more readable if you just used the value for unknown/invalid values, rather than the extra descriptive text.

src/devices/cpu/tlcs900/dasm900.cpp
```diff
@@ -1449,25 +1702,46 @@ offs_t tlcs900_disassembler::disassemble(std::ostream &stream, offs_t pc, const
...
 		switch( op & 0x07 )
 		{
-		case 0x00:  /* 0xC0 */
+		case 0x00:  /* 0xc0 */
 			imm = opcodes.r8( pos++ );
-			buf = string_format("0x%02x", imm);
+			if( imm < m_num_sfr )
+			{
+				buf = string_format("%s", m_sfr_names[imm]);
+			}
```
cuavas, line 1710: You don’t need `string_format` for unformatted strings, `<<` is a thing.

src/devices/cpu/tlcs900/tmp94c241.h
```diff
@@ -0,0 +1,419 @@
...
+#define TMP94C241_INTETC45 14
+#define TMP94C241_INTETC67 15
+#define TMP94C241_INTE0AD  16
+#define TMP94C241_INTNMWDT 17
+
+// Flip-Flop operations:
+#define FLIPFLOP_INVERT   0b00
+#define FLIPFLOP_SET      0b01
+#define FLIPFLOP_CLEAR    0b10
+#define FLIPFLOP_DONTCARE 0b11
```
cuavas, line 39: Please avoid polluting the global namespace in headers, and particularly avoid unnecessary use of preprocessor macros.

src/devices/cpu/tlcs900/tmp94c241.h
```diff
@@ -0,0 +1,419 @@
...
+	auto portf_write() { return m_portf_write.bind(); }
+	auto portg_read()  { return m_portg_read.bind(); }
+	auto porth_read()  { return m_porth_read.bind(); }
+	auto porth_write() { return m_porth_write.bind(); }
+	auto portz_read()  { return m_portz_read.bind(); }
+	auto portz_write() { return m_portz_write.bind(); }
+
+protected:
+	// device-level overrides
+	virtual void device_config_complete() override;
```
cuavas, line 92: Minor thing, but please don’t copy/paste this comment – it doesn’t even make sense.  Use “device_t implementation” or similar.

src/devices/cpu/tlcs900/tmp94c241.h
```diff
@@ -0,0 +1,419 @@
...
+	devcb_read8    m_port0_read;
+	devcb_write8   m_port0_write;
+
+	// Port 1: 8 bit I/O. Shared with d8-d15
+	devcb_read8    m_port1_read;
+	devcb_write8   m_port1_write;
+
+	// Port 2: 8 bit I/O. Shared with d16-d23
+	devcb_read8    m_port2_read;
+	devcb_write8   m_port2_write;
```
cuavas, line 303: Can you use a `devcb_read8::array` and `devcb_write8::array` for these to make it more concise?

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,2273 @@
...
+void tmp94c241_device::p1_w(uint8_t data)
+{
+	m_port_latch[1] = data;
+	m_port1_write(0, data, 0xff);
+}
+
+void tmp94c241_device::p1cr_w(uint8_t data)
+{
+	m_port_control[1] = data;
+}
```
cuavas, line 427: Can you use function templates to avoid copy/pasta?

cuavas (changes requested):
Thanks for updating this to build against the current codebase.  This seems to still have a number of the same issues as the previous pull request.

cuavas:
I cribbed a small part of this in 23e9abc97c8923f1d7e948a68dc2896084972d29 as a release is coming very soon.

scripts/src/bus.lua
felipesanches: You mean it should be like this, instead?

`--@src/devices/bus/technics/kn5000_extension.h,BUSES["TECHNICS"] = true`

src/devices/bus/technics/hdae5000.cpp
felipesanches: What do you mean by "the prefix header"? Are you referring to the "emu.h" one?

scripts/src/bus.lua
cuavas: Yes.  Look at any of the other optional buses.

Think about this logically:
* You need to trigger it for anything that may instantiate the slot, as that will require it whether any cards are used or not.
* The card headers `#include` the slot header anyway, so something trying to use a card without the slot would still trigger it.
* The slot header does not `#include` the card headers, so triggering on a card header won’t work if that particular card is not used.

src/devices/bus/technics/hdae5000.cpp
cuavas: https://docs.mamedev.org/contributing/cxx.html#structural-organization

src/devices/cpu/tlcs900/tmp94c241.h
felipesanches: Here I am following the existing code-style of the other TLCS900 variants in MAME:
https://github.com/mamedev/mame/blob/9c36794ded21e8869b50041981676b69f4a64412/src/devices/cpu/tlcs900/tmp95c061.cpp#L152-L162

felipesanches: Do you accept the usage of the same code-style as the other TLCS900 variants already in MAME to be used in this pull request as well?

If that's not acceptable, then what would be your suggestion of a different approach to name these MCU registers in the source?

felipesanches: @cuavas, what about using `constexpr u8 TMP95C061_INTE45 = 0x71;` here?

felipesanches: Or even just `constexpr u8 INTE45 = 0x71;` (without the prefix indicating the specific CPU model)

felipesanches:
This one is finally done (via a larger number of smaller PRs merged separately).

PR #12692, opened 2024-08-26
Add a newer Diet Go Go set (with alt graphics)

clawgrip:
New working systems
-------------------
Diet Go Go (Europe v1.1 1992.09.26 v4) [Tailsnic Retroworks]

### DIFF
--- a/src/mame/dataeast/dietgo.cpp
+++ b/src/mame/dataeast/dietgo.cpp
@@ -92,7 +92,7 @@ uint32_t dietgo_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap
 	m_sprgen->set_flip_screen(BIT(flip, 7));
 	m_deco_tilegen->pf_update(m_pf_rowscroll[0], m_pf_rowscroll[1]);
 
-	bitmap.fill(256, cliprect); /* not verified */
+	bitmap.fill(256, cliprect); // not verified
 
 	m_deco_tilegen->tilemap_2_draw(screen, bitmap, cliprect, TILEMAP_DRAW_OPAQUE, 0);
 	m_deco_tilegen->tilemap_1_draw(screen, bitmap, cliprect, 0, 0);
@@ -228,6 +228,15 @@ static INPUT_PORTS_START( dietgo )
 	PORT_DIPSETTING(      0x0000, DEF_STR( On ) )
 INPUT_PORTS_END
 
+static INPUT_PORTS_START( dietgon ) // New version with alt graphics
+	PORT_INCLUDE( dietgo )
+
+	PORT_MODIFY("DSW")
+	PORT_DIPNAME( 0x2000, 0x2000, "Alternative graphics" )
+	PORT_DIPSETTING(      0x2000, DEF_STR( Off ) )
+	PORT_DIPSETTING(      0x0000, DEF_STR( On ) )
+INPUT_PORTS_END
+
 static const gfx_layout tile_8x8_layout =
 {
 	8,8,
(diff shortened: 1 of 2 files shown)


mamehaze:
all the European sets will attempt to display the alt graphics on the title screen if you flip the same switch

the problem is they're not in the gfx ROMs for the other Euro sets.

I suspect this is actually a case of the other sets being partial dumps, and it was assumed they used the same video ROMs as the Jp and US versions when in reality they should have these updated ROMs.

clawgrip:
We found the new gfx ROMs on two "Euro V1.1 (v3)" PCBs, but we also have other Euro V1.1 v2/v1 PCBs with MAY-01 and MAY-02 gfx, and factory soldered.
I'd say we should merge V3 and V4 as the same set with the new gfx ROMs, but leaving the others as they are.

mamehaze:
where is 'v3' and 'v4' even coming from? are those official version numbers?

clawgrip:
I'd say it's "set 1", "set 2", etc.

mamehaze:
right, if those aren't official version numbers from the board I'd just merge the first 2 sets then, as only that graphic ROM differs.

it's odd that all the Euro sets upload the data needed for this alt title screen, read the dipswitches, and attempt to draw the sprites (which just happen to be blank tiles in the older GFX ROM revision) if they weren't made with it in mind, but if the older PCBs have been seen without the updated graphics, then yeah, leave those with the old one.

all the Euro sets should have the dipswitch at least though, as the effect is known, it changes the title screen, to either the new graphics, or blank tiles where the new graphics would be depending on which GFX ROM is installed.

src/mame/dataeast/dietgo.cpp
```diff
@@ -322,8 +331,31 @@ void dietgo_state::dietgo(machine_config &config)
...
+	ROM_LOAD( "may-03.11l", 0x00000, 0x80000, CRC(b6e42bae) SHA1(c282cdf7db30fb63340cc609bf00f5ab63a75583) )
+
+	ROM_REGION( 0x0600, "plds", 0 )
+	ROM_LOAD( "pal16l8b_vd-00.6h",  0x0000, 0x0104, NO_DUMP ) // PAL is read protected
+	ROM_LOAD( "pal16l8b_vd-01.7h",  0x0200, 0x0104, NO_DUMP ) // PAL is read protected
+	ROM_LOAD( "pal16r6a_vd-02.11h", 0x0400, 0x0104, NO_DUMP ) // PAL is read protected
+ROM_END
 
-ROM_START( dietgo ) // same version 1.1 and same date as dietgoe but newer version in rom labels
+ROM_START( dietgoe ) // same version 1.1 and same date as dietgoe but newer version in ROM labels
```
cuavas, line 358: This comment doesn’t make sense any more – “same version 1.1 and same date as dietgoe” when it *is* `dietgoe`.

Please check the rest of the file for things that refer to the different systems by short name when you change short names.

happppp:
Yeah, it looks like the western style title screen gfx works on all European versions. Does it make sense to (ab)use a BIOS setting like so?: https://github.com/mamedev/mame/commit/83f14ed4288d4968c125c460e1764ea2a5f1bfb7

mamehaze:
> Yeah, it looks like the western style title screen gfx works on all European versions. Does it make sense to (ab)use a BIOS setting like so?: [83f14ed](https://github.com/mamedev/mame/commit/83f14ed4288d4968c125c460e1764ea2a5f1bfb7)

I'm not really keen on the BIOS abuse personally

happppp:
Sure, I reverted it.
I'm 50/50 on it for this driver. But here's a more extreme example, where not having a bios setting would result in a quadruple amount of drivers (72 instead of 18): https://github.com/mamedev/mame/blob/master/src/mame/fidelity/elite.cpp

clawgrip:
Is still needed to remove "dietgoe" and assume that the later European ROMs should always use the new sprite ROMs?

happppp:
If jy_00-3+jy_01-3 with may-01+may-02 does not exist

clawgrip:
> If jy_00-3+jy_01-3 with may-01+may-02 does not exist

Cannot tell for sure, I'll leave it as is then.

Thanks!!!!

PR #12694, opened 2024-08-27
sega/y2.cpp - higurashi: Revert the word Jong back to Jan.

mikehyrus:
This change is to be close to true to the Japanese name possible.

### DIFF
--- a/src/mame/sega/y2.cpp
+++ b/src/mame/sega/y2.cpp
@@ -107,7 +107,7 @@ ROM_START( kof2002umj )
 	ROM_LOAD( "nand08gw3b2cn6.u102", 0x00000000, 0x42000000, CRC(db931dca) SHA1(1b1fc88732944e9ede09e584c7b07e28a59df3e2) )
 ROM_END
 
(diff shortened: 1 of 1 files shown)


mikehyrus:
As also heard about 0:36 in this video.
https://www.youtube.com/watch?v=5zs9JRynJFs

cuavas:
The thing is, we tend to use standard spelling for loan words, and “麻雀” isn’t a Japanese word as such.  The etymology is rather complicated.

The game was originally called “麻雀” in Chinese, read “máquè” in Mandarin or “maa zoek” in Cantonese.  Game sets exported to America in the 1940s were sold as “Mah-Jongg”.  Over time, this turned became “mahjong” in English, and “麻將” in Chinese (read as “májiàng in Mandarin).

In Japanese, the game is called “麻雀”, read as “maajan”, but this is a case of taking the English name used for the game while using the original Chinese characters to write it.  The character “雀” isn’t read as “jan” when normal rules apply.  The kun (Japanese) reading is “suzume), and the on (Chinese) readings are “jaku”, “shaku” and “saku”.  It’s only read as “jan” in the loanword “mahjong” and other terms derived from it.

So by regular MAME rules, since “麻雀” is a loanword from the English “mahjong”, it should be transliterated as “mahjong”, and the abbreviation “雀” should be transliterated as “jong”.

PR #12696, opened 2024-08-27
sega/y2.cpp - higurashi: Include 07th Expansion in the company side.

mikehyrus:
Good question whether or not it stays as is on this one or change it to (07th Expansion license)?


cuavas:
I don’t support adding owners of licensed intellectual property to all the manufacturers.  For example the 1991 Batman arcade game just has “Atari Games” in the manufacturer field, with no mention of DC Comics and Warner Bros.  Adding owners of licensed IP to everything is just clutter.

mikehyrus:
Because 07th Expansion has made novels on this series.

PR #12702, opened 2024-08-29
new WORKING machine (DDRGame)

mamehaze:
new WORKING machine
--------
16-bit TV Dance Pad with 15 songs (DDRGame) [David Haywood, Sean Riddle]

### DIFF
--- a/src/mame/tvgames/spg2xx.cpp
+++ b/src/mame/tvgames/spg2xx.cpp
@@ -1334,6 +1334,28 @@ static INPUT_PORTS_START( virtten )
 	PORT_MODIFY("P3")
 INPUT_PORTS_END
 
+
+static INPUT_PORTS_START( ddr33v )
+	PORT_INCLUDE( spg2xx )
+
+	PORT_MODIFY("P1")
+	PORT_BIT( 0x0001, IP_ACTIVE_HIGH, IPT_JOYSTICK_UP ) PORT_16WAY
+	PORT_BIT( 0x0002, IP_ACTIVE_HIGH, IPT_JOYSTICK_LEFT ) PORT_16WAY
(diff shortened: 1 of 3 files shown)


src/mame/tvgames/spg2xx.cpp
```diff
@@ -1851,6 +1873,26 @@ void spg2xx_game_lexiart_state::lexiart(machine_config &config)
...
+{
+	SPG24X(config, m_maincpu, XTAL(27'000'000), m_screen);
+	m_maincpu->set_addrmap(AS_PROGRAM, &spg2xx_game_ddr33v_state::mem_map_4m);
+
+	spg2xx_base(config);
+
+	m_maincpu->porta_in().set(FUNC(spg2xx_game_ddr33v_state::base_porta_r));
+	m_maincpu->portb_in().set(FUNC(spg2xx_game_ddr33v_state::base_portb_r));
+	m_maincpu->portc_in().set(FUNC(spg2xx_game_ddr33v_state::base_portc_r));
+}
```
cuavas, line 1893: Isn’t this identical to `spg2xx_game_state::spg2xx`?

mamehaze: well yes, but I think you're just showing ignorance of the reverse engineering process again.

additions here start by copying the boilerplate, this makes it easy to customize / research - in the case of this one, whatever the thing I've ended up patching out is, but ultimately after trying a few things I decided on a patch for now and the config just ended up being the same.

it's was left a timesaver for whoever comes next, but fine, it can go.

MooglyGuy:
Just let it through. The SPG2xx drivers are uniformly not great, and it's not Haze's responsibility to address a broader underlying implementation deficiency. Let him be good at what he's good at - RE - and let me be good at what I'm good at - code cleaup.

mamehaze:
@cuavas feedback was addressed, but this has now dropped off the first page of pull requests, should be a simple merge?

PR #12703, opened 2024-08-30
irem/olibochu.cpp: Tweaked refresh rate (~62hz), misc. cleanups (no functional change)

jackson2k2:


happppp:
Unless the driver style is an inconsistent mess, or you've done enough changes that you can claim you're one of the main copyright holders, don't go randomly changing the whole code style of the driver.

src/mame/irem/olibochu.cpp
```diff
@@ -298,41 +299,48 @@ u32 olibochu_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, c
...
-
-/***************************************************************************
-    Sound
-***************************************************************************/
-
+/*************************************
+ *
+ *  Audio hardware
+ *
+ *************************************/
```
happppp, line 323: Was somethingw wrong with header comments like that?

src/mame/irem/olibochu.cpp
```diff
@@ -503,91 +510,128 @@ static INPUT_PORTS_START( punchkid )
...
+  OBC15
+  OBC16
+
+  HC3-55536    8910
 
-/***************************************************************************
-    ROM Definitions
-***************************************************************************/
 
+CPU M-47A-A:
```
happppp, line 606: Like I said in another PR: If the notes are at the top of the driver, you don't have to more game notes to the ROM defs, they're fine in the comment block at the top.

src/mame/irem/olibochu.cpp
```diff
@@ -298,41 +299,48 @@ u32 olibochu_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, c
 
 TIMER_DEVICE_CALLBACK_MEMBER(olibochu_state::scanline)
 {
-	int scanline = param;
+	// see TODO
+	int const scanline = param;
 
-	if (scanline == 248) // vblank irq
-		m_maincpu->set_input_line_and_vector(0, HOLD_LINE, 0xd7); // Z80 - RST 10h
+	if (scanline == 248) // VBLANK
```
happppp, line 305: Are you sure that sound and main IRQs are triggered at the same time? Does it introduce a chance of main-sound comms going wrong?

src/mame/irem/olibochu.cpp
```diff
@@ -369,18 +377,17 @@ void olibochu_state::cvsd_tick(int state)
...
+ *
+ *************************************/
 void olibochu_state::main_map(address_map &map)
 {
 	map(0x0000, 0x7fff).rom();
 	map(0x8000, 0x83ff).ram().w(FUNC(olibochu_state::videoram_w)).share(m_videoram);
 	map(0x8400, 0x87ff).ram().w(FUNC(olibochu_state::colorram_w)).share(m_colorram);
-	map(0x9000, 0x901f).writeonly().share(m_spriteram[1]);
-	map(0x9020, 0x903f).nopw(); // discard?
+	map(0x9000, 0x903f).writeonly().share(m_spriteram[1]);
```
happppp, line 390: Is 9020-903f used? I'm the one that did that part and it wasn't on a whim.

src/mame/irem/olibochu.cpp
```diff
@@ -235,61 +247,50 @@ void olibochu_state::video_start()
 	m_bg_tilemap->set_transparent_pen(0);
 }
 
-void olibochu_state::draw_sprites(bitmap_ind16 &bitmap, const rectangle &cliprect)
+void olibochu_state::draw_sprites(bitmap_ind16 &bitmap, rectangle const &cliprect)
 {
-	// 8x8 sprites
-	for (int offs = m_spriteram[0].bytes() - 4; offs >= 0; offs -= 4)
+	auto draw = [this](bitmap_ind16 &bitmap, rectangle const &cliprect, int const slot, bool const large)
```
happppp, line 252: Why not simply a standard C++ function draw_sprite_group(...)?

src/mame/irem/olibochu.cpp
```diff
@@ -235,61 +247,50 @@ void olibochu_state::video_start()
...
-		int flipx = attr & 0x40;
-		int flipy = attr & 0x80;
-		int sy = src[2];
-		int sx = src[3];
-
+		int const bank = large ? 1 : 0;
+		int const offs = slot * 4;
+		u8 const code  = m_spriteram[bank][offs + 0];
+		u8 const attr  = m_spriteram[bank][offs + 1];
+		u8 const color = BIT(attr, 0, 6);
```
happppp, line 258: IMO for simple masks like that, you don't need the 3-param BIT macro, just use x & 0x3f.

happppp:
As for the question renaming the driver to m47: I'd use common sense, if the driver has only 1 game (and in this case it is), it doesn't make sense to rename the driver to the PCB hw product number.

src/mame/irem/olibochu.cpp
```diff
@@ -1,201 +1,220 @@
...
+	// initialization
 	virtual void machine_start() override;
 	virtual void machine_reset() override;
 	virtual void video_start() override;
 
 private:
+	// constants
+	static inline XTAL constexpr MASTER_CLOCK = 18.432_MHz_XTAL;
+	static inline u16 constexpr HTOTAL = 384, HBSTART = 256, HBEND = 0; // see TODO
+	static inline u16 constexpr VTOTAL = 256, VBSTART = 248, VBEND = 8;
```
happppp, line 99: You say probable raw screen params, I don't think VTOTAL of 256 is likely. If I wanted to take a guess (and I rather not want to blindly guess raw screen params): 18342000/3/(262*384)

mamehaze:
> Unless the driver style is an inconsistent mess, or you've done enough changes that you can claim you're one of the main copyright holders, don't go randomly changing the whole code style of the driver.

I don't think the changes here are necessarily bad though, a lot of more tightly defined types etc.

happppp:
> I don't think the changes here are necessarily bad though, a lot of more tightly defined types etc.

Your reply is after he force pushed an update (can't see the previous diff)

happppp:
I re-added the 62.5Hz params, since I think the chance reasonable it was correct anyway.
FYI reviewing is harder if you force-push changes, it hides the incremental diffs.

holub:
> I re-added the 62.5Hz params, since I think the chance reasonable it was correct anyway. FYI reviewing is harder if you force-push changes, it hides the incremental diffs.

Rule of thumb: don't force push after review/collaboration is started

PR #12705, opened 2024-08-30
igs/igs_m027.cpp: confirm lhzb3 uses the same internal code as lhdmg;…

HappyASR:
… remove tilemaps rom scrambling in some games


mamehaze:
@HappyASR 
looks like some of the Mahjong games don't need
ROM_LOAD16_WORD_SWAP
on the tilemap ROM, but should just use
ROM_LOAD

(A lot of the fruit machine/slots style ones do need WORD_SWAP though, which is why I changed them over earlier, but it seems like a few of the Mahjong ones need changing back to give good tilemap colours)

mamehaze:
@HappyASR   https://github.com/mamedev/mame/pull/12708
feel free to integrate into your own changes, but as it was my mistake in the first place changing some of these, I felt I should change them back

PR #12711, opened 2024-08-31
igs_m027.cpp - added internal ROM for lhzb4 (now boots) [XingXing]

mamehaze:
- added internal IGS027A ROMs for lhzb4, oceanpar, triplfev, extradrw [Ioannis Bampoulas, Team Europe, XingXing]
- added chessc2 IGS027A internal ROM [Siftware]
- example of using uploaded XOR table for decrypting some of the games

new clones
----------
Fruit Paradise (V206US) [Ioannis Bampoulas]


mamehaze:
still need the correct label from the sticker, but video ROM loading / decryption are also correct on this one

mamehaze:
I've also changed 2 of the sets to use the XOR table supplied by the internal ROM, for any set with a real 027A dump this can be done too (won't work for a bunch of the PGM games, as that part of the boot sequence was in the 'execute only' area that wasn't dumped at the time, and is bypassed with a bootstrap)

mamehaze:
note you have to disable the XA to get these shots as there's no XA dump in the current set so runs off the rails (still don't fully understand the function of the XA on these ones either)

mamehaze:
@HappyASR are you able to see what Extra Draw needs to boot, the encryption is clearly different (Peter said it's an 8-bit ROM, so maybe it's the same thing, but applies differently, with only the extra xor but with a different step due to 8-bit/16-bit difference, no xor table, as the internal ROM never even uploads a complete one)

Also I got a little stuck trying to hook up the XA on Crazy Bugs, it's mostly like Fearless, but it seems to need the proper interrupt source flags / status or it just ends up sending unwanted data to the XA write ports, so again would appreciate some help there if possible.

HappyASR:
@mamehaze 8-bit mode encryption does not require an XOR table, but the game will get stuck at the RESET screen due to the lack of timer support.

```cpp
void extradrw_decrypt(running_machine &machine)
{
	auto const src = util::little_endian_cast<u16>(reinterpret_cast<u32 *>(machine.root_device().memregion("user1")->base()));

	int const rom_size = 0x80000;

	for (int i = 0; i < rom_size / 2; i++)
	{
		uint16_t x = 0;

		IGS27_CRYPT1
		IGS27_CRYPT2_ALT
		IGS27_CRYPT3_ALT2
		IGS27_CRYPT4
		IGS27_CRYPT5
		IGS27_CRYPT6
		IGS27_CRYPT7
		IGS27_CRYPT8_ALT

		x |= (x << 8);
		src[i] ^= x;
	}
}
```
and rominfo(not sure):
```cpp
ROM_START( extradrw ) // IGS PCB 0326-05-DV-1
	ROM_REGION( 0x04000, "maincpu", 0 )
	// Internal rom of IGS027A ARM based MCU
	ROM_LOAD( "e1_igs027a", 0x00000, 0x4000, CRC(ebbf4922) SHA1(d2d196756317523db650bfe9e4bf2aa243e87a00) ) // has a 'E1' sticker

	ROM_REGION32_LE( 0x80000, "user1", 0 ) // external ARM data / prg?
	ROM_LOAD( "u21", 0x00000, 0x80000, CRC(c1641b14) SHA1(bd2525a5b38d4d8a39e99e43ef62e1d2fd3c044d) ) // 1ST AND 2ND HALF IDENTICAL, label not readable

	ROM_REGION( 0x80000, "igs017_igs031:tilemaps", 0 )
	ROM_LOAD( "igs m3001.u4",  0x000000, 0x080000, CRC(d161f8f7) SHA1(4b495197895fd805979c5d5c5a4b7f07a68f4171) ) // label barely readable

	ROM_REGION( 0x280000, "igs017_igs031:sprites", 0 )
	ROM_LOAD( "u12",0x000000, 0x200000, CRC(642247fb) SHA1(69c01c3551551120a3786522b28a80621a0d5082) ) // 1xxxxxxxxxxxxxxxxxxxx = 0xFF, label not readable
	ROM_LOAD( "u3", 0x200000, 0x80000, CRC(97227767) SHA1(c6a1916c0df1aceafbd488ecace5794390058c49) ) // FIXED BITS (xxxxxxx0xxxxxxxx), label not readable

	ROM_REGION( 0x200000, "oki", 0 )
	ROM_LOAD( "igs s3002.u18", 0x00000, 0x200000, CRC(48601c32) SHA1(8ef3bad80931f4b1badf0598463e15508602f104) ) // BADADDR   --xxxxxxxxxxxxxxxxxxx
ROM_END
```

src/mame/igs/igs_m027.cpp
```diff
@@ -114,9 +146,27 @@ void igs_m027_state::machine_start()
...
+
+void igs_m027_state::xor_write(offs_t offset, u8 data)
+{
+	m_xortable[offset] = data;
+
+	if (offset == 0xff) // might not be the trigger
+	{
+		do_decrypt();
+	}
 }
```
cuavas, line 170: The current implementation of this will produce unexpected behaviour with saved states:

If you start the system, let it upload an XOR table, then load a saved state where the XOR table wasn’t decrypted, it won’t restore the initial empty state of the decrypted `user1` region.

You only re-run the decryption on the last byte of the XOR table being written, but on loading a saved state, you’ll run the decryption with whatever the immediate contents of the table is.  If the game writes to some earlier position in the table after writing the final byte, you’ll end up with different decrypted content to what you had when the saved state was created.

To avoid the issues, it might be easier to just use `.readonly().share(...)` for the `user1` ROM in the address map, use an object finder to get a pointer to it, and pass that to the decryption function.  That way the memory system will just save and restore the decrypted content for you and you won’t get these kinds of issues.

src/mame/igs/pgmcrypt.cpp
```diff
@@ -1033,29 +1033,10 @@ void hauntedh_decrypt(running_machine &machine)
...
-	0x9d, 0x91, 0xad, 0xda, 0x13, 0x0a, 0x16, 0x86, 0x41, 0x18, 0x08, 0x01, 0xef, 0x97, 0x11, 0x1f,
-	0x1a, 0xe7, 0x0c, 0xc9, 0x6d, 0x9d, 0xb9, 0x49, 0x0b, 0x6b, 0x9e, 0xd4, 0x72, 0x4d, 0x1d, 0x59
-};
-
-void chessc2_decrypt(running_machine &machine)
+void chessc2_decrypt(running_machine &machine, u8* table)
 {
-	auto const src = reinterpret_cast<u16 *>(machine.root_device().memregion("user1")->base());
+	auto const src = reinterpret_cast<u16 *>(machine.root_device().memregion("user1_encrypted")->base());
+	auto const dst = reinterpret_cast<u16 *>(machine.root_device().memregion("user1")->base());
```
cuavas, line 1039: This is unportable – it won’t work on big Endian hosts because you’re interpreting the regions at a non-native width.  Please use `util::little_endian_cast` to do the `u32` to `u16` cast to make the code portable.  Please fix the issue in functions as you update them.

src/mame/igs/pgmcrypt.cpp
```diff
@@ -1767,34 +1638,17 @@ void mgcs3_decrypt(running_machine &machine)
...
-	0x04, 0x7b, 0x26, 0x6f, 0x2c, 0x18, 0x47, 0xe8, 0xfc, 0x99, 0x42, 0xd3, 0xc9, 0x4c, 0x26, 0x7f,
-	0xd2, 0x4f, 0x40, 0x32, 0x74, 0xb2, 0xe6, 0x6b, 0x90, 0xcf, 0x7f, 0x56, 0x3a, 0xe5, 0xd7, 0x8f
-};
 
-void mgzz_decrypt(running_machine &machine)
+void mgzz_decrypt(running_machine &machine, u8* table)
 {
-	auto const src = util::little_endian_cast<u16>(reinterpret_cast<u32 *>(machine.root_device().memregion("user1")->base()));
+	auto const src = reinterpret_cast<u16 *>(machine.root_device().memregion("user1_encrypted")->base());
+	auto const dst = reinterpret_cast<u16 *>(machine.root_device().memregion("user1")->base());
```
cuavas, line 1651: You’ve introduced an Endianness issue here where the code was previously doing things correctly.

src/mame/igs/pgmcrypt.cpp
```diff
@@ -2266,35 +2065,16 @@ void lthy_decrypt(running_machine &machine)
...
-	0x0d, 0x07, 0x0c, 0x9c, 0x68, 0x24, 0x51, 0x51, 0x1f, 0x8d, 0x8b, 0xd6, 0x2e, 0x67, 0x5f, 0xc3,
-	0x07, 0x00, 0x12, 0x61, 0x77, 0xa8, 0x15, 0xa1, 0xd6, 0xd0, 0xd3, 0x57, 0x73, 0x62, 0xb9, 0xbb
-};
-
-void lhdmg_decrypt(running_machine &machine)
+void lhdmg_decrypt(running_machine &machine, u8* table)
 {
-	auto const src = util::little_endian_cast<u16>(reinterpret_cast<u32 *>(machine.root_device().memregion("user1")->base()));
+	auto const src = reinterpret_cast<u16 *>(machine.root_device().memregion("user1_encrypted")->base());
+	auto const dst = reinterpret_cast<u16 *>(machine.root_device().memregion("user1")->base());
```
cuavas, line 2077: Same with this one – you’ve introduced an Endianness issue here.  There’s another case further down, too.

mamehaze:
I was trying to make improvements to the functionality, but due to nitpicking rather than helping I've reverted this back to just the actual additions.

PR #12713, opened 2024-08-31
Initial support for the Motorola MEX68KECB Educational Computer Board

eschaton:
Initial implementation of the Motorola 68K Educational Computer Board (MEX68KECB).

### DIFF
--- a/src/mame/motorola/mex68kecb.cpp
+++ b/src/mame/motorola/mex68kecb.cpp
@@ -1,0 +1,228 @@
+// license:BSD-3-Clause
+// copyright-holders:Chris Hanson
+/*
+ * mex68kecb.cpp - Motorola MEX68KECB
+ *
+ * Created on: August 31, 2024
+ *     Author: Chris Hanson
+ *
+ * Documentation:
+ *   http://www.bitsavers.org/components/motorola/68000/MEX68KECB/MEX68KECB_D2_EduCompBd_Jul82.pdf
+ *
+ * The Motorola MC68000 Educational Computer Board is a single-board computer with
+ * a 4MHz 68000 CPU, 32KB RAM, 16KB ROM, host and terminal serial ports, a
+ * parallel interface/timer, a cassette interface, and a prototyping area with
+ * full access to the 68000 bus. The ROM contains TUTOR, a debug and bootstrap
+ * system that was the predecessor of MACSBUG.
+ *
+ * Specifications:
+ * - 4MHz MC68000L4 CPU
+ * - MC6850 ACIA x 2
+ * - MC68230 PIT
+ *
+ * To Do:
+ * - Cassette I/O
+ * - Save/Restore
+ *
+ */
+
+#include "emu.h"
+
+#include "bus/rs232/rs232.h"
+#include "cpu/m68000/m68000.h"
+#include "machine/68230pit.h"
+#include "machine/6850acia.h"
+#include "machine/mc14411.h"
+
+
+class mex68kecb_state : public driver_device
+{
+public:
+	mex68kecb_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_maincpu(*this, "maincpu")
+		, m_pit(*this, "pit")
+		, m_brg(*this, "brg")
+		, m_acia1(*this, "acia1")
+		, m_acia2(*this, "acia2")
+		, m_acia1_baud(*this, "ACIA1_BAUD")
+		, m_acia2_baud(*this, "ACIA2_BAUD")
+		, m_terminal(*this, "terminal")
+		, m_host(*this, "host")
+	{ }
+
+	void mex68kecb(machine_config &config);
+
+private:
+	required_device<cpu_device> m_maincpu;
+	required_device<pit68230_device> m_pit;
+	required_device<mc14411_device> m_brg;
+	required_device<acia6850_device> m_acia1;
+	required_device<acia6850_device> m_acia2;
+	required_ioport m_acia1_baud;
+	required_ioport m_acia2_baud;
+
+	required_device<rs232_port_device> m_terminal;
+	required_device<rs232_port_device> m_host;
+
+	virtual void machine_start() override;
+	virtual void machine_reset() override;
+	void mem_map(address_map &map);
+
+	// Clocks from Baud Rate Generator
+	template <u8 bit> void write_acia_clock(int state);
+
+	// Pointer to System ROMs needed by bootvect_r and masking RAM buffer for post reset accesses
+	uint16_t *m_sysrom = nullptr;
+	uint16_t m_sysram[8]{};
+	uint16_t bootvect_r(offs_t offset);
+	void bootvect_w(offs_t offset, uint16_t data, uint16_t mem_mask = ~0);
+};
+
+/* Input ports */
+static INPUT_PORTS_START( mex68kecb )
+	PORT_START("ACIA1_BAUD")
+	PORT_DIPNAME(0xff, 0x80, "Terminal Baud Rate")
+	PORT_DIPSETTING(0x80, "9600") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x40, "4800") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x20, "2400") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x10, "1200") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x08,  "600") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x04,  "300") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x02,  "150") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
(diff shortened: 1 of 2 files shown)


src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,339 @@
...
+	// Reset BRG.
+	m_brg->rsa_w(CLEAR_LINE);
+	m_brg->rsb_w(ASSERT_LINE);
+
+	/* Reset pointer to bootvector in ROM for bootvector handler bootvect_r */
+	if (m_sysrom == &m_sysram[0]) /* Condition needed because memory map is not setup first time */
+		m_sysrom = (uint16_t*)(memregion("roms")->base());
+}
+
+void mex68kecb_state::write_f1_clock(int state)
```
rb6502, line 210: This would be much cleaner as a template function, with the bit number as the template parameter.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,339 @@
...
+#include "bus/rs232/rs232.h"
+#include "cpu/m68000/m68000.h"
+#include "machine/68230pit.h"
+#include "machine/6850acia.h"
+#include "machine/mc14411.h"
+
+
+#include "debugger.h"
+
+#define VERBOSE  1
```
rb6502, line 40: '1' for VERBOSE is LOG_GENERAL.  But also things normally shouldn't be checked in with logging enabled.

Also, I don't see any LOG() or LOGMASKED() usage in this file?

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,339 @@
...
+#define LOG_OUTPUT_FUNC printf
+#include "logmacro.h"
+
+
+#ifdef _MSC_VER
```
rb6502, line 45: FUNCNAME is also not used in this file?

rb6502 (changes requested):
Looks mostly decent.  A few instances of copy-pasting unnecessary things, and one function that would be much cleaner if it was templatized.  I used one of these boards as part of a university class in the early 90s so it's fun to see a driver for it.

src/mame/motorola/mex68kecb.cpp
eschaton: Guess I'll figure out how to do that! (This is the first C++ I've written since 2005...)

rb6502: I never ask people to do something I haven't already done myself :-)

Here's an example of a function that works that way: https://github.com/mamedev/mame/blob/049d10067fbfe79099c79860268b705f5f880ffc/src/mame/apple/iosb.cpp#L324

rb6502: The function itself is fairly straightforward.  The devilish detail is pre-instantiating all of the versions you're going to use, which you can see immediately after the function there at line 356.

src/mame/motorola/mex68kecb.cpp
eschaton: Yeah, I used some logging while I brought it up just to ensure things were what I expected, but it wasn't needed once I was done. Think I should delete the whole section?

eschaton:
OK, I think this has addressed everything.

rb6502 (approved):

rb6502:
Just gonna let the CI builds run as a final sanity check and then we're good.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+	PORT_START("ACIA1_BAUD")
+	PORT_DIPNAME(0xff, 0x80, "Terminal Baud Rate")
+	PORT_DIPSETTING(0x80, "9600") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x40, "4800") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x20, "2400") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x10, "1200") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x08,  "600") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x04,  "300") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x02,  "150") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
+	PORT_DIPSETTING(0x01,  "110") PORT_DIPLOCATION("J10:8,7,6,5,4,3,2,1")
```
cuavas, line 93: You’re supposed to put the location on the `PORT_DIPNAME` line so you aren’t repeating it for every value.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+	// Set up interrupts.
+
+	// Nothing at IRQ1
+	m_pit->timer_irq_callback().set_inputline("maincpu", M68K_IRQ_2);
+	m_pit->port_irq_callback().set_inputline("maincpu", M68K_IRQ_3);
+	// Optional 6800 peripherals at IRQ4
+	m_acia1->irq_handler().set_inputline("maincpu", M68K_IRQ_5);
+	m_acia2->irq_handler().set_inputline("maincpu", M68K_IRQ_6);
+	// ABORT Button at IRQ7
```
cuavas, line 139: You don’t need to use literal tags here.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+void mex68kecb_state::mem_map(address_map &map)
+{
+	map.unmap_value_high();
+	map(0x000000, 0x000007).ram().w(FUNC(mex68kecb_state::bootvect_w));       /* After first write we act as RAM */
+	map(0x000000, 0x000007).rom().r(FUNC(mex68kecb_state::bootvect_r));       /* ROM mirror just during reset */
+	map(0x000008, 0x007fff).ram(); /* 32KB RAM */
+	map(0x008000, 0x00bfff).rom().region("roms", 0); /* 16KB ROM */
+	map(0x010000, 0x01003f).rw("pit", FUNC(pit68230_device::read), FUNC(pit68230_device::write)).umask16(0x00ff);
+	map(0x010040, 0x010043).rw("acia1", FUNC(acia6850_device::read), FUNC(acia6850_device::write)).umask16(0xff00);
+	map(0x010040, 0x010043).rw("acia2", FUNC(acia6850_device::read), FUNC(acia6850_device::write)).umask16(0x00ff);
```
cuavas, line 169: You don’t need to use literal tags here, either.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+void mex68kecb_state::machine_start()
+{
+	/* Setup pointer to bootvector in ROM for bootvector handler bootvect_r */
+	m_sysrom = (uint16_t *)(memregion("roms")->base());
```
cuavas, line 175: You should really use object finders to find resources.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+/* Boot vector handler, the PCB hardwires the first 16 bytes from 0xfc0000 to 0x0 at reset. */
+uint16_t mex68kecb_state::bootvect_r(offs_t offset) {
+	return m_sysrom[offset];
+}
+
+void mex68kecb_state::bootvect_w(offs_t offset, uint16_t data, uint16_t mem_mask) {
+	m_sysram[offset % std::size(m_sysram)] &= ~mem_mask;
+	m_sysram[offset % std::size(m_sysram)] |= (data & mem_mask);
+	m_sysrom = &m_sysram[0]; // redirect all upcoming accesses to masking RAM until reset.
+}
```
cuavas, line 212: This kind of overlay should really use a view.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+ * system that was the predecessor of MACSBUG.
+ *
+ * Specifications:
+ * - 4MHz MC68000L4 CPU
+ * - MC6850 ACIA x 2
+ * - MC68230 PIT
+ *
+ * To Do:
+ * - Cassette I/O
+ * - Save/Restore
```
cuavas, line 25: You can flag unemulated features.

src/mame/motorola/mex68kecb.cpp
```diff
@@ -0,0 +1,228 @@
...
+	if (BIT(m_acia2_baud->read(), bit)) {
+		m_acia2->write_txc(state);
+		m_acia2->write_rxc(state);
+	}
+}
+
+/* Boot vector handler, the PCB hardwires the first 16 bytes from 0xfc0000 to 0x0 at reset. */
+uint16_t mex68kecb_state::bootvect_r(offs_t offset) {
+	return m_sysrom[offset];
+}
```
cuavas, line 206: The comment here doesn’t match the code – the comment says 16 bytes, but the code only overlays the first 8 bytes.  I didn’t fix the contradiction when I cleaned it up, but it should be fixed one way or the other.

eschaton:
Thanks @cuavas for all the advice! I can do some cleanup.

Tafoid:
@eschaton

As of this moment the rom data to support this added machine has not been forwarded to the development team.  All commits/PRs that are intended for addition to the official source and contain new data should have this data sent in as soon as code has been submitted so it can be verified.

If you could, please use the mail listed at this page:   https://wiki.mamedev.org/index.php/Submitting_Source_Code#Submitting_ROM_or_disk_dumps

Thank you!

PR #12717, opened 2024-09-02
konami/konamigx.cpp: Add a note about a common Racin' Force year hack

clawgrip:
Useful to avoid problems when looking for different sets


src/mame/konami/konamigx.cpp
```diff
@@ -3523,7 +3523,11 @@ ROM_START( le2j )
 	ROM_LOAD( "le2j.nv", 0x0000, 0x080, CRC(f6790425) SHA1(f233f3c09c4cdbd1c6e5204fc6554a4826b44c59) )
 ROM_END
 
-/* Racin' Force */
+/* Racin' Force.
+   For reference, there is a common "year hack" version with:
```
angelosa, line 3527: And the fake release year is ...?

PR #12719, opened 2024-09-02
mex68kecb: Add Abort button, minor cleanup

eschaton:
Add an Abort button that triggers an NMI, just like the hardware.

Also fix the boot vector handler's comment to refer to the correct ROM location and the correct number of bytes that are mapped from the ROM.

### DIFF
--- a/src/mame/motorola/mex68kecb.cpp
+++ b/src/mame/motorola/mex68kecb.cpp
@@ -55,6 +55,8 @@ public:
 
 	void mex68kecb(machine_config &config) ATTR_COLD;
 
+	DECLARE_INPUT_CHANGED_MEMBER(abort_button);
+
 protected:
 	virtual void machine_reset() override ATTR_COLD;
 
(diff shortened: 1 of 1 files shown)


src/mame/motorola/mex68kecb.cpp
```diff
@@ -104,6 +107,9 @@ static INPUT_PORTS_START( mex68kecb )
 	PORT_DIPSETTING(0x04,  "300")
 	PORT_DIPSETTING(0x02,  "150")
 	PORT_DIPSETTING(0x01,  "110")
+
+	PORT_START("ABORT")
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_BUTTON1) PORT_NAME("Abort button") PORT_CHANGED_MEMBER(DEVICE_SELF, mex68kecb_state, abort_button, 0)
```
cuavas, line 112: Do you want to give this a different default mapping, as I expect Control is used for typing control characters with the terminal?  You can use `PORT_CODE` to set the default assignment.

eschaton: Aha, excellent. That’ll be handy.

PR #12720, opened 2024-09-02
apple2e: Apple //c mouse fixes

as-tb-dev:
- Initializes Apple //c mouse related variables
- Sets Y0 interrupt flag on //c mouse Y-axis movement even if X-axis movement is still being processed (allowing both the X0 and Y0 interrupt flags to be set at the same time)
- Increases //c mouse port sensitivity (found in apple2e.cpp) from 20 to 40 to match that of the //e mouse (found in a2bus/mouse.cpp)

### DIFF
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -1037,6 +1037,13 @@ void apple2e_state::machine_start()
 	m_joystick_x1_time = m_joystick_x2_time = m_joystick_y1_time = m_joystick_y2_time = 0;
 	m_reset_latch = false;
 
+	last_mx = 0;
+	last_my = 0;
+	count_x = 0;
+	count_y = 0;
+	m_x0 = false;
+	m_y0 = false;
+
 	// setup save states
 	save_item(NAME(m_speaker_state));
 	save_item(NAME(m_cassette_state));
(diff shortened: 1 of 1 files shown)


src/mame/apple/apple2e.cpp
```diff
@@ -1165,6 +1165,14 @@ void apple2e_state::machine_reset()
...
 	m_x0edge = false;
 	m_y0edge = false;
+	last_mx = 0;
+	last_my = 0;
+	count_x = 0;
+	count_y = 0;
+	m_x0 = false;
+	m_y0 = false;
+	m_x1 = false;
+	m_y1 = false;
```
cuavas, line 1175: You shouldn’t be changing these on reset as they’re not registers affected by reset.  They should only be cleared on start.

src/mame/apple/apple2e.cpp
```diff
@@ -2776,12 +2786,17 @@ void apple2e_state::update_iic_mouse()
...
 			}
 		}
 
 		m_y0 = !m_y0;
 	}
+
+	if (raise_mousexy_irq) {
+		raise_irq(IRQ_MOUSEXY);
+		raise_mousexy_irq = false;
+	}
```
cuavas, line 2799: There’s no point clearing the local variable `raise_mousexy_irq` when it won’t be used again.  Also, this file uses Allman brace positioning – please follow the example of the existing code.

src/mame/apple/apple2e.cpp
as-tb-dev: I can move the first 6 from machine_reset to machine_start.  For the last 2, X1/Y1 are returned by soft switches C066, C067, C06E, C06F.  Should they be reset in this case?  If not, I will move those as well.

src/mame/apple/apple2e.cpp
as-tb-dev: Will change.

PR #12722, opened 2024-09-03
vcs_ctrl: add support for Atari Trak-ball

as-tb-dev:
Adds support for the Atari CX22/CX80 Trak-Ball in native trackball mode.

To test: 
mame a800 -ctrl1 trakball missile

Press: Ctrl-T
Press the "Start" key


src/devices/bus/vcs_ctrl/trakball.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+	vcs_joy_return =
+		m_trakballb->read() |
+		((diff_pos[0] > 0) ? 0x01 : 0x00) |
+		((m_last_pos_sent[0] > 0) ? 0x02 : 0x00) |
+		((diff_pos[1] > 0) ? 0x04 : 0x00) |
+		((m_last_pos_sent[1] > 0) ? 0x08 : 0x00);
+
+	return vcs_joy_return;
+}
```
cuavas, line 122: Given the wide variety of machines using the VCS control port, it really isn’t safe to make assumptions about when it’s going to be read.  You really need to use `INPUT_CHANGED_MEMBER` to get notifications when the axes are updated and use timers to generate the quadrature signals.

cuavas (changes requested):

src/devices/bus/vcs_ctrl/trakball.cpp
as-tb-dev: I agree that my first submission was not realistic - mainly due to this line "if (diff_pos[axis]) m_last_pos_sent[axis] = !m_last_pos_sent[axis];" which is affected by the machine-dependent rate that vcr_joy_r() is called.

While working on the updates I just checked-in now, I had initially added INPUT_CHANGED_MEMBER and removed any trackball read() calls from vcr_joy_r(), but noticed that INPUT_CHANGED_MEMBER would only get called about once per 16ms, whereas, on the A800, vcs_joy_r gets invoked several times in 1ms.  Using just the INPUT_CHANGED_MEMBER function for the read() calls makes the trackball less controllable due to a loss of granularity.  Note that I did not add any timers for that test.  I assume the timers are meant to be used if a clock signal is being sent from the home computer to the trackball.  If so, then this is not the case for this device.

The vic20 appears to call vcr_joy_r() at a much lower rate than the a800, so I agree that those cases need to be handled as well.  What I have changed for this commit is:
- relying now on the input device's positional port-bits to determine the X/Y motion values instead of simply toggling the values each time there was movement detected
- added an INPUT_CHANGED_MEMBER so that trackball X/Y directional changes are not missed on systems that do not call vcs_joy_r() in a consistent and/or frequent manner
- kept logic to read and track X/Y movement and directional changes under vcs_joy_r() as well, so that granularity is not lost on systems that call it at much faster rate than INPUT_CHANGED_MEMBER would be called

cuavas: It’s still completely unrealistic, though.  The quadrature outputs are connected directly to the connector, and will update at a rate that depends on the speed you move the trackball.  Reading the port should have no side effects on the connected device as it isn’t strobed in any way.

You only get updates for inputs once for each frame interval for the first emulated screen (limitation of MAME).  To simulate the quadrature output, you need to use a timer to update it at a rate proportional to the accumulated delta value.

The Namco System 22 driver uses a pair of timers for the trackball quadrature signals: https://github.com/mamedev/mame/blob/mame0270/src/mame/namco/namcos22.cpp#L2906

The Mac 128K/512K/Plus updates the mouse quadrature signals while it’s updating the video output, using the video line frequency as a crude timer: https://github.com/mamedev/mame/blob/mame0270/src/mame/apple/mac128.cpp

PR #12723, opened 2024-09-03
MS68K: Marion Systems MS68K

eschaton:
The Marion Systems MS68K was a single-board 68000 system designed by Tom Oberheim in 1985 with a DUART providing two serial ports and (in combination with a latch) a parallel printer port, WD1772 floppy controller, optional NCR53C80 SCSI controller, and up to 512KB of RAM and 128KB of ROM onboard. Also included is an expansion connector providing access to most of the 68000 bus as well as memory refresh signals and a single dedicated interrupt line.

The stock ROM is a monitor that also contains a port of Tiny BASIC 1.2 at 0xF04900 with a loader at 0xF04800. Thus while there is no direct monitor command to start it, a simple GF04800 will start Tiny BASIC (and as usual BYE will exit back to the monitor).

The monitor’s included floppy boot functionality will load the first sector of the disk to 0x800 and jump to it in supervisor mode. There is no SCSI boot functionality built in.

The MS68K was used in a number of college and university computer systems classes, and a variant of Peter Stark’s SK*DOS was available.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -45134,6 +45134,9 @@ mcg85                           //
 mk14                            //
 mk14vdu                         //
 
+@source:trainer/ms68k.cpp
+ms68k                           // Marion Systems MS68K
+
 @source:trainer/thaler.cpp
 ct65                            // 1982 Thaler CT-65
 mps65                           // 1984 Thaler MPS-65
--- a/src/mame/trainer/ms68k.cpp
+++ b/src/mame/trainer/ms68k.cpp
@@ -1,0 +1,327 @@
+// license:BSD-3-Clause
+// copyright-holders:Chris Hanson
+/*
+ * ms68k.cpp - Marion Systems MS68K
+ *
+ * Documentation:
+ *   http://bitsavers.org/pdf/marionSystems/Marion_Systems_MS68K_Single_Board_Computer_Users_Manual.pdf
+ *
+ *  The Marion Systems MS68K was a single-board MC68000 system designed by Tom
+ *  Oberheim in 1985 with two serial ports, a parallel port, WD1772 floppy,
+ *  optional SCSI, and up to 512KB of RAM and ROM each onboard, plus an
+ *  expansion connector providing full access to the 68000 bus. It was used in
+ *  a number of university-level computer systems classes, and a variant of
+ *  Peter Stark’s SK*DOS was also available.
+ *
+ *  The default ROM contains a port of Tiny BASIC 1.2 at 0xF04900, and a
+ *  loader for it at 0xF04800. Thus while there is no direct command to start
+ *  it, a simple GF04800 will start Tiny BASIC with it.
+ *
+ * Specifications:
+ * - 8MHz MC68000 CPU
+ * - 512KB maximum onboard RAM
+ * - 128KB maximum onboard EPROM
+ * - MC68681 DUART
+ * - TTL-supported parallel port
+ * - NCR53C80 SCSI controller
+ * - Expansion bus
+ *
+ * TODO:
+ * - Expansion bus
+ *
+ */
+
+#include "emu.h"
+
+#include "bus/centronics/ctronics.h"
+#include "bus/nscsi/devices.h"
+#include "bus/rs232/rs232.h"
+#include "cpu/m68000/m68000.h"
+#include "imagedev/floppy.h"
+#include "machine/mc68681.h"
+#include "machine/ncr5380.h"
+#include "machine/wd_fdc.h"
+
+
+namespace {
+
+
+class ms68k_state : public driver_device
+{
+public:
+	ms68k_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_bootvect(*this, "bootvect")
+		, m_maincpu(*this, "maincpu")
+		, m_duart(*this, "duart")
+		, m_terminal(*this, "terminal")
+		, m_modem(*this, "modem")
+		, m_fdc(*this, "fdc")
+		, m_floppy(*this, "fdc:%u", 0)
+		, m_floppy_selected_drive(0)
+		, m_scsi(*this, "scsibus:7:ncr5380")
+		, m_scsibus(*this, "scsibus")
+		, m_printer_conn(*this, "prn")
+		, m_printer_out(*this, "prn_out")
+	{ }
+
+	void ms68k(machine_config &config) ATTR_COLD;
+
+	DECLARE_INPUT_CHANGED_MEMBER(nmi_button);
+
+protected:
+	virtual void machine_start() override;
+	virtual void machine_reset() override ATTR_COLD;
+
+private:
+	memory_view m_bootvect;
+
+	required_device<cpu_device> m_maincpu;
+
+	required_device<mc68681_device> m_duart;
+	required_device<rs232_port_device> m_terminal;
+	required_device<rs232_port_device> m_modem;
+
+	required_device<wd1772_device> m_fdc;
+	required_device_array<floppy_connector, 2> m_floppy;
+	uint8_t m_floppy_selected_drive;
+
+	required_device<ncr5380_device> m_scsi;
+	required_device<nscsi_bus_device> m_scsibus;
+
+	required_device<centronics_device> m_printer_conn;
+	required_device<output_latch_device> m_printer_out;
+
+	void mem_map(address_map &map) ATTR_COLD;
+
+	uint8_t duart_r(offs_t offset);
+	void duart_w(offs_t offset, uint8_t data);
+
+	// Floppy
+	static void floppy_types(device_slot_interface &device);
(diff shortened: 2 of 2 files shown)


src/mame/mame.lst
```diff
@@ -25078,6 +25078,9 @@ x37                             //
 @source:makerbot/replicator.cpp
 replica1                        // Makerbot Replicator 1 desktop 3d printer
 
+@source:marion/ms68k.cpp
+ms68k                           // Marion Systems MS68K
+
```
cuavas, line 25083: This shouldn’t really be in a project by itself.  It probably belongs in trainer (miscellaneous development systems and systems for learning computer architecture), homebrew (designs published with the intent of allowing hobbyists to build their own systems), or misc (things that don’t really fit anywhere else).

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+	DECLARE_INPUT_CHANGED_MEMBER(nmi_button);
+
+protected:
+	virtual void machine_reset() override ATTR_COLD;
+
+private:
+	void mem_map(address_map &map) ATTR_COLD;
+
+	memory_view m_bootvect;
```
cuavas, line 78: Please put `mem_map` with the other member functions in the `private` section (either collect the member functions above or below the data members).

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+	required_device<cpu_device> m_maincpu;
+
+	required_device<mc68681_device> m_duart;
+	required_device<rs232_port_device> m_terminal;
+	required_device<rs232_port_device> m_modem;
+
+	required_device<wd1772_device> m_fdc;
+	required_device_array<floppy_connector, 2> m_floppy;
+	int m_floppy_selected_drive;
```
cuavas, line 88: Should `m_floppy_selected_drive` be registered for save states?  If so, it should be an explicitly sized integer (e.g. `uint8_t`, `int16_t` or whatever) for portability, and you should register it in `machine_start` with `save_item`.

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+	// Set up modem RS-232.
+
+	RS232_PORT(config, m_modem, default_rs232_devices, nullptr);
+	m_duart->b_tx_cb().set(m_modem, FUNC(rs232_port_device::write_txd));
+	m_modem->rxd_handler().set(m_duart, FUNC(mc68681_device::rx_b_w));
+	// DUART IP1 = CTS
+	m_modem->cts_handler().set(m_duart, FUNC(mc68681_device::ip1_w));
+	// DUART OP1 = RTS
+	m_duart->outport_cb().append(m_modem, FUNC(rs232_port_device::write_rts)).bit(1);
```
cuavas, line 148: We generally prefer to have devices’ output routing grouped with their instantiation.  Please move the `m_duart` configuration up to group it with the instantiation, before instantiating the RS232 ports.  You can refer to `m_terminal` and `m_modem` before the devices are instantiated.

Also, do you need to configure the default serial settings for the terminal or other devices?  If so, you can do it with `DEVICE_INPUT_DEFAULTS_START`, `set_option_device_input_defaults`, etc. (see src/mame/intel/intellec4.cpp for an example).

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+	// FDC Interrupt = 4
+	m_fdc->intrq_wr_callback().set_inputline(m_maincpu, M68K_IRQ_3);
+	// DUART OP3 = Side 1 Select
+	m_duart->outport_cb().append(FUNC(ms68k_state::floppy_side_w)).bit(3);
+	// DUART OP4 = Drive Select 0
+	m_duart->outport_cb().append(FUNC(ms68k_state::floppy_drive0_w)).bit(4);
+	// DUART OP5 = Drive Select 1
+	m_duart->outport_cb().append(FUNC(ms68k_state::floppy_drive1_w)).bit(5);
+	// DUART OP6 = Double Density
+	m_duart->outport_cb().append(m_fdc, FUNC(wd1772_device::dden_w)).bit(6);
```
cuavas, line 178: Same here, please move the DUART output routing up with the DUART instantiation.

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+	m_printer_conn->perror_handler().set(m_duart, FUNC(mc68681_device::ip3_w));
+	// DUART IP4 = Device Selected
+	m_printer_conn->select_handler().set(m_duart, FUNC(mc68681_device::ip4_w));
+	// DUART IP5 = Acknowledge
+	m_printer_conn->ack_handler().set(m_duart, FUNC(mc68681_device::ip5_w));
+
+	// DUART OP2 = Initialize
+	m_duart->outport_cb().append(m_printer_conn, FUNC(centronics_device::write_init)).bit(2);
+	// DUART OP7 = Data Strobe
+	m_duart->outport_cb().append(m_printer_conn, FUNC(centronics_device::write_strobe)).bit(7);
```
cuavas, line 201: These should go with the DUART instantiation, too.

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+	// Reset pointer to bootvector in ROM for bootvector view
+	m_bootvect.select(0);
+}
+
+/* DUART */
+
+uint8_t ms68k_state::duart_r(offs_t offset)
+{
+	return m_duart->read(offset);
+}
```
cuavas, line 235: You shouldn’t need this trampoline – you should be able to do something like:
```c++
	map(0xe80000, 0xefffff).r(m_duart, FUNC(mc68681_device::read)).w(FUNC(ms68k_state::duart_w)).umask16(0x00ff);
```

src/mame/marion/ms68k.cpp
```diff
@@ -0,0 +1,319 @@
...
+void ms68k_state::floppy_types(device_slot_interface &device) {
+	device.option_add("525sd", FLOPPY_525_SD);
+	device.option_add("525dd", FLOPPY_525_DD);
+	device.option_add("35dd", FLOPPY_35_DD);
+	device.option_add("35hd", FLOPPY_35_HD);
+}
+
+void ms68k_state::floppy_formats(format_registration &fr) {
+	floppy_image_device::default_mfm_floppy_formats(fr);
+}
```
cuavas, line 258: Please be consistent about placement of opening braces of function bodies.  You’ve got them at the end of the line in some cases, and on a line by themselves in other cases.

src/mame/mame.lst
eschaton: Sounds like `trainer` is the best place for it then. Thanks!

src/mame/marion/ms68k.cpp
eschaton: Thanks! The terminal defaults (9600 8N1) work just fine. Should I still specify them?

src/mame/marion/ms68k.cpp
eschaton: Oh, sick, that’s a lot more convenient! Thanks!

src/mame/marion/ms68k.cpp
eschaton: Doh! Yeah, I generally try to keep them on a line by themselves (and use opening brace on the same line for everything that’s not a function body).

src/mame/marion/ms68k.cpp
cuavas: If 9600 8N1 works you don’t need to override it.  The defaults aren’t going to change.

happppp: About that, I think it would be useful to have something like set_option_device_input_defaults("*", ...)
In case of rs232, if the driver expects serial comms to be at 1200 baud for example, it applies to all of the slot options.

eschaton:
All feedback should be addressed now.

src/mame/trainer/ms68k.cpp
```diff
@@ -0,0 +1,327 @@
...
+void ms68k_state::floppy_drive0_w(int state)
+{
+	int drive = (state == 1) ? 0 : 1;
+	floppy_drive_select(drive);
+}
+
+void ms68k_state::floppy_drive1_w(int state)
+{
+	int drive = (state == 1) ? 1 : 1;
+	floppy_drive_select(drive);
```
happppp, line 280: Is that correct? If yes, just omit that 1st line and do select(1)?

eschaton: Doh! Good catch, thanks!

src/mame/trainer/ms68k.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
```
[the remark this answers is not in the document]
eschaton: Yeah, that was the intent…

PR #12726, opened 2024-09-04
Add Linksys PAP2 and other MIPS-X improvements

Octocontrabass:
cpu/mipsx/mipsx.cpp: probably 32-bit addresses
cpu/mipsx/mipsxdasm.cpp: fix various bugs
tomy/kissite.cpp: memory map borrowed from Linksys PAP2

### DIFF
--- a/src/mame/tomy/kisssite.cpp
+++ b/src/mame/tomy/kisssite.cpp
@@ -20,7 +20,7 @@ S1L9223A01-Q0
 SAMSUNG C031
 S5L9284D01-Q0
 
-ASD  AE43BH40I16I-35
+ASD  AE43BH4016I-35
 50G00290919D
 
 F 037B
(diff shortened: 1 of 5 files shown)


src/mame/tomy/kisssite.cpp
```diff
@@ -77,17 +77,19 @@ class kisssite_state : public driver_device
...
 {
-	m_maincpu->set_state_int(STATE_GENPC, 0xe2cc); // temp, there is code here, but this is unlikely to be the entry point
+	m_maincpu->set_state_int(STATE_GENPC, 0x1cffff80); // might actually be 0x7fffff80 with a ROM mirror
 }
 
 void kisssite_state::mem(address_map &map)
 {
-	map(0x0000000, 0x003ffff).rom();
+	map(0x00000000, 0x0007ffff).ram();
+	map(0x1c000000, 0x1c03ffff).mirror(0x00fc0000).rom();
```
ajrhacker, line 86: You need to add `.region("maincpu", 0)` to make this ROM mapping work. (You also need to raise the address space dimension in the CPU device.)

mamehaze:
if there are dumps for the Linksys PAP2 can you add a skeleton for it too?

Octocontrabass:
I can do that. Would it just go in the "skeleton" directory?

mamehaze:
> I can do that. Would it just go in the "skeleton" directory?

yeah, that that's the best place to put things in a very early state, for example if they're not much more than ROM holders, you'll need to add a line/entry in mame.lst too

also contact the team using the form on https://www.mamedev.org/contact.html?team so that somebody knows where the ROM is.

when I initially wrote the disassembler here I didn't know there was anything more using the CPU type dumped, so I appreciate you fixing this up, and pointing out that there are other things that require the same disassembler.

ajrhacker:
What should Linksys PAP2 be credited to?

Octocontrabass:
Unfortunately I don't know who dumped it.

PR #12727, opened 2024-09-04
floppy.cpp: Use standard drives for hard-sectored systems

mgarlanger:
Here is an alternate approach for handling hard-sectored floppy drives. Instead of requiring duplicating the existing drives and updating the variants, the sectoring is set during device configuration. This allows the drive to select the proper variants based on the sectoring type.

### DIFF
--- a/src/devices/bus/s100/vectordualmode.cpp
+++ b/src/devices/bus/s100/vectordualmode.cpp
@@ -334,7 +334,7 @@ void s100_vector_dualmode_device::device_reset()
 
 static void vector4_floppies(device_slot_interface &device)
 {
-	device.option_add("525", FLOPPY_525_QD16);
+	device.option_add("525", FLOPPY_525_QD);
 }
 
 static void vector4_formats(format_registration &fr)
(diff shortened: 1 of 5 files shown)


galibert:
Was there cases where multiple drive types could have been connected to the same place, like different hard or both soft and hard?  Because if that's the case it's not configurable with this approach.

Also, typo (multiple times), Doubld -> Double.

mgarlanger:
I fixed the typo. 

As for drives connected to both type of controllers, I have seen it mentioned in an ad for one of the third party soft-sectored controllers, but don't have any documentation or software associated with that controller to allow it to be implemented in mame. The only advantage of connecting a drive to multiple controllers would be to save the cost of the floppy drives (at that point $300 to $500 per drive). And there were significant disadvantages, harder to copy between one format and the other.  In a virtual environment, there is no reason to not just instantiate separate drives for each controller. If someone really wanted to simulate a drive connected to both types, this approach could be extended to have a list of sectoring_types, instead of just a single value.  
The existing approach in the code would also require changes to support this, would need to add additional drive types.

galibert:
I think you got it the wrong way around.  I'm talking about a controller able to cope with different types of drives on the same connector.  Or I don't understand your answer :-)

mgarlanger:
At least for Heath systems I'm not aware of any single controller that both (I don't know of any that did hard-sectored except for the original Heath one). Here is an ad for Percom talking about it - https://archive.org/details/sextantissue2summer1982/page/n1/mode/2up

It says "The Z Controller co-resides with your H-89 disk drive controller. Your software can select either, and you don’t have to move drives around when switching between systems. "

There was also an article in REMark (the magazine associated with Heath's user's group), which showed how to modify and connect Heath's own soft-sectored and hard-sectored controllers so that they could share drives. I don't have a link but could track it down if you want it.

mgarlanger:
Here is a second company's soft-sectored controller (CDR) that offered an upgrade to allow both their soft-sector controller and heath's original hard-sectored controller to use the same disk drives - can see the full ad on page 76 - http://www.pestingers.net/pdfs/remark-scans/1984/remark-volume5-issue1-1984.pdf
Screenshot 2024-09-07 at 9 37 12 AM

PR #12731, opened 2024-09-04
Update igs_m027.cpp - bypass chessc2 startup check so decryption can be verified

mamehaze:
this is fairly crude, but it allows us to see that the decrypt functions are correct for program (was correct) and sprites (was not correct)

### DIFF
--- a/src/mame/igs/igs_m027.cpp
+++ b/src/mame/igs/igs_m027.cpp
@@ -1449,11 +1449,10 @@ void igs_m027_state::init_chessc2()
 	m_igs017_igs031->tarzan_decrypt_sprites(0, 0);
 
 	// bypass IGS025 'version' check
-	u16* ROM = (u16*)memregion("user1")->base();
-	ROM[0x207d8/2] ^= 0x0100;
+	m_external_rom[0x207d8/4] ^= 0x00000100;
 	// bypass external ROM checksum
-	u16* ROM2 = (u16*)memregion("maincpu")->base();
-	ROM2[(0x16a/2)] ^= 0x1000;
+	u32 *ROM2 = &memregion("maincpu")->as_u32();
+	ROM2[(0x168/4)] ^= 0x10000000;
 
 	// what lives here? coin error otherwise
 	m_maincpu->space(AS_PROGRAM).install_read_handler(0x4000000c, 0x4000000f, read32smo_delegate(*this, FUNC(igs_m027_state::lhdmg_unk2_r)));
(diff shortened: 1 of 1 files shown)


src/mame/igs/igs_m027.cpp
```diff
@@ -1446,6 +1446,17 @@ void igs_m027_state::init_chessc2()
...
 	chessc2_decrypt(machine());
 	m_igs017_igs031->sdwx_gfx_decrypt();
+	m_igs017_igs031->tarzan_decrypt_sprites(0, 0);
+
+	// bypass IGS025 'version' check
+	u16* ROM = (u16*)memregion("user1")->base();
+	ROM[0x207d8/2] ^= 0x0100;
+	// bypass external ROM checksum
+	u16* ROM2 = (u16*)memregion("maincpu")->base();
+	ROM2[(0x16a/2)] ^= 0x1000;
```
cuavas, line 1456: There’s an object finder for the `user1` region in the class already, and these two regions are both 32 bits wide, so interpreting them as `u16` will cause the code to break on big Endian hosts.

Please access the external ROM through `m_external_rom` and use `->as_u32(...)` to get a reference into the internal ROM.  Something like:
```c++
	// bypass IGS025 'version' check
	m_external_rom[0x207d8/4] ^= 0x00000100;
	// bypass external ROM checksum
	u32 *ROM2 = &memregion("maincpu")->as_u32();
	ROM2[(0x168/4)] ^= 0x10000000;
```

(I haven’t tested that, I’m doing the calculations in my head.)

cuavas (changes requested):

cuavas:
Thanks.

cuavas:
BTW I’ll probably get around to sorting out the rest of the lhdmg DIP switches eventually, but no guarantee on timeline.  You can view the settings (in Chinese) by pressing F2 now.

I didn’t look at hooking up the hopper yet – it might be easy.  You can trigger it by coining up, then hitting payout.  Unfortunately the default assignment for payout clashes with the default mahjong I.  We have special mahjong variants for double up, take score, etc. because the “gambling” ones clash with mahjong inputs, but not for payout.

cuavas:
Tag @Osso13 because he mentioned possibly looking at the driver.

Osso13:
I'll be 'out of commission' for a couple of weeks, starting this weekend. So if someone else wants to do it, don't hold off because of me :)

mamehaze:
I still have no plans to look at any of the inputs in this driver (or anything else in the short term)

Some of what is in igs_fear.cpp may also be needed (IRQ control/source) which is especially noticeable for the ones that require the XA hookup (igs_m027xa.cpp) as they require both the vblank interrupt, and interrupt from the XA for it to communicate with the XA properly.

He also mentioned the timer stuff for Extra Draw (again igs_fear.cpp has timers) so that might also be something present on all these boards that only a few of the games actually need.

PR #12732, opened 2024-09-04
ussr/pk32.cpp: Elektronika PK-32 - not working

holub:
*Elektronika PK32* is based on MPL1839 and wasn't produced to mass market.
at the current state PK32 runs in terminal mode with all internal tests pass.

\+ mpl1839/kl1839vm1.cpp: MPL's *L1839VM1* cpu device:
1839 cpu is able to execute code in user space fully compatible with VAX11

Some info available here (rus):
http://turbinium.com/forum/index.php?topic=16.0

### DIFF
--- a/src/devices/cpu/mpk1839/kl1839vm1.cpp
+++ b/src/devices/cpu/mpk1839/kl1839vm1.cpp
@@ -1,0 +1,686 @@
+// license:BSD-3-Clause
+// copyright-holders:Andrei I. Holub
+
+#include "emu.h"
+
+#include "kl1839vm1.h"
+#include "kl1839vm1dasm.h"
+
+
+#define VERBOSE ( LOG_GENERAL )
+#include "logmacro.h"
+
+#define UNIMPLEMENTED(msg) LOG("Unimplemented: %s\n", msg)
+
+#define MCA    m_mca.w.l // Micro Command Address
+#define VMA    m_vma.d   // Virtual Memory Address
+#define RV     m_rv.d
+#define SCH    m_sch.b.l // Program Counter
+#define RSP    m_rsp.b.l
+
+#define K(x)   m_consts[x & 0x0f]
+#define PCM    K(5)
+#define RC     K(6)
+
+/* registers of various sizes */
+#define R(x)   m_reg[x].d
+
+#define AP     R(0x0c)
+#define FP     R(0x0d)
+#define SP     R(0x0e)
+#define PC     R(0x0f)
+
+#define RNK    R(0x1c)
+#define RKA    R(0x1d) // KRSP?
+#define PSL    R(0x1e)
+#define PSW    m_reg[0x1e].w.l
+#define BO     R(0x1f)
+
+#define NF     0x08
+#define ZF     0x04
+#define VF     0x02
+#define CF     0x01
+
+DEFINE_DEVICE_TYPE(KL1839VM1, kl1839vm1_device, "kl1839vm1", "KL1839VM1")
+
+
+kl1839vm1_device::kl1839vm1_device(const machine_config &mconfig,  const char *tag, device_t *owner, u32 clock)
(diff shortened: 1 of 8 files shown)


src/devices/cpu/mpk1839/kl1839vm1.h
```diff
@@ -0,0 +1,94 @@
...
+
+
+enum
+{
+	KL1839_MCA = STATE_GENPC, KL1839_VMA, KL1839_IF, KL1839_RSP, KL1839_PCM, KL1839_RC, KL1839_RV, KL1839_SCH,
+	VAX_R0, VAX_R1, VAX_R2, VAX_R3, VAX_R4, VAX_R5, VAX_R6, VAX_R7, VAX_R8, VAX_R9, VAX_RA, VAX_RB,
+	VAX_AP, VAX_FP, VAX_SP, VAX_PC,
+	VAX_AK0, VAX_AK1, VAX_AK2, VAX_AK3, VAX_AK4, VAX_AK5, VAX_AK6, VAX_AK7, VAX_AK8,
+	VAX_RNK, VAX_RKA, VAX_PSL, VAX_BO
+};
```
happppp, line 17: If you want drivers to have access to this register list, a better place for new CPU emulators is under public:.

src/devices/cpu/mpk1839/kl1839vm1.cpp
```diff
@@ -0,0 +1,686 @@
...
+	save_item(NAME(m_mca));
+	save_item(NAME(m_vma));
+	save_item(NAME(m_vma_tmp));
+	save_item(NAME(m_rv));
+	save_item(NAME(m_sch));
+	save_item(NAME(m_rsp));
+	save_item(NAME(m_ppc));
+	save_item(NAME(m_fp));
+	save_pointer(NAME(m_reg), 0x20);
+	save_pointer(NAME(m_consts), 0x10);
```
happppp, line 562: save_item works on arrays (no idea if it works on PAIR).
save_pointer 2nd param is bytes, not array length.

src/devices/cpu/mpk1839/kl1839vm1.h
holub: Not sure how it's going to be atm. Suspecion is that drivers has no acess to this registers directly. User code is interprited by microcode of cpu.

src/devices/cpu/mpk1839/kl1839vm1.cpp
cuavas: No it isn’t, it’s the element count.

From emu/device.h:
```c++
        /// \brief Register a member of a structure in an array of
        ///   indeterminate for save states
        ///
        /// Registers data to be automatically saved/restored when the
        /// length of the outermost array cannot be automatically determined
        /// at compile time.  Can be used with C arrays of indeterminate
        /// length, pointers, and \c std::unique_ptr instances pointing to
        /// arrays.  Use #save_item if length of the array can be determined
        /// at compile time.
        /// \param value Pointer to the array of structures containing the
        ///   member to be saved/restored.  The #STRUCT_MEMBER macro can be
        ///   used to simplify specifying the \p element and \p valname
        ///   arguments at the same time.
        /// \param [in] element Pointer to the member of the structure to
        ///   save/restore.
        /// \param [in] valname The name of the saved item.  The combination
        ///   of the \p valname and \p index arguments must be unique across
        ///   saved items for a device.
        /// \param [in] count The number of elements in the outermost array.
        /// \param [in] index A numeric value to distinguish between saved
        ///   items with the same name.
        /// \sa save_item
        template <typename ItemType, typename StructType, typename ElementType>
        void ATTR_COLD save_pointer(ItemType &&value, ElementType StructType::*element, const char *valname, u32 count, int index = 0)
        {
                assert(m_save);
                m_save->save_pointer(this, name(), tag(), index, std::forward<ItemType>(value), element, valname, count);
        }
```

PR #12733, opened 2024-09-05
sun3, sun3x, sun4: Add NSCSI tape to default devices.

eschaton:
Add default NSCSI tape support to sun3, sun3x, and sun4.

### DIFF
--- a/src/mame/sun/sun3.cpp
+++ b/src/mame/sun/sun3.cpp
@@ -192,6 +192,7 @@ fefc34a - start of mem_size, which queries ECC registers for each memory board
 #include "machine/ncr5380.h"
 #include "bus/nscsi/cd.h"
 #include "bus/nscsi/hd.h"
+#include "bus/nscsi/tape.h"
 
 #include "bus/rs232/rs232.h"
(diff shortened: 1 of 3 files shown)


cuavas (commented):
Adding the tape drive as an option is definitely a good idea.  I’m not completely sold on having it connected by default, though.  No-one had tape drives connected to the “lunchbox” SPARCstation IPC and IPX workstations on their desks.

eschaton:
> The CD-ROM drives are there by default because you’re likely to need them for initial OS installation, and you can’t just use any old drive because it’s fussy about sector sizes.

Ah, that makes sense. Prior to the SPARCstation era, most Sun systems would've been installed from tape, unless they were set up over the network (from a system that was likely installed from tape). So I'd prefer to leave it in, but it's not a big deal to me if you prefer the opposite. :)

cuavas:
In sun4.cpp, can you add the tape drive by default for the sun4 machines, but not the sun4c machines?  It can stay for sun3 and sun3x.  Is that a reasonable compromise?

eschaton:
Sounds perfect to me!

PR #12734, opened 2024-09-05
sun3: Add DIAG/NORMAL switch

eschaton:
Adds a DIAG/NORMAL switch that affect how a Sun-3/60 (and other models) are configured.

### DIFF
--- a/src/mame/sun/sun3.cpp
+++ b/src/mame/sun/sun3.cpp
@@ -245,7 +245,8 @@ public:
 		m_rom(*this, "user1"),
 		m_idprom(*this, "idprom"),
 		m_ram(*this, RAM_TAG),
-		m_lance(*this, "lance")
+		m_lance(*this, "lance"),
+		m_diagsw(*this, "diagsw")
 	{ }
 
 	void sun3(machine_config &config);
@@ -273,6 +274,7 @@ private:
 	required_memory_region m_rom, m_idprom;
 	required_device<ram_device> m_ram;
 	required_device<am79c90_device> m_lance;
+	required_ioport m_diagsw;
 
 	uint32_t tl_mmu_r(offs_t offset, uint32_t mem_mask = ~0);
 	void tl_mmu_w(offs_t offset, uint32_t data, uint32_t mem_mask = ~0);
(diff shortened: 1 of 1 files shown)


eschaton:
I used "Diagnostic" and "Normal" for the values; `DIAG` and `NORM` is how the hardware is labeled. Which is preferable for MAME?

src/mame/sun/sun3.cpp
```diff
@@ -809,6 +814,23 @@ void sun3_state::vmetype3space_map(address_map &map)
...
+}
+
+void sun3_state::enable_w(uint32_t data, uint32_t mem_mask)
+{
+	// Mask out diag switch value, it comes from the switch on read.
+	const uint32_t diag = 0x01000000;
+	const uint32_t diag_norm_mask = ~diag;
+	COMBINE_DATA(&m_enable);
+	m_enable &= diag_norm_mask;
+}
```
cuavas, line 832: You can still end up with an issue doing it this way if someone injects a value with the bit set via a debugger memory viewer or something.  You’re better off masking the value on read – something like:

```c++
uint32_t sun3_state::enable_r()
{
	// Incorporate diag switch value.
	const uint32_t diagsw = m_diagsw->read() << 24;
	return m_enable & ~(u32(1) << 24) | diagsw;
}
```

and don’t bother with `enable_w` at all.

src/emu/ioport.cpp
```diff
@@ -235,6 +235,7 @@ const struct
 	{ INPUT_STRING_Reverse, "Reverse" },
 	{ INPUT_STRING_Alternate, "Alternate" },
 	{ INPUT_STRING_None, "None" },
+	{ INPUT_STRING_Diagnostic, "Diagnostic" },
```
cuavas, line 238: Please don’t add an input type for this.  Just use `PORT_CONFNAME` etc or `PORT_BIT( ..., ..., IPT_OTHER ) PORT_TOGGLE PORT_NAME(...)`

eschaton: Aha! I only did it this way because I hadn't figured that out. :)

PR #12735, opened 2024-09-05
Improvements to Yamaha MU50 internal layout

felipesanches:
Internal layout in this PR:
Screenshot from 2024-09-05 07-57-21

And with external layout (available at https://github.com/ArqueologiaDigital/MAME_synths_artwork/tree/main/mu50):
Screenshot from 2024-09-05 08-12-03

### DIFF
--- a/src/mame/layout/mu50.lay
+++ b/src/mame/layout/mu50.lay
@@ -1,15 +1,19 @@
 <?xml version="1.0"?>
 <!--
 license:CC0-1.0
-authors:hap
+authors:hap, Felipe Sanches
 -->
 <mamelayout version="2">
 
 <!-- define elements -->
 
 	<element name="led" defstate="0">
-		<disk state="0"><color red="0.07" green="0.1" blue="0.02" /></disk>
-		<disk state="1"><color red="0.7" green="1.0" blue="0.2" /></disk>
+		<disk state="0"><color red="0.2" green="0.25" blue="0.2" /></disk>
+		<disk state="1"><color red="0.43" green="1.00" blue="0.29" /></disk>
+	</element>
+
+	<element name="round_button">
+		<disk><color red="0.25" green="0.25" blue="0.25" /></disk>	
 	</element>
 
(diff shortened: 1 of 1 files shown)


cuavas:
I really don’t want to go in the direction of having SVG data that isn’t readily human-readable in internal layouts.  So far, I’ve been trying to limit it to relatively simple shapes, coordinates that are simple to understand at a glance (nothing that needs a large number of significant figures), no unnecessary CSS, and no unnecessary transforms.

The idea of supporting embedded SVG in internal layouts wasn’t to have them bloated with stuff that really belongs in external artwork – it was just to allow people to add geometric shapes beyond circles and rectangles.

galibert:
I'm not sure I agree with you.  That driver is hard to control without artwork, even if it's less horrible than mu2000 with its pile of buttons.  So we should have something usable in the main tree.  And it's a case where doing the editing in a svg program is way, way nicer.  

I get your point about embedded stuff that's just way too big in something that's supposedly human-readable (even if that is becoming even less real with lua sliders), but perhaps we need to be able to externalize the svg and put it next to the .lay?

rb6502:
I agree.  Layouts should be able to load external SVGs from the artwork folder or whatever.  That way you can do simple shapes and the interactive hookups in human-readable form but also leverage the SVG tooling available to make the things look nice.  And include a way to indicate fallback layout that only is used if an SVG load fails, so you can fall back to something ugly but usable.

happppp:
These are SVGs loaded from internal .lay files: https://github.com/mamedev/mame/tree/master/artwork/chess
Or did you mean differently?

rb6502:
I mean in a more formally packaged way, where all of those would just be in chess.zip.

rb6502:
Right, for the shared stuff like chess you'd need some way to override the parent name for .zip purposes.

galibert:
I'd rather not lose it, so I'm going to commit it keeping in mind to extract the (not that big) svg part once we make includes sane.

PR #12737, opened 2024-09-05
bus/msx/slot/panasonic08r.cpp: Add MSX Turbo-R internal firmware mapper.

wilbertpol:


src/devices/bus/msx/slot/panasonic08r.h
```diff
@@ -0,0 +1,52 @@
...
+	required_device<nvram_device> m_nvram;
+	required_memory_region m_rom_region;
+	memory_bank_array_creator<8> m_bank;
+	memory_view m_view[8];
+	required_device<msx_slot_ram_mm_device> m_mm;
+	u32 m_region_offset;
+	u16 m_selected_bank[8];
+	u8 m_control;
+	u8 m_bank9;
+	std::vector<u8> m_sram;
```
cuavas, line 48: Do you really need a `vector` for `m_sram`?  It should be allocated once and never resized.  If you make it `std::unique_ptr<u8 []>` you can’t inadvertently cause it to be reallocated and break things expecting the data not to move.

PR #12739, opened 2024-09-07
New working machine (Casio/Tandy Zoomer)

devinacker:


startaq:
Since the zoomer_rtc device will only be used by this machine it can be moved into the `casio` directory.

src/mame/casio/zoomer.cpp
```diff
@@ -0,0 +1,728 @@
...
+
+//**************************************************************************
+void zoomer_state::ems_mem_w(offs_t offset, u8 data)
+{
+	if (offset < 0xf8000)
+		m_nvram[offset] = data;
+	// TODO: PCMCIA at 0x1000000
+	else
+		logerror("%s: unmapped write to EMS @ %08x\n", machine().describe_context(), offset);
+}
```
angelosa, line 461: As far as I can tell this should really use a combination of 2 `address_map_bank_device`s with shared address_map, and a `memory_view` that blocks writes where it matters (NVRAM only?)

devinacker: It'd presumably need a separate bank device for each window (128kb windows at 0x20000 and 16kb windows at 0xa0000) but I can do that if it still sounds reasonable.

devinacker: Does this look better?

angelosa: Definitely, thanks :)

src/mame/layout/zoomer.lay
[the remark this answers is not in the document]
devinacker: Done - currently it's a separate view from the default one, but I can combine them (or just get rid of the analog pen control entirely) if you think it makes more sense to do so.

src/mame/mame.lst
```diff
@@ -16078,6 +16078,9 @@ sk10                            //
 @source:casio/sx1000.cpp
 sx1010                          // Casio SX1010
 
+@source:casio/zoomer.cpp
+zoomer                          // 1993 Casio/Tandy
```
cuavas, line 16082: This thing was also sold with GRiD and AST branding (as the GridPad 2390).  Is the firmware dump you’ve used Tandy-branded, or is it generic Casio firmware?  I wouldn’t try stuffing all the marketing names into the description and manufacturer if they all used the same firmware.

devinacker: I dumped it from a Casio unit, but the same mask ROMs appear in the parts list for the Tandy version. I don't know about the GridPad.

cuavas: Does the Casio XL7000 Zoomer variant have the same ROMs, too?  (IIRC the Z-7000 dropped the “Zoomer” branding, just keeping the stylised “Z” logo on the outside of the flip cover.)

devinacker: Also, while Casio didn't use the Zoomer name in packaging/marketing, the device itself still refers to itself as such all over the place, e.g.:

PR #12743, opened 2024-09-10
igs_m027xa : copy the XA logic from igs_fear.cpp for now + use uploaded XOR tables

mamehaze:
note, this stops crzybugs booting for now, as it fails the XA check rather than passing it by chance.  I've done some experimentation and I think this needs irq mask / source / pending registers hooking up too, otherwise the ARM will keep writing commands to the XA when it doesn't want them

the comms structure otherwise seems the same (command structure is the same, the method of shifting in/out values seems the same etc.)


mamehaze:
as mentioned previously, I think @HappyASR will likely have a better idea on how this should hook up, I suspect a few things will need tweaking at the very least for it to work here, but most of the boilerplate I've copied will be needed in some form.

mamehaze:
I'm leaning more and more towards making this a subclass of igs_m027.cpp though, it looks like the peripheral (timer + irq logic?) is in all those games too, as they do the 0x70000xxx writes as well (is this just a feature of some / all 027A chips?)

the per-frame IRQ we're generating in igs_m027.cpp is probably just the programmed timer (the NMI likely comes from the video still though)

cuavas:
It’s sufficiently different that I don’t think a derived class is really the best option.  Once the IGS027A is understood better, the common functionality should be encapsulated as a device.

PR #12744, opened 2024-09-10
TLCS900: reduce usage of preprocessor macros.

felipesanches:
Based on feedback from @cuavas at https://github.com/mamedev/mame/pull/12684#discussion_r1730039950, this PR replaces a bunch of `#define`s with `enum` and `static constexpr`.

### DIFF
--- a/src/devices/cpu/tlcs900/tmp95c063.cpp
+++ b/src/devices/cpu/tlcs900/tmp95c063.cpp
@@ -176,64 +176,63 @@ void tmp95c063_device::device_config_complete()
 	}
 }
 
+enum {
+	INTE0AD,
+	INTE12,
+	INTE34,
+	INTE56,
+	INTE78,
+	INTET01,
+	INTET23,
+	INTET45,
+	INTET67,
+	INTET89,
+	INTETAB,
+	INTES0,
+	INTES1,
+	INTETC01,
+	INTETC23
+};
 
-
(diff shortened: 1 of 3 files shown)


src/devices/cpu/tlcs900/tmp95c063.cpp
```diff
@@ -200,37 +201,52 @@ static const struct {
...
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
+	{ INTE34,   0x80, 0x38 },     // INT4
```
felipesanches, line 245: There's unnecessary repetition of the INT4 line here.

cuavas: Ah crap, did I screw up editing?  I’ll fix it on the next commit.

felipesanches: thanks!

PR #12745, opened 2024-09-10
taito_f2.cpp: Notes and game name change about different PCB configurations for PuLiRuLa [Tailsnic Retroworks]

clawgrip:


clawgrip:
> Do we know whether the single PCB version is older or newer?

No, sorry

clawgrip:
> Also out of curiosity, is the Japan set for a single or dual PCB setup?

I've asked Tailsnik to check it!!!

We're also checking other games... For instance, we've checked that "Hat Trick Hero (Japan)" only runs on double PCB, meanwhile "Football Champ" and "Hat Trick Hero" are for single PCB...

cuavas:
If you can add a comment at the ROM definition for whether they run on single- or dual-board setups that would be useful.

clawgrip:
> If you can add a comment at the ROM definition for whether they run on single- or dual-board setups that would be useful.

Yes, will do!

PR #12747, opened 2024-09-10
cpu/tlcs900 (tmp95c063): remove lines duplicated by mistake

felipesanches:


cuavas:
I batched this up with 587cb04c276f9577c6c87ab1e965230ffaf00ae1 – did I mess that up as well?

felipesanches:
> I batched this up with [587cb04](https://github.com/mamedev/mame/commit/587cb04c276f9577c6c87ab1e965230ffaf00ae1) – did I mess that up as well?

No, that's good.

PR #12748, opened 2024-09-10
igs_m027.cpp - guess at OKI banking for jking02, fruitpar, oceanpar + timer code backport

mamehaze:


mamehaze:
haunted house in 027xa is the same situation with the sample ROMs as some of the ones I've commented btw, although we're still waiting on a board of that being sent to dump the XA; the bad dump flag can likely be removed though.

clearly the OKI map is half and half banking, with upper and lower 128kb banks, the first few banks in the sample rom don't contain a sample table because they're meant for the upper 128kb of the oki memory, while the banks after that all contain sample tables because they're meant to be in the lower 128kb of OKI memory.

this means for those cases there will be a pair of bank registers, one for the lower part and one for the upper.

I guess it's possible all of them are like that, and just happen to bank both halves at the same time.

src/mame/igs/igs_m027.cpp
```diff
@@ -217,20 +306,22 @@ void igs_m027_state::lhdmg_xor_map(address_map &map)
 	igs_mahjong_xor_map(map);
 
 	map(0x4000000c, 0x4000000f).r(FUNC(igs_m027_state::lhdmg_unk2_r));
+	map(0x40000018, 0x4000001b).umask32(0x000000ff).w(FUNC(igs_m027_state::jking02_misc_w));
```
cuavas, line 309: This definitely isn’t right – bit 7 of PPI port C has something to do with sound banking, but it isn’t as simple as just banking the whole sample ROM.  There are very obvious issues with sounds during gameplay.  Unfortunately these games don’t have a sound test.

mamehaze: yeah, as I've said above, some of the other games point at it being a 128kb/128kb bank, so it could just be that fruitpar and oceanpar are using it in avery basic way, banking complete 256kb segments, and the others are not, or the logic differs between games (which would fit with everything else IGS loved to change for the sake of it)

mamehaze:
just a few things

mgzza is also 2003, not 2000  (I changed this in an earlier PR, but it got lost due to a merge conflict)

the supported revision of extradrw is 2001, the change to the sprite ROM loading from here is needed because they store the '1' of the 2001 copyright in the extra sprite ROM; it needs loading at 100000 not 200000 because the first sprite ROM doesn't contain any data in the 2nd half (probably just used a higher capacity chip than they needed)  The init change for extradrw is needed to fix the tile decoding (background graphics)

cuavas:
I think 573695d0a0dc106982dba0dbbe0320e615d9eec8 should cover the rest of the changes here.  I need to call it a night, haven’t had time to do more work on the mahjong sound banking.

You can play chessc2 now, although it could be missing sound banking.  It doesn’t seem to have an input test, sound test or DIP switch display screen.  It does read the three banks of DIP switches on start.  It seems to lack hopper or ticket payout features altogether (unless they’re in the password-protected setup menus).

Crazy Bugs now has the second coin, ticket dispenser and coin/key-in counters hooked up.  I added a layout, including clickable buttons that reflect the state of the lamp outputs.

mamehaze:
I'm kinda surprised chessc2 plays, but maybe they really did only use the 025 for startup checks rather than integrating it with the game.  The bypass hack is kinda brutal, just swapping it to treat a failure condition as a pass.  I wouldn't trust it 100% until the code doing the check has been studied a bit more.

cuavas:
To be honest, chessc2 feels like an even lower effort release than jking02.  It’s also “for amusement only”, so they probably didn’t even audit it for “fairness”.

cuavas:
Seems to all be covered as of #12766

PR #12749, opened 2024-09-11
hash/nes.xml: add Micro Mages to working software list

chungy:
New working software list (nes.xml)
Micro Mages
Micro Mages Second Quest

Micro Mages can be purchased in both physical and digital format; http://morphcat.de/micromages/ has links to the store fronts. I've dumped the physical cart and verified it is the same ROM as the digital release. Micro Mages Second Quest is a digital-only release.

I'll attach a photo of the PCB here too, but I don't think it has useful information. If I'm wrong, feel free to use it :)

IMG_20240910_174516

### DIFF
--- a/hash/nes.xml
+++ b/hash/nes.xml
@@ -24626,6 +24626,43 @@ license:CC0-1.0
 		</part>
 	</software>
 
+	<software name="micromages">
+		<description>Micro Mages</description>
+		<year>2019</year>
+		<publisher>Morphcat Games</publisher>
+		<info name="serial" value="NES-MM-MCG" />
+		<info name="release" value="20190212" />
+		<part name="cart" interface="nes_cart">
+		<feature name="slot" value="nrom" />
+		<feature name="pcb" value="NES-NROM-256" />
+		<feature name="mirroring" value="horizontal" />
+		<dataarea name="prg" size="32768">
+			<rom name="Micro Mages prg" size="32768" crc="1f4aa740" sha1="c1f0e973815687e06086e57a615abc116aa37e61" />
+		</dataarea>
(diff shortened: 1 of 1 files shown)


cuavas (changes requested):
Please indent with one tab per nested element level.  You’ve used a mixture of tabs and spaces for indentation here.

hash/nes.xml
```diff
@@ -24626,6 +24626,43 @@ license:CC0-1.0
...
+		  <rom name="Micro Mages chr" size="8192" crc="24144d94" sha1="e08bec8c93dce07262be37c7cceb7019817bea69" />
+		</dataarea>
+	  </part>
+	</software>
+
+	<software name="micromages2">
+	  <description>Micro Mages Second Quest</description>
+	  <year>2022</year>
+	  <publisher>Morphcat Games</publisher>
+	  <info name="release" value="20220915" />
```
cuavas, line 24652: This is less than three years old, and only available as a paid download.  As such, it should not be added to the software list.  We generally try to be nice to software developers.

cuavas (changes requested):

chungy:
> Please indent with one tab per nested element level. You’ve used a mixture of tabs and spaces for indentation here.

Thanks, didn't notice Emacs was doing this to me :)

hash/nes.xml
chungy: Makes sense, feel free to reject the merge. The two games come as a bundle.

cuavas:
I added your cartridge dump in 81bc5452e515e5c09751414e54eb8578fd43f07a.

chungy:
Awesome, thanks!

PR #12751, opened 2024-09-11
hash/{famicom_flop,nes,snes}.xml: add (s)NESert Golfing

chungy:
Homebrew game inspired by an iPhone game "Desert Golfing". There's a [YouTube video](https://www.youtube.com/watch?v=9UP7HImbAlA) about its development. It's pretty fun!

Old versions are not available for download anymore from [the official page](https://rainwarrior.itch.io/nesert-golfing), which only offers the latest 1.5. A workaround I found that I feel is acceptable, is that according to the [README in the source](https://github.com/bbbradsmith/NESertGolfing/), [cc65 commit 80a43d732](https://github.com/cc65/cc65/commit/80a43d732dc9551ad84343ea9644d8fab7ec0bd9) was used as the toolchain. By building this exact version, I was able to create byte-for-byte perfect replicas of the 1.5 ROMs; since this toolchain version is older than NESert Golfing, the developer likely never updated it and checking out old tags allows me to at least compile playable ROMs, and I have little reason to believe they aren't representative of the original releases, but it can't be said with absolute certainty unless old copies surface (and the very early versions only had a life-span of a few hours!).

Since this is an open source game anyway, I feel fine with attaching Zip files containing both the historic and current versions here. The SNES version was only added with 1.5 and the official download site has the only version of it, as such.

[dgolf-fds.zip](https://github.com/user-attachments/files/16973039/dgolf-fds.zip)
[dgolf-nes.zip](https://github.com/user-attachments/files/16973040/dgolf-nes.zip)

New working software list items (famicom_flop.xml)
--------------------------------------------------
NESert Golfing (v1.5)
NESert Golfing (v1.4)

New working software list items (nes.xml)
-----------------------------------------
NESert Golfing (v1.5)
NESert Golfing Tournament Edition (v1.4)
NESert Golfing (v1.4)
NESert Golfing (v1.3)
NESert Golfing (v1.2)
NESert Golfing (v1.1)
NESert Golfing (v1.0)

New working software items (snes.xml)
-------------------------------------
sNESert Golfing

### DIFF
--- a/hash/famicom_flop.xml
+++ b/hash/famicom_flop.xml
@@ -510,6 +510,38 @@ Re-releases (probably the same as the original release, but listed while waiting
 		</part>
 	</software>
 
+	<software name="dgolf">
+		<description>NESert Golfing (v1.5)</description>
+		<year>2022</year>
+		<publisher>rainwarrior</publisher>
+		<info name="release" value="20221014" /> <!-- From NESertGolfing Git tag -->
+		<part name="flop1" interface="floppy_5_25">
+			<dataarea name="flop" size="65500">
+				<rom name="dgolf.fds" size="65500" crc="e9e99977" sha1="d84cc8a87a9f078318385632fe91f8b0d36add19" />
+			</dataarea>
+		</part>
+	</software>
+
+	<!-- Rebuilt from Git using cc65 commit 80a43d732dc9551ad84343ea9644d8fab7ec0bd9.  A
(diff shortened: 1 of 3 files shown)


simzy39:
The same person created Lizard https://rainwarrior.ca/lizard/index.html

cuavas (commented):
If you can’t confirm that the recreated older versions actually match what was released, they should be marked as bad dumps.

PR #12753, opened 2024-09-11
msx/msxtr.cpp: Moved fsa1st and fsa1gt to msxtr driver file.

wilbertpol:


wilbertpol:
This makes the fsa1st and fsa1gt drivers boot into their firmware and some software becomes playable.
But due to the missing emulation of the PCM hardware and filter the overall status is still not working.

PR #12754, opened 2024-09-12
Update ci-windows.yml

gdicnng:
Change "make -j2" to "make -j4". 

the cpu core number is 4 now , you can see the docs here : 

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


chungy:
Is it possible to use `make -j$(nproc)`?

angelosa:
https://github.com/mamedev/mame/pull/12442#issuecomment-2150328425

I guess the actual fix that should happen here is a comment rationale on that line ...

PR #12755, opened 2024-09-12
bus/qbus: add bus error callback and an option to install_device into a

shattered:
memory view.


shattered:
@ajrhacker please take a look?

PR #12758, opened 2024-09-13
use templates for TLCS900 I/O ports

felipesanches:
Based on feedback from @cuavas at https://github.com/mamedev/mame/pull/12684#discussion_r1730040576, this PR uses templates to "avoid copy/pasta".


felipesanches:
@cuavas, I consider this ready for review.

felipesanches:
I noticed that it would, as usual, be problematic to declare the `enum Ports` outside of the device class, as that way it would be declared globally. And when I tried to declare it inside the `tmp95c063_device` class, it became uncomfortable to keep prepending "tmp95c061_device::" to all references to port names (like `tmp95c061_device::PORT_A`). Maybe I'm unaware of a better way of doing it, so I would appreciate if you have any suggestions.

One alternative approach that I pushed as a separate commit, so that it is easier to drop it if not considered good, was to use `static constexpr uint8_t` statements for the I/O port names, inside the cpu device classes (`tmp95c061_device` and `tmp95c063_device`). That works, but I am not sure what @cuavas thinks. Please let me know.

cuavas:
That’s much neater, thanks.

PR #12760, opened 2024-09-13
Add Eniac Sound Saurus BH-1000 dump

buffis:
Add dump of Eniac Sound Saurus BH-1000
https://piano.tyonmage.com/eniac/bh-1000.html

There's a previous dump of the Suzuki GM-1000 which seems to be pretty much the same device. That one lacks the mask rom dumps.

Pics: https://photos.app.goo.gl/G12YJPEWpDKQPYr69

Roms have been sent to DU. Let me know if you need them.

Credit: buffi


rb6502:
I'm confused about this PR - why does it include random psikyo changes?  Those seem fine in and of themselves, but a PR should do one thing :-)

buffis:
Simple reason, incompetence! ;)
No but in all seriousness, looks like the master branch of my clone repo is weird from earlier commits... i'll just go ahead and try to fix that and make a new pull request. Sorry for the noise.

buffis:
New PR at https://github.com/mamedev/mame/pull/12797

PR #12765, opened 2024-09-14
Add more FreeDOS versions

chungy:
Official releases: 1.3-RC5 and 1.3 were missing, I've added those in the first commit.

I've also added an unofficial 5150-optimized version from https://archive.org/details/free-dos-1.3-8086-minimized. While I'm wary of including an unofficial release too, it is very useful to making FreeDOS usable on the ibm5150 driver, the regular version being much more optimized for the 486/Pentium era of computers. It also only comes in two 360K floppy disks, instead of a dozen 720K disks. The second commit can easily be omitted if it's not suitable for the MAME software list.

### DIFF
--- a/hash/ibm5150.xml
+++ b/hash/ibm5150.xml
@@ -6187,6 +6187,33 @@ Stalls or throws either "Bad command or file name", "disk read error" or "divide
 		</part>
 	</software>
 
+	<!--
+		This is not an official FreeDOS distribution, it is modified
+		to run more optimally on the original IBM PC model 5150.
+
+		More information about this distribution is available at:
+		* https://www.youtube.com/watch?v=EOVLlMQs9f8
+		* https://archive.org/details/free-dos-1.3-8086-minimized
+	-->
+	<software name="freedos13_8086">
+		<description>FreeDOS 1.3 8086 Minimized</description>
+		<year>2022</year>
+		<publisher>AkBKukU</publisher>
+		<info name="version" value="1.3" />
+		<part name="flop1" interface="floppy_5_25">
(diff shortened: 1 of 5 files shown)


angelosa:
I'm fine with these addition in general, I just have doubts about adding these "release candidates" at all, given that (as the name implies) they are just buggy versions of the same thing. This obviously also applies retroactively, like I don't see the point in supporting FreeDOS 1.3RC1-4.

chungy:
They could perhaps be removed, I doubt FreeDOS is at great risk of losing its development history (being open source, it's widely mirrored...). MAME has plenty of other examples of prototype software/games available, on the other hand, and filling in the missing RC just made it more complete to that aspect.

holub:
> I'm fine with these addition in general, I just have doubts about adding these "release candidates" at all, given that (as the name implies) they are just buggy versions of the same thing. This obviously also applies retroactively, like I don't see the point in supporting FreeDOS 1.3RC1-4.

We must define official guidelines regarding this: https://github.com/mamedev/mame/pull/11466

angelosa:
> > I'm fine with these addition in general, I just have doubts about adding these "release candidates" at all, given that (as the name implies) they are just buggy versions of the same thing. This obviously also applies retroactively, like I don't see the point in supporting FreeDOS 1.3RC1-4.
> 
> We must define official guidelines regarding this: #11466

Let's put this way: these are Operating Systems, and the install phases are one offs. Supporting RCs is the equivalent of supporting non-LTS Linux distros, and we don't have the workforce for that unnecessary level of detail (read: non-negligible install times for testing).

PR #12766, opened 2024-09-14
igs_m027.cpp: Added OKI samples ROM banking with 128KB size

HappyASR:
- igs_m027.cpp: Corrected OKI samples ROM banking for lhdmg/lhzb3
- igs_m027.cpp: Added NVRAM support
- igs_m027xa.cpp: Corrected OKI samples ROM banking for jking02
- igs_m027xa.cpp: Added NVRAM support

### DIFF
--- a/src/mame/igs/igs_m027.cpp
+++ b/src/mame/igs/igs_m027.cpp
@@ -15,11 +15,10 @@
  To emulate these games the Internal ROM will need dumping
  There are at least 20 other games on this and similar platforms.
 
- Hold service on boot to access input test.
+ Hold service on boot to access input test and sond test when implemented.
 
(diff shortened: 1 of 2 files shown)


cuavas:
I could be missing something, but I can’t see where this corrects the Oki ROM banking for lhdmg/lhdmg3.  From reading the source, it doesn’t set a banked map for the Oki sound chip, and it doesn’t ever change the ROM bank.

cuavas:
Oh my bad, it’s added to `mahjong_output_w`.  So it applies to lthy, zhongguo, etc. as well?

cuavas:
Maybe it’s time to promote the games with complete I/O and sound.

PR #12769, opened 2024-09-14
add undumped note to seibuspi.cpp

simzy39:
Guru's blog post from the 5th Sept 2024 talks about an undumped  PAL16L8 marked 'MCTL01' on the first version of Seibu SPI hardware, so I added a note.

https://gurudumps.otenko.com/


cuavas:
You’re saying a PAL16L8 is undumped while editing a comment about a GAL16V8D…

Gurudumps:
There's a bunch of PALs/GALs on SPI, as listed in my board doc. The one in my blog is only one of them. Adding a note for just one of them is stupid as they are all undumped. There's literally tens of thousands of undumpeds PALs/GALs/CPLDs spread across 50000+ boards and almost all of them are undumped and those that are dumped are mostly not in MAME. Should we now add a comment to all of those to say they are 'undumped as of 2024' ?
No, of course not.

PR #12770, opened 2024-09-14
emu/diserial.cpp: Improved stop bit handling (fixes #12762)

### LINKED ISSUE #12762: Serial emulation : bytes are read too slow
### MAME version

0.263 (unknown) (0.263+dfsg.1-0ubuntu1~ppa1~mantic1)

### System information

Linux Ubuntu 23.04, x86_64, en_US.UTF-8 locale.

### INI configuration details

```INI
﻿#
# CORE CONFIGURATION OPTIONS
#
readconfig                1
writeconfig               1

#
# CORE SEARCH PATH OPTIONS
#
homepath                  .
rompath                   $HOME/mame/roms;/usr/local/share/games/mame/roms;/usr/share/games/mame/roms
hashpath                  files
samplepath                $HOME/mame/samples;/usr/local/share/games/mame/samples;/usr/share/games/mame/samples
artpath                   $HOME/mame/artwork;/usr/local/share/games/mame/artwork;/usr/share/games/mame/artwork
ctrlrpath                 /usr/share/games/mame/ctrlr
inipath                   $HOME/.mame;/etc/mame
fontpath                  /usr/share/games/mame/fonts
cheatpath                 $HOME/mame/cheat;/usr/local/share/games/mame/cheat;/usr/share/games/mame/cheat
crosshairpath             $HOME/mame/crosshair;/usr/local/share/games/mame/crosshair;/usr/share/games/mame/crosshair
pluginspath               /usr/share/games/mame/plugins
languagepath              /usr/share/games/mame/language
swpath                    software

#
# CORE OUTPUT DIRECTORY OPTIONS
#
cfg_directory             $HOME/.mame/cfg
nvram_directory           $HOME/.mame/nvram
input_directory           $HOME/.mame/inp
state_directory           $HOME/.mame/sta
snapshot_directory        $HOME/.mame/snap
diff_directory            $HOME/.mame/diff
comment_directory         $HOME/.mame/comments
share_directory           share

#
# CORE STATE/PLAYBACK OPTIONS
#
state                     
autosave                  0
rewind                    0
rewind_capacity           100
playback                  
record                    
exit_after_playback       0
mngwrite                  
aviwrite                  
wavwrite                  
snapname                  %g/%i
snapsize                  auto
snapview                  auto
snapbilinear              1
statename                 %g
burnin                    0

#
# CORE PERFORMANCE OPTIONS
#
autoframeskip             0
frameskip                 0
seconds_to_run            0
throttle                  1
sleep                     1
speed                     1.0
refreshspeed              0
lowlatency                0

#
# CORE RENDER OPTIONS
#
keepaspect                1
unevenstretch             1
unevenstretchx            0
unevenstretchy            0
autostretchxy             0
intoverscan               0
intscalex                 0
intscaley                 0

#
# CORE ROTATION OPTIONS
#
rotate                    1
ror                       0
rol                       0
autoror                   0
autorol                   0
flipx                     0
flipy                     0

#
# CORE ARTWORK OPTIONS
#
artwork_crop              0
fallback_artwork          
override_artwork          

#
# CORE SCREEN OPTIONS
#
brightness                1.0
contrast                  1.0
gamma                     1.0
pause_brightness          0.65
effect                    none

#
# CORE VECTOR OPTIONS
#
beam_width_min            1.0
beam_width_max            1.0
beam_dot_size             1.0
beam_intensity_weight     0
flicker                   0

#
# CORE SOUND OPTIONS
#
samplerate                48000
samples                   1
volume                    0
compressor                1
speaker_report            0

#
# CORE INPUT OPTIONS
#
coin_lockout              1
ctrlr                     
mouse                     0
joystick                  1
lightgun                  0
multikeyboard             0
multimouse                0
steadykey                 0
ui_active                 0
offscreen_reload          0
joystick_map              auto
joystick_deadzone         0.3
joystick_saturation       0.85
joystick_threshold        0.3
natural                   0
joystick_contradictory    0
coin_impulse              0

#
# CORE INPUT AUTOMATIC ENABLE OPTIONS
#
paddle_device             keyboard
adstick_device            keyboard
pedal_device              keyboard
dial_device               keyboard
trackball_device          keyboard
lightgun_device           keyboard
positional_device         keyboard
mouse_device              mouse

#
# CORE DEBUGGING OPTIONS
#
verbose                   0
log                       0
oslog                     0
debug                     1
update_in_pause           0
debugscript               
debuglog                  0

#
# CORE COMM OPTIONS
#
comm_localhost            0.0.0.0
comm_localport            15112
comm_remotehost           127.0.0.1
comm_remoteport           15112
comm_framesync            0

#
# CORE MISC OPTIONS
#
drc                       1
drc_use_c                 0
drc_log_uml               0
drc_log_native            0
bios                      
cheat                     0
skip_gameinfo             0
uifont                    default
ui                        cabinet
ramsize                   
confirm_quit              0
ui_mouse                  1
language                  
nvram_save                1

#
# SCRIPTING OPTIONS
#
autoboot_command          
autoboot_delay            0
autoboot_script           
console                   0
plugins                   1
plugin                    
noplugin                  

#
# HTTP SERVER OPTIONS
#
http                      0
http_port                 8080
http_root                 web

#
# OSD INPUT MAPPING OPTIONS
#
uimodekey                 DEL
controller_map            none
background_input          0

#
# OSD FONT OPTIONS
#
uifontprovider            auto

#
# OSD OUTPUT OPTIONS
#
output                    auto

#
# OSD INPUT OPTIONS
#
keyboardprovider          auto
mouseprovider             auto
lightgunprovider          auto
joystickprovider          auto

#
# OSD DEBUGGING OPTIONS
#
debugger                  auto
debugger_port             23946
debugger_font             auto
debugger_font_size        0
watchdog                  0

#
# OSD PERFORMANCE OPTIONS
#
numprocessors             auto
bench                     0

#
# OSD VIDEO OPTIONS
#
video                     bgfx
numscreens                1
window                    1
maximize                  1
waitvsync                 0
syncrefresh               0
monitorprovider           auto

#
# OSD PER-WINDOW VIDEO OPTIONS
#
screen                    auto
aspect                    auto
resolution                560x384
view                      auto
screen0                   auto
aspect0                   auto
resolution0               auto
view0                     auto
screen1                   auto
aspect1                   auto
resolution1               auto
view1                     auto
screen2                   auto
aspect2                   auto
resolution2               auto
view2                     auto
screen3                   auto
aspect3                   auto
resolution3               auto
view3                     auto

#
# OSD FULL SCREEN OPTIONS
#
switchres                 0

#
# OSD ACCELERATED VIDEO OPTIONS
#
filter                    1
prescale                  1

#
# OpenGL-SPECIFIC OPTIONS
#
gl_forcepow2texture       0
gl_notexturerect          0
gl_vbo                    1
gl_pbo                    1
gl_glsl                   0
gl_glsl_filter            1
glsl_shader_mame0         none
glsl_shader_mame1         none
glsl_shader_mame2         none
glsl_shader_mame3         none
glsl_shader_mame4         none
glsl_shader_mame5         none
glsl_shader_mame6         none
glsl_shader_mame7         none
glsl_shader_mame8         none
glsl_shader_mame9         none
glsl_shader_screen0       none
glsl_shader_screen1       none
glsl_shader_screen2       none
glsl_shader_screen3       none
glsl_shader_screen4       none
glsl_shader_screen5       none
glsl_shader_screen6       none
glsl_shader_screen7       none
glsl_shader_screen8       none
glsl_shader_screen9       none

#
# OSD SOUND OPTIONS
#
sound                     auto
audio_latency             2

#
# PORTAUDIO OPTIONS
#
pa_api                    none
pa_device                 none
pa_latency                0

#
# OSD MIDI OPTIONS
#
midiprovider              auto

#
# OSD EMULATED NETWORKING OPTIONS
#
networkprovider           auto

#
# BGFX POST-PROCESSING OPTIONS
#
bgfx_path                 /usr/share/games/mame/bgfx
bgfx_backend              auto
bgfx_debug                0
bgfx_screen_chains        crt-geom-deluxe
bgfx_shadow_mask          slot-mask.png
bgfx_lut                  lut-default.png
bgfx_avi_name             auto

#
# SDL PERFORMANCE OPTIONS
#
sdlvideofps               0

#
# SDL VIDEO OPTIONS
#
centerh                   1
centerv                   1
scalemode                 none

#
# SDL FULL SCREEN OPTIONS
#
useallheads               0
attach_window             

#
# SDL KEYBOARD MAPPING
#
keymap                    0
keymap_file               keymap.dat

#
# SDL JOYSTICK MAPPING
#
sixaxis                   0

#
# SDL LIGHTGUN MAPPING
#
lightgun_index1           auto
lightgun_index2           auto
lightgun_index3           auto
lightgun_index4           auto
lightgun_index5           auto
lightgun_index6           auto
lightgun_index7           auto
lightgun_index8           auto

#
# SDL LOW-LEVEL DRIVER OPTIONS
#
videodriver               auto
renderdriver              auto
audiodriver               auto
gl_lib                    auto

#
# FRONTEND COMMAND OPTIONS
#
dtd                       1

#
# SLOT DEVICES
#
printer                   null_modem
modem                     null_modem

#
# IMAGE DEVICES
#
floppydisk1               /home/colin/Documents/git-oss/a2tools/wozamp-65c02.po
floppydisk2               /home/colin/Documents/git-oss/a2tools/disks/qt100.dsk
bitbanger                 /home/colin/Documents/git-oss/a2tools/socket.localhost:2001
bitbanger2                /home/colin/Documents/git-oss/a2tools/socket.localhost:2000
```

### Emulated system/software

apple2c0

### Incorrect behaviour

Hi,
I am currently developing an audio/video player for the Apple II, and have noticed that MAME is feeding bytes from the bitbangers to the emulated Apple II too slowly. 

The symptoms are visible in this demo video : https://www.youtube.com/watch?v=cqWVqEN_OCw
It consists of three parts: 

- a 1 minute timer video, played on real hardware. You can notice that this part is precisely 1 minute long, and 1 second/10 seconds beeps are in phase with the displayed timer
- the same video, played through a MAME-emulated Apple IIc. You can see that the video part is correct, as the timer on screen and the video timestamp are in sync; but you can hear that the beeps are not in sync. This is especially audible with the 10 seconds beeps: the last one arrives almost 5 seconds late.
- A quick test to verify that the problem does not come from the rest of the serial emulation: feeding 11520*60 bytes of 'A' through it takes at most 5 seconds.

A bit more information about the serial emulation: My transcoder gets a video file, transcodes it to a video part and a sound part. The sound part is 5bit, 11520Hz PCM data, and is streamed to the Apple II's modem port, configured at 115200 8n1. 

That takes perfect care of the playback speed on real hardware, as 115200 8n1 means 11520 bytes/s, including 1 start bit, 8 data bits, 1 stop bit.

(The video is streamed to the printer port. In this demo it is not slowed down because it uses less bandwidth (~1400 bytes/s), so even if MAME reads it slowly, it's still fast enough to do 25fps).

On MAME, the serial stack is as follows:
- My proxy streams to /dev/tnt0 (provided by tty0tty, https://github.com/freemed/tty0tty. Everything written to /dev/tnt0 gets output to /dev/tnt1, and vice-versa)
- ser2net binds /dev/tnt1 to localhost:2000
- MAME reads that via `-modem null_modem -bitb2 socket.localhost:2000`

The last part of my test video is there to verify that the tty0tty + ser2net part of the emulation is not responsible for the slowdown (on the contrary, it is much too fast, transferring 60*11520 'A' characters in less than 6 seconds. This causes me other problems, but they're not MAME-related).

I would love to help fix this, but I have no idea where to start in the code. I hope that, if nobody can take care of this issue for me, maybe someone with a good knowledge of MAME's codebase could point me in the right direction and help me solve it?

Thanks a lot in advance!

### Expected behaviour

MAME sends 11520 bytes per second to the emulated serial port.

### Steps to reproduce

mame apple2c0 -window -flop1 wozamp-65c02.po -resolution 560x384 -printer null_modem -bitb socket.localhost:2001 -modem null_modem -bitb2 socket.localhost:2000 -nomouse

### Additional details

_No response_

colinleroy:
Inserting an extra stop bit does not seem necessary anymore (it was introduced in commit 198565b8), and it makes serial noticeably too slow at 115200 baud/s, pulling bandwidth from 11520 B/s down to 10472 B/s, with each byte needing 95µs instead of 86µs to be fetched.


cuavas:
I’d like to know *what* the extra stop bit was supposed to fix.  diserial.cpp is used for all sorts of things.

colinleroy:
> I’d like to know _what_ the extra stop bit was supposed to fix. diserial.cpp is used for all sorts of things.

Me too. That's the only things I can tell you about it:
- https://github.com/mamedev/mame/commit/198565b8
- https://github.com/mamedev/mame/commit/1076f566a43f55b3f6a026e03f06685cf433483a#diff-dbd6e1f26ff0c49c6f5adde5a2f231eb2c668e23a05fa8b300639be07c24f05c

colinleroy:
I can just say "it still works for me" using apple2c0 `-modem null_modem -bitb socket.localhost:2000`, apple2p or apple2ee `-sl2 ssc -sl2:ssc:rs232 null_modem -bitb2 socket.localhost:2000`...

colinleroy:
Maybe @JoakimLarsson can remember what his patch fixed?

colinleroy:
> I’d like to know _what_ the extra stop bit was supposed to fix. diserial.cpp is used for all sorts of things.

I found the original PR: https://github.com/mamedev/mame/pull/802

That is simply not true, there is no special small delay between bytes on a serial line. (The small delay is the stop bit(s)).
There's a start bit used for sync, then the data bits, then optionally the parity bit, then one, 1.5 or two stop bits, then the next start bit, etc.
This is both documented extensively on internet (eg. https://thecloudstrap.com/what-is-rs232-protocol-explained/) and in real life, using signal analysers or simply math.

Lord-Nightmare:
I'm wondering if pr #802 was a flawed attempt to fix a bug which was actually elsewhere, like the one fixed in c056522aabda6256f7ab24cc40a41ec134064acc

cuavas:
> That is simply not true, there is no special small delay between bytes on a serial line. (The small delay is the stop bit(s)). There's a start bit used for sync, then the data bits, then optionally the parity bit, then one, 1.5 or two stop bits, then the next start bit, etc. This is both documented extensively on internet (eg. https://thecloudstrap.com/what-is-rs232-protocol-explained/) and in real life, using signal analysers or simply math.

That’s only true if the system involved has a UART with a separate transmit data register and transmit shift register, though.  In that case, the UART can reload the transmit shift register from the transmit data register immediately after shifting out a byte.  If there isn’t a separate transmit data register, the UART needs to wait for the system to write the next byte to transmit after shifting out a byte, which inevitably leads to some delay.  Various 8-bit systems also had issues receiving serial data with no gaps between bytes.

But that aside, glancing at diserial.cpp, I think the actual issue is that the stop bits setting is affecting reception as well as transmission.  This isn’t how actual UARTs work.  The stop bits setting only affects transmission.  The receiver will treat a received byte as complete if it detects a space when sampling the stop bit and return to waiting for a start bit.  (If the receiver detects a mark when sampling the stop bit, it typically waits an additional half-bit interval before returning to waiting for a start bit.)

What #802 was presumably trying to fix was a case where the emulated system is capable of transmitting serial bytes back-to-back, but requires a delay between received bytes.

The proper fix would be to change the receive side to be unaffected by the stop bits setting, then remove the extra delay between bytes, and add real support for 1.5 stop bits rather than just making it behave like 2 stop bits.

colinleroy:
I'm having a hard time understanding the travel a byte makes through the various elements of MAME's serial emulation, but from what I understand since yesterday, and from your message:

- The transmit side, in my case, is null_modem.cpp, backed by diserial.cpp
- The receive side, in my case, is mos6551.cpp

- The current implementation of diserial.cpp makes it so the bitbanger-fed null_modem has no separate transmit data register and transmit shift register, and has to delay the start of the next byte ?
- And my patch makes it the contrary ?

The part I don't understand is the end:
> change the receive side to be unaffected by the stop bits setting
How can the receive side not be affected by the stop bits setting? if the transmit side sends two stop bits over the "wire", it will make the next byte start one bit later than if there were only one stop bit?

ajrhacker:
> If there isn’t a separate transmit data register, the UART needs to wait for the system to write the next byte to transmit after shifting out a byte, which inevitably leads to some delay.

This was the case, for instance, with the Z8's built-in SIO. It was described as a feature allowing one to send break signals on P37 by repeatedly writing zero to the SIO data register. (I think I might have made its non-diserial emulation transmit one too many stop bits, though the hardware is specified to transmit "at least two.")

cuavas:
A UART contains a transmitter and a receiver.  Changing the stop bits setting on a real UART only affects the behaviour of the transmitter, it has no effect on the bahaviour of the receiver.

However, changing the stop bits setting in `device_serial_interface` affects its receiver behaviour, causing it to expect to receive the same stop bit interval it transmits.  This is bad behaviour – the stop bits setting should not affect the behaviour of the receiver, only the transmitter.

The issue #802 was trying to address was dealing with an emulated system that can be set to transmit 1 stop bit, but which requires some delay between received bytes.  If you set the `device_serial_interface` to 1 stop bit, the emulated system wouldn’t be able to keep up with the data it received.  But if you set `device_serial_interface` to 1.5 or 2 stop bits, it would reject bytes received from the emulated system as it would see the following start bit as a framing error.

As-is, this patch produces the same issue that #802 was attempting to fix.  In order to avoid reintroducing the issue, the receive side of `device_serial_interface` needs to be fixed to ignore the stop bits setting as well as removing the “extra” stop bit that #802 introduced.

colinleroy:
Now I get it (I think)! thanks for your patience. I'll see if I can do something about that, then :)

ajrhacker:
> Changing the stop bits setting on a real UART only affects the behaviour of the transmitter, it has no effect on the bahaviour of the receiver.

This may depend on the UART. SCN2651 is documented as searching two stop bits if the framing parameters are so configured, though SCN2661 only checks one.

cuavas:
Zilog SIO and SCC only ever look for a single stop bit.  INS8250 is explicitly documented as only checking for one stop bit.  (“The Receiver checks the first Stop-bit only, regardless of the number of Stop bits selected.”)  The SCN2651 is the odd one out if it really checks for multiple stop bits.

cuavas:
Just checked the 6850 ACIA again, and it also only checks a single stop bit.

colinleroy:
So in addition to the change already in this PR,  this is `device_serial_interface::set_data_frame()` and `device_serial_interface::receive_register_update_bit()` that I should change to handle 0, 1 or 2 stop bits?

cuavas:
Well, the minimum additional change is to make `device_serial_interface::receive_register_update_bit` ignore `m_df_stop_bit_count` and always check a single stop bit.

The zero stop bits setting doesn’t really make sense anyway, and making 1.5 stop bits work properly can be a job for another day – it’S been broken forever anyway.

ajrhacker:
> The SCN2651 is the odd one out if it really checks for multiple stop bits.

Yes, this was probably a bug that Signetics thought was worth fixing in its successor. The original-model Intel 8251 possibly made the same mistake, judging from early manuals; if so, it was fixed in the 8251A along with several more serious bugs.

colinleroy:
@cuavas Can you tell me where is the code that feeds diserial.cpp from the bitbanger?
TIA :)

ajrhacker:
> The zero stop bits setting doesn’t really make sense anyway

Not for asynchronous transmission, no, but the present I8251 emulation also uses diserial for its synchronous mode.

> Can you tell me where is the code that feeds diserial.cpp from the bitbanger?

Take a look in `devices/bus/rs232/null_modem.cpp`.

colinleroy:
Thanks for your help @cuavas and @ajrhacker.
I've pushed a second commit which should implement what @cuavas wanted (but not requiring any stop bits at all).

Hope this helps!

ajrhacker:
No, that looks wrong. In asynchronous reception, you need to examine at least one stop bit or you can't detect framing errors.

colinleroy:
> No, that looks wrong. In asynchronous reception, you need to examine at least one stop bit or you can't detect framing errors.

How do I know whether we are in asynchronous or synchronous reception?

ajrhacker:
Synchronous reception uses STOP_BITS_0 here and asynchronous reception never does.

By the way, I don't know of any device in MAME that absolutely relies on framing error detection for correct operation under normal conditions. Parity error detection, on the other hand, is required by several of the keyboard interfaces used by Sanyo.

colinleroy:
Thanks for your help, @ajrhacker , it is appreciated.
We should be closer to a good solution with this third commit?

colinleroy:
Hi there! 
What do you think of the current state of that pr?
For my part, it fixes the extra bit slowness, and everything works with one or two stop bits each side.

rb6502:
I like this PR, but I'm going to wait until after the branch for the next release to merge it (should be tomorrow, Sep. 20th).

cuavas:
That looks like a relatively low-impact way to get this working for now.

The UART trying to use `device_serial_interface` for synchronous communication really should be weaned off it.  Trying to use it for emulating real UARTs never ends well.

colinleroy:
Thanks !

PR #12776, opened 2024-09-16
Add FreeDOS 1.3, remove release candidates

chungy:
An alternative pull request to #12765, this one removing software list items.

This pull includes the final FreeDOS 1.3 release, an unofficial minimalist version, and removes all the release candidates from the list. Being that FreeDOS is an open source project with plentiful of mirrors, it is not in danger of losing its development history and does not really needs MAME's help to preserve them.

From the older pull request, on the minimalist version:
> I've also added an unofficial 5150-optimized version from https://archive.org/details/free-dos-1.3-8086-minimized. While I'm wary of including an unofficial release too, it is very useful to making FreeDOS usable on the ibm5150 driver, the regular version being much more optimized for the 486/Pentium era of computers. It also only comes in two 360K floppy disks, instead of a dozen 720K disks. The ~~second~~ third commit can easily be omitted if it's not suitable for the MAME software list.


chungy:
I pushed hard disk image definitions here, which are tricky to handle since no such image exists in the upstream distributions.

I'll attach here both the disk images and INP files that can be used to recreate the disk images.

[freedos13.zip](https://github.com/user-attachments/files/17009698/freedos13.zip)

hash/ibm5170_hdd.xml
```diff
@@ -32,13 +32,13 @@ license:CC0-1.0
...
+		<description>FreeDOS 1.3</description>
+		<year>2022</year>
 		<publisher>The FreeDOS Project</publisher>
 		<part name="hdd" interface="ide_hdd">
 			<diskarea name="harddriv">
-				<disk name="freedos13rc4" sha1="bb81d708746f4e5cfd4207af598cd7691a8e33bf" writeable="yes"/>
+				<disk name="freedos13" sha1="72eef7bf0ad53b8a240ce05244173df54db923f0" writeable="yes"/>
 			</diskarea>
 		</part>
 	</software>
```
angelosa, line 44: Note: FreeDOS 1.3 (at least) has a out-the-box bug around CPU detection (which comes from the `vinfo /m` calls), so for example it will wrongly goto as 80186 in `ct486`.

https://gitlab.com/FreeDOS/issue-reporting/-/issues/26

I guess this is worth mentioning in a`<notes>` line for this HDD entry, giving an hint to the end user about it.

chungy: I could work around it by editing fdauto.bat to always go into the 286/386 routines (this disk image is made for ibm5170, after all), but that'd both require me to make a new inp (install from scratch, then edit), and I'd feel kind of dirty for making a modified OS image.

At least a notes line wouldn't involve either of those...

chungy: Adding the notes is easy enough, I think, but could you give me a clue to where in the MAME UI it's supposed to display? I can't find it at all.

wilbertpol: If I'm not mistaken, the info usage elements should be displayed in UI.

The notes field is just for notes. We use it mostly for documenting things that are broken with the emulation.

angelosa: The trouble with info usage is that it doesn't allow for multiline commentary afaik, so long lines will go way after any conventional file width limit. fwiw frontend has no limit ...
image

chungy: I'm not even sure it's necessary. It's not really MAME's job to point out all software bugs, and this one doesn't cause FreeDOS to be unusable.

hash/ibm5170.xml
```diff
@@ -590,221 +590,106 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="freedos13rc1">
-		<description>FreeDOS 1.3 Release Candidate 1 (bootdisk)</description>
-		<year>2019</year>
+	<software name="freedos13">
```
angelosa, line 593: The bootdisk is tied with the CDROM release, should be moved in `ibm5170_cdrom.xml`

chungy: > The bootdisk is tied with the CDROM release, should be moved in `ibm5170_cdrom.xml`

Agreed :) This was one of my earlier PRs, I'll fix this up now.

PR #12777, opened 2024-09-17
hash/ibm5170{,_cdrom}.xml: Pristine Windows NT 3.1 builds

chungy:
The image that was listed here is a hacked-up warez version from 2003 that doesn't function properly (won't install on most computers NT 3.1 should be able to install on).

In its stead, this is a list of pristine images direct out of Microsoft, no modifications have been made to them, and NT 3.1's (admittedly limited) full hardware compatibility is supported.

Taking from the BeOS example, floppy images that are intimately tied to the CD-ROM are listed in the same entry in ibm5170_cdrom.xml, since they are only useful in booting a computer in order to install from CD-ROM (which requires a compatible SCSI controller and SCSI CD-ROM drive).  Installing from MS-DOS is always another possibility.

### DIFF
--- a/hash/ibm5170.xml
+++ b/hash/ibm5170.xml
@@ -4565,153 +4565,241 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="winnt31w">
-		<description>Windows NT 3.1 Workstation (3.10.511.1)</description>
-		<year>1993</year>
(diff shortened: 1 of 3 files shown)


chungy:
I'm going to close this pull request. I expanded my own scope to all of (pre-2000) NT. May as well open a new pull request with a proper branch name. :-)

PR #12779, opened 2024-09-17
new NOT WORKING machines (IGS Triple Slot + Triple Fever clones)

mamehaze:
new NOT WORKING machines
-----------
Triple Slot (V200VE) [Ioannis Bampoulas, TeamEurope]

new NOT WORKING clones
----------
Triple Fever (V108US) [Ioannis Bampoulas, TeamEurope]
Triple Fever (V105US) [Ioannis Bampoulas, TeamEurope]

- improve ROM labels for extradrw, add note about extra 8255 [Jorge Silva, Team Europe]
- added XA dump to Triple Fever (matches crzybugs) [Ioannis Bampoulas, TeamEurope]

### DIFF
--- a/scripts/src/bus.lua
+++ b/scripts/src/bus.lua
@@ -2865,6 +2865,8 @@ if (BUSES["VME"]~=null) then
 		MAME_DIR .. "src/devices/bus/vme/mvme187.h",
 		MAME_DIR .. "src/devices/bus/vme/mvme327a.cpp",
 		MAME_DIR .. "src/devices/bus/vme/mvme327a.h",
+		MAME_DIR .. "src/devices/bus/vme/mvme328.cpp",
+		MAME_DIR .. "src/devices/bus/vme/mvme328.h",
 		MAME_DIR .. "src/devices/bus/vme/mvme350.cpp",
 		MAME_DIR .. "src/devices/bus/vme/mvme350.h",
 		MAME_DIR .. "src/devices/bus/vme/mzr8105.cpp",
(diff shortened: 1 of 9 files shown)


mamehaze:
mostly needs proper OKI banking, and inputs/lamps sorting out correctly, it's somewhat playable as-is though

cuavas:
I’ve got I/O hooked up for the Triple Fever games, but they trigger a fatal error after paying out tickets.  If you win a game, press TAKE, and then press Ticket (I by default), it will pay out the tickets then crash with “Fatal error: ADD.w R0, [R7+#$02]” from the ARM CPU core.

cuavas:
Triple Slot is another nasty one with two supported wiring options (like Jungle King).  I’ll get to it later.

mamehaze:
the error is from the XA, not the ARM.  I've added the missing opcode, but it still hangs at that point, no longer responding to any inputs and still showing 'payout'
I wonder if something has already failed at that point, hence asking the XA to execute a command it otherwise wouldn't.

mamehaze:
alternatively maybe the XA is meant to reset the ARM at that point (probably worth studying the specific command it calls) and they changed the logic on later revisions

cuavas:
> tripslot mostly needs proper OKI banking, and inputs/lamps sorting out correctly, it's somewhat playable as-is though. tripfev seems to be basically the same game on the XA based boards, sound works there, so could just need the same sound emulation copied to tripslot

IGS would never make it that easy.  The sound bank value is split across two different I/O ports for Triple Slot.  It isn’t quite the same, though – Triple Fever seems to be a ticket payout game for the US, while Triple Slot is a cash payout game for Latin America.

PR #12780, opened 2024-09-18
trs/coco3.cpp: Extend display from 639 to 640 pixels

tlindner:
Fixes bug id 8903.


ajrhacker:
Does the vertical display window need to be enlarged too?

tlindner:
I'm going to say "no", becuase all of the pixels are showing when drawing a box to the maximum size.

PR #12783, opened 2024-09-18
Windows NT 3.x/4.0/2000 pristine images

chungy:
MAME had a completely broken warez version of Windows NT 3.1, that was replaced with a known-good unmodified copy. Language variants of Windows NT 3.1 were also added (I don't have the non-English copies of future versions, can't do those at this time). Advanced Server is added to the list.

Windows NT 3.5 was added new.

Windows NT 3.51's CD-ROM image was not the original copy, and likewise updated to a known-good unmodified copy. Server was added to the list.

Windows NT 4.0 was added new, all known-good copies and editions of such.

### DIFF
--- a/hash/ibm5170.xml
+++ b/hash/ibm5170.xml
@@ -4565,153 +4565,241 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="winnt31w">
-		<description>Windows NT 3.1 Workstation (3.10.511.1)</description>
-		<year>1993</year>
-		<publisher>Microsoft</publisher>
-		<part name="flop" interface="floppy_3_5">
-			<!-- Origin: WinWorld -->
-			<dataarea name="flop" size="1474560">
-				<rom name="cdinstall.img" size="1474560" crc="6501bf63" sha1="386d83a2e9fd758272cbbd0beb4148f9e238626e"/>
-			</dataarea>
-		</part>
-	</software>
-
-	<software name="winnt31w_35" cloneof="winnt31w">
+	<software name="winnt31_35">
 		<description>Windows NT 3.1 Workstation (3.10.511.1) [3.5" floppy]</description>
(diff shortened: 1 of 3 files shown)


chungy:
It started to bother me that Windows 2000 was alone without these kind of additions and fixups (ignoring the non-NT Windowses, at least for now... those are an even bigger mess to unravel). Being that Windows 2000 is NT 5.0, it seems to fit the theme of the pull request. I swear it's my last change here. :-)

chungy:
I'd like to note also that NT 3.1, 3.5, 3.51, and 4.0 are all multi-architecture disc images. All of them run on IA-32 (ibm5170_cdrom here), MIPS (mmr4000le), and PowerPC (afaik no MAME driver supports it); NT 3.1 build 528 and all the rest also support Alpha (likewise no MAME driver supporting it). Maybe the images should be duplicated to other software lists. The floppy images are exclusive to IA-32, but the other architectures have firmware that allow direct CD installation.

Windows 2000 is the only single-architecture (IA-32) version here. Actually it's the only version of NT ever that got released only for a single architecture. :)

ajrhacker:
This looks OK, except that it needs a commit message properly describing all the additions.

chungy:
What is lacking in the current commit messages?

angelosa:
Need a brief message in a commit or in the PR body about what has been added here:
```
ibm5170_cdrom:  Windows NT 3.1 (3.10.528.1, MSDN, en-US) [whoever is to be credited for the addition]
```

chungy:
> Need a brief message in a commit or in the PR body about what has been added here:
> 
> ```
> New working software list items
> ---------------------------------------
> ibm5170_cdrom:  Windows NT 3.1 (3.10.528.1, MSDN, en-US), ... [whoever is to be credited for the addition]
> ```

Thanks, I've done this.

PR #12789, opened 2024-09-19
chore: add ATTR_COLD to common lifecycle methods

holub:
add ATTR_COLD to following methods:
```
void x(address_map &y) ATTR_COLD;
void machine_reset() override ATTR_COLD;
void machine_start() override ATTR_COLD;
void video_start() override ATTR_COLD;
void device_start() override ATTR_COLD;
void device_reset() override ATTR_COLD;
const tiny_rom_entry *device_rom_region() const override ATTR_COLD;
void device_add_mconfig(machine_config &config) override ATTR_COLD;
```


happppp:
CI says it doesn't compile, error in src/mame/ibm/rtpc.cpp:147:62.
But that aside: After doing so many of them at once, do you start to see any benefit? In MAME executable size or RAM usage perhaps. Or maybe MAME's startup got a couple of microseconds slower?

Or is this like the old 'fastcall' in the C days where almost all of the time the compiler was smart enough to know, and in the end it's source code clutter?

holub:
I don't expect to see difference in sturtup. Just planned to compare all binaries size after CI build.

happppp:
Does it have any performance loss on lambdas inside address maps or are those unaffected?

holub:
Good question. I didn't plan maps in the first place, but added as we have ATTR_COLD on a lot of them.

cuavas:
In theory, it should marginally improve cache locality while running by grouping the “cold” together within an object file, so they aren’t mixed in with the rest of the functions.  It’s hard to measure these kinds of things with MAME given it’s so big overall.

It shouldn’t affect optimisation of lambdas or local class member functions within a cold function.

The old `DECLARE_` macros for machine config, start, reset, driver init, etc. included `ATTR_COLD`, but it wasn’t carried across most of the time when the code was demacrofied.  It’s something that used to be there but got lost in most places.

> Or is this like the old 'fastcall' in the C days where almost all of the time the compiler was smart enough to know, and in the end it's source code clutter?

What?  The compiler can’t automatically apply “fastcall” for targets that support it (mostly 32-bit x86) as it changes the calling convention to pass arguments in registers.  It definitely does make a difference to use it for leaf functions that are called on the critical path, as it reduces stack churn.  It makes no difference if the function actually ends up being inlined, though.

The “fastcall” calling convention doesn’t make sense for x86-64 where arguments are always passed in registers anyway, but perhaps to compensate, Microsoft has invented “[vectorcall](https://learn.microsoft.com/en-us/cpp/cpp/vectorcall)” for when you want more arguments passed in vector registers.  Once again, it makes a difference if you apply it appropriately, but it doesn’t affect functions that get inlined.

PR #12790, opened 2024-09-19
Add a skeleton for DASH4 Debug Adapter for SH CPUs

clawgrip:
DASH4 Debug Adapter [Arcade Hacker]

### DIFF
--- a/src/mame/skeleton/dash4.cpp
+++ b/src/mame/skeleton/dash4.cpp
@@ -1,0 +1,85 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/***********************************************************************************************************
+Skeleton driver for Cross Products Ltd. DASH4 Debug Adapter.
+
+DASH4 is the hardware debug adapter of CodeScape development environment for the Hitachi SuperH family of
+microprocessors.
+It's an intelligent debug pod that offers a complete low-level debug capability. The DASH4 uses the JTAG
+interface to communicate with the target at 1.5 Mbytes/s via the Hitachi User Debug Interface (HUDI)
+built-in to the SH7750. The DASH4 connects to the host computer via Ethernet to allow high-speed
+communications with the debugging software.
+
+The CodeScape Development system encompasses both software tools and hardware interfaces to provide
+development support for Hitachi SH microprocessors from initial low-level system troubleshooting through
+to the development of complex applications in C++.
+
+It came bundled with the Altera Design Software for PCs "Quartus II":
+https://archive.org/details/altera-quartus-ii-version-2.2-manuals-software/
+
+DASH4 Rev. B PCB (1999):
+      ____________________________________________________________________
+     |                                  _____________       ______       |
+     |     ____________         Xtal    |NEC         |     |LT1129       |___
+     |    |Altera     |      11.059MHz  |D4564323G5__|     |_____|          _|__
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/dash4.cpp
```diff
@@ -0,0 +1,85 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/***********************************************************************************************************
+Skeleton driver for Cross Products Ltd. DASH4 Debug Adapter.
+
+DASH4 is the hardware debug adapter of CodeScape development environment for the Hitachi SuperH family of
+microprocessors.
+It's an intelligent debug pod that offers a complete low-level debug capability. The DASH4 uses the JTAG
+interface to communicate with the target at 1.5 Mbytes/s via the Hitachi User Debug Interface (HUDI)
+built-in to the SH7750. The DASH4 connects to the host computer via Ethernet to allow high-speed
```
angelosa, line 10: `SH7750` doesn't make sense here: the SoC mounted on the provided PCB ASCII below is a HD6417604F28, which is a SH2, cfr. https://github.com/mamedev/mame/blob/master/src/mame/psikyo/psikyosh.cpp and https://github.com/mamedev/mame/blob/master/src/mame/seibu/feversoc.cpp 
Is the `DASH4` printed on PCB?

src/mame/skeleton/dash4.cpp
```diff
@@ -0,0 +1,85 @@
...
+	ROM_REGION(0x00100, "eeprom", 0)
+	ROM_LOAD( "93lc56b.u17",       0x00000, 0x00100, CRC(a05ee68a) SHA1(5dd120cb4a818826ebec8bde59225ccdbf28a57d) )
+
+	ROM_REGION(0x002e5, "pld", 0)
+	ROM_LOAD( "gal22v10d-15lj.u6", 0x00000, 0x002e5, CRC(80cc0146) SHA1(e28da16e18d0c9de48a08651f1c670f946d188e7) )
+ROM_END
+
+} // anonymous namespace
+
+SYST( 1999, dash4, 0, 0, dash4, dash4, dash4_state, empty_init, "Cross Products Ltd.", "DASH4 Debug Adapter", MACHINE_IS_SKELETON )
```
angelosa, line 85: We don't use trading entities for manufacturer field. Please ditch the `Ltd.`

clawgrip: Fixed, thanks

src/mame/skeleton/dash4.cpp
clawgrip: > Is the DASH4 printed on PCB?
Yes, see attached pic.

This text was based on the (archived) product web page...

DASH4 PCB front


DASH4 Codescape information

clawgrip: The 7750 is not the CPU on the DASH4, but the CPU externally connected to the DASH4

p1pkin: I'm wondering how much similar it is to Katana (Dreamcast) Dev.box's Debug Adapter PCB? expected to be quite similar device but with SCSI instead of Ethernet, uses CodeScape software as well.
SegaDC_KatanaMBb_1

yz70s:
This is very intersting, expecially to explore the undocumented sh4 jtag debugging protocol.

If nobody complains, i will merge it myself on 11 october.

PR #12792, opened 2024-09-19
hash/bml3_flop.xml: bulk convert all MP-1802 5.25-inch images to DD

angelosa:
- hitachi/bml3.cpp: drop bml3 prefix for all bus slot options;

### DIFF
--- a/hash/bml3_flop.xml
+++ b/hash/bml3_flop.xml
@@ -17,7 +17,7 @@ Compatibility flags:
 		<sharedfeat name="compatibility" value="5" />
 		<part name="flop1" interface="floppy_5_25">
 			<dataarea name="flop" size="346800">
-				<rom name="hiwriter.d88" size="346800" crc="637f3b94" sha1="0bf7fdb528e03aa2409f386e63e257f5aa013c8f" />
+				<rom name="hiwriter.d88" size="346800" crc="53015bf2" sha1="73cbde8f921203335c1afa68ce4accb5d0be4ce5" />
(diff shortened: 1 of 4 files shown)


hash/bml3_flop.xml
```diff
@@ -17,7 +17,7 @@ Compatibility flags:
 		<sharedfeat name="compatibility" value="5" />
 		<part name="flop1" interface="floppy_5_25">
 			<dataarea name="flop" size="346800">
-				<rom name="hiwriter.d88" size="346800" crc="637f3b94" sha1="0bf7fdb528e03aa2409f386e63e257f5aa013c8f" />
+				<rom name="hiwriter.d88" size="346800" crc="53015bf2" sha1="73cbde8f921203335c1afa68ce4accb5d0be4ce5" />
```
angelosa, line 20: All these images gets interpreted by [d88_disk.cpp](https://github.com/mamedev/mame/blob/f845a619e5ef7fecac6ea03154609e1f5ff8065e/src/lib/formats/d88_dsk.cpp#L440) as QD, which doesn't look the case by looking at the MP-1802 schematics.

image

ajrhacker:
I added the d88 track check recently in 425573a0cbfa357c44e35a4a60c68ae77962a1f4. I may have made it too strict.

angelosa:
I relaxed d88 check in https://github.com/mamedev/mame/commit/26ea9181f4e2ca8a225f6d6aff9ab846e1604787

PR #12793, opened 2024-09-19
Add correct and complete Quake dumps, and id Anthology 4

chungy:
The "quake" entry was a bad dump, it lacked the game's audio tracks even though official shareware releases include them. The main "quake" entry has been replaced with the 1.06 full version, and correct shareware dumps are added in addition.

id Anthology had previously marked discs 3 and 4 as "nodump"; disc 3 of the id Anthology is identical to the standalone Quake 1.06 release (here as "quake"), and disc 4 is Macintosh software, so it is moved to mac_cdrom.xml with the proper hash.

I could, perhaps, include id Anthology disc 3, but since it's identical to "quake", I don't think that duplication is appropriate. It might be worthwhile putting note for the frontend pointing to where it is. I don't know the proper procedure for a disc that's duplicated in two releases.

### DIFF
--- a/hash/ibm5170_cdrom.xml
+++ b/hash/ibm5170_cdrom.xml
@@ -4368,8 +4368,8 @@ Playable demos: "Gobliins 2", "Goblins 3", "Inca", "Lost", "Ween"
 CD-ROM 1: "Catacomb 3-D", "Commander Keen: Invasion of the Vorticons", "Commander Keen: Goodbye Galaxy", "Commander Keen: Aliens Ate My Babysitter!", "Commander Keen: Keen Dreams",
 		  "Dangerous Dave", "Hover Tank 3D", "Rescue Rover", "Rescue Rover 2", "Shadow Knights", "Slordax: The Unknown Enemy", "Spear of Destiny", "The Catacomb", "Wolfenstein 3D"
 CD-ROM 2: "DOOM II: Hell on Earth", "Final DOOM", "Master Levels for DOOM II", "The Ultimate DOOM"
-CD-ROM 3: "Quake"
-CD-ROM 4: "DOOM II: Hell on Earth", "Final DOOM", "The Ultimate DOOM", "Wolfenstein 3D"
+CD-ROM 3: "Quake" (see "quake" -- identical disc to the standalone 1.06 release)
+CD-ROM 4 (mac_cdrom.xml): "DOOM II: Hell on Earth", "Final DOOM", "The Ultimate DOOM", "Wolfenstein 3D"
 ]]></notes>
 		<info name="language" value="English" />
 		<info name="region" value="USA" />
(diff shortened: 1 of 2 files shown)


ICEknigh7:
> and disc 4 is Macintosh software, so it is moved to mac_cdrom.xml with the proper hash.

Would this be a good time to create a `multiplatform_cdrom.xml` softlist for these cases?

Another solution for it could be to have a more general `cdrom.xml` with the compatibility flags choosing which entries to show for each system, but that might be slower and wouldn't look as nice as having separate per-system folders.

rb6502:
PC/Mac multiplatform CDs were common enough in the 90s that we do want to create some sort of shared software list, like how my upcoming "Apple Developer CDs" software list is Apple II/Mac.  I'm not completely sure what it should be named, but the need is definitely there.

FoxhackDN:
> Another solution for it could be to have a more general `cdrom.xml` with the compatibility flags choosing which entries to show for each system, but that might be slower and wouldn't look as nice as having separate per-system folders.

Considering the amount of hybrid PC+Mac discs that are out there, let alone ones with binaries for multiple CPU architectures and operating systems, I think this would be the better choice.

It would take a really long time to set up but it would be worth it.

chungy:
> > and disc 4 is Macintosh software, so it is moved to mac_cdrom.xml with the proper hash.
> 
> Would this be a good time to create a `multiplatform_cdrom.xml` softlist for these cases?
> 
> Another solution for it could be to have a more general `cdrom.xml` with the compatibility flags choosing which entries to show for each system, but that might be slower and wouldn't look as nice as having separate per-system folders.

id Anthology was a single release, but the discs were split clean by platform. The first three were only for the PC, while the fourth was only for the Macintosh.

Multi-platform CDs are definitely a thing, and one such was added in bebfd944ae08470a4ce7e4ce5cead59aaa83b042 -- I just duplicated Chex Quest across both systems. As a singular example, it's maybe OK, but it's a problem in need of a solution when more and more multi-platform discs get added.

I also commented on my [NT pull request](https://github.com/mamedev/mame/pull/12783#issuecomment-2366947490) shortly before your own comment, since NT 3.x/4.0 also need a good multi-platform solution. They don't officially support the Mac platform, but they do support four different architectures with the same CD images.

rb6502:
I feel like it'd be weird to break up id Anthology, regardless of the platforms being a clean break by disc.  So I think it'd be a good thing to start off a multi-platform software list with.

And someone recently wrote the missing HAL to make  NT install and run on some Power Macintoshes.  https://github.com/Wack0/maciNTosh

ICEknigh7:
> id Anthology was a single release, but the discs were split clean by platform. The first three were only for the PC, while the fourth was only for the Macintosh. (and the third is just identical to the "quake 1.06.chd" image I added here, so I opted against duplicating it in the same system...)

For these cases (and other compilations with the original CDs included such as Lucasarts Archives), perhaps allowing MAME to use part aliases could be a solution, something like:
```
		<part name="cdrom3" interface="cdrom" aliasof="ibm5170_cdrom/quake/cdrom1>
		</part>
```

chungy:
> And someone recently wrote the missing HAL to make NT install and run on some Power Macintoshes. https://github.com/Wack0/maciNTosh

I know, but I said _official_ support. Someone else writing the drivers to make NT PowerPC work on a Mac doesn't count as official. :)

PR #12795, opened 2024-09-20
added missing XA opcode for tripfevb, although it still hangs

mamehaze:


mamehaze:
so yeah, the XA command that it calls, which uses this opcode also ends up trying to push out a byte to the serial port (I think)

not sure why this set calls the command and the others don't (probably need to check if it's intentional or not)

I can think of a few possible things that it could need, maybe a system reset after payout (would be a bit weird, but we've seen slot machines that do a reset of the main CPU before) or maybe it needs timers on the XA side emulating, could be it wants to output more bytes over serial and doesn't send the 'finished' signal back to the ARM properly until that has happend.  Oddly I don't think it ever sets anything else up related to the serial port though.

Maybe @HappyASR would know more in this case? I think he made some notes about timer related irqs on the XA when looking at it before.

cuavas:
If it’s trying to send serial data, it might be trying to talk to a centralised audit/bookkeeping system to record payouts.

HappyASR:
@mamehaze It looks like the serial buffer of XA is full, so maybe at least the **UART0_TX/UART 0 Transmitter** interrupt  needs to be implemented here.

PR #12799, opened 2024-09-21
Fixes coin bit order in avengers, except for avengersc

jotego:
The coin slot bit order was reversed in all versions of this machine except for avengersc. This change fixes that on five versions of the game but now breaks avengers. As the driver had 5 versions broken/1 good before, I think this is a better balance.

I do not know how to keep the old order just for avengersc. I have tried a couple of things but cannot get the macros to do what I want.


jotego:
Just realized that the game then reacts to coin 2 credits/coin settings on coin 1. So the service screen reports coin events wrong. No need to make any changes.

PR #12804, opened 2024-09-22
trojan/avengers: sound CPU reset control by main CPU

jotego:
Adds control of the sound CPU reset line by the main CPU, following original Trojan schematics.

image


happppp:
The schematics say bit 5, not bit 4 (and that agrees with writes to this register). I've updated the PR.

jotego:
Thank you. I copied the reset code from the Commando driver and misread it. Thanks for correcting it.

Issue #12806, opened 2024-09-22
CHD CDs are not reproducible between 2048B/2352B sector images that convey the same information


chungy:
### MAME version

0.269

### System information

n/a

### INI configuration details

```INI
n/a
```

### Emulated system/software

n/a

### Incorrect behaviour

In CD-ROM images, sync, address, EDC, and ECC subchannels are normally constructed from well-known methods and MODE1 discs that do not perform any tricks should be reconstructable to the original physical format image from a 2048B sector ("cooked") image. This is the case for the vast majority of MODE1 discs without physical DRM applied to them.

The bin2ecm utility is an example program that takes advantage of this fact to offer high compressability to 2352B sector ("raw") images: https://github.com/chungy/cmdpack/blob/master/bin2ecm.txt

### Expected behaviour

I believe that since CHDs are supposed to represent CD-ROMs, the input of 2048B sector images should end up with the same CHD as a 2352B sector image conveying the same data without any subchannel manipulation.

### Steps to reproduce

Here is an image for demonstration purposes:

[test.zip](https://github.com/user-attachments/files/17091625/test.zip)

If you create a CHD in the two ways, `chdman createcd -i test.toc ...` and `chdman createcd -i test.iso ...`, you get distinct CHDs with different SHA1s. If you mount or burn test.iso and use `cdrdao read-cd --device /dev/srN --read-raw --datafile new.dat new.toc`, you will notice the `new.dat` will match with _identical_ data as in the test archive. The subchannel data is not manipulated and can be reproduced exactly, hence proving its identity with the "cooked" image.

PR #12807, opened 2024-09-23
hash/ibm5170_cdrom: Add Zork Nemesis

chungy:
Ripped from my own CDs :)

### DIFF
--- a/hash/ibm5170_cdrom.xml
+++ b/hash/ibm5170_cdrom.xml
@@ -7686,4 +7686,29 @@ Wave Studio v3.12
 		</part>
 	</software>
 
+	<software name="zorknem">
+		<description>Zork Nemesis: The Forbidden Lands</description>
+		<year>1996</year>
+		<publisher>Activision</publisher>
+		<info name="language" value="English" />
+		<part name="cdrom1" interface="cdrom">
+			<feature name="part_id" value="Disc I" />
+			<diskarea name="cdrom">
+				<disk name="zork nemesis disc 1" sha1="85a6cac392f3648ba02320455690df1196b0bd79" />
+			</diskarea>
+		</part>
+		<part name="cdrom2" interface="cdrom">
+			<feature name="part_id" value="Disc II" />
+			<diskarea name="cdrom">
+				<disk name="zork nemesis disc 2" sha1="566903d9419cb929ed6fd4f230b71bbfb608245a" />
+			</diskarea>
+		</part>
+		<part name="cdrom3" interface="cdrom">
+			<feature name="part_id" value="Disc III" />
+			<diskarea name="cdrom">
+				<disk name="zork nemesis disc 3" sha1="356fcaef31915484cb444529159cb9b69caefef5" />
+			</diskarea>
+		</part>
+	</software>
+
 </softwarelist>
--- a/hash/ibm5170_cdrom.xml
+++ b/hash/ibm5170_cdrom.xml
@@ -7692,21 +7692,21 @@ Wave Studio v3.12
 		<publisher>Activision</publisher>
 		<info name="language" value="English" />
 		<part name="cdrom1" interface="cdrom">
-			<feature name="part_id" value="Disc I" />
+			<feature name="part_id" value="I" />
 			<diskarea name="cdrom">
-				<disk name="zork nemesis disc 1" sha1="85a6cac392f3648ba02320455690df1196b0bd79" />
+				<disk name="zork nemesis i" sha1="85a6cac392f3648ba02320455690df1196b0bd79" />
 			</diskarea>
 		</part>
 		<part name="cdrom2" interface="cdrom">
-			<feature name="part_id" value="Disc II" />
+			<feature name="part_id" value="II" />
 			<diskarea name="cdrom">
-				<disk name="zork nemesis disc 2" sha1="566903d9419cb929ed6fd4f230b71bbfb608245a" />
+				<disk name="zork nemesis ii" sha1="566903d9419cb929ed6fd4f230b71bbfb608245a" />
 			</diskarea>
 		</part>
 		<part name="cdrom3" interface="cdrom">
-			<feature name="part_id" value="Disc III" />
+			<feature name="part_id" value="III" />
 			<diskarea name="cdrom">
-				<disk name="zork nemesis disc 3" sha1="356fcaef31915484cb444529159cb9b69caefef5" />
+				<disk name="zork nemesis iii" sha1="356fcaef31915484cb444529159cb9b69caefef5" />
 			</diskarea>
 		</part>
 	</software>


FoxhackDN:
How are you ripping these, just straight bin+cues? Or using Redump tools?

chungy:
> How are you ripping these, just straight bin+cues? Or using Redump tools?

`cdrdao read-cd --device /dev/sr0 --read-raw --datafile zork1.{dat,toc}` (and repeat for discs 2 and 3)

Then also checked to see if it can be accurately represented as a 2048B sector image (these can). Use that version to make the CHD.

ICEknigh7:
Is cdrdao suitable for game preservation? These comments I just found at the VGPC Discord have me a bit worried about it:

rb6502:
That's the old audio offsets chestnut, which applies to every method of reading a CD (including playing the games on hardware).  There is not a widely available solution for it, just some closed source hoarded software made by one of the dumping groups that you have to use with a specific 25 year old Plextor drive.

angelosa:
https://github.com/superg/redumper I don't think it's closed source if you mean this, the problem being yes, it wants very specific drives and needs cue editing because chdman doesn't know what to do with "MODE/0" it outputs in cuesheets.

FoxhackDN:
Yeah the Redump folks used DiscImageCreator with Plextor drives, a little while back a dev created Redumper, which was intended to be a replacement for DIC, and it works just as well if not better in most cases. It's open source and the developer accepts patches to improve on it.

Since a lot of the software on the lists uses Redump dumps, I just add those. I also dump things in their format using their tools, and I *have* had issues with the cues but it's nothing a quick edit doesn't fix. I haven't really added any CD items to the lists in a while, mostly because the stuff I dump is Windows XP or higher, or PC+Mac hybrids and there still isn't an agreed way of handling those in the lists.

chungy:
> Is cdrdao suitable for game preservation? These comments I just found at the VGPC Discord have me a bit worried about it:

I've never encountered a situation where cdrdao ends up with an image that differs from known-good dumps in the redump.org database, _even for mixed-mode CDs_. Perhaps I got lucky with my choice of disc drive, but I'm not buying the fearmongering messages. :)

Zork Nemesis is as simple as it gets. A single MODE1 data track. No cause for alarm about audio anyway.

chungy:
> https://github.com/superg/redumper I don't think it's closed source if you mean this, the problem being yes, it wants very specific drives and needs cue editing because chdman doesn't know what to do with "MODE/0" it outputs in cuesheets.

well, despite that `HL-DT-ST - BD-RE WH14NS40` is listed in "supported drives" and being my exact drive... it claims that I have an unsupported drive and refuses to do anything.

Now I'm really going to not care. xD

DopefishJustin:
> That's the old audio offsets chestnut, which applies to every method of reading a CD (including playing the games on hardware). There is not a widely available solution for it, just some closed source hoarded software made by one of the dumping groups that you have to use with a specific 25 year old Plextor drive.

This has not been the case for a long time now. [Redumper](https://github.com/superg/redumper) and its predecessor [DiscImageCreator](https://github.com/saramibreak/DiscImageCreator) are open source. There is also ongoing work to support currently sold models of drives from ASUS and LG.

> well, despite that `HL-DT-ST - BD-RE WH14NS40` is listed in "supported drives" and being my exact drive... it claims that I have an unsupported drive and refuses to do anything.

It's a supported drive but you also need to flash it with a supported firmware version: http://wiki.redump.org/index.php?title=Flashing_Asus_BW-16D1HT_firmware

chungy:
> It's a supported drive but you also need to flash it with a supported firmware version: http://wiki.redump.org/index.php?title=Flashing_Asus_BW-16D1HT_firmware

Yeah that's not happening. Not only have I not been convinced that cdrdao produces inaccurate dumps (my own experience shows otherwise), but I'm not going to introduce the possibility of bricking my drive just to use something else.

Besides, Zork just has a single MODE1 data track. All of this debate is pointless.

DopefishJustin:
> Besides, Zork just has a single MODE1 data track. All of this debate is pointless.

Agreed, it is not relevant to this PR specifically and I don't have any objection to merging it. I just wanted to push back against some of the misleading information being posted.

hash/ibm5170_cdrom.xml
```diff
@@ -7686,4 +7686,26 @@ Wave Studio v3.12
...
+				<disk name="zork nemesis disc 2" sha1="566903d9419cb929ed6fd4f230b71bbfb608245a" />
+			</diskarea>
+		</part>
+		<part name="cdrom3" interface="cdrom">
+			<diskarea name="cdrom">
+				<disk name="zork nemesis disc 3" sha1="356fcaef31915484cb444529159cb9b69caefef5" />
+			</diskarea>
+		</part>
+	</software>
+
```
cuavas, line 7710: * Can you try to make the `name` a bit more terse?  We try to jam them into eight characters if possible so there’s plenty of room for suffixes for clones if more versions show up.
* Please add `part_id` for each part to give human-readable disc labels.

Like:
```xml
		<part name="cdrom1" interface="cdrom">
			<diskarea name="cdrom">
				<feature name="part_id" value="What it says on the disc label" />
				<disk name="zork nemesis disc 1" sha1="85a6cac392f3648ba02320455690df1196b0bd79" />
			</diskarea>
		</part>
```

cuavas (changes requested):

hash/ibm5170_cdrom.xml
angelosa: Labels will be displayed when selecting the entry in File Manager. Pretty useful if the game specifically prompts for those.

chungy:
> I just wanted to push back against some of the misleading information being posted. (Which keeps happening - no, a Windows program is not required to flash.)

I had to wait some time to try to write something in a calm mood, but I still find this statement deeply ironic. What you linked contains links to a bunch of *.exe files (aka, Windows programs), which also happen to be all 404s. Even if I wanted to try a dangerous operation on my hardware, I couldn't.

DopefishJustin:
As is often the case with these kinds of things, the wiki documentation is out of date and unclear. Thank you for pointing out the broken links, I have now corrected them. The currently recommended method for all of the JB8 drive models is to use the modified 3.10 firmware, which is available via a [direct .bin link](https://web.archive.org/web/20231007213714/https://cdn.discordapp.com/attachments/429374270607720461/960431754357927946/ASUS-BW-16D1HT-3.10-WM01601-211901041014_MK_edited.bin) at the top of the page. My understanding is that this can be flashed using the [Linux port of MakeMKV](https://forum.makemkv.com/forum/viewtopic.php?f=3&t=224) using the `makemkvcon` command-line tool which is referenced elsewhere on the wiki page, but I have not tried this personally (I last went through the procedure a couple of years ago using Windows). I will see if I can get someone more knowledgeable to update the page.

The issue is that redumper has a fairly small [whitelist](https://github.com/superg/redumper/blob/main/drive.ixx) of known good drive and firmware combinations and the settings to use for each, and yours is most likely not in the list, even though it might have the needed functionality. It should be possible to run redumper without changing the firmware by manually specifying the correct values for `--drive-type`, `--drive-sector-order`, `--drive-pregap-start`, and `--drive-read-method` at the command line, but the details are getting getting beyond the scope of this thread. If you would like to discuss further, I would recommend the [VGPC Discord server](https://www.preservegames.org/p/contact-donate.html) which has a very active PC dumping community.

I mainly just wanted to get across that the Redump project methods are not some kind of unattainable secret sauce, they have been used by hundreds of dumpers on over 100,000 discs with consistent results. For simple data discs it may not be an issue but the CD standard is messy and there are a lot of edge cases that can come up.

sairuk:
@DopefishJustin efforts to correct some of the misinformation in this thread (particularly around availability of project tooling) overall is a good thing.

I have gone through this ASUS crossflashing method on both of my drives. If interested my notes on crossflashing my `LG BH16NS55` to `ASUS_BW-16D1HT_3.02` are [here](https://www.mameau.com/notes/gamehist/dumping/hw/BH16NS55/) but i expect it is the media that will necessitate need.

More generally In lieu of any official software submission specs (that I know of), @ICEknigh7 querying the dumping method seems a reasonable enough particularly when there is no currently known good hash for a submission, for new hashes methods should probably be recorded (in pull requests?) and if necessary challenged where applicable to ensure software lists remain as accurate as possible.

FoxhackDN:
I also asked the question because of the whole 2048/2352 sector size thing with disc images. I believe I was told in an earlier software list submission that disc images *had* to be 2352 sectors. I haven't added anything in a while because all my stuff is too new for MAME (unsupported OS and hardware specs.)

Since these discs were made from the submitter's own discs I wasn't sure if they would match existing dumps. They *might*, though, and that would be helpful for determining which version of the game was dumped and if they matched either of those. (Redump currently has two English versions of the game.) Single track dumps from drives that haven't been tested by the dumping tool developers will usually match Redump verified dumps.

chungy:
> Since these discs were made from the submitter's own discs I wasn't sure if they would match existing dumps.

They match other dumps already confirmed on redump.org, but yes I converted them to 2048B sector images prior to making the CHDs (which affects the CHD SHA1, see #12806). My basic method of doing that is this:

```
$ cdrdao read-cd --device /dev/sr0 --read-raw --datafile zork.dat zork.toc
$ cdemu load 0 zork.toc
$ pv /dev/sr1 > zork.iso
$ eject /dev/sr1 && cdemu load 0 zork.iso
$ cdrdao read-cd --device /dev/sr1 --read-raw --datafile zork2.dat zork2.toc
$ sha256sum zork.dat zork2.dat
```

Basically: With CDemu loading from a 2048B image, it will _regenerate_ all the address, sync, EDC, ECC information on-the-fly as requested by the application (cdrdao in this case). If the original disc never had malformed sectors due to DRM or other schemes, the dat file will end up identical, hence being able to know that the 2048B sector image can accurately represent the original disc.

It gets a little trickier with mixed-mode CDs because I'd have to use `dd` to cut out the data track and perform the same test. I usually don't bother when the data track is small (eg, the quake dumps I submitted in #12793), but I've sometimes done the same on other images stored on my computer.

PR #12809, opened 2024-09-23
thinkpad8xx.xml: Fix floppies sizes

clawgrip:
Also change the driver to use BIOS version as tag instead of using the build code


cuavas:
Please open a new pull request against the release0270 branch with just the floppy size fixes.

clawgrip:
> Please open a new pull request against the release0270 branch with just the floppy size fixes.

Ok!!

PR #12811, opened 2024-09-23
thinkpad8xx.xml: Fix AIX 2nd CD name

clawgrip:
Sorry for the inconvenience…

### DIFF
--- a/hash/thinkpad8xx.xml
+++ b/hash/thinkpad8xx.xml
@@ -19,7 +19,7 @@ license:CC0-1.0
 		</part>
 		<part name="cdrom2" interface="cdrom">
 			<diskarea name="cdrom">
-				<disk name="aix_4.1.5_disc_01" sha1="228afedd90619e89c6b466b4970bda0e5a3e9a13" />
+				<disk name="aix_4.1.5_disc_02" sha1="228afedd90619e89c6b466b4970bda0e5a3e9a13" />
 			</diskarea>
 		</part>
 	</software>
--- a/hash/thinkpad8xx.xml
+++ b/hash/thinkpad8xx.xml
@@ -13,11 +13,13 @@ license:CC0-1.0
 		<publisher>IBM</publisher>
 		<sharedfeat name="compatibility" value="800,820,850" />
 		<part name="cdrom1" interface="cdrom">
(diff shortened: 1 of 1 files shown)


cuavas:
There’s also AIX 4 here: https://archive.org/details/tp850-media/aix-v4/aix-v4-cd2.jpg

Oh, and it would be nice to have proper labels and part numbers for these discs.

clawgrip:
> There’s also AIX 4 here: https://archive.org/details/tp850-media/aix-v4/aix-v4-cd2.jpg

I'll add it for the next release!

Also, the PowerPC boots normally on Big Endian, see the pic (48 00 3F 24 -> branch):

f6f1e5be-e0b9-4c63-b99a-cb0a7987f908

Being AIX and Solaris also Big Endian... Is the Little endian needed for emulating those operating systems?

Thanks!!

(and sorry for the mistakes on this driver)

cuavas:
You can’t run a little Endian OS on a big Endian board and vice versa with older PowerPC systems.  The eight bytes on the 64-bit bus are wired the opposite way around.  The reason is that the chips don’t actually swap the bytes at all, they just swizzle the low bits of the address for accesses smaller than 64 bits to provide little Endian semantics for aligned accesses (kind of like the way MAME handles emulating a system that doesn’t match host Endianness).

Windows NT and OS/2 Warp could only ever run on little Endian platforms, neither was ever portable to big Endian systems.  PReP machines and the ThinkPad 800 series were wired for little Endian operating systems, as Windows NT was envisaged as being a major OS for them.

Solaris for PowerPC only runs on little Endian boards.  You can’t run it on big Endian RS/6000 workstations or PowerMacs.  It was only ever distributed in development kits for the ThinkPad 800 series, you couldn’t buy a copy to use with a PReP desktop system.  (SunOS runs big Endian on SPARC, but it was always portable to little Endian platforms.  The 1988 Sun386i ran SunOS 4 on little Endian 80386 CPUs.)

AIX runs in little Endian mode on the ThinkPad 800 series and PReP desktop systems.  It’s still capable of running big Endian applications as PowerPC CPUs have always supported automatically switching big/little Endian mode on a user/supervisor mode switch.  However, drivers and anything that runs in kernel mode obviously needs to deal with the little Endian system boards.

PR #12815, opened 2024-09-23

angelosa:
Hitachi MB-S1 [archive.org]


angelosa:
Opening a PR for feedback:
1. I feel like I have to move bml3_state and children in an header, then move this system in an own file. By quick glance this will become big and potentially unmanageable pretty fast.
2. I don't like the full memory_map repetition, I've thought that declaring `.m` should be enough to grant memory_views the address it needs?

happppp:
Hmm? Did I accidentally rebase to this branch via sourcetree? Sorry about the trouble.

PR #12816, opened 2024-09-24

angelosa:
Hitachi MB-S1 [Angelo Salese, archive.org]


angelosa:
As per:
> I feel like I have to move bml3_state and children in an header, then move this system in an own file. By quick glance this will become big and potentially unmanageable pretty fast.

Done this way, they changed a lot about video in S1 mode ...

> I don't like the full memory_map repetition, I've thought that declaring .m should be enough to grant memory_views the address it needs?
From what I can tell from documentation, it also needs an extra mirroring for work RAM. For now the repetition stands.

PR #12821, opened 2024-09-25
spec_snqk.cpp: fix wrong format detection; code cleanup

holub:


src/mame/sinclair/spectrum.h
```diff
@@ -154,6 +155,7 @@ class spectrum_state : public driver_device
 	void spectrum_map(address_map &map);
 	void spectrum_data(address_map &map);
 
+	optional_device<snapshot_image_device> m_snapshot;
```
cuavas, line 158: Why add the this member?  Is it ever going to be used after configuration time?

holub: Honestly I don't remember after 2 years. Clearly had some issue when was adding spg but can't tell for sure.

cuavas: I’ll merge it for now anyway as it would be good to get the bug fixes in.  But optional object finders often lead to trouble as they obviously can’t verify that the thing exists as they’re optional, and you still need to be careful about dereferencing them because the object might not exist.  Also, object finders for things that aren’t used after configuration time make things like `-validate` slightly slower for relatively little benefit.

holub: reverted

PR #12831, opened 2024-09-30
Update megadriv.xml for ESPN Hockey Night

PsychoFox11:
I've been looking into this ROM as the trusted dump and scene dumps on no-intro did not match and were marked bad as Mame considers them bad. I did my own testing, as I don't think one db should rely on another or why be separate? Dumping 3 copies on Sanni's OSCR reader, plus desoldering a chip and dumping it with a Backbit chip tester, I get the hashes I'm updating it to here.

While I don't think MAME should consider it bad because No-intro does or vice-versa, I'm the dumper and glad to answer any questions to both projects. If I should be noting this elsewhere please let me know.

### DIFF
--- a/hash/megadriv.xml
+++ b/hash/megadriv.xml
@@ -16674,7 +16674,7 @@ https://segaretro.org/Elitserien_96
 		<part name="cart" interface="megadriv_cart">
 			<feature name="slot" value="rom_sram"/>
 			<dataarea name="rom" width="16" endianness="big" size="2097152">
-				<rom name="espn national hockey night (usa).bin" size="2097152" crc="1d08828c" sha1="cfd65e3ffb17e1718356ef8de7c527e2c9fd8940"/>
+				<rom name="espn national hockey night (usa).bin" size="2097152" crc="401dc618" sha1="cb312d6617574f5b9cccd56a83a0a45b734671ed"/>
 			</dataarea>
 			<dataarea name="sram" size="16384">
 			</dataarea>


cuavas:
So how many bits is “a few”?  Where are they located (e.g. code or data)?  What are the differences?

ICEknigh7:
More details can be found at its no-intro page: https://datomatic.no-intro.org/index.php?page=show_record&s=32&n=0437

PsychoFox11:
> So how many bits is “a few”? Where are they located (e.g. code or data)? What are the differences?

I listed the locations of the two bytes, but here's a diff showing the difference, highlighted in red. What Mame has recorded on top, my dump on the bottom. Again, confirmed from 3 carts and a desoldered chip (and matching other dumps). I suppose its one byte and some bits to be technical. On a binary level, it is 11101111 00000000 vs 11111111 11111111.
image

Happy to answer any other questions! My pics are on my no-intro dumps!

Tafoid:
@PsychoFox11 This was recently committed to the official MAME source files but the data added was not sent in for verification purposes.  Please use the email address mentioned at the link below to send this data or a link to it so that this PR can be completed.  Thank you for your contribution!

Robbbert:
Information has been uploaded.

PsychoFox11:
I just emailed the info as requested but wasn't sure if I needed to based on the last comment (was it uploaded by someone else)? Worst case you got it twice I guess? First time contributing!

PR #12832, opened 2024-09-30
hash/thinkpad8xx.xml: Add AIX 4 (not working)

clawgrip:
thinkpad8xx.xml: IBM AIX 4 [ClawGrip]

Also added some metadata to AIX 4.1.5

### DIFF
--- a/hash/thinkpad8xx.xml
+++ b/hash/thinkpad8xx.xml
@@ -15,14 +15,14 @@ license:CC0-1.0
 		<part name="cdrom1" interface="cdrom">
 			<feature name="part_id" value="AIX 4 Volume 1 of 2" />
 			<feature name="form" value="LK2T0241" /> <!-- FORM on internal .Version -->
-			<feature name="date" value="970227070036" /> <!-- September 29, 2000, on internal .Version -->
+			<feature name="date" value="970227070036" /> <!-- 1997-02-27, on internal .Version -->
 			<diskarea name="cdrom">
 				<disk name="aix_4.1.5_disc_01" sha1="4c49c8190e33c2c7f13a74d7332d384edd6b3042" />
 			</diskarea>
 		</part>
 		<part name="cdrom2" interface="cdrom">
 			<feature name="part_id" value="AIX 4 Volume 2 of 2" />
-			<feature name="date" value="970331130110" /> <!-- September 30, 2000, on internal .Version  -->
+			<feature name="date" value="970331130110" /> <!-- 1997-03-31, on internal .Version  -->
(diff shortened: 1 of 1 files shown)


cuavas (commented):
Can you please not remove the space before the `/>` in self-closing XML tags?

hash/thinkpad8xx.xml
```diff
@@ -14,33 +14,84 @@ license:CC0-1.0
...
 			<feature name="part_id" value="AIX 4 Volume 1 of 2" />
+			<feature name="form" value="LK2T0241" /> <!-- FORM on internal .Version -->
+			<feature name="date" value="970227070036" /> <!-- September 29, 2000, on internal .Version -->
 			<diskarea name="cdrom">
 				<disk name="aix_4.1.5_disc_01" sha1="4c49c8190e33c2c7f13a74d7332d384edd6b3042" />
 			</diskarea>
 		</part>
 		<part name="cdrom2" interface="cdrom">
 			<feature name="part_id" value="AIX 4 Volume 2 of 2" />
+			<feature name="date" value="970331130110" /> <!-- September 30, 2000, on internal .Version  -->
```
cuavas, line 25: Are you sure these are supposed to be UNIX timestamps?  They don’t make sense interpreted that way.  Wouldn’t it be more likely to be 1997-03-31T13:01:10?

hash/thinkpad8xx.xml
```diff
@@ -14,33 +14,84 @@ license:CC0-1.0
...
+		<description>AIX 4</description>
+		<year>1995</year> <!-- Printed on first AIX CD cover -->
+		<publisher>IBM</publisher>
+		<sharedfeat name="compatibility" value="800,820,850" />
+		<part name="cdrom1" interface="cdrom">
+			<feature name="part_id" value="AIX Version 4 for Clients, Ultimedia Services for AIX and Softwindows Volume 1 of 2" />
+			<feature name="part_number" value="40H1045" /> <!-- PN on CD cover -->
+			<feature name="part" value="11H6464" /> <!-- PART on internal .Version -->
+			<feature name="form" value="Power Desktop" /> <!-- FORM on internal .Version -->
+			<feature name="date" value="951031114200" /> <!-- February 20, 2000 -->
```
cuavas, line 42: Same with this – the date doesn’t make sense as a UNIX timestamp, but 1995-10-31T11:42:00 makes sense as a build date.

hash/thinkpad8xx.xml
clawgrip: Yes, absolutely. I've fixed it.

PR #12842, opened 2024-10-02
bus/spectrum: Added The Music Machine

holub:
ports are different from Amstrad version and must be confirmed later


Pernod70:
Are you testing the passthru slot? You need to be overriding all slot handlers to ensure everything is passed through, see https://github.com/mamedev/mame/blob/master/src/devices/bus/spectrum/vtx5000.cpp

Pernod70 (approved):

cuavas:
This is missing save state registration for the data member, but I’ll fix that.

PR #12847, opened 2024-10-03
machines promoted to working (Crazy Bugs Japan ver)

mamehaze:
machines promoted to working
------------------
Crazy Bugs (V103JP) [buffi, Peter Wilhelmsen, XingXing]


cuavas:
Did you actually test this?  It has different DIP switches, and seems to support hopper payout.  It definitely needs its own inputs.

PR #12854, opened 2024-10-04
Add Linksys PAP2 again

Octocontrabass:
I forgot the license header last time.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -42456,6 +42456,9 @@ p8000_16                        //
 @source:skeleton/palestra.cpp
 palestra
 
+@source:skeleton/pap2.cpp
+pap2
+
 @source:skeleton/pcksurfer.cpp
 pcksurfer                       // (c) 2006 Datawind
 
--- a/src/mame/skeleton/pap2.cpp
+++ b/src/mame/skeleton/pap2.cpp
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:David Haywood, Joseph Khilfeh
+// copyright-holders:Joseph Khilfeh
 
 /*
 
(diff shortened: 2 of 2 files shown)


happppp:
Could you use mipsx_cpu_device instead of cpu_device?
What is the source of the 27MHz clock? An XTAL? If yes, use one from emu/xtal.cpp, otherwise maybe add a note clarifying it?

MAME doesn't do "rom dump of chip" for filenames. It does [what's on the label].[pcb location such as ic1, or empty if none]

Octocontrabass:
Yeah, I can use mipsx_cpu_device. Should I make that change in kisssite.cpp too?

The 27MHz clock comes from a crystal. Actually, thanks for reminding me to check that, it should have a multiplier.

Unfortunately I didn't dump the ROM, so I don't know how it was labeled or if it even had a label. It's at position U51 on the board.

happppp:
No you don't need to update other drivers, or is that one by you too?
It's only an issue when the (cpu) device has its own setters/getters that aren't in cpu_device.

Octocontrabass:
Most of the code in pap2.cpp is copied from kisssite.cpp, so I figure they should both use mipsx_cpu_device.

I'm open to suggestions for a better file name, but that's (approximately) the name it was given by whoever dumped it.

cuavas:
Please put the content of the file besides the `SYST` macro in an anonymous namespace.

It isn’t really necessary to use `mipsx_cpu_device` – it will work fine with `cpu_device` (or another base class) if you aren’t actually using anything specific to the MIPS-X CPU, which at this point is basically nothing.

Also, I don’t think this file has any creative content from kisssite.cpp in it, so it doesn’t need David Haywood as a copyright holder.

Octocontrabass:
I won't claim to understand C++ but I'm pretty sure it's already inside an anonymous namespace.

I'll update the copyright header.

PR #12855, opened 2024-10-05
snk/ngp.cpp: Improve rtc behavior. Fixes MT05279.

wilbertpol:


cuavas:
As further cleanup, could the k1ge be modified to implement `device_palette_interface` to make the `"k1ge:palette"` unnecessary?

PR #12856, opened 2024-10-06
k054539: correctly handles reverb RAM as 32kB during test and initialization

jotego:
The external RAM has indeed 32kB, as comments in the MAME source file say. However, the k054539 class is dealing with it as 16kB. The connection to the chip has the upper address bit connected to the K054539 address bit 16, not the expected 14. Checkout the schematics extracted by my team:

image

If you inspect Konami's X-Men test sequence, you will see that it keeps advancing the internal test address register until bit 16 toggles, then it writes test content again. After that, it keeps advancing it until it wraps around and it checks at the two 16kB halves: the one at A16=0 and then A16=1. It stops the test after checking the first 16kB at A16=1 without writting the pointer again (as it is not needed)

This is a simulation of my verilog implementation of X-Men, but you can check the behavior in MAME debugger easily too. See how the orange signal (the test address) is increased until it wraps around and how the test_we signal (indicating writes to the RAM) is active at two specific segments.

image

The PR implements this behavior, deleting the different treatment previously given to RAM and ROM checks. I have tested it with X-Men, Run&Gun (RAM & ROM checks) and Pirate Ship (just booting up the game).

### DIFF
--- a/src/devices/sound/k054539.cpp
+++ b/src/devices/sound/k054539.cpp
@@ -25,7 +25,6 @@ k054539_device::k054539_device(const machine_config &mconfig, const char *tag, d
 	, flags(0)
 	, reverb_pos(0)
 	, cur_ptr(0)
-	, cur_limit(0)
 	, rom_addr(0)
 	, stream(nullptr)
 	, m_timer(nullptr)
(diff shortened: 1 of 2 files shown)


cuavas:
You should really be using unsigned types for the address maths.  I’ll clean it up.

PR #12862, opened 2024-10-08
misc/multfish*.cpp: Minor cleanups

cam900:
- Reduce preprocessor defines
- Reduce runtime tag lookups
- Fix naming
- Make video RAM to unique_ptr
- Fix type values for variables
- Fix spacings
- Reduce static functions

machine/timekpr.cpp: Suppress side effect in debugger read

### DIFF
--- a/src/mame/misc/multfish.cpp
+++ b/src/mame/misc/multfish.cpp
@@ -187,99 +187,95 @@ below are simply made to the banking address to run on other boards.
 #include "speaker.h"
 
 
-TILE_GET_INFO_MEMBER(igrosoft_gamble_state::get_igrosoft_gamble_tile_info)
+TILE_GET_INFO_MEMBER(igrosoft_gamble_state::get_static_tile_info)
 {
-	int code = m_vid[tile_index*2+0x0000] | (m_vid[tile_index*2+0x0001] << 8);
-	int attr = m_vid[tile_index*2+0x1000] | (m_vid[tile_index*2+0x1001] << 8);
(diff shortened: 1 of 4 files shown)


src/mame/misc/multfish.cpp
```diff
@@ -424,256 +419,247 @@ A10 <-> A18
 A12 <-> A13
 */
 
-	uint32_t i,j,jscr,romoffset;
-	uint8_t *igrosoft_gamble_gfx = memregion("gfx")->base();
-	std::vector<uint8_t> temprom(igrosoft_gamble_ROM_SIZE);
+	uint32_t jscr,romoffset;
+	uint8_t *gfx = memregion("gfx")->base();
+	std::vector<uint8_t> temprom(ROM_SIZE);
```
cuavas, line 424: Is `ROM_SIZE` the actually the size of the ROM region?  If it is, you should just get the size of the ROM region directly.  If it’s potentially less than the size of the ROM region, you should add a check to ensure the ROM region is at least that big to avoid potentially writing of the end of the region.

cuavas (changes requested):

mamehaze:
while out of scope for the PR, it should be noted that the banking addresses are different for each board/game, it's part of the protection - the way MAME just hooks up every address right now is incorrect.  At the time I thought it was likely just some kind of mirroring, but it isn't.

some of the bootleg sets are specifically recoded to use different bank addresses

src/mame/misc/multfish.cpp
cam900: It's size of each gfx ROMs (0x80000 bytes * 8 ROMs).

PR #12865, opened 2024-10-08
new WORKING machines (Chaoji Dou Dizhu Plus)

mamehaze:
new WORKING machines
----------
Chaoji Dou Dizhu Plus (S300CN) [XingXing]

### DIFF
--- a/src/mame/igs/igs_m027.cpp
+++ b/src/mame/igs/igs_m027.cpp
@@ -95,6 +95,7 @@ public:
 	void init_lhzb4() ATTR_COLD;
 	void init_gonefsh2() ATTR_COLD;
 	void init_cjddz() ATTR_COLD;
+	void init_cjddzp() ATTR_COLD;
 	void init_zhongguo() ATTR_COLD;
 	void init_klxyj() ATTR_COLD;
 	void init_slqz3() ATTR_COLD;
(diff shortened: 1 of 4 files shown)


mamehaze:
and before any rude/snarky 'did you even test this' comments, yes, I tested it, it boots, I can coin it up, I can select things.

could the dips differ, maybe, but we can hardly trust the test modes on half of these anyway.

mamehaze:
I also removed the clone status on Long Hu Da Manguan Plus as these 'plus' sets are distinct releases, not revisions.

little0000bag:
it is playing same as cjddz.
 key 1 start： playing card
Ctrl : select cards
Alt:  bet or cancel selected carda
Blank: pass

cuavas:
> I have no idea how to play the game.

https://en.wikipedia.org/wiki/Dou_dizhu

cuavas:
The input tests have been comprehensive on all the games that have them.

Also, I don’t think it’s unreasonable to expect that you’d notice that this:
image

is completely different to this:
image

little0000bag:
Yes, the dip switch for Joy stick and Mahjong key changing is SW 2-2  for this new game.

mamehaze:
> The input tests have been comprehensive on all the games that have them.
> 
> Also, I don’t think it’s unreasonable to expect that you’d notice that this: image
> 
> is completely different to this: image

Given that it's all in Chinese, there's nothing I can do with that information anyway, one might be more complete than the other, documenting dips were there anyway but didn't work.

You really expect FAR too much of your devs who do this for fun, and your condescending attitude just stinks.

cuavas:
> Yes, the dip switch for Joy stick and Mahjong key changing is SW 2-2 for this new game.

Yes, which requires its own input port definitions because all the `PORT_CONDITION` settings have to change, because the wiring is different for outputs as well as inputs when using a mahjong panel.

It also has a link feature that needs to be flagged as unemulated.

little0000bag:
I played the  games with Joy stick,not used the Majiang Panel.
But I guess the outputs for playing with Majiang Panel are same.
Here are the manuals for these two games with Chinese language for your refence.

cuavas:
> I played the games with Joy stick,not used the Majiang Panel. But I guess the outputs for playing with Majiang Panel are same. Here are the manuals for these two games with Chinese language for your refence.

They aren’t – see the code here: https://github.com/mamedev/mame/blob/mame0270/src/mame/igs/igs_m027.cpp#L1106

The hopper is controlled with different PPI lines depending on whether it’s wired for a mahjong matrix panel or JAMMA joystick panel.

cuavas:
> one might be more complete than the other, documenting dips were there anyway but didn't work.

It isn’t.  The settings that were on the DSW1 in cjddz are all on DSW2 on this game because DSW1 has been repurposed to support the link feature.  All the DIP switch settings that show in the test modes for either of these games do work.  The input tests 

> You really expect FAR too much of your devs who do this for fun, and your condescending attitude just stinks.

You’re the guy who regularly complains about premature promotions and developers not testing things.  Can you not see the hypocrisy of adding things as working when you aren’t in a position to actually test them?

mamehaze:
A few missing / incorrect DIPs is not the same as something 'not working' or being prematurely promoted.  It's something that's easy enough for somebody with a better understanding of the games / languages to pick up on.

I'll complain about premature promotion if a game plays a single hung note over everything, or has completely broken screens, or visuals that make parts of it literally unplayable.  Likewise for computer systems with no media support at all, or no ability to load the media you can actually find or systems that run so badly performance-wise that there's no proper way to know if they really work as intended.

Your insinuation that I didn't bother to test the last IGS PR when I'd put a significant amount of work in already, and even noted that the ROM label didn't match the test mode was nothing but shitting over those doing the work.

PR #12870, opened 2024-10-11
Allow querying for alternate Qt6 libexec directory

cho-m:
This is closer to how Qt6's moc.prf finds moc, i.e.

```
qtPrepareLibExecTool(QMAKE_MOC, moc)

...

defineTest(qtPrepareLibExecTool) {
    isEmpty(instloc): instloc = "$$[QT_HOST_LIBEXECS]"
    qtPrepareTool($$1, $$2, $$3, $$4, $$instloc)
}
```

---

Specifically adding for Homebrew - https://github.com/Homebrew/homebrew-core/pull/193751

However, I think this will help other repositories/distros that don't use `QT_HOME / libexec / moc`. For example, Arch Linux uses `QT_HOME / moc` and thus needs to modify the script[^1].

Similar handling in Meson[^2] though they also fall back on `QT_INSTALL_LIBEXECS`. Not sure if that is actually needed so left out for now.

[^1]: https://gitlab.archlinux.org/archlinux/packaging/packages/mame/-/blob/main/PKGBUILD?ref_type=heads#L58-59
[^2]: https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/qt.py#L65-L68

### DIFF
--- a/scripts/src/osd/modules.lua
+++ b/scripts/src/osd/modules.lua
@@ -384,12 +384,16 @@ function qtdebuggerbuild()
 			if _OPTIONS["QT_HOME"]~=nil then
 				MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/bin/moc --version 2>/dev/null")
 				if (MOCTST=='') then
-					MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/libexec/moc --version 2>/dev/null")
+					local qt_host_libexecs = backtick(_OPTIONS["QT_HOME"] .. "/bin/qmake -query QT_HOST_LIBEXECS")
+					if (string.starts(qt_host_libexecs,"/")==false) then
+						qt_host_libexecs = _OPTIONS["QT_HOME"] .. "/libexec"
+					end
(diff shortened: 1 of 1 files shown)


scripts/src/osd/modules.lua
```diff
@@ -384,12 +384,16 @@ function qtdebuggerbuild()
 			if _OPTIONS["QT_HOME"]~=nil then
 				MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/bin/moc --version 2>/dev/null")
 				if (MOCTST=='') then
-					MOCTST = backtick(_OPTIONS["QT_HOME"] .. "/libexec/moc --version 2>/dev/null")
+					local qt_host_libexecs = backtick(_OPTIONS["QT_HOME"] .. "/bin/qmake -query QT_HOST_LIBEXECS")
+					if (qt_host_libexecs=='') then
```
cho-m, line 388: May actually need to check this is a path as `qmake` can output messages to stdout, e.g. on non-existent property it outputs `**Unknown**`

Perhaps
```suggestion
					if (string.starts(qt_host_libexecs,"/")~=true) then
```

PR #12876, opened 2024-10-13
bus/a2bus/suprterminal.cpp: allow mapping control via C0Bx reads

SpecLad:
Accessing the C0B2, C0B4 and C0B6 locations is supposed to work whether you read or write to them, according to the Sup'R'Terminal manual.

### DIFF
--- a/src/devices/bus/a2bus/suprterminal.cpp
+++ b/src/devices/bus/a2bus/suprterminal.cpp
@@ -171,7 +171,10 @@ u8 a2bus_suprterminal_device::read_c0nx(u8 offset)
 {
 	if (offset < 8)
 	{
-		access_c0nx(offset);
+		if (!machine().side_effects_disabled())
+		{
+			access_c0nx(offset);
+		}
 	}
 	else if (offset == 9)
 	{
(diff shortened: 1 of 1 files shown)


src/devices/bus/a2bus/suprterminal.cpp
```diff
@@ -173,14 +164,36 @@ void a2bus_suprterminal_device::write_c0nx(u8 offset, u8 data)
...
+u8 a2bus_suprterminal_device::read_c0nx(u8 offset)
+{
+	if (offset < 8)
+	{
+		access_c0nx(offset);
+	}
+	else if (offset == 9)
+	{
+		return m_crtc->register_r();
+	}
```
cuavas, line 179: You need to check `machine().side_effects_diabled()` before doing things with side effects in response to a read.  If you don’t, then having a debugger memory window open will wreak havoc with the system.

cuavas (changes requested):

src/devices/bus/a2bus/suprterminal.cpp
SpecLad: It looks like this check is already done in `apple2_state::c080_r`.

cuavas: Having the check there is a problem in and of itself, as it means you can’t see what would be read from the C0nx range when debugging without forcing a read with side effects.

The slot cards really should all be audited for doing the checks themselves so the check can be removed from the host system and the debugger can become more useful.  Adding more cards without checks means more work later.

SpecLad: Okay, makes sense. Updated.

PR #12877, opened 2024-10-13
New working clone

Osso13:
New working clones
------------------
Chaoji Dou Dizhu (V219CN) [dyq, little0, Guru]

- igs/igs_m027.cpp: renamed Long Teng Hu Yue to Long Teng Hu Yue Duizhan Jiaqiang Ban, as per title screen [dyq]


Osso13:
done as PR because I'm in no position to check if the DIP settings are the same

PR #12882, opened 2024-10-15
machine/mpu401.cpp: return ACK byte from data port when command is re

angelosa:
…

* fix pc98:gaialord MIDI detection

### DIFF
--- a/src/devices/machine/mpu401.cpp
+++ b/src/devices/machine/mpu401.cpp
@@ -206,7 +206,13 @@ uint8_t mpu401_device::asic_r(offs_t offset)
 {
 	if (offset == 0)
 	{
-		m_gatearrstat &= ~STAT_TX_FULL;
+		if (!machine().side_effects_disabled())
+		{
+			m_gatearrstat &= ~STAT_TX_FULL;
+			// pc98:gaialord expects an ACK byte at driver init time
+			// for the command (0xff) it sends.
+			m_mpudata = 0xfe;
+		}
 		return m_command;
 	}
 	else if (offset == 1)


angelosa:
As per <http://midi.teragonaudio.com/tech/mpu.htm> , the host is supposed to receive an ACK byte when a command is issued.

rb6502 (approved):

ajrhacker:
I don't think it makes sense. The MusicQuest PC MIDI Card only sends FE when it receives FF as the command. Then again, it doesn't have an ASIC, just a few bits of glue logic (including a small PLD).

angelosa:
0xff is the command that `pc98:gaialord` sends from host. ACK-ing may be narrower?

That website also mentions:
> Errata
As mentioned, most cards implement only UART mode. But some cheap game cards don't even bother implementing a bi-directional COMMAND/STATUS port. They implement only the STATUS port. (The assumption is that, because UART mode doesn't support any commands except the FF command to knock an MPU back into intelligent mode, there's no reason to have a COMMAND port in a UART only card. Instead, the manufacturer saves a few pennies on parts). The implication here is that if you call set_uart(), it will lock up forever in the again loop (ie, while waiting for the ACK). Why? Because the card will never process any command and therefore never ACK to something that the card completely ignores. That's why I recommend adding the timeouts as noted above to set_uart() if you're not sure that you're dealing with a properly designed MPU Uart mode.

angelosa:
> ACK-ing may be narrower?

ETA: it's actually not the case as per page 13 of the MPU tech manual:
image

ajrhacker:
I just double-checked the Z8 disassembly of the MusicQuest card, and I found this old note of mine:

> 1af5: [...] LD   R8, #FEh [N.B. MPU-401 ACK (not supposed to be returned here!)]

This is part of the routine setting up UART mode, though it waits until the status latch is empty before actually sending this byte. The exit from UART mode upon receipt of an FF command is, as described above, just a normal system reset with no ACK sent.

ajrhacker:
That looks like an editing error. $3F is the command to *enter* UART mode; $FF is the reset command needed to exit it.

angelosa:
I'll close this and acknowledge the issue in the pc98 entry, pending further investigation later on.

PR #12885, opened 2024-10-16
mjsenpu.cpp: Minor cleanups

cam900:
- Use palette_device handlers for palette
- Use std::unique_ptr for VRAM variables
- Make screen update function into cliprect-stricted
- Correct VRAM data type
- Reduce literal tag usage
- Use logerror for logging features
- Reduce duplicates
- Improve function and ROM region namings
- Reduce unnecessary lines
- Add notes for in-game text and voices
- Fix key matrix function, Improve namings for IO related variables

### DIFF
--- a/src/mame/misc/mjsenpu.cpp
+++ b/src/mame/misc/mjsenpu.cpp
@@ -67,11 +67,7 @@ public:
 		, m_hopper(*this, "hopper")
 		, m_mainram(*this, "mainram")
 	//  , m_vram(*this, "vram")
-		, m_mux_9e(*this, "MUX_9E")
-		, m_mux_9d(*this, "MUX_9D")
-		, m_mux_9b(*this, "MUX_9B")
-		, m_mux_97(*this, "MUX_97")
-		, m_mux_8f(*this, "MUX_8F")
+		, m_io_key(*this, "KEY%u", 0U)
 	{
 	}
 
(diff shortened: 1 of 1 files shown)


src/mame/misc/mjsenpu.cpp
```diff
@@ -205,43 +184,45 @@ uint32_t mjsenpu_state::muxed_inputs_r()
...
+		return m_mux_9b->read();
 
 	case 0x97:
-		return ioport("MUX_97")->read();
+		return m_mux_97->read();
 
 	case 0x8f:
-		return ioport("MUX_8F")->read();
+		return m_mux_8f->read();
 	}
```
cuavas, line 200: This isn’t a multiplexer – it’s a switch matrix, using the low five bits to select rows.  Could you please:
* Use a `required_ioport_array` for the five mahjong keyboard rows
* Convert this function to read the rows as a switch matrix

You can find examples of switch matrix input in many mahjong drivers.  The code should end up something like:
```c++
	uint32_t result = 0xffffffff;
	for (unsigned i = 0; m_matrix_inputs.size() > i; ++i)
	{
		if (!BIT(m_mux, i))
			result &= m_matrix_inputs[i]->read();
	}
	return result;
```

cuavas (changes requested):

PR #12889, opened 2024-10-19
ussr/bk: attach floppy

shattered:
cpu/t11: VM1 does not use RMW bus transaction for MOV and MOVB insns (fixes writing to floppy)

### DIFF
--- a/scripts/src/bus.lua
+++ b/scripts/src/bus.lua
@@ -4748,6 +4748,8 @@ end
 
 if (BUSES["QBUS"]~=null) then
 	files {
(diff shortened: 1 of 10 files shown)


Tafoid:
@shattered 

Seems we are missing a couple additional BIOS versions for bk_kmd:
bk_kmd: Unable to find matching ROM disk_253.rom (7a21b887)
bk_kmd: Unable to find matching ROM disk_326.rom (fde7583d)

If you would be so kind as to send the data or a link to it via the email detailed on this page:

Thank you

PR #12890, opened 2024-10-20
scv.xml: Mark 1 item as not working, remove zero offsets, and use hexadecimal sizes.

wilbertpol:
Items marked as not working:
Star Speeder


ArcadeShadow:
No sure about the supported="no" flagged sets. If you try using the Super Vision Cassette (PAL) system, they work well. 
I don't know if those sets were designed to work in PAL or if it's a bug. 
So, maybe the "partial" flag is more adequate.

cuavas:
Maybe these systems need compatibility filters to show which games work on which variant?

wilbertpol:
It seems to be a timing related thing. The scv driver has the cpu at 4MHz and uses a 14.3MHz input clock (which can be seen on pictures of the mainboard) for the video display at 60Hz.
The scv_pal driver has the video running using a 13.4MHz input clock for a display at 42Hz; which is quite likely to not be correct.

Overclocking the scv cpu to 6MHz makes wheelrcr work on the scv driver; changing the display parameters for the scv_pal driver closer to 50Hz also makes it impossible to start a game.

ajrhacker:
Does changing line 9462 of `cpu/upd7810/upd7810_opcodes.cpp` from

> m_ovc0 = 16 * ( TM0 + ( ( TM1 & 0x0f ) << 8 ) );

to

> m_ovc0 = 8 * ( TM0 + ( ( TM1 & 0x0f ) << 8 ) );

improve the situation here?

ArcadeShadow:
A few years back, I've reported a bug concerning that ([MT 08381](https://mametesters.org/view.php?id=8381))

ajrhacker:
940671b78c77cc519348d1a219cd54433efdecee seems to fix most of these games, though not Star Speeder.

hash/scv.xml
```diff
@@ -355,25 +356,26 @@ Information found at http://www.rhod.fr/yeno_epoch.html
...
-				<rom size="8192" offset="0x4000" loadflag="reload" />
-				<rom size="8192" offset="0x6000" loadflag="reload" />
-				<rom name="starspeeder.0" size="32768" crc="511576ee" sha1="1b61132af26dc8d5c1b5e0f72c2e96192bba7e20" offset="0x8000" />
+			<feature name="slot" value="rom64k"/>
+			<dataarea name="rom" size="0x10000">
+				<rom name="starspeeder.1" size="0x2000" crc="7978c4a6" sha1="6bb5ff3946867fa7adb6561dc084ef5a7a92c507"/>
+				<rom size="0x2000" offset="0x2000" loadflag="reload"/>
+				<rom size="0x2000" offset="0x4000" loadflag="reload"/>
+				<rom size="0x2000" offset="0x6000" loadflag="reload"/>
+				<rom name="starspeeder.0" size="0x8000" crc="511576ee" sha1="1b61132af26dc8d5c1b5e0f72c2e96192bba7e20" offset="0x8000"/>
```
cuavas, line 378: I’m not really a fan of having implicit offset for some ROMs and explicit offsets for others, particularly within a single `part` like this.

hash/scv.xml
```diff
@@ -192,11 +193,11 @@ Information found at http://www.rhod.fr/yeno_epoch.html
...
-			<dataarea name="rom" size="65536">
-				<rom name="kungfurd.0" size="32768" crc="9cfc5a67" sha1="1c34470c1f389db7be755e130db298f6b475c7ce" offset="0" />
-				<rom size="24576" offset="0x8000" loadflag="reload" />
-				<rom name="kungfurd.1" size="8192"  crc="a895375a" sha1="644ab7c7fb85d0d80aa3a416b0a5359d395e48a3" offset="0xe000" />
+			<feature name="slot" value="rom64k"/>
+			<dataarea name="rom" size="0x10000">
+				<rom name="kungfurd.0" size="0x8000" crc="9cfc5a67" sha1="1c34470c1f389db7be755e130db298f6b475c7ce"/>
+				<rom size="0x6000" offset="0x8000" loadflag="reload"/>
+				<rom name="kungfurd.1" size="0x2000" crc="a895375a" sha1="644ab7c7fb85d0d80aa3a416b0a5359d395e48a3" offset="0xe000"/>
 			</dataarea>
```
cuavas, line 201: Here’s another case of mixing implicit and explicit offsets withing a `part`.

hash/scv.xml
```diff
@@ -140,10 +141,10 @@ Information found at http://www.rhod.fr/yeno_epoch.html
...
 		<part name="cart" interface="scv_cart">
-			<feature name="slot" value="rom64k" />
-			<dataarea name="rom" size="65536">
-				<rom name="doraemon.0" size="32768" crc="ee0b874d" sha1="425eb6ed31faa2afdecc216145616f577f72db9a" offset="0" />
-				<rom name="doraemon.1" size="32768" crc="d2de91a6" sha1="6aba5ee3d08afad222bec9a176c2607829fa5cc6" offset="0x8000" />
+			<feature name="slot" value="rom64k"/>
+			<dataarea name="rom" size="0x10000">
+				<rom name="doraemon.0" size="0x8000" crc="ee0b874d" sha1="425eb6ed31faa2afdecc216145616f577f72db9a"/>
+				<rom name="doraemon.1" size="0x8000" crc="d2de91a6" sha1="6aba5ee3d08afad222bec9a176c2607829fa5cc6" offset="0x8000"/>
 			</dataarea>
```
cuavas, line 148: And another.

cuavas (commented):
In general, it's a lot easier to read if the offset and size are kept together and come before the stuff that can vary in length.

PR #12891, opened 2024-10-20
mac_cdrom, ibm5170_cdrom: new CD-ROM dumps

chungy:
New working software list items (ibm5170_cdrom)
-----------------------------------------------
My First Math Adventure: Adding and Subtracting [chungy]
Ultimate Solitaire [chungy]

New NOT WORKING software list items (ibm5170_cdrom)
---------------------------------------------------
Multimedia Middle School Grammar [chungy]
Mavis Beacon Teaches Typing Special Edition 10 [chungy]

New working software list items (mac_cdrom)
-------------------------------------------
My First Math Adventure: Adding and Subtracting [chungy]

### DIFF
--- a/hash/ibm5170_cdrom.xml
+++ b/hash/ibm5170_cdrom.xml
@@ -5023,6 +5023,34 @@ https://www.pcgamingwiki.com/wiki/The_Adventures_of_Lomax
 		</part>
 	</software>
 
+	<!-- Windows 95/NT4 -->
+	<software name="mfma_addsub">
+		<description>My First Math Adventure: Adding and Subtracting</description>
+		<year>1998</year>
+		<publisher>Global Software Publishing</publisher>
+		<info name="developer" value="Dorling Kindersley" />
+		<part name="cdrom" interface="cdrom">
+			<diskarea name="cdrom">
+				<disk name="my first math adventure - adding and subtracting" sha1="860fe59ec122aff801de2c724e40a5f0045f88be" />
+			</diskarea>
+		</part>
+	</software>
+
+	<!-- Windows 3.1/95 -->
+	<!-- Program fails to launch properly on ct486/Win95 -->
+	<software name="mmmsgram" supported="no">
+		<description>Multimedia Middle School Grammar</description>
+		<year>1995</year>
+		<publisher>Pro One</publisher>
+		<info name="developer" value="Sofsource" />
+		<info name="language" value="English" />
+		<part name="cdrom" interface="cdrom">
+			<diskarea name="cdrom">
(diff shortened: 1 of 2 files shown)


hash/ibm5170_cdrom.xml
```diff
@@ -5127,6 +5155,21 @@ The Curse of Monkey Island (Windows 95 demo version)
 		</part>
 	</software>
 
+	<!-- Windows 95/98/NT4 -->
+	<!-- Graphics bugged on svga_s3, sound not working properly on sblaster_16 -->
```
angelosa, line 5159: Need more context on this: is it requiring a true color display hence the "bugged"? Post a picture.

angelosa: That's typical true color bug for the (hackish) `svga_s3` (or even, the underlying ibm8514/a adapter). What are the minimum requirements, I expect this to be better suited for `pcipc` considering the '99 release?

chungy: The CD doesn't state any minimums beyond Windows 95 and NT 4.0, but installing on ct486 is extremely slow. I have not tried to do it in pcipc :)

angelosa: Well I'll eventually look at this, just please be more specific with the comment with the `True Color' tag (not asking to translate to a `notes` field because that needs upgrading for `ibm5170_cdrom.xml` anyway).

chungy: Been trying pcipc, but I think the driver is nearly completely broken. Couldn't get Win95 nor NT4 to install; got Windows 2000 installed but even the basic OS will frequently crash the driver. Can't even try to get to the point of installing extra software.

I can append the comment, I didn't know there were known true color bugs; Windows 95 wasn't even set to True Color, but I used svga_s3 because it was the only one Win95 really has a driver for.

angelosa: `pcipc` BIOS is pretty finnicky, it's definitely possible to install Windows 98 on it (and the Jp DOS/V variant) but first off you have to replace PS/2 mouse with a serial one (including disabling former in CMOS). Legit never tried 95/NT4, I know 2000 has issues with the [virge card](https://github.com/mamedev/mame/blob/51a8ec962bef3a44fa6686faeb851722973e2014/src/devices/video/s3virge.cpp#L10) it uses.

chungy: I could try 98, I didn't figure it was worth trying if 95 didn't work. Windows 2000 is fine with real ViRGE cards, so that is almost surely a MAME bug itself :p

chungy: Well, tried twice, no good. Both Windows 98 and Me are locking up in the 16-bit GUI installer just as Win95 does, just as Windows 2000 does with certain events (opening Solitaire or the command prompt are among them).

PR #12895, opened 2024-10-20
igs_m027_023vid.cpp - tweak interrupt timing

mamehaze:


cuavas:
It’s still rather sus with the very approximate vertical blanking time.  Can anyone with a logic analyser scope up some of the IGS games so we can get the actual counts for the video timings?

mamehaze:
well it roughly ties to vblank in / out now, rather than a random scanline, which is maybe more logical than it was before, but yeah, having it measured would be better

PR #12896, opened 2024-10-21
PPC: Fix qnan check

toxieainc:
a) all constants were one digit too short
b) the payload was always expected to be 0 (which is just a special case qnan (Real Indefinite) though)


toxieainc:
Workflow runs failing are not due to my minimal change from what i can tell.

PR #12904, opened 2024-10-24
osd: better socket handling.

SailorSat:
there was a commit (quite some time back) that changed how read operations work (if no data available) -- https://github.com/mamedev/mame/commit/937e416814c9fac8a02efb43ccb4cd2a12b1cf0d

the issue is, that change was made to not confuse "no data available" with "connection closed".
right now, "no data available" will return an error, while "connection closed" returns "no error" (and size of 0)

also it still is confusing, because when a listening socket gets connected to, it ALSO returns "no error" and a size of 0.

as the code already checks if a socket is "readable" via select - and both winsock and posix api state that a recv return value of 0 means a socket was gracefully closed, so rather check for that and return an error.
also "no data available" is - at least for me - not an error at all.

as of my changes, reading "no data" (or a listening socket getting connected) simply returns "no error" with a size of 0 - while reading a closed socket will return connection-reset error.

this makes error handling a lot cleaner (and more logical).

### DIFF
--- a/src/devices/bus/coco/coco_dwsock.cpp
+++ b/src/devices/bus/coco/coco_dwsock.cpp
@@ -143,7 +143,7 @@ u8 beckerport_device::read(offs_t offset)
 			{
 				// Try to read from dws
 				std::error_condition filerr = m_pSocket->read(m_buf, 0, sizeof(m_buf), m_rx_pending);
-				if (filerr && (std::errc::operation_would_block != filerr))
+				if (filerr)
 					osd_printf_error("%s: coco_dwsock.c: beckerport_device::read() socket read operation failed with error %s:%d %s\n", tag(), filerr.category().name(), filerr.value(), filerr.message());
 				else
 					m_head = 0;
--- a/src/osd/modules/file/posixsocket.cpp
+++ b/src/osd/modules/file/posixsocket.cpp
@@ -77,6 +77,10 @@ public:
 				{
 					return std::error_condition(errno, std::generic_category());
 				}
+				else if (result == 0)
+				{
+					return std::errc::connection_reset;
+				}
 				else
 				{
 					actual = std::uint32_t(size_t(result));
@@ -106,7 +110,7 @@ public:
 		{
 			// no data available
 			actual = 0;
-			return std::errc::operation_would_block;
+			return std::error_condition();
 		}
 	}
 
--- a/src/osd/modules/file/winsocket.cpp
+++ b/src/osd/modules/file/winsocket.cpp
@@ -74,6 +74,10 @@ public:
 				{
 					return wsa_error_to_file_error(WSAGetLastError());
 				}
+				else if (result == 0)
+				{
+					return std::errc::connection_reset;
+				}
 				else
 				{
 					actual = result;
@@ -103,7 +107,7 @@ public:
 		{
 			// no data available
 			actual = 0;
-			return std::errc::operation_would_block;
+			return std::error_condition();
 		}
 	}
 


cuavas:
937e416814c9fac8a02efb43ccb4cd2a12b1cf0d fixed a regression from e8bbea1fc6e94e14768509d322f6c624403ffb36.

Prior to e8bbea1fc6e94e14768509d322f6c624403ffb36, reading a socket with no data available returned an error, while a clean end-of-stream returned length zero with no error.  After e8bbea1fc6e94e14768509d322f6c624403ffb36, reading a socket with no data available returned length zero with no error, making it indistinguishable from end-of-stream.

Connection reset isn’t the same as end-of-stream.  Connection reset happens on receiving an RST (bypassing the half-closed state) rather than a FIN-ACK which is the normal sequence to terminate a TCP connection.  It often doesn’t actually come from the peer but from some intermediate router when a connection times out or a route is lost.  You don’t want to confuse connection reset with end of stream.

Also, zero length with no error is the standard way to indicate end of stream, and `operation_would_block` corresponds to `EWOULDBLOCK`, which is what you get if you try to read from a non-blocking file descriptor that has no data ready.  Changing that would make MAME’s behaviour unintuitive as it would go against what developers exist.

This is not the correct fix.

SailorSat:
While I understand your point, but "this is not the correct fix" feels kinda... ignorant.
I agree that a connection reset is different from a clean shutdown, maybe an end-of-file error would be more apropiate.

What would be the "right" fix?
Right now, if I assume "No Error, Size 0" means connection closed, none of my comm board code would work, as every time a listening socket accepts an incoming connection the result also is "No Error, Size 0".
Should a socket that accepted an incoming connection throw some random error?

As far as I know there currently is no way to "peek" an osd file if there is any data available, and I think stuff like
"if (NoError && Size == 0) || (Error && !OperationWouldBlock)" way more unintuitive.

To be honest, I spent debuggin my comm code for quite a while, wondering why it would literally freeze MAME in some cases - read on a "closed socket" did not return any error.

While it is true that a socket api recv returns 0 if a socket was closed, in almost every situation I as a developer would expect an error - be it an EOF error on normal file operations, or some SocketException - in basically everything other enviroment.

In any case, accepting incoming connections must not be confused with "socket shutdown".

cuavas:
> While I understand your point, but "this is not the correct fix" feels kinda... ignorant. I agree that a connection reset is different from a clean shutdown, maybe an end-of-file error would be more apropiate.

There is no such thing as an "end-of-file error" because POSIX signals an end-of-file by returning no error with a length of zero.

> What would be the "right" fix? Right now, if I assume "No Error, Size 0" means connection closed, none of my comm board code would work, as every time a listening socket accepts an incoming connection the result also is "No Error, Size 0". Should a socket that accepted an incoming connection throw some random error?

What happened when a connection was accepted before e8bbea1fc6e94e14768509d322f6c624403ffb36? Did it behave the same way, or was it returning an error in that situation?

The TCP sockets are supposed to be transparent to things that use files. A connection being accepted is another situation where no data is availabl. Logically, it should also return `EWOULDBLOCK` equivalent for compatibility.

> As far as I know there currently is no way to "peek" an osd file if there is any data available, and I think stuff like "if (NoError && Size == 0) || (Error && !OperationWouldBlock)" way more unintuitive.

But that's what you need to do with POSIX `read` and hence what developers expect:
* `EINTR` - interrupted by asynchronous signal before any data was read
* `EAGAIN` or `EWOULDBLOCK` - no data available on a non-blocking file descriptor
* Some other error - "real" error
* Length zero - end of stream
* Length non-zero - data read

> To be honest, I spent debuggin my comm code for quite a while, wondering why it would literally freeze MAME in some cases - read on a "closed socket" did not return any error.

But that's what happens when you reach the end of a file or stream. Reading returns no error with zero length. That happens for regular files as well.

> While it is true that a socket api recv returns 0 if a socket was closed, in almost every situation I as a developer would expect an error - be it an EOF error on normal file operations, or some SocketException - in basically everything other enviroment.

"Normal file operations" don't give an "EOF error". Both POSIX `read` and Win32 `ReadFile` return no error with length zero on reaching the end of a stream. MAME's read/write streams are modelled on POSIX semantics.

> _EDIT_ also, trying to "send" to a closed socket WOULD return an error instead of "written 0". _EDIT2_ there is only a hand full of code in MAME that actually uses those file-sockets, and none of them handles "no error size 0"

That isn't even true - the Model 1 comm device detects this condition as the remote side closing the connection: https://github.com/mamedev/mame/blob/mame0270/src/mame/sega/m1comm.cpp#L591

That's how the regression in e8bbea1fc6e94e14768509d322f6c624403ffb36 was detected in the first place - the Model 1 games were treating "no data available" as the remote side disconnecting.

The TCP sockets are supposed to be transparent to things that work with files, hence the ability to create TCP sockets by specifying magic filenames. Anything that uses files expects a read returning no error with zero length to indicate end-of-file.

> In any case, accepting incoming connections must not be confused with "socket shutdown".

Well you can distinguish between the two with the current behaviour. If you get no error with zero length from a read, it could be connection accepted or end-of-stream. But if you try to read again and get no error with zero length a second time, it's definitely end-of-stream, because a socket-backed `osd_file` will only accept a single connection - if the first one was caused by accepting a connection, the second read will return `EWOULDBLOCK` equivalent or non-zero length in the absence of an error.

So once again, what happened when a connection was accepted before e8bbea1fc6e94e14768509d322f6c624403ffb36? The only time it returned non error and zero length for no data available was for the six days between those two commits, which was clearly a regression that was my fault. No code should have been written assuming behaviour that only last one week.

SailorSat:
I now see why I only stumbled upon that issue recently.

my m1comm code only worked because it actually ignores any errors until the connection was closed.
In linkup phase (m_linkalive == 0) it also ignores "no error, recv 0".

when I started work on the k056230 I got EWOULDBLOCK errors for the first time.
having tinkered with the namco_c139 lately, I decided to update my older stuff to actually check for errors, and those freezes would start to appear, and i failed to see the relation.

Mhm... Alright, I'll change my code accordingly.

I still believe incoming sockets should behave like "no data available" then. but that is a story for another PR i guess :)

PR #12905, opened 2024-10-24
video/upd7220.cpp: command reset shouldn't touch RA parameters

angelosa:
- fix `pc9801:mightyhd` text screen display


angelosa:
Risky change close to 0.271 release, hence the PR. 
image

PR #12911, opened 2024-10-26
ussr/bk: quickload and switchable b&w/color output

shattered:


MooglyGuy:
To make the code more informative, I would rename "extra button" to "color toggle button" or something similar, and rename `update_monitor` to `update_monitor_type`.

Also, why the rename from `screen_update` to `screen_update_10`? Although it's not an override, by convention the function tends not to be changed from `screen_update` unless there is a reason to do so, and in this case it's the only screen-update function in the driver.

shattered:
extra_button rename makes sense.

bk0011 model has a bit more complex video hardware (two framebuffers, palettes, vblank irq).  Some recent releases use mid-screen palette switching, so I'm planning to use a separate screen_update_11 and a  scanline timer to implement that.

MooglyGuy:
At that point, would it make more sense to subclass the driver and simply override the base `screen_update` entirely?

shattered:
Maybe, but I'd like to get the WIP code out of the door first

ajrhacker:
I don't understand why there needs to be a hacky extra input in the keyboard device here.

PR #12915, opened 2024-10-26
Add -rtcdate option to set initial date and/or time for RTCs

ajrhacker:
* Remove system_time class from emu.h and rebase it on util::arbitrary_datetime
* util/timeconv.cpp: Add struct tm converter and day_of_week and day_of_year functions to arbitrary_datetime

### DIFF
--- a/docs/source/commandline/commandline-all.rst
+++ b/docs/source/commandline/commandline-all.rst
@@ -4068,6 +4068,23 @@ Core Misc Options
 
             mame galaga88 -nonvram_save
 
+.. _mame-commandline-rtcdate:
+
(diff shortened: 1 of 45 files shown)


ajrhacker:
I'm putting this out here as a draft PR because, while I think the concept is worthwhile, the code changes are a bit messier than one might hope for. In particular, `system_time::customize` could probably be done in several neater ways, though the approach I used does enforce the format more strictly than others might.

I also had a hard time coming up with a good example for the docs, and finally remembered one of my own drivers. A lot of systems with RTCs don't display the time quickly or easily enough.

PR #12917, opened 2024-10-27
Driver for Moog Source, including an interactive layout. No sound.

m1macrophage:
Hello,

This is my first contribution. I have multiple other synthesizer drivers in various stages of development, which will follow soon. I intend for these drivers to be an educational tool on the digital-analog interface of synthesizers, with a focus on classic ones.

This driver comes with an interactive layout. The digital and digital-analog interface are almost fully emulated.  TODOs in the driver outline some of the gaps.  There is no attempt to emulate the analog audio circuits.

I have been looking at past PRs to get a sense of the feedback, and tried to incorporate it. But I'm sure I missed things, so happy to fix as needed.

I do not have access to ROMs. 

Interactive layout:


rb6502:
I didn't realize you were serious about not having the ROM.  There are 3 versions dumped at http://dbwbp.com/index.php/9-misc/37-synth-eprom-dumps (3.2, 3.3, and 4.11).  That would probably greatly help with getting the digital side correct :-)

m1macrophage:
Thanks for taking a look!

I intentionally avoid downloading ROMs . I have confidence in my schematic-reading abilities, and I expect that
revision 3.3, which is apparently the latest official release, should work just fine :).

I think revision 4.11 is for a third-party extension, so I would not expect it to work with this driver.

galibert:
That's going to make things... complicated.  The aim of Mame is to run the systems, so if you're not going to add the roms we're going to have to, and the instant we do that you can't run the driver anymore.  That would be annoying.

m1macrophage:
I see.

Here is why I hesitate. According to an FAQ on mamedev.org, it might or might not be OK for me to download ROMs. So this requires more research to verify.

If someone has a proper way to retrieve the rev 3.3 checksum, they could follow up in a subsequent PR. If not, this driver still serves its informational and educational goals, but there will be caveats (what @galibert mentioned).

As mentioned earlier, I expect rev 3.3 (from DBWBP or elsewhere) will work just fine :). All non-corrupt rev 3.* images should.

Sorry to be weird about this. Just making sure to stick to emulation and avoid ROMs :)

Unrelated: @galibert, ded2280e3fc62dcce443a93839d5f5a554a983d6 caught my eye. Is it correct to infer it will be possible to map MIDI events to MAME inputs in the future? That would be pretty cool!

MooglyGuy:
With all due respect, I do not understand your position.

Even the version of Moog Music behind this particular synth was defunct as a corporate entity around 35 years ago. The MAME team has to do what is right for itself, which is to not distribute copyrighted content, and to that end it does not.

MAME is not, however, responsible for what individuals across the globe choose to do, with some regularity I might add. I cannot give legal advice, but if you're trying to stay on the correct side of the law, I would suggest never exiting your dwelling. So many places have so many un-repealed, ancient laws that you will quite possibly violate at least some of them simply by taking a quick jaunt down to the corner store.

Therefore this concern, itself, seems a bit concerning, as I wonder if you apply it equally to all aspects of your life.

Robbbert:
Other things to consider are the fact that schematics often contain errors and omissions. How do you know that your code doesn't contain a runtime error or some other bug? This can't be found out until the roms are added and testing performed. And if you can't be bothered to do live testing, then why should the devs bother to add your submission? Think about it.

m1macrophage:
> With all due respect, I do not understand your position.

I get that it might seem excessive, and I agree that this causes a lot of inconvenience. But to each their own.


>  ... if you can't be bothered to do live testing ...

This is a valid concern. I would not accept unverified code, and I don't expect others to do so. 

But I can assure you the necessary due diligence has been done. This is not a case of one getting the code to compile, and shipping it off to others to verify. Too proud for that :). On a related note, the choice of firmware revision in the driver is deliberate.

I have multiple other synth drivers in various stages of development. Mostly classics. This was a large investment in time, including a lot of tedious work (non-trivial layouts) and running into unexpected issues (bug in the mcs51 core). So I fully stand behind this work.

I would be excited to have these drivers become part of MAME. But I recognize that I might be asking for the benefit of the doubt, which I haven't yet earned. So if missing checksums is unacceptable, then that's that. Let me know if that's the case, so I can consider next steps.

If someone has comments about the code or layout, or if someone gave it a spin, very interested to hear about that as well.

Thank you for your time!

galibert:
It's not disqualifying.  We will add the checksums, and I'm sure you'll find a way to handle that in the future.

I'm working on the ladder filter on the polysix, which should make building the sound for this system a reasonable endeavour.  Just been a little busy lately...

m1macrophage:
An owner shared the checksum. It has been added to the PR.

galibert:
Sorry, I was busy with other stuff...

m1macrophage:
Thanks for looking into it!

And no problem at all. I know this was a non-trivial PR. And I understand such reviews compete with hobby time, which itself competes with "real life" time :)

PR #12924, opened 2024-10-30
namco/namcos22.cpp: Move C71 internal ROM and device defines into namco_dsp.cpp, Cleanups:

cam900:
- Fix typename value for some bools
- use BIT helper for bitfields
- Reduce literal tag usages
- Simplify gfx decode layouts
- Fix naming
- Fix some notes


happppp:
FYI shifting has higher priority than & or |, so putting it between parentheses is unnecessary.
It'd be same as changing "a * b + c" to "(a * b) + c"

happppp:
Like I said, I prefer without parentheses. I'll revert that part.

PR #12929, opened 2024-11-01
heathzenith/h89.cpp: Reduce duplicate BIOS declarations

mgarlanger:
Making the BIOS definitions DRY. 

I have several more BIOSes to add. This should make it clearer on followup PRs.


src/mame/heathzenith/h89.cpp
```diff
@@ -1040,122 +1040,133 @@ void h89_mms_state::h89_mms(machine_config &config)
 	m_intr_socket->set_fixed(true);
 }
 
+#define ROM_H17 \
```
rb6502, line 1043: This could be simplified down a bit.  Only h88 has the H88 monitor so it doesn't need to be macro-ized, and the h89 variants have a common set of options (Z17, Ultimeth, etc) that could all be one macro.

mgarlanger: I can undo the H88 one. But for the `h89` and `h89_sigmasoft`, once the left bus is fully implemented, we will be able to remove the h89_sigmasoft system and then all systems would have a unique set of roms.

PR #12935, opened 2024-11-03
Update toaplan2.cpp

PabloArcade:
TP-030SP
The original chip has that exact silkscreen.


ajrhacker:
In the future, please give your pull requests descriptive titles.

PR #12936, opened 2024-11-03
nec/pc9801_kbd.cpp: convert to serial interface

angelosa:
* fix `pc9801vm` getting stuck when reset;
* implements KBDE signal: fixes at least `pc9801:gamepac1:16` to not skip title screen.


angelosa:
1. It doesn't improve anything, it actually manages to make stuck key even more prominent (maybe it's just a `device_serial_interface` without FIFO?). cfr. `pc9801f bokosuka`.
2. keyboard never receives the actual byte (i.e. any write from i8251 gets translated to an 0xff in received_byte)
3. breaks `pc9801rs`: pretends to write a 0x37 to i8251 mode (6 data bits, even parity, no stop), host writes to TX, never gets a TXRDY back.
4. No idea how to connect `RTY` / `RDY` / `RST` to/from i8251 end.

src/mame/nec/pc9801.cpp
```diff
@@ -2142,8 +2148,18 @@ void pc9801_atapi_devices(device_slot_interface &device)
...
+	m_sio_kbd->rxrdy_handler().set(m_pic1, FUNC(pic8259_device::ir1_w));
+
+	clock_device &kbd_clock(CLOCK(config, "kbd_clock", 19'200));
+	kbd_clock.signal_handler().set(m_sio_kbd, FUNC(i8251_device::write_rxc));
+	kbd_clock.signal_handler().append(m_sio_kbd, FUNC(i8251_device::write_txc));
+
+	PC9801_KBD(config, m_keyb, 0);
+	m_keyb->rxd_callback().set("sio_kbd", FUNC(i8251_device::write_rxd));
+	m_keyb->rdy_callback().set("sio_kbd", FUNC(i8251_device::write_cts));
+	m_keyb->rty_callback().set("sio_kbd", FUNC(i8251_device::write_dsr));
```
angelosa, line 2162: image

I may have swapped these two, would be very simple to check with schematics I guess ...

ajrhacker:
Perhaps the chipset in the PC-9801RS uses a degenerate 8251 that omits the mode register, or moves it to a separate address like the V5X SCU.

angelosa:
I guess this is okay for now. It doesn't seem to neither improve nor regress on individual use scenarios described in [hash/pc98.xml](https://raw.githubusercontent.com/mamedev/mame/refs/heads/master/hash/pc98.xml) with `[keyboard]` tags. 
- `alantia` uses `SHIFT` for shooting, but that require another key press already, can't shoot unless you move.
- `dbleagle` just breaks keyboard handling specifically on course select. Need to backtrace code around PC=0x28c0e, supect either copy protection or timing events going awry on transitions.

BIOS traps for keyboard for games just checks for framing/overrun/parity errors. I guess RDY should go high when starting a transmission, I have no clue where that should happen and how it really connects with i8251, would be far easier with actual schematics available.

angelosa:
Put in draft again: typematic works _in alternating make/break sequences_ ...

angelosa:
Pushing to ready for review: I've disabled key repeat as per header notes in pc9801_kbd.cpp.

PR #12942, opened 2024-11-05
added IGS027A and CG ROM dumps to icescape [TeamEurope, Peter Wilhelmsen, XingXing]

mamehaze:


mamehaze:
unfortunately TE can't dump the XA, and the sender didn't dump it beforehand, so leaving it in a non-booting state for now.

mamehaze:
I may add the missing XA ops it needs to boot with the Fearless XA ROM too, as it's more likely if either of them is correct for this game, that one is (as SuperKids has very specific customizations for the trackball) but it's very frustrating to work without a dump that should have been done before the PCB was sent off.

src/devices/cpu/xa/xa_ops.cpp
```diff
@@ -1593,7 +1593,35 @@ void xa_cpu::mulu_byte_rd_rs(u8 rd, u8 rs) { fatalerror( "MULU.b %s, %s", m_regn
...
+
+	if (result & 0xffff0000)
+		set_v_flag();
+	else
+		clear_v_flag();
+
+	if (result == 0x0000)
+		set_z_flag();
+	else
+		clear_n_flag();
```
cuavas, line 1611: This looks suspect – set zero, clear negative?

src/devices/cpu/xa/xa_ops.cpp
```diff
@@ -1593,7 +1593,35 @@ void xa_cpu::mulu_byte_rd_rs(u8 rd, u8 rs) { fatalerror( "MULU.b %s, %s", m_regn
...
+
+	if (result & 0x8000)
+		set_n_flag();
+	else
+		clear_n_flag();
+
+	sr16(rd, result & 0xffff);
+	sr16(rd + 1, result >> 16);
```
cuavas, line 1619: Should `rd + 1` be masked here?

mamehaze: possibly should mask out the lowest bit, manual does say you should use an even rd, but behavior could well be undefined

src/devices/cpu/xa/xa_ops.cpp
mamehaze: looks like a copy+paste from the div, yeah

src/devices/cpu/xa/xa_ops.cpp
```diff
@@ -1593,7 +1593,36 @@ void xa_cpu::mulu_byte_rd_rs(u8 rd, u8 rs) { fatalerror( "MULU.b %s, %s", m_regn
...
+{
+	u32 fullval = gr16(rd);
+	u16 rsval = gr16(rs);
+
+	u32 result = fullval * rsval;
+
+	if (result & 0xffff0000)
+		set_v_flag();
+	else
+		clear_v_flag();
```
cuavas, line 1606: Is this also copy/paste from the divide, where a result that doesn’t fit in 16 bits is an overflow condition, or is the multiply documented to do this?  (A 16 by 16 multiply can’t give a result that doesn’t fit in 32 bits, so it can’t actually cause an overflow.)

mamehaze: muluterms

src/devices/cpu/xa/xa_ops.cpp
mamehaze: I've added a check to make sure this is the case (and that we're not meant to be shifting rd by 1 or anything)

src/devices/cpu/xa/xa_ops.cpp
mamehaze: I believe we match the behavior specified in the document here

cuavas: Yep, that’s the documented behaviour.  It’s just a little unusual compared to x86, etc.

PR #12944, opened 2024-11-06
epson/qx10.cpp: Revert commits that broke zoom handling on the QX10

brijohn:
I noticed when testing #12943 that the zoom diag tests were broken on the QX10, bisecting showed that it was these two commits that broke things.

Revert "epson/qx10.cpp: rectify previous commit"

This reverts commit 35f1ec96be9e097247f13680acb4f62d6e63e538.

Revert "epson/qx10.cpp: use copyrozbitmap for (external to 7220) zoom handling"

This reverts commit a266a60699b4fe34fe14a8d11f2415ee4b48ef46.

### DIFF
--- a/src/mame/epson/qx10.cpp
+++ b/src/mame/epson/qx10.cpp
@@ -40,7 +40,6 @@
 #include "bus/rs232/rs232.h"
 #include "cpu/z80/z80.h"
 #include "imagedev/floppy.h"
-#include "imagedev/snapquik.h"
 #include "machine/am9517a.h"
 #include "machine/i8255.h"
 #include "machine/mc146818.h"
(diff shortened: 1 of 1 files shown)


angelosa:
1. Is `cpm22mf:flop2` diag the only known test for using this?
2. by manually writing thru debugger to port $3a after test ends then the lines will look like this with the revert:

brijohn:
> 1. Is `cpm22mf:flop2` diag the only known test for using this?
> 
>     2. by manually writing thru debugger to port $3a after test ends then the lines will look like this with the revert:
>        

Yes, this is the only known test I am currently aware of.  I believe what you are seeing when manually writing through to the 0x3a register after the test completes is actually correct. The GDC is what actually duplicates the horizontal lines based on the value you set with its zoom command, writing to io register 0x3a only tells the video hardware to basically stretch lines out in the X direction, which is basically what you are seeing by only writing to port 0x3a without also telling the GDC it should zoom.

If you look a the second zoom test you can see how this breaks things, the picture should have maintained its aspect ratio, but instead it gets stretched out vertically after these patches were applied.
zoom

angelosa:
Ok, got fooled by the confusing nomenclature here. What the QX10 `m_zoom` actually does is acting on 2xWCLK interactions, which is a different entity compared to the 7220 internal `zoom` / `m_disp`. I'll rename these variables later.

PR #12948, opened 2024-11-06
Fix cue, toc, gdi parsers

987123879113:
For `parse_cue` and `parse_toc`: The last line of the file gets dropped because the code was checking for `feof` directly after doing a `fgets`, so if you have something like an INDEX command as the very last line without a newline at the end of the file then it would get dropped causing an error, or potentially an improper CHD depending on what data was dropped. The diff looks large for those parts but it's just because I changed the indentation level.

For `parse_gdi`: I rewrote the GDI parser to have a lot more sanity checks + just be more robust than before. Previously it would just trust that there were X number of tracks as specified in the first line of the file. If you change that to be a number larger than could be stored in `outtoc.tracks` then it would segfault. And if the file was blank it'd just cause chdman to get stuck compressing at nan%.

### DIFF
--- a/src/lib/util/cdrom.cpp
+++ b/src/lib/util/cdrom.cpp
@@ -2104,7 +2104,9 @@ std::error_condition cdrom_file::parse_iso(std::string_view tocfname, toc &outto
 
 std::error_condition cdrom_file::parse_gdi(std::string_view tocfname, toc &outtoc, track_input_info &outinfo)
 {
-	int i, numtracks;
+	char token[512];
+	int i = 0;
+	int trackcnt = 0;
 
 	std::string path = std::string(tocfname);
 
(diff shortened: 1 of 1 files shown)


src/lib/util/cdrom.cpp
```diff
@@ -2125,89 +2127,171 @@ std::error_condition cdrom_file::parse_gdi(std::string_view tocfname, toc &outto
...
+		return chd_file::error::INVALID_DATA;
 	}
 
 	if (EXTRA_VERBOSE)
-		for(i=0; i < numtracks; i++)
+		for(int i = 0; i < numtracks; i++)
 		{
-			printf("%s %d %d %d (true %d)\n", outinfo.track[i].fname.c_str(), outtoc.tracks[i].frames, outtoc.tracks[i].padframes, outtoc.tracks[i].physframeofs, outtoc.tracks[i].frames - outtoc.tracks[i].padframes);
+			printf("'%s' %d %d %d (true %d)\n", outinfo.track[i].fname.c_str(), outtoc.tracks[i].frames, outtoc.tracks[i].padframes, outtoc.tracks[i].physframeofs, outtoc.tracks[i].frames - outtoc.tracks[i].padframes);
 		}
```
cuavas, line 2295: If you use the `osd_printf_X` functions, you don’t need `.c_str()` here.

src/lib/util/cdrom.cpp
```diff
@@ -2283,270 +2367,266 @@ std::error_condition cdrom_file::parse_cue(std::string_view tocfname, toc &outto
...
-					{
-						fclose(infile);
-						printf("ERROR: couldn't read [%s] or not a valid .WAV\n", lastfname.c_str());
-						return chd_file::error::INVALID_DATA;
-					}
-				}
-				else
+			/* keep the filename */
+			std::string prevfname = lastfname;
+			lastfname.assign(path).append(token);
```
cuavas, line 2452: Since you’re going to assign to `lastfname` immediately, you should move out of it – `std::string prevfname(std::move(lastfname));`

src/lib/util/cdrom.cpp
```diff
@@ -2283,270 +2367,266 @@ std::error_condition cdrom_file::parse_cue(std::string_view tocfname, toc &outto
...
+			if (prevfname.length() > 0 && lastfname.compare(prevfname) != 0)
+				is_multibin = true;
```
cuavas, line 2454: `std::string` has `empty()` and comparison operators.  Also, you can skip the check if `is_multibin` is already set: `if (!is_multibin && ~prevfname.empty() && (lastfname == prevfname))`.

But actually, can you skip the whole keeping the previous path part if `is_multibin` is already set?

987123879113:
To make the `osd_printf_X` funcs work with chdman I had to add a `output_callback` for `osd_output`. I copied the `output_callback` from `osd_common_t` because `osd_common_t` has way too much unrelated stuff going on for me to be able to just use `osd_common_t`.

src/lib/util/cdrom.cpp
```diff
@@ -2125,89 +2127,170 @@ std::error_condition cdrom_file::parse_gdi(std::string_view tocfname, toc &outto
...
+	const int numtracks = atoi(token);
 
-	for(i=0;i<numtracks;++i)
+	if (numtracks > 0 && numtracks - 1 > std::size(outinfo.track))
 	{
-		char *tok;
-		int trknum;
-		int trksize,trktype;
-		int sz;
+		osd_printf_error("GDI expects too many tracks. Expected %d tracks but only up to %zu tracks allowed\n", numtracks, std::size(outinfo.track) + 1);
```
cuavas, line 2143: You don’t actually need the size qualifiers with the `osd_printf_*` functions.  Using them will cause whatever type is passed in to be cast to the specified size, but omitting them just uses it as-is.

PR #12951, opened 2024-11-07
sdl/window.cpp: Remove assert(0 <= info->clickcnt)

ajrhacker:


ajrhacker:
This assertion triggered rather too often in my testing of att630, using right-click and drag to create windows in the emulated GUI, on a DEBUG=1 build. What is its purpose supposed to be?

cuavas:
It’s inherited from `win_window_info::expire_pointer` (called when handling `WM_POINTERLEAVE` and `WM_MOUSELEAVE`) where that was supposed to be a condition that can’t occur (and apparently doesn’t under normal circumstances as I use Windows debug builds most of the time).  Thinking about it again, the assert may not be that useful on Windows either, but I can’t think of a way to trigger, besides perhaps unplugging a USB pen tablet while the pen is down on the window.  Unfortunately I don’t have a suitable pen tablet with me for testing right now, and won’t for several weeks at least.

PR #12953, opened 2024-11-07
vt1682.cpp: various improvements for the Classic Max Pocket MX-10 and MX-11 sets (plus other units) promote to working

mamehaze:
machines promoted to working
-----------------
Classic Max Pocket PCMX11 - 12 in 1 Colour Games Console (horizontal, France) [David Haywood, AJR]
Classic Max Pocket Mx-10 - 12 in 1 (vertical) [David Haywood, AJR]
NJ Pocket 60-in-1 handheld 'X zero' (NTSC)  [David Haywood, AJR]
Portable Game Station 268-in-1  [David Haywood, AJR]
unknown VT1682-based 101-in-1 handheld (PAL) [David Haywood, AJR]
Toy Story 3 (Lexibook) [David Haywood, AJR]

- hook up a port to allow cmpmx11 to boot
- correct ROM mapping for cmpmx11 so all games can be selected
- add support for SoundCpu -> MainCpu interrupt request so that the SCU can request sample data when it needs it
- added guard against partial vector addresses being fetched which was causing crashes due to invalid IRQ destinations
- set the sound 6502 Stack Pointer to a known value on reset for cmpmx10 / cmpmx11 as the sound programs they upload don't set it and it was causing a huge amount of instability due to the stack corrupting critical data
- added support for an undocumented alt VRAM autoinc mode used by many of the games
- added rowscroll support (games using it work with the latest ALU fixes)
- updated logging in places

along with the ALU fix from AJR this brings many games on the units to playable state with sound, although a few still need work.


src/mame/nintendo/vt1682.cpp
```diff
@@ -72,11 +72,12 @@
 #include "screen.h"
 #include "speaker.h"
 
+#define LOG_DMA              (1U << 4)
 #define LOG_VRAM_WRITES      (1U << 1)
 #define LOG_SRAM_WRITES      (1U << 2)
 #define LOG_OTHER            (1U << 3)
```
cuavas, line 78: Please keep the bits used by the macros in ascending order for the order of declaration.  It’s a lot easier to see that there are no conflicts that way.

src/mame/nintendo/vt1682.cpp
```diff
@@ -240,6 +246,12 @@ class vt_vt1682_state : public driver_device
 	uint8_t m_202e;
 	uint8_t m_2030;
 
+	int m_scu_to_main_irq_active;
+	int m_timera_to_main_irq_active;
+	int m_current_main_vector;
+	int m_new_main_vector;
```
cuavas, line 252: Please use explicitly-sized integers for things that need to be registered for saved states to help keep saved states portable.

src/mame/nintendo/vt1682.cpp
```diff
@@ -5185,6 +5261,21 @@ void vt_vt1682_state::vt_vt1682_sound_map(address_map& map)
 	map(0xfffe, 0xffff).r(FUNC(vt_vt1682_state::soundcpu_irq_vector_hack_r)); // probably need custom IRQ support in the core instead...
 }
 
+uint8_t vt_vt1682_state::vt1682_211c_scuirq_clear_r()
+{
+	m_scu_to_main_irq_active = 0;
+	update_main_interrupts();
+	return 0x00;
+}
```
cuavas, line 5269: Remember to suppress read side effects for debugger accesses.  Also, should this return `space.unmap()`?

src/mame/nintendo/vt1682.cpp
```diff
@@ -5353,8 +5444,19 @@ uint8_t vt_vt1682_state::soundcpu_irq_vector_hack_r(offs_t offset)
...
+
+		uint8_t ret = rom_8000_to_ffff_r((m_current_main_vector - 0x8000) + offset);
+		// redirect to required IRQ!
+		return ret;
+	}
+	else
+	{
+		return rom_8000_to_ffff_r((0xfffe - 0x8000) + offset);
+	}
 }
```
cuavas, line 5460: This isn’t going to show the value that will be read in the debugger, which is confusing.  You’re going to need to do something like:
```c++
	auto vector = m_current_main_vector;
	if (offset == 0)
	{
		vector = m_new_main_vector;
			m_current_main_vector = vector;
	}

	// redirect to required IRQ!
	return rom_8000_to_ffff_r((vector - 0x8000) + offset);
```

cuavas (changes requested):

src/mame/nintendo/vt1682.cpp
mamehaze: well I intentionally coded it because I wanted to see the value in actual ROM space there, that was a choice

src/mame/nintendo/vt1682.cpp
mamehaze: I don't think this really exists, it isn't the address used, and the later documentation has a correction which relocates it to the write address actually used by the software we have, I've removed it

src/mame/nintendo/vt1682.cpp
mamehaze: I'd prefer to leave this as it is, the vectors can be viewed from the drop down menu in the memory viewer anyway, and I think showing a modified version of ROM is more confusing.

mamehaze:
the latest code attempts to use the rowscroll registers, but the only games using rowscroll (all of which are based off the same engine) don't populate the table properly.  The values used to fill the table go through the ALU, so it could be the ALU still has a few other quirks / tricks up its sleeve.

mamehaze:
with AJR's fix to the ALU (already in) and the rowscroll I just added I felt it time to promote a few of the other units, which were being held back due to the racing games on them being entirely broken before.

one similar multi-game unit (wowwg) hasn't been promoted yet as it uses the high colour mode for the menu screen which I still need to add

a few of these have games that are confirmed to be broken on real hardware as they used PAL versions instead of NTSC (or sometimes even mixed them in the same unit!) but obviously as those don't work as expected on hardware they don't block the sets concerned from getting a working flag.

some VT features aren't emulated yet, such as the palette cutthrough effect (not seen it used) and the ability to output the 'palette1' and 'palette2' (which I'm assuming are framebuffers) to different devices rather than mixing them to a single output (again not seen it used, very niche use-case that I doubt anybody took advantage of)

the gm235upc set needs IO stuff on the sound-side, which I believe is separate from I/O on the main side (manual suggests it has different shared pins) but that also has a keyboard etc. which will need emulating (can't remember if it's a touchscreen device too)

intg5410 hasn't been promoted either as it contains some gun games that need the gun emulating (there are likely other standalone units just designed for the gun games, but they're not dumped yet)

I still need to understand the off-by-1 / 2 scroll offsets in certain cases, the VT manual suggests there is no workaround for the problem, but applying it universally is wrong vs real hardware even between games in the same unit, so there probably is something that triggers it (I have suspicions that it might be a side-effect of how the hardware fetches rowscroll data, but when rowscroll is disabled as rowscroll feels like an odd hardware hack with it fetching the table from main ram, not vram)

overall quality control on these VT168 games was poor, which combined with the hardware issues makes knowing what is / isn't an issue difficult at times.  some of the sets that are marked as working such as exsprt48 have very 'buzzy' audio, which again the manual mentions, but tells you to not use the top bit of the DAC if you want to avoid it; these games do use the top bit so I don't know if they coded it in a way that avoids the problem on hardware, but causes it if you handle the DAC normally, or if there's yet something else with the ALU (the sound programs use it heavily when expanding the data, maybe it gives unexpected results if it overflows?)

src/mame/nintendo/vt1682.cpp
cuavas: Viewing the address space is supposed to show what would be read from that address.  It’s confusing for developers if viewing the address space or reading from the address space in a Lua script or debugger script doesn’t  If you want to view ROM, you can view the ROM region in the debugger.

mamehaze: I still think you're overstepping the mark here, I developed this in the way that is least confusing to me, the developer of the driver.  I find what you're suggesting more confusing, and do not wish to change it to be like that.

Again you're coming off as incredibly patronizing here.

mamehaze: Anyway, I've made my point above, I won't be changing this.  It is more beneficial the way I have it.  Does that mean I should just close this whole PR and abandon work as it is never going to be merged as per seems the way around here whenever anybody disagrees with you?

cuavas: Suppressing side effects is supposed to do what it says, i.e. suppress side effects.  If it doesn’t do that, it breaks expectations and confuses people.  When you’re debugging, viewing the address space should show what the CPU will see if it reads any given location.  Side effects have to be suppressed to make that non-intrusive.

Returning a different value for the debugger and scripts breaks that expectation.  At best it will leave people asking, “Why isn’t the CPU seeing the value I’m seeing?”  But more likely it will just leave them confused that the CPU is doing something strange with the value it reads because they won’t consider that the driver would be coded in such a confusing way.

If someone attempting to debug code running on an emulated system in MAME has to wade through MAME’s source to work out why the value read by the CPU isn’t what they see in the address space view in the debugger, that’s broken.

> It is more beneficial the way I have it.

Breaking expectations is not beneficial.  It makes MAME confusing and unapproachable.

> Does that mean I should just close this whole PR and abandon work as it is never going to be merged as per seems the way around here whenever anybody disagrees with you?

If you want to throw a tantrum, fine.

mamehaze: Once the logic is moved to the CPU core the debugger view will be identical to what you see now.

This CPU pulls vectors from difference places, even on the sound CPU side there is no vector mirror at fffx from what people have told me, it simply pulls from 3ffx or 1ffx depending on if internal ROM is enabled or not.

With the IRQ controller in the CPU core, there will be no overlay, it will pull the vectors from the address associated with that interrupt level.

The move to the CPU core is (was) scheduled for a later PR.  Why you insist on this temporary (incorrect) view an interfering with driver development I do not know.

Just don't complain about having too much work to do when you won't allow devs to actually develop without interfering.

Feedback is meant to be a 2 way process, mentioning things that the author might have missed.  In this case it is not something I missed, it is an intentional design.  If you can't respect your developers you're going to very quickly be left with none.

rb6502: If you want this to go in, change it to do what Vas said.  You're perfectly free to make your local copy do the weird stuff again afterwards, but what lands in master has to make sense, and the debugger not showing what the CPU actually sees qualifies as not making sense.

mamehaze: The real CPU sees what the debugger shows now.  It doesn't see the overlay.

I will not be changing this.

mamehaze: If somebody merges this, and the work gets done on the CPU core to just fetch from the proper address it's going to look exactly like it looks now, does that get blocked too?

If you know this uses the VT SoC then you know where to look for the correct vectors, you don't want to see the overlay in the debugger.

mamehaze: I can't stop somebody else from making the change post-merge, which honestly is the most likely thing to happen here, because there's absolutely no reason I should be bending to demands to make the code I have less correct.

I can still not appreciate that level of interference and micro-management and act accordingly.

mamehaze: As a compromise I've added a comment explaining why that logic is correct, and why it is as it is.

PR #12954, opened 2024-11-08
t10mmc: Fix CD-TEXT detection, cdrom: Parse control flags from toc

987123879113:


src/devices/machine/t10mmc.cpp
```diff
@@ -622,8 +626,8 @@ void t10mmc::ExecCommand()
 			auto track_type = m_image->get_track_type(trk);
 
 			// If there's a transition between CD data and CD audio anywhere in the requested range then return an error
-			if ((last_track_type == cdrom_file::CD_TRACK_AUDIO && track_type != cdrom_file::CD_TRACK_AUDIO)
-			|| (last_track_type != cdrom_file::CD_TRACK_AUDIO && track_type == cdrom_file::CD_TRACK_AUDIO))
+			if (last_track_type != -1 && ((last_track_type == cdrom_file::CD_TRACK_AUDIO && track_type != cdrom_file::CD_TRACK_AUDIO)
+			|| (last_track_type != cdrom_file::CD_TRACK_AUDIO && track_type == cdrom_file::CD_TRACK_AUDIO)))
```
987123879113, line 630: `last_track_type != cdrom_file::CD_TRACK_AUDIO` was causing problems here when `last_track_type` was -1 (haven't parsed any tracks yet).

987123879113, line 630: When no tracks have been seen yet (`last_track_type` = -1) and an audio track is encountered as the first track, it would throw an error. Only applies to the READ CD command.

PR #12962, opened 2024-11-11
Various namco 2d drivers cleanups, updates

cam900:
- Various drivers:
  - Fix spacing
  - Use BIT macro for bit flags
  - Suppress side effects in debugger read
  - Reduce literal tag usages
  - Reduce unnecessary lines
- namco/namco_c123tmap.cpp:
  - Use generic gfx layout
  - Use reference in delegate
  - Fix notes
- namco/namco_c169roz.cpp:
  - Use generic gfx layout
  - Use reference in delegate
- namco/namcofl.cpp:
  - Fix maincpu ROM region size match to memory map
  - Fix notes
- namco/namconb1.cpp:
  - Split namco NB2 hardware and Point Blank/Gun Bullet driver state
  - Reduce preprocessor defines
  - Fix typename values
- namco/namcos1.cpp:
  - Split quester driver state for reduce optional finders
  - Fix typename value for save states
- namco/namcos2.cpp:
  - Fix typename consistency
  - Simplify gfx decode layout
- namco/namcos2_roz.cpp:
  - Use generic gfx layout
- namco/namcos2_sprite.cpp:
  - Use device_gfx_interface for gfx decode
- shared/namco_c355spr.cpp:
  - Use generic gfx layouts
  - Fix notes


src/mame/namco/namconb1.h
```diff
@@ -39,44 +38,31 @@ class namconb1_state : public driver_device
...
 	virtual void machine_reset() override ATTR_COLD;
 	virtual void video_start() override ATTR_COLD;
 
-private:
 	int m_gametype = 0;
 	enum
 	{
 		/* Namco NB1 */
-		NAMCONB1_NEBULRAY = 0x2000,
+		NAMCONB1_NEBULRAY = 0,
```
cuavas, line 65: This changes behaviour so it defaults to applying hacks for Nebulas Ray if the game type is not set.  Previously, no hacks would be applied if the game type was not set (keeping initial zero value).  The default value zero should not correspond to any per-game hack configuration.

src/mame/namco/namcos1_m.cpp
```diff
@@ -942,12 +954,20 @@ u8 namcos1_state::berabohm_buttons_r(offs_t offset)
...
+			}
 			if (m_strobe == 0)
-				m_input_count = (m_input_count + 1) % 5;
+			{
+				if (!machine().side_effects_disabled())
+					m_input_count = (m_input_count + 1) % 5;
+			}
 		}
 
 		// status bit, used to signal end of pressure sensitive button reads
```
cuavas, line 973: This has a couple of issues.

Firstly, the way this is here, everything inside the strobe update block is conditional on side effects being disabled.  You may as well just put the `!machine().side_effects_disabled()` test around the whole block.

But if you do that, you’ll still have the issue that debug accesses won’t return what a real read would return, because a real read would update the strobe value.  You need calculate the updated values, and only commit them if side effects are not disabled.  Something like this (please test this if you use it, I’m coding in a browser here):

```c++
		u8 strobe_count = m_strobe_count + 1;
		u8 strobe = m_strobe;
		u8 input_count = m_input_count;
		if (strobe_count >= 4)
		{
			strobe_count = 0;
			strobe ^= 0x40;
			if (!strobe)
				input_count = (input_count + 1) % 5;
		}

		{
			m_strobe_count = strobe_count;
			m_strobe = strobe;
			m_input_count = input_count;
		}

		if (input_count == 3) res |= 0x10;
		res |= strobe;

		return res;
```

src/mame/namco/namcos1_m.cpp
```diff
@@ -996,9 +1016,12 @@ u8 namcos1_state::faceoff_inputs_r(offs_t offset)
...
-		if (++m_strobe_count > 8)
+		if (!machine().side_effects_disabled())
+			++m_strobe_count;
+		if (m_strobe_count > 8)
 		{
-			m_strobe_count = 0;
+			if (!machine().side_effects_disabled())
+				m_strobe_count = 0;
 
 			res |= m_input_count;
```
cuavas, line 1027: The code immediately below this sets `m_stored_input` – isn’t that going to be a side effect you don’t want triggered by debugger accesses?

src/mame/namco/namcos2_m.cpp
```diff
@@ -264,7 +266,7 @@ uint16_t namcos2_state::namcos2_68k_key_r(offs_t offset)
 		case 1: return 0x110;
 		case 4: return 0xBE;
 		case 6: return 0x1001;
-		case 7: return (m_sendval==1)?0xBE:1;
+		case 7: return (m_sendval == true) ? 0xBE : 1;
```
cuavas, line 269: You don’t need to test Booleans with `== true`, you can just do `m_sendval ? 0xbe : 1`

src/mame/namco/namcos2_m.cpp
```diff
@@ -279,14 +281,18 @@ uint16_t namcos2_state::namcos2_68k_key_r(offs_t offset)
...
 			break;
 		case 7:
-			if( m_sendval == 1 ){
-				m_sendval = 0;
+			if (m_sendval == true)
+			{
+				if (!machine().side_effects_disabled())
+					m_sendval = false;
 				return 0x13F;
 			}
```
cuavas, line 297: Same here – just test `if (m_sendval)`.

src/mame/namco/namcos1_m.cpp
cam900: Fixed via https://github.com/mamedev/mame/pull/12962/commits/c114f4ac1c47f926bff57c636d37821595e50a72 .

src/mame/namco/namcos1_m.cpp
cam900: Fix side effect via https://github.com/mamedev/mame/pull/12962/commits/c114f4ac1c47f926bff57c636d37821595e50a72 .

src/mame/namco/namcos1_m.cpp
```diff
@@ -996,39 +1018,49 @@ u8 namcos1_state::faceoff_inputs_r(offs_t offset)
...
+					stored_input[1] = m_io_in[3]->read() & 0x18;
 					break;
 
 				default:
-					m_stored_input[0] = 0x1f;
-					m_stored_input[1] = 0x1f;
+					stored_input[0] = 0x1f;
+					stored_input[1] = 0x1f;
 					break;
 			}
```
cuavas, line 1050: You don’t need the local `stored_input` – you can skip this switch statement entirely if side effects are disabled since the values of the stored inputs aren’t used for calculating the `res` value in this path of the `if (strobe_count > 8)`.

PR #12963, opened 2024-11-11
Game Boy: add new BIOS variants and gbplayer (not working GBA clone)

chungy:
These files are all floating around the No-Intro sets.

For gbcolor, a Japan-exclusive version of the BIOS is added.  The variant of the BIOS from the Game Boy Advance is also included, which unlocks features in a few games that use it to detect running on GBA hardware.

A beta version of the GBA BIOS has been added, which features a different boot up sound and lacks the ability to escape to network boot mode (start+select in the released console).

The Game Boy Player has been added as a GBA clone.  At least the BIOS checksum is documented here, but it fails to boot in the MAME gba driver.

### DIFF
--- a/src/mame/nintendo/gb.cpp
+++ b/src/mame/nintendo/gb.cpp
@@ -1243,8 +1243,23 @@ ROM_END
 
 ROM_START(gbcolor)
 	ROM_REGION(0x800, "maincpu", 0)
-	ROM_LOAD("gbc_boot.1", 0x0000, 0x0100, CRC(779ea374) SHA1(e4b40c9fd593a97a1618cfb2696f290cf9596a62)) // Bootstrap code part 1
-	ROM_LOAD("gbc_boot.2", 0x0100, 0x0700, CRC(f741807d) SHA1(f943b1e0b640cf1d371e1d8f0ada69af03ebb396)) // Bootstrap code part 2
+
+	// The international Game Boy Color hardware (CGB-001)
+	ROM_SYSTEM_BIOS(0, "cgb-v1", "Game Boy Color v1 CGB-001")
+	ROMX_LOAD("gbc_v1_boot.1", 0x0000, 0x0100, CRC(779ea374) SHA1(e4b40c9fd593a97a1618cfb2696f290cf9596a62), ROM_BIOS(0) ) // Bootstrap code part 1
+	ROMX_LOAD("gbc_v1_boot.2", 0x0100, 0x0700, CRC(f741807d) SHA1(f943b1e0b640cf1d371e1d8f0ada69af03ebb396), ROM_BIOS(0) ) // Bootstrap code part 2
+
+	// (Older?) Japanese units used another version of the BIOS
+	ROM_SYSTEM_BIOS(1, "cgb-v0", "Game Boy Color v0 (Japan) CGB-001")
+	ROMX_LOAD("gbc_v0_boot.1", 0x0000, 0x0100, CRC(7edc621b) SHA1(8fc5f2e25b2cf480d378e4463c89bd0aecdfb21f), ROM_BIOS(1) )
+	ROMX_LOAD("gbc_v0_boot.2", 0x0100, 0x0700, CRC(2a16a329) SHA1(03c0cbfcb7e6a239e25cfb90cafae81087c465f6), ROM_BIOS(1) )
+
+	// Game Boy Advance includes a modified version of the BIOS code.
+	// Some games (shantae, wendy, zeldaage, zeldasea) use CPU register values
+	// set by it to enable GBA-exclusive features.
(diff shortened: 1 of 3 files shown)


ICEknigh7:
I'd suggest replacing "beta" with "prototype" in the code and descriptions.

cuavas (commented):
I really don’t want a hacky Gameboy Player as a clone of the GBA.  The Super Game Boy as a clone of the Game Boy is already bad enough – it can’t possibly emulate the parts that rely on SNES-side functionality.  We don’t need another one of these.

I’m not sure “GBC with bootstrap from GBA” is really something we want to support, either.  It isn’t a configuration you’d ever actually see.  Ultimately we’ll want to support the backwards compatibility mode on the GBA.  But adding that bootstrap ROM so it doesn’t get lost might not be too bad, as it isn’t creating a fake machine.

chungy:
> I really don’t want a hacky Gameboy Player as a clone of the GBA. The Super Game Boy as a clone of the Game Boy is already bad enough – it can’t possibly emulate the parts that rely on SNES-side functionality. We don’t need another one of these.

It's fair enough, but the Super Game Boy really does add a lot of integration between host console and expansion device. The Game Boy Player is only a GBA and doesn't take any functionality from the GameCube system (for the most part, the player disc adds a little bit of UI).

Can drivers interface with each other? the SNES driver should probably include the Super Game Boy expansion, referencing the Game Boy driver for the core Z80 et al emulation. If the GameCube driver ever stops being a skeleton, the same could be done for Game Boy Player.

> I’m not sure “GBC with bootstrap from GBA” is really something we want to support, either. It isn’t a configuration you’d ever actually see. Ultimately we’ll want to support the backwards compatibility mode on the GBA. But adding that bootstrap ROM so it doesn’t get lost might not be too bad, as it isn’t creating a fake machine.

It's the configuration you get from playing GB/GBC games on a GBA/GBA-SP, and it provides actual game play changes in a few titles (desiring to get to the Advance shop in Zelda Oracles was the motivator for me). Yeah, maybe the GBA driver could offer the backwards compatibility mode that effectively loads the GBC driver anyway. If that's possible, let's go for it. (I suppose then the L/R buttons to stretch/unstretch the screen could even work...)

src/mame/nintendo/gba.cpp
```diff
@@ -1496,7 +1496,12 @@ void gba_robotech_state::gbadv_robotech(machine_config &config)
 
 ROM_START( gba )
 	ROM_REGION( 0x4000, "maincpu", 0 )
-	ROM_LOAD( "gba.bin", 0x000000, 0x004000, CRC(81977335) SHA1(300c20df6731a33952ded8c436f7f186d25d3492) )
+
+	ROM_SYSTEM_BIOS(0, "agb-v1", "Game Boy Advance (AGB-001)")
+	ROMX_LOAD( "gba.bin", 0x000000, 0x004000, CRC(81977335) SHA1(300c20df6731a33952ded8c436f7f186d25d3492), ROM_BIOS(0) )
+
+	ROM_SYSTEM_BIOS(1, "agb-proto", "Game Boy Advance (prototype)")
```
angelosa, line 1503: Is this just a ROM swap or there's more that warrants a separate romset ID?

src/mame/nintendo/gb.cpp
```diff
@@ -1243,8 +1243,23 @@ ROM_END
...
+	ROM_SYSTEM_BIOS(1, "cgb-v0", "Game Boy Color v0 (Japan) CGB-001")
+	ROMX_LOAD("gbc_v0_boot.1", 0x0000, 0x0100, CRC(7edc621b) SHA1(8fc5f2e25b2cf480d378e4463c89bd0aecdfb21f), ROM_BIOS(1) )
+	ROMX_LOAD("gbc_v0_boot.2", 0x0100, 0x0700, CRC(2a16a329) SHA1(03c0cbfcb7e6a239e25cfb90cafae81087c465f6), ROM_BIOS(1) )
+
+	// Game Boy Advance includes a modified version of the BIOS code.
+	// Some games (shantae, wendy, zeldaage, zeldasea) use CPU register values
+	// set by it to enable GBA-exclusive features.
+	ROM_SYSTEM_BIOS(2, "agb", "Game Boy Advance AGB-001")
+	ROMX_LOAD("gbc_gba_boot.1", 0x0000, 0x0100, CRC(3c2ef5a4) SHA1(8c43360930ebc42a909cff736800606c697d59ea), ROM_BIOS(2) )
+	ROMX_LOAD("gbc_gba_boot.2", 0x0100, 0x0700, CRC(5f940766) SHA1(ba7ab25014e23d8b902b514571e1e21cc42d370f), ROM_BIOS(2) )
```
angelosa, line 1262: This needs to be loaded in gba (eventually) not in gbcolor.

PR #12970, opened 2024-11-13
Rename to copa 98 marywu.cpp

flama12333:
https://youtu.be/DRYMaNwW5BY
same pcb https://www.castillodealadino.com/images/fotos/FOTOS%20DE%20PCB/Copa%2098.JPG im not sure if the name are right?


mamehaze:
from what I can tell it's just a generic board that you can buy different face skins for

PR #12973, opened 2024-11-13
sinclair/byte.cpp: separate clone

holub:
Added:
* КР580ВИ53 sound
* keyboard mapping
* ROM management


src/mame/sinclair/byte.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+	bool m_1f_gate;
+};
+
+u8 byte_state::spectrum_rom_r(offs_t offset)
+{
+	if (m_1f_gate)
+	{
+		const u16 adr66 = ((offset >> 7) & 0xff) | ((m_io_comp->read() & 1) << 8);
+		const u8 dat66 = memregion("dd66")->base()[adr66];
```
ajrhacker, line 57: It would be better to use a `required_region_ptr` here.

src/mame/sinclair/byte.cpp
```diff
@@ -0,0 +1,281 @@
...
+		if (~dat66 & 0x10)
+		{
+			u16 adr71 = ((dat66 & 0x0f) << 7) | (offset & 0x7f);
+			return memregion("dd71")->base()[adr71];
```
ajrhacker, line 61: ...and here, too.

src/mame/sinclair/spectrum.cpp
```diff
@@ -628,14 +628,14 @@ INPUT_PORTS_START( spectrum )
 	PORT_START("LINE0") /* 0xFEFE */
 	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("CAPS SHIFT") PORT_CODE(KEYCODE_LSHIFT) PORT_CHAR(UCHAR_SHIFT_1)
 	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("z    Z    :      LN       BEEP   COPY") PORT_CODE(KEYCODE_Z)     PORT_CHAR('z') PORT_CHAR('Z') PORT_CHAR(':')
-	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME(u8"x    X    £      EXP      INK    CLEAR") PORT_CODE(KEYCODE_X)  PORT_CHAR('x') PORT_CHAR('X') PORT_CHAR(U'£')
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("x    X    £      EXP      INK    CLEAR") PORT_CODE(KEYCODE_X)    PORT_CHAR('x') PORT_CHAR('X') PORT_CHAR(U'£')
```
cuavas, line 631: The `u8` qualifier is needed for portability when the string contains characters that don’t fit in eight bits (pound sterling symbol).

holub: Ok. Will update the driver as well as it has manu utf8

cuavas:
Some of the characters in key labels will get nuked by srcclean – I’ll fix it after merging.

PR #12976, opened 2024-11-14
igs_fear.cpp : confirmed Icescape XA as the same as fearless [Team Europe]

mamehaze:
- confirmed Icescape XA as the same as fearless [Team Europe]
- added some XA opcode groups that get executed with this now


mamehaze:
I've made no attempt to map the rest of the inputs, I don't know if the Touchscreen is optional or required for this game.

PR #12981, opened 2024-11-16
fuuki/*.cpp: Split tilemap features from fuuki hardware to device

cam900:
fuuki/fuukispr.cpp: Use generic gfx decode layout, Improve naming (it just emulates sprite hardware)
fuuki/fuukifg2.cpp: Fix comments
fuuki/fuukifg3.cpp: Reduce duplicates

### DIFF
--- a/src/mame/fuuki/fuukifg2.cpp
+++ b/src/mame/fuuki/fuukifg2.cpp
@@ -48,7 +48,8 @@ To Do:
 
 #include "emu.h"
 
-#include "fuukifg.h"
+#include "fuukispr.h"
+#include "fuukitmap.h"
 
 #include "cpu/z80/z80.h"
 #include "cpu/m68000/m68000.h"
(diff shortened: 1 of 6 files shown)


src/mame/fuuki/fuukifg2.cpp
```diff
@@ -358,39 +209,20 @@ void fuuki16_state::sound_command_w(u8 data)
...
 	map(0x800000, 0x800001).portr("SYSTEM");
 	map(0x810000, 0x810001).portr("P1_P2");
 	map(0x880000, 0x880001).portr("DSW");
 	map(0x8a0001, 0x8a0001).w(FUNC(fuuki16_state::sound_command_w));
-	map(0x8c0000, 0x8c001f).ram().w(FUNC(fuuki16_state::vregs_w)).share(m_vregs);
-	map(0x8d0000, 0x8d0003).ram().share(m_unknown);
-	map(0x8e0000, 0x8e0001).ram().share(m_priority);
+	map(0x8c0000, 0x8c001f).rw(m_fuukitmap, FUNC(fuukitmap_device::vregs_r), FUNC(fuukitmap_device::vregs_w));
+	map(0x8d0000, 0x8d0003).rw(m_fuukitmap, FUNC(fuukitmap_device::unknown_r), FUNC(fuukitmap_device::unknown_w));
+	map(0x8e0000, 0x8e0001).rw(m_fuukitmap, FUNC(fuukitmap_device::priority_r), FUNC(fuukitmap_device::priority_w));
```
cuavas, line 225: Is the mapping similar enough in the two drivers that you could implement a device map function and install it like:
```c++
	map(0x8c0000, 0x8effff).m(m_fuukitmap, FUNC(fuukitmap_device::regs_map));
```
and avoid needing to know the details of the layout in each driver?

cuavas (commented):
The sprite and tilemap devices are both fairly small – should they be kept together in a single header/source file (e.g. fuukifgvid.h and fuukifgvid.cpp) rather than having a pair of smaller files for each one?

PR #12988, opened 2024-11-17
cdicdic: Fix restarting audio_map after stopping it

Slamy:
Fixes:
https://mametesters.org/view.php?id=8833

I also thought it would fix this one but my assumption was wrong:
https://mametesters.org/view.php?id=8955

Further tests might be needed but I'm not a big fan of Hotel Mario and therefore unable to complete this game as it is quite repetitive.

### DIFF
--- a/src/mame/philips/cdicdic.cpp
+++ b/src/mame/philips/cdicdic.cpp
@@ -706,7 +706,7 @@ void cdicdic_device::process_audio_map()
 	else
 	{
 		m_decode_addr = 0xffff;
-		m_audio_sector_counter = m_audio_format_sectors;
+		m_audio_sector_counter = 0;
 	}
 
 	if (was_decoding)


MooglyGuy:
Can you please check all of the Quizard parent sets in MAME with this change as well (quizard, quizard2, quizard3, quizard4)?
Have you at least played Hotel Mario far enough to get to the same point in the game as the video clip which is on MT 08833?

I suggested that you submit a pull-request for this as you had told me that this fixed MT 08955. If there isn't a guarantee that it fixes the Hotel Mario issue, and there are no other games you can cite that this clearly fixes, then without more comprehensive proof that this doesn't cause other titles to *re*gress, I'm not sure this should be merged.

What, quite simply, does this actually, provably, fix?

Slamy:
Here is a video of Hotel Mario with the Fix:
https://youtu.be/9400Xb6LdMQ?t=45

Here is a video of Hotel Mario as it was in c75845b1ef01d76379bcc0a6937f1ca678484c68:
https://youtu.be/sRrDfg3F2u4?t=37

The error can be caused with this version of MAME by "wiggling" the joystick around even after the first levels score screen as shown in the video.
You can also clearly hear the audio issue as well.
On mametesters I've provided an audio sample recorded from a CDi 210/20 to compare this against to.

The CDIC driver enters an invalid state. I don't yet 100% understand why this happens but the effects are visible.
Having a coding of 0xff seems to disable the audiomap and the current implementation doesn't reflect that.

I can't play Hotel Mario until the specified location because this game makes me aggressive and it would increase my blood pressure to unhealthy amounts.
I also don't know anything about the Quizard hardware and therefore I don't know how it should behave.

The change in question was derived by reading through the source code of the CDIC component of cdiemu and noticing the discrepancy between both emulators.

Slamy:
One thing that might still be important. Even with the fix, the "coin sample" starts very sharp on MAME while it does have some in-fade on real hardware. This is visible when looking at the sample in audacity:
image
The upper is the real machine, the lower is MAME after the fix. I assume that the missing Attenuation is causing this but I'm not sure yet.

Slamy:
I've checked quizard. The game shows no signs of problems. The reason for it being not affected is a certain pause after ever 0xff coding until it is restarted again.
I then went over to quizard2. The engine seems to be the same. The coding for playback is always 0x04. 0xff is followed by a pause again.
I then loaded quizard3. They finally had another opening music. The game has different sounds but the engine was again the same. Same behavior.
Then over to quizard4. The same as before.

In general CD-i titles that use the audiomap can be divided in two. One kind does continues supply of sampledata. This would be Zelda Wand of Gamelon or Nobilia. Or there are games like Frog Feast or the Score Screen of Hotel Mario which permanently stops and restarts it.

PR #12992, opened 2024-11-18
konami/bottom9.cpp: Use memory view instead trampolines

cam900:


src/mame/konami/bottom9.cpp
```diff
@@ -300,7 +276,10 @@ void bottom9_state::sound_bank_w(uint8_t data)
 void bottom9_state::main_map(address_map &map)
 {
 	map(0x0000, 0x3fff).rw(FUNC(bottom9_state::k052109_051960_r), FUNC(bottom9_state::k052109_051960_w));
-	map(0x0000, 0x07ff).rw(FUNC(bottom9_state::bankedram1_r), FUNC(bottom9_state::bankedram1_w));
+	map(0x0000, 0x07ff).view(m_k051316_view);
+	m_k051316_view[0](0x0000, 0x07ff).rw(m_k051316, FUNC(k051316_device::read), FUNC(k051316_device::write));
+	m_k051316_view[0](0x0000, 0x07ff).view(m_k051316_rom_view);
+	m_k051316_rom_view[0](0x0000, 0x07ff).r(m_k051316, FUNC(k051316_device::rom_r));
```
cuavas, line 282: Is it better to use a view nested inside a view like this, or would it be better to have a single layer view with one entry for regular 051316 read/write and a second entry for ROM readback mode?

The code would end up something like:

```c++
	// bit 5 = RAM bank
	if (BIT(data, 5)
	{
		m_palette_view.disable();
		m_k051316_view.disable();
	}
	else
	{
		m_palette_view.select(0);

		// bit 4 = enable 051316 ROM reading
		m_k051316_view.select(BIT(data, 4));
	}

...

	map(0x0000, 0x07ff).view(m_k051316_view);
	m_k051316_view[0](0x0000, 0x07ff).rw(m_k051316, FUNC(k051316_device::read), FUNC(k051316_device::write));
	m_k051316_view[1](0x0000, 0x07ff).rw(m_k051316, FUNC(k051316_device::rom_r), FUNC(k051316_device::write));
```

PR #12995, opened 2024-11-19
ibm5170_cdrom.xml: New software items added

FoxhackDN:
Corel Linux Deluxe [Foxhack]
Civilization: Call to Power (Limited Edition, v1.1, Linux) [Foxhack]

### DIFF
--- a/hash/ibm5170_cdrom.xml
+++ b/hash/ibm5170_cdrom.xml
@@ -2267,6 +2267,35 @@ Offset [CDDA] audio, does not stop when exiting game (including getting to Windo
 		</part>
 	</software>
 
+	<!-- Linux -->
+	<!-- Source: http://redump.org/disc/118951/ -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition).cue" size="1368" crc="66417039" sha1="b7eab95eb3fbf42cbd2bec4a433ed1214d16039f" /> -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition) (Track 1).bin" size="418517232" crc="e7a144e7" sha1="539d5a0ea16970d6e9420a0a077e741d2f0ff6c3" /> -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition) (Track 2).bin" size="22407504" crc="b3d73bca" sha1="ffdb673f3c697176500dd6c5e5cee4c76cc83cc6" /> -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition) (Track 3).bin" size="21525504" crc="6e0add33" sha1="e53db937c7f84afb9e3c743a17fb753ada608738" /> -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition) (Track 4).bin" size="29816304" crc="e6510dc0" sha1="4d545259bf8eadad87e736c756704776829396d9" /> -->
+	<!-- <rom name="Civilization - Call to Power (USA) (Linux) (Limited Edition) (Track 5).bin" size="20819904" crc="9bc13589" sha1="b3fd40f818bbfd84c5a05c838266d427eaeba3d8" /> -->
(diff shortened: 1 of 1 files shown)


FoxhackDN:
I dumped all of these discs myself using Redump methods (Plextor + Redumper) except for the floppy, which I pulled from the installation disc iso. The floppy booted fine in ct486 but I couldn't get past the "starting linux" screen. Some of the CDs are bootable, but I don't know how to create a bootable PC, haha.

I set everything to supported="partial" since things might not work very well.

hash/ibm5170_cdrom.xml
```diff
@@ -6983,6 +7011,59 @@ Incredibly slow in tearing down even on pcipcs7 (bypass with fast forward)
...
+			<diskarea name="cdrom">
+				<disk name="corel_linux_os_deluxe_source_code" sha1="f435e52979b833c3ac00aaa05ddfdaf41bee5d8a"/>
+			</diskarea>
+		</part>
+		<part name="flop1" interface="floppy_3_5">
+		<!-- Floppy image extracted from Installation CD-ROM, located in BOOT\BOOT1440.IMG -->
+			<dataarea name="flop" size="1474560">
+				<rom name="corel_linux_boot_floppy.img" size="1474560" crc="4ac616ea" sha1="2f7999a3fd24b59b6939b94af47cf6cc599f3da0" />
+			</dataarea>
+		</part>
```
angelosa, line 7064: I assume the installer is an [El-Torito](https://wiki.osdev.org/El-Torito) complaint CD, so it's very expected that it won't boot on `ct486` as-is, but rather needs using something like https://github.com/chenall/grub4dos . I have my doubts that it will boot anyway (commercial Linux released in '99 has probably already dropped 486 support), so it would rather just need either `pcipc -bios 2` or `shutms11`.

In short, please just ditch this floppy entry.

hash/ibm5170_cdrom.xml
```diff
@@ -2267,6 +2267,34 @@ Offset [CDDA] audio, does not stop when exiting game (including getting to Windo
...
+]]></notes>
+		<info name="version" value="1.1" />
+		<info name="language" value="English" />
+
+		<part name="cdrom" interface="cdrom">
+			<diskarea name="cdrom">
+				<disk name="civilization_call_to_power_le" sha1="94ad4b9e24bec7b32e32d1efe801c45025f673f5" />
+			</diskarea>
+		</part>
+	</software>
```
angelosa, line 2297: Entry needs the `<sharedfeat name="platform" value="Linux" />`

hash/ibm5170_cdrom.xml
FoxhackDN: The box did include a floppy disk, but I couldn't dump it (I don't have the hardware) so I added this in its place. Should I still remove it, or add a note that it needs dumping from the correct media?

corellinux2

angelosa: Ok, mark it as `status="baddump"` and add a note, along the lines of `<!-- baddump: has an actual boot disk for non El-Torito systems -->`

hash/ibm5170_cdrom.xml
```diff
@@ -6983,6 +7012,59 @@ Incredibly slow in tearing down even on pcipcs7 (bypass with fast forward)
...
+			<diskarea name="cdrom">
+				<disk name="corel_linux_os_deluxe_source_code" sha1="f435e52979b833c3ac00aaa05ddfdaf41bee5d8a"/>
+			</diskarea>
+		</part>
+		<part name="flop1" interface="floppy_3_5">
+		<!-- Package included a boot floppy for non El-Torito systems, but could not be dumped. This image was extracted from the Installation CD-ROM, located in BOOT\BOOT1440.IMG -->
+			<dataarea name="flop" size="1474560">
+				<rom name="corel_linux_boot_floppy.img" size="1474560" crc="4ac616ea" sha1="2f7999a3fd24b59b6939b94af47cf6cc599f3da0" status="baddump" />
+			</dataarea>
+		</part>
```
cuavas, line 7065: FWIW that’s the “virtual floppy” image used for El Torito floppy emulation mode.  There’s no guarantee it’s the same as the real supplied floppy.

FoxhackDN: It actually *is* a custom version, with a Corel Linux bootup screen. I agree that it might not match the floppy 1:1, but it's probably close to the real thing.

PR #12996, opened 2024-11-19
segacd.xml: Added a dump of Sonic CD (USA)

FoxhackDN:
Sonic CD (USA, alt 3) [Foxhack]

### DIFF
--- a/hash/segacd.xml
+++ b/hash/segacd.xml
@@ -3252,7 +3252,7 @@ Requires [disc swap]
 		<rom name="Sonic CD (USA) (Track 34).bin" size="16233504" crc="b9f018f8" />
 		<rom name="Sonic CD (USA) (Track 35).bin" size="32111856" crc="4194928d" />
 		-->
-		<description>Sonic the Hedgehog CD (USA)</description>
+		<description>Sonic the Hedgehog CD (USA, R1C)</description>
 		<year>1993</year>
 		<publisher>Sega</publisher>
 		<info name="serial" value="4407"/>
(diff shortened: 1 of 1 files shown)


FoxhackDN:
Disc was dumped from my own copy with Redump methods (Plextor + Redumper); like the other alts, the only difference is in Track 13, and the ringcode.

cuavas:
Please give pull requests descriptive titles.

cuavas:
So is this a different version, a different pressing of the same version, or different in some other way?  If possible it’s better to have descriptive disambiguating text rather than nondescript “alt 1”, “alt 2”, etc.

FoxhackDN:
> So is this a different version, a different pressing of the same version, or different in some other way? If possible it’s better to have descriptive disambiguating text rather than nondescript “alt 1”, “alt 2”, etc.

The only difference in all the USA alt versions is one single music track. For some reason, Track 13 has a different CRC despite being the same song. If you look at the redump hashes in the software list...

```
(soniccd) <rom name="Sonic CD (USA) (Track 13).bin" size="14013216" crc="c41a848e" />
(soniccda) <rom name="Sonic CD (USA) (RE125) (Track 13).bin" size="14013216" crc="282d4f23" />
(soniccdb) <rom name="Sonic CD (USA) (RE125) (Alt) (Track 13).bin" size="14013216" crc="8e9aaeaf" />
(soniccdc) <rom name="Sonic CD (USA) (RE125) (Alt 2) (Track 13).bin" size="14013216" crc="50722cff" />
```

Every other file matches 100%. I was actually confused why the disc I've owned for over 15 years had this one different track until I looked at the ringcodes:

```
soniccd: <feature name="cd_matrix" value="SEGA4407 R1C MFD BY JVC"/>
soniccda: <feature name="cd_matrix" value="SEGA4407RE125 R7D MFD BY JVC"/>
soniccdb: <feature name="cd_matrix" value="SEGA4407RE125 R8C MFD BY JVC"/>
soniccdc: <feature name="cd_matrix" value="SEGA4407RE125 R4C MFG BY JVC"/>
```

Redump also has a dump with "SEGA4407 R3C MFD BY JVC" in the ringcode. - http://redump.org/disc/94140/

I didn't want to change the setnames based on the ringcode because it'd probably cause problems so I just added it as alt 3.

cuavas:
What format is the track ripped in?  Plain audio samples, or including subchannel data?  Is it a difference in audio, or only in subcode data?

FoxhackDN:
Standard Bin+cue with Redumper. I *also* dumped it with DiscImageCreator to make sure it wasn't a fluke, and it matched.

cuavas:
> I didn't want to change the setnames based on the ringcode because it'd probably cause problems so I just added it as alt 3.

Better to change the descriptions to include R1C, R7D, etc. rather than “alt 1”, “alt 2”, etc. to make it more obvious which is which.  Short names can stay the same.

cuavas:
> Standard Bin+cue with Redumper. I _also_ dumped it with DiscImageCreator to make sure it wasn't a fluke, and it matched.

That doesn’t answer the question though as I’m not intimately familiar with those tools – are the audio tracks in a format that only includes the audio samples, or does it also include subcode data?

FoxhackDN:
It's just audio samples AFAIK, and redumper gives you a scrambled data file.

PR #12999, opened 2024-11-19
cpu/lh5801: Don't add PC value twice when disassembling relative addressing modes

goldnchild:
The calculations for RelP and RelM include adding the pc which is incorrect I think.  Removing the "pc+" seems to fix the calculation.


goldnchild:
pos is first set to the pc, then it's incremented when it gets the opcode, then pos is incremented when it gets the offset.  I think the offset is calculated from that position.

... 
offs_t lh5801_disassembler::disassemble(std::ostream &stream, offs_t pc, const data_buffer &opcodes, const data_buffer &params)
{
        offs_t pos = pc;
     ...

        oper=opcodes.r8(pos++);       //   incremented when it get the opcode
        entry=table+oper;

   ...

case RelP:
                        temp=opcodes.r8(pos++);         // incremented again when it gets the offset, which is put into temp

cuavas:
Oh, it’s adding the PC *twice*.  That makes more sense.

PR #13003, opened 2024-11-20
Wrap tilemap - fix harddunk background in attract

mahoneyt944:
Hi, 

Referencing the open question removed in this commit, adding wraparound on X correctly displays the background when the players are presented during attract mode.

Should this also be added to the Y?

Thank you.

### DIFF
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -33,10 +33,6 @@
       effect to draw the boxing ring over NBG0.
       Same deal for ga2 when in stage 2 cave a wall torch is lit.
 
-    - harddunk draws solid white in attract mode when the players are presented.
-      NBG0 is set with $200 on center X/Y, same as above or perhaps missing
-      tilemap wraparound?
-
     - Wrong priority cases (parenthesis for the level setup):
       dbzvrvs: draws text layer ($e) behind sprite-based gauges ($f).
       dbzvrvs: Sheng-Long speech balloon during Piccoro ending (fixme: check levels).
@@ -744,7 +740,7 @@ void segas32_state::update_tilemap_zoom(screen_device &screen, segas32_state::la
 	srcy += (m_videoram[0x1ff14/2 + 4 * bgnum] & 0xfe00) << 4;
 
 	/* then account for the destination center coordinates */
-	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum], 10) * srcxstep;
+	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum] & 0x1ff, 10) * srcxstep;
 	srcy -= util::sext(m_videoram[0x1ff32/2 + 2 * bgnum], 9) * srcystep;
 
 	/* finally, account for destination top,left coordinates */


mahoneyt944:
Fixes harddunk but breaks alien3 towards the end of the tunnel level when the platform drops

rb6502:
There's likely a register bit somewhere that determines if this should wrap or not.

mamehaze:
this is the problem with system32

it's very much like Taito F3, in that unless you've actually got the boards to test with you're going to find 50 cases where you can fix one thing, but something elsewhere will break by that very change, with no obvious register responsible for a change in behavior.

trying to fix it without boards will just result in stacks of hacks upon hacks.

it's further muddied by the protected games, because the later suicide protections, which have full bus access are doing suspiciously little, and could even be responsible for copying register data about (system 16 games with similar style MCUs do exactly that)

also when it comes to things like 'wraparound' MAME presents a very simplified concept, when in reality there can be other factors involved depending on the hardware implementation - some won't actually wrap *infinitely* and with wraparound turned off, especially with zooming also involved it might not be obvious which co-ordinates come into play in the calculations.  The Namco drivers seem to suffer from that (problems with the battleship in Metal Hawk, which clearly isn't meant to wraparound as you get repeats and 'phantom' ships, but where turning wraparound off causes it to not render in many positions properly at all)

once you get to the 95% correct state for emulation the only real way forward is usually with the boards in front of you and extensive custom programmed tests.

mamehaze:
Title Fight is odd, I'm sure I remember the ring being reasonable at one point, before Aaron's rewrite of the driver (same for some of the line effects used later in Jurassic Park)

However I could be thinking of code I had that was never submitted even back then because it caused problems elsewhere.

galibert:
It may be interesting to note that the closest to system32 seems to be
saturn.  Perhaps looking at the vdp2 could give some inspiration.

mahoneyt944:
With jpark, specifically in the raptor tunnel, what's interesting is the flash light effect seems to display correctly for player 2, but not player 1. Both players must be alive to see player 2's effect though. That maybe a point of interest. Not sure how many people noticed that as most testing was probably on player 1. In the pic below I moved the player 1 crosshairs up so you can see player 2's effect.
jpark-241122-012043

angelosa:
That honestly sounds two cliprects where one of the two is overriding the composition. Something is also amiss here, it shouldn't cut at that line.

Title Fight ring/ga2 torches are known to use a linescroll/lineselect cliprect. I remember trying it but the underlying code didn't cooperate at all, how things are built under the hood doesn't have much room for extensions ...

mahoneyt944:
> it shouldn't cut at that line.

Player 1's flash light effect is drawing the entire length of the bitmap in X while the Y height is more proportional. The previous pictures I posted was to illustrate player 2's effect working with the player 1 crosshairs moved out of the way. Player 1 by itself draws like this if that's the cutoff line you are referring to in the previous picture.

jpark-241122-053057

mahoneyt944:
Referencing:
https://github.com/mahoneyt944/mame/blob/1049f59d19805385467ec7b7a40b50dd968cab8e/src/mame/sega/segas32_v.cpp#L21

Using 10 bit resolution causes harddunk's attract mode background issue while it displays correctly at 9. It seems hit or miss for which games work at which resolution. Could it be that there's variable resolution based on the zooming factor?

angelosa:
Likely controlled by a video register, cfr. sections 4.6 and 4.7 of the Saturn VDP2 manual.

mahoneyt944:
I found this video which describes hooking up a generic jamma harness to a jpark board and notes that moving the joystick up and down changed the camera view, and button 2 lowered the health bars. I confirmed this behavior by swapping the inputs on jpark with a generic two player set. This makes me think that the "flash light" effect in the raptor tunnel might not be the video code at all but an input issue since the effect follows the positioning of the input.

Panning higher and lower:
jpark-241210-234240
jpark-241210-234210

Flash light effect wrong for player1 and correct for player2 (both players must be alive to see player 2's effect.)
jpark-241210-234555
jpark-241210-234635


Video starting around 1:50
https://youtu.be/l__YufFcaSQ?si=7NPpF26CkZO8tiwz

Should this behavior be documented?

mahoneyt944:
I found this video which describes hooking up a generic jamma harness to a jpark board and notes that moving the joystick up and down changed the camera view, and button 2 lowered the health bars. I confirmed this behavior by swapping the inputs on jpark with a generic two player set. This makes me think that the "flash light" effect issue in the raptor tunnel might not be the video code at all but an input issue since the effect follows the positioning of the input.

Panning higher and lower:
jpark-241210-234240
jpark-241210-234210

Flash light effect wrong for player1 and correct for player2 (both players must be alive to see player 2's effect.)
jpark-241210-234555
jpark-241210-234635


Video starting around 1:50
https://youtu.be/l__YufFcaSQ?si=7NPpF26CkZO8tiwz

Should this behavior be documented?

PR #13010, opened 2024-11-23
scv: Implement video register 0 bit 2: hide sprites 64-127

ReverendGumby:
Two games -- Kung Fu Road and Super Golf -- set this bit to hide half of the sprites.

This replaces the "screen_sprite_start_line" hack, which only worked on Kung Fu Road.


wilbertpol (approved):
That makes way more sense than the current hack. Thank you for submitting this change.

ReverendGumby:
> That makes way more sense than the current hack. Thank you for submitting this change.

You're welcome! I saw your recent fix for Ton Ton Ball, and was inspired to share the fixes I've been discovering.

PR #13011, opened 2024-11-23
misc/compucranes.cpp: Fix hash for 'mastcranea' program ROM

clawgrip:


cuavas:
Can you please slow down and check your changes *before* opening pull requests?  There’s no need to rush them, and these kinds of errors make life more difficult for everyone.

PR #13018, opened 2024-11-26
tetrisp2.cpp: Use memory_bank finder for sound bankswitching, Cleanups:

cam900:
- Split state via configurations
- Use machine_start(), video_start() overrides for video initialization
- Reduce runtime tag lookups, literal tag usages
- Fix typename value consistency
- Fix naming

### DIFF
--- a/src/mame/jaleco/tetrisp2.cpp
+++ b/src/mame/jaleco/tetrisp2.cpp
@@ -82,20 +82,12 @@ u16 rockn_state::rockn_adpcmbank_r()
 void rockn_state::rockn_adpcmbank_w(u16 data)
 {
 	m_rockn_adpcmbank = data;
-	int bank = ((data & 0x001f) >> 2);
-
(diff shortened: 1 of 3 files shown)


src/mame/jaleco/tetrisp2.cpp
```diff
@@ -74,33 +74,27 @@ To Do:
...
-
 	m_rockn_adpcmbank = data;
-	bank = ((data & 0x001f) >> 2);
+	int bank = ((data & 0x001f) >> 2);
 
 	if (bank > 7)
 	{
 		popmessage("!!!!! ADPCM BANK OVER:%01X (%04X) !!!!!", bank, data);
 		bank = 0;
 	}
```
cuavas, line 91: It’s completely impossible for `bank` to be greater than 7 here – `0x1f >> 2 == 0x07`.  The test is useless.

PR #13019, opened 2024-11-26
video/tmep038.cpp: Use device_gfx_interface for gfx decode, Fix notes

cam900:
atlus/cave.cpp: Reduce duplicates

### DIFF
--- a/src/devices/video/tmap038.cpp
+++ b/src/devices/video/tmap038.cpp
@@ -30,11 +30,12 @@
 
     Offset:     Bits:                   Value:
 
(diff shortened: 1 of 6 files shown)


cuavas (commented):
The changes look OK, but is it really a win in this case?
* The 038 tilemap device doesn’t use a fixed tile format, so the actual graphics decoding layout still needs to be supplied by the user.
* It seems to require *more* configuration this way than it did before.
* If multiple chips use the same tiles, they’ll end up with duplicated copies of the decoded tiles.

Overall, this doesn’t look like a clear improvement.

PR #13021, opened 2024-11-27
ap2_dsk.cpp/h: clean up constants

SpecLad:
* convert `#define` to `constexpr`;
* remove unused constants;
* replace magic numbers with constants where appropriate.

### DIFF
--- a/src/lib/formats/ap2_dsk.cpp
+++ b/src/lib/formats/ap2_dsk.cpp
@@ -437,11 +437,11 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
 		uint8_t sectdata[APPLE2_SECTOR_SIZE*16];
 		memset(sectdata, 0, sizeof(sectdata));
 		int nsect = 16;
-		if constexpr (VERBOSE_SAVE) {
+		if(VERBOSE_SAVE) {
 			fprintf(stderr,"DEBUG: a2_16sect_format::save() about to generate bitstream from track %d...", track);
 		}
 		auto buf = generate_bitstream_from_track(track, head, 3915, image);
-		if constexpr (VERBOSE_SAVE) {
+		if(VERBOSE_SAVE) {
 			fprintf(stderr,"done.\n");
 		}
 		int pos = 0;
@@ -475,10 +475,10 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
 				uint8_t tr = gcr4_decode(h[2],h[3]);
 				uint8_t se = gcr4_decode(h[4],h[5]);
 				uint8_t chk = gcr4_decode(h[6],h[7]);
-				if constexpr (VERBOSE_SAVE) {
+				if(VERBOSE_SAVE) {
 					uint32_t post = get_u24be(&h[8]);
 					printf("Address Mark:\tVolume %d, Track %d, Sector %2d, Checksum %02X: %s, Postamble %03X: %s\n",
-						vl, tr, se, chk, (chk ^ vl ^ tr ^ se)==0?"OK":"BAD", post, (post&0xFFFF00)==0xDEAA00?"OK":"BAD");
+							vl, tr, se, chk, (chk ^ vl ^ tr ^ se)==0?"OK":"BAD", post, (post&0xFFFF00)==0xDEAA00?"OK":"BAD");
 				}
 				// sanity check
 				if (tr == track && se < nsect) {
@@ -544,15 +544,14 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
 							// but only write it if the bitfield of the track shows datagood is NOT set.
 							// if it is set we don't want to overwrite a guaranteed good read with a bad one
 							// if past read had a bad checksum or bad postamble...
-							if constexpr (USE_OLD_BEST_SECTOR_PRIORITY) {
+							if(USE_OLD_BEST_SECTOR_PRIORITY) {
 								if ((visualgrid[se][track]&DATAGOOD)==0) {
 									for(int i=0x56; i<0x156; i++) {
 										uint8_t dv = data[i];
 										*dest++ = dv;
 									}
 								}
-							}
-							else {
+							} else {
 								if (((visualgrid[se][track]&DATAGOOD)==0)||((visualgrid[se][track]&DATAPOST)==0)) {
 									// if the current read is good, and postamble is good, write it in, no matter what.
 									// if the current read is good and the current postamble is bad, write it in unless the postamble was good before
@@ -572,10 +571,10 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
 								}
 							}
 							// do some checking
-							if constexpr (VERBOSE_SAVE) {
+							if(VERBOSE_SAVE) {
 								if ((data[0x156] != c) || (dpost&0xFFFF00)!=0xDEAA00)
 									fprintf(stderr,"Data Mark:\tChecksum xpctd %d found %d: %s, Postamble %03X: %s\n",
-										data[0x156], c, (data[0x156]==c)?"OK":"BAD", dpost, (dpost&0xFFFF00)==0xDEAA00?"OK":"BAD");
+											data[0x156], c, (data[0x156]==c)?"OK":"BAD", dpost, (dpost&0xFFFF00)==0xDEAA00?"OK":"BAD");
 							}
 							if (data[0x156] == c) visualgrid[se][track] |= DATAGOOD;
 							if ((dpost&0xFFFF00)==0xDEAA00) visualgrid[se][track] |= DATAPOST;
@@ -603,7 +602,7 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
 		//printf("\n");
 	}
 	// display a little table of which sectors decoded ok
-	if constexpr (VERBOSE_SAVE) {
+	if(VERBOSE_SAVE) {
 		int total_good = 0;
 		for (int j = 0; j < APPLE2_TRACK_COUNT; j++) {
 			printf("T%2d: ",j);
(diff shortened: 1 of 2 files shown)


src/lib/formats/ap2_dsk.cpp
```diff
@@ -542,37 +544,39 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
...
+									}
 								}
 							}
-#endif
 							// do some checking
-							#ifdef VERBOSE_SAVE
-							if ((data[0x156] != c) || (dpost&0xFFFF00)!=0xDEAA00)
-								fprintf(stderr,"Data Mark:\tChecksum xpctd %d found %d: %s, Postamble %03X: %s\n", data[0x156], c, (data[0x156]==c)?"OK":"BAD", dpost, (dpost&0xFFFF00)==0xDEAA00?"OK":"BAD");
-							#endif
+							if constexpr (VERBOSE_SAVE) {
```
MooglyGuy, line 575: I don't think it's necessary to prepend these with `constexpr`, as MAME has the appropriate compiler flags set to handle effective `if (0)` and `if (1)` situations without warnings.

SpecLad: Sure, but is there any harm in it? I think it's a useful reminder that it's supposed to be a compile-time selection.

cuavas: That’s not what `if constexpr` is for.  The purpose of `if constexpr` is for when one branch is malformed when the expression is false.  For example see use in src.emu/device.h:
```c++
    template <class DeviceClass>
    DeviceClass *subdevice(std::string_view tag) const
    {
        device_t *const found = subdevice(tag);
        if constexpr (std::is_base_of_v<device_t, DeviceClass>)
        {
            return downcast<DeviceClass *>(found);
        }
        else
        {
            auto const result = dynamic_cast<DeviceClass *>(found);
#if defined(MAME_DEBUG)
            if (found && !result)
                report_bad_device_cast(found, typeid(device_t), typeid(DeviceClass));
#endif
            return result;
        }
    }
```

The code can use the `downcast` helper if `DeviceClass` is derived from `device_t`, but this is invalid otherwise (e.g. if it is derived from `device_interface` but not `device_t`).

Or consider this code in src/lib/util/delegate.h where it needs to make a decision on whether it can cast the call operator to a usable function type or it needs to wrap it in a lambda to convert the return value:
```c++
    template <typename T>
    static functoid_setter make_functoid_setter()
    {
        if constexpr (matching_non_const_call(&unwrapped_functoid_t<T>::operator()))
        {
            return
                    [] (delegate &obj)
                    {
                        obj.basetype::operator=(
                                basetype(
                                    static_cast<ReturnType (unwrapped_functoid_t<T>::*)(Params...)>(&unwrapped_functoid_t<T>::operator()),
                                    obj.unwrap_functoid<T>()));
                    };
        }
        else if constexpr (matching_const_call(&unwrapped_functoid_t<T>::operator()))
        {
            return
                    [] (delegate &obj)
                    {
                        obj.basetype::operator=(
                                basetype(
                                    static_cast<ReturnType (unwrapped_functoid_t<T>::*)(Params...) const>(&unwrapped_functoid_t<T>::operator()),
                                    obj.unwrap_functoid<T>()));
                    };
        }
        else
        {
            return
                    [] (delegate &obj)
                    {
                        obj.basetype::operator=(
                                basetype(
                                    [] (unwrapped_functoid_t<T> &f, Params... args) { return ReturnType(f(std::forward<Params>(args)...)); },
                                    obj.unwrap_functoid<T>()));
                    };
        }
    }
```

Depending on the signature of the call operator, one or two of the branches will be malformed.

src/lib/formats/ap2_dsk.h
```diff
@@ -21,12 +21,9 @@
...
 
-#define APPLE2_NIBBLE_SIZE 416
-#define APPLE2_SMALL_NIBBLE_SIZE 374
-#define APPLE2_STD_TRACK_COUNT 35
-#define APPLE2_TRACK_COUNT 40
-#define APPLE2_SECTOR_COUNT 16
-#define APPLE2_SECTOR_SIZE 256
+constexpr int APPLE2_STD_TRACK_COUNT = 35;
+constexpr int APPLE2_TRACK_COUNT = 40;
+constexpr int APPLE2_SECTOR_SIZE = 256;
```
cuavas, line 26: Can the scope of these be reduced?  Do they need to be in the global namespace?  Are the nibble sizes useful for documentation even if they’re currently unused?

src/lib/formats/ap2_dsk.cpp
```diff
@@ -542,37 +544,39 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
...
 										*dest++ = dv;
 									}
 								}
 							}
-#else
-							if ((visualgrid[se][track]&DATAGOOD)==0) {
-								for(int i=0x56; i<0x156; i++) {
-									uint8_t dv = data[i];
-									*dest++ = dv;
+							else {
```
cuavas, line 555: This file is already inconsistent with K&R vs Allman formatting, but I don’t think any of it uses this style:
```c++
    }
    else {
```

It’s all either Allman:
```c++
    }
    else
    {
```
or K&R:
```c++
    } else {
```

You should use the predominant style of the file.

src/lib/formats/ap2_dsk.cpp
```diff
@@ -471,19 +475,17 @@ bool a2_16sect_format::save(util::random_read_write &io, const std::vector<uint3
...
 				uint8_t se = gcr4_decode(h[4],h[5]);
 				uint8_t chk = gcr4_decode(h[6],h[7]);
-				#ifdef VERBOSE_SAVE
-				uint32_t post = get_u24be(&h[8]);
-				printf("Address Mark:\tVolume %d, Track %d, Sector %2d, Checksum %02X: %s, Postamble %03X: %s\n", vl, tr, se, chk, (chk ^ vl ^ tr ^ se)==0?"OK":"BAD", post, (post&0xFFFF00)==0xDEAA00?"OK":"BAD");
-				#endif
+				if constexpr (VERBOSE_SAVE) {
+					uint32_t post = get_u24be(&h[8]);
+					printf("Address Mark:\tVolume %d, Track %d, Sector %2d, Checksum %02X: %s, Postamble %03X: %s\n",
+						vl, tr, se, chk, (chk ^ vl ^ tr ^ se)==0?"OK":"BAD", post, (post&0xFFFF00)==0xDEAA00?"OK":"BAD");
```
cuavas, line 481: We use an additional indentation level for statement continuations vs scope levels, e.g.:
```c++
    if (x)
    {
        f( // indented one level because scope depth increased
                a, // two levels for continuation
                b);
        g();
    }
```

cuavas (commented):
I tend to agree with @MooglyGuy that `if constexpr` should be used for situations where one branch may be malformed depending on the condition being tested, not just for situations where the expression is a compile-time constant.

src/lib/formats/ap2_dsk.cpp
cuavas: You can find plenty of examples of testing compile-time constant expressions with regular `if`, e.g. in src/devices/cpu/e132xs/e132xsop.hxx:

```c++
template <hyperstone_device::reg_bank DST_GLOBAL, hyperstone_device::reg_bank SRC_GLOBAL>
void hyperstone_device::hyperstone_and()
{
        check_delay_PC();

        const uint32_t dst_code = DST_GLOBAL ? DST_CODE : ((DST_CODE + GET_FP) & 0x3f);
        const uint32_t sreg = SRC_GLOBAL ? m_core->global_regs[SRC_CODE] : m_core->local_regs[(SRC_CODE + GET_FP) & 0x3f];
        const uint32_t dreg = (DST_GLOBAL ? m_core->global_regs : m_core->local_regs)[dst_code] & sreg;

        if (dreg == 0)
                SR |= Z_MASK;
        else
                SR &= ~Z_MASK;

        if (DST_GLOBAL)
                set_global_register(dst_code, dreg);
        else
                m_core->local_regs[dst_code] = dreg;

        m_core->icount -= m_core->clock_cycles_1;
}
```

SpecLad: Okay, I removed it.

src/lib/formats/ap2_dsk.cpp
SpecLad: FWIW, this style is used in this function too (see the block starting with `if(v == 0xff) {`). But sure, I changed it.

src/lib/formats/ap2_dsk.cpp
SpecLad: Okay, updated.

src/lib/formats/ap2_dsk.h
SpecLad: > Can the scope of these be reduced? Do they need to be in the global namespace?

I thought about moving them to the `.cpp` file, but since I used them in the definition of `a2_nib_format`, I can't do that anymore. (Technically, I could still move `APPLE2_SECTOR_SIZE`, but it seems better for readability to keep these constants together.)

> Are the nibble sizes useful for documentation even if they’re currently unused?

I don't even know what these values are supposed to be, to be honest. I thought they were the sector data sizes after GCR encoding (in the 13-sector and 16-sector formats), but on second look, that doesn't seem to actually be the case.

ajrhacker:
> I tend to agree with @MooglyGuy that `if constexpr` should be used for situations where one branch may be malformed depending on the condition being tested, not just for situations where the expression is a compile-time constant.

Unfortunately the C++ language doesn't allow `if constexpr` to have malformed branches unless the condition being tested involves some sort of template-dependent stuff.

peterferrie:
>>+constexpr int APPLE2_STD_TRACK_COUNT = 35;
>>+constexpr int APPLE2_TRACK_COUNT = 40;
>>+constexpr int APPLE2_SECTOR_SIZE = 256;
...
>I don't even know what these values are supposed to be, to be honest. I
thought they were the sector data sizes after GCR encoding (in the
13-sector and 16-sector formats), but on second look, that doesn't seem to
actually be the case.

They are nothing to do with the nib format.  There are 35 tracks on a
regular disk, 40 on an extended disk, and decoded sectors are 256 bytes.
16-sector nib encoded sectors are 342 bytes long, without headers.
13-sector nib sectors are even bigger.

PR #13022, opened 2024-11-28
split monster 'toaplan2.cpp' file into smaller per-board parts + slight modernization

mamehaze:
this chops down the incredibly entangled toaplan2.cpp into individual drivers for each board making better use of required_device rather than optional_device.  toaplan2.cpp/toaplan2.h are killed off in the process.

modernizations have also been done, removing legacy VIDEO_START / MACHINE_START / MACHINE_RESET type macros

this isn't perfect, and there's definitely room for some further improvements, such as maybe moving the bootleg sound systems to a device, or the extra tilemaps, and some notes could probably do with re-homing, but overall I feel this already provides greater clarity compared to the unwieldy toaplan2.cpp driver.

otherwise the code is pretty much untouched, so I kindly ask that this PR isn't judged on existing code that has just ended up in a new location, as such clean-ups can be scheduled for a later date.


MooglyGuy:
Looks great to me, just some boilerplate feedback. I only left the feedback on a few files for reference purposes, as a few things stand out as nice-to-haves, but which I do not think should hold up the merging of this PR, as I'd be happy to sweep for and apply these bits of feedback as a separate PR myself:
- Drivers which are .cpp-only tend to be put in an anonymous namespace, see src/mame/alpha/equites.cpp for an example.
- There's at least one driver (tekipaki.cpp) which has an optional shared-RAM finder which goes unused. It could be worth double-checking for other instances of unneeded shared RAM.
- Now that the drivers are self-contained, in many cases the vestigial function prefixes and suffixes can be removed, e.g. `screen_update_toaplan2` can be changed back to the more-typical `screen_update`. In some cases the function suffixes are still useful (e.g. pipibibis and its bootlegs).
- There are some cases where formerly-optional finders can now be switched to required finders, usually audio devices and audio CPUs, and occasionally the audio shared-RAM finder.

mamehaze:
> Looks great to me, just some boilerplate feedback. I only left the feedback on a few files for reference purposes, as a few things stand out as nice-to-haves, but which I do not think should hold up the merging of this PR, as I'd be happy to sweep for and apply these bits of feedback as a separate PR myself:
> 
>     * Drivers which are .cpp-only tend to be put in an anonymous namespace, see src/mame/alpha/equites.cpp for an example.
> 
>     * There's at least one driver (tekipaki.cpp) which has an optional shared-RAM finder which goes unused. It could be worth double-checking for other instances of unneeded shared RAM.
> 
>     * Now that the drivers are self-contained, in many cases the vestigial function prefixes and suffixes can be removed, e.g. `screen_update_toaplan2` can be changed back to the more-typical `screen_update`. In some cases the function suffixes are still useful (e.g. pipibibis and its bootlegs).
> 
>     * There are some cases where formerly-optional finders can now be switched to required finders, usually audio devices and audio CPUs, and occasionally the audio shared-RAM finder.

I think the majority of this has been addressed, most were simple oversights (I thought I'd already got rid of everything that wasn't used, but that tekipaki one slipped through)

the optional -> required cases were times I forgot to do an extra pass I was planning on doing.

anonymous namespaces have been added, _toaplan2 and _truxton2 extensions from the old code have been dropped, notes have been moved to match the drivers now containing the games.

mamehaze:
In terms of future PRs, I think what I'll do with the tilemaps is create a device for the setups with rowscroll / rowselect, and use that for the original games, and just have the bootleg classes create their own tilemaps without any of that in cases where they don't support them.

I'll probably also make the 'banked' OKI some of the Raizing games use a device too.

For now I'd like to consider this a milestone however, as further work is a little more invasive and has a higher risk of accidentally breaking something, while this set of changes should generally be safe.

PR #13024, opened 2024-11-28
heathzenith/h89.cpp: Convert SigmaSoft parallel port into an h89 left card.

mgarlanger:
tlb.cpp/.h:

- move into the src/devce/bus directory so that other bus devices can reference the classes.

cards.cpp:

- add new sigmasoft_parallel_port option to left cards

h89bus.cpp:

- move left card processing for read/write outside of PROM bit check
- mask off unavailable address bits for the left card devices.

sigmasoft_parallel_port.cpp:

- convert into an h89 left card and move source to devices/bus/heathzenith/h89

h89.cpp:

- Update h89_sigmasoft system to use h89 left card for parallel port


src/devices/bus/heathzenith/h89/sigmasoft_parallel_port.cpp
```diff
@@ -0,0 +1,245 @@
...
+{
+	// connect callbacks to TLB
+	ctrl_r_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_ctrl_r));
+	video_mem_r_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_video_mem_r));
+	video_mem_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_video_mem_w));
+	io_lo_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_io_lo_addr_w));
+	io_hi_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_io_hi_addr_w));
+	window_lo_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_window_lo_addr_w));
+	window_hi_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_window_hi_addr_w));
+	ctrl_cb().set(":tlbc", FUNC(heath_tlb_connector::sigma_ctrl_w));
```
ajrhacker, line 242: It would be better not to use absolute tags here.

mgarlanger: What is the alternative? Before both the card and TLBC was created in h89.cpp, and could be directly linked. But here, when the parallel card is created, it does not have direct access to the tlbc.

rb6502: There are relative tags.

cuavas: It’s still a device making assumptions about the surrounding system, which is fragile, and we’re trying to eliminate.  If these signals can pass through the slot connector, they need to go through the slot device somehow.

It seems these are being added faster than existing ones are fixed.  At this rate I think I’m going to need to come up with a way to trigger validation warnings for them.

mgarlanger: I received feedback from some other mame devs that resulted in my followup PR based on their suggestions - https://github.com/mamedev/mame/pull/13040 which you have also commented on. I guess we can have the conversation on that PR.

PR #13030, opened 2024-11-29
Converted Williams SC1/SC2 blitter to a device.

MooglyGuy:
First step towards breaking up the annoyingly spaghetti-y williams.cpp drivers.
Moving some of the control-bit checks outside of the inner blit loop seems to have given a slight improvement to performance, which is nice.


ajrhacker:
The new source files need to be added to tiny.lua. The midway/wmg.cpp driver also needs to be updated.

MooglyGuy:
Thanks for the heads-up. Addressing.

MooglyGuy:
Fixed. I like the unnecessary function prefixes in wmg.cpp, very retro-MAME-standards.

ajrhacker:
Not properly fixed with regard to tiny.lua.

MooglyGuy:
Ugh, you're right. I found the first result for "williams" and wasn't paying attention. Having manufacturer-named files in shared is just awful.

ajrhacker:
There's just an unused variable issue left. I'll take care of that.

PR #13033, opened 2024-11-30
igs/igs011.cpp: Split vbowl configuration via trackball inputs, Cleanups:

cam900:
- Fix drgnwrldv40k metadata due to title screen/game name
- Simplify key matrix input read functions
- Reduce duplicates
- Use generic gfx decode layout for debugging
- Suppress side effects in debugger reads
- Fix naming, Fix spacings
- Fix/Add docs

### DIFF
--- a/src/mame/igs/igs011.cpp
+++ b/src/mame/igs/igs011.cpp
@@ -1673,7 +1673,7 @@ void igs011_state::lhb_inputs_w(offs_t offset, u16 data, u16 mem_mask)
 
 	if (ACCESSING_BITS_0_7)
 	{
-		machine().bookkeeping().coin_counter_w(0, BIT(data, 5) );
+		machine().bookkeeping().coin_counter_w(0, BIT(data, 5));
 		//  coin out        BIT(data, 6)
 		m_hopper_bit = BIT(data, 7);
 	}
--- a/src/mame/igs/igs011.cpp
+++ b/src/mame/igs/igs011.cpp
@@ -94,12 +94,10 @@ public:
 		, m_oki(*this, "oki")
 		, m_screen(*this, "screen")
 		, m_palette(*this, "palette")
-		, m_ics(*this, "ics")
 		, m_priority_ram(*this, "priority_ram")
-		, m_vbowl_trackball(*this, "vbowl_trackball")
 		, m_maincpu_region(*this, "maincpu")
 		, m_gfx(*this, "blitter")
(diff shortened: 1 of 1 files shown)


src/mame/igs/igs011.cpp
```diff
@@ -433,10 +456,10 @@ u32 igs011_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, con
 
 u16 igs011_state::igs011_layers_r(offs_t offset)
 {
-	int layer0 = ((offset & (0x80000/2)) ? 4 : 0) + ((offset & 1) ? 0 : 2);
+	const int layer0 = ((offset & (0x80000 / 2)) ? 4 : 0) + (BIT(offset, 0) ? 0 : 2);
```
cuavas, line 459: You could express this as `bitswap<2>(offset, 18, 0) << 1` couldn’t you?

src/mame/igs/igs011.cpp
```diff
@@ -446,17 +469,15 @@ u16 igs011_state::igs011_layers_r(offs_t offset)
 
 void igs011_state::igs011_layers_w(offs_t offset, u16 data, u16 mem_mask)
 {
-	u16 word;
-
-	int layer0 = ((offset & (0x80000/2)) ? 4 : 0) + ((offset & 1) ? 0 : 2);
+	const int layer0 = ((offset & (0x80000 / 2)) ? 4 : 0) + (BIT(offset, 0) ? 0 : 2);
```
cuavas, line 472: Same here, this looks like a bit of a convoluted way to extract two bits.

src/mame/igs/igs011.cpp
```diff
@@ -651,24 +664,36 @@ void igs011_state::machine_start()
...
-u16 igs011_state::igs_dips_r()
+u16 igs011_state::dips_r()
 {
-	u16 ret=0;
+	u16 ret = 0;
 
 	for (int i = 0; i < Num; i++)
-		if ((~m_igs_dips_sel) & (1 << i))
+		if (BIT(~m_dips_sel, i))
 			ret = m_io_dsw[i]->read();
```
cuavas, line 697: This is another switch matrix – you should AND together the active rows:

```c++
	u16 ret = 0x00ff;

	for (int i = 0; i < Num; i++)
		if (BIT(~m_dips_sel, i))
			ret &= m_io_dsw[i]->read();
```

PR #13036, opened 2024-12-01
update notes seoul88.cpp

flama12333:
when the original is dumped. the driver need to be rename to lucky 75. also this is bootleg of lucky 75


Osso13:
The fact that they have some similar GFX doesn't necessarily make them related. If/when the Lucky 75 program gets dumped conclusions can be drawn,

cuavas:
This is unsubstantiated speculation.

PR #13038, opened 2024-12-01
trs/gime.cpp: fix some legacy video modes

tlindner:
The legacy GIME mode lines per row calculation is very different from the MC6847. Also fix a small border color issue when compared with real hardware.

### DIFF
--- a/src/mame/trs/gime.cpp
+++ b/src/mame/trs/gime.cpp
@@ -1286,7 +1286,7 @@ void gime_device::update_border(uint16_t physical_scanline)
 			// graphics, green or white
 			border = (~m_ff22_value & MODE_CSS) ? 0x12 : 0x3F;
 		}
-		else if (m_ff22_value & MODE_GM2)
+		else if ((m_ff22_value & MODE_GM2) && !(m_ff22_value & MODE_GM1))
 		{
 			// text, green or orange
 			border = (~m_ff22_value & MODE_CSS) ? 0x12 : 0x26;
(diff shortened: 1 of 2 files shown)


src/mame/trs/gime.cpp
```diff
@@ -1329,42 +1329,26 @@ inline uint16_t gime_device::get_lines_per_row()
...
-					11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 4, 3, 2, 1, 12
-				};
-
-				int i = m_gime_registers[0x0C] & 0x0F;
-				//lines_per_row = 12;
-				lines_per_row = ff9c_lines_per_row[i];
-				break;
-			}
+				3, 3, 3, 2, 2, 1, 1, 1
+			};
```
cuavas, line 1337: This is non-constant static data in a function.  @ajrhacker can you please review code before merging it?

src/mame/trs/gime.cpp
```diff
@@ -1329,42 +1329,26 @@ inline uint16_t gime_device::get_lines_per_row()
...
+			int i = m_sam_state & (SAM_STATE_V0|SAM_STATE_V1|SAM_STATE_V2);
+			lines_per_row = gime_legacy_lines_per_row_graphic[i];
+		}
+		else
+		{
+			// http://cocogamedev.mxf.yuku.com/topic/4299238#.VyC6ozArI-U
+			static int ff9c_lines_per_row_alpha[16] =
+			{
+				11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 4, 3, 2, 1, 12
+			};
```
cuavas, line 1348: So is this.

src/mame/trs/gime.cpp
```diff
@@ -1329,42 +1329,26 @@ inline uint16_t gime_device::get_lines_per_row()
...
 
-			default:
-				fatalerror("Should not get here\n");
+			int i = m_gime_registers[0x0C] & 0x0F;
+			lines_per_row = ff9c_lines_per_row_alpha[i];
```
cuavas, line 1351: MAME standard is to use lowercase hex digits.

cuavas (commented):
@ajrhacker please fix the obvious issues with this.

PR #13039, opened 2024-12-02
dt7.cpp - improve text layer handling (adds rowselect, fixes course select screen) [David Haywood]

mamehaze:
(resynced to latest code)

### DIFF
--- a/src/mame/toaplan/dt7.cpp
+++ b/src/mame/toaplan/dt7.cpp
@@ -55,6 +55,10 @@ public:
 		, m_shared_ram(*this, "shared_ram")
 		, m_tx_videoram(*this, "tx_videoram")
 		, m_lineram(*this, "lineram")
+		, m_eepromport(*this, "EEPROM")
+		, m_sysport(*this, "SYS")
+		, m_p1port(*this, "IN1")
+		, m_p2port(*this, "IN2")
 	{ }
 
 public:
@@ -90,8 +94,13 @@ private:
 	uint8_t read_port_2();
 	void write_port_2(uint8_t data);
 
+	uint8_t eeprom_r();
+	void eeprom_w(uint8_t data);
+
+
 	u8 dt7_shared_ram_hack_r(offs_t offset);
 	void shared_ram_w(offs_t offset, u8 data);
+	void shared_ram_audio_w(offs_t offset, u8 data);
 
 	void screen_vblank(int state);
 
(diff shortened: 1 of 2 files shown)


src/mame/toaplan/dt7.cpp
```diff
@@ -470,34 +468,48 @@ void dt7_state::video_start()
...
+		//u16 scroll2 = m_lineram[((y * 8) + (table * 2) + 1) & 0x7ff]; // xscroll
+
+		scroll1 &= 0x7fff; // 0x8000 might be enable? or per-line flip if it's more similar to other Toaplan drivers
+		scroll1 -= 0x0900; // are all these scroll bits?
+
+		m_tx_tilemap[table/2]->set_scrolly(0, scroll1 - y);
+		m_tx_tilemap[table/2]->draw(screen, bitmap, clip, 0);
 	}
-}
+ }
```
cuavas, line 495: Stray space.

src/mame/toaplan/toaplan_v25_tables.h
```diff
@@ -109,15 +109,15 @@ class toaplan_v25_tables
 	// some kind of branch, not sure which
 	// it's used after compares in blocks, sometimes with a 'be' then a 'br' straight after, so it must be a condition that could also fail a be and fall to the br
 	//#define G_B0  0x74
-	#define G_B0  0x79
+	#define G_B0  0x79 // very low confidence
```
ajrhacker, line 112: The context suggests that this might instead be 0x77 (it follows comparing a bit position against 7).

src/mame/toaplan/dt7.cpp
```diff
@@ -141,10 +154,41 @@ uint8_t dt7_state::read_port_t()
 uint8_t dt7_state::read_port_2()
 {
 	logerror("%s: read port 2\n", machine().describe_context());
-
 	return 0xff;
```
ajrhacker, line 157: At least a few of the bits written to this port need to be read back here, since the code is using bit manipulation instructions and the V25 device doesn't enforce port directions.

mamehaze: ahh, indeed, I think I had that in the code at one point, but this has been sat around for so long I must have forgot about it, anyway, it's added now.

happppp:
Is it intentional that the region is Korea now?

mamehaze:
> Is it intentional that the region is Korea now?

it's meant to come from the eeprom, which still isn't hooked up properly, so kinda irrelevant at this point

Issue #13041, opened 2024-12-02
Emscripten build fails due to variable name clashing with Emscripten macro


algestam:
### MAME version

0.272

### System information

Linux host with Emscripten 3.1.35

### Incorrect behaviour

When building MAME 0.272 on Linux using Emscripten 3.1.15, the build fails with the following error:

```
Compiling src/devices/cpu/z80/tmpz84c011.cpp...
In file included from ../../../../../src/devices/cpu/z80/t6a84.cpp:10:
../../../../../src/devices/cpu/z80/t6a84.h:44:35: error: expected member name or ';' after declaration specifiers
        static inline constexpr uint32_t PAGE_SIZE = 0x10000;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
In file included from ../../../../../src/devices/cpu/z80/t6a84.cpp:1:
In file included from /src/build/projects/sdl/mamearcade/gmake-asmjs/../../../../../src/emu/emu.h:24:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/list:202:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/memory:860:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__memory/shared_ptr.h:39:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/atomic:523:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/__chrono/duration.h:15:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/ratio:82:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/climits:42:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/c++/v1/limits.h:57:
In file included from /emsdk/upstream/lib/clang/17/include/limits.h:21:
In file included from /emsdk/upstream/emscripten/cache/sysroot/include/limits.h:40:
/emsdk/upstream/emscripten/cache/sysroot/include/bits/limits.h:3:19: note: expanded from macro 'PAGE_SIZE'
#define PAGE_SIZE 65536
                  ^
```

The issue seems to be that Emscripten defines a macro, `PAGE_SIZE`, which is expanded when compiling and replaces the variable name `PAGE_SIZE` in `src/devices/cpu/z80/t6a84h` into `65536` and causes the compile error.

[...]

Issue #12479, opened 2024-06-14
[Android] Fail to compile msxdos2.cpp:  expected member name or ';' after declaration specifiers 


FlykeSpice:
### MAME version

85624c486db19493fbbb87c091ca1a77559651d5 

### System information

Cross-compiling to Android

### Emulated system/software

msxdos2

### Incorrect behaviour

I'm trying to cross-compile mame (latest git head commit) to Android so I can contribute to Mame4Droid, because Mame4Droid author doesn't provides the build scripts or instructions necessary to build the core I had to make them myself.

My buildscripts do nothing than setup the build system to use mame4droid OSD on android target, I managed to sucessfully compile(and run) SUBTARGET=tiny and now I'm trying to compile the mame one.

However after long hours it failed on this exact spot:

../../../../../src/devices/bus/msx/cart/msxdos2.cpp:42:35: error: expected member name or ';' after declaration specifiers
        static inline constexpr unsigned PAGE_SIZE = 0x4000;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
In file included from ../../../../../src/devices/bus/msx/cart/msxdos2.cpp:1:
In file included from /home/flyke/mame/build/projects/mame4droid/mame/gmake-android-arm64/../../../../../src/emu/emu.h:24:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/list:198:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/__memory/allocator.h:23:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/stdexcept:48:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/iosfwd:100:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/__mbstate_t.h:29:
In file included from /home/flyke/android-ndk-r26d/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/wchar.h:124:
[...]

PR #13042, opened 2024-12-02
cpu/z80/t6a84: Fix emscripten compile issue

algestam:
This PR changes fixes a emscripten compile issue described in #13041.

### DIFF
--- a/src/devices/cpu/z80/t6a84.cpp
+++ b/src/devices/cpu/z80/t6a84.cpp
@@ -139,7 +139,7 @@ bool t6a84_device::memory_translate(int spacenum, int intention, offs_t &address
 
 uint32_t t6a84_device::code_address(uint16_t address)
 {
-	const uint32_t page_address = PAGE_SIZE * m_code_page + address;
+	const uint32_t page_address = DEVICE_PAGE_SIZE * m_code_page + address;
 	LOGMASKED(LOG_MEM, "CODE @ %06x => %06x\n", address, page_address);
 
 	return page_address;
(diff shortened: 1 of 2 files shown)


cuavas:
Can you push back on Emscripten and tell them to:
* Stop polluting standard headers with architecture headers, and
* Use reserved names for their macros that start with either two underscores or an underscore followed by an uppercase letter?

If they used reserved names like they’re supposed to, we wouldn’t have to worry about this crap happening.

algestam:
Agree that the macro should've had a better name in the first place. However, I'm not sure if Emscripten is to blame entirely here as they are bundling musl and have added an emscripten specific arch variant with it. All arch variants in musl seems to define `PAGESIZE` (changed from `PAGE_SIZE` in 2018: https://github.com/kraj/musl/commit/c9c2cd3e6955cb1d57b8be01d4b072bf44058762). The emscripten arch variant was most likely added before the name was changed.

We can of course ask Emscripten to change the variable name in the same manner as the upstream musl change but as MAME isn't building on the latest Emscripten builds that still won't resolve the issue with building MAME with emscripten.

holub:
Cuavas' concern was not about who added this variable first.
Build environment shouldn't leak into client's environment. That can create all sort of dangerous results in unpredictable places.

I think we can merge the PR. I'll try to fined the way to deliver "the message" to Emscropten.community.

happppp:
I agree the generic "PAGE_SIZE" shouldn't have been named that way.
But I think it was unneeded here, it's removed: https://github.com/mamedev/mame/commit/991bdf14be74ac69f9012534613433e317f95917

holub:
> Can you push back on Emscripten and tell them to:
> 
>     * Stop polluting standard headers with architecture headers, and
> 
>     * Use reserved names for their macros that start with either two underscores or an underscore followed by an uppercase letter?
> 
> 
> If they used reserved names like they’re supposed to, we wouldn’t have to worry about this crap happening.

@cuavas can you give us a hand? https://github.com/emscripten-core/emscripten/issues/23082

FlykeSpice:
This is the very same issue I came across when compiling mame to android. Android ndk leaks PAGE_SIZE into standard headers aswell: https://github.com/mamedev/mame/issues/12479

PR #13044, opened 2024-12-03
extra info on zstd, zlib, lzma

checktext00:
add extra info about zstd, about decompression speeds for zstd,zlib,lzma, and about using zstd and zlib with the `chdman create...` commands (some info from https://github.com/mamedev/mame/issues/7402#issuecomment-1849132657)

### DIFF
--- a/docs/source/tools/chdman.rst
+++ b/docs/source/tools/chdman.rst
@@ -418,16 +418,43 @@ Compression algorithms
 The following compression algorithms are supported:
 
 zlib – zlib deflate
-   Compresses data using the zlib deflate algorithm.
+   Compresses data using the zlib deflate algorithm.  Decompression performance
+   is around 3½ times slower than Zstandard and 4 times faster than the
+   Lempel-Ziv-Markov chain algorithm (LZMA).
+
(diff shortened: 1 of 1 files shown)


crashGG:
The decompression speed of zstd is fast, but not as fast as expected. A simple test:
Use  single algorithm to compress a DVD iso with almost no duplicate data
chdman createdvd -c zstd -i sample.iso -o zstd.iso
chdman craatedvd -c zlib -c zstd -i sample.iso -o zlib.chd
chdman craatedvd -c flac -c zstd -i sample.iso -o flac.chd
chdman craatedvd -c haff -c zstd -i sample.iso -o haff.chd
chdman craatedvd -c lzma -c zstd -i sample.iso -o lzma.chd


Then use **chdman verify** to decompress 5 chd files with script to calculate the time difference (accurate to milliseconds) and divide it by the file size to get the decompression speed of 5 algorithms. According to my hardware（amd 7840h）, flac 110M/s ,huff 87M/s, zlib 97M/s, zstd 115.7M/s lzma 24.6M/s

PR #13047, opened 2024-12-04
Segas32 - Resolution change when zooming

mahoneyt944:
It seems that zooming requires more resolution but uses 9 bit otherwise.

This should fix the missing background for harddunk in attract mode, while also working with radr map selection, alien3 map between stages and so on. Has not been verified to real hardware but it seems to test correctly with all games. Let me know if there's a specific transition / case you would like tested.

### DIFF
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -17,9 +17,8 @@
       The theory is that opaque pens should go above background layer and
       behind everything else like System 24.
 
-    - radr uses $1A0 as the X center for zooming; however, this
-      contradicts the theory that bit 9 is a sign bit. For now, the code
-      assumes that the X center has 10 bits of resolution.
+    - Verify that X/Y center has 10 bits of resolution when zooming and
+      9 when not.
 
     - In svf (the field) and radr (on the field), they use tilemap-specific
       flip in conjunction with rowscroll AND rowselect. According to Charles,
@@ -740,8 +739,8 @@ void segas32_state::update_tilemap_zoom(screen_device &screen, segas32_state::la
 	srcy += (m_videoram[0x1ff14/2 + 4 * bgnum] & 0xfe00) << 4;
 
 	/* then account for the destination center coordinates */
-	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum] & 0x1ff, 10) * srcxstep;
-	srcy -= util::sext(m_videoram[0x1ff32/2 + 2 * bgnum], 9) * srcystep;
+	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum] & 0x1ff, (dstxstep != 0x200)?10:9) * srcxstep;
+	srcy -= util::sext(m_videoram[0x1ff32/2 + 2 * bgnum], (dstystep != 0x200)?10:9) * srcystep;
 
 	/* finally, account for destination top,left coordinates */
 	srcx_start += cliprect.min_x * srcxstep;
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -742,7 +742,7 @@ void segas32_state::update_tilemap_zoom(screen_device &screen, segas32_state::la
 	/*
 	   Then account for the destination center coordinates - We currently expand the resolution
 	   from 9 bit to 10 bit while zooming which correctly centers the bg in harddunk during attract
-	   mode at the low resolution and the course selection screen in radr at the higher resolution.
+	   mode at the lower resolution and the course selection bg in radr at the higher resolution.
 	   This behavior has not been verified yet on real hardware and might be a hack.
 	*/
 	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum], (dstxstep != 0x200) ? 10 : 9) * srcxstep;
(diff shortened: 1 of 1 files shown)


mahoneyt944:
Update to #13003

cuavas:
That seriously looks like a hack that works by coincidence.  It’s the kind of thing that I’d want to see some pretty convincing evidence for.

mahoneyt944:
Well I don't know how to prove it as I don't have the hardware or means to investigate the hardware even if I had it, but if there's a tangible way to test this I'll be happy to try. It seems to me that that the code was written with the presumption that it should be using 9 bit, and was only altered to use 10 as some games such as the rad Rally's map selection screen did not work properly at 9. Unfortunately this alternation seems to cause harddunks background to not line up, trading one game for another. My theory is that the resolution should increase while zooming, not locked at 10 as the current code is, all games seem to behave properly with this. I'm happy to test or compare any games you feel do not.

angelosa:
At best this could have the Saturn VDP2 Reduction Enable bit: if I gather right that is supposed to control minimum possible value when incrementing on the coordinate _X axis_. And, even then, it doesn't mention anything about Y axis so if by removing the srcy modification harddunk don't work then you are back to square one here.

mahoneyt944:
Completed full play throughs of every game in the driver and confirmed correct behavior with all games.

Some games that suggest increased resolution only when zooming:
- radr (map selection screen) uses 10 bit - zooming
- alien3 (map between levels, start of 3rd level) uses 10 bit - zooming
- harddunk (attract mode, player selection) uses 9 bit - no zoom

A lot of games use zooming very little, so changing all these instances to 9 bit, should yield an incorrect display somewhere but that doesn't seem to be the case.

This PR is minimally invasive so I can't see much harm in it and has been noted in the open questions. Increasing the resolution was already an accepted commit from 9 to 10 prior to this request, this just tightens the belt on when that is necessary.

Is there any further testing you would like done for this to be accepted?

rb6502:
I think given the very limited amount of software for this machine this is acceptable given it's very limited in scope and it definitely helps the results.  That said I'd still like there to be a comment that it's possibly a hack pending someone testing it on hardware.

mahoneyt944:
> I think given the very limited amount of software for this machine this is acceptable given it's very limited in scope and it definitely helps the results. That said I'd still like there to be a comment that it's possibly a hack pending someone testing it on hardware.

Thank you. I did add a note to the open questions, however maybe it needs more detail in this instance?

mahoneyt944:
Also just noticed what seems to be a typo on the page select register notes. Have not checked to see if that translated anywhere else in the code though or if those are accurate. The last page 1 register was missing.

src/mame/sega/segas32_v.cpp
```diff
@@ -744,8 +740,8 @@ void segas32_state::update_tilemap_zoom(screen_device &screen, segas32_state::la
 	srcy += (m_videoram[0x1ff14/2 + 4 * bgnum] & 0xfe00) << 4;
 
 	/* then account for the destination center coordinates */
-	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum], 10) * srcxstep;
-	srcy -= util::sext(m_videoram[0x1ff32/2 + 2 * bgnum], 9) * srcystep;
+	srcx_start -= util::sext(m_videoram[0x1ff30/2 + 2 * bgnum], (dstxstep != 0x200)?10:9) * srcxstep;
+	srcy -= util::sext(m_videoram[0x1ff32/2 + 2 * bgnum], (dstystep != 0x200)?10:9) * srcystep;
```
cuavas, line 744: If this is going to be merged, it needs a comment here to note that this is effectively a per-game hack masquerading as a global hack to fix this one case.  Also, please don’t code like your space bar is broken.

mahoneyt944: I've updated the code with a more detailed note and added some spacing.

happppp: Does the hw not have a zoom flag?
I see the same thing here, where it thinks zoom is enabled only if that value is not 0x200:

	/* enable this code below to display zoom information */
#if 0
	if (dstxstep != 0x200 || dstystep != 0x200)
	(...)

If a value of 0x200 really disables zoom and there is no specific flag for it, then no I don't think it's a hack.

cuavas: It’s not the way it’s checking for zoom being enabled that looks like a hack.  I still think ignoring one bit of the value unless zoom is enabled looks a lot like a hack, and I’d want to see some pretty convincing evidence to the contrary to believe otherwise.

mahoneyt944: The code originally ran on 9 for all games before it was changed to 10 as a "temporary patch" for radr 20 something years ago. I wouldn't say we are ignoring a bit at 9...I would say it just uses more resolution when zooming.

It explains why all the games seem to suggest there's 9 bits of resolution and explains why the couple of outliers would use more. It tests working with all games at least...

I did update the PR to note that it might be a hack and explained to two most relevant cases it aims to correct. So hopefully that satisfies both sides to this.

mahoneyt944:
@rb6502 @cuavas 

Does this updated PR, spacing and notes added, adequately cover what you wanted amended?

mahoneyt944:
Hi, is this still being considered? I can close this PR if not. Thank you.

PR #13050, opened 2024-12-06
added Boku wa Plarail Untenshi Shinkansen de Ikou! (playable but SEEPROM still needs hooking up)

mamehaze:


mamehaze:
it should be the prequel (non-plus version) of prailpls in generalplus_gpl16250_spi.cpp

mamehaze:
this might be of interest to @sp1187 too as it has sound looping issues (as in the engine sounds don't loop properly at all, leaving a lot of the game in silence)

mamehaze:
I'm not entirely sure where / how the SEEPROM hooks up, I *think* the writes are on port C, but the read pattern doesn't match what I'd expect to actually read from it.  Going to mark this as ready and just revisit later.

cuavas (commented):
There are no comments or warning flags to indicate that the serial EEPROM isn’t hooked up.  It should at least be noted in the source.

mamehaze:
added a note

cuavas:
Also missing credits, but I assume it’s another TeamEurope dump.

mamehaze:
yeah, sorry, this is another I picked up and TeamEurope dumped

PR #13051, opened 2024-12-06
added dumps from 6 more Bandai Karaoke Station units [David Haywood, Team Europe]

mamehaze:


mamehaze:
I want to try and extract lists of what is on each unit before undrafting this.

one of them might be a virgin / unused unit, the rest seem to have been used at least once, and have downloaded data / session data on them.

it does seem like the downloads in the flash are tied to the unit somehow, hopefully it's nothing too complex; the bootloader ROMs differ only by the bytes at the end which may or may not be some kind of key.

mamehaze:
writing generic code to extract the song lists is a bit tricker than expected due to the different data around them in each case (probably some per-device scrambling of downloaded data) so I'll just mark this as ready for now.

PR #13052, opened 2024-12-07
apple2gs: implement the LANGSEL soft switch

as-tb-dev:
This PR implements the Apple IIgs LANGSEL soft switch that permits alternate character sets to be displayed

It requires a change in character ROM.  The current apple2gs.chr does not have all the alternate character sets.

The megaii.chr is created based on data from the Apple IIgs Diagnostic, V3.1 that is also part of this PR.  Since it is not an actual dump, it is marked as BAD.

To validate, run mame with: mame apple2gs a2diag31

Pick these options at the diagnostics menu:
c) Choose Tests -> h) Video Pattern Tests -> b. Character Generator Test
b) Execute Selected Tests

This alternates between a hi-res image of the character set and the text mode displaying the characters.  There should no longer be any flashing characters.


rb6502:
Neat!  There is an undocumented way on the IIgs to read the character set data out of the Mega II but I don't think anyone's released the data yet and my IIgs isn't currently set up.  I believe it's the same character data as the various international IIes though, and the match with the hires versions backs that up.

as-tb-dev:
Apple documentation seems to also confirm that the international character set assignments are the same between the IIgs and IIe.  The IIgs adds a Danish character set which I do not see referenced in any IIe documentation.

The character bit patterns may be a bit different in a few cases.  apple2eeuk character bit-patterns match OK.  With the apple2eefr, the pound sign looks a bit different between the IIgs and IIe using 342-0274-a.e9.

hash/apple2_flop_misc.xml
```diff
@@ -49,6 +49,25 @@ Generated by SLIST 0.2.1
...
+		    Apple IIc Diagnostic, V3.0
+		    Apple IIe Diagnostic, V3.0
+		-->
+		<part name="flop1" interface="floppy_3_5">
+			<dataarea name="flop" size="819200">
+				<rom name="apple ii diagnostic v3_1.po" size="819200" crc="50d3243d" sha1="23881d0096d428f2ea8568c854dc25f6b66ff051" />
+			</dataarea>
+		</part>
+	</software>
+
```
cuavas, line 70: There’s no dump credit for this – where did it come from?

as-tb-dev: > There’s no dump credit for this – where did it come from?

It has been a while since I had it.  I see it is available from at least two sites without any credits given on either one. 

Do all submitted dumps need a credit?  If so, I will remove it since I mostly just added it to support testing the PR.

happppp: Credit yourself or anonymous.

as-tb-dev: I have credited anonymous in the main PR description just now since I did not create the dump myself.  Is it just in the PR form that I do this?

as-tb-dev:
> I believe it's the same character data as the various international IIes though, and the match with the hires versions backs that up.

While working on another PR, I found another difference.  On the IIc and IIe (except perhaps the Platinum IIe), the Swedish keyboard has what looks to be a macron symbol instead of the tilde symbol that it has on the IIgs.

rb6502:
@as-tb-dev I just happened to see someone talking about how to read the real Mega II font data again.  Hit me up: messdrivers [at] gmail [dot] com and I can send you the details.  You'll need to modify the code to flip the LANGSEL switch and get all of the character sets, but it should work otherwise.  Then we can get rid of the BAD_DUMP once and for all.

rb6502:
@as-tb-dev Hi!  I just noticed this broke the initial text screen in Total Replay somehow.  Can you take a look?

as-tb-dev:
> @as-tb-dev Hi! I just noticed this broke the initial text screen in Total Replay somehow. Can you take a look?

Hi.  I tested it using mame-0.275-1.fc41.x86_64 and mame-0.270-1.fc41.x86_64 just now using these two commands:
mame apple2ee -sl7 cffa2 -hard1 "Total Replay v5.2.hdv"
mame apple2gs -sl7 cffa2 -hard1 "Total Replay v5.2.hdv"

The only difference in the text screen I see is that on the Apple IIe, it says "TOTAL REPLAY", "128K", "Mockingboard in slot 4", "Stereo", and "v5.2".  On the Apple IIgs, it says "TOTAL REPLAY", "IIgs", and "v5.2".  Both enter the same graphical screen, and I can press ? to enter help mode.  I am able to start a game.  I see the same results with both 270 and 275.

Am I testing it correctly to reproduce the issue?

rb6502:
I was testing on 5.1.  I just downloaded 5.2 and it does seem to be fine, and so does 5.0.  I guess it's something they temporarily broke in 5.1.  False alarm, sorry!

PR #13054, opened 2024-12-08
Sync bgfx, bx and bimg with upstream as of 20241207

belegdol:
Sync bgfx, bx and bimg with upstream:
- bgfx to https://github.com/bkaradzic/bgfx/commit/f6baa4bdaf7dad27785889f6adcefa2ec81ada92
- bx to https://github.com/bkaradzic/bx/commit/1faeaee6615a3548efbe04ca092386af2e10c165
- bimg to https://github.com/bkaradzic/bx/commit/1faeaee6615a3548efbe04ca092386af2e10c165

The rationale for the sync are recent wayland improvements [1], [2], [3]. As an unfortunate side effect, macOS 13 or newer is required.
This seems to pass the first tests but more are needed.

[1] https://github.com/bkaradzic/bgfx/commit/3303b451a1b3b1b264a854dbd4b29dffab4f7fbd
[2] https://github.com/bkaradzic/bgfx/commit/a6e372ead9daac0e8d867180cc06bc624a5e6f1a
[3] https://github.com/bkaradzic/bgfx/commit/e488a07f1be82cb6fbb2120f0d4f0a4f0139c558


belegdol:
I have found one regression so far: when running with `--bgfx_backend vulkan --bgfx_debug`, video will be messed up until first button press. It does not happen with `--bgfx_backend opengl` or without `--bgfx_debug`. The colours live are different than when taking a screenshot but the pattern is the same.
vulkan_bgfx_debug

belegdol:
I have just tested on windows. I have found no obvious issues with dx11, dx12 or opengl renderer, but the vulkan renderer exhibits the same issue as mentioned in my previous comment.

ICEknigh7:
Just in case that it's something which could be fixed with this update, here's a link to an easy to reproduce bug that only happens when using bgfx: [Brightness settings are randomly ignored when using bitmap_rgb32 in the screen_update function](https://mametesters.org/view.php?id=8586)

cuavas:
Bumping the requirement to macOS 13 “Ventura” is a non-starter.  We’ve only recently bumped it to macOS 11 “Big Sur”.  It’s too soon to bump it up by two major versions.

Also, after the debacle with the last attempted bgfx update, I’m pretty nervous about updating bgfx in general.

belegdol:
> Bumping the requirement to macOS 13 “Ventura” is a non-starter. We’ve only recently bumped it to macOS 11 “Big Sur”. It’s too soon to bump it up by two major versions.
> 
> Also, after the debacle with the last attempted bgfx update, I’m pretty nervous about updating bgfx in general.

Fair enough. Would you be willing to keep this PR open? This case, once we do want to update, pre-work would not need to be redone from scratch.

belegdol:
> Just in case that it's something which could be fixed with this update, here's a link to an easy to reproduce bug that only happens when using bgfx: [Brightness settings are randomly ignored when using bitmap_rgb32 in the screen_update function](https://mametesters.org/view.php?id=8586)

It does not look like it is fixed. Just as with 0.272, changing the gamma in `mslug2` has no effect.

PR #13055, opened 2024-12-08
ap2_dsk.cpp: fix loading bardstl3

SpecLad:
I accidentally broke this in 47d8c69b, where I forgot to account for the "oddball" images from the software list. AFAICS, bardstl3 is the only software to be broken, because its images are _shorter_ than they're supposed to be.

In addition, disk 3 of this image has a different size that's not listed in `identify()`, so it couldn't even be loaded before that commit. Fix that too.

### DIFF
--- a/src/lib/formats/ap2_dsk.cpp
+++ b/src/lib/formats/ap2_dsk.cpp
@@ -222,7 +222,7 @@ int a2_16sect_format::identify(util::random_read &io, uint32_t form_factor, cons
 	if (
 		size != APPLE2_TRACK_COUNT * 16 * APPLE2_SECTOR_SIZE
 		&& size != APPLE2_STD_TRACK_COUNT * 16 * APPLE2_SECTOR_SIZE
-		&& size != 143403 && size != 143363 && size != 143358
+		&& size != 143403 && size != 143363 && size != 143358 && size != 143195
 	)
 	{
 		return 0;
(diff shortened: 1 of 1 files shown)


src/lib/formats/ap2_dsk.cpp
```diff
@@ -313,7 +313,9 @@ bool a2_16sect_format::load(util::random_read &io, uint32_t form_factor, const s
 		uint8_t sector_data[APPLE2_SECTOR_SIZE*16];
 
 		auto const [err, actual] = read_at(io, fpos, sector_data, sizeof sector_data);
-		if (err || actual != sizeof sector_data)
+		// Some supported images have oddball sizes, where the last track is incomplete.
+		// Skip the `actual` check to avoid rejecting them.
+		if (err /* || actual != sizeof sector_data */)
 			return false;
```
cuavas, line 319: Isn’t this going to behave non-deterministically on these images with incomplete final tracks, because the sector data buffer is uninitialised?

What do these images represent?  Disks with fewer than the standard number of sectors on the final track?  Truncated sectors implicitly zero-filled?  Something else?

Should it be checking that actual length read is a whole number of sectors?  Should it only allow the read to return short on the final track?  Should it generate a track with fewer than the standard number of sectors if the read returns short?

SpecLad: > Isn’t this going to behave non-deterministically on these images with incomplete final tracks, because the sector data buffer is uninitialised?

Yeah, but that's the way it was before 47d8c69b. I just reverted to the original behavior.

> What do these images represent? Disks with fewer than the standard number of sectors on the final track? Truncated sectors implicitly zero-filled? Something else?

My best guess is that they are simply incorrectly truncated images. I could later double-check with the woz images whether the missing bytes are actually zeroes or if any data was lost.

> Should it be checking that actual length read is a whole number of sectors? Should it only allow the read to return short on the final track? Should it generate a track with fewer than the standard number of sectors if the read returns short?

Personally, I think that `bardstl3` should be removed from the list, given that its images are malformed, and there's a better replacement (`brdtale3`). Then the check can be restored. But I don't know what the MAME policy is for such cases.

galibert:
There is no policy.  Softlists are an unholy mixup of high-quality curated images and stuff so bad tosec looks goods in comparison.  In apple land, I'd say woz is curated and the rest... not as much.

PR #13056, opened 2024-12-08
Adjust macOS CI to use three simultaneous jobs

belegdol:
As of December 2024, all the used gitlab hosted runners have 7 GB of RAM and 14 GB of storage. macOS runner has 3 CPUs whereas Linux and Windows runners have 2. Since Windows and Linux jobs already use 3 jobs, it should also work for macOS.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -34,7 +34,7 @@ jobs:
       env:
         USE_LIBSDL: 1
         TOOLS: 1
-      run: make -j2
+      run: make -j3
     - name: Validate
       run: ./mame -validate
     - uses: actions/upload-artifact@main


belegdol:
macOS job only got two minutes faster which could just as well be the normal variance. Looks like this might not be worth it.

cuavas:
I don’t want to do this, because on a Linux VM with 6GB of RAM and lots of things stopped to free up memory (e.g. graphical session manager stopped), a full build hits swap at a couple of points even with `-j4` due to the way it ends up trying to build the memory system templates in `libemu` and the Lua bindings templates in `libfrontend` at the same time, and the M68000 CPU core files.  I want to stick with `-j2` on these 7GB VMs to ensure there’s some headroom.  I’ve explained this every time someone opens a PR to do this.

PR #13061, opened 2024-12-09
hash/ibm51*.xml: New FreeDOS 1.3 hard disk image, FreeDOS 1.4 images

chungy:
This one is installed with the CD-ROM version of the operating system, and the “Full” installation option selected.  It is much more representative of what FreeDOS is supposed to be than using the floppy disk set as in the previous image.

I've also made a repository to preserve and document the INP files used to create the FreeDOS disk images: https://github.com/chungy/mame-reprod-chd/tree/master/freedos13

### DIFF
--- a/hash/ibm5170_hdd.xml
+++ b/hash/ibm5170_hdd.xml
@@ -55,7 +55,7 @@ license:CC0-1.0
 		<info name="version" value="1.4" />
 		<part name="hdd" interface="ide_hdd">
 			<diskarea name="harddriv">
-				<disk name="freedos14" sha1="c1ffb2c1cbf3f9c8f6193d5ba9c5c01ce74d3d4e" writable="yes" />
+				<disk name="freedos14" sha1="c1ffb2c1cbf3f9c8f6193d5ba9c5c01ce74d3d4e" writeable="yes" />
 			</diskarea>
 		</part>
 	</software>
--- a/hash/ibm5170_hdd.xml
+++ b/hash/ibm5170_hdd.xml
@@ -47,6 +47,19 @@ license:CC0-1.0
 		</part>
 	</software>
 
+	<software name="freedos14">
+		<description>FreeDOS (version 1.4)</description>
+		<year>2025</year>
+		<publisher>The FreeDOS Project</publisher>
+		<info name="language" value="English" />
+		<info name="version" value="1.4" />
+		<part name="hdd" interface="ide_hdd">
+			<diskarea name="harddriv">
+				<disk name="freedos14" sha1="c1ffb2c1cbf3f9c8f6193d5ba9c5c01ce74d3d4e" writable="yes" />
+			</diskarea>
+		</part>
+	</software>
+
 	<software name="msdos331">
 		<description>MS-DOS (Version 3.31)</description>
 		<year>1987</year>
--- a/hash/ibm5170_hdd.xml
+++ b/hash/ibm5170_hdd.xml
@@ -42,7 +42,7 @@ license:CC0-1.0
 		<info name="usage" value="FreeDOS contains a CPU detection bug in VINFO.COM, always reporting a 80186.  Edit C:\FDAUTO.BAT to remove calls to VINFO and always run Support286 or Support386 routines to work around the problem." />
 		<part name="hdd" interface="ide_hdd">
 			<diskarea name="harddriv">
-				<disk name="freedos13" sha1="72eef7bf0ad53b8a240ce05244173df54db923f0" writeable="yes"/>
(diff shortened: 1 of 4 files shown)


angelosa:
In order to make repeated actions more portable between MAME versions you should use LUA script, not .inp files. Once latter breaks for any kind of timing change then you have to rebuild said .inp from scratch.

chungy:
> Once latter breaks for any kind of timing change then you have to rebuild said .inp from scratch.

Wouldn't that same problem exist with Lua scripts?

angelosa:
Well, _if_ API changes yes. But imo it's the kind of breakage akin to an integration test, and you don't necessarily need the full API spec for such task (i.e. for computers is barely `emu::keypost`, gathering the frame number then eventually check RAM/VRAM program flow).

chungy:
I don't really understand. If emulation timing changes to break an INP recording, wouldn't the same apply to a Lua script?

angelosa:
Lemme put an example, following is a border flicker quick test for c64 (run with `-autoboot_script <filename>` in CLI options).

```lua
local frame_num = 0

local function process_frame()
	frame_num = frame_num + 1

	if frame_num == 1 then
		manager.ui.show_fps = true
		manager.machine.video.throttled = false
	end
	if frame_num == 120 then
		emu.keypost("10 FOR I=0 TO 16\n")
		emu.keypost("20 POKE 53280, I\n")
		emu.keypost("30 NEXT I\n")
		emu.keypost("40 GOTO 10\n")
		emu.keypost("RUN\n")
	end
	if frame_num == 60*10 then
		manager.machine.video.throttled = true
	end
end

emu.register_frame_done(process_frame)
```

afaik works for all c64 and clones for what's intended for (making border to colorize at scanline level). 
However this won't work for c65 off the bat, I cheated at the time and just used an .inp for making it to fallback to c64 at boot. Said .inp was made in 0.268 cycle, it still works atm just because nobody touched c64/c65 emulation since then, but once that happens I'm positive that I'll either have to redo the .inp or just get rid of that and make it query the machine type for what I need here.

Bottom line: a very good reason to use lua script for automation tasks is for _portability_, like having FreeDOS to install on `ct486` / `pcipc` / `shutms11` et al., something you can't guarantee with .inp files by definition.

angelosa:
Another thing: [FreeDOS 1.4](https://www.freedos.org/download/announce.html) has been released. Do you mind adding it from this very PR? I'll take a look over the weekend, thanks in advance.

chungy:
I can do that. The installation media will be trivial. I may try to do a Lua-based install script, though it seems like it'll be a bit of a nightmare of trial-and-error to get everything right in it. Haha.

sairuk:
@angelosa I'd started building a demo out for this [here](https://github.com/mameau/mame-plugins/tree/main/plugins/act486auto) a while ago but its not guaranteed to run atm, although i did test the msdos622 at486 installer this morning.  I still have a number of changes from last year to push but i also need time to review beforehand, i think it may be work related to modifier key support. With my tool mapping an install right now is a tedious process, but an inp convertor or keylogger was in the plans before i moved on to other things.

i ran into the (documented) shortcomings of `emu.keypost` pretty early on so ended up using `ioport` where i may have been able to use `natkeyboard`.

it would be good if this or something else was available to support some project external to MAME where there could be a larger automation library for folks to build out a "playlist" (for lack of a better word) to install/run things.

chungy:
With some experimentation done in using Lua scripting, especially with @sairuk's plugin, I've decided against using it for FreeDOS 1.4.  I added a post to the forum thread I linked in the previous comment, and I also added a [wiki page explanation](https://chiselapp.com/user/chungy/repository/mame-reprod-chd/timeline?r=act486auto-test) to a "act486auto-test" branch I created to represent the last state of my efforts to go that route.

tl;dr: besides being tedious, MAME also desynced too frequently to depend on it. I am falling back to the INP method until Lua scripting can be improved enough.

angelosa:
Sounds like some drift latency happening, from the IDE as first suspect. Will investigate.

hash/ibm5170_hdd.xml
```diff
@@ -42,7 +42,20 @@ license:CC0-1.0
...
+
+	<software name="freedos14">
+		<description>FreeDOS (version 1.4)</description>
+		<year>2025</year>
+		<publisher>The FreeDOS Project</publisher>
+		<info name="language" value="English" />
+		<info name="version" value="1.4" />
+		<part name="hdd" interface="ide_hdd">
+			<diskarea name="harddriv">
+				<disk name="freedos14" sha1="c1ffb2c1cbf3f9c8f6193d5ba9c5c01ce74d3d4e" writable="yes" />
```
angelosa, line 58: Typo: `writeable` not `writable`

PR #13067, opened 2024-12-10
jaleco/megasys1.cpp: Hook up MCU program for EDF

sergiopolog:
* MCU internal ROM dumped by "Darren Olafson"


src/mame/jaleco/megasys1.h
```diff
@@ -51,11 +51,13 @@ class megasys1_state : public driver_device
...
 
+	void system_base(machine_config &config);
+
+	void system_B(machine_config &config);
 	void system_B_monkelf(machine_config &config);
+	void system_Bbl(machine_config &config);
 
 	void system_C(machine_config &config);
-	void system_Bbl(machine_config &config);
-	void system_base(machine_config &config);
```
cuavas, line 58: The abstract machine configurations (e.g. `system_base`) should be `protected` rather than `public`.  Also may as well put `ATTR_COLD` on them while you’re moving them around.

sergiopolog: Done. When all MCUs Internal ROMs will be dumped, "iosim" class could be deleted and those abstract methods will not be needed anymore.

I also reordered methods, grouping them in a more clear way.

PR #13069, opened 2024-12-10
radikalb: add output start lamp to Gaelco Radikal Bikers

KabukiFlux:
This PR adds the Start Lamp output for Radikal Bikers, now it can be used on external devices to simulate the original European Gaelco cabinet version.


startaq:
You can directly connect the lamp to the latch, no need for a trampoline: `m_outlatch->q_out_cb<2>().set_output("start_lamp")`

The comment "Arbitrary data written" makes me think that this isn't just simple lamp data though.

KabukiFlux:
It can be connected directly to the latch if no buffer is needed, I can assure you that the output of this bit in radikalb is the correct start lamp.

KabukiFlux:
@startaq cleanup code and verified behavior in Gaelco games:

- speedup
- surfplnt
- radikalb
- footbpow

Thank you.

PR #13070, opened 2024-12-11
Cherry-pick wayland improvements from upstream bgfx

belegdol:
As full sync with upstream is not an option at the moment (see https://github.com/mamedev/mame/pull/13054#issuecomment-2531577035), cherry-pick recent wayland fixes from upstream bgfx instead.

This uses the API more correctly as well as moves wayland detection to runtime, reducing the build requirements slightly.


src/osd/modules/render/drawbgfx.cpp
cuavas, line 1: What’s the rationale for this change?  Does SDL always create the Wayland EGL window now?  If it does, when did that change happen?  Does it raise the minimum SDL requirement?

cuavas (commented):
This seems to break building on macOS:
```
Compiling 3rdparty/bgfx/src/renderer_gl.cpp...
../../../../../3rdparty/bgfx/src/renderer_gl.cpp:7148:59: error: too many arguments to function call, expected single argument '_nwh', have 3 arguments
                m_swapChain = s_renderGL->m_glctx.createSwapChain(_nwh, _width, _height);
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^~~~~~~~~~~~~~~
In file included from ../../../../../3rdparty/bgfx/src/renderer_gl.cpp:9:
In file included from ../../../../../3rdparty/bgfx/src/renderer_gl.h:1158:
../../../../../3rdparty/bgfx/src/glcontext_nsgl.h:29:16: note: 'createSwapChain' declared here
                SwapChainGL* createSwapChain(void* _nwh);
                             ^
1 error generated.
make[2]: *** [../../../../osx_clang/obj/x64/Release/3rdparty/bgfx/src/renderer_gl.o] Error 1
```

src/osd/modules/render/drawbgfx.cpp
belegdol: The way I understand it, wayland detection is moved from compile-time to run-time with these changes. If SDL is built with Wayland support (SDL_VIDEO_DRIVER_WAYLAND), then the required Wayland library will be dlopened at runtime and things will just work. There is some explanation in the three upstream bgfx PRs, which have prompted me to submit this PR:
- https://github.com/bkaradzic/bgfx/pull/3358
- https://github.com/bkaradzic/bgfx/pull/3359
- https://github.com/bkaradzic/bgfx/pull/3360

belegdol: It appears that creating wayland EGL window is taken care by glcontext. Quoting https://github.com/bkaradzic/bgfx/pull/3358:
> In this PR, the EGL context now creates this wl_egl_window when the NativeWindowHandleType is set to Wayland.

belegdol:
> This seems to break building on macOS:
> 
> ```
> Compiling 3rdparty/bgfx/src/renderer_gl.cpp...
> ../../../../../3rdparty/bgfx/src/renderer_gl.cpp:7148:59: error: too many arguments to function call, expected single argument '_nwh', have 3 arguments
>                 m_swapChain = s_renderGL->m_glctx.createSwapChain(_nwh, _width, _height);
>                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       ^~~~~~~~~~~~~~~
> In file included from ../../../../../3rdparty/bgfx/src/renderer_gl.cpp:9:
> In file included from ../../../../../3rdparty/bgfx/src/renderer_gl.h:1158:
> ../../../../../3rdparty/bgfx/src/glcontext_nsgl.h:29:16: note: 'createSwapChain' declared here
>                 SwapChainGL* createSwapChain(void* _nwh);
>                              ^
> 1 error generated.
> make[2]: *** [../../../../osx_clang/obj/x64/Release/3rdparty/bgfx/src/renderer_gl.o] Error 1
> ```

Looks like this is fixed. I do not have access to a mac though so testing would be appreciated - especially the GL renderer.

cuavas:
What’s the prevailing opinion on this?  @galibert, @rb6502 do you have thoughts on it?

rb6502:
> What’s the prevailing opinion on this? @galibert, @rb6502 do you have thoughts on it?

I'll pull the branch and smoke test it on at least macOS.  I believe there's an issue with MAME itself that's causing a lot of the BGFX issues where we're misreporting the width and/or stride and that's why so many drivers segfault during a memcpy().

rb6502:
The branch does not compile.

3rdparty/sol2/sol/sol.hpp:19875:31: error: address of overloaded function 'call' does not match required type 'int (lua_State *)'

cuavas:
If we want to merge it, it’s probably better to do it sooner rather than later so Linux users are exposed to it for a while.

angelosa:
Gonna test it locally.

PR #13074, opened 2024-12-12
superxavix: some research on the bitmap layer, slight visual improvement where it gets used + general XaviX refactoring and improvements + 7 new sets

mamehaze:
- Started to refactor XaviX bus accesses, moved some in-cartridge logic for e-Kara SEEPROMs to cartridge space
- XaviX external bus accesses can now go through memory, avoiding the speed-up bypass
- Mapped RAM in external space for Super TV-PC sets, as well as a ROM bank, allowing bitmaps in RAM to render
- Added Bitmap plotter, again used by Super TV-PC sets for constructing some bitmaps and the paint program in Super TV-PC Hello Kitty
- Added previously unsupported video mode used by Excite Stadium
- Swapped background order when priorities are equal, issue higlighted by Excite Golf 
- Added some XaviX2000 (2nd Gen and SuperXaviX) opcodes needed by Let's! TV Play Touin Ougi Onmyou Taisenki: Mezase Saikyou Toushinshi for sprites to show
- Added reduced size sprite DMA transfers, used by Choro-Q Byun Byun Racer 
- Allow SuperXaviX sets to render in higher horizontal resolutions for proper display of the bitmap layers, doubling the standard XaviX output to match
- Began some research on SuperXaviX extended ROM modes, used by anpanmdx for sprites and the bitmap layer

new NOT WORKING machines
----------------------------------
Anpanman Kazoku De Ikunou Mat DX (Japan) [TeamEurope, David Haywood]
Let's TV Play Dragon Ball Z Battle Experience Kamehameha 2 Ossu Ome Goku Tenkaichi Budokai (Japan) [TeamEurope, David Haywood]
Doraemon Moving! Oekaki (Japan) [TeamEurope, David Haywood]
Anpanman Pyon-Pyon Ikunou Mat (Japan) [TeamEurope, David Haywood]
Doraemon anywhere - Japan travel game DX experience! Where is the Dragon Grand Prix! (Japan) [TeamEurope, David Haywood]
Let's! TV Play Futari wa PreCure MaxHeart Dance on the mat Let's go to MaxHeart (Japan) [TeamEurope, David Haywood]
Let's! TV Play Disney Characters Oto! Iro! Ton-Ton! Miracle Parade [TeamEurope, David Haywood]
Hello Kitty Piano PC (Japan)  [TeamEurope, David Haywood]


mamehaze:
the title of the tmy_thom set could probably be improved now the title screen renders more correctly

ajrhacker:
The Japanese title is Soreike Thomas - Sodor Tou no Nakamatachi (それいけトーマス ソドー島のなかまたち). The English title (Thomas & Friends on the Island of Sodor) should be noted too, though.

mamehaze:
there are some baby steps in here regarding a general XaviX cleanup, with a view to moving the SoC logic into a core, rather than the driver at some point.  There will likely be a performance hit to doing this, but it feels like the right time to start doing it.

mamehaze:
While there are definitely still some issues, being able to map RAM into external space, and adding the bitmap plotter allows the Super TV-PC units to start drawing to their bitmap layer, which is important as they decompress data there.  Some of the data that gets copied is corrupt though, I suspect due to bad XaviX2000+ opcodes, although there could also be a data conversion feature I'm missing.






This will eventually lead to a cleaner implementation of the XaviX memory handling, even if it'll be a few PRs beyond this one before we're fully there.

mamehaze:
The bitmap plotter was further improved, might not be perfect, but the different width data can now be written which fixes things like the filling in the Hello Kitty paint demo in attract

mamehaze:
Outside of that, this set of changes gives improvements to various other SuperXaviX games using the bitmaps.



rockman_xavix_1
rockman_xavix_2

DopefishJustin:
> The bitmap plotter was further improved, might not be perfect, but the different width data can now be written which fixes things like the filling in the Hello Kitty paint demo in attract

Definitely some issues with the text rendering still. It's readable but seems to be missing pixels on the thinner lines, and there seem to be extra pixels on the parentheses in the answers making them look doubled up. E.g. I think the top left answer should say "本（ほん）"

Also, are the colours supposed to be so desaturated?

mamehaze:
It's interlaced, and probably meant to be twice the x resolution, but until
I understand the crtc stuff this is the best I can get, I'll need to pixel
double the standard xavix output to get the higher red bitmaps superxavix
supports. One step at a time.

mamehaze:
cq1
cq2
cq3

the current changes improve the visuals on Choro Q, although it still crashes shortly after this (it doesn't crash if you return 0 from the math unit, although then the cars don't move either)

mamehaze:
there do still seem to be some plotter issues, but it is a bit better with the full horizontal resolution for the bitmap layer
https://www.youtube.com/watch?v=IDu_wXxyOow&t=224s

mamehaze:
FWIW, a future PR will likely move the XaviX chips to be SoC/CPU implementations, but that will require significantly more refactor and is out of the scope for this.

The work done here is basically tweaks and improvements based on new knowledge gathered from recent dumps.

mamehaze:
some of the other Super TV-PC carts can be booted now

mamehaze:
epo_golf title screen has tiles of the same priority, and expects the draw order to be swapped compared to the old code or the menu doesn't show.  I don't think any other cases used the same priority values before this.

mamehaze:
ban_ordj required a missing XaviX2000+ opcode (cmc / complement carry) to be implemented in order to show sprites

mamehaze:
Super TV-PC Hello Kitty & Super TV-PC Doraemon require a machine specific ROM bank to be hooked up in order to display the boot screen and desktop backgrounds properly.  (unclear why they do this, but the code seems explicit in what it expects)

mamehaze:
epo_stad uses a tile addressing mode we haven't seen anywhere else in one place, behind some of the dialog boxes

src/mame/tvgames/xavix.cpp
```diff
@@ -2014,134 +2047,134 @@ void xavix_state::init_xavix()
 ***************************************************************************/
 
 ROM_START( taitons1 )
-	ROM_REGION( 0x200000, "bios", ROMREGION_ERASE00 )
+	ROM_REGION( 0x800000, "bios", ROMREGION_ERASE00 )
 	ROM_LOAD( "taitonostalgia1.u3", 0x000000, 0x200000, CRC(25bd8c67) SHA1(a109cd2da6aa4596e3ca3abd1afce2d0001a473f) )
 ROM_END
```
cuavas, line 2052: The fact that you need to change every single one of these to pad the ROM region out to 8 MiB should indicate that you’re doing this in a very fragile way that creates a lot of extra work when something changes.  Can you think of a way to do this that doesn’t require doing stuff like this?

src/mame/tvgames/xavix.h
```diff
@@ -854,25 +921,51 @@ class xavix_cart_state : public xavix_state
...
+	{
+		if (m_cartslot->has_cart())
+		{
+			m_cartslot->write_cart(offset, data);
+		}
+		else
+		{
+			logerror("%s: unhandled write access to cart area with no cart installed %08x %02x\n", machine().describe_context(), offset, data);
+		}
+	}
```
cuavas, line 948: To avoid taking the performance hit on every memory access, can you install the cartridge ROM on start if a cartridge is loaded?

src/mame/tvgames/xavix.h
```diff
@@ -854,25 +921,51 @@ class xavix_cart_state : public xavix_state
...
-					return m_cartslot->read_cart(offset);
+					return m_rgn[(offset) & (m_rgnlen - 1)];
 				}
 				else
 				{
-					return m_rgn[(offset) & (m_rgnlen - 1)];
+					return cart_r(offset);
 				}
 			}
 		}
```
cuavas, line 971: This is another place you’re hurting performance by doing a pile of checks on every access that should be possible to avoid entirely by installing things in the right place on start (`offset & 0x8000`, `(offset & 0x7fffff) >= m_cartlimit`, etc.) or using a memory view (`m_disable_memory_bypass`).

src/mame/tvgames/xavix.h
```diff
@@ -884,19 +977,19 @@ class xavix_cart_state : public xavix_state
...
-				return m_cartslot->read_cart(offset);
+				return m_rgn[(offset) & (m_rgnlen - 1)];
 			}
 			else
 			{
-				return m_rgn[(offset) & (m_rgnlen - 1)];
+				return cart_r(offset);
 			}
 		}
 	}
```
cuavas, line 995: Here you’ve got dual maintenance because this code is duplicated.

src/mame/tvgames/xavix.h
mamehaze: I'm aware of that, this is why I've started refactoring it, again that's for a future PR, because I don't want this one to be too disruptive.

src/mame/tvgames/xavix.h
mamehaze: likewise, that's what is planned for a future PR, not this one

src/mame/tvgames/xavix.cpp
```diff
@@ -1871,6 +1899,11 @@ void xavix_cart_state::xavix_cart_daig(machine_config &config)
 	SOFTWARE_LIST(config, "cart_list_daig").set_original("takara_daigunder_dx_cart");
 }
 
+void xavix_cart_gcslottv_state::xavix_extbus_map(address_map &map)
+{
+	map(0x000000, 0x7fffff).rw(FUNC(xavix_cart_gcslottv_state::cart_r), FUNC(xavix_cart_gcslottv_state::cart_w)).mirror(0x800000);
+}
```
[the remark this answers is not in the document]
mamehaze: technically the mirroring is no longer needed, so I'll remove it

mamehaze:
basically the longer term strategy is to move the entire Xavix / SuperXaviX implementations into a device, then split the drivers up into individual files, rather than the huge ones we have now, each with its own memory map for the external bus, and any specific extras that need emulating for that unit.

part of doing that will be to get rid of the memory bypass stuff anyway, which was useful during earlier development (as it helped get around a few other issues) but now things are better understood, and there's enough evidence to be confident in changing how things work a little, it will likely end up going away.

the motivation for doing this part of the changes is that it allows for better testing of the new dumps, especially the SuperXaviX ones which really need more complex external bus setups (banking, RAM etc.) which means what gets dumped can be more easily verified, and there's less chance of adding a bad dump / TeamEurope removing something from the dumping rig if there are still issues.  I'm already pretty sure the Double Mouse Party dump is bad based on this research - I did have suspicions before, but the code was in no shape to run it at all then.

mamehaze:
I've added a better summary of the changes too

mamehaze:
The Anpanman game is the first SuperXaviX one we know to use the extended secondary bus feature which appears to be usable by the sprites and bitmap layer at least.  We have a few others with extended ROMs, but it isn't clear if they're doing the same, or have banking, because those don't boot far enough yet.  Unfortunately implementing this requires some extra checks in places, which slows things down, but the current line of thought is "make it work" while re-optimizing it will be looked into later.

It also uses the bitmap plotter to plot to main ram instead of extra RAM, which will need some work (for the graph in one scene)

There seem to be some opcode / logic bugs with a few of the mini-games too, but in this case I'd consider that a positive because the more 'not quite right logic' places we find, the better when it comes to debugging it.

mamehaze:
Doraemon Moving! Oekaki (Japan) [TeamEurope, David Haywood]
Anpanman Pyon-Pyon Ikunou Mat (Japan) [TeamEurope, David Haywood]
Doraemon anywhere - Japan travel game DX experience! Where is the Dragon Grand Prix! (Japan) [TeamEurope, David Haywood]
Let's! TV Play Futari wa PreCure MaxHeart Dance on the mat Let's go to MaxHeart (Japan) [TeamEurope, David Haywood]

likely need better translations at least, maybe the other Anpanman mat too

mamehaze:
Disney needs a better name

ajrhacker:
Oto! Iro! Don-Don! Miracle Parade (オト！イロ！トン・トン！ミラクルパレード)

mamehaze:
thanks, replaced that one.  I've left the "Let's TV Play" part because it's on the box even if it isn't on the screen

ajrhacker:
Correction: the transliteration should be Ton-Ton, not Don-Don.

mamehaze:
@cuavas @galibert 
I would appreciate a merge of this so that I can start to think about refactoring some of the other bus/memory access stuff that was mentioned, moving away from the code we've had for 6+ years.  As I've said above, I don't really want to commit to doing that until this part goes in, because it will make tracking regressions easier (as I have a feeling that part will be more disruptive)

There will likely be a few more SuperXaviX sets to add (and the odd XaviX2 set) afterwards, but this is the bulk of what was sent to Klaus now added.

PR #13082, opened 2024-12-14
heathzenith/h19/tlb.cpp: Make the page 2 memory option configurable for UltraROM

mgarlanger:
tlb.cpp: 

- Add machine configuration option to install/remove the optional page 2 memory for the UltraROM.

h89.cpp:

- Use class variable instead of tag for memory_view region routine.


mgarlanger:
This has been verified with the self-diag in the ROM (`ESC |`). It correctly reports the presence or absence of the page 2 memory. 
Screenshot 2024-12-29 at 1 08 51 PM
Screenshot 2024-12-29 at 1 09 16 PM

PR #13090, opened 2024-12-16
Advanced skeleton driver for the Memorymoog.

m1macrophage:
This driver is marked as ~a skeleton~ "not working", because it is not really usable without a layout. Nonetheless, a large portion of the digital logic and digital-analog interface is emulated. There is no attempt to emulate audio.

Following the Source (#12917), this is the second in a series of 8 (at the moment) synth drivers.

Similar to the previous PR, I've incorporated feedback I've seen in other people's PRs, but I am sure I missed things. Happy to fix as needed and proactively apply any feedback to future drivers.


angelosa:
"Advanced Skeleton driver" ... I guess you want to just mark the driver a `MACHINE_NOT_WORKING` instead, given it's essentially a double negation in wording:
https://github.com/mamedev/mame/blob/c41a16b3a8054eac3a68f9af79be23bd51f1f0b3/src/emu/gamedrv.h#L92-L93
We commented about this a couple months ago [here](https://github.com/mamedev/mame/commit/8b46595df803b0935c5f11b2e426f0accdb57e81#commitcomment-148113945) , de facto is a deprecated flag.

m1macrophage:
Done. Had to also add "| MACHINE_NO_SOUND" to pass validation.

m1macrophage:
While on the subject, any guidance on when one would use MACHINE_IS_INCOMPLETE instead of MACHINE_NOT_WORKING?

Robbbert:
MACHINE_IS_INCOMPLETE refers to a machine (most likely a homebrew) that was never finished. So some intended features won't work because they never got implemented. The inventor gave up on the project.

m1macrophage:
Ah, makes sense.

m1macrophage:
Hi, and happy new year!

An etiquette question: if I don't hear back on a PR after multiple weeks, should I assume that it is still on someone's radar, and they just haven't gotten around to it? Or should I assume it got forgotten, and send occasional reminders, say every few weeks?

Larger turnaround times are not a huge deal, as I can still make progress on my [backlog](https://github.com/m1macrophage/synthesizerdocs?tab=readme-ov-file#status)

The only concern is that someone else might end up doing duplicate work. So if others are working on synthesizers, please take a look at the link above, and let's coordinate if there are conflicts :).

By the way, I incorporated changes from commit 9e463d2 (in moog/source.cpp) into this driver (and all other drivers in my WIP list for that matter). Hopefully this helps with the review.

galibert:
Sorry, my fault...

PR #13095, opened 2024-12-18
namco/namco_cus4xtmap.cpp: Device-fied CUS42 + CUS43 tilemap hardware, Cleanups:

cam900:
namco/baraduke.cpp:
- Use video/resnet.h for palette initialization
- Simplify gfx decode layout
- Fix address map order

namco/namcos86.cpp:
- Use video/resnet.h for palette initialization
- Use set_pen_indirect and set_indirect_color for indirect color palette beheviour
- Simplify gfx decode layout
- Reduce unnecessary pointer
- Reduce runtime tag lookups
- Reduce duplicates
- Fix type value for some variables
- Fix address map order
- Fix spacing
- Fix naming

### DIFF
--- a/src/mame/namco/baraduke.cpp
+++ b/src/mame/namco/baraduke.cpp
@@ -107,10 +107,11 @@ DIP locations verified for:
 
 #include "emu.h"
 
+#include "namco_cus4xtmap.h"
+
 #include "cpu/m6800/m6801.h"
 #include "cpu/m6809/m6809.h"
 #include "machine/watchdog.h"
-#include "namco_cus4xtmap.h"
 #include "sound/namco.h"
 #include "video/resnet.h"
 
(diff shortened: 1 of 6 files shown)


src/mame/namco/baraduke.cpp
```diff
@@ -110,7 +110,9 @@ DIP locations verified for:
 #include "cpu/m6800/m6801.h"
 #include "cpu/m6809/m6809.h"
 #include "machine/watchdog.h"
+#include "namco_cus4xtmap.h"
 #include "sound/namco.h"
+#include "video/resnet.h"
```
cuavas, line 115: Local project headers (in this case, `"namco_cus4xtmap.h"`) should be included before headers from src/devices.

PR #13097, opened 2024-12-18
chore: Reduced MSVC warning about an "empty controlled statement"

holub:


holub:
@pmackinlay we are closing this. right?

PR #13099, opened 2024-12-19
Add documentation comments about the effects of the CVS sound registers observed by poking values on a real PCB.

prswan:
Comment update to add notes about the four sound registers 0x1884 -> 0x1887 as observed on a real PCB.


happppp:
"Mikes comment above about a high pitched whistle didn't match my PCBs. His was faulty?"
That written by me from me hearing your recording ;D

PR #13103, opened 2024-12-20
gave ninjemakt its own state as it requires galivan background decoding

mamehaze:
this set is a conversion of Galivan, but the text layer handling seems to be closer to Ninja Emaki (done with PCB mods) so easier to handle as a subclass of Galivan

also the 'red' PROM is a bad dump, while the background tilemap data ROM (which was marked as 'BAD' due to fixed bits) is actually correct, as that bit goes unused.

### DIFF
--- a/src/mame/nichibutsu/galivan.cpp
+++ b/src/mame/nichibutsu/galivan.cpp
@@ -209,6 +209,8 @@ protected:
 	void common(machine_config &config);
 	void video_config(machine_config &config);
 
+	virtual TILE_GET_INFO_MEMBER(get_bg_tile_info);
+
 private:
 	uint8_t m_scrollx[2]{}, m_scrolly[2]{};
 	uint8_t m_layers = 0U;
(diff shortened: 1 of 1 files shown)


src/mame/nichibutsu/galivan.cpp
```diff
@@ -272,12 +273,28 @@ class ninjemak_state : public galivan_state
...
+		galivan_state(mconfig, type, tag)
+	{ }
+
+protected:
+	virtual void video_start() override ATTR_COLD;
+
+private:
+	virtual TILE_GET_INFO_MEMBER(get_tx_tile_info) override;
+
+};
```
cuavas, line 295: Why have you made the tile info functions `virtual` when every class that wants a different implementation overrides `video_start` to create the tilemaps itself, and hence will pick up the implementation from the most derived class anyway?  It doesn’t need to go through the vtable at all.

PR #13104, opened 2024-12-21
bus/z29_kbd: Move z29 keyboard bus to bus/heathzenith directory

mgarlanger:
Moves the heath/zenith z29 keyboard bus from the top level bus directory into the bus/heathzenith directory.


ajrhacker:
As the author of this bus, I don't really see a compelling need for this move.

mgarlanger:
> As the author of this bus, I don't really see a compelling need for this move.

Agree, there isn't a compelling need for this to move, but I think it makes it better organized to have it under the company name with the other buses for heath/zenith equipment, if I'm in the minority, I can close the PR.

mgarlanger:
Closing since the move of the code is not wanted.

PR #13105, opened 2024-12-21
bus/heathzenith/h89: Add Z-89-11 Multi-Mode Interface Card option

mgarlanger:
When this new card is selected, the CP/M configur.com program properly detects it.

Screenshot 2024-12-20 at 10 41 00 PM


mgarlanger:
My changes doesn't seem to be the cause of the build failure.

PR #13106, opened 2024-12-21
Replace uPD1771c high level emulation with a cpu core.

wilbertpol:
cpu/upd177x/upd177x.cpp: Add NEC uPD177x cpu core.

scv.xml:
Star Speeder


src/devices/cpu/upd177x/upd177x.cpp
```diff
@@ -0,0 +1,797 @@
+// license:BSD-3-Clause
+// copyright-holders:Wilbert Pol
+/*
+
+TODO:
+- Only upd1771c supported
+- Noise, time, and external interrupts not implemented
+- CH1/CH2 not implemented
+- Not all instructions implemented
+- Are PNC1 and PNC2 updated every machine cycle or only when the noise counter overflows?
```
ReverendGumby, line 10: They're updated on noise interrupt trigger: rising edge of selected noise counter bit.

ReverendGumby, line 10: PNC1/2 are updated when a specific bit of the noise counter (selected by MD_NSF2/3) is cleared (negedge). That specific bit is bit 8 by default, so the update event precisely coincides with the noise counter overflowing. (The PNC1/2 update event is also the noise interrupt trigger.)

src/devices/cpu/upd177x/upd177x.cpp
```diff
@@ -0,0 +1,797 @@
...
+static constexpr u16 VECTOR_EXT   = 0x60;
+static constexpr u16 VECTOR_TIME  = 0x80;
+
+static constexpr u8 STACK_START = 0x20;
+
+
+upd177x_cpu_device::upd177x_cpu_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock)
+	: cpu_device(mconfig, type, tag, owner, clock)
+	, device_sound_interface(mconfig, *this)
+	, m_program_config("program", ENDIANNESS_BIG, 16, 16, -1, address_map_constructor(FUNC(upd177x_cpu_device::program_map), this))
```
ReverendGumby, line 67: Any hope that I can convince you to align with my MiSTer FPGA SCV core, and change the ROM to little-endian? Having two emulators which require different ROM endianness could cause chaos. :(

For my emulator, I chose endianness based on the behavior of the TBL* instructions. They byte-address the ROM in little-endian order, by selecting the lower byte when PR bit 0 is low (even). It also matches the RAM endianness for 16-bit stack accesses.

ReverendGumby:
Fantastic work on getting Star Speeder working and playing its lovely digital effects in MAME! ❤️

src/devices/cpu/upd177x/upd177x.cpp
```diff
@@ -0,0 +1,797 @@
...
+
+	set_icountptr(m_icount);
+
+	m_channel = stream_alloc(0, 1, clock() / 4);
+}
+
+
+void upd177x_cpu_device::device_reset()
+{
+	m_pc = 1; // Yes, this is odd, but otherwise there are sync issues between writing and clearing PA in scv kungfurd when playing back adpcm.
```
ReverendGumby, line 151: That's an elegant hack.

FWIW, on silicon, an external write (from the main CPU) takes priority over 'OUT PA'. And because the 'OUT PA' completes while the external write of 0x1F is still in progress, the external write wins.

src/devices/cpu/upd177x/upd177x.cpp
```diff
@@ -0,0 +1,797 @@
...
+
+void upd177x_cpu_device::op0xxx(u16 opcode)
+{
+	switch (opcode)
+	{
+	case 0x0000:  // NOP
+		break;
+	case 0x0002:  // OUT PA
+		m_pa = m_a;
+		m_pa_out_cb(m_pa);
```
ReverendGumby, line 191: There isn't much use for a callback here. There's no external indication of the 'OUT PA' at instruction execution time; the port pins don't immediately change. In order for the outside world to observe this instruction's effect, an external read (from the main CPU) must happen.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: That is how I have it currently implemented.
In the description of the MIX command there is some text mentioning that the PNC2 output is updated every machine cycle. That is where my confusion came from. 
Thanks for confirming that following the noise counter is correct.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: Just to confirm: the file you use has a crc of 975bd68d?

wilbertpol:
> Fantastic work on getting Star Speeder working and playing its lovely digital effects in MAME! ❤️

Thank you. It is not just Star Speeder, also Kung Fu Road and Pole Position 2 now have some speech being played back.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: Thanks for that background information. I already thought it should be something to do with low level timing between the chips.
I have updated the comment on that hack.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: You are right. I have removed the callback.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: Answering my own question:
I see we would have the same checksums.
What an interesting coincidence, I was not aware that you were also working on a cpu core for the upd1771.

scripts/src/cpu.lua
```diff
@@ -3929,6 +3929,8 @@ end
 --------------------------------------------------
 
 if opt_tool(CPUS, "UPD177X") then
+	table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/upd177x/upd177x.cpp")
+	table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/upd177x/upd177x.h")
 	table.insert(disasm_files , MAME_DIR .. "src/devices/cpu/upd177x/upd177xd.cpp")
```
cuavas, line 3934: You shouldn’t be adding the CPU core files to the disassembler files.  If unidasm depends on the CPU core files, something is very wrong.

src/devices/cpu/upd177x/upd177x.cpp
```diff
@@ -0,0 +1,794 @@
...
+		m_tn_int = false;
+		break;
+
+	case 0x0005:  // STF
+	case 0x0101:  // MON
+	case 0x0501:  // JMPA
+	case 0x0602:  // OFF
+	case 0x0801:  // RETS
+	default:
+		fatalerror("%04x: Unsupported instruction %04x\n", m_pc - 1, opcode);
```
cuavas, line 256: You should at least include the device tag in the fatal error messages – it isn’t included automatically like it is for `logerror`.

src/devices/cpu/upd177x/upd177x.h
```diff
@@ -0,0 +1,128 @@
...
+	// configuration helpers
+	auto pb_out_cb() { return m_pb_out_cb.bind(); }
+
+	void pa_w(u8 data) { m_pa = data; }
+	void pb_w(u8 data) { m_pb = data; }
+
+protected:
+	upd177x_cpu_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock);
+
+	// device-level overrides
```
cuavas, line 47: Please don’t copy/paste this comment.

src/mame/nec/apc.cpp
```diff
@@ -972,7 +972,7 @@ void apc_state::apc(machine_config &config)
 
 	/* sound hardware */
 	SPEAKER(config, m_speaker).front_center();
-	UPD1771C(config, m_sound, MAIN_CLOCK).add_route(ALL_OUTPUTS, "mono", 1.00); //uPD1771C-006
+	UPD1771C_HLE(config, m_sound, MAIN_CLOCK).add_route(ALL_OUTPUTS, "mono", 1.00); //uPD1771C-006
```
cuavas, line 975: Given this is the only other use of the µPD1771C in MAME, is it possible to migrate it and avoid having the extra implementation?

src/devices/cpu/upd177x/upd177x.cpp
ReverendGumby: I was also not aware of your work. No worries! It's a good time to be working on the Super Cassette Vision. The more people involved, the merrier. Happy to take what I've learned from the experience and help you.

src/mame/nec/apc.cpp
wilbertpol: Updated the app driver to use the new cpu core and removed the old implementation.

scripts/src/cpu.lua
wilbertpol: That was an error from my end, there is no dependency.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: I have updated all fatalerrors to include the tag.

src/mame/epoch/scv.cpp
```diff
@@ -587,6 +601,9 @@ ROM_START(scv)
 
 	ROM_REGION(0x400, "charrom", 0)
 	ROM_LOAD("epochtv.chr.s02", 0, 0x400, BAD_DUMP CRC(db521533) SHA1(40b4e44838c35191f115437a14f200f052e71509))
+
+	ROM_REGION(0x400, "upd1771c", 0)
+	ROM_LOAD16_WORD("upd1771c-017", 0, 0x400, CRC(975bd68d) SHA1(e777217c622331677ac1d6520a741d48ad9133c0))
```
ReverendGumby, line 606: I'm not sure what MAME's rules are for ROM filenames, but should it have an extension? We've been naming this ROM "upd1771c-017.s03"... just because ".s01" and ".s02" are listed above it.

src/devices/cpu/upd177x/upd177x.cpp
wilbertpol: No worries, no one knew I was working on this; it started out as an experiment. I am normally quite bad at sound emulation so I am very happy that I got this far with it.
Your comments are very much appreciated. Thank you.

src/mame/epoch/scv.cpp
wilbertpol: Normally the location on the pcb is put in the extension like u1, or u2, but I have not found any pictures of the pcb where such markings are visible for upd1771c chip.

ReverendGumby: TIL something new, thanks!
Here's a pic of my NTSC board. What a funny coincidence 😂 
image

ReverendGumby (approved):
Great work!

src/devices/cpu/upd177x/upd177x.h
cuavas: It’s the “level” comments that are a problem (“device-level overrides”, “interface-level overrides, etc.) – they make no sense.  If you want a comment for this section, write “device_t implementation” or similar.

cuavas:
> Great work!

Not a great idea to approve a pull request that breaks the build due to build scripts not being updated…

src/devices/cpu/upd177x/upd177x.h
wilbertpol: I have removed those comments

PR #13107, opened 2024-12-22
cpu/mpk1839/kl1839vm1.cpp: Improved VAX-11 code decoder in user space

holub:
Good progress here. CPU is able to decode/process VAX-11 instructions in user space.
image
image

### DIFF
--- a/src/devices/cpu/mpk1839/kl1839vm1.cpp
+++ b/src/devices/cpu/mpk1839/kl1839vm1.cpp
@@ -5,22 +5,26 @@
 
 #include "kl1839vm1.h"
 #include "kl1839vm1dasm.h"
+#include "cpu/vax/vaxdasm.h"
+#include <regex>
 
 
-#define VERBOSE ( LOG_GENERAL )
+#define LOG_VAX    (1U << 1)
+
+#define VERBOSE ( LOG_GENERAL | LOG_VAX )
 #include "logmacro.h"
 
 #define UNIMPLEMENTED(msg) LOG("Unimplemented: %s\n", msg)
+#define LOGVAX(...)        LOGMASKED(LOG_VAX, __VA_ARGS__)
 
-#define MCA    m_mca.w.l // Micro Command Address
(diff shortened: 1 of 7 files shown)


src/devices/cpu/mpk1839/kl1839vm1.cpp
```diff
@@ -5,22 +5,26 @@
...
 #include "kl1839vm1.h"
 #include "kl1839vm1dasm.h"
+#include "cpu/vax/vaxdasm.h"
+#include <regex>
 
 
-#define VERBOSE ( LOG_GENERAL )
+#define LOG_VAX    (1U << 1)
+
+#define VERBOSE ( LOG_GENERAL | LOG_VAX )
```
rb6502, line 14: Don't check in with VERBOSE enabled.

holub: Is it ok to keep LOG_GENERAL?

holub: done

src/devices/cpu/mpk1839/kl1839vm1.h
```diff
@@ -75,17 +95,24 @@ class kl1839vm1_device :  public cpu_device
...
+	PAIR            m_amc;    // Microdode PC
+	PAIR            m_ppc;    // previous program counter
+	bool            m_fp;
+	bool            m_jzdra_waiting;
+	u32             m_consts[0x10] = { 0x4, 0x2, 0x8, 0x1, 0x0, 0, 0, 0x66, 0, 0xc00000, 0xffffffff, 0x1f0000, 0x4000000, 0, 0, 0 };
+	PAIR            m_reg[0x20];
+	int             m_icount;
+	u32             m_op_size;
+
+	std::deque<u32> m_pcm_queue;
```
cuavas, line 114: You’ll need to change this to something else to support save states.  You can’t register a `std::deque` for save states.

PR #13108, opened 2024-12-22
Fixes to UML and DRC backends

987123879113:
I've been working on a project involving the DRC lately and found several bugs and inconsistencies when trying to compare results so I tried to fix bugs and make things more consistent between all of the backends.

Most of the fixes are to mostly unused code paths so I'm not expecting it to fix many, if any, of the known DRC issues.

I'd appreciate it if people could test the DRC in various games or operating systems before merging so I'm only sending this as a draft PR for now. I've tested mostly SH-4 (Naomi) and PPC (Firebeat) games and I'm not noticing any differences.

- Added new opcodes
    - BREAK: Hardware breakpoint that can be caught in a debugger to see exactly what assembly is being generated at any given time. (Warning: will crash the program if used without a debugger attached so use carefully)
    - SETFLGS: Set the hardware flags to a specific state. Meant for debugging such as testing opcodes in various states
    - MULUH and MULSH: Unsigned and signed multiplication where the result is only the lower half and the status flags are calculated based on the lower half. Previously the MULU and MULS opcodes would change how status flags were calculated based on whether the upper half output destination was set to something different than the lower half output destination so I decided to split them off to make their intended usages and status register calculations more explicit.
- General fixes
    - Standardized how shifts are calculated between the various backends so status flags and results should be the same now
- UML fixes
    - Fixed broken multiplication and division in the simplifier
    - Fixed shift calculations in the simplifier to work the same as the backends
    - Changed GETFLGS definition such that it will not change status flags in the process of reading out the status flags (and made appropriate changes in backends)
- C backend
    - Some opcodes wouldn't be recognized depending on the surrounding operation flag usages so I added the cases where I encountered unrecognized opcodes
    - Fixed various broken flag calculations
    - Fixed output when a shift is called with a shift of 0 (previously wouldn't set the output value at all)
- x86 backend
    - Many of the 64-bit operations were outright broken so I fixed those up as much as I could
    - Fixes to status flag calculations
- x64 backend
    - Fixed some unsafe code in the FNEG implementation that'd cause segfaults if the case was actually used anywhere (it wasn't)
    - Fixes to status flag calculations
- Added some extra documentation in uml.cpp for opcodes with cryptic namings and some status flag behavior
- Removed a lot of useless comments in the format of `a.test(ecx, ecx); // test ecx,ecx`. I didn't get them all because there's too many so mostly just on the code I touched. The comments almost 99% of the time are exactly the same as the assembly code it's beside so they aren't needed.


ajrhacker:
> MULUH and MULSH: Unsigned and signed multiplication where the result is only the lower half

I think this nomenclature is a bit misleading. On the Xtensa architecture, instructions with exactly these mnemonics return the most significant 32 bits of the product.

987123879113:
>I think this nomenclature is a bit misleading. On the Xtensa architecture, instructions with exactly these mnemonics return the most significant 32 bits of the product.

Do you have a better suggestion for naming then? I used the H to mean half here since it's only returning half of the result of the operation.

ajrhacker:
I would probably use 'L' (low) as the suffix instead, though that's unfortunately not unambiguous either because ARM uses 'L' for long multiply. Xtensa, for its part, has only one MULL instruction because, like some other RISC architectures, it doesn't bother detecting signed or unsigned overflow.

987123879113:
Playing off using L then: we don't *need* to keep the opcode names as minimal as possible especially since it's just a language used for MAME's DRC, so what about something like MULULOW/MULSLOW instead? I doubt that should clash with any existing arch's naming and it feels like it should be unambiguous as to what the opcodes are doing.

rb6502:
Yours and current fail on my PPC opcode validator on the C backend, but not x64.  Looks purely like flag update issues.

Mismatch: instr=SUBFMEO, src1=0x7fffffff, src2=0x0
expected: dest=0x7fffffff, XER=0xe0000000, CR=0x0
got: dest=0x7fffffff, XER=0x20000000, CR=0x0
opcode is 0x7C6305D0 

Mismatch: instr=SUBFMEO., src1=0x7fffffff, src2=0x0
expected: dest=0x7fffffff, XER=0xe0000000, CR=0x50000000
got: dest=0x7fffffff, XER=0x20000000, CR=0x40000000
opcode is 0x7C6305D1

Mismatch: instr=SUBFZEO, src1=0x7fffffff, src2=0x0
expected: dest=0x80000000, XER=0x0, CR=0x0
got: dest=0x80000000, XER=0xc0000000, CR=0x0
opcode is 0x7C630590

Mismatch: instr=SUBFZEO., src1=0x7fffffff, src2=0x0
expected: dest=0x80000000, XER=0x0, CR=0x80000000
got: dest=0x80000000, XER=0xc0000000, CR=0x90000000
opcode is 0x7C630591

987123879113:
@rb6502 How can I run that validator? I have an idea of what might be happening but I need to verify it first.

cuavas:
> BREAK: Hardware breakpoint that can be caught in a debugger to see exactly what assembly is being generated at any given time. (Warning: will crash the program if used without a debugger attached so use carefully)

You could call `osd_break_into_debugger` for this, which will check for an attached debugger before blowing up on macOS and Windows.

cuavas:
> > I think this nomenclature is a bit misleading. On the Xtensa architecture, instructions with exactly these mnemonics return the most significant 32 bits of the product.
> 
> Do you have a better suggestion for naming then? I used the H to mean half here since it's only returning half of the result of the operation.

The PowerPC instruction mnemonics for these kinds of operations are:
* `mullw` (multiply low word) – Computes the least significant 32 bits of the 64-bit product of two 32-bit integers.
* `mulhw` (multiply high word) – Computes the most significant 32 bits of the 64-bit product of two 32-bit integers.
* `mulhwu` (multiply high word unsigned) – Computes the most significant 32 bits of the 64-bit product of two unsigned 32-bit integers.

It’s possibly just conditioning, but the PowerPC names feel logical for me.

src/devices/cpu/drcbex64.cpp
```diff
@@ -3332,160 +3568,195 @@ void drcbe_x64::op_muls(Assembler &a, const instruction &inst)
...
+
+			a.cmp(rcx, hireg);
+			a.pushfd();
+			a.pop(hireg);
+			a.and_(hireg, 0x40); // zero
+			a.xor_(hireg, 0x40);
+			a.shl(hireg, 5); // turn into overflow flag
+			a.or_(dword_ptr(rsp, 0), hireg);
+
+			a.popfd();
```
cuavas, line 3759: There are potential issues with storing flags on the stack:
* x86-64 ABIs expect the stack pointer to be static and remain 16-byte aligned outside function prologues and epilogues.  Violating this can cause problems with signal handler and interrupts.
* `pushfd` and `popfd` tend to cause pipeline stalls and hurt performance.

Can you use `lahf` and `sahf` instead?

src/devices/cpu/drcbex86.cpp
```diff
@@ -417,12 +421,30 @@ inline void drcbe_x86::emit_combine_z_flags(Assembler &a)
...
-	a.popfd();                                                                          // popf
-	a.lea(esp, ptr(esp, 4));                                                            // lea   esp,[esp+4]
+	a.pushfd();
+
+	a.mov(ecx, dword_ptr(esp, 4)); // zero flag
+	a.or_(ecx, ~0x40);
+	a.and_(dword_ptr(esp, 0), ecx);
+
+	a.popfd();
+	a.lea(esp, ptr(esp, 4));
```
cuavas, line 431: Probably more complex to do given the code is already structured around using the stack, but this would get a performance benefit from avoiding `pushfd`/`popfd`, too.

987123879113:
I rewrote the logic for the C backend's SUBB (32-bit and 64-bit) to match the results of x86/x64. I checked the i386 CPU's SBB32 implementation and that was using a uint64_t to check for the carry flag which makes sense for 32-bit but wouldn't work for the 64-bit version of SUBB, so I ended up just checking for the specific case where x+carry would result in 0 when carry was non-0 (for the x=0, carry=0 case). The results between x86, x64, C backend, and ARM (sbcs) are all agreeing with the results of a few edge cases I tested against, including the above mention 0xffffffff - 0x7fffffff with carry set. If the PPC instruction still fails in the test program then I think it needs to be handled in the frontend.

987123879113:
>You could call osd_break_into_debugger for this, which will check for an attached debugger before blowing up on macOS and Windows.

Never seen that function before. I tested it out and it worked when I had x32dbg/x64dbg attached to mame.exe and did nothing without the debugger attached, so definitely an improvement. Thanks.

src/devices/cpu/drcbex64.cpp
```diff
@@ -1484,7 +1484,7 @@ void drcbe_x64::op_nop(Assembler &a, const instruction &inst)
 
 void drcbe_x64::op_break(Assembler &a, const instruction &inst)
 {
-	a.int3();
+	smart_call_r64(a, (x86code *)(uintptr_t)&osd_break_into_debugger, rax);
```
cuavas, line 1487: Do you need to put a pointer to a string for the function argument in RCX/RDI (depending on the calling convention) for the argument to `osd_break_into_debugger`?

src/devices/cpu/drcbex86.cpp
```diff
@@ -2387,7 +2387,7 @@ void drcbe_x86::op_nop(Assembler &a, const instruction &inst)
 
 void drcbe_x86::op_break(Assembler &a, const instruction &inst)
 {
-	a.int3();
+	a.call(imm(&osd_break_into_debugger));
```
cuavas, line 2390: Do you need to store a pointer to a string at (%esp) for the argument to `osd_break_into_debugger`?

src/devices/cpu/drcbex64.cpp
987123879113: Yeaaah completely forgot it took a parameter. 

I ended up just implementing trampoline functions because these commands will be called very infrequently so performance isn't an issue, and it lets the compiler handle anything that it needs to. For x86 in particular it was doing stack pointer adjustments in the prologue and epilogue instead of just putting it at esp like I implemented initially, so it seems the safer choice.

src/devices/cpu/drcbex64.cpp
```diff
@@ -1482,9 +1591,16 @@ void drcbe_x64::op_nop(Assembler &a, const instruction &inst)
...
+	static const char *message = "break from drc";
+	osd_break_into_debugger(message);
+}
+
 void drcbe_x64::op_break(Assembler &a, const instruction &inst)
 {
-	smart_call_r64(a, (x86code *)(uintptr_t)&osd_break_into_debugger, rax);
+	a.int3();
+	smart_call_r64(a, (x86code *)(uintptr_t)&break_into_debugger, rax);
 }
```
cuavas, line 1604: You shouldn’t need the trampoline function, e.g. see what `op_callc` does:
```c++
void drcbe_x64::op_callc(Assembler &a, const instruction &inst)
{
    // validate instruction
    assert(inst.size() == 4);
    assert_any_condition(inst);
    assert_no_flags(inst);

    // normalize parameters
    const parameter &funcp = inst.param(0);
    assert(funcp.is_c_function());
    be_parameter paramp(*this, inst.param(1), PTYPE_M);

    // skip if conditional
    Label skip = a.newLabel();
    if (inst.condition() != uml::COND_ALWAYS)
        a.short_().j(X86_NOT_CONDITION(inst.condition()), skip);                        // jcc   skip

    // perform the call
    mov_r64_imm(a, Gpq(REG_PARAM1), (uintptr_t)paramp.memory());                        // mov   param1,paramp
    smart_call_r64(a, (x86code *)(uintptr_t)funcp.cfunc(), rax);                        // call  funcp

    // resolve the conditional link
    if (inst.condition() != uml::COND_ALWAYS)
        a.bind(skip);                                                               // skip:
}
```

Also, missing one `const` for the pointer.

You should be able to do something like:
```c++
    static char const *const message = "break from drc";
    mov_r64_imm(a, Gpq(REG_PARAM1), (uintptr_t)message);
    smart_call_r64(a, (x86code *)(uintptr_t)&break_into_debugger, rax);
```

src/devices/cpu/drcbex64.cpp
987123879113: https://github.com/mamedev/mame/pull/13108/commits/25b8ab97d96e07b4569702bf90b072ee4164d5b5

I used r10 and r11 as scratch registers to handle juggling register values. I [double checked](https://learn.microsoft.com/en-us/cpp/build/x64-software-conventions?view=msvc-170#x64-register-usage) and for Windows r10 and r11 considered volatile so I think it should be safe to use them. The header of the x64 DRC also says for Linux/MacOS the r10 and r11 registers are volatile.

x86 uses pushfd/popfd a lot more and there aren't really any readily available scratch registers to make it easier to implement so I don't think it's worth the time to optimize that for x86. Someone else can handle that if they think it's worth the effort though.

src/devices/cpu/drcbex64.cpp
987123879113: I mentioned it in a different comment, but that's what I was doing for x64 but x86 was doing something a bit different so I thought it'd be easier to implement them the same way and leave it up to the compiler.

src/devices/cpu/drcbex86.cpp
```diff
@@ -2385,9 +2385,15 @@ void drcbe_x86::op_nop(Assembler &a, const instruction &inst)
...
+{
+	static const char *message = "break from drc";
+	osd_break_into_debugger(message);
+}
+
 void drcbe_x86::op_break(Assembler &a, const instruction &inst)
 {
-	a.call(imm(&osd_break_into_debugger));
+	a.call(imm(&break_into_debugger));
 }
```
cuavas, line 2397: Same here, it should be possible to do something like:
```c++
    static char const *const message = "break from drc";
    a.mov(dword_ptr(esp, 0), imm(message));
    a.call(imm(&osd_break_into_debugger));
```

src/devices/cpu/drcbex64.cpp
cuavas: Sorry, should be:
```c++
    static char const *const message = "break from drc";
    mov_r64_imm(a, Gpq(REG_PARAM1), (uintptr_t)message);
    smart_call_r64(a, (x86code *)(uintptr_t)&osd_break_into_debugger, rax);
```

src/devices/cpu/drcbex64.cpp
cuavas: The trampoline for x86 needs to do that to keep the stack aligned because it’s a non-leaf function.  The stack will already be aligned if you haven’t done any unbalanced pushes/pops, so you can just place the value at (%esp) in the same way that `op_callc` does.

src/devices/cpu/drcbex64.cpp
987123879113: Ok. https://github.com/mamedev/mame/pull/13108/commits/7da0d8e2bf0ec88ce2422f0bee3f4f52d21c639f

987123879113:
Just to make it more permanent for the PR, RB tested the PPC test again with the new SUBB implementation and it appears to have cleared up all of the failing tests: "ppctest passes 100% now with the C backend, same as it did with x64. Great!"

cuavas:
I don’t have any other objections to the code change itself at this point.

I do agree with @ajrhacker that people familiar with “real” CPU architectures are likely to be confused by using “H” in mnemonics to mean an instruction calculates the least significant part of the result.  In general “real” architectures use “h”/“l” suffixes of some sort for instructions that yield the upper and lower parts of the result of an operation.

987123879113:
I don't really have any particular preference for naming, so I'll change it to `MULULW` and `MULSLW` to keep a more familiar naming while also keeping it tied to the existing MULU and MULS commands in the DRC.

cuavas:
> I don't really have any particular preference for naming, so I'll change it to `MULULW` and `MULSLW` to keep a more familiar naming while also keeping it tied to the existing MULU and MULS commands in the DRC.

Do you actually need two instructions, though?  The least significant half of the full multiplication result should be identical for signed and unsigned multiplication.

987123879113:
>Do you actually need two instructions, though? The least significant half of the full multiplication result should be identical for signed and unsigned multiplication.

The previous implementation wanted to calculate the status flags differently based on whether it was meant to be a 32x32=32 (where the output register for the lower half matched the upper half) or 32x32=64 (when the lower and upper half had different output registers) multiplication. For example, if the top half register wasn't specified in the MULU/MULS opcode then the sign bit would be set if the multiplication resulted in 0x00000001ffffffff and only the 0xffffffff is being returned. When the top half register was specified it would set the sign bit to 0 for the same input values.

I think it warrants splitting off the commands because I think having status flags calculated based on only the lower 32-bits when the upper half's register isn't specified isn't all that explicit in what it's going to do, and that unintuitiveness resulted in the MULS opcode in the C backend being broken for like the last 12 years.

src/devices/cpu/uml.cpp
```diff
@@ -177,9 +177,9 @@ opcode_info const instruction::s_opcode_info_table[OP_MAX] =
...
+	OPINFO3(MULULW,  "!mululw",  4|8, false, NONE, SZV,  ALL,  PINFO(OUT, OP, IRM), PINFO(IN, OP, IANY), PINFO(IN, OP, IANY)) // Unsigned 32x32=32 and 64x64=64 multiplication (overflow set based on 32x32=64 calculation but zero and sign based on 32-bit result)
 	OPINFO4(MULS,    "!muls",    4|8, false, NONE, SZV,  ALL,  PINFO(OUT, OP, IRM), PINFO(OUT, OP, IRM), PINFO(IN, OP, IANY), PINFO(IN, OP, IANY)) // Signed 32x32=64 and 64x64=128 multiplication
-	OPINFO3(MULSH,   "!mulsh",   4|8, false, NONE, SZV,  ALL,  PINFO(OUT, OP, IRM), PINFO(IN, OP, IANY), PINFO(IN, OP, IANY)) // Signed 32x32=32 and 64x64=64 multiplication (overflow set based on 32x32=64 calculation but zero and sign based on 32-bit result)
+	OPINFO3(MULSLW,  "!mulslw",  4|8, false, NONE, SZV,  ALL,  PINFO(OUT, OP, IRM), PINFO(IN, OP, IANY), PINFO(IN, OP, IANY)) // Signed 32x32=32 and 64x64=64 multiplication (overflow set based on 32x32=64 calculation but zero and sign based on 32-bit result)
```
cuavas, line 182: I’m pretty tired at the moment, but the comment doesn’t make sense: “overflow set based on 32x32=64 calculation”.

It shouldn’t be possible for a 32\*32->64 multiply to overflow.  The largest possible values will always fit in a 64-bit result.

For unsigned:
* 0xffff'ffff \* 0xffff'ffff = 0xffff'fffe'0000'0001

For signed:
* 0x7fff'ffff * 0x7fff'ffff = 0x3fff'ffff'0000'0001
* 0x7fff'ffff * 0x8000'0000 = 0xC000'0000'8000'0000
* 0x8000'0000 * 0x8000'0000 = 0x4000'0000'0000'0000

So is “overflow” representing something other than arithmetic overflow?  Because if it represents arithmetic overflow, it will always be zero, and the results will be the same for the signed and unsigned versions of this instruction (which is why e.g. PowerPC only has `mullw`, since results for the lower half are identical for signed/unsigned, while it has `mulhw` and `mulhwu` because the upper half needs different handling).

987123879113: Going based off the x86 documentation because the flags are calculated the same way:
`If the high-order bits of the product are 0, the CF and OF flags are cleared; otherwise, the flags are set.`

cuavas: x86 flags are pure evil.

peterferrie:
This is sign-extending and checking if the result is zero? Why not just sub
instead and not pushf at all?

987123879113:
@peterferrie You're commenting on old code that has since been changed. Did you read the code at all beyond the snippet shown in the email message? You need context to actually understand what the code was trying to do and that snippet didn't give you the required context.

Having said that, I looked over the code again and did find a better way to simplify the logic for that specific part of the code.

peterferrie:
I was commenting on what was in front of me at the time. Now that I see
lahf/sahf being used for flags, I think you need to try that again.

rb6502:
Where are we at on this?  I'd like to see it get in ASAP so we can move on to getting the AArch64 backend in.

src/devices/cpu/drcbec.cpp
```diff
@@ -972,14 +988,27 @@ int drcbe_c::execute(code_handle &entry)
...
+				break;
+
+			case MAKE_OPCODE_SHORT(OP_MULSLW, 4, 1):
+				temp64 = mul_32x32(PARAM1, PARAM2);
+				temp32 = (int32_t)temp64;
+				flags = FLAGS32_NZ(temp32);
+				PARAM0 = temp32;
+				if (temp64 != (int32_t)temp64)
+					flags |= FLAG_V;
+				break;
```
cuavas, line 1010: These are playing fast and loose with undefined corners of integer conversion.  Will it work if you do it like this?

```c++
			case MAKE_OPCODE_SHORT(OP_MULSLW, 4, 0):      // MULSLW   dst,src1,src2[,f]
				temp64 = mul_32x32(PARAM1, PARAM2);
				PARAM0 = (uint32_t)temp64;
				break;

			case MAKE_OPCODE_SHORT(OP_MULSLW, 4, 1):
				temp64 = mul_32x32(PARAM1, PARAM2);
				temp32 = (uint32_t)temp64;
				flags = FLAGS32_NZ(temp32);
				PARAM0 = temp32;
				if ((int64_t)temp64 != (int32_t)temp32)
					flags |= FLAG_V;
				break;
```

src/devices/cpu/drcbec.cpp
```diff
@@ -1579,6 +1619,15 @@ int drcbe_c::execute(code_handle &entry)
...
 				break;
 
+			case MAKE_OPCODE_SHORT(OP_MULULW, 8, 0):      // DMULULW  dst,src1,src2[,f]
+				dmulu(*inst[0].puint64, *inst[0].puint64, DPARAM1, DPARAM2, false);
+				break;
+
+			case MAKE_OPCODE_SHORT(OP_MULULW, 8, 1):
+				flags = dmulu(*inst[0].puint64, *inst[0].puint64, DPARAM1, DPARAM2, true);
+				flags = FLAGS64_NZ(DPARAM0) | (flags & FLAG_V);
+				break;
```
cuavas, line 1629: This would appear to be inconsistent with the 32-bit version in that it doesn’t set V if the result is truncated.

src/devices/cpu/drcbec.cpp
```diff
@@ -1587,6 +1636,15 @@ int drcbe_c::execute(code_handle &entry)
...
 				break;
 
+			case MAKE_OPCODE_SHORT(OP_MULSLW, 8, 0):      // DMULSLW  dst,src1,src2[,f]
+				dmuls(*inst[0].puint64, *inst[0].puint64, DPARAM1, DPARAM2, false);
+				break;
+
+			case MAKE_OPCODE_SHORT(OP_MULSLW, 8, 1):
+				flags = dmuls(*inst[0].puint64, *inst[0].puint64, DPARAM1, DPARAM2, true);
+				flags = FLAGS64_NZ(DPARAM0) | (flags & FLAG_V);
+				break;
```
cuavas, line 1646: Same for this – it’s inconsistent with the 32-bit version.

src/devices/cpu/uml.cpp
```diff
@@ -137,45 +137,49 @@ opcode_info const instruction::s_opcode_info_table[OP_MAX] =
 
 	// Control Flow Operations
 	OPINFO0(NOP,     "nop",      4,   false, NONE, NONE, NONE)
-	OPINFO1(DEBUG,   "debug",    4,   false, NONE, NONE, ALL,  PINFO(IN, OP, IANY))
+	OPINFO1(DEBUG,   "debug",    4,   false, NONE, NONE, ALL,  PINFO(IN, OP, IANY)) // MAME debugger breakpoint
+	OPINFO0(BREAK,   "break",    4,   false, NONE, NONE, NONE) // (for debugging) Hardware breakpoint for debugging the output assembly (will cause MAME to crash without a debugger attached)
```
cuavas, line 141: Comment on the second one is incorrect now – it should be downgraded to “may cause MAME to crash…” as it some platforms check for a debugger.

src/devices/cpu/uml.cpp
```diff
@@ -504,13 +508,26 @@ void uml::instruction::simplify()
...
+			if (m_param[1].is_immediate_value(0) || m_param[2].is_immediate_value(0))
+				convert_to_mov_immediate(0);
+			else if (m_param[1].is_immediate() && m_param[2].is_immediate())
+			{
+				if (m_size == 4)
+					convert_to_mov_immediate(u32(u32(m_param[1].immediate()) * u32(m_param[2].immediate())));
+				else if (m_size == 8)
+					convert_to_mov_immediate(u64(u64(m_param[1].immediate()) * u64(m_param[2].immediate())));
+			}
+			break;
```
cuavas, line 529: Will the code generated by `convert_to_mov_immediate` set the overflow flag appropriately if a condition code update is requested?

src/devices/cpu/uml.cpp
```diff
@@ -520,13 +537,26 @@ void uml::instruction::simplify()
...
+			if (m_param[1].is_immediate_value(0) || m_param[2].is_immediate_value(0))
+				convert_to_mov_immediate(0);
+			else if (m_param[1].is_immediate() && m_param[2].is_immediate())
+			{
+				if (m_size == 4)
+					convert_to_mov_immediate(s32(s32(m_param[1].immediate()) * s32(m_param[2].immediate())));
+				else if (m_size == 8)
+					convert_to_mov_immediate(s64(s64(m_param[1].immediate()) * s64(m_param[2].immediate())));
+			}
+			break;
```
cuavas, line 558: Same here – will the overflow flag be updated correctly by the simplified version?

cuavas (commented):
Had another look, trying to think about corner cases, etc.

src/devices/cpu/drcbec.cpp
```diff
@@ -359,9 +363,10 @@ void drcbe_c::generate(drcuml_block &block, const instruction *instlist, uint32_
 				m_labels.set_codeptr(inst.param(0).label(), (drccodeptr)dst);
 				break;
 
-			// ignore COMMENT and NOP opcodes
+			// ignore COMMENT, NOP, and BREAK opcodes
 			case OP_COMMENT:
 			case OP_NOP:
+			case OP_BREAK:
 				break;
```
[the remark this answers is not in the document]
987123879113: I'm not against it but for the C backend I don't think you can use a debugger to get anything meaningful so I don't think it matters compared to something generating actual assembly where you want to be able to debug the generated assembly. Can you think of a usecase where you'd want to break into a debugger for the C backend?

src/devices/cpu/drcbec.cpp
987123879113: FLAG_V is the overflow flag that I mentioned before, where it sets overflow/FLAG_V if the upper half of the result isn't zero. `dmulu` sets FLAG_V using `((hi != 0) << 1)` so I think it's the same.

Can you think of a specific case where you think it'd fail?

src/devices/cpu/uml.cpp
987123879113: The simplify function in uml.cpp only works for opcodes that won't update status flags, so this code is fine.

See: https://github.com/mamedev/mame/blob/master/src/devices/cpu/uml.cpp#L370-L374

src/devices/cpu/uml.cpp
987123879113: Same as above comment so I'll resolve this comment and leave the other for discussion.

src/devices/cpu/drcbec.cpp
987123879113: From a quick test, yeah this works fine. All of my test cases pass with the change. I'll change it.

987123879113:
>Where are we at on this? I'd like to see it get in ASAP so we can move on to getting the AArch64 backend in.

The aarch64 backend isn't really ready quite yet. I'm still writing test cases, this time for mapvar usages. You already responded in the shoutbox so you're aware now but it's very likely that the mapvar usages in the C backend are broken so I need to fix that for this PR (+ fix x86 and x64 backends if needed) before I can move on to other things.

src/devices/cpu/drcbec.cpp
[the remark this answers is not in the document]
cuavas: It still might be useful for examining the state of the emulator when particular generated code is hit. It's considered generally useful enough that we have a convenient function for it.

src/devices/cpu/drcbec.cpp
cuavas: Oh sorry, I forgot `dmulu` was already a DRC thing that set flags and not just a multiplication.

src/devices/cpu/drcbec.cpp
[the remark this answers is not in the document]
987123879113: Ok. I don't think there's any reason *not* to implement it for the C backend so I'll add it.

cuavas:
> > Where are we at on this? I'd like to see it get in ASAP so we can move on to getting the AArch64 backend in.
> 
> The aarch64 backend isn't really ready quite yet. I'm still writing test cases, this time for mapvar usages. You already responded in the shoutbox so you're aware now but it's very likely that the mapvar usages in the C backend are broken so I need to fix that for this PR (+ fix x86 and x64 backends if needed) before I can move on to other things.

But if we can get this PR to a state where we're confident it isn't making things worse, can we merge it to get a new baseline for testing and let additional fixes wait for another PR?

987123879113:
>But if we can get this PR to a state where we're confident it isn't making things worse, can we merge it to get a new baseline for testing and let additional fixes wait for another PR?

Next release is at the end of January right? The more testing before next release the better so I'm fine with that. It can always be reverted before next release if needed.

rb6502:
Right.  My suspicion is that regressions caused by this are most likely to be errors in the actual UML generation by the CPU cores, but obviously the best way to find regressions is going to be to get this landed.

src/devices/cpu/drcbec.cpp
```diff
@@ -2232,8 +2231,7 @@ void drcbe_c::output_parameter(drcbec_instruction **dstptr, void **immedptr, int
 
 		// convert mapvars to immediates
 		case parameter::PTYPE_MAPVAR:
-			temp_param = m_map.get_last_value(param.mapvar());
-			return output_parameter(dstptr, immedptr, size, temp_param);
+			return output_parameter(dstptr, immedptr, size, param.mapvar());
```
987123879113, line 2234: drcuml.cpp already has an optimization step which replaces mapvar references with constants when possible, except if the opcode is OP_RECOVER which requires the mapvar register ID. The other backends rely on that to get mapvar values. 

The C backend had an extra step where it'd try to get mapvar values during generation which would result in OP_RECOVER breaking because instead of getting the mapvar register ID, it'd get some other value. For example, if `uml::M1` was set to 1234 and OP_RECOVER was called for `uml::M1`, it'd try looking up the value from `uml::M0 + 1234` instead of `uml::M0 + 1`) if `uml::M1` was set within the same block, or if you're in a sub call somewhere it would fail to get the value of `uml::M1` and return 0 so it'd try to get the value from `uml::M0 + 0` which would also give a different value than you want.

https://github.com/mamedev/mame/blob/master/src/devices/cpu/drcuml.cpp#L421-L429

987123879113:
@cuavas @rb6502 I think the above C backend mapvar fix is probably the last major issue I'm going to find in UML and the existing backends, so feel free to merge whenever you're ready.

987123879113:
I found another potential issue but I'm not really sure what the "correct" way to handle it is.

```
// value = 0x1111111187654321
// value2 = 0x4444444412345678

UML_DMOV(block, uml::I0, 0x2222222212348765);

UML_DAND(block, mem(&m_state->value), mem(&m_state->value), 0xffffffff00000000U);
UML_AND(block, uml::I0, mem(&m_state->value2), 0xffffffff);
UML_DOR(block, mem(&m_state->value), mem(&m_state->value), uml::I0);
```

The C backend and x86 backend does not clear out the upper half of the 64-bit register if a 32-bit operation is executed, so if you use `UML_AND` then it only writes to the lower 4 bytes of the I0 register, and if you use `UML_DAND` it'll write the entire 8 bytes. On the C backend and x86 backend the above code results in `0x3333333312345678` and on the x64 backend it results in `0x1111111112345678`

I found a place in sharcdrc.cpp that does pretty much the same thing so it could in theory be happening already, depending on what is in I0:
```
case 0x00:  // MR0F
	UML_DAND(block, MRF, MRF, 0xffffffff00000000U);
	UML_AND(block, I0, REG(rn), 0xffffffff);
	UML_DOR(block, MRF, MRF, I0);
	break;
```

987123879113:
Turns out trying to get the C backend to clear the upper bits of a register would require quite a bit of rewriting to the point it's probably just better to leave that particular thing undefined and just have the frontend make sure it's using the proper width commands for what it wants to do.

cuavas:
Is that possibly why stuff like the Hyperstone DRC has `#ifndef PTR64` around explicitly isolating the low 32 bits of a register?  For example here:

https://github.com/mamedev/mame/blob/master/src/devices/cpu/e132xs/e132xsdrc_ops.hxx#L633

There are a bunch of cases of that happening, often following a load.  Here’s one of them:
```c++
			UML_LOAD(block, I0, (void *)m_core->global_regs, src_code, SIZE_DWORD, SCALE_x4);
			UML_AND(block, I1, DRC_SR, 1);
#ifndef PTR64
			UML_DAND(block, I0, I0, 0x00000000ffffffffULL);
			UML_DAND(block, I1, I1, 0x00000000ffffffffULL);
#endif
			UML_DADD(block, I0, I0, I1);
```

Note that it does a 32-bit load into I0 before zeroing the upper half with `UML_DAND` if it isn’t a 64-bit target.  Is this trying to work around the difference in behaviour between the i686 and x86-64 backends?  If it is, it’s broken for the C backend which apparently behaves like the i686 backend.  It’s also a pretty bad situation if the front-ends need to know about backend quirks.

My thoughts regarding this:
* A 32-bit load leaving the upper 32 bits of the register undefined seems like very bad behaviour to me.  No real-world CPU behaves like this.  As you’re no doubt aware, PowerPC, Itanium, x86-64 and AArch64 zero the upper 32 bits.  Alpha AXP and RISC-V sign extend the value into the upper 32 bits.  (Yes, some architectures do give you the option to use the other alternative, but others don’t – PowerPC can sign-extend on load, but Itanium can’t sign-extend on load and Alpha AXP can’t zero-extend on load.)
* I’d prefer if at the very least, 32-bit loads were defined as clearing the upper 32 bits of the register.
* I don’t like front-ends depending on backend quirks.  I’d honestly rather take a performance hit than have stuff like that waiting to rear its head.
* Is the x86-64 backend smart enough to recognise contiguous bit patterns in constant arguments to bitwise AND that can be optimised?  For example converting `x & 0x0000ff` or `x & 0x00ffff` to a `movzx` instruction?

At any rate, cleaning up that mess isn’t in scope for this PR.

987123879113:
Yeah, that ifndef stuff looks exactly like what I'd expect a fix for these issues to be.

For the C frontend specifically: there's a union with a bunch of pointers (https://github.com/mamedev/mame/blob/master/src/devices/cpu/drcbec.cpp#L209-L234). If there was a way to tell apart a PTYPE_INT_REGISTER vs some other type like PTYPE_MEMORY then it would be easier to fix but currently you can't tell what the parameter type was at the time you write the result value. A quick hacky fix could be something like clearing the upper bits if `output_parameter` is called with for a PTYPE_INT_REGISTER with a size of 4 but there's no way to tell if it's an output parameter or input parameter so that has its own issues. So all of the quick options I can think of aren't very good.

987123879113:
https://github.com/mamedev/mame/blob/master/src/devices/cpu/drcbex86.cpp#L1021-L1022

x86 could probably largely be fixed by adding a mov outside of the if statement there.
`a.mov(MABS(m_reghi[param.ireg()], 4), 0);`

rb6502:
Isn't it defined that UML_LOAD zero-extends and UML_LOADS sign-extends?  I'll note that none of the DRC frontends actually use UML_LOADS, but looking at the x64 definitions of those operations, that seems like the intent.

rb6502:
Actually never mind on that, LOAD/LOADS are purely for register/cache access.  And in that case, while I suppose it's possible, I can't really think of a case where you'd want to sign extend an access like that.

PR #13112, opened 2024-12-23
Fix 2 Visual Studio Warnings

Meerkov:
Two are for implicit casts (which the visual studio build elevates to Errors) and one is for an unused variable.

Since these break compilation from a fresh build in default conditions, and the project is set to elevate warnings to errors, these likely should be fixed.

### DIFF
--- a/3rdparty/lsqlite3/lsqlite3.c
+++ b/3rdparty/lsqlite3/lsqlite3.c
@@ -1317,7 +1317,6 @@ static void db_update_hook_callback(void *user, int op, char const *dbname, char
     sdb *db = (sdb*)user;
     lua_State *L = db->L;
     int top = lua_gettop(L);
-    lua_Number n;
 
     /* setup lua callback call */
     lua_rawgeti(L, LUA_REGISTRYINDEX, db->update_hook_cb);    /* get callback */
--- a/3rdparty/lsqlite3/lsqlite3.c
+++ b/3rdparty/lsqlite3/lsqlite3.c
@@ -1317,6 +1317,7 @@ static void db_update_hook_callback(void *user, int op, char const *dbname, char
     sdb *db = (sdb*)user;
     lua_State *L = db->L;
     int top = lua_gettop(L);
+    lua_Number n;
 
     /* setup lua callback call */
     lua_rawgeti(L, LUA_REGISTRYINDEX, db->update_hook_cb);    /* get callback */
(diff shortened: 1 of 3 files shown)


3rdparty/lsqlite3/lsqlite3.c
```diff
@@ -1317,7 +1317,6 @@ static void db_update_hook_callback(void *user, int op, char const *dbname, char
     sdb *db = (sdb*)user;
     lua_State *L = db->L;
     int top = lua_gettop(L);
-    lua_Number n;
```
cuavas, line 1320: I don’t want to diverge from upstream for something trivial like this.  I’ve already had to patch it to build cleanly as C++, I want to keep the patches minimal.

src/mame/nec/pc88va.cpp
```diff
@@ -256,7 +256,7 @@ void pc88va_state::sys_port5_w(u8 data)
 
 u8 pc88va_state::sys_port5_r()
 {
-	return m_rstmd | 8;
+	return (u8)m_rstmd | 8;
```
cuavas, line 259: Better to use `(m_rstmd ? 1 : 0) | 8` to explicitly map `bool` to integer values.

src/mame/moog/source.cpp
```diff
@@ -408,7 +408,7 @@ float source_state::get_keyboard_v() const
 
 	// *** Convert pressed key to a voltage.
 
-	static constexpr const float KEYBOARD_VREF = 8.24;  // From schematic.
+	static constexpr const float KEYBOARD_VREF = 8.24f;  // From schematic.
```
cuavas, line 411: Does MSVC really require literal suffixes for constant initialisation of `float` variables?  That doesn’t strike me as useful.

cuavas:
We don’t support building with Visual Studio without `NOWERROR=1` – MSVC generates too many spurious warnings.

src/mame/moog/source.cpp
Meerkov: It treats it as a cast from double to float, (loss of precision).

Meerkov (commented):
I'll undo lsqlite3 and change to a ternary.

pmackinlay:
All of these are better addressed by adjusting the MSVC warning settings and not the code.

happppp:
If I ignore the context of trying to work around compiler warnings:
- f prefix for const float, ok
- don't do a binary OR for a boolean, good

PR #6993, opened 2020-07-23
Add new option beam_dot_size

aaronsgiles:
Add new option beam_dot_size that controls the rendered size of 'dots' in vector games. [Aaron Giles]

PR #13116, opened 2024-12-24
vector_device: render dots as fuzzy circles.

pmarks-net:
When emulating games like Asteroids with increased -beam_dot_size, MAME's various renderers draw the dots differently:

rendersw (gdi and libretro) draws a vertical line
bgfx draws a diamond
d3d draws a diamond
opengl draws a circle (reasonable)

This change replaces add_line() with add_quad() and a fuzzy circle texture, so that dots look the same on all renderers.

Also increase the brightness of dots by 2X (capped to 255) because I once played Asteroids on a real cabinet and found the dots surprisingly bright. The Asteroids high score screen uses intensity 112 for the words and 64 for the dots, which suggests that a factor of 2 is approximately correct.

Related to Issue 399


pmarks-net:
Screenshots before and after this patch, with `mame asteroid -beam_dot_size 2 -video gdi`:

asteroids_swrender_unpatched

asteroids_swrender_new_dot

mamehaze:
I'm not entirely convinced this is the correct way, the after shot looks kinda terrible IMHO

pmarks-net:
> the after shot looks kinda terrible IMHO

Do you agree that the dots should have circular symmetry?  I can use this technique to draw circles of any size/sharpness, so I'm open to suggestions for fine tuning.

Note that it's possible to apply shader effects using rendersw->RetroArch or bgfx->vkBasalt, so vector_device just needs to do something reasonable, not look perfectly realistic.

mamehaze:
Well they should likely be a point, but adding a glow to them like this just makes it look like you've thrown a bitmap in place of them with artificial glow effect when nothing else is being rendered that way.

pmarks-net:
Here's a screenshot with plain non-glowing circles:
asteroids_swrender_plain_circles

mamehaze:
the smaller dots look like a more reasonable approximation without the out of place 'fuzzy glow' but ultimately it won't be my decision that determines if people think this is the way forward

pmarks-net:
I fixed the problem.  Here's a before/after screenshot of `mame tempest -beam_dot_size 1 -video gdi`:

tempest_swrender_unpatched
tempest_swrender_new_dot

pmarks-net:
Mentioning @aaronsgiles and @rb6502, because this code was last touched in https://github.com/mamedev/mame/pull/6993.

MooglyGuy:
GDI is a last-resort renderer anyway and should only be reliably used for ground truth on raster games.

Ideally, BGFX and D3D should be fixed instead, but as my time to look at the renderers lately has been scarce, the small-dot version of this gets the green light from me as a visually-pleasant stopgap.

pmarks-net:
> GDI is a last-resort renderer anyway and should only be reliably used for ground truth on raster games. Ideally, BGFX and D3D should be fixed instead

I've been demonstrating with GDI because it's the worst, and lots of people are using the software renderer via RetroArch, but this patch also improves BGFX and D3D.

before/after screenshots of `mame asteroid -beam_dot_size 2 -video bgfx`:

asteroids_bgfx_unpatched
asteroids_bgfx_new_dot

pmarks-net:
Is there someone specific I should ask to get this merged?  I'm happy to make changes if requested.

PR #13117, opened 2024-12-24
namco/namcos2.cpp: Added PORT_MINMAX to fix steering left range to 127

UnclePetros:
Resolve Bug ID 08690: https://mametesters.org/view.php?id=8690

- In the current version of dirtfoxj, steering range is (128,127). 
The fix adjust the range to (127,127)


src/mame/namco/namcos2.cpp
Gurudumps, line 1: This isn't really doing anything and the mid-point is likely to be incorrect.
The middle value should be 0x80 (decimal 128) since it's mid-way between 0 and 255 (256 total range). The minmax thing does nothing. It's probably technically accurate to limit the range but the game only goes to +-128 even without it.
In any case without a manual and accurate info about the real steering control mechanism there's no way to know what the real min-max values are and as far as I can see there's no manual available for this game. If someone can provide the info from the manual it'll help more than simply guessing the values and changing them by a pointless 0x01 as was done here.
The game isn't playable (or at least not easily playable) in its current state unless you use a real spinner or a paddle which are not common PC controllers. An Xbox360 analog stick doesn't help.
I suppose this 'fix' can go in with correct 0x80 value and minmax set to 0x00,0xff but it's still not playable as-is and really doesn't help anything.

UnclePetros: This fix aim to just equal left steering and right steering.
Otherwise you have 128 Left steering and 127 Right steering.
I know it's a pointless 0x01, but I think the correct behaviour is to have the same steering left and right (unless proven otherwise).
If you use all 256 values you can't have an equal steering left and right, because of the central 0, that is shared by left and right.
So you have to limit the values to 255 (0 to 0xfe) and choose 0x7f as central value.

happppp: No, the central value of 0 is very likely correct.

UnclePetros: > No, the central value of 0 (0x80) is very likely correct.

Are you sure?
It is like a 1-byte signed integer range (-128,127).
Obviously 0 is not the midpoint.
Using 0x80 is impossible to have equal steering left and right.
Anyway, nothing really change in the game. Changing value by 0x01 is pointless (as @Gurudumps noticed).
So no problem at all.

happppp: Yes, see service mode, when it's 0x7f it says it's moving to the left.
It's probably a free spinning wheel, that MAME does not emulate. <0x80 = spinning left, 0x80 = not moving, >0x80 = spinning right.

UnclePetros: > Yes, see service mode, when it's 0x7f it says it's moving to the left. It's probably a free spinning wheel, that MAME does not emulate. <0x80 = spinning left, 0x80 = not moving, >0x80 = spinning right.

Where do you see this?
I have this screen:
image
It is the same for both 0x80 and 0x7f.
The only difference is on the left steering limit. 128 for current version. 127 with the fix.

happppp: Exactly there, and if I copy-paste your change, the range becomes: 128-001-126.
Maybe you accidentally recalibrated the controls?

UnclePetros: > Exactly there, and if I copy-paste your change, the range becomes: 128-001-126. Maybe you accidentally recalibrated the controls?

I never experience this.
Whatever value I put as central value, I always see 000 in service mode adjustement control screen.
What changes are only the bounds of the steering range.

Gurudumps: Of course if you recalibrate it the value will always be 0.
The 'Adjustment Of Control' screen is only for use when changing the wheel mechanism on a real machine or fixing the movement if the central (no-movement) value is not zero on the normal test screen. For MAME you don't need to recalibrate it.
A value of 127 left and 128 right (or some other numbers) does not make any difference at all to the gameplay so changing it does nothing.
Find the real cabinet control info (buy the manual etc), pass that info onto the MAME team (ideally scan and upload to archive.org) then it will be possible to fix the game controls properly. Until then there's no point messing with it.

UnclePetros: Ok, clear.
But I don't understand why thia bug is open and confirmed on mametesters.
Please close it: https://mametesters.org/view.php?id=8690

Gurudumps: Not everyone who makes mametester issues actually understands why things are so ;-)
The adjustment screen should never be anything other than 000 for all 3 values. That is not where the movement range is actually checked.
The idea is the controls are untouched and centered (autmatically with a spring) then the adjustment screen is activated which makes those values 0 then the game uses that as a zero reference. Look at Suzuka 8 Hours 2,... same type of adjustment is available in that game too, and many other Namco games with analog controls.
Check the manual of Suzuka or Dirt Dash to see how that adjustment screen works.
To fix this insignificant and pointless 'bug' all that needs to be done is set the range to 0x00,0xff with the center value untouched (0x80).
It's also highly likely the game uses a normal 5K potentiometer as it doesn't need much movement based on the video of the real cabinet. It certainly doesn't need any more movement than what a pot provides which is 270 degrees. 

Additionally Namco System 2 doesn't have any chip on the board that could decode a mouse-like segmented wheel encoder for steering so it has to be using a normal 270 degrees pot for the wheel like all the other System 2 driving games.

UnclePetros:
Someone should close insignificant bug, not confirm them.
Eventually explaining why they are pointless and should be closed.
There are bugs from 2016 and before, all confirmed.
I'm a novice contributer to Mame.
So I begin with DIP Switches bugs.
But many of these ones are very misleading.

Inviato da Outlook per Android<https://aka.ms/AAb9ysg>
________________________________

src/mame/namco/namcos2.cpp
UnclePetros: Someone should close insignificant bug, not confirm them.
Eventually explaining why they are pointless and should be closed.
There are bugs from 2016 and before, all confirmed.
I'm a novice contributer to Mame.
So I begin with DIP Switches bugs.
But many of these ones are very misleading.

Gurudumps: You should close this PR since highly likely it will not be processed further.

happppp:
I reduced brake pedal range here https://github.com/mamedev/mame/commit/5d13ae79d7c72681d2746887fcab3a34e089000f
And while I was at it, also made steering wheel left/right range the same.

UnclePetros:
> I reduced brake pedal range here [5d13ae7](https://github.com/mamedev/mame/commit/5d13ae79d7c72681d2746887fcab3a34e089000f) And while I was at it, also made steering wheel left/right range the same.

Ok. It was just the other solution I was posting in order to mantain 0x80 as midpoint. Fine.

PR #13122, opened 2024-12-25
spg format + [tsconf.xml] added 28 items

holub:


holub:
Knock... knock

holub:
@cuavas is it good time to finish with the leftovers?

holub:
Hello, I'm PR #13122. I'm 8 months old... well, technically 2y3m old because I'm reincarnation of #11274. Can anybody take care of me?

hash/tsconf.xml
```diff
@@ -0,0 +1,456 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="tsconf" description="TS-Configuration (All Formats)">
```
angelosa, line 6: Wouldn't `TS-Configuration snapshots` be a better description for this?

angelosa:
These lacks whatsnew credits if they are new additions.

hash/tsconf.xml
holub: I merged this with existing tr-dos list, so we have both in same list now.
The thing is, tsconf doesn't have many prods and it doesn't change too much over the years. Using multiply lists doesn't make sense imho.

angelosa: If we want to pursue this road then at least separate media types as having their own block, so to make future insertions less prone to errors.
```xml
<!-- !Snapshots -->

	<software name="aaaa">
[...]
	</software>

<!-- !Floppies -->
	<software name="abbb">
[...]
	</software>
```

holub:
knock-knock?

src/mame/sinclair/spectrum.h
```diff
@@ -96,6 +96,7 @@ class spectrum_state : public driver_device
 	virtual void video_start() override ATTR_COLD;
 
 	// until machine/spec_snqk.cpp gets somehow disentangled
+	virtual void bank3_set_page(u8 page) { }
```
angelosa, line 99: Sounds like a `TODO`?

angelosa (approved):

cuavas:
This still has issues with copyright, doesn’t it?  It pulls code from elsewhere into an existing source file.

src/mame/sinclair/spectrum.h
holub: It's painful heritage. Not sure if anybody brave enough to change this

holub: The problem which originally driven this... intention to make snapshot loadable by any of clones. Any of methods in this batch are not supported in 48K machine clearly. Hence making it work as intended without issues is practically impossible.

holub:
> This still has issues with copyright, doesn’t it? It pulls code from elsewhere into an existing source file.

I don't think so. After further investigation I discovered that any emulators which support hrust compression (e.g. ZESarUX, Unreal) just copied this from original hrust's cpp examples which were released more than 10 years ago.

cuavas:
> > This still has issues with copyright, doesn’t it? It pulls code from elsewhere into an existing source file.
> 
> I don't think so. After further investigation I discovered that any emulators which support hrust compression just copied this from original hrust's cpp examples which were released more than 10 years ago.

But that still creates a copyright issue, doesn’t it?  The example code is substantial in this case and has some implicit license (all rights reserved if not otherwise specified), and this pulls it into a source file, which then complicates copyright on that source file.

holub:
This is one of original authors' repo: https://github.com/psbhlw/hrust13-hrum35-pc/blob/master/dehrust13/dehrust13.cpp
I'll try to find if he still active and receive his approval.
or... is it possible just to add him to our copyright header?

cuavas:
> This is one of original authors' repo: https://github.com/psbhlw/hrust13-hrum35-pc/blob/master/dehrust13/dehrust13.cpp I'll try to find if he still active and receive his approval. or... is it possible just to add him to our copyright header?

Well, looking at that repo, it has “Hrumer, Nikita Burnashev, HalfElf, psb.” in the readme file, and no license specified, which means we *can’t* use that code legally – it’s implicitly “all rights reserved”.

If they’re prepared to provide it under an acceptable license, I’d rather have it confined to a separate source file to avoid making the copyright situation for sinclair/spec_snqk.cpp being made murkier.

If they can’t/won’t provide the code under an acceptable license, it needs a clean reimplementation (“Chinese wall”, etc.).

holub:
What would be appropriate license example which can not cause the issue for us?

holub:
I was talking to psb, he told that HalfElf is original author of dehrust plugin.
I received verbal consent for usage it under BSD-3 in MAME from HalfElf

he1f:
Hi, I'm HalfElf and as original author of dehrust sourses I'm granting permission to use them under BSD-3 license.

holub:
@cuavas here you go. I also added HalfElf as a co-holder, what else is missing?

holub:
Shall we?

he1f:
as you wish :)

cuavas:
Let me look at this again…

cuavas:
OK, can we please not use this “class split across multiple files and empty header voodoo” approach?  Can you come up with a way to encapsulate the hrust stuff into a class or function that doesn’t need to directly be part of the `spectrum_state`?

What does it actually need from `spectrum_state`?  How can we make a clean interface to it?

Also, where does `mlz_decompress_block` come from?  Is that from some other source, or is that a clean implementation of the format?

he1f:
I can share my original sources if it helps. What is the prefered way to share? Actually they are located at [theX.unterdrund.net/look](https://thex.untergrund.net/look/)

holub:
mlz: https://web.archive.org/web/20080224212635/http://lvd.nm.ru/MegaLZ/

holub:
> OK, can we please not use this “class split across multiple files and empty header voodoo” approach? Can you come up with a way to encapsulate the hrust stuff into a class or function that doesn’t need to directly be part of the `spectrum_state`?
> 
> What does it actually need from `spectrum_state`? How can we make a clean interface to it?
> 
> Also, where does `mlz_decompress_block` come from? Is that from some other source, or is that a clean implementation of the format?

I refactored per your suggestion and dropped MegaLZ support.

PR #13123, opened 2024-12-26
MT08472: heap overflow

holub:
Can't reproduce but have high confidence that should fix the issue


firewave:
It does not fix the issue.

The failure happens on the access of `m_palette_data` which only contains a single entry but `get_border_color()` will obviously return values bigger than 0...

holub:
> It does not fix the issue.
> 
> The failure happens on the access of `m_palette_data` which only contains a single entry but `get_border_color()` will obviously return values bigger than 0...

Great. Could you check this one?

firewave:
> Great. Could you check this one?

That fixes it.

PR #13125, opened 2024-12-27
Remove redundant children device reset for most drivers

FlykeSpice:
This removes the manual children device reset that was happening under ``machine_reset()`` for most drivers, it's redundant since all the children devices are automatically resetted when the user invokes the reset command before the call to``machine_reset()``

There were some drivers I didn't debilatery touch such as the commodore machines because they reuse the machine_reset function to manually reset [selected] children devices when the RESET input port is changed and I didn't want to break functionality.

### DIFF
--- a/src/mame/act/victor9k.cpp
+++ b/src/mame/act/victor9k.cpp
@@ -149,7 +149,6 @@ private:
 	required_shared_ptr<uint8_t> m_video_ram;
 
 	virtual void machine_start() override ATTR_COLD;
-	virtual void machine_reset() override ATTR_COLD;
 
 	void via1_pa_w(uint8_t data);
 	void write_nfrd(int state);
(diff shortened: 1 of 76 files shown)


happppp:
There are some useless cases yes, victor9k looks like a driver author misunderstanding for example. But without thoroughly testing everything, you won't know for sure.

Reasons a driver deliberately resets a device:
- reinitialize something and reset device again to make it catch the changes
- race condition: reset order is device a,b,c,d, but you want device b to reset after device c, so reset it again later
- cancel the effects of an event(timer trigger) that happened between device resets and machine_resets, like a vblank rising edge (that realistically a device wouldn't detect in mid-reset)

Robbbert:
The drivers written by me only have a reset when it's needed, and removing them will break things. I'm guessing that extensive testing hasn't been done, and I'd suggest this PR be rejected.

FlykeSpice:
My bad, I didn't account for this scenario that some drivers may actually need to manually reset their children device under ``machine_reset`` after setting properties to apply changes. I thought it was some dodgy code.

PR #13127, opened 2024-12-27
CD-i: Fix DYUV Format

Vincent-Halver:
This change fixes the following issues with the DYUV format:

1. The DYUV format incorrectly was too dark by 16/256. This fix corrects a previously incorrect attempt to fix that issue.
2. The DYUV format does not allow Mosaic features.

To Validate this change:
1. Use the Validation CD (Europe).
2. Click "Philips Tests"
3. Click "Video Representations"
4. Click "DYUV Decoder"

Result:
The DYUV format is slightly brighter. Mosaic effects will work more correctly. Unfortunately I don't know of a good example of how to see the mosaic behavior difference, however the spec makes it clear this configuration is not allowed.

Uncertain Issues:
The previous midpoint interpolation of 129+129 would be 1. The spec agrees with that formula. As the expected behavior is not yet well understood, a modified algorithm is used to produce more sensible midpoints that do not have this rollover artifact.

Known Issues:
There are further transparency issues that are out of scope for this update.

RGB format fix is not included in this update. Previous discussion about color/noise can be safely ignored for now, and discussed in the RGB update later.

There are further format issues.

Updated Test Images:
image
image

Compare to previous broken decoder:
image
image

### DIFF
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -724,7 +724,7 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
 
 						uint32_t entry = (limit_rgb[m_dyuv_v_to_r[v0]] << 16) | (limit_rgb[m_dyuv_u_to_g[u0] + m_dyuv_v_to_g[v0]] << 8) | limit_rgb[m_dyuv_u_to_b[u0]];
 						pixels[x] = entry;
-						transparent[x] = (transp_always || (use_region_flag && region_flags[x])) != invert_transp_condition;
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
 
 						x++;
 
@@ -732,7 +732,7 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
 
 						entry = (limit_rgb[m_dyuv_v_to_r[v1]] << 16) | (limit_rgb[m_dyuv_u_to_g[u1] + m_dyuv_v_to_g[v1]] << 8) | limit_rgb[m_dyuv_u_to_b[u1]];
 						pixels[x] = entry;
-						transparent[x] = (transp_always || (use_region_flag && region_flags[x])) != invert_transp_condition;
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
 
 						byte = data[(vsr++ & 0x0007ffff) ^ 1];
 
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -724,7 +724,7 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
 
 						uint32_t entry = (limit_rgb[m_dyuv_v_to_r[v0]] << 16) | (limit_rgb[m_dyuv_u_to_g[u0] + m_dyuv_v_to_g[v0]] << 8) | limit_rgb[m_dyuv_u_to_b[u0]];
 						pixels[x] = entry;
-						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x << 1])) != invert_transp_condition;
 
 						x++;
 
@@ -732,7 +732,7 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
 
 						entry = (limit_rgb[m_dyuv_v_to_r[v1]] << 16) | (limit_rgb[m_dyuv_u_to_g[u1] + m_dyuv_v_to_g[v1]] << 8) | limit_rgb[m_dyuv_u_to_b[u1]];
 						pixels[x] = entry;
-						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x << 1])) != invert_transp_condition;
 
 						byte = data[(vsr++ & 0x0007ffff) ^ 1];
 
@@ -1408,7 +1408,7 @@ void mcd212_device::device_start()
 
 	for (uint16_t w = 0; w < 0x300; w++)
 	{
-		const uint8_t limit = (w < 0x100) ? 0 : w < 0x200 ? w - 0x100 : 0xff;
+		const uint8_t limit = (w < 0x100) ? 0 : (w < 0x200) ? (w - 0x100) : 0xff;
 		m_dyuv_limit_rgb_lut[w] = limit;
 	}
 
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -661,9 +661,9 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
 	const uint8_t mosaic_factor = get_mosaic_factor<Channel>();
 
 	const uint32_t dyuv_abs_start = m_dyuv_abs_start[Channel];
-	const uint8_t start_y = (dyuv_abs_start >> 16) & 0x000000ff;
-	const uint8_t start_u = (dyuv_abs_start >>  8) & 0x000000ff;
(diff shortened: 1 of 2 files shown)


src/mame/philips/mcd212.h
```diff
@@ -171,7 +171,7 @@ class mcd212_device : public device_t,
 
 		ICM_OFF              = 0x0,
 		ICM_CLUT8            = 0x1,
-		ICM_RGB555           = 0x2,
+		ICM_RGB555           = 0x1,
```
angelosa, line 174: This change looks unintentional, and I presume it will break `magicard` Impera logo at boot.

Vincent-Halver: While this change wasn't intended to be included in this change, it is necessary for the RGB format to work correctly, which will be added in a future change.

angelosa: Well, you can touch it on your next PR then :)

cuavas: But it clashes with `ICM_CLUT8` now.

src/mame/philips/mcd212.cpp
```diff
@@ -719,48 +716,23 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
...
-						uint32_t *limit_r = m_dyuv_limit_r_lut + y0 + 0xff;
-						uint32_t *limit_g = m_dyuv_limit_g_lut + y0 + 0xff;
-						uint32_t *limit_b = m_dyuv_limit_b_lut + y0 + 0xff;
+						uint32_t* limit_rgb = m_dyuv_limit_rgb_lut + y0 + 0x100;
 
-						uint32_t entry = limit_r[m_dyuv_v_to_r[v0]] | limit_g[m_dyuv_u_to_g[u0] + m_dyuv_v_to_g[v0]] | limit_b[m_dyuv_u_to_b[u0]];
+						uint32_t entry = (limit_rgb[m_dyuv_v_to_r[v0]] << 16) | (limit_rgb[m_dyuv_u_to_g[u0] + m_dyuv_v_to_g[v0]] << 8) | limit_rgb[m_dyuv_u_to_b[u0]];
 						pixels[x] = entry;
-						transparent[x] = (transp_always || (use_region_flag && region_flags[x << 1])) != invert_transp_condition;
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
```
cuavas, line 727: Please don’t take away spaces around binary operators unnecessarily.

src/mame/philips/mcd212.cpp
```diff
@@ -719,48 +716,23 @@ void mcd212_device::process_vsr(uint32_t *pixels, bool *transparent)
...
-						if (mosaic_enable)
-						{
-							for (int mosaic_index = 1; mosaic_index < mosaic_factor && (x + mosaic_index) < width; mosaic_index++)
-							{
-								pixels[x + mosaic_index] = pixels[x];
-								transparent[x + mosaic_index] = transparent[x];
-							}
-							x += mosaic_factor - 1;
-						}
+						transparent[x] = (transp_always || (use_region_flag && region_flags[x<<1])) != invert_transp_condition;
```
cuavas, line 735: Same here, please put the spaces around the shift operator back.

src/mame/philips/mcd212.cpp
```diff
@@ -1434,12 +1406,10 @@ void mcd212_device::device_start()
...
 
-	for (uint16_t w = 0; w < 3 * 0xff; w++)
+	for (uint16_t w = 0; w < 0x300; w++)
 	{
-		const uint8_t limit = (w < 0xff + 16) ?  0 : w <= 16 + 2 * 0xff ? w - 0x10f : 0xff;
-		m_dyuv_limit_r_lut[w] = limit << 16;
-		m_dyuv_limit_g_lut[w] = limit << 8;
-		m_dyuv_limit_b_lut[w] = limit;
+		const uint8_t limit = (w < 0x100) ? 0 : w < 0x200 ? w - 0x100 : 0xff;
+		m_dyuv_limit_rgb_lut[w] = limit;
```
cuavas, line 1412: Please put use parentheses around non-trivial arguments to ternary conditional operators to make them easier to parse mentally, like:
```c++
		const uint8_t limit = (w < 0x100) ? 0 : (w < 0x200) ? (w - 0x100) : 0xff;
```

src/mame/philips/mcd212.h
Vincent-Halver: The chip uses a different variable to decide whether to use RGB555 or CLUT8. Regardless since the rest of the code wasn't included in this update, it's not easy to understand why it was changed. I'll undo it for now.

cuavas:
@MooglyGuy can you have a quick look at this, since you’re the main CD-i guy these days?

MooglyGuy:
I'd feel better if the comparison screenshots came from MAME, and not some other person's CD-i emulator in fucking RetroArch of all things, but they are using MAME's mcd212 implementation, and my original objection to the previously-closed PR no longer applies, as it appears to stick to integer math, so do what you will.

cuavas:
@Vincent-Halver can you please post comparison screenshots from MAME as a sanity check?

Vincent-Halver:
I'm afraid at the moment I don't quite know how to add support for Validation Disk (Europe) to the games list of the CD-i. If support was added, I can look into taking the comparison screenshots.

However, since this question is more about a sanity check, here is the bios screen comparison between retroarch and mame. As you can see, it's nearly identical.

Here is MAME 0.272
image

here is Retro Arch (running 0.239, but with these edits and a few more).
image

SAME_CDI in retro arch is in fact running the same core, just using a shim layer. Pressing tab opens the MESS configuration menu, seen here:
image

I hope that helps with the sanity check that the cores are in fact the same

FoxhackDN:
> I'm afraid at the moment I don't quite know how to add support for Validation Disk (Europe) to the games list of the CD-i. If support was added, I can look into taking the comparison screenshots.

You don't need to add support, if you have the disk image in bin+cue (which I found rather quickly), just copy it into the folder with the MAME executable and run it from the command line like this:

`mame cdisystemgoeshere -cdrom "Validation Disc (Europe) (Disc 1).cue"`

Or you can just run the cdi model you're trying to test and use the built in file manager: Press tab, select "File Manager", then cdrom, then navigate to where you have your disc image, select the cue, hit enter, then select "Reset System" to reboot it with the disc in the drive.

Vincent-Halver:
Thanks, that saved me some time, however even compiling and running as suggested, I don't see my changes taking effect as expected.
image
I'm not exactly sure why the format is still broken in MAME. The mosaic feature has clearly been deleted, but is still visible in the blocky discolored DYUV sample texture.

I think it is user error on my end. @FoxhackDN can you confirm what the test sample looks like for you when using this change?

Vincent-Halver:
Oh I understand the problem. I'm misinterpreting the control image. The blocky discolored part is actually RGB encoded data (not included in this change).

Adding the RGB fix back in will fix that, however it adds quite a few more changes. Rather than discuss the RGB format here, this is only targeting the DYUV format to reduce complexity of the change.

The screenshot indeed is correct, sorry for the confusion/mislabeling of the discolored pixels.

cuavas:
So the initial screenshots you posted weren’t with the code in this PR at all, and the screen you showed isn’t actually using DYUV.  Can you please:
* Find something that demonstrates what this PR fixes
* Show before/after screenshots in MAME, not some other emulator
* Not use anything involving RetroArch or other downstream projects for reference here

Vincent-Halver:
It seems that I ended up describing this poorly, and I got some things wrong in my screenshots.

The change removes support for Pixel Hold (Mosaic) which is banned for DYUV images, and cleans up the code to make it more readable.
Here is the before
image

Here is the after
image

These are pixel-identical. As such, you can consider this change is mostly about code readability and correctness in terms of removing a feature which does not exist on the chip (Pixel Hold Mosaic for DYUV).

Vincent-Halver:
Ok I definitely messed up explain this multiple times.
Here is the before:
image

Here is the after:
image

Here is the diff
image

The primary difference is that the DYUV color scale is 16/256 brighter.

You likely may see a white rectangle also marked as different. You can ignore that, as that is part of the RGB color scale, and is likely related to uninitialized memory or another spurious difference (RGB color scale is unimplemented, so the behavior is not well defined at the moment).

PR #13128, opened 2024-12-28
Make fs::err_t enum a subcategory of std::error_condition; use the st…

ajrhacker:
…andard codes in some places

### DIFF
--- a/src/lib/formats/fs_vtech.cpp
+++ b/src/lib/formats/fs_vtech.cpp
@@ -310,7 +310,6 @@ std::pair<std::error_condition, std::vector<u8>> vtech_impl::file_read(const std
 	while(pos < len) {
 		if(track >= 40 || sector >= 16)
 			break;
-		fprintf(stderr, "Reading to pos %d from track %d sector %d\n", pos, track, sector);
 		auto dblk = m_blockdev.get(track*16 + sector);
 		int size = len - pos;
 		if(size > 126)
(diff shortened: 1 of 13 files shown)


cuavas:
I think making the errors an `error_condition` category is a sensible change.

However, I don’t think using `std::errc` error conditions for errors relating to the emulated filesystem is a great idea.  It makes it impossible to distinguish between errors accessing the underlying storage holding the disk image and errors from accessing the filesystem withing the disk image.  If anything, I would have considered using `std::errc::not_supported` in place of `ERR_UNSUPPORTED` but kept `ERR_NOT_FOUND`, `ERR_NOT_EMPTY` and `ERR_NO_SPACE` equivalents in the filesystem error category.

src/lib/formats/fs_vtech.cpp
```diff
@@ -310,6 +310,7 @@ std::pair<err_t, std::vector<u8>> vtech_impl::file_read(const std::vector<std::s
 	while(pos < len) {
 		if(track >= 40 || sector >= 16)
 			break;
+		fprintf(stderr, "Reading to pos %d from track %d sector %d\n", pos, track, sector);
 		auto dblk = m_blockdev.get(track*16 + sector);
```
cuavas, line 314: Is this extra `printf` intentional?  It looks like the kind of thing that would get spammy fast.

ajrhacker: It's a leftover from an unrelated debugging effort. I'll remove it.

PR #13132, opened 2024-12-29
DRC backend fixes pt 2

987123879113:
- Fixed an issue with x64 LOAD not clearing the rest of the register in rare cases
- Fixed x86 LOADS not sign extending properly for 64-bit loads
- Fixed x86 data widths for FLOAD/FSTORE to support doubles
- Bandaid fix for clearing x86 upper half of register when writing to a 32-bit ireg. This won't cover all cases but should cover a lot of the common usages (I0-I3)

### DIFF
--- a/src/devices/cpu/drcbec.cpp
+++ b/src/devices/cpu/drcbec.cpp
@@ -391,10 +391,14 @@ void drcbe_c::generate(drcuml_block &block, const instruction *instlist, uint32_
 					psize[2] = 4;
 				if (opcode == OP_STORE || opcode == OP_FSTORE)
 					psize[1] = 4;
-				if (opcode == OP_READ || opcode == OP_READM || opcode == OP_FREAD)
+				if (opcode == OP_READ || opcode == OP_FREAD)
 					psize[1] = psize[2] = 4;
-				if (opcode == OP_WRITE || opcode == OP_WRITEM || opcode == OP_FWRITE)
+				if (opcode == OP_WRITE || opcode == OP_FWRITE)
 					psize[0] = psize[2] = 4;
+				if (opcode == OP_READM)
+					psize[1] = psize[3] = 4;
+				if (opcode == OP_WRITEM)
+					psize[0] = psize[3] = 4;
 				if (opcode == OP_SEXT && inst.param(2).size() != SIZE_QWORD)
 					psize[1] = 4;
 				if (opcode == OP_FTOINT)
@@ -1455,20 +1459,20 @@ int drcbe_c::execute(code_handle &entry)
 				DPARAM0 = m_space[PARAM2]->read_dword(PARAM1);
 				break;
 
-			case MAKE_OPCODE_SHORT(OP_READ8, 8, 0):     // DREAD   dst,src1,space_QOWRD
+			case MAKE_OPCODE_SHORT(OP_READ8, 8, 0):     // DREAD   dst,src1,space_QWORD
 				DPARAM0 = m_space[PARAM2]->read_qword(PARAM1);
 				break;
 
 			case MAKE_OPCODE_SHORT(OP_READM2, 8, 0):    // DREADM  dst,src1,mask,space_WORD
-				DPARAM0 = m_space[PARAM3]->read_word(PARAM1, PARAM2);
+				DPARAM0 = m_space[PARAM3]->read_word(PARAM1, DPARAM2);
 				break;
 
 			case MAKE_OPCODE_SHORT(OP_READM4, 8, 0):    // DREADM  dst,src1,mask,space_DWORD
-				DPARAM0 = m_space[PARAM3]->read_dword(PARAM1, PARAM2);
+				DPARAM0 = m_space[PARAM3]->read_dword(PARAM1, DPARAM2);
 				break;
 
 			case MAKE_OPCODE_SHORT(OP_READM8, 8, 0):    // DREADM  dst,src1,mask,space_QWORD
-				DPARAM0 = m_space[PARAM3]->read_qword(PARAM1, PARAM2);
+				DPARAM0 = m_space[PARAM3]->read_qword(PARAM1, DPARAM2);
 				break;
 
 			case MAKE_OPCODE_SHORT(OP_WRITE1, 8, 0):    // DWRITE  dst,src1,space_BYTE
--- a/src/devices/cpu/drcbec.cpp
+++ b/src/devices/cpu/drcbec.cpp
@@ -324,19 +324,39 @@ void drcbe_c::reset()
 
 void drcbe_c::generate(drcuml_block &block, const instruction *instlist, uint32_t numinst)
 {
+	// Calculate the max possible number of register clears required
+	uint32_t regclears = 0;
+
+	for (int inum = 0; inum < numinst; inum++)
+	{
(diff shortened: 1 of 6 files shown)


src/devices/cpu/drcbec.cpp
```diff
@@ -391,10 +391,14 @@ void drcbe_c::generate(drcuml_block &block, const instruction *instlist, uint32_
...
-				if (opcode == OP_READ || opcode == OP_READM || opcode == OP_FREAD)
+				if (opcode == OP_READ || opcode == OP_FREAD)
 					psize[1] = psize[2] = 4;
-				if (opcode == OP_WRITE || opcode == OP_WRITEM || opcode == OP_FWRITE)
+				if (opcode == OP_WRITE || opcode == OP_FWRITE)
 					psize[0] = psize[2] = 4;
+				if (opcode == OP_READM)
+					psize[1] = psize[3] = 4;
+				if (opcode == OP_WRITEM)
+					psize[0] = psize[3] = 4;
```
987123879113, line 401: OP_READ and OP_WRITE set the space size parameter (3rd parameter) to a fixed size of 4, but for OP_READM and OP_WRITEM that parameter is the 4th parameter because the 3rd parameter is the mask parameter.

987123879113:
@cuavas Please check https://github.com/mamedev/mame/pull/13132/commits/d2289603bcda9e0dac8dd6bcc41692f90446290d. I tried to make it so it'll automatically do an `reg &= 0xffffffff` if it detects that there was a 64 bit register write since the last time the register was written to, instead of doing it after every 32 bit register write just to reduce the amount of overhead.

src/devices/cpu/uml.h
```diff
@@ -398,6 +398,8 @@ namespace uml
 		// construction/destruction
 		constexpr instruction() : m_param{ } { }
 
+		static bool const is_param_out(opcode_t opcode, int paramnum) { assert(opcode < OP_MAX); assert(paramnum < m_numparams); return (s_opcode_info_table[opcode].param[paramnum].output & 0x02) != 0; }
```
cuavas, line 401: `const`-qualifying the return type doesn’t do anything when returning by value.

cuavas:
> @cuavas Please check [d228960](https://github.com/mamedev/mame/commit/d2289603bcda9e0dac8dd6bcc41692f90446290d). I tried to make it so it'll automatically do an `reg &= 0xffffffff` if it detects that there was a 64 bit register write since the last time the register was written to, instead of doing it after every 32 bit register write just to reduce the amount of overhead.

It looks reasonable.

987123879113:
Thanks. That covers the last of the things I found in the existing DRC/UML code so I'll open the PR for review then.

cuavas:
Can some of the hackery in the Hyperstone recompiler be removed with this?  There are 21 `#ifndef PTR64` guarding 29 `UML_DAND` ops, and with a quick look I count 20 that follow 32-bit loads.  That would leave just nine of them.

src/devices/cpu/drcbex86.cpp
```diff
@@ -5839,25 +5845,25 @@ void drcbe_x86::op_fload(Assembler &a, const instruction &inst)
...
 	// other index
 	else
 	{
 		Gp const indreg = indp.select_register(ecx);
 		emit_mov_r32_p32(a, indreg, indp);
-		a.mov(eax, ptr(u64(basep.memory(0)), indreg, 2));                               // mov   eax,[basep + 4*indp]
+		a.mov(eax, ptr(u64(basep.memory(0)), indreg, inst.size() == 8 ? 3 : 2));
 		if (inst.size() == 8)
-			a.mov(edx, ptr(u64(basep.memory(4)), indreg, 2));                           // mov   edx,[basep + 4*indp + 4]
+			a.mov(edx, ptr(u64(basep.memory(4)), indreg, inst.size() == 8 ? 3 : 2));
```
cuavas, line 5860: Can you please put explicit parentheses around non-trivial arguments of ternary conditional operators?  They can be difficult to read otherwise.  (Like `(inst.size() == 8) ? 3 : 2`.)

src/devices/cpu/drcbex86.cpp
```diff
@@ -5878,26 +5884,26 @@ void drcbe_x86::op_fstore(Assembler &a, const instruction &inst)
...
 	// other index
 	else
 	{
 		Gp const indreg = indp.select_register(ecx);
 		emit_mov_r32_p32(a, indreg, indp);
-		a.mov(ptr(u64(basep.memory(0)), indreg, 2), eax);                               // mov   [basep + 4*indp],eax
+		a.mov(ptr(u64(basep.memory(0)), indreg, inst.size() == 8 ? 3 : 2), eax);
 		if (inst.size() == 8)
-			a.mov(ptr(u64(basep.memory(4)), indreg, 2), edx);                           // mov   [basep + 4*indp + 4],edx
+			a.mov(ptr(u64(basep.memory(4)), indreg, inst.size() == 8 ? 3 : 2), edx);
```
cuavas, line 5906: Same with these ternary conditional operators.

src/devices/cpu/drcumlsh.h
```diff
@@ -146,8 +146,8 @@
 /* ----- 32-bit Floating Point Arithmetic Operations ----- */
 #define UML_FSLOAD(block, dst, base, index)                 do { using namespace uml; block.append().fsload(dst, base, index); } while (0)
 #define UML_FSSTORE(block, base, index, src1)               do { using namespace uml; block.append().fsstore(base, index, src1); } while (0)
-#define UML_FSREAD(block, dst, src1, space)                 do { using namespace uml; block.append().fsread(dst, src1, AS_##space); } while (0)
-#define UML_FSWRITE(block, dst, src1, space)                do { using namespace uml; block.append().fswrite(dst, src1, AS_##space); } while (0)
+#define UML_FSREAD(block, dst, src1, space)                 do { using namespace uml; block.append().fsread(dst, src1, space); } while (0)
+#define UML_FSWRITE(block, dst, src1, space)                do { using namespace uml; block.append().fswrite(dst, src1, space); } while (0)
```
cuavas, line 150: This is a nasty leftover from when only the predefined address spaces could be used – nice work catching it and cleaning it up.

cuavas (approved):
Apart from the trivial thing with ternary conditional operators, this looks OK to me.

987123879113:
>Can some of the hackery in the Hyperstone recompiler be removed with this? There are 21 #ifndef PTR64 guarding 29 UML_DAND ops, and with a quick look I count 20 that follow 32-bit loads. That would leave just nine of them.

I want to say yes but more testing is needed. I tried out btplay2k (dgpix) and with the `#ifndef PTR64` stuff removed it won't boot with the x86 DRC or C DRC in master, but will boot on both x86 DRC and C DRC with all of the recent changes + this PR.

987123879113:
Tested some more games using a clean built of master (before this PR) and with the PR, and before the PR they would hang but with the PR the games boot and seem to be running without issue. So I think all of those `#ifndef` can be removed now.

cuavas:
I removed all the ones immediately following a 32-bit load to I0-I3 in d9a1cc5dc9d91c4071ff8a0ba40bdd928d4a6587.  That leaves six following logical operations.

PR #13133, opened 2024-12-29
Various Atari 2d drivers updates & cleanups:

cam900:
- Various drivers:
  - Reduce tilemap size for reduce unnecessary glitchy area
  - Use GFXENTRY_SCALE for scaled gfxs
  - Fix namings
  - Fix notes
- atari/arcadecl.cpp:
  - Use gfxdecode into arcadecl_state only (sparkz_state hasn't sprites)
- atari/atarimo.cpp:
  - Remove unused parameter in config
  - Remove gfx information modification (_raw function ignores them)
  - Fix spacing
- atari/atarisy1.cpp:
  - Reduce preprocessor defines
- atari/atarisy2.cpp:
  - Reduce preprocessor defines
- atari/skullxbo.cpp:
  - Remove unnecessary init functions


src/mame/atari/xybots.cpp
```diff
@@ -243,7 +242,9 @@ void xybots_state::video_int_ack_w(uint16_t data)
 uint16_t xybots_state::special_port1_r()
 {
 	int result = m_ffe200->read();
-	result ^= m_h256 ^= 0x0400;
+	result ^= m_h256;
+	if (!machine().side_effects_disabled())
+		m_h256 ^= 0x0400;
 	return result;
```
cuavas, line 248: Isn’t this going to change behaviour?  Think about associativity – assignment operators are right-associative in C++ (unlike other binary operators which are left-associative).  This:
```c++
	result ^= m_h256 ^= 0x0400;
```
Will be parsed as:
```c++
	result ^= (m_h256 ^= 0x0400);
```
Equivalent to:
```c++
	m_h256 ^= 0x0400;
	result ^= m_h256;
```

To do it with the side effects check, you’ll need to do:
```c++
	result ^= m_h256 ^ 0x0400;
		m_h256 ^= 0x0400;
```

src/mame/atari/atarivad.cpp
```diff
@@ -312,10 +313,10 @@ void atari_vad_device::internal_control_write(offs_t offset, uint16_t newword)
 		// latch enable
 		case 0x0a:
 			// check for palette banking
-			if (m_palette_bank != (((newword & 0x0400) >> 10) ^ 1))
+			if (m_palette_bank != (BIT(newword, 10) ^ 1))
 			{
 				screen().update_partial(screen().vpos());
-				m_palette_bank = ((newword & 0x0400) >> 10) ^ 1;
+				m_palette_bank = BIT(newword, 10) ^ 1;
```
cuavas, line 319: The `^ 1` just inverts the extracted bit.  Would it be more readable as:
```c++
			// check for palette banking
			if (m_palette_bank != BIT(~newword, 10))
			{
				screen().update_partial(screen().vpos());
				m_palette_bank = BIT(~newword, 10);
```

src/mame/atari/batman.cpp
```diff
@@ -452,7 +451,7 @@ void batman_state::batman(machine_config &config)
 
 	TILEMAP(config, "vad:playfield", "gfxdecode", 2, 8, 8, TILEMAP_SCAN_COLS, 64, 64).set_info_callback(FUNC(batman_state::get_playfield_tile_info));
 	TILEMAP(config, "vad:playfield2", "gfxdecode", 2, 8, 8, TILEMAP_SCAN_COLS, 64, 64, 0).set_info_callback(FUNC(batman_state::get_playfield2_tile_info));
-	TILEMAP(config, "vad:alpha", "gfxdecode", 2, 8, 8, TILEMAP_SCAN_ROWS, 64, 32, 0).set_info_callback(FUNC(batman_state::get_alpha_tile_info));
+	TILEMAP(config, "vad:alpha", "gfxdecode", 2, 8, 8, TILEMAP_SCAN_ROWS, 64, 30, 0).set_info_callback(FUNC(batman_state::get_alpha_tile_info));
```
cuavas, line 454: Why the reduction in size here?  How much tile map memory is there backing the alpha tilemap?

src/mame/atari/cybstorm.cpp
```diff
@@ -489,7 +488,7 @@ void cybstorm_state::round2(machine_config &config)
 	m_vad->scanline_int_cb().set_inputline(m_maincpu, M68K_IRQ_4);
 	TILEMAP(config, "vad:playfield", m_gfxdecode, 2, 8, 8, TILEMAP_SCAN_COLS, 64, 64).set_info_callback(FUNC(cybstorm_state::get_playfield_tile_info));
 	TILEMAP(config, "vad:playfield2", m_gfxdecode, 2, 8, 8, TILEMAP_SCAN_COLS, 64, 64, 0).set_info_callback(FUNC(cybstorm_state::get_playfield2_tile_info));
-	TILEMAP(config, "vad:alpha", m_gfxdecode, 2, 8, 8, TILEMAP_SCAN_ROWS, 64, 32, 0).set_info_callback(FUNC(cybstorm_state::get_alpha_tile_info));
+	TILEMAP(config, "vad:alpha", m_gfxdecode, 2, 8, 8, TILEMAP_SCAN_ROWS, 64, 30, 0).set_info_callback(FUNC(cybstorm_state::get_alpha_tile_info));
```
cuavas, line 491: Same for this one – does the game have enough tilemap memory for 32 rows?

cuavas (commented):
I don’t particularly like reducing the tilemap area to less than what the hardware supports just because the games don’t use all of it.  If the hardware has RAM to support the full area, it should be used.  If nothing else, it helps when developing ROM hacks if you can visualise the entire contents of tile map memory.

src/mame/atari/batman.cpp
cam900: In this case, alpha tilemap memory size is less than 64 * 32 words, And it results of potentially crash and glitch at bottom most area. That's why I reduced tilemap size.

src/mame/atari/cybstorm.cpp
cam900: Same case at above, Tilemap memory size is 0xf00 bytes (64 columns x 30 rows x 2 bytes per tile).

PR #13135, opened 2024-12-29
imagedev/flopdrv.cpp: Remove code supporting functionality that's no longer used

plaes:
Seems like the list of machines still using legacy floppy drives has shrinked rapidly..


cuavas:
@galibert do you agree with this in principle?  It removes functionality that was used to support systems that have been migrated to the “new floppy” device, leaving just the code used by systems still using the legacy floppy drive devices.

PR #13141, opened 2024-12-30
Update spg2xx_jakks.cpp - add 3rd button for jak_spd3

mamehaze:


mamehaze:
(missed this at first, but once you get to the 4th stage you need the C button)

PR #13143, opened 2024-12-30
Update 3dom2.cpp - add a note next to the hack causing the extreme slowdown, and turn it off as a test

mamehaze:


mamehaze:
I just wanted to document in public that this is the source of the extreme slowdown.  It might be worth turning the hack off to see if anything breaks at this point (maybe it was a workaround for something else that is no longer an issue?)

The games all still seem to boot with the hack turned off, but it could be a stability thing later on.

happppp:
If you determined the hack (whatever was it for?) is obsolete, just remove it altogether?
If a regression is found in the future, the usual method is to try version xyz in the past, and a git history check would reveal this. I think there's no need to keep the if 0 in.

MooglyGuy:
Either way it's been a pretty standard point of feedback for several years now to not put code behind #if 0/#endif blocks due to the high likelihood of code rot. Aside from hap's point of feedback (just remove the line), put code cutouts in an `if (SOME_CONSTEXPR_BOOL_MEMBER)` instead. It will ensure the inner code continues to get compiled and therefore does not rot over time.

angelosa:
https://github.com/mamedev/mame/blob/d985c984ae0321bf8ec85d5dab92a451a9bac19d/src/mame/konami/konamim2.cpp#L1614

We have five games here, if the real reason is "I'm not gonna bother testing in full because they don't run at acceptable speeds" then the problem is at root: premature promotion to working or an actual MAME machine flag that acknowledges there's a very high requirement from host. 

As an aside, there's a way excessive pressure over the [QA team](https://mametesters.org/rules.html) over rules written in another time age: I had to repeat again lately that the `MACHINE_NOT_WORKING` isn't a valid reason to reject a report about making MAME to outright crash either implicitly or explictily thru `emu_fatalerror`.

mamehaze:
This is just here so that it has visibility, hence leaving it in draft status.  How you want to take if forward is up to you; I think maybe it's legacy from a time the DRC wasn't picking up changes properly or something, but I'm not sure.

There's public interest in the driver (to the point people have been making videos about it) and those people would probably be best positioned to do any testing to see if this causes issues or not, but that essentially requires at least one public release build where the hack is turned off to see.

happppp:
Instead of forever marking it as draft, the correct place for posting this slowdown issue is: https://mametesters.org/view_all_bug_page.php or https://github.com/mamedev/mame/issues

Hydreigon223:
> Instead of forever marking it as draft, the correct place for posting this slowdown issue is: https://mametesters.org/view_all_bug_page.php or https://github.com/mamedev/mame/issues

Done by me before but closed by @cuavas . You can view the report here: https://mametesters.org/view.php?id=8980

happppp:
Ok, let's just remove that hack.

PR #13148, opened 2024-12-31
Add comments on how the Uniwars Warp effect works.

prswan:
Add documentation on the Irem hardware hack that implements the warp effect in Uniwars.
- What the effect is
- How it's triggered.

### DIFF
--- a/src/mame/galaxian/galaxian.cpp
+++ b/src/mame/galaxian/galaxian.cpp
@@ -710,6 +710,52 @@ t) 'bagmanmc'
   - DSW bit 6 was previously used for "Bonus Lives" settings, but it has no effect
     in this set because of 'NOP' instructions from 0x3501 to 0x3507.
 
+u) uniwars
+
+    Uniwars Warp Analysis on an Irem PCB from PrSwan (Paul Swan)
+    ------------------------------------------------------------
+
+    - What is the warp effect?
+      The game PCB has several fly wires of which Red & Green relate to the warp effect.
+
+      Video PCB connections
+      Red   => IC K8 (7420)  pin 12
+      Green => IC F6 (74161) pin 14
+
+      The Green signal is a clock generated on the video PCB and the 161 is likely one of the video H/V counters.
+      It measured 10.44us.
+
+      The Red signal is input to the video PCB:
+      - Warp off => idle high
+      - Warp on  => Green clock
+
+      The effect can be manually triggered by disconnecting the plug and bridging the red & green wire.
+      The warp comes from the star field generator - with warp enabled all other graphics elements are unchanged.
+
+      There is a cut track isolating K8 pin 12 to allow the Red wire to control pin 12.
+
+      Trace of the K8 pin 12 cut track:
+      => IC K9 (7402)  pin  3 (input, 2-in NOR)
+      => IC E9 (7432)  pin 12 (input, 2 in OR)
+      => IC D9 (74173) pin  4 (Q output, latch)
+
+      Comparing with the Midway schematic suggests Irem IC K8 === Midway IC 3C and K8 pin 12 corresponds to ~256H* feeding Midway IC 3C pin 9.
+
+    - How is the effect triggered?
+      A spare IC position is populated with a 74LS157 mux connected as follows:
+
+      - pin 1 - S => IC C2 (LS259) pin 9
+      - pin 2 - A => not connected but evidence of prior solder, wire may have fallen off.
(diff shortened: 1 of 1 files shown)


src/mame/galaxian/galaxian.cpp
```diff
@@ -710,6 +710,52 @@ t) 'bagmanmc'
...
+
+    - What is the warp effect?
+      The game PCB has several fly wires of which Red & Green relate to the warp effect.
+
+      Video PCB connections
+      Red   => IC K8 (7420)  pin 12
+      Green => IC F6 (74161) pin 14
+
+      The Green signal is a clock generated on the video PCB and the 161 is likely one of the video H/V counters.
+      It measured 10.44us.
```
cuavas, line 726: *What* measured 10.44 microseconds?  The pulse width?  The period?  Something else?  What’s the duty cycle?  What divider chain on the board could be producing this signal?  What master clock is it derived from?

src/mame/galaxian/galaxian.cpp
```diff
@@ -710,6 +710,52 @@ t) 'bagmanmc'
   - DSW bit 6 was previously used for "Bonus Lives" settings, but it has no effect
     in this set because of 'NOP' instructions from 0x3501 to 0x3507.
 
+u) uniwars
+
+    Uniwars Warp Analysis on an Irem PCB from PrSwan (Paul Swan)
+    ------------------------------------------------------------
+
+    - What is the warp effect?
+      The game PCB has several fly wires of which Red & Green relate to the warp effect.
```
cuavas, line 719: This doesn’t actually answer the question of what the warp effect is.  Is it a sound effect?  A video effect?  Something else?

src/mame/galaxian/galaxian.cpp
```diff
@@ -710,6 +710,52 @@ t) 'bagmanmc'
...
+      The warp comes from the star field generator - with warp enabled all other graphics elements are unchanged.
+
+      There is a cut track isolating K8 pin 12 to allow the Red wire to control pin 12.
+
+      Trace of the K8 pin 12 cut track:
+      => IC K9 (7402)  pin  3 (input, 2-in NOR)
+      => IC E9 (7432)  pin 12 (input, 2 in OR)
+      => IC D9 (74173) pin  4 (Q output, latch)
+
+      Comparing with the Midway schematic suggests Irem IC K8 === Midway IC 3C and K8 pin 12 corresponds to ~256H* feeding Midway IC 3C pin 9.
```
cuavas, line 742: This is more confusing than anything else without an explanation of what those logic gates do or what those signals represent on the Midway board.

src/mame/galaxian/galaxian.cpp
```diff
@@ -710,6 +710,52 @@ t) 'bagmanmc'
...
+
+    - How is the effect triggered?
+      A spare IC position is populated with a 74LS157 mux connected as follows:
+
+      - pin 1 - S => IC C2 (LS259) pin 9
+      - pin 2 - A => not connected but evidence of prior solder, wire may have fallen off.
+      - pin 3 - B => Green wire
+      - pin 4 - Q => Red wire
+      - pin 15 - ~E => GND
```
cuavas, line 751: A floating input will behave like a logic high with poor noise immunity on a TTL chip due to how the common base amplifier input works.  It’s almost definitely supposed to be connected to something.

src/mame/galaxian/galaxian.cpp
prswan: Updated with clarification & TODO.

src/mame/galaxian/galaxian.cpp
prswan: Added text description of the effect.

src/mame/galaxian/galaxian.cpp
prswan: Rewritten to make it a bit clearer that this is the analysis of why we think this is 256H, however if you prefer I can remove the tracing context entirely and just leave the conclusions? LMK.

src/mame/galaxian/galaxian.cpp
prswan: Added a TODO note to confirm if another PCB shows up and the suggestion o use 256H*.

src/mame/galaxian/galaxian.cpp
prswan: @cuavas can you rereview on if this is clearer? Thx :)

PR #13151, opened 2024-12-31
atari/atarig1.cpp, atarig42.cpp, atarigt.cpp, atarigx2.cpp, atarirle.cpp: Cleanups:

cam900:
- Reduce preprocessor defines
- Fix save states
- Fix naming
- Fix spacing
- Reduce literal tag usages
atari/atarigx2.cpp:
- Reduce runtime tag lookups

### DIFF
--- a/src/mame/atari/atarigt.h
+++ b/src/mame/atari/atarigt.h
@@ -14,11 +14,6 @@
 #include "emupal.h"
 #include "tilemap.h"
 
-#define CRAM_ENTRIES        0x4000
-#define TRAM_ENTRIES        0x4000
-#define MRAM_ENTRIES        0x8000
-
-#define ADDRSEQ_COUNT   4
 
 class atarigt_state : public atarigen_state
 {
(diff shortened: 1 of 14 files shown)


src/mame/atari/atarirle.h
```diff
@@ -69,6 +50,20 @@ class atari_rle_objects_device : public device_t,
...
+	static const uint16_t PRIORITY_MASK = ((0xffff << PRIORITY_SHIFT) & 0xffff);
+	static const uint16_t DATA_MASK     = (PRIORITY_MASK ^ 0xffff);
+
+	static const int CONTROL_MOGO       = 1;
+	static const int CONTROL_ERASE      = 2;
+	static const int CONTROL_FRAME      = 4;
+
+	static const int COMMAND_NOP        = 0;
+	static const int COMMAND_DRAW       = 1;
+	static const int COMMAND_CHECKSUM   = 2;
```
cuavas, line 65: You should use `static inline constexpr` for these.

src/mame/atari/atarigt.h
```diff
@@ -52,8 +47,8 @@ class atarigt_state : public atarigen_state
 	virtual void video_start() override ATTR_COLD;
 
 private:
+	static const unsigned ADDRSEQ_COUNT = 4;
```
cuavas, line 50: This should be `static inline constexpr` as well.

PR #13153, opened 2024-12-31
new NOT WORKING machines (JAKKS Pacific EA Sports Classic NHL/Fifa + Bob the Builder) (SPG110 work)

mamehaze:
new NOT WORKING machines
EA Sports Classics: NHL 95 & FIFA Soccer 96 [David Haywood, Team Europe]
Bob the Builder - Project: Build It [David Haywood, Team Europe]

- improved SPG110 interrupt handling, tile y flip, and prepare for other improvements


mamehaze:
(boots, but hangs going ingame, probably due to improperly emulated SPG110 differences)

mamehaze:
Bob The Builder mostly works apart from the 'Can We Hear It?' minigame which fails to play the sound clip, and then hangs, presumably waiting for it to finish.  Not surprising as the SPG110 sound is currently just kludged to somewhat work with the SPG2xx core and probably needs its own instead.

EA Sports is mostly functional now, but NHL has no collisions at the top-right and bottom of the playfield, which is problematic.  I wonder if this is a math or flag bug in the unSP core.

src/devices/machine/spg110_video.cpp
```diff
@@ -531,7 +641,7 @@ void spg110_video_device::tmap1_regs_w(offs_t offset, uint16_t data)
 void spg110_video_device::map_video(address_map &map)
 {
 	// are these addresses hardcoded, or can they move (in which case tilemap system isn't really suitable)
-	map(0x00000, 0x03fff).ram(); // 2fff?
+	map(0x00000, 0x02fff).ram(); // 2fff?
```
cuavas, line 644: Is the comment still relevant?

src/devices/machine/spg110_video.cpp
```diff
@@ -727,27 +893,29 @@ void spg110_video_device::palette_w(offs_t offset, uint16_t data, uint16_t mem_m
...
-
-	const uint32_t page1_addr = 0x40 * m_tilebase;//0x40 * m_video_regs[0x20];
-	const uint32_t page2_addr = 0x40 * m_tilebase;//0x40 * m_video_regs[0x21];
-	uint16_t *page1_regs = tmap0_regs;
-	uint16_t *page2_regs = tmap1_regs;
+	const pen_t *pens = m_palette->pens();
+	const uint32_t page1_addr = 0x40 * m_tilebase[0];//0x40 * m_video_regs[0x20];
+	const uint32_t page2_addr = 0x40 * m_tilebase[0];//0x40 * m_video_regs[0x21];
+	uint16_t* page1_regs = tmap0_regs;
+	uint16_t* page2_regs = tmap1_regs;
```
cuavas, line 900: What purpose do the `page1_reg` and `page2_regs` locals serve?

src/devices/machine/spg110_video.cpp
```diff
@@ -756,23 +924,14 @@ uint32_t spg110_video_device::screen_update(screen_device &screen, bitmap_rgb32
...
 	}
-
-	if (m_video_irq_enable & 1)
+	else
 	{
-		m_video_irq_status |= i;
-		check_video_irq();
+		m_blk_irq_flag = 1;
+		update_video_irqs();
 	}
```
cuavas, line 936: This is equivalent to:
```c++
	m_blk_irq_flag = state;
	update_video_irqs();
```

src/mame/tvgames/spg110.cpp
```diff
@@ -56,6 +60,20 @@ class spg110_game_state : public driver_device
...
+	spg110_easports_game_state(const machine_config &mconfig, device_type type, const char *tag) :
+		spg110_game_state(mconfig, type, tag)
+	{ }
+public:
+	void easports_pal(machine_config &config);
+
+protected:
+private:
+	virtual void mem_map(address_map &map) override ATTR_COLD;
+};
```
cuavas, line 75: Redundant extra `public:`, unnecessary `protected:`.

src/devices/machine/spg110_video.cpp
mamehaze: why are you complaining about code I haven't even changed?

src/devices/machine/spg110_video.cpp
mamehaze: no, none of the comments around there are at this point, was confirmed from documentation

src/devices/machine/spg110_video.cpp
cuavas: You’ve reformatted every one of those lines, and changed other code in the function.  If the locals don’t serve any purpose, now is a great time to get rid of them.

src/mame/tvgames/spg110.cpp
mamehaze: I've always been told to put a protected block so that it's ready for things to be added later, it's good coding practice at most workplaces.  I can remove it if you really want though.  as for the public, yeah, that can go.

src/devices/machine/spg110_video.cpp
cuavas: Well please either clarify why the comment repeats the upper limit of the RAM mapping with a question mark, or remove it.  It isn’t going to help someone reading the code like this.

src/devices/machine/spg110_video.cpp
mamehaze: essentially this kind of thing will get addressed in a clean-up PR after more work has been done, but it's part of the existing codebase, it's nothing to do with this PR and I find it annoying when you start commenting on out-of-scope things like this.

src/mame/tvgames/spg110.cpp
cuavas: It’s visual clutter that serves no purpose if there’s nothing in the section.  If protected members need to be added, it can be added later.

mamehaze: interesting that you seem to take the opposite approach to every company I've worked for, but fair enough.

src/devices/machine/spg110_video.cpp
mamehaze: well I'll change it for now anyway, in reality this section needs more work and it was due to be changed then, as there are factors not being taken into account right now.

PR #5929, opened 2019-11-19
mb8421.cpp : Support variable size and data width, Use template for base device, Add notes

cam900:
twinkle.cpp, firebeat.cpp : Device-fied CY7C131 Dual port SRAM, Add notes
esqmr.cpp : Add IDT7130 Dual port SRAM placeholder, Add notes

PR #13167, opened 2025-01-03
dynax/ddenlovr.cpp: DIP Switches full improvements starting from new manuals

UnclePetros:
Ref. Bug ID [#05929](https://mametesters.org/view.php?id=5929)

Fixed DIP Switches for the following games (using dynax operator manuals): hginga, hgokou, hgokbang, hkagerou, hanakanz, htengoku, mjmyornt, mjmyorntr, mjmyster, mjmyorn2, mjchuuka, mjschuka, daimyojn, kotbinyo, kotbinsp, mjreach1, jongtei.
- hginga
  - uploaded manual on [#05929](https://mametesters.org/view.php?id=5929) 
  - improved DIP Switches using hginga operator manual
  - SW1:8 Credits for Note: not sure how it works, to check in-game
  - updated function hginga_dsw_r, common to hginga and hgokou, in order to correctly manage 10-bit DIP Switches
- hgokou, hgokbang
  - improved DIP Switches using similar game hginga operator manual
  - SW1:8 Credits for Note: not sure how it works, to check in-game
  - updated function hginga_dsw_r, common to hginga and hgokou, in order to correctly manage 10-bit DIP Switches
- hkagerou
  - uploaded manual on [#05929](https://mametesters.org/view.php?id=5929) 
  - improved DIP Switches using hkagerou operator manual
  - SW1:8 Credits for Note: not sure how it works, to check in-game
  - updated function hanakanz_dsw_r, common to hkagerou and hanakanz, in order to correctly manage 10-bit DIP Switches
- hanakanz
  - improved DIP Switches using similar game hkagerou operator manual
  - SW1:8 Credits for Note: not sure how it works, to check in-game
  - updated function hanakanz_dsw_r, common to hkagerou and hanakanz, in order to correctly manage 10-bit DIP Switches
- htengoku
  - uploaded manual on [#05929](https://mametesters.org/view.php?id=5929) 
  - improved DIP Switches using htengoku operator manual
  - SW3:3 Credits per Note: not sure how it works, to check in-game
  - updated function htengoku_dsw_r, in order to correctly manage 10-bit DIP Switches
- mjmyornt
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches
  - added common function mjmyster_dsw_r in mjmyornt config function, in order to correctly manage 10-bit DIP Switches
- mjmyster
  - Using manual present on [#05560](https://mametesters.org/view.php?id=5560) to adjust some DIP Switches
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches
  - updated function mjmyster_dsw_r, in order to correctly manage 10-bit DIP Switches
- mjmyorn2
  - updated Coinage DIP Switch to adapt to mjmyster DIP Switches improvements
- mjchuuka
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches
  - Fixed Coinage, Bets and Hopper BIP switches
  - Odds rate seems to have no effect in-game
- mjschuka, mjmyorntr
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches
  - Fixed Coinage, Bets and Hopper DIP switches
  - added common function mjschuka_dsw_r
- daimyojn
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches
- kotbinyo, kotbinsp, mjreach1, jongtei
  - Removed DSW5, and added 2 high bits to DSW1-4 to correctly manage 10-bit DIP switches

For all games, checked almost all DIP Switches values in-game.
Some of them are supposed to be correct, but I can't test in-game, because is a matter of luck to reply the right game flow.

Note: as default values for DIP Switches I've used the last configuration suggested by the manuals


cuavas:
The changes you’ve made to how DIP switches are read and how I/O ports are organised are regressive.  I’ll look at the manuals myself at some point, but this is not the way to do this.

UnclePetros:
> The changes you’ve made to how DIP switches are read and how I/O ports are organised are regressive. I’ll look at the manuals myself at some point, but this is not the way to do this.

Ok mate, but it would be useful if you sent me a CR, so that I can improve the changes, and merge the pull request, or at least explain the problems with my code.
Otherwise it will be impossible for a novice to begin to contribute.

cuavas:
I’ve been working on Dynax DIP switches for weeks, you can look at any of my changes in that area to get an idea of what to do.  If you’re not even prepared to do that, it doesn’t instill much confidence.

But anyway:
* Removing object finders in favour of using run-time tag lookups is the opposite of what we’re been doing for years.  You’ll notice numerous commits from @Osso13 and @cam900 replacing tag lookups with object finders.  That should be a pretty good hint of the direction things are supposed to be going.
* Switch matrices are not multiplexers.  I’d implemented the DIP switch matrix properly for Hanafuda Hana Tengoku.  You changed it back.  Once again, this is something we’ve been fixing as we find examples.  It would be more sensible to fix other cases you find, not make regressive changes to examples that have been fixed.
* It’s better to have I/O ports organised how they’re logically wired in hardware if possible.  The bit manipulations you need to do after reorganising I/O ports should be a pretty good hint that it isn’t an improvement.
* DIP switch locations are independent of I/O ports.  You can have switches from the same bank on different ports, or different banks on a single port.  You can even split bits for a single setting across banks with syntax like `PORT_DIPLOCATION("SW 1:3,4,SW 2:3,4")`.  There’s no reason to force DIP switches from the same bank into a single I/O port if they aren’t wired that way.
* We use the first set of recommended settings, adjusted to have attract sound on to facilitate testing.  You could see this from any of my commits to dynax.cpp or ddenlovr.cpp.
* There’s no point messing with the fake DIP switches for hopper switches – either hook the hoppers up properly, or don’t bother.  I did this for a few games, but the code is a tangled mess so I put off fixing the remaining ones where the dependencies made it difficult.
* Credits per note in Hanafuda Hana Tengoku is pretty easy to test.  If you didn’t test basic stuff like that, it suggests you weren’t very thorough.
* I’d have to check the whole thing and add notes with the Japanese descriptions from the manuals myself anyway, and by the time I’ve done that, I may as well do it myself anyway.

UnclePetros:
Ok, thank you for the detailed answer. 
I'll take my time to check all the points.
But what it looked strange to me is the use of "DSW5" for representing the high 2 bits of the first 4 DSW.
From manuals there are only 4 x 10-bit DSW. There is also a DSW5, but it's only 4-bit and not needed to be emulated.
I've removed DSW5 in code, because, there was some single settings that use bits across port DSW1 and DSW5 (as example), so you have to set 2 values in  ui menu to change only one settings. It was obviously cumbersome.

UnclePetros:
Ok, thank you for the detailed answer. 
I'll take my time to check better all the points.
But, in the meantime, I'll give you a quick feedback on all points:

> * Removing object finders in favour of using run-time tag lookups is the opposite of what we’re been doing for years.  You’ll notice numerous commits from @Osso13 and @cam900 replacing tag lookups with object finders.  That should be a pretty good hint of the direction things are supposed to be going.

What do you refer to in detail? Sorry, but I don't understand here.  

> * Switch matrices are not multiplexers.  I’d implemented the DIP switch matrix properly for Hanafuda Hana Tengoku.  You changed it back.  Once again, this is something we’ve been fixing as we find examples.  It would be more sensible to fix other cases you find, not make regressive changes to examples that have been fixed.

Yes, It was implemented properly. My change was to remove DSW5, because there are cases in which this fake DIP Switch should not be used (that is when a setting is made up of bits from DSW5 and another DSW).
So, to standardize the implementation, I removed DSW5 from the code.

> * It’s better to have I/O ports organised how they’re logically wired in hardware if possible.  The bit manipulations you need to do after reorganising I/O ports should be a pretty good hint that it isn’t an improvement.

I agree. But the main need of bit manipulation was to put togheter bit 8,9,10 in order to have a unique setting voice in menu in-game. I've copied implementation from other part of the code. Nothing new.
Anyway, honestly it was strange for me the use of "DSW5" for representing the 2 high bits of the 4 DSW.
Because the pcb has only 4 10-bit DSW. And has a DSW5, but of only 4-bit and used for other purposes (not needed to be emulated). 

> * DIP switch locations are independent of I/O ports.  You can have switches from the same bank on different ports, or different banks on a single port.  You can even split bits for a single setting across banks with syntax like `PORT_DIPLOCATION("SW 1:3,4,SW 2:3,4")`.  There’s no reason to force DIP switches from the same bank into a single I/O port if they aren’t wired that way.

The reason was there are some settings that are of the type SW1:8,9,10. So If we have bit 8 in DSW1 and bit 9,10 in DSW5, then in the in-game DSW Menu, We have to combine the two item corresponding to DSW1 bit 8 and DSW5 bit 9,10 to have the desired setting. A little bit confusing.  

> * We use the first set of recommended settings, adjusted to have attract sound on to facilitate testing.  You could see this from any of my commits to dynax.cpp or ddenlovr.cpp.

Ok, clear. I understand.

> * There’s no point messing with the fake DIP switches for hopper switches – either hook the hoppers up properly, or don’t bother.  I did this for a few games, but the code is a tangled mess so I put off fixing the remaining ones where the dependencies made it difficult.

Ah Ok. In fact I don't understood why Hopper and Bet, sometimes resulted repeated in DIP Switches menu UI.
Anyway, I just test it in Service Mode to verify it correctly toggle ON/OFF, and at least have changed the name, copying from other games implementations.

> * Credits per note in Hanafuda Hana Tengoku is pretty easy to test.  If you didn’t test basic stuff like that, it suggests you weren’t very thorough.

Yes, I've tested it. But, 
1. sometimes it is represented in code with only two basic values x5, x10. Other times is used MAHJONG_NOTE_CREDITS #define to represent the value multiplied the current coinage chosen.
2. If using the define, if doesn't take care of the alternate coinage case.
For these reasons, I was confused.
I had assumed the use of MAHJONG_NOTE_CREDITS #define was the right one, because it was used in a DIP Switches well organised and with all japanese description. 

> * I’d have to check the whole thing and add notes with the Japanese descriptions from the manuals myself anyway, and by the time I’ve done that, I may as well do it myself anyway.

Ok, sorry for that, I don't know I had to report japanese desc as comments in code for all DIP Switches.
Next time, I'll do it. Otherwise, you can open a CR to me, and I report them from the manual (even though a bit harder to extract text from image, or try to manual copy, as manuals are in a poor resolution).

PR #13172, opened 2025-01-04
pasogo.xml: Metadata updates

FoxhackDN:
Added missing cart numbers and an additional undumped game.


cuavas:
That’s not metadata, that’s a comment.

PR #13175, opened 2025-01-04
konami/88games.cpp: DIP Switches update 

UnclePetros:
DIP Switch SW3:4 update (from HyperSports Special manual)
image

The switch is shared for 88games, konami88 and hypsptsp.
It works fine for konami88 and hypsptsp.
Not work as expected for 88games (after the end of game it always restart the game).
May be it has a different use in 88games?

I have uploaded hypsptsp manual [here](https://www.arcade-museum.com/manuals-videogames/K/Konami88HyperSportsSpecialManual_00004.pdf) on arcade-museum.


cuavas:
If the setting doesn’t apply to 88 Games, it should be left as “Unknown” for 88 Games.

When it says that it repeats three times, is that three repetitions *after* the first round (i.e. four total), or three repetitions *including* the first round (i.e. three total)?  I added the setting in f8324bf6da5e38935fff9fba993b389ea3e1854f, assuming the first interpretation.  The only even I can win in this game is the 100m Dash, so I can’t test it.

PR #13176, opened 2025-01-04
misc/4enraya.cpp: swapped difficulty and speed DIP Switches

UnclePetros:
Swapped _Difficulty_ and _Speed_ DIP Switches as per manual, for games **4enraya** and **4enrayaa**.
Tested in-game by measuring time to shoot 5 arrows. Works as expected by the manual.
image


cuavas:
This still seems wrong:
* The “Demo Sounds” DIP switch doesn’t seem to affect sound in attract mode.
* The DIP switches are in the opposite order to that table.

UnclePetros:
> This still seems wrong:
> 
> * The “Demo Sounds” DIP switch doesn’t seem to affect sound in attract mode.
Right. This bit is part of the Pieces setting.
Let me check, test and commit PR.
> * The DIP switches are in the opposite order to that table.
Yes, I noticed. Don't know if an error in manual, or in the implementation (related to the order).

PR #13177, opened 2025-01-04
oberheim/dmx.cpp: Adding mouse control to faders.

m1macrophage:


felipesanches:
@m1macrophage I was not able to drag the sliders. I have the impression that the embedded Lua script is not even running at all in my setup (even though I do have the layout plugin enabled).

I'm testing on Debian. In which operating system have you tested this?

felipesanches:
In my setup, the sliders (faders) in the layour react to changes made in the "Slider Controls" dialog. But not to mouse click & drag events.

Screenshot From 2025-01-08 15-32-51

m1macrophage:
It works fine for me (on Ubuntu). If you haven't done so already, try adding some `print`s in different parts of the script, to check if it is even being initialized, and activated by mouse events.

It might also be worth trying other layouts with scripts, in case this is a broader issue with your build or setup.

m1macrophage:
Also, I don't need to explicitly enable plugins. They seem to be enabled by default. But just in case, try `./mame -plugins -plugin layout` as suggested in https://docs.mamedev.org/techspecs/layout_script.html#introduction .

felipesanches:
ok, got it!

It was `-plugin  layout` that was missing indeed!
That's weird, because I though this was on by default.

Anyway, now I got it working. Thanks for the help! ;-)

src/mame/layout/oberheim_dmx.lay
```diff
@@ -490,5 +502,88 @@ copyright-holders:m1macrophage
...
+
+						-- Button just pressed. Find affected fader.
+						if dn & 1 ~= 0 then
+							for i = 1, #faders do
+								if faders[i].bounds:includes(x, y) then
+									selected = i
+									break
+								end
+							end
+						end
```
cuavas, line 562: This won’t work with multiple pointing devices or a multi-touch display.  At the very least, you need to keep track of the `id` of the active pointer and reject other pointers until it’s released.  Ideally, you should be able to track multiple pointers simultaneously so multiple faders can be adjusted simultaneously with multiple fingers on a multi-touch display.

Also, you need to implement the other pointer callbacks in order to know when pointer IDs may be recycled, or when you need to abandon all pointer input due to the view being switched, the menu being displayed, etc.

cuavas: FWIW this layout shows how to track a single pointer at a time and reject additional pointers until it’s released: https://github.com/mamedev/mame/blob/master/src/mame/layout/microtouch.lay

But that’s less than ideal for systems where multi-touch control is desirable.

m1macrophage: Thanks for pointing it out. I'll take a stab at fixing it.

cuavas: I’ll probably eventually get to it myself, but I’m thin on the ground.  I still have more recompiler stuff to deal with after the release is done.

m1macrophage: I don't mind giving it a shot. It would be cool to have it work with multi-touch. I was planning on working on this script anyway, to make fine slider adjustments a bit easier. I'll try this weekend, and if I run into issues or fail, you are welcome to pick it up when you have the bandwidth.

galibert: It would be nice if you two could reach a reference implementation for that stuff, and that we turn it into a library preloaded when rendering artwork.

m1macrophage: Agreed. It would be nice to avoid the copy-pasting. I have a couple of layouts coming up that will require slider support. My plan was to have the same script work for all of them (and the d70). After 2-3 of these, it should hopefully be mature enough to turn it into a library.

cuavas: I need to add a better approach for plugin libraries at some point, too.

PR #13185, opened 2025-01-05
misc/4enraya.cpp: Demo Sound DSW updated to Unknown

UnclePetros:
SW1:2 (Demo Sound) not working as expected.
Updated to _Unknown_ with a comment note.


UnclePetros:
> Where does the table on #13176 come from? Is it from the manual for this game, or for a similar game?

It's from a one page pdf. As you can see it seems to be a copy from a fax or something similar, with the name written on with pen, so may be also a similar game (if so probably not documented in Mame).

> That table suggests this switch is part of the “No. Nanaros” setting, along with the one adjacent to it. Have you tested to see whether those two switches do in fact allow selection of 10, 16, 20 or 30 pieces?

Yes, I've answered in the last comment to the PR #13716, [here](https://github.com/mamedev/mame/pull/13176#issuecomment-2571703168).
Pieces in-game are only 16 or 30, so manual is wrong with this.

> Also, according to that table, the DIP switches are wired with switch 8 on the least significant bit and switch 1 on the most significant bit, while they’re currently the other way around in MAME. Is there any evidence that MAME is correct?

I wouldn't know. I can't see any evidence.

PR #13187, opened 2025-01-06
Improvements to Roland D-70

felipesanches:
* Chose more accurate colors for the LCD. 
* Declared more versions of the program ROMs.
* Split the ROMs to match actual chips and renamed them to indicate their corresponding IC references (IC4 for the even ROMs and IC9 for the odd ones).

Before:
d70_old_colors

After:
Screenshot From 2025-01-05 22-07-59

(@giulioz)

### DIFF
--- a/src/mame/roland/roland_d70.cpp
+++ b/src/mame/roland/roland_d70.cpp
@@ -32,13 +32,19 @@
 namespace {
 
 // unscramble address: ROM dump offset -> proper (descrambled) offset
-#define UNSCRAMBLE_ADDR_INT(_offset) \
-	bitswap<19>(_offset, 18, 17, 15, 14, 16, 12, 11, 7, 9, 13, 10, 8, 3, 2, 1, 6, 4, 5, 0)
+template <typename T>
+constexpr auto UNSCRAMBLE_ADDR_INT(T offset)
+{
+	return bitswap<19>(offset, 18, 17, 15, 14, 16, 12, 11, 7, 9, 13, 10, 8, 3, 2, 1, 6, 4, 5, 0);
+}
 // scramble address: proper offset -> ROM dump offset
-#define SCRAMBLE_ADDR_INT(_offset) \
-	bitswap<19>(_offset, 18, 17, 14, 16, 15, 9, 13, 12, 8, 10, 7, 11, 3, 1, 2, 6, 5, 4, 0)
+template <typename T>
+constexpr auto SCRAMBLE_ADDR_INT(T offset)
+{
+	return bitswap<19>(offset, 18, 17, 14, 16, 15, 9, 13, 12, 8, 10, 7, 11, 3, 1, 2, 6, 5, 4, 0);
+}
 
-#define UNSCRAMBLE_DATA(_data) bitswap<8>(_data, 1, 2, 7, 3, 5, 0, 4, 6)
+constexpr u8 UNSCRAMBLE_DATA(u8 data) { return bitswap<8>(data, 1, 2, 7, 3, 5, 0, 4, 6); }
 
 // Bitmasks for the display board interface via PORT1
 static constexpr u8 CONT_MASK = 0b10000000;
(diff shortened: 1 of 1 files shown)


giulioz (approved):
Cool! I have no merge rights but the code looks good

giulioz:
By the way: if you have further interest into emulating the D70, I got all the custom sound chips decapped and imaged.
PCM 1: https://siliconpr0n.org/archive/doku.php?id=furrtek:roland:r15229894
PCM 2: https://siliconpr0n.org/archive/doku.php?id=furrtek:roland:r15229895
TVF: https://siliconpr0n.org/archive/doku.php?id=furrtek:roland:r15239115
RCC: https://siliconpr0n.org/archive/doku.php?id=furrtek:roland:r15239126
At least something like the PCM interpolation table can be dumped

src/mame/roland/roland_d70.cpp
```diff
@@ -510,8 +513,33 @@ void roland_d70_state::descramble_rom_internal(u8 *dst, const u8 *src) {
 
 ROM_START(d70)
 	ROM_REGION(0x20000, "maincpu", 0)
-	ROM_SYSTEM_BIOS(0, "v203", "Version 2.03")
-	ROMX_LOAD("roland_d70_v110_combined.bin", 0x00000, 0x20000, CRC(52deab1e) SHA1(87d7196888edec65c9feddd16d4c715f6992abc7), ROM_BIOS(0))
```
cuavas, line 514: Why was this called “Version 2.03” to begin with?

giulioz: I think wrong copy paste from my side, sorry

src/mame/roland/roland_d70.cpp
```diff
@@ -510,13 +517,13 @@ void roland_d70_state::descramble_rom_internal(u8 *dst, const u8 *src) {
...
-	ROM_SYSTEM_BIOS( 3, "v1.12", "Version 1.12 - August 8, 1990" )
-	ROM_SYSTEM_BIOS( 4, "v1.10", "Version 1.10 - April 19, 1990" )
-	ROM_SYSTEM_BIOS( 5, "v1.00", "Version 1.00 - March 10, 1990" )
+	ROM_DEFAULT_BIOS("v119")
+	ROM_SYSTEM_BIOS( 0, "v119", "Version 1.19 - March 9, 1993" )
+	ROM_SYSTEM_BIOS( 1, "v116", "Version 1.16 - January 28, 1991" )
+	ROM_SYSTEM_BIOS( 2, "v114", "Version 1.14 - September 20, 1990" )
+	ROM_SYSTEM_BIOS( 3, "v112", "Version 1.12 - August 8, 1990" )
+	ROM_SYSTEM_BIOS( 4, "v110", "Version 1.10 - April 19, 1990" )
+	ROM_SYSTEM_BIOS( 5, "v100", "Version 1.00 - March 10, 1990" )
```
felipesanches, line 522: What's the reasoning for omitting the dots?

How does one deal with the ambiguity when some ROM eventually reaches version `v10.0`?

cuavas: It’s unnecessary clutter when the minor version is always two digits and there’s no patch version.  If they went to two-digit major versions, it would still be obvious that `v1000` is `10.00`.

For other cases, see e.g. this (major version and minor version always one digit, patch version may or may not be present):
https://github.com/mamedev/mame/blob/mame0273/src/mame/osborne/osborne1.cpp#L1069

Or this:
https://github.com/mamedev/mame/blob/mame0273/src/mame/acorn/z88.cpp#L609

Or this:
https://github.com/mamedev/mame/blob/mame0273/src/mame/alesis/alesis.cpp#L480

We don’t put punctuation in the BIOS short names unless it’s really necessary for it to work as a mnemonic.

galibert:
> PCM 2: https://siliconpr0n.org/archive/doku.php?id=furrtek:roland:r15229895

Unless I missed something, only that one has something dumpable.  It's a table that's similar but not identical to tanh.  If I were to guess, I'd say it's used to handle saturation in the simulated filter.
[c.txt](https://github.com/user-attachments/files/18339278/c.txt)

giulioz:
@galibert thank you for dumping it! I don't think it's filter related, since the filter is handled in another chip (TVF). In other similar Roland chips, the ROM stores coefficients for a 4-point interpolation algorithm, maybe it's possible that it has a similar usage. That PCM2 chip does in fact decoding and interpolation of DPCM samples.

> Unless I missed something, only that one has something dumpable.

Correct, but the others can still be traces and reverse engineered (if anybody else is interested in doing it)

PR #13191, opened 2025-01-06
Adding IBM System/23 Datamaster

RetroAND:
I have written this driver to emulate the System/23 Datamaster. It halts at test 05, so it is still not working, unfortunately. The cause of this is the implementation of the Intel 8275 (i8275.cpp), which does not implement the behaviour of the IC when not initialized. This computer relies on this behaviour in order to perform tests so it stalls. Some components and memory regions haven't still been added as the current ones were added as needed in order to keep the code clean. Basically this pull request is for allowing MAME testers to test the 8275 with the example that made it fail. Whenever the issue with such component is resolved, programation of this driver will resume.

Thank you for your time and patience.

### DIFF
--- a/hash/amiga_cd.xml
+++ b/hash/amiga_cd.xml
@@ -5,7 +5,7 @@ license:CC0-1.0
 
 http://redump.org/discs/system/acd
 -->
-<softwarelist name="amigacd" description="Commodore AmigaCD images">
+<softwarelist name="amiga_cd" description="Commodore AmigaCD images">
 
 <!-- !Games -->
 
--- a/src/devices/video/i8275.cpp
+++ b/src/devices/video/i8275.cpp
@@ -27,7 +27,6 @@
     TODO:
 
     - double spaced rows
-    - preset counters - how it affects DMA and HRTC?
 
 */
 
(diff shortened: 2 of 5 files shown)


src/mame/ta/alphatpx.cpp
```diff
@@ -5,10 +5,16 @@
     Triumph-Adler Alphatronic Px series
     ===================================
 
-    The Px series was designed by SKS (Steinmetz-Krischke Systemtechnik), like the ITT3030 and the SKS Nano,
+    The Px series was an OEM design by SKS (Steinmetz-Krischke Systemtechnik), like the ITT3030 and the SKS Nano 2500,
```
angelosa, line 8: Please make a separate PR about modifying `ta/alphatpx.cpp` so that people (presumably @startaq ) can review without depending on IBM System/23 for merging.

RetroAND: Sorry for being a noob. Please, could you indicate me how to do that? I know the bare basics of git, but not many more things.

startaq: `alphatpx.cpp` was written by @cracyc I think, I have no idea why my name is on that driver. In any case, this would need a run of `srcclean` to fix spacing before committing.

RetroAND: I managed to revert that specific commit. This is the only way I found to separate tha Alphatronic series from the System/23.

PR #13192, opened 2025-01-06
Updated Swedish translation

yeager:


yeager:
Is this the right way to commit translation updates?

PR #13196, opened 2025-01-07
bfm\bfcobra.cpp: Add attract mode rom loading to Inquizitor

Paul-Arnold:
inqvypp2 is the attract mode rom but wasn't being loaded in "user1" rom region

### DIFF
--- a/src/mame/bfm/bfcobra.cpp
+++ b/src/mame/bfm/bfcobra.cpp
@@ -2809,6 +2809,14 @@ void bfcobjam_state::bfcobjam_with_dmd(machine_config &config)
 
   Note: Two different versions of each 6809 ROM exist: standard and protocol
   It appears sets can be a combination of disk images, 6809 and Z80 ROMs!
+  
+  I have just dumped the roms from a board with the labels "95760002" and "viper".
+  The "viper" rom is the same as 9576002.bin below and 95760002 is the same
+  as inqvypp2!
+  The labels are original and don't appear to have been removed from the
+  chips before.
+  I'm unsure where 7 digit prom numbering would've come from. BFM always
+  used 8 digit numbers.
 
 ***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


src/mame/bfm/bfcobra.cpp
```diff
@@ -2809,20 +2809,35 @@ void bfcobjam_state::bfcobjam_with_dmd(machine_config &config)
...
   Note: Two different versions of each 6809 ROM exist: standard and protocol
   It appears sets can be a combination of disk images, 6809 and Z80 ROMs!
+  
+  I have just dumped the roms from a board with the labels "95760002" and "viper".
+  The "viper" rom is the same as 9576002.bin below and 95760002 is the same
+  as inqvypp2!
+  The labels are original and don't appear to have been removed from the
+  chips before.
+  I'm unsure where 7 digit prom numbering would've come from. BFM always
+  used 8 digit numbers.
```
angelosa, line 2819: "I have just" will become outdated the very moment this PR hits master branch. Also you may as well credit yourself on this block of note: the part where this very driver has "Anonymous" as copyright holder won't be helpful in the long run.

src/mame/bfm/bfcobra.cpp
```diff
@@ -2809,20 +2809,35 @@ void bfcobjam_state::bfcobjam_with_dmd(machine_config &config)
...
 	ROM_REGION( 0x20000, "altuser1", 0 )
-	ROM_LOAD( "9576028.bin", 0x10000, 0x10000, CRC(2d85682c) SHA1(baec47bff4b8beef5afbb737dc57b22bf93ebcf8) )
+	ROM_LOAD( "9576028.bin", 0x00000, 0x10000, CRC(2d85682c) SHA1(baec47bff4b8beef5afbb737dc57b22bf93ebcf8) )
+	ROM_LOAD( "inqvypp2", 0x10000, 0x010000, CRC(f9cd196c) SHA1(0ac31d87462cbee6f41e19aefe740d876910bdf5) )
 
-		// these look quite different.. (like they belong together) but booting with these gives a checksum error (banking?)
+	// inqvypp1.bin is Version 1.1 but fails the rom check so appears to be corrupt :(
+	// inqvypp2 is the attract mode rom.
 	ROM_LOAD( "inqvypp1", 0x00000, 0x010000, CRC(9bac8c6e) SHA1(15e24d60c2f3997e637694f60daa552b22628766) )
 	ROM_LOAD( "inqvypp2", 0x10000, 0x010000, CRC(f9cd196c) SHA1(0ac31d87462cbee6f41e19aefe740d876910bdf5) )
```
angelosa, line 2842: `user1` is a legacy "do not use" name, it indicates nothing about what the ROM area contains. The fact that exists an `altuser1` is possibly even worse, but anyway you don't reload the same contents twice, you rather use `ROM_COPY` once referencing a previously allocated `ROM_REGION`.

On most basic gist and following what's in this very driver that would be:

```
	ROM_REGION( <joined_length>, "altrevs", 0 )
    // other roms here
    ROM_LOAD( "inqvypp2", <linear_source_address>, 0x010000, CRC(f9cd196c) SHA1(0ac31d87462cbee6f41e19aefe740d876910bdf5) )

    ROM_REGION( 0x20000, "user1", 0 )
   	ROM_LOAD( "9576002.bin", 0x00000, 0x10000, CRC(5b8c8a04) SHA1(af5328fee79c370f45bff36f534aaf50964b6900) )
    ROM_COPY( "altrevs", <linear_source_address>, 0x10000, 0x10000 )
```

mamehaze:
the 'alt' ROM regions need splitting up into clones at some point, the issue was the romsets around were such a mess it was difficult to work out what was part of one set and what wasn't, so prematurely splitting them made no sense at all.

src/mame/bfm/bfcobra.cpp
Paul-Arnold: OK, I'll add my name to that part of the comment block.

As the anonymous copyright holder was also me I'll change that too.

Paul-Arnold:
I've now got dumps from multiple boards.
9576002.bin is definitely assigned to the wrong ROM dump in previous versions of this driver. It was assigned to the game ROM but it is actually the name of the attract mode ROM. We don't know the correct part number of this V1.2 ROM or whether this ROM or 9576068 is the latest.

inqvypp1 is Version 1.1 and is definitely corrupt.
I have a valid dump of V1.1, 95760001,bin which I'll submit.

I think that should be everything.

mamehaze:
any new files should be submitted.  files that have just been renamed / moved to different sets they don't need to be submitted.

Paul-Arnold:
Thanks for the help. The new file has been submitted so hopefully no more changes

james-wallace-ghub:
Is there anything here that's still an issue, it looks like an overall improvement to me, albeit hamstrung by the lack of disk data.

PR #13197, opened 2025-01-07
Fix infinite looping on reading corrupted OSD files.

Bob-Z:
On Linux, if an OSD file contains a 0x0. It is interpreted as a terminating null character by mbstowcs. So osd_uchar_from_osdchar returns 0. This seems legit.

core_text_file::getc is calculating a seek value from this return value (charlen) like this:

seek(std::int64_t(charlen * sizeof(default_buffer[0])) - readlen, SEEK_CUR);

Since charlen is 0, we end up having seek(-readlen) which create an infinite loop preventing MAME from starting.

It happens to me with a corrupted .ini file in the "folders" directory.


cuavas:
The proper fix is to special-case NUL in `osd_uchar_from_osdchar` – see 007948a96dc356541aa2140a23f0d228d2783c20.

PR #13198, opened 2025-01-07
Add a new War of the Bugs clone from Radar Ltd.

clawgrip:
New working clones
------------------
Rocket Convoy [Luis Arrufat]

### DIFF
--- a/src/mame/galaxian/galaxian.cpp
+++ b/src/mame/galaxian/galaxian.cpp
@@ -10732,6 +10732,22 @@ ROM_START( spactrai )
 	ROM_LOAD( "stk.bin", 0x0000, 0x0020, CRC(6a0c7d87) SHA1(140335d85c67c75b65689d4e76d29863c209cf32) )
 ROM_END
 
+// Found on a PCB from the Spanish company "Sede 3"
+ROM_START( rocktcnvy )
+	ROM_REGION( 0x10000, "maincpu", 0 )
+	ROM_LOAD( "1.bin",   0x0000, 0x1000, CRC(fbefae0a) SHA1(23eceded14286d6b61eee34819fb46318009054a) )
+	ROM_LOAD( "2.bin",   0x1000, 0x1000, CRC(311a47cb) SHA1(3f72010da0aa4d73a6160bffa7ee7dd3f944dd4b) )
+	ROM_LOAD( "3.bin",   0x2000, 0x1000, CRC(822749cb) SHA1(92e617088d462911118842f3f68b7ff8ac77fcf5) )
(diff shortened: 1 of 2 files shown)


src/mame/galaxian/galaxian.cpp
```diff
@@ -10732,6 +10732,22 @@ ROM_START( spactrai )
...
+
+	ROM_REGION( 0x1000, "gfx1", 0 )
+	ROM_LOAD( "c1.bin",  0x0000, 0x0800, CRC(a59a9f3f) SHA1(9564f1d013d566dc0b19762aec66119e2ece0b49) )
+	ROM_LOAD( "c2.bin",  0x0800, 0x0800, CRC(9b75b40a) SHA1(4ad94db3f1d6b45a3de1ed9b51d361f20c6706e4) )
+
+	ROM_REGION( 0x0020, "proms", 0 )
+	ROM_LOAD( "prom.bin", 0x0000, 0x0020, BAD_DUMP CRC(6a0c7d87) SHA1(140335d85c67c75b65689d4e76d29863c209cf32) ) // Not dumped, borrowed from spactrai
+ROM_END
 
 /*  Galaxian hardware mods for War of the Bugs warofbug/warofbugg
```
cuavas, line 10752: Are the short names in this comment still correct, or have the short names been shuffled while this comment hasn’t been updated?  It’s now immediately above the ROM definitions for `warofbug` whiles `warofbug` and `warofbugg` are earlier in the file.  The ROM definitions here aren’t in the order of the system definitions, either.

cuavas: So can we work out when the comment got out-of-sync with the code and move it so it’s immediately above the sets it applies to, and refers to the correct short names?  This kind of rot is really frustrating.

clawgrip: I'd say it's OK now... At least this section, this driver is huge

src/mame/galaxian/galaxian.cpp
```diff
@@ -10682,7 +10650,24 @@ ROM_START( omegab )
...
+8E, 74LS139 - cut pin 11 only, join the stub left ON THE CHIP to pin 13 on the chip
+
+Cut the track on the bottom of the board going to pin 21 of IC 1K (2716 EPROM).
+
+Cut the track on the bottom of the board going to pin 21 of IC 1H (2716 EPROM).
+
+Join pins 21 of IC's 1H and 1K together and connect both to +5 volts (although
+I've seen one set of docs saying to connect to pin 1 of ic 2N (7408) instead). */
+
+ROM_START( warofbug )  // Shows 20c as the base of currency. I know of no US games that used dimes - unless it's another country?
```
cuavas, line 10670: Probably an Australian set – 20c was the standard price for one game in the ’80s.

PR #13199, opened 2025-01-07
megacd: merge the three software lists into one

chungy:
@cuavas had mentioned on IRC that these lists should be merged, and I spent some hours doing just that. Was it all worth it?

No images have been added nor removed compared to the current HEAD version, but several releases have been renamed and reparented in order to merge them all into a singular "megacd" list. The `mdconsole.cpp` file has also been updated to filter the software list based on region, so "segacd" only shows American games, "megacdj" only shows Japanese games, "megacd" only shows European games.


etabeta78:
These lists were separated on purpose, because 
- if you try loading a PAL disc on any other region, the system does not display anything
- if you try loading a NA disc on any other region, the system does not display anything
- if you try loading a JPN disc on any other region, the system does not display anything

This is not the typical situation with e.g. MD/Genesis or SNES/SFC carts, where there is some interoperability, which is why we have cross-region software lists in general, but still it is the sad reality for MegaCD / SegaCD

I believe that merging them in a single list would create confusion among non-expert users, who would question why most of the titles does not load (especially when launched with -cdrm argument, which bypasses compatibility filters)

cuavas:
If you put a PAL SNES cartridge in an SuFami or vice versa, it shows nothing due to CIC lockout, but those are a single list.  If you put a Turbografx HuCard in a PC Engine it won’t work at all because it’s wired the opposite way around, but those are a single list.  Users already have to be aware of the “system parent is US but software parent is Europe or APAC” quirk.

etabeta78:
I don't think at this stage we emulate the CIC lockout in SFC/SNES, and
this is the output when running smwj on snespal...
 [image: 0000.png]

Anyway, refraining from nitpicking, what I had mentioned was the
motivation, agreed upon by the team ~10 years ago, to keep those lists
separated.
If you think it is not a good reason anymore, I have no additional
objections of course :-)

PR #13202, opened 2025-01-09
sinclair/tsconf.cpp: Added TR-DOS with virutal drives

holub:


src/mame/sinclair/tsconf.cpp
```diff
@@ -77,27 +76,32 @@ void tsconf_state::tsconf_mem(address_map &map)
 }
 
 void tsconf_state::tsconf_io(address_map &map)
+{
+	map(0x0000, 0xffff).lrw8(
+		NAME([this](offs_t offset) { return m_ioext.read_byte((m_beta->dos_io_r() << 16) | offset); }),
+		NAME([this](offs_t offset, u8 data) { m_ioext.write_byte((m_beta->dos_io_r() << 16) | offset, data); }));
+}
```
cuavas, line 83: Can you find a way to do this with a `memory_view`, rather than effectively creating an `address_map_bank_device` in all but name?

src/mame/sinclair/tsconf_beta.h
```diff
@@ -0,0 +1,72 @@
+// license:BSD-3-Clause
+// copyright-holders:Andrei I. Holub
+#ifndef MAME_SINCLAIR_TSCONF_BETA_H
+#define MAME_SINCLAIR_TSCONF_BETA_H
+
+#pragma once
+
+#include "machine/wd_fdc.h"
+#include "imagedev/floppy.h"
```
cuavas, line 9: Please sort these.

src/mame/sinclair/tsconf_beta.h
```diff
@@ -0,0 +1,72 @@
...
+	void turbo_w(int state);
+
+    static void beta_ioext(const device_finder<tsconf_beta_device, true> &finder, address_map &map);
+    void on_m1_w();
+    bool dos_r() { return m_dos; }
+    bool vdos_r() { return m_vdos; }
+    bool dos_io_r() { return m_dos || m_io_forced; }
+    void enable_w(bool state);
+    void fddvirt_w(u8 fddvirt);
+    void io_forced_w(bool io_forced);
```
cuavas, line 39: Mixed space/tab indentation.  Also, the way to install a device in an address map is with:

```c++
   map.m(taglike, FUNC(whatever_device::mapfunc);
```

Where `mapfunc` has a signature like `void mapfunc(address_map &)` – there are numerous examples.

src/mame/sinclair/tsconf_beta.h
```diff
@@ -0,0 +1,72 @@
...
+private:
+	required_device<kr1818vg93_device> m_wd179x;
+	required_device_array<floppy_connector, 4> m_floppy;
+	output_finder<4> m_floppy_led;
+	void fdc_hld_w(int state);
+	void motors_control();
+	u8 m_control;
+	bool m_motor_active;
+
+	static void floppy_formats(format_registration &fr);
```
cuavas, line 58: Does this need to be exposed here?  You may as well throw it in an anonymous namespaece in the .cpp file.

src/mame/sinclair/tsconf_beta.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+tsconf_beta_device::tsconf_beta_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
+	: device_t(mconfig, TSCONF_BETA, tag, owner, clock)
+    , m_out_dos_cb(*this)
+    , m_out_vdos_m1_cb(*this)
+	, m_wd179x(*this, "wd179x")
+	, m_floppy(*this, "wd179x:%u", 0U)
+	, m_floppy_led(*this, "fdd%u_led", 0U)
+	, m_control(0)
+	, m_motor_active(false)
```
cuavas, line 25: This file has a lot more mixed space/tab indentation.

src/mame/sinclair/tsconf.cpp
holub: That is mainly to support configuration through zxbus. I took this approach previously because zxbus is required to handle 17 bit IO where all 3 combinations are available: shadow only, only !shadow, and both.
It's not clear if such case is possible to implement using banks or memviews.

cuavas: Could you have a view with two entries, one for shadow and one for !shadow, then have devices install to one or both of the entries?  Are there actually cases where something uses the 17th bit as an address bit rather than just a select bit?

holub: Sorry, it still not clear for me how can we pull desired view in zxbus when installing device in IO space?
https://github.com/mamedev/mame/blob/c5c18833c896afedb52ab9e79a85dfe1dad1a99b/src/devices/bus/spectrum/zxbus.h#L91

holub: I've switched this PR to the views. I'll revisit zxbus mapping later as it's not critical for this work.

PR #13205, opened 2025-01-10
ussr/juku: Convert to memory_view, add mouse & fix timer

boamaod:
* Use memory_view instead of address_map_bank_device
* Add juku_mouse_device (joystick like mouse with interrupt)
* Make third PIT8253 clk2 follow the measured ~49.92Hz


src/mame/ussr/jukumouse.h
```diff
@@ -0,0 +1,30 @@
+// license: BSD-3-Clause
+// copyright-holders:Märt Põder
+
+#ifndef MAME_USSR_JUKU_H
```
plaes, line 4: `MAME_USSR_JUKUMOUSE_H` (here and end of file)

boamaod: Renamed, thank you, @plaes!

src/mame/ussr/juku.cpp
```diff
@@ -102,7 +110,11 @@ class juku_state : public driver_device
 
 private:
 	required_device<i8080a_cpu_device> m_maincpu;
-	required_device<address_map_bank_device> m_bank;
+	required_region_ptr<uint8_t> m_rom;
+	optional_region_ptr<uint8_t> m_exp;
+	required_shared_ptr<uint8_t> m_ram;
+	optional_shared_ptr_array<uint8_t, 1> m_ext;
```
cuavas, line 116: You could change this into a `memory_share_array_creator<uint8_t, 1>` and supply the size at construction to simplify things.

src/mame/ussr/juku.cpp
```diff
@@ -645,18 +639,35 @@ void juku_state::pio0_portc_w(uint8_t data)
...
+	//m_mode[2].install_ram(0x4000, 0xbfff, &m_ext[0][0x0000]);
+	// no info on programs actually using extra 32kb "memory window"
+	m_mode[2].install_writeonly(0x4000, 0xbfff, &m_ext[0][0x0000]);
+	// could also refer to usual ram (specs not clear)
+	//m_mode[2].install_writeonly(0x4000, 0xbfff, &m_ram[0x4000]);
+	m_mode[2].install_ram(0xc000, 0xd7ff, &m_ram[0xc000]);
+	m_mode[2].install_rom(0xd800, 0xffff, &m_rom[0x1800]);
+	m_mode[2].install_writeonly(0xd800, 0xffff, &m_ram[0xd800]);
+	// could as well disable the view, but for coherence
+	m_mode[3].install_ram(0x0000, 0xffff, &m_ram[0x0000]);
```
cuavas, line 666: It should be possible to do this in the address map function.  Also remember that if there’s nothing in the selected view slot, accesses will fall through to whatever’s installed underneath. I think your address map should look like:

```c++
	map(0x0000, 0xffff).ram().share(m_ram);
	map(0x0000, 0xffff).view(m_mode);
	m_mode[0](0x0000, 0x3fff).rom().region("maincpu", 0x0000);
	m_mode[1](0xd800, 0xffff).rom().region("maincpu", 0x1800);
	// optional BASIC expansion cartridge
	m_mode[2](0x4000, 0xbfff).rom().region("expcart", 0x0000);
	//m_mode[2](0x4000, 0xbfff).ram(m_ext[0]);
	// no info on programs actually using extra 32kb "memory window"
	m_mode[2](0xd800, 0xffff).rom().region("maincpu", 0x1800);
	m_mode[3]; // let everything fall through to RAM
```

Then you can avoid all this here.

Also, to allow swapping RAM and ROM cartridges, you should turn them into slot devices, but that can wait.

src/mame/ussr/juku.cpp
```diff
@@ -645,18 +639,35 @@ void juku_state::pio0_portc_w(uint8_t data)
...
+	
+	// check if mouse is plugged
+	if (m_mouse.found() && strcmp(mconfig().options().mouse_device(), "none") != 0)
+		m_maincpu->space(AS_IO).install_read_handler(0x80, 0x80, read8smo_delegate(*m_mouse, FUNC(juku_mouse_device::mouse_port_r)));
```
cuavas, line 670: If you want to make the mouse optional, you need to turn it into a slot device.  See for example the Sharp X68000 mouse, or Sun workstation mouse.  The `mouse_device` option is for setting which host input devices should be enabled when the emulated system has mouse axis inputs.  You shouldn’t be using it like this.

cuavas (changes requested):

src/mame/ussr/juku.cpp
boamaod: Technically nothing very bad happens if I just remove the mouse check and move the port allocation back to `io_map`. This will make the mouse always enabled, but on original system reading port 80 without a mouse attached stalled the system, so all the programs have to explicitly ask if the mouse is attached anyway. 

The authentic slot device would be a Multibus, which seems to be a lot of work that I would leave for next iterations.

cuavas: We do have at least some level of support for Multibus in src/devices/bus/multibus, so you hopefully won’t have to do everything from scratch for it.

PR #13207, opened 2025-01-10
debugger/qt/debuggerview.cpp: fix deprecation warning on Qt6

angelosa:
- chore: bump Linux CI to Qt6

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y libsdl2-dev libsdl2-ttf-dev libasound2-dev libxinerama-dev libxi-dev qtbase6-dev qtbase6-dev-tools
+        sudo apt-get install -y libsdl2-dev libsdl2-ttf-dev libasound2-dev libxinerama-dev libxi-dev qt6-base-dev qt6-base-dev-tools
     - name: Install clang
       if: matrix.compiler == 'clang'
       run: sudo apt-get install -y clang


angelosa:
Courtesy branch just to check it compiles everywhere:

```
src/osd/modules/debugger/qt/debuggerview.cpp: In member function ‘virtual void osd::debugger::qt::DebuggerView::mousePressEvent(QMouseEvent*)’:
src/osd/modules/debugger/qt/debuggerview.cpp:311:67: warning: ‘int QMouseEvent::x() const’ is deprecated: Use position() [-Wdeprecated-declarations]
  311 |         clickViewPosition.x = (std::min)(int(topLeft.x + (event->x() / fontWidth)), topLeft.x + visibleCharDims.x - 1);
      |                                                           ~~~~~~~~^~
In file included from /usr/include/qt6/QtGui/QKeyEvent:1,
                 from src/osd/modules/debugger/qt/debuggerview.cpp:14:
/usr/include/qt6/QtGui/qevent.h:226:16: note: declared here
  226 |     inline int x() const { return qRound(position().x()); }
      |                ^
src/osd/modules/debugger/qt/debuggerview.cpp:312:67: warning: ‘int QMouseEvent::y() const’ is deprecated: Use position() [-Wdeprecated-declarations]
  312 |         clickViewPosition.y = (std::min)(int(topLeft.y + (event->y() / fontHeight)), topLeft.y + visibleCharDims.y - 1);
      |                                                           ~~~~~~~~^~
/usr/include/qt6/QtGui/qevent.h:228:16: note: declared here
  228 |     inline int y() const { return qRound(position().y()); }
```

angelosa:
Given the Qt5 EoL in May 2025 may as well bump to Qt6 anyway. Thoughts?

rb6502:
I agree we should bump to Qt 6, but it needs to compile successfully first obviously.

cuavas:
I’m not set up to deal with Qt6 for Windows test builds, and doing so would mean bumping the compiler version etc. for releases due to the mess of dependencies in MSYS2’s MinGW.  I don’t really want to do that at the moment.

angelosa:
> I agree we should bump to Qt 6, but it needs to compile successfully first obviously.

I recognize the current error: somehow this section will go qt5 regardless of what's really installed (and locally I just populate the `QT_HOME` env)

> Package Qt5Widgets was not found in the pkg-config search path.
Perhaps you should add the directory containing `Qt5Widgets.pc'
to the PKG_CONFIG_PATH environment variable

https://github.com/mamedev/mame/blob/cd2d3f417717fba6cf92a0aa96074451afc24d2b/scripts/src/osd/modules.lua#L509-L531

> I’m not set up to deal with Qt6 for Windows test builds, and doing so would mean bumping the compiler version etc. for releases due to the mess of dependencies in MSYS2’s MinGW. I don’t really want to do that at the moment.

It's not urgent sure, it also needs Mac OS equivalent for using Qt6 anyway (and I'm not setup for that atm either).

ajrhacker:
A Qt6 fix for the MacOS build was committed in c709a0f124d6028005bae7ec7c2b1c92b46a3c3a.

angelosa:
An Arch distro I have does something dull: maps `qt5` to `/usr/bin/qmake` and `qt6` to `/usr/bin/qmake6`. Need to investigate over that front.

cuavas:
2550f57948a4637167f6124673e4e9b086eed297 should fix the deprecation warning without breaking Qt5.

angelosa:
Nothing left of value here, closing.

PR #13212, opened 2025-01-11
Adding a layout to the Roland D-70 driver.

felipesanches:
Support moving sliders with mouse click & drag.

* **Note 1:** LEDs declared and positioned in the layout, but the driver still does not control them.
* **Note 2:** The rotary encoder is also not functional yet. I'll leave that one to a future separate commit.

(@giulioz, @m1macrophage)

### DIFF
--- a/src/mame/layout/d70.lay
+++ b/src/mame/layout/d70.lay
@@ -1,0 +1,633 @@
+<?xml version="1.0"?>
+<!--
+license:CC0-1.0
+copyright-holders:Felipe Sanches, m1macrophage
+-->
+<mamelayout version="2">
+
+<!-- define elements -->
+
+	<element name="red_led" defstate="0">
+		<rect state="1">
+			<color red="1" green="0.0" blue="0.0" />
+		</rect>
+		<rect state="0">
+			<color red="0.2" green="0.0" blue="0.0" />
+		</rect>
+	</element>
+
+	<element name="cpanel">
+		<rect><color red="0.13" green="0.13" blue="0.11"/></rect>
+	</element>
+
+	<element name="clear_rect">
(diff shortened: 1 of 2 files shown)


felipesanches:
Using the internal layout (included in this PR):
Screenshot From 2025-01-10 18-23-57

Using the external layout (available at https://github.com/ArqueologiaDigital/MAME_synths_artwork/tree/main/d70):
Screenshot From 2025-01-10 18-23-27

src/mame/layout/d70.lay
```diff
@@ -0,0 +1,633 @@
...
+
+
+	<script><![CDATA[
+		file:set_resolve_tags_callback(
+			function()
+				-- These constants need to match the "slider" and "slider-knob"
+				-- element attributes in the "Mixer" section.
+				local slider_height <const> = 82
+				local knob_height <const> = 24
+				local knob_slider_delta_y <const> = 9  -- slider y - knob y
```
m1macrophage, line 557: I guess this wasn't well documented. Using the IDs in this layout , this can be set to slider-well.y - slider-knob.y, or 357-352 = 5.

This will ensure the mouse uses the entire range of the inner well.

felipesanches: thanks!

m1macrophage: Ugh, apologies. I see now that your layout setup is a bit different.

In your case, I think you need to set `slider_height` to `slider-wider-well.height`, get rid of `knob_height` and `knob_slider_delta_y`, and set `slider_deadzone` to `slider-well.y - slider-wider-well.y`. 

I don't think it will make a huge difference, but the movement and response might be a bit off otherwise.

If you are having trouble getting it perfect, I can take a look after it is submitted.

PR #13213, opened 2025-01-11
excellent/excellent_spr.cpp: Fix sprite delay, Cleanups/updates:

cam900:
excellent/excellent_spr.cpp:
- Reduce duplicates
- Remove unused variable
- Use decode_gfx for gfx decoding routine
- Fix docs

excellent/gcpinbal.cpp:
- Fix background tilemap bankswitching
- Use ROMREGION_ERASE00 for Fill empty ROM area


src/mame/excellent/gcpinbal.cpp
cuavas, line 1: This seems a bit hacky.  Should the Excellent System sprite device be using double-buffered sprite RAM and flip buffers on each frame?  Does Aquarium need sprite delay as well?

cam900: aquarium has sprite delay (tested with reference video in comment), I will move this feature into excellent/excellent_spr.cpp.

src/mame/excellent/excellent_spr.cpp
```diff
@@ -72,29 +79,35 @@ void excellent_spr_device::device_reset()
 {
 }
 
+void excellent_spr_device::vblank(int state)
+{
+	if (state)
+		std::copy_n(&m_ram[0], 0x1000, &m_ram_buffered[0]);
+}
```
cuavas, line 86: Does the hardware appear to use DMA to copy the data like this, or does it simply swap the buffer currently accessible to the CPU with the buffer that it draws from?

You may be able to work this out by looking at what the main program expects to see when reading from sprite RAM.  Does it need to see the contents of the most recent frame, or will it work if it sees the contents from the previous frame?

If it works seeing the contents from the previous frame, it’s likely just exchanging buffers, and you can implement it by exchanging a pair of pointers rather than copying from  one buffer to the other.  If it needs to see the contents of the most recent frame, it’s likely using DMA, which means the copy here is accurate.

If it’s using DMA, is the CPU halted or prevented from accessing sprite RAM during DMA?  (For example some Capcom games like Ghosts ’n Goblins block CPU bus accesses during sprite DMA, although MAME does not emulate this.)

cam900: I think it's DMA, there's 2 6116 (2kword x 8bit) RAMs (possibly sprite RAM) are mentioned in gcpinbal and aquarium PCB docs. But I don't think the CPU is halted or prevented from accessing sprite RAM during DMA.

cuavas: How would it work in that case?  It’s unlikely the RAM is clocked at double the CPU speed to allow transparent sprite RAM DMA.

PR #13216, opened 2025-01-11
kaneko/expro02.cpp: dip switches update and fix

UnclePetros:
image
According to fantasia manual ([arcade-museum](https://www.arcade-museum.com/manuals-videogames/F/Fantasia-pin--DIP.pdf)):

- Fixed Coin-2p settings by using switch 4 as "Alternate Coinage" (this affects galsnew, galsnewu, fantasia, fantasiaa, fantasiab)
- Fixed Switches 1,2 (this affects fantasia, fantasiaa, fantasiab)


src/mame/kaneko/expro02.cpp
```diff
@@ -739,8 +822,8 @@ void expro02_state::fantasia_map(address_map &map)
 {
 	expro02_video_base_map(map);
 	map(0x000000, 0x4fffff).rom();
-	map(0x800000, 0x800001).portr("DSW1");
-	map(0x800002, 0x800003).portr("DSW2");
+	map(0x800000, 0x800001).portr("DSW2");
+	map(0x800002, 0x800003).portr("DSW1");
```
cuavas, line 826: There’s no reason to reverse these.  DIP switch locations are independent of port names.

src/mame/kaneko/expro02.cpp
```diff
@@ -538,12 +542,91 @@ INPUT_PORTS_END
...
+	PORT_BIT( 0x0400, IP_ACTIVE_LOW, IPT_JOYSTICK_LEFT ) PORT_8WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,EQUALS,0x0008)
+	PORT_BIT( 0x0800, IP_ACTIVE_LOW, IPT_JOYSTICK_RIGHT ) PORT_8WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,EQUALS,0x0008)
+	PORT_BIT( 0x0100, IP_ACTIVE_LOW, IPT_JOYSTICK_UP ) PORT_4WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,NOTEQUALS,0x0008)
+	PORT_BIT( 0x0200, IP_ACTIVE_LOW, IPT_JOYSTICK_DOWN ) PORT_4WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,NOTEQUALS,0x0008)
+	PORT_BIT( 0x0400, IP_ACTIVE_LOW, IPT_JOYSTICK_LEFT ) PORT_4WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,NOTEQUALS,0x0008)
+	PORT_BIT( 0x0800, IP_ACTIVE_LOW, IPT_JOYSTICK_RIGHT ) PORT_4WAY PORT_PLAYER(2) PORT_CONDITION("DSW2",0x000c,NOTEQUALS,0x0008)
+	PORT_BIT( 0x1000, IP_ACTIVE_LOW, IPT_BUTTON1 ) PORT_PLAYER(2)
+	PORT_BIT( 0x2000, IP_ACTIVE_LOW, IPT_BUTTON2 ) PORT_PLAYER(2)   /* "Shot2" in "test mode" */
+	PORT_BIT( 0x4000, IP_ACTIVE_LOW, IPT_UNUSED )
+	PORT_BIT( 0x8000, IP_ACTIVE_LOW, IPT_UNUSED )
```
cuavas, line 583: I think all these conditions on the joystick inputs are pointless.  You may as well just use `PORT_MODIFY` to change the joystick inputs to `PORT_8WAY`.

src/mame/kaneko/expro02.cpp
```diff
@@ -538,12 +542,91 @@ INPUT_PORTS_END
...
+
+	PORT_START("DSW2")
+	PORT_DIPNAME( 0x0003, 0x0003, DEF_STR( Difficulty ) )   PORT_DIPLOCATION("SWB:1,2")
+	PORT_DIPSETTING(      0x0002, DEF_STR( Easy ) )
+	PORT_DIPSETTING(      0x0003, DEF_STR( Normal ) )
+	PORT_DIPSETTING(      0x0001, DEF_STR( Hard ) )
+	PORT_DIPSETTING(      0x0000, DEF_STR( Hardest ) )
+	PORT_DIPNAME( 0x000c, 0x0004, "Draw a diagonal" )  PORT_DIPLOCATION("SWB:3,4")
+	PORT_DIPSETTING(      0x000c, DEF_STR( Off ) )
+	PORT_DIPSETTING(      0x0008, DEF_STR( On ) ) //Other combinations are undocumented in manual
```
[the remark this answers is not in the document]
UnclePetros: Yes, I know.
But my reasoning was the fact if the manual document a "Draw a diagonal" choice, then there will be same logic on PCB allowing to enable or disable diagonal input from a 8-way Joystick.
Otherwise we have to assume manual is wrong.
Anyway, I would leave it fixed to SW3,4 (ON,OFF) until I found further info from PCB.

src/mame/kaneko/expro02.cpp
UnclePetros: Ok, I would leave it fixed to PORT_8WAY unitl we found further info from PCB.

src/mame/kaneko/expro02.cpp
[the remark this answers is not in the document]
cuavas: The manual doesn’t give a choice, though.  It only gives one option.  It doesn’t list an option for disabling the feature.

src/mame/kaneko/expro02.cpp
UnclePetros: ok

src/mame/kaneko/expro02.cpp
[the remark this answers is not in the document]
mamehaze: maybe there's an undumped set where it has an effect

for the Fantasia games which show diagonal movement in attract mode we should definitely be allowing 8-way though.

Osso13:
as far as I can see this was the only use case of set_invert_flip, so you can remove the method from the device.

cuavas (commented):
This still seems to add a lot of copy/paste in I/O port definitions.  Is it not possible to use `PORT_MODIFY`?

src/mame/kaneko/expro02.cpp
```diff
@@ -538,12 +542,34 @@ INPUT_PORTS_END
...
 
 static INPUT_PORTS_START( fantasia )
-	PORT_INCLUDE( galsnew )
+	PORT_INCLUDE( expro02 )
 
 	PORT_MODIFY("DSW1")
-	PORT_DIPNAME( 0x0080, 0x0080, DEF_STR( Unknown ) )  PORT_DIPLOCATION("SWA:8")
-	PORT_DIPSETTING(      0x0080, DEF_STR( Off ) )
+	PORT_DIPNAME( 0x000c, 0x0004, "Draw a diagonal" )  PORT_DIPLOCATION("SWA:3,4")
+	PORT_DIPSETTING(      0x0008, DEF_STR( On ) ) //Other combinations are undocumented in manual
```
angelosa: This is setting an invalid default wrt what's available. Also the convention is to add all the possible options even if manual states `<reserved>` and similiar junk.

UnclePetros: OK.
I've removed the DIP switch 0x000c (“Draw a diagonal”) definition and added two new separate switches (0x0004 and 0x0008), since the manual’s indication does not appear to be correct (as suggested by hap on Discord).
Updated the definition of PORT_DIPNAME(0x0080, 0x0080, "Unknown") to use DEF_STR(Unknown) instead of the literal string.

PR #13220, opened 2025-01-12
new tlcs900 variant: TMP94C241

felipesanches:
Used on the KN5000 driver.

### DIFF
--- a/src/devices/cpu/tlcs900/tmp94c241.cpp
+++ b/src/devices/cpu/tlcs900/tmp94c241.cpp
@@ -1393,6 +1393,18 @@ void tmp94c241_device::tlcs900_handle_timers()
 
 void tmp94c241_device::execute_set_input(int input, int level)
 {
+	auto update_int_reg = [this, level, input](uint8_t reg, uint8_t mask)
+	{
+		if (level != m_level[input])
+		{
+			m_level[input] = level;
+			if (level == ASSERT_LINE)
+				m_int_reg[reg] |= mask;
+			else
+				m_int_reg[reg] &= ~mask;
+		}
+	};
+
 	switch( input )
 	{
 		case INPUT_LINE_NMI:
@@ -1418,97 +1430,23 @@ void tmp94c241_device::execute_set_input(int input, int level)
 					m_halted = 0;
 					m_int_reg[INTE0AD] |= 0x08;
 				}
+				m_level[TLCS900_INT0] = level;
 			}
 			else
 			{
 				/* Level detect */
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE0AD] |= 0x08;
-				else
-					m_int_reg[INTE0AD] &= ~0x08;
-			}
-			m_level[TLCS900_INT0] = level;
-			break;
-
-		case TLCS900_INT4:
-			if (level != m_level[TLCS900_INT4]) {
-				m_level[TLCS900_INT4] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE45] |= 0x08;
-				else
-					m_int_reg[INTE45] &= ~0x08;
+				update_int_reg(INTE0AD, 0x08);
 			}
 			break;
 
-		case TLCS900_INT5:
-			if (level != m_level[TLCS900_INT5]) {
-				m_level[TLCS900_INT5] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE45] |= 0x80;
-				else
-					m_int_reg[INTE45] &= ~0x80;
-			}
-			break;
-
-		case TLCS900_INT6:
-			if (level != m_level[TLCS900_INT6]) {
-				m_level[TLCS900_INT6] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE67] |= 0x08;
-				else
-					m_int_reg[INTE67] &= ~0x08;
-			}
-			break;
-
-		case TLCS900_INT7:
-			if (level != m_level[TLCS900_INT7]) {
-				m_level[TLCS900_INT7] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE67] |= 0x80;
-				else
-					m_int_reg[INTE67] &= ~0x80;
-			}
-			break;
-
-		case TLCS900_INT8:
-			if (level != m_level[TLCS900_INT8]) {
-				m_level[TLCS900_INT8] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE89] |= 0x08;
-				else
-					m_int_reg[INTE89] &= ~0x08;
-			}
-			break;
-
-		case TLCS900_INT9:
-			if (level != m_level[TLCS900_INT9]) {
-				m_level[TLCS900_INT9] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTE89] |= 0x80;
-				else
-					m_int_reg[INTE89] &= ~0x80;
-			}
-			break;
-
-		case TLCS900_INTA:
-			if (level != m_level[TLCS900_INTA]) {
-				m_level[TLCS900_INTA] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTEAB] |= 0x08;
-				else
-					m_int_reg[INTEAB] &= ~0x08;
-			}
-			break;
-
-		case TLCS900_INTB:
-			if (level != m_level[TLCS900_INTB]) {
-				m_level[TLCS900_INTB] = level;
-				if (level == ASSERT_LINE)
-					m_int_reg[INTEAB] |= 0x80;
-				else
-					m_int_reg[INTEAB] &= ~0x80;
-			}
-			break;
+		case TLCS900_INT4: update_int_reg(INTE45, 0x08); break;
+		case TLCS900_INT5: update_int_reg(INTE45, 0x80); break;
+		case TLCS900_INT6: update_int_reg(INTE67, 0x08); break;
+		case TLCS900_INT7: update_int_reg(INTE67, 0x80); break;
+		case TLCS900_INT8: update_int_reg(INTE89, 0x08); break;
+		case TLCS900_INT9: update_int_reg(INTE89, 0x80); break;
+		case TLCS900_INTA: update_int_reg(INTEAB, 0x08); break;
+		case TLCS900_INTB: update_int_reg(INTEAB, 0x80); break;
 
 		default:
 			/* invalid */
--- a/src/devices/cpu/tlcs900/tmp94c241.cpp
+++ b/src/devices/cpu/tlcs900/tmp94c241.cpp
@@ -1088,29 +1088,44 @@ void tmp94c241_device::tlcs900_handle_ad()
 //  tlcs900_handle_timers -
 //-------------------------------------------------
 
+/* Prescaler shift ammounts corresponding
+    to each possible timer input clock sources: */
+static constexpr uint8_t T1 = 3;
+static constexpr uint8_t T4 = 5;
+static constexpr uint8_t T16 = 7;
+static constexpr uint8_t T256 = 11;
+
 void tmp94c241_device::tlcs900_handle_timers()
 {
 	uint32_t  old_pre = m_timer_pre;
+
+	auto update_timer_count = [this, old_pre](
+		uint8_t timer_index,
+		uint8_t input_clk_select,
+		uint8_t s1,
+		uint8_t s2,
+		uint8_t s3)
+	{
+		switch( input_clk_select )
+		{
+			case 0:
+			/* Not yet implemented.
+			    - For the 8 bit timers: TIO, TO0TRG, invalid and TO2TRG
+			    - For all 16 bit timers: TIA
+			*/
+			break;
+			case 1: m_timer_change[timer_index] += ((m_timer_pre >> s1) - (old_pre >> s1)); break;
+			case 2: m_timer_change[timer_index] += ((m_timer_pre >> s2) - (old_pre >> s2)); break;
+			case 3: m_timer_change[timer_index] += ((m_timer_pre >> s3) - (old_pre >> s3)); break;
+		}
+	};
 
 	if ( BIT(m_t16run, 7) ) /* prescaler is active */
 		m_timer_pre += m_cycles;
 
 	if ( BIT(m_t8run, 0) ) /* Timer 0 is running */
 	{
-		switch( m_t01mod & 3 ) /* T0_INPUT_CLOCK */
-		{
-			case 0:  /* TIO */
-				break;
-			case 1:  /* T1 */
-				m_timer_change[0] += ( m_timer_pre >> 3 ) - ( old_pre >> 3 );
-				break;
-			case 2:  /* T4 */
-				m_timer_change[0] += ( m_timer_pre >> 5 ) - ( old_pre >> 5 );
-				break;
-			case 3:  /* T16 */
-				m_timer_change[0] += ( m_timer_pre >> 7 ) - ( old_pre >> 7 );
-				break;
-		}
+		update_timer_count(0, m_t01mod & 3, T1, T4, T16);
 
 		for( ; m_timer_change[0] > 0; m_timer_change[0]-- )
 		{
@@ -1136,20 +1151,7 @@ void tmp94c241_device::tlcs900_handle_timers()
 
 	if ( BIT(m_t8run, 1) ) /* Timer 1 is running */
 	{
-		switch( (m_t01mod >> 2) & 3 ) /* T1_INPUT_CLOCK */
-		{
-			case 0:  /* TO0TRG */
-				break;
-			case 1:  /* T1 */
-				m_timer_change[1] += ( m_timer_pre >> 3 ) - ( old_pre >> 3 );
-				break;
-			case 2:  /* T16 */
-				m_timer_change[1] += ( m_timer_pre >> 7 ) - ( old_pre >> 7 );
-				break;
-			case 3:  /* T256 */
-				m_timer_change[1] += ( m_timer_pre >> 11 ) - ( old_pre >> 11 );
-				break;
-		}
+		update_timer_count(1, (m_t01mod >> 2) & 3, T1, T16, T256);
 
 		for( ; m_timer_change[1] > 0; m_timer_change[1]-- )
 		{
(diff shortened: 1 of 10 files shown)


felipesanches:
## Before/after comparison

On current git master, the screen gets almost completely black with just a glitch of half a dozen colored pixels at the top line:
image

With the new CPU variant implemented in this PR, we get a good-looking system message regarding control-panel MCU communications error:
Screenshot From 2025-01-12 04-01-12

felipesanches:
This PR is a followup to previous work done at #12684. I believe that I have already addressed all issues pointed out by @cuavas in his previous code review. In this PR I have included only the implementation of the new CPU variant and the small snippets of code in the KN5000 driver that use features on this variant.

The rest of the code for the kn5000 driver will be submitted as separate PRs later to make it easier to review and merge smaller chunks of code.

@wilbertpol and/or @ajrhacker, as you both are listed as authors of other variants of the TLCS900 family, could you please take a look at this PR?

felipesanches:
**Note:** The wording of the `** ERROR in CPU data transmission **` message is a bit misleading. I know it is related to the control panel MCUs due to this description in the service manual:

image

felipesanches:
To make that work, I'll need to implement UART support in the TMP94C241 CPU device and then use it to attach the MCUs (with HLE initially). But I'd like to land this PR before attempting that, as the amount of code needing review keeps growing.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1935 @@
...
+	{
+		UPCOUNTER_3 = 0;
+		TIMER_CHANGE_3 = 0;
+	}
+}
+
+void tmp94c241_device::treg01_w(offs_t offset, uint8_t data)
+{
+
+    if (BIT(0, offset))
```
wilbertpol, line 636: This should be `BIT(offset, 0))` ?

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1935 @@
...
+		break;
+	}
+
+	SFR_T02FFCR = data | 0xcc;
+}
+
+void tmp94c241_device::treg23_w(offs_t offset, uint8_t data)
+{
+
+    if (BIT(0, offset))
```
wilbertpol, line 745: This should be `BIT(offset, 0)` ?

wilbertpol:
> @wilbertpol and/or @ajrhacker, as you both are listed as authors of other variants of the TLCS900 family, could you please take a look at this PR?

I don't need to be in the copyright-holders section of the tmp94c241.cpp/h files since it is all code made by you.

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Yes, indeed. I've just pushed a commit fixing it in both places.

felipesanches:
> > @wilbertpol and/or @ajrhacker, as you both are listed as authors of other variants of the TLCS900 family, could you please take a look at this PR?
> 
> I don't need to be in the copyright-holders section of the tmp94c241.cpp/h files since it is all code made by you.

OK. I've pushed a commit removing your name from the copyright holders list of those files.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1935 @@
...
+
+enum
+{
+	FF_INVERT,
+	FF_SET,
+	FF_CLEAR,
+	FF_DONTCARE
+};
+
+void tmp94c241_device::change_timer_flipflop(uint8_t flipflop, uint8_t operation)
```
galibert, line 666: That method makes no sense.  Look at function 4 on PORT_D, INVERT clears bit 0 (never sets it), SET clears bit 1-7 if bit0=1, CLEAR sets bits 1-7 if bit0=1.  If it's the expected semantics (which I doubt), the names make no sense whatsoever.

Plus it feels like it's always use with constant parameters, which select which line to activate.  Is there a real reason to group them together rather than briging each line up at the calling point?

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1935 @@
...
+}
+
+
+//-------------------------------------------------
+//  tlcs900_check_irqs -
+//-------------------------------------------------
+
+void tmp94c241_device::tlcs900_check_irqs()
+{
+	int irq_vectors[8] = { -1, -1, -1, -1, -1, -1, -1, -1 };
```
galibert, line 1399: Please put variable declarations closer to their first use, C99 is 26 years old at this point and C++ had them even earlier :-)

The int i is particularly egregious, an index variable should escape its loop if and only if its value is used after the loop.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1935 @@
...
+	case TLCS900_INT6:
+		if (m_level[TLCS900_INT6] == CLEAR_LINE && level == ASSERT_LINE)
+		{
+			m_int_reg[INTE67] |= 0x08;
+		}
+		else if (m_level[TLCS900_INT6] == ASSERT_LINE && level == CLEAR_LINE)
+		{
+			m_int_reg[INTE67] &= ~0x08;
+		}
+		m_level[TLCS900_INT3] = level;
```
galibert, line 1865: pretty sure that's wrong

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Fixed this now.

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Fixed this now.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1798 @@
...
+		}
+	}
+}
+
+void tmp94c241_device::dmav_w(offs_t offset, uint8_t data)
+{
+	m_dma_vector[offset] = data;
+}
+
+void tmp94c241_device::b0cs_w(offs_t offset, uint8_t data)
```
galibert, line 434: Why don't you use 16-bits handlers and COMBINE_DATA ?

felipesanches: I don't know how that works. Can you help me?

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: I've just pushed a re-implementation of this method that I consider to be much better now.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1769 @@
...
+}
+
+void tmp94c241_device::tffcr_w(uint8_t data)
+{
+	change_timer_flipflop( 1, (data >> 2) & 3 );
+	change_timer_flipflop( 3, (data >> 6) & 3 );
+	m_tffcr = data | 0xcc;
+}
+
+void tmp94c241_device::treg23_w(offs_t offset, uint8_t data)
```
galibert, line 708: why isn't that two independant handlers?

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1769 @@
...
+{
+	return m_trdc;
+}
+
+void tmp94c241_device::trdc_w(uint8_t data)
+{
+	m_trdc = data;
+}
+
+void tmp94c241_device::treg45_w(offs_t offset, uint16_t data)
```
galibert, line 736: Obviously untested code is bad, mmmkay?  offset is going to be 0 or 1, never 2.

src/devices/cpu/tlcs900/tmp94c241.cpp
galibert: void tmp94c241_device::b0cs_w(offs_t offset, uint16_t data, uint16_t mem_mask)
{
    COMBINE_DATA(&m_block_cs[0]);
}

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: `map(0x000090, 0x000093).w(FUNC(tmp94c241_device::treg45_w));`

I expect offset to be between 0 and 3.

felipesanches: oh, I see! data is `uint16_t`... ouch!

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: yes, you're right! Some timers SFRs pack together data for a couple timers, but that's not this case for these ones. I'll fix this!

Screenshot From 2025-02-09 18-46-55

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: I'll make additional commits later tonight. Thanks for finding this bug!

felipesanches: I think I fixed it now.

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: I think I fixed this one as well.

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Also implemented this suggestion. Thanks.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1768 @@
...
+			    for 16-bit timers 4, 6, 8 and A
+			*/
+			m_timer16[i] = 0;
+			m_timer_change[4 + i] = 0;
+		}
+	}
+}
+
+
+uint8_t tmp94c241_device::cap45_r(offs_t offset)
```
galibert, line 872: Similarly to write, you can do:
uint16_t tmp94c241_device::cap45_r(offs_t offset)
{
    return m_t16_cap[offset+0];
}

or even a:

template<int timer> uint16_r cap_r()
{
   return m_t16_cap[timer];
}

and change the address map accordingly.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1768 @@
...
+	map(0x000082, 0x000082).rw(FUNC(tmp94c241_device::tffcr_r), FUNC(tmp94c241_device::tffcr_w));
+	map(0x000084, 0x000084).rw(FUNC(tmp94c241_device::t01mod_r), FUNC(tmp94c241_device::t01mod_w));
+	map(0x000085, 0x000085).rw(FUNC(tmp94c241_device::t23mod_r), FUNC(tmp94c241_device::t23mod_w));
+	map(0x000088, 0x000088).w(FUNC(tmp94c241_device::treg0_w));
+	map(0x000089, 0x000089).w(FUNC(tmp94c241_device::treg1_w));
+	map(0x00008a, 0x00008a).w(FUNC(tmp94c241_device::treg2_w));
+	map(0x00008b, 0x00008b).w(FUNC(tmp94c241_device::treg3_w));
+	map(0x000090, 0x000091).w(FUNC(tmp94c241_device::treg4_w));
+	map(0x000092, 0x000093).w(FUNC(tmp94c241_device::treg5_w));
+	map(0x000094, 0x000097).r(FUNC(tmp94c241_device::cap45_r));
```
galibert, line 1175: that would become:
map(0x000094, 0x000095).r(FUNC(tmp94c241_device::cap_r<0>));
map(0x000096, 0x000097).r(FUNC(tmp94c241_device::cap_r<1>));

and the following indexes for cap67_r and friends.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1768 @@
...
+	map(0x000046, 0x000046).w(FUNC(tmp94c241_device::port_cr_w<PORT_H>));
+	map(0x000047, 0x000047).w(FUNC(tmp94c241_device::port_fc_w<PORT_H>));
+	map(0x000068, 0x000068).rw(FUNC(tmp94c241_device::port_r<PORT_Z>), FUNC(tmp94c241_device::port_w<PORT_Z>));
+	map(0x00006a, 0x00006a).w(FUNC(tmp94c241_device::port_cr_w<PORT_Z>));
+	map(0x000080, 0x000080).rw(FUNC(tmp94c241_device::t8run_r), FUNC(tmp94c241_device::t8run_w));
+	map(0x000081, 0x000081).rw(FUNC(tmp94c241_device::trdc_r), FUNC(tmp94c241_device::trdc_w));
+	map(0x000082, 0x000082).rw(FUNC(tmp94c241_device::tffcr_r), FUNC(tmp94c241_device::tffcr_w));
+	map(0x000084, 0x000084).rw(FUNC(tmp94c241_device::t01mod_r), FUNC(tmp94c241_device::t01mod_w));
+	map(0x000085, 0x000085).rw(FUNC(tmp94c241_device::t23mod_r), FUNC(tmp94c241_device::t23mod_w));
+	map(0x000088, 0x000088).w(FUNC(tmp94c241_device::treg0_w));
```
galibert, line 1169: there you could have a 8-bit treg03_w and a 16-bits treg45_w both using offset to know which timer to use and remove duplication.  There probably are others too in the map.

The general idea is, if you have essentially the same action on consecutive timers and consecutive addresses, build a handler at the "natural" single timer size, and use offset to select.  If you have essentially the same action, and it's at regular intervals in the memory map and the number of entries is a power of two, using .select().  If you have essentially the same action, but it's all over the place in the memory map, use a template.

felipesanches: done

felipesanches:
I've also just rebased this with current master branch.

felipesanches:
I've been making commits here to facilitate code-reviews. But not that I've addressed all feedback, perhaps it would be useful for me to git squash it all into a single commit.

felipesanches:
I've been making commits here to facilitate code-reviews. But now that I've addressed all feedback, perhaps it would be useful for me to git squash it all into a single commit.

felipesanches:
I've squashed all 30 commits into a single one and rebased it against current MAME master branch. The original commits are still available at https://github.com/felipesanches/mame/tree/tlcs900_tmp94c241_before_squashing (just in case it can be useful for the code-reviews).

felipesanches:
Also use templates for the two serial channels (squashed-in code corresponding to https://github.com/felipesanches/mame/commit/eeca47ade44155353956c0729a0a9d26b35d75b5)

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1520 @@
...
+			new_port_value = m_port_latch[PORT_E] & 0xfe;
+			if (ff_state) new_port_value |= 0x01;
+			port_w<PORT_E>(new_port_value);
+			break;
+		case 0xa:
+			new_port_value = m_port_latch[PORT_E] & 0xef;
+			if (ff_state) new_port_value |= 0x10;
+			port_w<PORT_E>(new_port_value);
+			break;
+		}
```
cuavas, line 530: Same here – braces should be at the same level as the `switch` keyword.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1520 @@
...
+				m_check_irqs = 1;
+
+				if ( (m_tffcr & 3) == 3 ) /* "FF1 Invert Enable" && "Invert by timer 1" */
+					change_timer_flipflop( 1, FF_INVERT );
+
+				/* In 16bit timer mode also reset timer 0 */
+				if ( ((m_t01mod >> 6) & 3) == 1 ) /* TO1_OPERATING_MODE == MODE_16BIT_TIMER */
+					m_timer_8[1] = 0;
+			}
+		}
```
cuavas, line 1170: This seems to be repeated quite a bit.  Can you encapsulate at least some of it with a local lambda or something?

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -0,0 +1,1520 @@
...
+
+	case TLCS900_INT4:
+		if (level != m_level[TLCS900_INT4]) {
+			m_level[TLCS900_INT4] = level;
+			if (level == ASSERT_LINE)
+				m_int_reg[INTE45] |= 0x08;
+			else
+				m_int_reg[INTE45] &= ~0x08;
+		}
+		break;
```
cuavas, line 1441: Similarly, there are eight almost identical copies of this.  Can you make a local lamba that takes the index to `m_level`, index to `m_int_reg` and bit mask for `m_int_reg` as parameters?

src/devices/cpu/tlcs900/tmp95c061.cpp
```diff
@@ -229,6 +229,12 @@ void tmp95c061_device::device_start()
 	save_item(NAME(m_mem_start_mask));
 	save_item(NAME(m_dram_refresh));
 	save_item(NAME(m_dram_access));
+
+	m_nmi_state = CLEAR_LINE;
+	for( int i = 0; i < TLCS900_NUM_INPUTS; i++ )
+	{
+		m_level[i] = CLEAR_LINE;
+	}
```
cuavas, line 237: You really want to do this in `device_resolve_objects` to avoid annoyances when devices push output state on start.

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Thanks!

Is this better now? https://github.com/mamedev/mame/pull/13220/commits/77cb08609ab7bad59a4fbbb8fec2d773a0993395

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: Sure! Do you like it this way? https://github.com/mamedev/mame/pull/13220/commits/38610081dc0d230795c465194f4fda069ef0a7fb

src/devices/cpu/tlcs900/tmp94c241.cpp
felipesanches: OK. The remainder of the implementation of timers was generalized (at https://github.com/mamedev/mame/pull/13220/commits/9fcd79c530983d7893120408fa2e5c27257e53bf) with a couple local lambdas: One for the 8bit timers that can also optionally operate in pairs as 16 bit timers. And the other lambda is for the 16-bit only timers.

felipesanches:
@cuavas, I believe that with the few last commits pushed here, I've addressed all of the things you brought up and this PR is again ready for a review.

felipesanches:
Awesome! :-D

PR #13222, opened 2025-01-12
formats/h17disk.cpp: Add format for Heath hard-sectored formats

mgarlanger:
Add support for h17disk format needed for the Heath H17 controller.

### DIFF
--- a/src/lib/formats/h17disk.cpp
+++ b/src/lib/formats/h17disk.cpp
@@ -1,0 +1,248 @@
+// license:BSD-3-Clause
+// copyright-holders:Mark Garlanger
+/*********************************************************************
+
+Heath H17D disk image format (version 2.0.0)
+
+   Format for Heath hard-sectored 5.25" disk images.
+
+   See https://heathkit.garlanger.com/diskformats/ for more information
+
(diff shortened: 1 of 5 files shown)


src/lib/formats/h17disk.cpp
```diff
@@ -0,0 +1,246 @@
...
+			buf.clear();
+		}
+	}
+
+	return true;
+}
+
+void heath_h17d_format::fm_reverse_byte_w(std::vector<uint32_t> &buffer, uint8_t val) const
+{
+	static unsigned char lookup[16] = { 0x0, 0x8, 0x4, 0xc, 0x2, 0xa, 0x6, 0xe, 0x1, 0x9, 0x5, 0xd, 0x3, 0xb, 0x7, 0xf };
```
cuavas, line 241: Non-const static data is a bug waiting to happen…

mgarlanger: Thanks, I'll create a PR to address this.

mgarlanger: Oh, I see you already cleaned it up, thanks!

PR #13224, opened 2025-01-12
cpu/mcs51.cpp: Implement PES bit for DS5002FP

cam900:
PES bit is peripheral enable select on bytewide bus (technically it controlled by PE1-PE4 pin when PES bit enabled, each pin controls 16KByte block each.), it works same when partition mode enabled or disabled.

Reference: https://www.analog.com/media/en/technical-documentation/user-guides/secure-microcontroller-users-guide.pdf (It covers DS5001, DS5002, etc.)


angelosa:
Is any of the systems in Gaelco dir requiring this PES bit? I see stuff in `gaelco/goldart.cpp` that may require an update over that driver:

https://github.com/mamedev/mame/blob/cd2d3f417717fba6cf92a0aa96074451afc24d2b/src/mame/gaelco/goldart.cpp#L132-L139

PR #13225, opened 2025-01-12
ef9345: implement service row for the TS9347 variant

fabio-d:
With this change, the minitel2 can correctly display its status row at the top of the screen.

Screenshots before:
 

Screenshots after:
 

Corresponding picture from a real Minitel 2: http://hxc2001.free.fr/minitel/minitel2.jpg (from @jfdelnero's website).

### Explanation

The ef9345.cpp file provides support for both the EF9345 and the TS9347 video chips. Compared to the EF9345, the TS9347 has a slightly different logic for the "service row" (i.e. a text line that is conceptually separate from the main contents of the screen). In particular:
- the EF9345 lets one select the memory location for fetching data for the service row, but not where to display it (it's always displayed at the top of the screen)
- the TS9347 lets one select where to display (either at the top or at the bottom), but not where to fetch it from (it's always fetched from the beginning of the video memory block)

Before this change, the TS9347 peculiarities were not implemented. The EF9345 code path is not affected by this change.

Relevant extract from the **EF9345** datasheet (i.e. the one that is already implemented):

Relevant extract from the **TS9347** datasheet (i.e. the variant implemented by this change):

### DIFF
--- a/src/devices/video/ef9345.cpp
+++ b/src/devices/video/ef9345.cpp
@@ -378,7 +378,23 @@ void ef9345_device::zoom(uint8_t *pix, uint16_t n)
 uint16_t ef9345_device::indexblock(uint16_t x, uint16_t y)
 {
 	uint16_t i = x, j;
-	j = (y == 0) ? ((m_tgs & 0x20) >> 5) : ((m_ror & 0x1f) + y - 1);
+
+	if (m_variant == EF9345_MODE::TYPE_EF9345)
+	{
+		// On the EF9345 the service row is always displayed at the top, and
+		// it can be fetched from either Y=0 or Y=1.
+		j = (y == 0) ? ((m_tgs & 0x20) >> 5) : ((m_ror & 0x1f) + y - 1);
+	}
+	else
+	{
+		// On the TS9347 the service row is displayed either at the top or at
+		// the bottom, and it is always fetched from Y=0.
+		if (m_tgs & 1)
+			j = (y == 24) ? 0 : ((m_ror & 0x1f) + y);
+		else
+			j = (y == 0) ? 0 : ((m_ror & 0x1f) + y - 1);
+	}
+
 	j = (j > 31) ? (j - 24) : j;
 
 	//right side of a double width character


src/devices/video/ef9345.cpp
```diff
@@ -378,7 +378,23 @@ void ef9345_device::zoom(uint8_t *pix, uint16_t n)
...
+	else
+	{
+		// On the TS9347 the service row is displayed either at the top or at
+		// the bottom, and it is always fetched from Y=0.
+		if (m_tgs & 1)
+			j = (y == 24) ? 0 : ((m_ror & 0x1f) + y);
+		else
+			j = (y == 0) ? 0 : ((m_ror & 0x1f) + y - 1);
+	}
+
```
angelosa, line 397: I'm puzzled about two things in here, which aren't necessarily PR author fault:
1. using enums to determine what device should do what is bad abstraction in general, where it should really `virtual override` a getter instead, and I guess from the context this path is hot so it's a branch true/false that has performance implications already.
2. `i` and `j` are poorly named variables. What are supposed to be, latches for x and y, respectively?

fabio-d: > i and j are poorly named variables. What are supposed to be, latches for x and y, respectively?

This function maps screen coordinates of characters (x, y) into the corresponding offset (0x40 * j + i) within the current "block" of the EF9345's video memory.

The memory organization of the device is described in the datasheet in the first column of this page of the datasheet:

**Figure 8** alone was somewhat confusing to me at the beginning, but I think I understand it now, thanks to the further definition of "district" given in the right column: the device's addressable RAM is hierarchically organized as follows:
* D level: 4 districts
  * B level: containing 4 blocks each
    * Y level: containing 26 buffers each (identified by discontinuous indices 0, 1 and 8 to 31)
      * X level: containing 40 bytes each

The Y=1 buffers are special because they are only partially populated in odd blocks.

**Figure 11** later shows how to assemble a "logical address" out of these indices (D, B, Y and X):


This is where the `0x40` factor in the `return` statement comes from: it's concatenating the Y value to the X as shown in this figure.

Back to your original question about the naming of the variables: MAME's `i` and `j` are actually the `X` and `Y` of the datasheet. The screen coordinates (MAME's `x` and `y`) do not seem to have an official name in the datasheet.

If we want to match the naming in the datasheet, I guess we could rename our variables as follows:
```c
// calculate the address of the char at screen coordinates (col, row)
uint16_t ef9345_device::indexblock(uint16_t col, uint16_t row)
{
	uint16_t x = col, y;

	if (m_variant == EF9345_MODE::TYPE_EF9345)
	{
		// On the EF9345 the service row is always displayed at the top, and
		// it can be fetched from either Y=0 or Y=1.
		y = (row == 0) ? ((m_tgs & 0x20) >> 5) : ((m_ror & 0x1f) + row - 1);
	}
	else
	{
		// On the TS9347 the service row is displayed either at the top or at
		// the bottom, and it is always fetched from Y=0.
		if (m_tgs & 1)
			y = (row == 24) ? 0 : ((m_ror & 0x1f) + row);
		else
			y = (row == 0) ? 0 : ((m_ror & 0x1f) + row - 1);
	}

	...
	return 0x40 * y + x;
}
```
If it works for you, I can either do it as part of this PR or as a separate one.

fabio-d: > using enums to determine what device should do what is bad abstraction in general, where it should really virtual override a getter instead, and I guess from the context this path is hot so it's a branch true/false that has performance implications already

This function is called once per character for every generate frame (each frame consisting of 25x40 characters). I suspect it's not that hot, but I don't really have a data on the performance implications.

angelosa: I missed the first comment, but yeah, it can go as a separate PR essentially.

PR #13228, opened 2025-01-13
oberheim/dmx.cpp: AA filters, VCA as a device, optimizations, bug fix.

m1macrophage:
- Refactored the gain and decay logic into its own VCA device, to make it easy to add the anti-aliasing filters.
- Added the anti-aliasing / reconstruction filters.
- Optimization: Gain and decay-RC-constant variations computed at initialization.
- Bugfix: voices with pitch control have 1 instead of 3 decay variations. Fixes decay speed in some TOM variations.

### DIFF
--- a/src/mame/oberheim/dmx.cpp
+++ b/src/mame/oberheim/dmx.cpp
@@ -430,7 +430,8 @@ dmx_voice_card::dmx_voice_card(const machine_config &mconfig, const char *tag, d
 	, m_dac(*this, "dac_u8")
 	, m_vca(*this, "dmx_vca")
 	, m_filters(*this, "aa_sk_filter_%d", 0)
-	, m_config(dmx_voice_card_config{})
+	// Need non-zero entries for the filter for validation to pass.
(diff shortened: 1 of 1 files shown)


m1macrophage:
Got a [validation error](https://github.com/m1macrophage/mamefork/actions/runs/12742779857/job/35511478189#step:6:7), which I did not get on my setup.

Attempting to fix with deabca428f37b7864c877c9f62a978f47f63ba8f, but let me know if this is a better way.

Out of curiosity, what calls those "default" device constructors? And more importantly, are there any expectations on which methods need to work in this case?  For instance, will `device_start()`, `device_reset()` etc get called in those cases?

m1macrophage:
Thanks for the info!

I also verified `device_start()` does not get called when doing a `-validate`. And I figured out why I didn't run into this: I used `-validate obdmx`, which doesn't use the "default" constructor. But a plain `-validate` does. I'll make sure to also run that in the future.

I initialized the filter parameters to non-zero for the "unused" constructor, and verified that this passes validation. The checks here now pass as well. But let me know if there is a better way to solve this.

galibert:
That way sounds good enough to me.

PR #13237, opened 2025-01-16
Add Casiotone 8000 and related systems

devinacker:
This adds the Casiotone 8000 keyboard, FK-1 bass pedals, and (non-working) MB-1 recording/playback unit all belonging to the "Symphonytron" organ system, as well as an adapter device to make them all controllable over MIDI.

The stereo chorus is also implemented, but tends to sound noticeably aliased due to the lack of resampling on the BBD output. I don't really have the bandwidth to try fixing it at the moment though.

New working machines
--------------------
Casiotone 8000 [=CO=Windler, Devin Acker]
Casiotone FK-1 [BCM, Devin Acker]

New NOT_WORKING machines
------------------------
Casiotone MB-1 [=CO=Windler, Devin Acker]

### DIFF
--- a/src/devices/sound/flt_biquad.cpp
+++ b/src/devices/sound/flt_biquad.cpp
@@ -17,7 +17,7 @@
     * Optional direct control of the 5 normalized biquad parameters for a custom/raw parameter filter.
 
     Possibly useful features which aren't implemented because nothing uses them yet:
-    * More Sallen-Key filter variations (band-pass, high-pass)
+    * More Sallen-Key filter variations (band-pass)
 
 */
 #include "emu.h"
@@ -202,7 +202,56 @@ filter_biquad_device::biquad_params filter_biquad_device::opamp_sk_lowpass_calc(
 	return r;
 }
 
-// TODO when needed: Sallen-Key high-pass filter
+/* Setup a biquad filter structure based on a single op-amp Sallen-Key high-pass filter circuit.
+*
+*                   .----------------------------.
+*                   |                            |
+*                   Z   r1                       |
+*                   Z                            |
+*                   Z                            |
+*            c1     |   c2                |\     |
+*   In >-----||-----+---||----+--------+  | \    |
+*                             |        '--|+ \   |
+*                             Z   r2      |   >--+------> out
+*                             Z        .--|- /   |
+*                             Z        |  | /    |
+*                             |        |  |/     |
+*                            gnd       |         |
+*                                      |   r4    |
+*                                      +--ZZZZ---'
+*                                      |
+*                                      Z
+*                                      Z r3
+*                                      Z
+*                                      |
+*                                     gnd
+*/
+filter_biquad_device& filter_biquad_device::opamp_sk_highpass_setup(double r1, double r2, double r3, double r4, double c1, double c2)
+{
+	filter_biquad_device::biquad_params p = opamp_sk_highpass_calc(r1, r2, r3, r4, c1, c2);
+	return setup(p);
+}
+
+void filter_biquad_device::opamp_sk_highpass_modify(double r1, double r2, double r3, double r4, double c1, double c2)
+{
+	filter_biquad_device::biquad_params p = opamp_sk_highpass_calc(r1, r2, r3, r4, c1, c2);
+	modify(p);
+}
+
+filter_biquad_device::biquad_params filter_biquad_device::opamp_sk_highpass_calc(double r1, double r2, double r3, double r4, double c1, double c2)
+{
+	filter_biquad_device::biquad_params r;
+	if ((r1 == 0) || (r2 == 0) || (r3 == 0) || (r4 == 0) || (c1 == 0) || (c2 == 0))
+	{
+		fatalerror("filter_biquad_device::opamp_sk_highpass_calc() - no parameters can be 0; parameters were: r1: %f, r2: %f, r3: %f, r4: %f, c1: %f, c2: %f", r1, r2, r3, r4, c1, c2); /* Filter can not be setup.  Undefined results. */
+	}
+	r.type = biquad_type::HIGHPASS;
+	r.gain = 1.0 + (r4 / r3); // == (r3 + r4) / r3
+	r.fc = 1.0 / (2 * M_PI * sqrt(r1 * r2 * c1 * c2));
+	r.q = sqrt(r1 * r2 * c1 * c2) / ((r1 * c1) + (r1 * c2) + ((r2 * c2) * (1.0 - r.gain)));
+	LOGMASKED(LOG_SETUP,"filter_biquad_device::opamp_sk_highpass_calc(%f, %f, %f, %f, %f, %f) yields: fc = %f, Q = %f, gain = %f\n", r1, r2, r3, r4, c1*1000000, c2*1000000, r.fc, r.q, r.gain);
+	return r;
+}
 
 
 // Multiple-Feedback filters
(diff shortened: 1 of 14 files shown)


src/devices/sound/flt_biquad.cpp
```diff
@@ -202,7 +202,56 @@ filter_biquad_device::biquad_params filter_biquad_device::opamp_sk_lowpass_calc(
...
+	{
+		fatalerror("filter_biquad_device::opamp_sk_highpass_calc() - no parameters can be 0; parameters were: r1: %f, r2: %f, r3: %f, r4: %f, c1: %f, c2: %f", r1, r2, r3, r4, c1, c2); /* Filter can not be setup.  Undefined results. */
+	}
+	r.type = biquad_type::HIGHPASS;
+	r.gain = 1.0 + (r4 / r3); // == (r3 + r4) / r3
+	r.fc = 1.0 / (2 * M_PI * sqrt(r1 * r2 * c1 * c2));
+	r.q = sqrt(r1 * r2 * c1 * c2) / ((r1 * c1) + (r1 * c2) + ((r2 * c2) * (1.0 - r.gain)));
+	LOGMASKED(LOG_SETUP,"filter_biquad_device::opamp_sk_highpass_calc(%f, %f, %f, %f, %f, %f) yields: fc = %f, Q = %f, gain = %f\n", r1, r2, r3, r4, c1*1000000, c2*1000000, r.fc, r.q, r.gain);
+	return r;
+}
```
Lord-Nightmare, line 254: The calculations for the lowpass and highpass Sallen-Key filters are actually the same and can share a function, the only difference would be the r.type for the filter. I have a local change which does exactly this, if you'd like to rebase on top of that?

devinacker: Isn't Q supposed to be calculated differently for highpass vs. lowpass?

Lord-Nightmare: You are indeed correct, and I was wrong about that.

devinacker: Thanks for taking a look anyway, I'm definitely not an expert when it comes to (active) filters.

Lord-Nightmare: `filter_biquad_device::biquad_params filter_biquad_device::opamp_sk_lphp_calc(biquad_type type, double r1, double r2, double r3, double r4, double c1, double c2)
{
	filter_biquad_device::biquad_params r;
	if ((r1 == 0) || (r2 == 0) || (r3 == 0) || (r4 == 0) || (c1 == 0) || (c2 == 0))
	{
		fatalerror("filter_biquad_device::opamp_sk_lphp_calc() - no parameters can be 0; parameters were: r1: %f, r2: %f, r3: %f, r4: %f, c1: %f, c2: %f", r1, r2, r3, r4, c1, c2); /* Filter can not be setup.  Undefined results. */
	}
	r.type = type;
	r.gain = 1.0 + (r4 / r3); // == (r3 + r4) / r3
	r.fc = 1.0 / (2 * M_PI * sqrt(r1 * r2 * c1 * c2));
	if (type == biquad_type::LOWPASS)
		r.q = sqrt(r1 * r2 * c1 * c2) / ((r1 * c2) + (r2 * c2) + ((r2 * c1) * (1.0 - r.gain)));
	else if (type == biquad_type::HIGHPASS)
		r.q = sqrt(r1 * r2 * c1 * c2) / ((r1 * c1) + (r1 * c2) + ((r2 * c2) * (1.0 - r.gain)));
	else
		r.q = M_SQRT2/2.0; // we shouldn't get here but fail gracefully if we do.
	LOGMASKED(LOG_SETUP,"filter_biquad_device::opamp_sk_lphp_calc(%f, %f, %f, %f, %f, %f) yields: fc = %f, Q = %f, gain = %f\n", r1, r2, r3, r4, c1*1000000, c2*1000000, r.fc, r.q, r.gain);
	return r;
}`
This could work.

Lord-Nightmare: I'm doublechecking my math again, vs the ti databooks as well as the okawa-denshi filter design tools.

src/devices/sound/upd931.cpp
```diff
@@ -0,0 +1,575 @@
...
+	else if (m_addr == 0xf4)
+	{
+		// 0xf4 - filter control
+		m_filter_cb(m_data);
+	}
+	else
+	{
+	//	logerror("%02x %x\n", m_addr, m_data);
+	}
+}
```
angelosa, line 299: This is clearly an `address_map_space`

cuavas: You think it’s worth implementing `device_memory_interface` just for a write-only register space?

angelosa: Yes: what happens if HW by any chance accesses unmapped space?

angelosa: I've YOLO-ed worse conversions you know ...
```c++
void upd931_device::regs_map(address_map &map)
{
	// 0x20-27 - select voice note
	map(0x20, 0x27).lw8(NAME([this] (offs_t offset, u8 data) { m_voice[offset].m_note = data; }));
	// 0x30-37 - select voice octave
	map(0x30, 0x37).lw8(NAME([this] (offs_t offset, u8 data) { m_voice[offset].m_octave = data; }));
	// 0x40 (ct8000), 0x60 (mt65) - select waveform position
	map(0x40, 0x40).mirror(0x20).lw8(NAME([this] (offs_t offset, u8 data) { m_wave_pos = data; }));
	// 0xa7 - write to waveform
	map(0xa7, 0xa7).lw8(NAME([this] (offs_t offset, u8 data) { 
		const u8 sel = BIT(m_flags, FLAG_WAVE_SEL); 
		m_wave[sel][m_wave_pos & 0xf] = data; 
	}));
	// 0xb0-b7 - shift into flags register
	map(0xb0, 0xb7).lw8(NAME([this] (offs_t offset, u8 data) { 
		m_flags >>= 1;
		m_flags |= ((data & 1) << FLAG_WAVE_SEL);
 	}));
	// 0xb8-bf - get voice status
	map(0xb8, 0xbf).lw8(NAME([this] (offs_t offset, u8 data) { 
		// TODO: more details. ct8000 only checks if bits 1-3 are all zero or not
		if (m_voice[m_addr & 7].m_env_state == ENV_IDLE)
			m_status = 0xf;
		else
			m_status = 0;
 	}));
	// 0xc0-c3 (ct8000), 0xd0-d3 (mt65) - vibrato
	map(0xc0, 0xc3).mirror(0x10).lw8(NAME([this] (offs_t offset, u8 data) { 
		// TODO: implement this. ct8000 always writes 0 since it uses the external VCO vibrato instead
		m_vibrato = data & 3;
 	}));
	// 0xc4-c7 (ct8000), 0xd4-d7 (mt65) - sustain/reverb
	map(0xc4, 0xc7).mirror(0x10).lw8(NAME([this] (offs_t offset, u8 data) { 
		// TODO: implement this. ct8000 always writes 0 since it uses the external VCO vibrato instead
		m_sustain = data & 3;
		m_reverb = BIT(data, 1, 2) ? 1 : 0;
 	}));
	// 0xe0-e7 - note on/off
	map(0xe0, 0xe7).lw8(NAME([this] (offs_t offset, u8 data) { 
		voice_t &voice = m_voice[m_addr & 7];
		if (BIT(data, 0))
			note_on(voice);
		else
			voice.m_env_state = ENV_RELEASE;

		// mt65 turns off sustain and reverb when changing tones, ct8000 does this instead
		// (and also when playing a new note that is already playing)
		voice.m_force_release = BIT(data, 3);
	}));
	// 0xf4 - filter control
	map(0xf4, 0xf4).lw8(NAME([this] (offs_t offset, u8 data) { m_filter_cb(data); }));
}
```

PR #13238, opened 2025-01-17
alesis/midiverb.cpp: Adding non-working driver for midiverb.

m1macrophage:
Driver for MIDIverb, including interactive layout. No audio emulation yet. The UI portion of the device is fully emulated, though there isn't much to it.


galibert:
It uses an adc/dac pair I guess?

m1macrophage:
It just has a DAC, which it uses as both a DAC and ADC, by combining with a successive approximate register. I'll include more info in future PRs.

PR #13244, opened 2025-01-19
alesis/midiverb.cpp: Routed entire audio path. DSP just a passthrough.

m1macrophage:
- Routed the entire audio path.
- DSP is mostly a passthrough. Just does quantization (ADC) right now.
- Using 'samples' mechanism for audio input.
- Other style and code organization fixes.

It is now possible to "process" audio through the midiverb (see instructions near the top of the driver). There won't be any reverb, but there should still be some difference between wet and dry.


galibert:
I'm working on adding live audio input to mame, but as usual it's more complicated than it should.  Samples are an amusing stopgap.

m1macrophage:
Haha, they get the job done :). Speaks to the versatility of MAME. Aye to live audio input.

m1macrophage:
> I'm working on adding live audio input to mame [...]

Looks like it's done! 🎉

galibert:
Yep, and I'm having fun adding real-time reverb to my voice :-)

PR #13248, opened 2025-01-19
- when doing prefetch, we need to save/restore DA just as we do for r…

Elektraglide:
…egular instr fetch.

The previous code did the 68010 prefetch and just reset the m_mmu_tmp_buserror_occurred.  Althought the prefetch addr was invalidated the machine code that invoked the buserr would be unaware that this bus error was being ignored.  Copying the same code that wraps the regular fetch of saving/restoring the internal DA state means we treat the prefetch just like other fetches.

Tested on:
tek4404 - totally fixes spurios crashes
harddriv 
gauntlet

### DIFF
--- a/src/devices/cpu/m68000/m68kcpu.h
+++ b/src/devices/cpu/m68000/m68kcpu.h
@@ -604,11 +604,28 @@ inline u32 m68ki_read_imm_16()
 	result = MASK_OUT_ABOVE_16(m_pref_data);
 	m_pc += 2;
 	if (!m_mmu_tmp_buserror_occurred) {
+
+		u32 tmp_dar[16];
+		for (int i = 15; i >= 0; i--)
+		{
+			tmp_dar[i] = REG_DA()[i];
+		}
+
 		// prefetch only if no bus error occurred in opcode fetch
 		m_pref_data = m68ki_ic_readimm16(m_pc);
 		m_pref_addr = m_mmu_tmp_buserror_occurred ? ~0 : m_pc;
-		// ignore bus error on prefetch
-		m_mmu_tmp_buserror_occurred = 0;
+
+		// restore cpu address registers to value at start of instruction
+		if (m_mmu_tmp_buserror_occurred)
+		for (int i = 15; i >= 0; i--)
+		{
+			if (REG_DA()[i] != tmp_dar[i])
+			{
+				REG_DA()[i] = tmp_dar[i];
+			}
+		}
+
+
 	}
 
 	return result;


src/devices/cpu/m68000/m68kcpu.h
```diff
@@ -604,11 +604,28 @@ inline u32 m68ki_read_imm_16()
...
 		m_pref_data = m68ki_ic_readimm16(m_pc);
 		m_pref_addr = m_mmu_tmp_buserror_occurred ? ~0 : m_pc;
-		// ignore bus error on prefetch
-		m_mmu_tmp_buserror_occurred = 0;
+
+		// restore cpu address registers to value at start of instruction
+		if (m_mmu_tmp_buserror_occurred)
+		for (int i = 15; i >= 0; i--)
+		{
+			if (REG_DA()[i] != tmp_dar[i])
```
ajrhacker, line 622: Is this conditional really necessary?

rb6502:
Just as a note, this core is only used for 68020/68030/68040.  gauntlet and harddriv don't actually use it and aren't valid tests.  Something like dendego or sidebs are 68040 arcade machines with the  MMU enabled so that's a more correct smoke test here.

src/devices/cpu/m68000/m68kcpu.h
```diff
@@ -604,11 +604,28 @@ inline u32 m68ki_read_imm_16()
 	result = MASK_OUT_ABOVE_16(m_pref_data);
 	m_pc += 2;
 	if (!m_mmu_tmp_buserror_occurred) {
+
+		u32 tmp_dar[16];
+		for (int i = 15; i >= 0; i--)
+		{
+			tmp_dar[i] = REG_DA()[i];
+		}
```
cuavas, line 612: Why copy in reverse order?  Also, this file uses K&R formatting.

cuavas:
@galibert can you please look at this?

rb6502:
Musashi in general is more me than OG these days.  Just to clarify, the problem here is when there's a bus error/MMU fault on a prefetch?

pmackinlay:
The issue is that on the 68010, bus errors raised during prefetch should not be cancelled. On later variants, prefetch bus errors are suppressed until/unless the prefetched instruction is actually executed. There was discussion on discord that the register save/restore should not be required (because they would not be altered by the call to `m68ki_ic_readimm16()`, and that simply not clearing the bus error flag for 68010 should be sufficient.

rb6502:
Thanks @pmackinlay .  @Elektraglide can you try the suggested version that just doesn't clear the bus error flag and if that also works for tek4404 that would be a slam-dunk to apply.

pmackinlay:
Seems like the corrected version is part of https://github.com/mamedev/mame/pull/13264, but there are other things in that PR that are not ready for merging.

PR #13249, opened 2025-01-19
Fedora 42 build fixes

belegdol:
Fedora 42 plans to ship gcc-15 and to replace native SDL2 with sdl2-compat. These changes have exposed minor header issues which this PR addresses.

### DIFF
--- a/src/lib/util/corestr.cpp
+++ b/src/lib/util/corestr.cpp
@@ -15,6 +15,7 @@
 #include <memory>
 
 #include <cctype>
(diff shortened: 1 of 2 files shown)


belegdol:
Draft state because I still need to confirm whether the remaining build problem is the final one and whether it is going to be resolved by updating to a newer sdl2-compat snapshot.

rb6502:
That's pretty wild given SDL 3 hasn't had an official release yet, but we'll see how it goes.

PR #13253, opened 2025-01-21
zxbus more refactoring

holub:
@cuavas that's almost everything you suggested, but I need:
#13202  - to complete tsconf
~~... and work on solution for the Sprinter, because it requires setup through ISA's io_space, which atm a bit challenging for me.~~

~~PS automerge is conflicting because it contains both merged and updated commits~~

### DIFF
--- a/scripts/src/bus.lua
+++ b/scripts/src/bus.lua
@@ -1896,6 +1896,8 @@ if (BUSES["ISA"]~=null) then
 		MAME_DIR .. "src/devices/bus/isa/xsu_cards.h",
 		MAME_DIR .. "src/devices/bus/isa/xtide.cpp",
 		MAME_DIR .. "src/devices/bus/isa/xtide.h",
+		MAME_DIR .. "src/devices/bus/isa/zxbus_adapter.cpp",
+		MAME_DIR .. "src/devices/bus/isa/zxbus_adapter.h",
 	}
 end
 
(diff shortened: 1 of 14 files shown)


holub:
Given none else but Sprinter needs this, I'll create local space with desired view there and install to ISA.
This will keep zxbus implementation clean.

holub:
something among this lines 28a719e31829943d3500f6658d251becad02f9fa for Sprinter/ISA but I can't figure out proper lifecycle where I can install device to io space.

cuavas:
Can you explain the slots on the Sprinter?  It has an ISA slot and a ZX Bus slot?  Does the ISA card’s I/O space appear directly in the Z80 I/O space?  Does the ZX Bus slot share its I/O space with the ISA card?  The ISA card is going to be unaware of the state of the “shadow” line on the ZX Bus slot, right?

holub:
Hold on looking at Sprinter part. I'm adding isa_zxbus card for adapter. Will see if it works.

holub:
@cuavas when ever you have time, can you check please zxbus_adapter? Looks like all installations happen on time now but for some reason reads and writes are unmaped now.

holub:
I see `m_address_mask.m_address_mask == 0` and not sure how to fix it

holub:
ready for review. just waiting for #13253

holub:
@cuavas are we good here #13372 or anything else I can do?

PR #13255, opened 2025-01-21
emscripten: fixes to get bgfx working

algestam:
This PR adds support for the bgfx video backend when building for emscripten.

In order to get bgfx build and run with mame, the following changes has been done:
* `src/osd/modules/render/bgfx/shadermanager.cpp`: Change `make_path_string(...)` to always return the path to the `essl` shaders. This wasn't working properly since the call to the function is done before bgfx has been initialised and therefore returned the renderer type `Noop` instead (see https://bkaradzic.github.io/bgfx/bgfx.html#_CPPv4N4bgfx15getRendererTypeEv).
* `src/osd/modules/render/drawbgfx.cpp`: Change `set_platform_data(...)` to return platform data with a window handle that includes a HTML target selector for the canvas element to use (currently set to `#canvas`).
* `scripts/genie.lua`: Update emscripten compiler flags to match recent emscripten versions and set memory allocation to work with the drivers I've been testing with. Also embedded a file, `artwork/lut-default.png`, which is required for the lut and hlsl chains.

Additional changes:
* `src/osd/modules/font/font_sdl.cpp`: Change SDL_ttf include to use SDL2 since recent emscripten supports this.
* `docs/source/initialsetup/compilingmame.rst`: Update emscripten version to use.

The change has been tested by me and works well with the drivers I've tested. The performance is greatly improved when using bgfx compared to the soft renderer and many drivers that struggled to maintain a decent framerate are now running smoothly at 100%.

image

All builds has been done using emscripten 3.1.35. When trying later versions various other issues has been encountered which I haven't addressed or investigated further.

Setting this as a draft and ask kindly for other Mame developers with more insight into the OSD and build system to please take a look at this PR and suggest how these changes can be improved.

### DIFF
--- a/src/osd/modules/render/bgfx/shadermanager.cpp
+++ b/src/osd/modules/render/bgfx/shadermanager.cpp
@@ -83,6 +83,15 @@ std::string shader_manager::make_path_string(const osd_options &options, const s
 {
 	std::string shader_path(options.bgfx_path());
 	shader_path += PATH_SEPARATOR "shaders" PATH_SEPARATOR;
+
+#if defined(SDLMAME_EMSCRIPTEN)
+	// Hard-code renderer type to OpenGL ES for emscripten builds since the
+	// bgfx::getRendererType() is called here before BGFX has been
+	// initialized and therefore gives the wrong renderer type (Noop).
+	shader_path += "essl" PATH_SEPARATOR;
+	return shader_path;
+#endif
+
 	switch (bgfx::getRendererType())
 	{
 		case bgfx::RendererType::Noop:
--- a/src/osd/modules/render/drawbgfx.cpp
+++ b/src/osd/modules/render/drawbgfx.cpp
@@ -395,6 +395,17 @@ bool video_bgfx::set_platform_data(bgfx::PlatformData &platform_data, osd_window
 	platform_data.ndt = nullptr;
 	platform_data.nwh = GetOSWindow(dynamic_cast<mac_window_info const &>(window).platform_window());
 #else // defined(OSD_*)
+
+#if defined(SDLMAME_EMSCRIPTEN)
+	platform_data.ndt = nullptr;
+	platform_data.nwh = (void *)"#canvas"; // HTML5 target selector
+	platform_data.context = nullptr;
+	platform_data.backBuffer = nullptr;
+	platform_data.backBufferDS = nullptr;
+	bgfx::setPlatformData(platform_data);
+	return true;
+#endif
+
 	SDL_SysWMinfo wmi;
 	SDL_VERSION(&wmi.version);
 	if (!SDL_GetWindowWMInfo(dynamic_cast<sdl_window_info const &>(window).platform_window(), &wmi))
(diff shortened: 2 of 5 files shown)


DopefishJustin:
Looks OK functionality-wise. I was able to compile with Emscripten 3.1.35 and run with both software rendering and bgfx with shader chains. bgfx had been working previously but Emscripten and bgfx are both moving targets, so thanks for chasing the issues down!

holub:
Looks good here as well. I'm glad we can move forward from 3.1.25 (my latest emscripten version which only worked without noticeable problems with MAME)

src/osd/modules/render/bgfx/shadermanager.cpp
```diff
@@ -83,6 +83,15 @@ std::string shader_manager::make_path_string(const osd_options &options, const s
...
 	shader_path += PATH_SEPARATOR "shaders" PATH_SEPARATOR;
+
+#if defined(SDLMAME_EMSCRIPTEN)
+	// Hard-code renderer type to OpenGL ES for emscripten builds since the
+	// bgfx::getRendererType() is called here before BGFX has been
+	// initialized and therefore gives the wrong renderer type (Noop).
+	shader_path += "essl" PATH_SEPARATOR;
+	return shader_path;
+#endif
+
```
cuavas, line 94: Why doesn’t it also need a hack on macOS or Linux where DirectX 9 is unsupported?  Shouldn’t bgfx be equally uninitialised on those platforms?

algestam: I’m not sure why the behavior differs, but this could be related to platform-specific handling within the bgfx codebase or how certain APIs are abstracted across different systems. The bgfx docs states that the library (bgfx) must be initialised before calling the function. As it is now, mame will call the function before bgfx has been initialised and for emscripten the wrong rendering type is returned. After bgfx has been initialised, the function returns the correct type for emscripten.

algestam:
Thanks for testing the change out and other comments. bgfx has been working in the past but has been non-functional for a while. Changing the state of this PR to be ready for review.

algestam:
Perhaps this can be merged as-is if you guys are fine with it. Otherwise as I said, PR is ready for review.

holub:
Sorry, we are experiencing bandwidth with review authorities atm. It will take a bit time to return on track. Please be patient unless PR is verbally declined.

src/osd/modules/render/drawbgfx.cpp
```diff
@@ -395,6 +395,17 @@ bool video_bgfx::set_platform_data(bgfx::PlatformData &platform_data, osd_window
 	platform_data.ndt = nullptr;
 	platform_data.nwh = GetOSWindow(dynamic_cast<mac_window_info const &>(window).platform_window());
 #else // defined(OSD_*)
+
+#if defined(SDLMAME_EMSCRIPTEN)
```
galibert, line 399: Please go for #elif there, compilers can get cranky when you have a pile of stuff after a return and it's bad form in any case.

algestam: Thanks for the suggestion. The emscripten specific code has been updated to an #elif statement, making the code align with the approach used by other system for setting up bgfx 👍

PR #13260, opened 2025-01-22
Update actions

blu3berryys:
The actions now use the latest release. Older actions were using the main branch, since there can be any malicious code and bugs in main repository. Also I removed fetch depts from checkouts. Because its unneeded. It checkouts all branches. But there is no other branches. 

ALSO, I would highly recommend you guys to use git submodules for 3rdparty directory and setup dependabot to update them. So there can be no requirement to update them manually.

### DIFF
--- a/.github/workflows/bgfxshaders.yml
+++ b/.github/workflows/bgfxshaders.yml
@@ -3,12 +3,12 @@ name: Rebuild BGFX shaders
 on:
   push:
     paths:
-    - '.github/workflows/**'
-    - 'src/osd/modules/render/bgfx/shaders/**'
+      - '.github/workflows/**'
+      - 'src/osd/modules/render/bgfx/shaders/**'
   pull_request:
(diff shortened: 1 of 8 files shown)


cuavas:
Please revert all the formatting changes so the content of the change can be discussed.

blu3berryys:
Well I know its hard to understand the changes but it will take abit time (i'm lazy aswell lol) so can you try to review it like this? Or do you still want me to revert those?

cuavas:
Yeah, judging from the rate they’re opening pull requests like this, I think this is another “person” spamming repos to get their contribution count up.

happppp:
I think what holub meant that blu3berryys may have used AI to reformat the code, not that this is an AI bot.

holub:
I actually ment she is a bot. But maybe it worth somebody take a look at the chenges and try to make sense of them

PR #13262, opened 2025-01-22
hash/pc8801_flop.xml: babylon updates

wwark:
- Required V2 Mode set so change PC8801 to PC8801mk2SR
- Parent set the scenario disk is user disk so make user is not possible (add a note)

If I set the scenario disk on XM8, the message indicate the disk is not the correct one.
image

If I set Mode n88v1h
image

The user disk of this set seems a blank disk.

Thanks,

### DIFF
--- a/hash/pc8801_flop.xml
+++ b/hash/pc8801_flop.xml
@@ -4070,7 +4070,8 @@ A few kanji drawings in gameplay are odd to look at (cfr. signs, sports as full 
 		<!-- PC8801mk2SR -->
 		<info name="release" value="198612xx"/>
 		<info name="alt_title" value="バビロン"/>
-		<info name="usage" value="Needs BASIC V2, check Crimson usage information's to create user disk"/>
+		<info name="usage" value="Needs BASIC V2"/>
+		<!-- check Crimson usage information for instructions on how to create user disk -->
 		<!--combined image-->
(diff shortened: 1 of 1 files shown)


angelosa:
Several Xtal Soft games requires an user disk initialization. There is an how to around `crimson` entry:
```xml
        <!-- TODO: brief explaination about how to create an user disk -->
        <!-- It's Xtal Soft typical mess: -->
        <!-- boot it normally then select user disk creation item -->
        <!-- Insert scenario disk in drive 1 and a non-write protected copy of user disk in drive 2 -->
        <!-- Press space then "NOW COPYING" appears -->
        <!-- If everything goes correct then "user disk created" appears, otherwise restart from scratch -->
        <!-- Finally select new game with the newly created user disk in drive 2 -->
        <!-- NB: "scenario" disk seems to be a default copy of user disk -->
        <!--     in some Xtal games (i.e. Babylon) having the scenario disk in drive 2 -->
        <!--     is enough to make it to boot, not in Crimson case tho. -->
```
The problem here is all this wall of text won't cope well with single liner that `<info name="usage" value="<explaination here>" />` allows ...

angelosa:
Feel like this info should go in a MAMEdev wiki subsection at https://wiki.mamedev.org/index.php?title=System-Specific_Setup_and_Information, and usage to point there instead. @cuavas thoughts?

cuavas:
Eventually I need to change the software list XML schema to support this better, but I keep having to fight fires, so it gets postponed.

PR #13263, opened 2025-01-23
3rdparty/ymfm: sync to latest

angelosa:
https://github.com/aaronsgiles/ymfm/commit/d334ab1a957351ce2a0aa2777b8c704664be3189


angelosa:
This eventually goes after release. Need a base for `pc98:roliroli` detection bug later.

happppp:
ymfm_opx.h is a preliminary/placeholder file and won't compile. MAME has no need for it.

angelosa:
Ref: https://github.com/aaronsgiles/ymfm/pull/64

happppp:
Even if Aaron looks at that PR, this file is useless since there's no cpp file. So, no need to put this one on hold, I'd suggest just removing that h file.

angelosa:
Ready to merge after CI/CD passes.

PR #13264, opened 2025-01-23
Tek4404 now works

Elektraglide:
Boots to Uniflex, demos runs, compiler runs, graphics works.

- fix for ncr5385 always waiting full 250ms timeout
- fixes for tek4404 custom MMU
- adds ns32081
- adds am7990
- adds x2210
- keyboard mapping fixed
- mc68010 doesn't clear buserr on prefetch
- 4 LEDs are on CPU motherboard on real device
- palette colors as close as I can get to real display


src/devices/machine/am9513.cpp
```diff
@@ -976,7 +986,7 @@ void am9513_device::write_gate_alt(int c, bool level)
 {
 	if (bus_is_16_bit())
 	{
-		logerror("Gate %dA written when configured as DB%d\n", c + 1, c + 8);
+		LOGWARN("Gate %dA written when configured as DB%d\n", c + 1, c + 8);
```
cuavas, line 989: Seriously, please stop trying to do this.  This is the third PR you’ve tried to do this on, I’ve already explained why it’s a bad idea.

src/devices/machine/ns32081.cpp
```diff
@@ -194,6 +194,11 @@ template <typename T> void ns32081_device_base::write(T data)
 		execute();
 }
 
+// for building using MinGW on Windows, this is neccessary to avoid a unresolved link error
+template u16 ns32081_device_base::read<u16>();
+template void ns32081_device_base::write<u16>(u16 data);
```
cuavas, line 199: The comment isn’t true – I with MinGW on Windows all the time without such an error.  Aren’t these functions only called from within the translation unit where the templates are defined?

src/devices/machine/ncr5385.cpp
cuavas, line 1: Please separate this change so it can be discussed independently.

src/devices/machine/nscsi_bus.cpp
```diff
@@ -80,16 +80,16 @@ void nscsi_bus_device::regen_ctrl(int refid)
...
 							dev[i].ctrl & nscsi_device::S_ACK ? "K" : "",
 							dev[i].ctrl & nscsi_device::S_REQ ? "Q" : "",
+							dev[i].ctrl & nscsi_device::S_SEL ? "S" : "",
+							dev[i].ctrl & nscsi_device::S_BSY ? "B" : "",
 							dev[i].ctrl & nscsi_device::S_MSG ? "M" : "",
-							dev[i].ctrl & nscsi_device::S_INP ? "I" : "",
 							dev[i].ctrl & nscsi_device::S_CTL ? "C" : "",
-							dev[i].ctrl & nscsi_device::S_SEL ? "S" : "",
-							dev[i].ctrl & nscsi_device::S_BSY ? "B" : "");
+							dev[i].ctrl & nscsi_device::S_INP ? "I" : "");
```
cuavas, line 92: What’s the rationale for shuffling these.  @galibert I assume you ordered them the way you did for a reason?

src/mame/tektronix/tek410x_kbd.cpp
```diff
@@ -249,14 +249,14 @@ static INPUT_PORTS_START(tek410x_keyboard)
...
+	PORT_BIT(0x40, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_CHAR(']') PORT_CHAR('}') PORT_CODE(KEYCODE_CLOSEBRACE)
 	PORT_BIT(0x80, IP_ACTIVE_LOW, IPT_UNUSED)
 
 	PORT_START("X3")
 	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Right Shift") PORT_CODE(KEYCODE_RSHIFT)
-	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_CHAR('\\') PORT_CHAR('`') PORT_CODE(KEYCODE_OPENBRACE)
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_CHAR('\\') PORT_CHAR('`') PORT_CODE(KEYCODE_BACKSLASH)
 	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_CHAR('\'') PORT_CHAR('"') PORT_CODE(KEYCODE_QUOTE)
-	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Back Space") PORT_CHAR(0x08) PORT_CODE(KEYCODE_CLOSEBRACE)
+	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Back Space") PORT_CHAR(0x08) PORT_CODE(KEYCODE_BACKSPACE)
```
cuavas, line 259: The default assignments are supposed to keep the physical arrangement of the keys.  The keyboard doesn’t have the same arrangement as a PC keyboard.

src/devices/machine/nscsi_bus.cpp
Elektraglide: would love to know.  I made them so they match the other logs in the order they present info.  It was doing my head in having to switch where I was looking every time.

src/mame/tektronix/tek440x.cpp
```diff
@@ -57,17 +60,454 @@
 #include "machine/mos6551.h"    // debug tty
 #include "machine/ncr5385.h"
 #include "machine/ns32081.h"
+#include "machine/am79c90.h"
+#include "machine/i8255.h"
+#include "machine/x2212.h"
 #include "machine/nscsi_bus.h"
 #include "sound/sn76496.h"
```
cuavas, line 67: These should be kept sorted.

src/mame/tektronix/tek440x.cpp
```diff
@@ -57,17 +60,454 @@
...
+#define MAP_SYS_WR_ENABLE 5
+// mapcntrl result bits
+#define MAP_BLOCK_ACCESS 6
+#define MAP_CPU_WR 7
+
+#define OFF8_TO_OFF16(A)	((A)>>1)
+#define OFF16_TO_OFF8(A)	((A)<<1)
+
+#define MAXRAM 0x200000	// +1MB
+//#define MAXRAM 0x400000	// +3MB (which was never a Thing)
```
cuavas, line 93: Don’t use preprocessor macros unnecessarily.  `constexpr` is a thing.

src/mame/tektronix/tek410x_kbd.cpp
Elektraglide: Its probably me not understanding but they existing binding were totally unworkable for tek4404.  As they are, editors etc works as expected in tek4404

src/devices/machine/am9513.cpp
Elektraglide: what is a bad idea?  Using LOGWARN?

src/devices/machine/ns32081.cpp
pmackinlay: I've addressed this issue here 3706f0602cfd255854c73d5a131c1c82f77a4021. The problem was due to gcc inlining the virtual methods, while the template functions were not exported, generating a link error. The problem does not occur with MSVC or MinGW/clang on Windows.

src/mame/tektronix/tek410x_kbd.cpp
rb6502: @Elektraglide The theory in MAME is that you keep the physical arrangement accurate to the original keyboard (not a PC keyboard) and let MAME's -natural function do the remapping so what you type on the PC keyboard is what you get.  I've soured on that approach over time but I do understand why it was done that way originally.

src/devices/machine/ns32081.cpp
Elektraglide: fixed

src/devices/machine/nscsi_bus.cpp
```diff
@@ -14,8 +14,8 @@
 #define LOG_DATA        (1U << 4)
 #define LOG_DATA_SENT   (1U << 5)
 
-#define VERBOSE (LOG_UNSUPPORTED)
-//#define LOG_OUTPUT_FUNC osd_printf_info
+//#define VERBOSE (LOG_GENERAL | LOG_STATE | LOG_CONTROL | LOG_DATA)
+//#define VERBOSE (LOG_UNSUPPORTED)
```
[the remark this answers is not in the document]
Elektraglide: fixed

src/mame/tektronix/tek440x.cpp
Elektraglide: fixed

src/mame/tektronix/tek440x.cpp
Elektraglide: fixed

src/devices/machine/am9513.cpp
Elektraglide: fixed

src/devices/machine/am9513.cpp
```diff
@@ -1423,19 +1445,20 @@ u16 am9513_device::read16(offs_t offset)
...
+
+		// NB testing afterwards because this command may have been changing bus width
 		if ((data & 0x00ff) == 0x00e7)
-			logerror("8-bit data bus selected with 16-bit write\n");
+			LOGWARN("8-bit data bus selected with 16-bit write\n");
 		else if ((data & 0x00ff) == 0x00ef && !bus_is_16_bit())
-			logerror("16-bit data bus selected\n");
-
-		command_write(data & 0x00ff);
+			LOGWARN("16-bit data bus selected\n");
```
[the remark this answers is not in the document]
Elektraglide: fixed

src/devices/machine/ncr5385.cpp
Elektraglide: done

src/mame/tektronix/tek410x_kbd.cpp
happppp: There is no such enforced rule, it's up to the developer. eg. if I made a driver that has an AZERTY keyboard, I would not put key A on KEYCODE_Q, no matter cuavas or rb's preference.
Or a more common example in my case since I wrote a lot of chess computer drivers: If one key is "Move", one button is "Option", i use default keys M and O, I don't put them in physical order.

This file was by @ajrhacker if he thinks your change is ok, then go for it. If he does not, then remove the change.

PR #13268, opened 2025-01-24
machine/ncr5385.cpp: fix timeout

Elektraglide:
- dont wait full 250ms if not IDLE
- fixes very low performance seen in Tek4404
- fix suggested by @shattered

### DIFF
--- a/src/devices/machine/ncr5385.cpp
+++ b/src/devices/machine/ncr5385.cpp
@@ -179,7 +179,7 @@ void ncr5385_device::scsi_ctrl_changed()
 				ctrl & S_REQ ? " REQ" : "", ctrl & S_ACK ? " ACK" : "");
 
 		if (m_state != IDLE)
-			m_state_timer->adjust(attotime::zero);
+			m_state_timer->adjust(attotime::from_usec(40));
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


cuavas:
Can someone familiar with these SCSI chips please consider the implications of this, and suggest some systems to test with?  IIRC the 5380 is used by the Mac Plus/SE and the Sun 3 family, and the 53C80 is used by later Mac models.

ajrhacker:
While NCR did develop 5385 and 5380 at more or less the same time, they are different enough that their MAME emulations share absolutely nothing except some SCSI base classes. The few MAME systems that use 5385 are all marked MACHINE_NOT_WORKING.

rb6502:
AJR is correct.  Unlike the NCR5390/94/96 chips, the 5380 and 5385 might as well be from different companies.  The Tek driver would be the first working system in MAME with a 5385.  Obviously that doesn't mean we shouldn't be careful with the changes, but there are no working systems currently to test the changes against.

pmackinlay:
I will take a close look at these changes this week; I think I have another WIP driver which can at least smoke test.

Elektraglide:
Looking at ncr5385.cpp again, there are many places where the timer is cancelled.  Is it possible this case was simply an omission?

pmackinlay:
See https://github.com/mamedev/mame/pull/13407.

PR #13270, opened 2025-01-24
seta/simple_st0016.cpp: Remove unnecessary trampoline

cam900:
seta/st0016.cpp: Fix naming

### DIFF
--- a/src/mame/seta/simple_st0016.cpp
+++ b/src/mame/seta/simple_st0016.cpp
@@ -77,7 +77,6 @@ private:
 
 	void rom_bank_w(uint8_t data);
 
-	uint32_t screen_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect);
 	TIMER_DEVICE_CALLBACK_MEMBER(interrupt);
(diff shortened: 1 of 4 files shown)


src/mame/seta/simple_st0016.cpp
```diff
@@ -547,7 +540,7 @@ void st0016_state::st0016(machine_config &config)
 	m_screen->set_vblank_time(ATTOSECONDS_IN_USEC(0));
 	m_screen->set_size(48*8, 48*8);
 	m_screen->set_visarea(0*8, 48*8-1, 0*8, 48*8-1);
-	m_screen->set_screen_update(FUNC(st0016_state::screen_update));
+	m_screen->set_screen_update(m_maincpu, FUNC(st0016_cpu_device::update));
```
angelosa, line 543: The call signature should really be `st0016_cpu_device::screen_update` as canon with 99% of devices that uses a form of `device_video_interface` or another.
Put in another way: `st0016_cpu_device::update` is confusing in the context of a CPU ...

PR #13271, opened 2025-01-24
seta/speglsht.cpp: Cleanups:

cam900:
- Remove PLOT_PIXEL_RGB macro
- Reduce duplicates
- Reduce literal tag usage
- Remove unnecessary gfxdecode and palette (both are already handled in ST0016)
- Fix save state
- Fix spacing
- Fix naming
- Fix filename
- Fix notes


src/mame/seta/speglsht.cpp
```diff
@@ -228,68 +226,69 @@ void speglsht_state::cop_w(offs_t offset, uint32_t data, uint32_t mem_mask)
 {
 	COMBINE_DATA(&m_cop_ram[offset]);
 
-	if(m_cop_ram[offset]&0x8000) //fix (sign)
+	if (m_cop_ram[offset] & 0x8000) //fix (sign)
 	{
-		m_cop_ram[offset]|=0xffff0000;
+		m_cop_ram[offset] |= 0xffff0000;
```
angelosa, line 231: The comment is misleading, I believe it's doing a 16 to 32 sign extension?

cam900: Yes, it's sign extension bit behavior.

src/mame/seta/speglsht.cpp
```diff
@@ -376,23 +372,24 @@ void speglsht_state::machine_reset()
...
 	{
-		for(int x=0;x<512;x++)
+		for (int x = 0; x < 512; x++)
 		{
-			int tmp=dy+y*512+x;
-			PLOT_PIXEL_RGB(x-67,y-5,(m_framebuffer[tmp]>>0)&0xff,(m_framebuffer[tmp]>>8)&0xff,(m_framebuffer[tmp]>>16)&0xff);
+			int const tmp = dy + y * 512 + x;
+			PLOT_PIXEL_RGB(x - 67, y - 5, (m_framebuffer[tmp] >> 0) & 0xff, (m_framebuffer[tmp] >> 8) & 0xff, (m_framebuffer[tmp] >> 16) & 0xff);
 		}
 	}
```
angelosa, line 394: This `PLOT_PIXEL_RGB` macro has no right to live: it should respect cliprect when looping, and do pitch start base addressing before inner x instead. At bare minimum unrolling is desired.

src/mame/seta/speglsht.cpp
cuavas: It seems pretty weird – sign-extending negative numbers but not clearing the upper bits for positive numbers.  Does the game ever write meaningful data to the upper half?  Is it just 16-bit RAM on a 32-bit bus?

src/mame/seta/speglsht.cpp
```diff
@@ -228,68 +226,69 @@ void speglsht_state::cop_w(offs_t offset, uint32_t data, uint32_t mem_mask)
...
 {
-	int32_t *cop=(int32_t*)&m_cop_ram[0];
+	int32_t *cop = (int32_t*)&m_cop_ram[0];
 
 	union
 	{
 		int32_t  a;
 		uint32_t b;
-	}temp;
+	} temp;
```
cuavas, line 244: This union is unnecessary.  Just use a local `int32_t` and cast to `uint32_t` when returning it.

```c++
	int32_t result = 0;
...
			result = ...;
...
	return uint32_t(result);
}
```

src/mame/seta/speglsht.cpp
```diff
@@ -228,68 +226,69 @@ void speglsht_state::cop_w(offs_t offset, uint32_t data, uint32_t mem_mask)
...
 		}
 
-		case 0x48/4:
+		case 0x48 / 4:
 		{
-			temp.a=((cop[0x9]*cop[0x0]+cop[0xa]*cop[0x1]+cop[0xb]*cop[0x2])>>14)+cop[0xe];
+			temp.a = ((cop[0x9] * cop[0x0] + cop[0xa] * cop[0x1] + cop[0xb] * cop[0x2]) >> 14) + cop[0xe];
 			return temp.b;
 		}
 	}
```
cuavas, line 265: This looks overly convoluted.  Can you reduce these by actually using the offset?  Like:
```c++
	switch (offset)
	{
	case 0x10:
	case 0x11:
	case 0x12:
		{
			unsigned displacement = (offset & 0x03) * 3;
			result = ((cop[3 + displacement] * cop[0] + cop[4 + displacement] * cop[1] + cop[5 + displacement] * cop[2]) >> 14) + cop[0xc];
			break;
		}
	}
```

src/mame/seta/speglsht.cpp
```diff
@@ -376,23 +372,23 @@ void speglsht_state::machine_reset()
...
+		for (int x = 0; x < 512; x++)
 		{
-			int tmp=dy+y*512+x;
-			PLOT_PIXEL_RGB(x-67,y-5,(m_framebuffer[tmp]>>0)&0xff,(m_framebuffer[tmp]>>8)&0xff,(m_framebuffer[tmp]>>16)&0xff);
+			int const tmp = dy + y * 512 + x;
+			int const dstx = x - 67;
+			if (cliprect.contains(dstx, dsty))
+				dstline[dstx] = rgb_t((m_framebuffer[tmp] >> 0) & 0xff, (m_framebuffer[tmp] >> 8) & 0xff, (m_framebuffer[tmp] >> 16) & 0xff);
 		}
 	}
```
cuavas, line 393: Can you rework this to loop over the clip rectangle and calculate the source location from that instead?  Something like:
```c++
	int y = cliprect.top() + 5 + (BIT(m_videoreg, 5) ? (256 * 512) : 0);
	for (int dsty = cliprect.top(); dsty <= cliprect.bottom(); dsty++, y += 512)
	{
		uint32_t *const dstline = &bitmap.pix(dsty);
		for(int dstx = cliprect.left(); dstx <= cliprect.right(); dstx++)
		{
			uint32_t const pix = m_framebuffer[y + dstx + 67];
			dstline[dstx] = rgb_t((pix >> 0) & 0xff, (pix >> 8) & 0xff, (pix >> 16) & 0xff);
		}
	}
```

That gets the multiplications and extra tests out of the inner loop.

src/mame/seta/speglsht.cpp
cam900: I think it's (signed) 16-bit RAM, and LSB only.

PR #13274, opened 2025-01-24
hooking the sound cpu for captain lucky sigmab31.cpp

flama12333:
same sound cpu memory maps of b52. 
only the sound cpu are hooked up except the main cpu.

### DIFF
--- a/src/mame/sigma/sigmab31.cpp
+++ b/src/mame/sigma/sigmab31.cpp
@@ -107,7 +107,7 @@ public:
 	sigmab31_state(const machine_config &mconfig, device_type type, const char *tag) :
 		driver_device(mconfig, type, tag),
 		m_maincpu(*this, "maincpu"),
-		m_audiocpu(*this, "audiocpu"),
+		m_audiocpu(*this, "audiocpu")
 
 	{ }
 
--- a/src/mame/sigma/sigmab31.cpp
+++ b/src/mame/sigma/sigmab31.cpp
@@ -119,11 +119,13 @@ protected:
 
 private:
 	void main_map(address_map &map) ATTR_COLD;
-	void sound_map(address_map &map) ATTR_COLD;
+	void sound_map (address_map &map) ATTR_COLD;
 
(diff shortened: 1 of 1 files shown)


src/mame/sigma/sigmab31.cpp
```diff
@@ -174,7 +187,11 @@ void sigmab31_state::sigmab31(machine_config &config)
 {
 	MC6809(config, m_maincpu, 8_MHz_XTAL);
 	m_maincpu->set_addrmap(AS_PROGRAM, &sigmab31_state::prg_map);
+    
+	MC6809(config, m_audiocpu, XTAL(8'000'000));
+	m_audiocpu->set_addrmap(AS_PROGRAM, &sigmab52_state::sound_prog_map);
```
angelosa, line 192: This won't compile: it's `sigmab31_state::sound_prog_map`. Also it's not a Prog music genre, just call it `sound_map` instead.

src/mame/sigma/sigmab31.cpp
```diff
@@ -117,9 +119,20 @@ class sigmab31_state : public driver_device
...
+	map(0x6020, 0x6027).rw(m_6840ptm_2, FUNC(ptm6840_device::read), FUNC(ptm6840_device::write));
+//	map(0x6030, 0x6030).w(FUNC(sigmab31_state::audiocpu_irq_ack_w));
+//	map(0x6050, 0x6050).r("soundlatch", FUNC(generic_latch_8_device::read));
+	map(0x6060, 0x6061).rw("ymsnd", FUNC(ym3812_device::read), FUNC(ym3812_device::write));
+	map(0x8000, 0xffff).rom().region("audiocpu", 0);
+}
+
 
 
 void sigmab31_state::prg_map(address_map &map)
```
angelosa, line 138: Below the `prg_map` please, that should really be renamed to `main_map` or something ...

src/mame/sigma/sigmab31.cpp
flama12333: I did not realize

PR #13276, opened 2025-01-24
Add a skeleton driver for VTech Genius Color Pocket console

clawgrip:
VTech Genio Color Pocket (Spanish) [jordigahan, ClawGrip]

### DIFF
--- a/src/mame/vtech/geniuscolor.cpp
+++ b/src/mame/vtech/geniuscolor.cpp
@@ -6,7 +6,7 @@
 
     Skeleton driver for VTech Genius Color Pocket / Super Color Pocket / Genio Color Pocket.
 
-    VTech 35-140500-100 203 PCB with MX25L3206E and N25S10 serial ROMs on one side and two globs on the other.
+    VTech 35-140500-100-203 PCB with MX25L3206E and N25S10 serial ROMs on one side and two globs on the other.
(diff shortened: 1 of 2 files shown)


src/mame/vtech/geniuscolor.cpp
```diff
@@ -0,0 +1,80 @@
...
+};
+
+uint32_t geniuscolor_state::screen_update_geniuscolor(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect)
+{
+	return 0;
+}
+
+// 45-keys foldable keyboard, 8 activity buttons, two direction keys (right, left) and home, OK and power buttons.
+INPUT_PORTS_START( geniuscolor )
+INPUT_PORTS_END
```
cuavas, line 47: Is the keyboard foldable, or does it slide behind the screen on these?  Pictures look like it’s a “slider” form factor like some smart phones and PDAs used.

clawgrip: Yes! Not foldable, but "slider" type. I've changed the text.

Have you seen the ROM? Any idea about the CPU or its encoding?

Thanks!

cuavas: Sorry, I haven’t looked at the ROM dump.

PR #13280, opened 2025-01-26
amiga_demos.xml: Added 2 not-working entries

ICEknigh7:
amiga_demos: Interference, Puggs In Space [pouet.net]

### DIFF
--- a/hash/amiga_demos.xml
+++ b/hash/amiga_demos.xml
@@ -27,6 +27,20 @@ TODO:
 		</part>
 	</software>
 
+	<software name="interfer">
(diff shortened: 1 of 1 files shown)


angelosa:
> Best thing at the moment is collecting .adf in `amiga_demos.xml`
Eventually those needs to run side by side, before even attempting a QA

That implied: anything that gets added here is `supported="no"` by default, when I get the chance to put together 2 lines for a bash script for the side by side then and only then the status will be updated accordingly.

ICEknigh7:
`supported="no"` it is, then.

PR #13282, opened 2025-01-26
Minimize tools-only build

skrinakron:
Currently building with `EMULATOR=0` and `TOOLS=1` entails processing layouts and translations, building libraries like BGFX and SoftFloat and dependencies like SDL_ttf, none of which appear to be necessary to build only the tools. The changes proposed here would omit the extra dependencies and avoid the unnecessary load in tools-only builds (without affecting emulator builds), leaving formats, netlists and disassemblers as the only really time-consuming parts. As an additional benefit this would make the tools much easier to package separately from the emulator for source-based Linux distributions like Gentoo.

### DIFF
--- a/scripts/src/cpu.lua
+++ b/scripts/src/cpu.lua
@@ -9,6 +9,12 @@
 --
 ---------------------------------------------------------------------------
 
+function cpufiles(filepaths)
+	for _, filepath in ipairs(filepaths) do
+		table.insert(cpu_core_files, filepath)
+	end
+end
+
 --------------------------------------------------
 -- Dynamic recompiler objects
 --------------------------------------------------
@@ -24,7 +30,7 @@ end
 
 
 if (CPU_INCLUDE_DRC) then
-	files {
+	cpufiles {
 		MAME_DIR .. "src/devices/cpu/drcbec.cpp",
 		MAME_DIR .. "src/devices/cpu/drcbec.h",
 		MAME_DIR .. "src/devices/cpu/drcbeut.cpp",
@@ -42,7 +48,7 @@ if (CPU_INCLUDE_DRC) then
 		MAME_DIR .. "src/devices/cpu/drcumlsh.h",
 	}
 	if not _OPTIONS["FORCE_DRC_C_BACKEND"] then
-		files {
+		cpufiles {
 			MAME_DIR .. "src/devices/cpu/drcbearm64.cpp",
 			MAME_DIR .. "src/devices/cpu/drcbearm64.h",
 			MAME_DIR .. "src/devices/cpu/drcbex64.cpp",
@@ -67,7 +73,7 @@ end
 --------------------------------------------------
 
 if CPUS["8X300"] then
-	files {
+	cpufiles {
 		MAME_DIR .. "src/devices/cpu/8x300/8x300.cpp",
 		MAME_DIR .. "src/devices/cpu/8x300/8x300.h",
 	}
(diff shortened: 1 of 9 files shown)


holub:
What is a real necessary to build just tools? Does adding all this "if not" conditions worth it?
I think adding dummy cpp and use it in SOURCES can make build relatively quick.

skrinakron:
Building only the tools and not the emulator was enabled in d5340b8; this PR would just skip more emulator-specific things not considered in that change.

My current use case for this kind of build is to get chdman for use with MiSTer, where I have no need for MAME itself.

crashGG:
Yes, it is very helpful for people who need to quickly build mame tool and test it.

scripts/src/osd/modules.lua
```diff
@@ -35,6 +35,20 @@ function addoptionsfromstring(str)
...
+		   (filepath:find("none%.cpp$") or
+		    filepath:find("_module%.h$") or
+		    filepath:find("/lib/[^/]+$") or
+		    filepath:find("/monitor/[^/]+$")) then
+			files {
+				filepath
+			}
+		end
+	end
+end
```
cuavas, line 50: This is really too gross.

scripts/src/cpu.lua
cuavas, line 1: Won’t nothing depend on `liboptional.a` if the emulator isn’t being built anyway?

scripts/src/osd/modules.lua
skrinakron: It seemed this small localized bit of grossness was better than either of the alternatives I contemplated:
1. Defining `NO_MODULE_IMPLS` for the preprocessor and ifdefing things out at the source code level (like
   is done with e.g. `NO_USE_MIDI`). This would smear all over the OSD module sources something that feels
   more like a build-level thing, namely which modules to enable for a specific target.
2. Using `if` blocks or a separate `files` function for module files to include only for the emulator,
   based on the `with-emulator` Genie option. This would separate the otherwise neatly bundled module file
   lines and require you to know which blocks to place new modules as they are added.

On further reflection the `if` route of alternative 2 may be cleaner after all. Would that be more appealing?

scripts/src/cpu.lua
skrinakron: Indeed, without the emulator liboptional isn't needed, but currently it still gets built anyway. Also due to how the `CPUS["I386"]` block is positioned (*after* the corresponding `opt_tool` lines rather than before like with every other CPU), the athlon and i386 CPUs are always built when `TOOLS=1`, which in turn requires the otherwise unnecessary SoftFloat; flipping the order to eliminate these CPUs causes linker errors for liboptional, which appears to mean some CPU must be built for liboptional to work out. This is where I gave up and went for the more drastic solution of eliminating liboptional.

It would of course be ideal if projects simply were only built if something depends on them – that might also eliminate the conditionals I added in 3rdparty.lua –, but in my (admittedly limited) spelunkings I didn't find how to do this. Is there a way?

skrinakron: The operational details of Genie aren't well documented, so I did some closer reading of its source and the corresponding Premake documentation. For what it's worth, I couldn't find any functionality that would omit or filter out a project from the generated solution (in my case the makefiles) once it's been added with `project` in Lua. The only way to skip liboptional, then, appears to be to not do the `project` to begin with, which is slightly complicated by cpu.lua listing files for both the CPU cores (for `project ("optional")` needed by `EMULATOR=1`) and the disassemblers (for `project ("dasm")` needed by `TOOLS=1`).

PR #13283, opened 2025-01-27
alesis/midiverb.cpp: DSP emulation. Transitioned to working.

m1macrophage:


galibert:
Ideally, the dsp should be a cpu, with disassembler and execution interface.  It would be really nice to see the effects running in the debugger.

m1macrophage:
Agreed!

Interested in adding this to my backlog. On one hand, it might be easy due to the minimal instruction set. On the other hand, there are some quirks that might make it less straightforward: instructions in the ROM don't match the instructions being executed, and neither do their addresses. Also, specific PC values change the effect of instructions.

For cases like this, where there isn't a CPU IC per se, and the circuit is custom to one device, should the implementation reside alongside the driver? Or should it be located under `src/device/cpu`?

Regardless, leaving this PR open for your consideration. At the very least, it is a working, accurate reference to the discrete logic circuit that makes up the DSP.

galibert:
For the dsp subpart of the swp30 I'm putting it into devices/audio, at least for now.  But it's mostly because there's no ROM (it's all uploaded from the main cpu and changes dynamically) so there's no real sense in having it in unidasm.  if you want it in unidasm I think so building sake it should be in devices/cpu.

src/mame/alesis/midiverb.cpp
```diff
@@ -123,39 +137,169 @@ void midiverb_dsp::device_start()
...
+	save_item(NAME(m_ram_offset));
+	save_item(NAME(m_ram));
+
+	// There seem to be 2 different ROM sizes. An 27128 and an 27256. When the
+	// 27256 is installed, only its upper half is accessible.
+	assert(m_microcode->bytes() == 0x8000 || m_microcode->bytes() == 0x4000);
+	if (m_microcode->bytes() == 0x8000)
+		m_rom_base = 0x4000;
+	else
+		m_rom_base = 0;
```
galibert, line 151: A detail that changes the expected implementation:  does the chip have 15 address lines but the top on is always 1, or does it have 14 lines and the top address line of the rom is tied to +5V?

Because if it's the latter we'd rather have a rom region length of 0x4000 and ROM_IGNORE the first half (and no special code in the device, because the device has no say).

galibert: Not that I'm not sure we can have an IGNORE at the start, but it can always be done with a ROM_CONTINUE over the first half.

src/mame/alesis/midiverb.cpp
```diff
@@ -123,39 +137,169 @@ void midiverb_dsp::device_start()
...
+			if (ld_dac)
+			{
+				if (dac_left)
+					LOG_DSP(" [DAC LEFT]");
+				else
+					LOG_DSP(" [DAC RIGHT]");
+			}
+			LOG_DSP("\n");
+
+			m_ram_offset = (m_ram_offset + ram_offset_delta) % m_ram.size();
```
galibert, line 271: There's usually no modulo in hardware.  Is the ram size a power of two?  Then it should be & (m_ram.size() - 1).

src/mame/alesis/midiverb.cpp
m1macrophage: Yeah, that's cleaner. It was the latter.

src/mame/alesis/midiverb.cpp
m1macrophage: Fixed, along with another instance. The actual hardware is just an adder that overflows and wraps around.

PR #13286, opened 2025-01-27
Fixes for MC68EZ328

isotherm:
In the course of fleshing out the AlphaSmart 3000 emulation, I've made these two changes to more accurately emulate the underlying hardware. They are separate commits to help in case of future regression testing.

- The Port A change should have no impact on existing emulation, as none of the machines (cowtipping, iqunlim, clie_db, palm) use port A.
- The timer change should be functionally equivalent, though the timer tick callback will be called more often. It might affect emulation speed on very low power host systems.

### DIFF
--- a/src/devices/machine/mc68328.cpp
+++ b/src/devices/machine/mc68328.cpp
@@ -972,6 +972,15 @@ u16 mc68328_device::csd_lsw_r() // 0x11e, 0x12e, 0x13e, 0x14e
 //  MMU/chip-select hardware - EZ variant
 //-------------------------------------------------
 
+void mc68ez328_device::scr_w(u8 data)
+{
+	if (data & SCR_WDTH8)
+	{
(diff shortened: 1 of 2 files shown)


cuavas:
You’re better off updating the timer register value when it’s read rather than using a high-frequency MAME timer.

isotherm:
> You’re better off updating the timer register value when it’s read rather than using a high-frequency MAME timer.

@cuavas, I made the change this way because it more accurately emulates the timer. In particular, the previous code did not account for cases where the clock and/or compare target are changed while the timer is enabled. I compared to the [68340 timer module](https://github.com/mamedev/mame/blob/master/src/devices/machine/68340tmu.cpp) which uses a similar implementation.

However, the emulation is still not perfect, so if you think the speed is more important than accuracy, I'm ok reverting and adding only the change to calculate the timer value when read.

PR #13289, opened 2025-01-27
formats/tzx_cas.cpp: Cut data size in case header requesting more than available (MT8952)

holub:
Images reported in MT8952 have wrong headers and requesting more than available. Added guards for segfault prevention.


holub:
Wow @ajrhacker, that's the fastest merged PR

happppp:
Too fast to spot your lenght typo ;)

holub:
That's considered as an accent.
But thing are getting worse as I blindly changed some stuff and didn't perform full build.

cuavas:
Looks like this doesn’t build at all.  Did you actually do a full build or at least a tools build for the API change?

PR #13290, opened 2025-01-27
tidy: fix build after #13289

holub:
sorry about that!

### DIFF
--- a/src/lib/formats/oric_tap.cpp
+++ b/src/lib/formats/oric_tap.cpp
@@ -348,7 +348,7 @@ static int oric_cassette_calculate_size_in_samples(const uint8_t *bytes, int len
 static int oric_cassette_fill_wave(int16_t *buffer, int length, const uint8_t *bytes)
 {
 	unsigned char header[9];
-	uint8_t *data_ptr;
+	const uint8_t *data_ptr = bytes;
 	int16_t *p;
 	int i;
 	uint8_t data;
@@ -373,7 +373,6 @@ static int oric_cassette_fill_wave(int16_t *buffer, int length, const uint8_t *b
 	length = length - ORIC_WAVESAMPLES_TRAILER;
 
 	oric.cassette_state = ORIC_CASSETTE_SEARCHING_FOR_SYNC_BYTE;
-	data_ptr = bytes;
 
 	while ((data_ptr<(bytes + oric.tap_size)) && (p < (buffer+length)) )
 	{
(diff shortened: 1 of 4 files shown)


src/lib/formats/oric_tap.cpp
```diff
@@ -348,7 +348,7 @@ static int oric_cassette_calculate_size_in_samples(const uint8_t *bytes, int len
 static int oric_cassette_fill_wave(int16_t *buffer, int length, const uint8_t *bytes)
 {
 	unsigned char header[9];
-	uint8_t *data_ptr;
+	const uint8_t *data_ptr = bytes;
 	int16_t *p;
 	int i;
```
cuavas, line 353: `data_ptr` doesn’t need to be declared here, it can go inside the `else`.

Do you want to fix the screwed up indentation in this function while you’re here?

src/lib/formats/uef_cas.cpp
```diff
@@ -249,29 +249,31 @@ static int16_t* uef_cas_fill_bit( uint8_t loops, int16_t *buffer, bool bit )
 
 static int uef_cas_fill_wave( int16_t *buffer, int length, const uint8_t *bytes )
 {
+	const uint8_t *b = bytes;
```
cuavas, line 252: Isn’t this unnecessary?  `bytes` is `uint8_t const *`, not `uint8_t const *const` so you can change the value of the pointer, just not what it points to.

src/lib/formats/uef_cas.cpp
```diff
@@ -249,29 +249,31 @@ static int16_t* uef_cas_fill_bit( uint8_t loops, int16_t *buffer, bool bit )
...
 	if ( bytes[0] == 0x1f && bytes[1] == 0x8b ) {
 		if ( gz_ptr == nullptr ) {
 			return 1;
 		}
-		bytes = gz_ptr;
+		b = gz_ptr;
```
holub, line 257: That's to support this. Or am I completely dumb?

src/lib/formats/oric_tap.cpp
holub: just did

src/lib/formats/uef_cas.cpp
cuavas: You don’t need the extra `b` because you can assign to `bytes` since the pointer is not `const`, only what it points to is `const`.

See, this is why `uint8_t const *` is clearer than `const uint8_t *`, because you can follow the simple rule that the `const` applies to the thing immediately to the left.

* `uint8_t const *` – the `const` applies to the `uint8_t` part, the pointer is not `const`.
* `uint8_t *const` – the `const` applies to the pointer, the `uint8_t it points to is not `const`.
* `uint8_t const *const` – the first `const` applies to the `uint8_t`, the second `const` applies to the pointer.

holub:
Sorry again for this disaster and thank you'll!

PR #13291, opened 2025-01-27
Update ddenlovr.cpp

Gently28:
麻雀大明神Mahjong Daimyojin（ Deluxe） DIP-SW   （daimyojn）

user-attachments/assets/59c4d6cc-83f6-43b5-83f7-db66163a92c0)


Gently28:
Here it is ,I changed 2 only,earch other waiting @ Vas Crabb to doing,  Ha ha……，excuse me.
A

Gently28:
Here it is ,I changed 2 only,earch other waiting @ Vas Crabb to doing,  Ha ha……，excuse me.
b

cuavas:
Thanks, I improved the settings in baea718933dc386aec1c940d912d7815aea5891a.

image

I might have another look at it later.  It’s unfortunate that the text isn’t very legible in the images.

PR #13298, opened 2025-01-30
paia/midi2cv8.cpp: Adding driver for MIDI2CV8.

m1macrophage:
Includes layout that displays the output voltages.

### DIFF
--- a/src/mame/layout/paia_midi2cv8.lay
+++ b/src/mame/layout/paia_midi2cv8.lay
@@ -1,0 +1,282 @@
+<?xml version="1.0"?>
+<!--
+license:CC0-1.0
+copyright-holders:m1macrophage
+-->
+<mamelayout version="2">
+	<!-- Text elements -->
+
+	<element name="text_power">
+		<text string="POWER" align="1"><color red="0.7" green="0.7" blue="0.7"/></text>
+	</element>
+
+	<element name="text_system">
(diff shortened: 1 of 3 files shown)


galibert:
In which way is it non-working?

m1macrophage:
There are no "physical" voltage outputs. Voltages are just printed on the screen. In some ways, this is similar to systems marked as MECHANICAL. Totally open to alternative of course.

galibert:
There won't be physical voltage outputs in any case.  I find it annoying if there's no state of the driver which we would consider working.  If it just can't be better, why not classify it as working, FVSO ?

m1macrophage:
Sure, tagged as working. Should it also be tagged with some other flag to convey the missing "physical" functionality? Maybe "mechanical" (even thought it doesn't match that exactly)? Whatever you think is right.

rb6502:
For cases like this as long as it's coming up with the correct voltage numbers it can be considered working.  Interfacing to the real world (or to other instances of MAME on the same computer or a LAN) can be done later.  I can't actually find any modern CV/Gate interfaces for computers, just MIDI to CV/Gate interfaces which duplicate this device's functionality, but I haven't looked extensively.

m1macrophage:
Understood. Thanks for the background!

I am also not aware of any non-MIDI computer-to-CV interfaces.

PR #13300, opened 2025-01-30
Various PPC based konami hardware updates/cleanups:

cam900:
Common:
- Fix save states
- Reduce literal tag usage
- Use logerror for logging
- Fix filename
- Use BIT for single bit variables
- Fix initialization

konami/gticlub.cpp:
- Remove unnecessary function

konami/hornet.cpp:
- Split driver state class via configuration
- Move bank config into machine_start()

konami/konppc.cpp:
- Remove hardcoded tags
- Reduce duplicates
- Use memory_bank for texture bank init function
- Fix naming
- Fix spacing

konami/nwk-tr.cpp:
- Split palette per CG boards (K001604 also has palette RAM interface?)


src/mame/konami/konppc.h
```diff
@@ -38,70 +43,70 @@ class konppc_device :  public device_t
...
 
-	std::unique_ptr<uint32_t[]> dsp_shared_ram[MAX_CG_BOARDS];
+	std::unique_ptr<uint32_t[]> m_dsp_shared_ram[MAX_CG_BOARDS];
 
-	uint32_t dsp_state[MAX_CG_BOARDS]{};
-	uint32_t nwk_device_sel[MAX_CG_BOARDS]{};
-	const char *texture_bank[MAX_CG_BOARDS]{};
+	uint32_t m_dsp_state[MAX_CG_BOARDS]{};
+	uint32_t m_nwk_device_sel[MAX_CG_BOARDS]{};
+	memory_bank *m_texture_bank[MAX_CG_BOARDS]{};
```
cuavas, line 81: Would it be a good idea to use `optional_memory_bank_array<2>` for `m_texture_bank` and allow the system using it to configure the tags?

src/mame/konami/zr107.cpp
```diff
@@ -394,7 +394,7 @@ void zr107_state::sysreg_w(offs_t offset, uint8_t data)
 			break;
 
 		case 2: // Parallel data register
-			osd_printf_debug("Parallel data = %02X\n", data);
+			logerror("Parallel data = %02X\n", data);
```
cuavas, line 397: How frequently will this logging happen?  `osd_printf_debug` does nothing in non-debug builds, while `logerror` still sends the output to the debugger, log file and/or console, so it may have a performance impact.  Would it be better to put these behind `logmacro.h` conditional logging?

src/mame/konami/zr107.cpp
```diff
@@ -455,13 +455,16 @@ uint32_t zr107_state::ccu_r(offs_t offset, uint32_t mem_mask)
 			// Midnight Run polls the vertical counter in vblank
 			if (ACCESSING_BITS_24_31)
 			{
-				m_ccu_vcth ^= 0xff;
+				int32_t const vcth = m_ccu_vcth ^ 0xff;
+				if (!machine().side_effects_disabled())
+					m_ccu_vcth = vcth;
 				r |= m_ccu_vcth << 24;
```
cuavas, line 461: Previously the updated value was being used.  Did you mean:
```c++
				int32_t const vcth = m_ccu_vcth ^ 0xff;
					m_ccu_vcth = vcth;
				r |= vcth << 24;
```

src/mame/konami/zr107.cpp
```diff
@@ -455,13 +455,16 @@ uint32_t zr107_state::ccu_r(offs_t offset, uint32_t mem_mask)
...
 			}
 			if (ACCESSING_BITS_8_15)
 			{
-				m_ccu_vctl++;
-				m_ccu_vctl &= 0x1ff;
+				int32_t const vctl = (m_ccu_vctl + 1) & 0x1ff;
+				if (!machine().side_effects_disabled())
+					m_ccu_vctl = vctl;
 				r |= (m_ccu_vctl >> 2) << 8;
```
cuavas, line 468: Same here – previously the updated value was used.  Maybe you meant:
```c++
				int32_t const vctl = (m_ccu_vctl + 1) & 0x1ff;
					m_ccu_vctl = vctl;
				r |= (vctl >> 2) << 8;
```

src/mame/konami/gticlub.cpp
```diff
@@ -420,7 +413,7 @@ uint8_t gticlub_base_state::sysreg_r(offs_t offset)
 		}
 
 		default:
-			osd_printf_debug("sysreg_r %d\n", offset);
+			logerror("sysreg_r %d\n", offset);
```
cuavas, line 416: Same as the other comment about `osd_printf_debug` – is this hit frequently enough to have performance implications that would make it worth making it conditional?

src/mame/konami/hornet.cpp
```diff
@@ -574,7 +611,7 @@ void hornet_state::sysreg_w(offs_t offset, uint8_t data)
 			break;
 
 		case 2: // Parallel data register
-			osd_printf_debug("Parallel data = %02X\n", data);
+			logerror("Parallel data = %02X\n", data);
```
cuavas, line 614: Same as the other comments about `osd_printf_debug` for the cases in this file.

src/mame/konami/hornet.cpp
```diff
@@ -641,8 +678,8 @@ void hornet_state::sysreg_w(offs_t offset, uint8_t data)
 				m_adc12138->sclk_w(BIT(data, 0));
 			}
 
-			bool sndres = (data & 0x80) ? true : false;
-			m_audiocpu->set_input_line(INPUT_LINE_RESET, (sndres) ? CLEAR_LINE : ASSERT_LINE);
+			bool const sndres = BIT(data, 7) ? true : false;
+			m_audiocpu->set_input_line(INPUT_LINE_RESET, sndres ? CLEAR_LINE : ASSERT_LINE);
```
cuavas, line 682: You don’t really need the conditional operator here – `bool const sndres = BIT(data, 7);` should work.

src/mame/konami/zr107.cpp
Hydreigon223: I may suggest adding a todo stating ccu here is identical to k053252.cpp (here in zr107 it's driven by a cpld instead of the custom IC). Did some reversing of that area to figure that out.

No clue if it's possible to add the current k053252 device to zr107.

cuavas: @Hydreigon223 please don’t suggest adding speculation in comments.  Just leave this PR alone.

src/mame/konami/konppc.h
cam900: but nwk-tr.cpp uses both SHARC banks are mapped as board 0, it will be not as easy...
```
void nwktr_state::init_racingj()
{
	init_nwktr();
	m_konppc->set_cgboard_texture_bank(0, membank("master_cgboard_bank"), memregion("master_cgboard")->base());
	m_konppc->set_cgboard_texture_bank(0, membank("slave_cgboard_bank"), memregion("slave_cgboard")->base()); // for some reason, additional CG roms are located on the slave CG board...
}

void nwktr_state::init_thrilld()
{
	init_nwktr();
	m_konppc->set_cgboard_texture_bank(0, membank("master_cgboard_bank"), memregion("master_cgboard")->base());
	m_konppc->set_cgboard_texture_bank(0, membank("slave_cgboard_bank"), memregion("master_cgboard")->base()); // ...while this is not the case for thrilld
}
```

Hydreigon223: It's a copy/paste error on my end. The slave bank should have been set to 1 and I failed to notice that small detail.

src/mame/konami/hornet.cpp
```diff
@@ -704,40 +753,38 @@ void hornet_state::sysreg_w(offs_t offset, uint8_t data)
...
-	if (m_comm_board_rom.found())
-		m_comm_bank->set_entry(bank & 0x7f);
+	m_comm_bank->set_entry((data >> 24) & 0x7f);
 }
 
-uint32_t hornet_state::comm0_unk_r(offs_t offset, uint32_t mem_mask)
+uint32_t sscope2_state::comm0_unk_r(offs_t offset, uint32_t mem_mask)
 {
-//  printf("comm0_unk_r: %08X, %08X\n", offset, mem_mask);
+//  LOGCOMM("comm0_unk_r: %08X, %08X\n", offset, mem_mask);
```
cuavas, line 782: You don’t need to have these commented out now that they’re conditional and disabled by default.

PR #13302, opened 2025-01-30
machine/ncr5385.cpp: delay after XFI_OUT_ACK

Elektraglide:
Without a delay after state XFI_OUT_ACK, tek4404 cannot load anything.
Trial and error determined a delay of <3.5us does not work, more than 3.5us does work.
I dont understand exactly why, but tek4404 scsi code just never sees the state change without this.

[I appreciate code changes and not understanding root-cause are not great, but this does resolve all issues with scsi for tek4404.]

### DIFF
--- a/src/devices/machine/ncr5385.cpp
+++ b/src/devices/machine/ncr5385.cpp
@@ -759,7 +759,9 @@ int ncr5385_device::state_step()
 			else
 				m_sbx = false;
 
-			delay = 3'500;		// >=3.5us delay works, < 3.5us fails
+			// tek4404 fails to load without a delay here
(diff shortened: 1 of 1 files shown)


Elektraglide:
OK, found the smoking gun in the tek4404 bootrom.  It does not expect the signal to arrive so quickly so clears the variable tracking IRQ_SCSI_complete and then waits for it to be set.  which it now never will be.  The delay means it has time to clear IRQ_SCSI_complete and *then* go into its busy wait loop.

Screenshot 2025-01-31 at 07 40 56

pmackinlay:
See https://github.com/mamedev/mame/pull/13407.

PR #13303, opened 2025-01-31
cpu/kl1839vm1.cpp: Updated save state

holub:


src/devices/cpu/mpk1839/kl1839vm1.cpp
```diff
@@ -768,16 +768,23 @@ void kl1839vm1_device::vax_decode_pc()
...
 	} while ((arg_n < 6) && (args[arg_n] != vax_disassembler::mode::none));
 
+	m_pcm_queue.clear();
 	if (m_op_size > 0) // above completed without failure
 	{
+		while (!tmp_args.empty())
+		{
+			m_pcm_queue.push_back(tmp_args.back());
+			tmp_args.pop_back();
+		}
```
cuavas, line 786: Is there any chance this can cause `m_pcm_queue` to grow beyond six items?  Because if it can, it will break save states.

Registering a `std::vector` for save states saves the size/address of the storage **at the time it’s registered**.  If you do anything that causes it to be resized later, it breaks, because the new size won’t be saved/restored, and if the storage was reallocated, the address won’t be updated.  To be honest, I think allowing `std::vector` to be registered for save states was a mistake, given the pitfalls.

If the PCM queue always has a size of six, you can use an array of some kind.  If the size never exceeds six but is variable at the time you return to the scheduler, you can use an array and store the size or start/end separately.  (Of course, if the PCM queue is always empty when you return to the scheduler, you don’t actually need to save it, and you can add an `assert` to check that it’s empty when exiting the execute loop.)

holub: It's always 6. I tried to avoid extra internal state, but as you insist - I'll do.

cuavas: If it’s always exactly 6, you can just maintain a local pointer to the current position within this function, can’t you?

holub: A bit tricky as it's <=6 but done.

PR #13304, opened 2025-01-31
philips/mcd212.cpp: Rename matte flag constants and refactor matte handling code

Vincent-Halver:
This intentionally has no visible effects. The purpose is to make the code easier to understand and maintain.

1. The non-canonical name "Regions" was changed to the official name "Mattes". Corresponding abbreviation "R" is changed to "M".
2. Deprecated function get_plane_width is deleted.
3. mix_lines is refactored.
4. Minor changes.

### DIFF
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -59,22 +59,20 @@ void mcd212_device::update_matte_arrays()
 	uint8_t latched_wfb = m_weight_factor[1][0];
 	const int width = get_screen_width();
 
-	if (BIT(m_image_coding_method, ICM_NM_BIT))
+	int x = 0;
+	const int num_mattes = BIT(m_image_coding_method, ICM_NM_BIT) ? 2 : 1;
+	const bool matte_flag = BIT(m_matte_control[0], MC_MF_BIT); // MF bit must be the same. See 5.10.2 Matte Commands
+	
+	int matte_idx = 0;
+	for (; x < width; x++)
 	{
-		if (get_matte_op(0) == 0 && get_matte_op(4) == 0)
+		for (int f1 = 0; f1 < num_mattes; f1++)
 		{
-			std::fill_n(m_weight_factor[0], std::size(m_weight_factor[0]), latched_wfa);
-			std::fill_n(m_weight_factor[1], std::size(m_weight_factor[1]), latched_wfb);
-			std::fill_n(m_matte_flag[0], std::size(m_matte_flag[0]), false);
-			std::fill_n(m_matte_flag[1], std::size(m_matte_flag[1]), false);
-			return;
-		}
-
-		for (int x = 0; x < width; x++)
-		{
-			for (int flag = 0; flag < 2; flag++)
+			const int max_matte_id = (0x10 >> num_mattes) + (f1 << 2);
+			const int flag = (num_mattes == 2) ? f1 : matte_flag;
+			if (num_mattes == 2)
 			{
-				for (int matte = 0; matte < 4; matte++)
+				for (int matte = 0; matte < max_matte_id; matte++)
 				{
 					const int matte_idx = (flag << 2) + matte;
 					const uint32_t matte_ctrl = m_matte_control[matte_idx];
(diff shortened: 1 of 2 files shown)


src/mame/philips/mcd212.cpp
```diff
@@ -42,175 +42,165 @@
...
 	uint8_t latched_wfb = m_weight_factor[1][0];
 	const int width = get_screen_width();
 
-	if (BIT(m_image_coding_method, ICM_NR_BIT))
+	int x = 0;
+	const int num_mattes = BIT(m_image_coding_method, ICM_NM_BIT) ? 2 : 1;
+	const bool matte_flag = BIT(m_matte_control[0], MC_MF_BIT); // MF bit must be the same. See 5.10.2 Matte Commands
+	
+	int matte_idx = 0;
+	for (; x < width; x++)
```
angelosa, line 67: Any reason about to have `int x = 0;` outside of the for block? i.e. isn't just `for (int x = 0; x < width; x++)` possible here?

Vincent-Halver: Yes, that's possible for now. It's pre-emptively moved out as that will be needed later to fully deduplicate the switch statements around lines 88 and 143. But the loop logic was getting tricky, so I chose to not include the full refactoring.

Vincent-Halver: By tricky, I mean that the loops contain a functional bug in the original version due to incorrect order of execution of the loops. However I wanted to avoid adding functional changes in this request. First I want to align on the renaming and refactoring before continuing with bug fixes.

Slamy:
@Vincent-Halver Are you sure you want to rename regions to mattes? At least according to the datasheet of the MCD212 the name region is actually used in there. I'm asking as this might confuse future readers of mcd212.cpp

Vincent-Halver:
@Slamy I feel reasonably sure, though I see why you ask. Let me explain. Mattes and Regions are closely related in the spec:

From the Greenbook Spec:
> 2.3.2.6 Regions
Regions divide the set of points in the UCM coordinate space into two disjoint sets
of points. Their main uses are for graphics drawing and construction of mattes.

> 2.3.4.3 Regions
Regions divide a drawmap into two disjoint sets of pixels. Their primary use is to
limit the area of drawing in a drawmap. Regions can also be drawn. Regions may
be of any arbitrary shape or size. Regions are created from the basic shapes of
rectangle, polygon, circle, circular wedge, ellipse, and elliptical wedge. More
complex shapes are created from these basic shapes using intersection, union,
exclusive-or, and difference operations. (see VII.2.3.2.6 for more information).

The various function names in the OS call these shapes and unions/intersections/etc as "regions".
> RG_Creat - Create Region
RG_Creat creates a region using one of the basic shapes.
Input: d0.w = Path number
d1.w = SS_RG setstat code
d2.w = RG_Creat function code
d3.w = Region type (rectangle, polygon, etc.)
d4.l = Region type dependent
d5.l = Region type dependent
d6.l = Region type dependent
d7.l = Region type dependent
(a0) = Region type dependent

The Validation Disc also uses this terminology. You may see debug output when using that disc that refers to these function names. Regions are small files occupy space in the file manager and developers have to allocate and deallocate them.

I do see that the MCD212 PDF doesn't use the word Matte, and instead calls them regions. That means that a choice would have to be made: do you use the Motorola chip's terms, or the official Philips terms?

I think it would be good to consider who would actually be reading and writing the code. When working on the emulator, the Validation Disc, Operating System, and official CD-i specification all use the term Matte. It reduces developer confusion to use the same terms that the software uses. So I come to the conclusion it should follow the Greenbook terminology.

Slamy:
As I'm not the maintainer of this code, I also don't have a saying in this. I was just wondering, considering MAME is not using the HLE approach and sees the systems more low level. In theory CD-i is a software standard and there I would totally understand the naming being done like in the green book.
The same problem exists in the src/mame/philips/cdicdic.cpp. The green book says it is a soundmap while the mame code calls it the audiomap. As there is no public datasheet, nobody really knows... :-(
Are you a member of the CD-i discord server? I'm actually observing your actions, considering the state of CD-i emulation is not the best right now. I've also talked to MooglyGuy some time ago about the MAME CD-i code. He seems to have lost interest though...

angelosa:
Green Book is a standard, and it's not confined to Philips CD-i alone.

> The same problem exists in the src/mame/philips/cdicdic.cpp. The green book says it is a soundmap while the mame code calls it the audiomap.

Those are interchangeable names tbh, there isn't a real convention there. It's all about personal preferences, I prefer to use actual doc naming for stuff, except where that turns out to be impractical or confusing for MAME.

Vincent-Halver:
> Are you a member of the CD-i discord server

I am not, but I would join. Something that would help me a lot would be someone to record all the visual tests from Validation Disc (E) on real hardware. 

It sounds like we're ok with using the Greenbook spec naming, so if that's alright, let's pull in the change, and I'll clean up the next section of code that fixes the Matte execution order that I alluded to above.

PR #13312, opened 2025-02-01
konami/simpsons.cpp: Cleanups

cam900:
- Fix save state supports
- Reduce literal tag usage, runtime tag lookups
- Reduce unnecessary lines

konami/konami_helper.cpp: Allow variable type for konami_sortlayers* function, Make swap function inline


src/mame/konami/simpsons.cpp
```diff
@@ -593,16 +598,15 @@ INPUT_PORTS_END
...
-	int counter, num_inactive;
-	uint16_t *src, *dst;
+	uint16_t *dst;
 
 	m_k053246->k053247_get_ram(&dst);
 
-	src = m_spriteram.get();
-	num_inactive = counter = 256;
+	uint16_t *src = m_spriteram.get();
+	int num_inactive = 256, counter = 256;
```
cuavas, line 606: Can it be `uint16_t const *src;`?  I don’t think it should be writing to the source, just updating the pointer.

src/mame/konami/simpsons.cpp
```diff
@@ -593,16 +598,15 @@ INPUT_PORTS_END
...
 
 	do {
-		if ((*src & 0x8000) && (*src & 0xff))
+		if (BIT(*src, 15) && (*src & 0xff))
 		{
 			memcpy(dst, src, 0x10);
 			dst += 8;
```
cuavas, line 612: Is there a good reason not to express this as a `for` loop?  Wouldn’t it work as:
```c++
	for (int counter = 256; counter; --counter)
	{
		...
```

As far as I can tell, `counter` is only used as the loop counter.

Also, the loop below seems to be a convoluted way of forcing a `do` loop.  You could just use:
```c++
	while (num_inactive--)
	{
		*dst = 0;
		dst += 8;
	}
```

PR #9329, opened 2022-02-22
add TSX support for MSX

imulilla:
I add support for the TSX format. A TSX file is a TZX file (http://k1.spdns.de/Develop/Projects/zasm/Info/TZX%20format.html) that contains a new block, 4B, (https://tsx.eslamejor.com/res/tsx4b.png) specific to MSX but can also be used for protections on other systems.
I have added a new entry in the MSX software list, "Space Harrier 2" in TSX format

PR #13317, opened 2025-02-02
formats/tzx_cas.cpp: Added block 4B support (TSX format) / Fixed bug in block 35

holub:
https://github.com/nataliapc/makeTSX/wiki/Tutorial-How-to-generate-TSX-files#14-the-new-4b-block


holub:
Even if we are strictly against adding tsx format, at least block https://github.com/mamedev/mame/pull/35 fix has to go through. ref: https://worldofspectrum.net/features/TZXformat.html#CUSTOMBLOCK

As for me in such situation without mutual agreement I would accept the format itself but restrict such items in softlist.

rb6502:
Are we strictly against it?  I haven't heard any such thing.

holub:
Somebody pointed to this after I opened PR:  #9329

holub:
So, what would be directions? Pull #35 block fix and drop the rest (tsx support)?

angelosa (approved):

src/lib/formats/tzx_cas.cpp
```diff
@@ -736,6 +799,30 @@ static int tzx_cas_do_work( int16_t **buffer )
...
+		case 0x4b:  // TSX Data Block
+			{
+				data_size = get_u32le(&cur_block[1]) - 12;
+				pause_time = get_u16le(&cur_block[5]);
+				pilot = get_u16le(&cur_block[7]);
+				pilot_length = get_u16le(&cur_block[9]);
+				uint16_t zero = get_u16le(&cur_block[11]);
+				uint16_t one = get_u16le(&cur_block[13]);
+				bit0 = (cur_block[15] >> 4) ?: 16;
+				bit1 = (cur_block[15] & 0x0f) ?: 16;
```
cuavas, line 811: This is non-standard GNU crap, and hence not portable.

happppp: https://github.com/mamedev/mame/pull/14206

PR #13319, opened 2025-02-02
oberheim_dmx.lay: Proper tracking of pointer IDs. Easier fine adjustments.

m1macrophage:
- Click and adjustment state is now tracked for each pointer ID individually, rather than globally.
- If a user selects the slider knob, the adjustment will be proportional to the knob movement. This allows for finer adjustments. If the user clicks elsewhere on the slider, the adjustment is based on the absolute position of the click.

This is a followup to the discussion in #13177.  Tested on a single-mouse setup.

This should also work with multiple pointer devices, and touchscreens (assuming the latter generate "button press" events). But I was not able to verify. I tried testing with multiple mouse cursors (using `xinput` and multiple mice), but it seems like SDL2 does not support that. I also tried emulating a touchscreen with `evdev`, but this didn't work either. If someone has  a way to test multiple cursors or emulate a touchscreen on linux, let me know. Alternatively, if someone with a touchscreen can validate, that would be great!

### DIFF
--- a/src/mame/layout/oberheim_dmx.lay
+++ b/src/mame/layout/oberheim_dmx.lay
@@ -543,13 +543,11 @@ copyright-holders:m1macrophage
 			function()
 				local id_port_index <const> = string.len("slider_knob_") + 1
 
-				-- Local state used by the pointer update handler.
-				local sliders = {}
-				local slider_knobs = {}
-				local slider_fields = {}
-				local selected = 0
+				-- State used by pointer handlers.
+				local sliders = {}  -- Info about all sliders (constant after initialization).
+				local pointers = {}  -- Pointer tracking state.
 
-				-- Gather relevant elements and inputs into local state.
+				-- Gather relevant elements and inputs into `sliders`.
 				local view = file.views["Default Layout"]
 				for i = 1, #view.items do
 					local item = view.items:at(i)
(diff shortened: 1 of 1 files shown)


galibert:
It reacts a little weirdly.  Using "./mame obdmx -now -screen screen1
-enable_touch" (screen1 is the one in my three screens that has touch),
multitouch works.  For instance I can move two sliders independently at the
same time.  But the mame mouse appears, moves kinda randomly, and considers
there is a click when touching the screen somewhere else.  For instance
with the dmx, if the mouse happens to be over one of the pads and you touch
another, both pads will be triggered.  It's kinda weird.

I've not been able to understand how the mouse moved.  Without
-enable_touch, it moves following relative movement when touching the
screen (e.g. no absolute positioning).  With it... it's just weird.

m1macrophage:
Thanks for taking a look! To rule out the script itself as the source
of weirdness, consider running with `-enable_touch -noplugin layout`. Sliders
won't work of course, but if you are able to reproduce the issue with the pads,
then the source of weirdness might be elsewhere.

galibert:
Yeah, the issue is elsewhere.  And I solved the weirdness: the mame mouse moves by exactly how much you moved the finger while touching, but in one step once you release the touch.

Still needs to be solved before (multi?) touched is fully usable.  It's very nice to have a real driver to try with.

galibert:
Ok, it was an xinput misconfiguration.  It works 100%.

m1macrophage:
That's great to hear!

PR #13321, opened 2025-02-02
scc68070: limit external memory addresses to 24 bits

devinacker:
Fixes `cdi:zombie` and clones crashing after the intro, and possibly other stuff.

Not sure if `cpu/m68000/scc68070.cpp` is the ideal place to put this, or if somewhere else would work better for if/when the MMU gets implemented.


angelosa:
SCC68070 still has two clients only: philips/cdi.cpp and misc/magicard.cpp. If both don't have any relevant regression(s) then this is good to go (and I don't think there will be by looking at the `address_maps`

PR #13325, opened 2025-02-03
cpu/drcbex64.cpp: Directly dispatch memory accesses to specific handlers

cuavas:
At the urging of @galibert, this makes the x86-64 recompiler back-end dispatch memory accesses to specific handlers directly, avoiding one level of out-of-line function calls.  From my testing on an Intel 8th generation i7, the performance improvement is measurable, but less than 5% at best.

I’d like people to test this with other x86-64 CPUs (e.g. AMD or different generations of Intel CPUs) to see what kind of difference it makes (if any).

If it does seem to produce a useful improvement, I can follow up by adding code to directly dispatch reads smaller than native width, and add scaffolding for carrying over the same optimisations to AArch64.  If it doesn’t seem to be a worthwhile improvement, I’ll just move across the removal of the static address space accessors and ditch the recompiler changes.

### DIFF
--- a/scripts/genie.lua
+++ b/scripts/genie.lua
@@ -570,14 +570,6 @@ configuration { "gmake or ninja" }
 
 dofile ("toolchain.lua")
 
-if _OPTIONS["targetos"]=="windows" then
-	configuration { "x64" }
-		defines {
-			"X64_WINDOWS_ABI",
-		}
-	configuration { }
-end
-
 -- Avoid error when invoking genie --help.
 if (_ACTION == nil) then return false end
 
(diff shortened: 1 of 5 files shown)


cuavas:
@MooglyGuy if you could see if this yields improvements on your AMD CPU, I’d appreciate it.

cuavas:
I’ve updated this pull request to also inline reads smaller than native width.  However, in my testing this actually hurts performance for `pmac6100`.

If/when people try this out, please test and benchmark the two commits separately.  Only the first commit requires an emu.h rebuild – If you can check out the first commit, build, copy the binary, check out the second commit, and build again, the second build will only need to recompile src/devices/cpu/drcbex64.cpp.

MooglyGuy:
I'll have a look this coming weekend, just remind me.

cuavas:
I sorted out the performance issues with my initial implementation of reads smaller than the native width.  I’m pretty confident this is a modest but useful improvement now.

PR #13328, opened 2025-02-03
konami/mystwarr_v.cpp, k053246_k053247_k055673.cpp: improve sprite blending

FredYeye:
This pull request is split into three commits:

1. d0faecd
The sprite effect attribute bits in mystwarr, metamrph and viostorm are used for sprite blending purposes. I changed the sprite callbacks to attach the attribute bits.

2. e7f911e
Blending was being done in reverse. There are probably several reasons why this was never noticed, but two reasons would be that 0% and 100% blending worked as intended and almost no games use sprite blending (at least in current MAME). Simple solution: reverse the operands to *alpha_blend_r32()*.

3. 8184130
Optional commit: minor cleanup of *zdrawgfxzoom32GP()*, the function that contains the *alpha_blend_r32()* calls. Initially I had planned to make bigger changes, like condensing the switch-cases, but I wasn't satisfied with my attempts at doing so. Instead I opted for a more conservative cleanup pass. It appears to give a marginal speedup as a bonus.
This is the least important commit of the three, so if it's deemed unnecessary it can be dropped.
---
This fixes transparent sprites in mystwarr, metamrph and viostorm. Mystwarr & viostorm make little use of these; it's a bit more apparent in metamrph.
I assume this affects other systems too - like konami GX? I ran all GX games with -bench 400 and didn't run into any sprites that use any transparency. Either GX games do not have transparent sprites or they're not tagged as such in current MAME. This is hardly the perfect testing method, so please let me know if I should be testing any other games and systems.


angelosa:
I'm not gonna judge the changes here (they looks reasonable to me), you may give a shot at `dadandrn` Centaur and Fungus stages (both uses blending in one form or another). Don't think `sexyparo` stages 3L and 3W (i.e. Castlevania stage and its mission 2 quota satisfied equivalent) are affected here but you may try anyway.

FredYeye:
Good call on dadandrn! I had trouble finding real HW footage so I didn't dig into it earlier. Found a [video here](https://www.youtube.com/watch?v=FTvMWYgtLLo) to compare against. I switched it over to using the mystwarr sprite callback and blending does seem to work! It was using the gaiapols callback which is slightly different, so I will play through the rest of the game and see if anything looks out of place.

Will check sexyparo soonish, whenever I have time.

FredYeye:
I pushed the `dadandrn` change. This makes blending happen. It should be mentioned that many effects still don't look quite right due to the brightness registers going unused in mame at the moment.

There is some other problem with blending, maybe overlapping blended sprites? This can be seen in `dadandrn` (centaur light pillar) and `metamrph` (4 solid colors overlay in the intro). Fixing this in a timely manner may be out of scope for this PR though...

FredYeye:
(accidentally closed request somehow, don't mind that)

happppp:
While you're at it, could you rename "eax" to something more contextual like "delta"?
It's Acho's code, he had a habit of naming temp variables after Intel X86 registers.

FredYeye:
Yes, you're right, that should be done. It seems like most of the "eax" vars are either a distance between a clipping rect and a sprite coord, or a sprite tile byte used as a palette index. I've tried to at least make the names a little more descriptive.

FredYeye:
I took a look at `sexyparo`. For stage 3, there's a layer of solid ink that should be transparent. I assume this is a problem with tile / layer blending. I want to get to that eventually but I haven't made much headway there.

The first boss in the boss rush is some blue slime with a coin inside. That thing appears to use blending and seems to work as intended in mame.

angelosa:
iirc the very same path is used by Castlevania stage for the glass background meshes (which should blend, around the boss). 
`dadandrn` use a form of color mixing, specifically a powered up version of how SNES PPU does things i.e. a main screen and a sub screen where a color DAC offset is added/subtracted depending on what's happening on screen. 
You can cross check what I mean with the attract intro cycle, where on basic form it just fade-in or out the full video output, while enabling finer tuned control around the Moai punching the camera.

src/mame/konami/k053246_k053247_k055673.cpp
```diff
@@ -441,46 +441,20 @@ void k053247_device::k053247_sprites_draw(bitmap_rgb32 &bitmap, const rectangle
...
 void k053247_device::zdrawgfxzoom32GP(
 		bitmap_rgb32 &bitmap, const rectangle &cliprect,
-		u32 code, u32 color, int flipx, int flipy, int sx, int sy,
+		u32 code, u32 color, bool flipx, bool flipy, int sx, int sy,
 		int scalex, int scaley, int alpha, int drawmode, int zcode, int pri, u8* gx_objzbuf, u8* gx_shdzbuf)
 {
 #define FP     19
 #define FPONE  (1<<FP)
 #define FPHALF (1<<(FP-1))
 #define FPENT  0
```
cuavas, line 450: Can you turn these into `constexpr int` or `constexpr unsigned` while you’re at it?  Preprocessor macros have a habit of causing difficult to diagnose errors.

FredYeye: I changed these to `constexpr int` and realized that `FPONE` and `FPHALF` go unused. Should I comment them out or outright remove them?

src/mame/konami/mystwarr_v.cpp
```diff
@@ -77,29 +77,21 @@ K056832_CB_MEMBER(mystwarr_state::game4bpp_tile_callback)
...
-	{
-		*color = c;
-		*priority_mask = (attr & 0xe0) >> 2;
-	}
-	else
-	{
-		*color = c | 3<<K055555_MIXSHIFT | K055555_SKIPSHADOW; // reflection?
-		*priority_mask = 0x1c;
-	}
+	if (shadow == 0b11) { *color |= K055555_SKIPSHADOW; }
```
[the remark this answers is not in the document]
FredYeye: I didn't pay enough attention there, I've changed it to use your suggestion, thanks!

FredYeye:
> `dadandrn` use a form of color mixing, specifically a powered up version of how SNES PPU does things i.e. a main screen and a sub screen where a color DAC offset is added/subtracted depending on what's happening on screen. You can cross check what I mean with the attract intro cycle, where on basic form it just fade-in or out the full video output, while enabling finer tuned control around the Moai punching the camera.

I see!! Yeah, I can tell there's still things to improve on.

src/mame/konami/k053246_k053247_k055673.cpp
cuavas: Just remove them, the definitions contain no unique information as they’re just derived from `FP`.

FredYeye: Alright, done.

TheCoolPup:
someone trying to fix video emulation for konami stuff??? its a mircale!

TheCoolPup:
also i have noticed one regression, if you look at eagle's teeth, they seem to be the wrong color 
Monster Maulers (ver EAA)  mmaulers  - MAME 0 274 (LLP64) 2_7_2025 5_46_35 PM
And also in terms of the water missing in this scene could this be a blending issue or something else entirely?
Monster Maulers (ver EAA)  mmaulers  - MAME 0 274 (LLP64) 2_7_2025 5_48_10 PM

FredYeye:
Yep, noticed the teeth as well! Will have to investigate

TheCoolPup:
so the water is a blend issue?

TheCoolPup:
also here is some pcb footage if you want it https://youtu.be/iuLQnJeBKZo?feature=shared&t=6197

FredYeye:
The teeth regression appears to have been a priority mask issue. I updated the `mystwarr_sprite_callback` priority mask to use one less bit, which fixes it.  This fix seems to make sense: the mask bits were using a bit that I guess is used for palette indexing (*color & 0x1f).

As a bonus, it also fixes the final boss's monitors in mystic warriors! Thanks for pointing it out.

> so the water is a blend issue?

It's possible! This PR is only focusing on sprites, though.

---
I think this is all I will get done for this PR (unless there are other suggestions / objections). `dadandrn` is far from perfect as angelosa pointed out, but I think fixing that won't be as simple.

TheCoolPup:
So happy someone is working on fixing the issues these games have had, thank you again

FredYeye:
Alright, so I went ahead and made one last addition. I'm sure I'm being annoying now 😅.
The whole reason I started changing `zdrawgfxzoom32GP` in the first place was because there were 4 different call sites to the alpha blending function. Almost all the switch cases are very similar so I condensed them down. Drawmodes 0-3 now take the same path, while mode 4 is still separate. There's still different code paths for zoom though, but this reduces some code duplication.

In my testing all current changes come out ~7% faster than current mame, so it seems the compiler can work with drawmode checks inside the loop.

happppp:
After having changed that switch to if, should you add a drawmode==4 check to the last one?

FredYeye:
If I'm reading this right, `drawmode` gets created in `konamigx_mixer_draw()`, set based on the variables `temp2` and `temp4` in `konamigx_mixer()`. `temp2` can be 0-3, while `temp4` can be 4-5. Mode 5 either gets toggled to 4 or makes an early return in `zdrawgfxzoom32GP()`.
So it would seem that 0-5 are possible on entry and 0-4 at the if statement. Would you like me to add the check anyway? Or add a note about this? `drawmode` could be enum material, though I don't know about making that change right now.

happppp:
I see, nevermind =)

happppp:
@FredYeye this caused a regression with screen fade during stage transitions in Violent Storm. See https://mametesters.org/view.php?id=9131

K055555_SKIPSHADOW when shadow flags == 3 smells like a workaround to me for what is probably a priority issue? If I remove that, the fade works again, but it reintroduces a bug with the fireworks after finishing stage 2. And as expected, it'll also make shadows on the ice level worse in Metamorphic Force.

FredYeye:
Ah! I will take a look. I agree that the K055555_SKIPSHADOW check is probably a hack. This is something that existed before I changed the function, let me find the old version...

Here: https://github.com/mamedev/mame/blob/ab6237da824a95898209ecbb568095981c4dbe69/src/mame/konami/mystwarr_v.cpp#L85

I'm a bit handicapped at the moment so everything takes me a bit more time right now, apologies if i'm doing things slowly.

FredYeye:
@happppp There are many things that are either missing or working incorrectly concerning shadows as far as I can tell. I want to improve this but it will definitely take time.

happppp:
Understood.

PR #13331, opened 2025-02-04
Add a skeleton driver for DEC MicroVAX 3100

clawgrip:
MicroVAX 3100 [OldComputers ES]


src/mame/dec/microvax3100.cpp
```diff
@@ -0,0 +1,170 @@
...
+	// DUSCC68562(...) // Signetics SCN68562C4N48
+
+	// Video hardware
+	GENERIC_TERMINAL(config, m_terminal, 0);
+}
+
+ROM_START( mvax3100 )
+	ROM_REGION( 0x40000, "maincpu", ROMREGION_ERASEFF )
+	ROMX_LOAD( "dec89_23-116e8-00_system_rom_lo_word.e24", 0x00000, 0x20000, CRC(69ef8cf5) SHA1(a59a500921278dc356a519cb435641426c844779), ROM_NIBBLE | ROM_SHIFT_NIBBLE_LO | ROM_SKIP(1) )
+	ROMX_LOAD( "dec89_23-117e8-00_system_rom_hi_word.e25", 0x00000, 0x20000, CRC(df572ac3) SHA1(5e91d0f4fc8442e3ebc2424d9f85078ba00d2de5), ROM_NIBBLE | ROM_SHIFT_NIBBLE_HI | ROM_SKIP(1) )
```
ajrhacker, line 146: I don't think this loading is correct. These are supposed to be 65536x16 EPROMs, not 131072x4.

src/mame/dec/microvax3100.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+	ROM_REGION( 0x00100, "prom", ROMREGION_ERASEFF ) // On the main PCB
+	ROM_LOAD(  "dec84_fx9206_365a1.bin",                   0x00000, 0x00100, NO_DUMP )
+ROM_END
+
+} // anonymous namespace
+
+
+//    YEAR  NAME      PARENT  COMPAT  MACHINE       INPUT         STATE               INIT        COMPANY                          FULLNAME         FLAGS
+COMP( 1989, mvax3100, 0,      0,      microvax3100, microvax3100, microvax3100_state, empty_init, "Digital Equipment Corporation", "MicroVAX 3100", MACHINE_NOT_WORKING | MACHINE_NO_SOUND)
```
ajrhacker, line 170: Unfortunately this is not specific enough, since DEC released several different (though feature-comparable) systems under the "MicroVAX 3100" name. The PCB description suggests that this one is a Model 10 or 20 (there was very little difference between these two).

ajrhacker (changes requested):

src/mame/dec/microvax3100.cpp
clawgrip: Fixed!

src/mame/dec/microvax3100.cpp
clawgrip: Fixed! It's model 10. We'll dump a Model 20 soon

src/mame/dec/microvax3100.cpp
```diff
@@ -0,0 +1,172 @@
...
+	microvax3100_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_maincpu(*this, "maincpu")
+		, m_terminal(*this, "terminal")
+	{ }
+
+	void microvax3100(machine_config &config);
+
+private:
+	required_device<t11_device> m_maincpu;
```
cuavas, line 120: Doesn’t this need a VAX CPU, not PDP11?

clawgrip: Do we have a VAX CPU core? I've only seen the disassembler...

PR #13333, opened 2025-02-06
nwk-tr.cpp: Revert back to one K001604 to properly show graphics during POST

Hydreigon223:
racingj and racingj2 had a problem where some portions of POST wouldn't properly show with two k001604 tilemap ICs. This is because of the hardware configured in a way so that each CG board had its own separate bit (EXID0 and EXID1) to set for a read/write. This means the main board could set both bits on for a broadcast read/write to both boards simultaneously which is exactly what racingj does before POST to initialize both k001604 registers at once.

I myself have no clue if this kind of broadcasting is supported when it comes to address maps as it would be nice to convert k001604's register area to use a memory map instead of switch case mess. Regardless, it makes more sense to revert to one k001604 as the registers initialized to both are the same by default.

### DIFF
--- a/src/mame/konami/nwk-tr.cpp
+++ b/src/mame/konami/nwk-tr.cpp
@@ -259,11 +259,11 @@ public:
 		m_audiocpu(*this, "audiocpu"),
 		m_dsp(*this, "dsp%u", 1U),
 		m_k056800(*this, "k056800"),
-		m_k001604(*this, "k001604_%u", 1U),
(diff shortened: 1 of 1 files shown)


galibert:
Writing to two devices at the same time is not supported and will not be.  It requires hardware support so funky it ends up being extremely rare and a case-by-case basis.  Not something suitable to generic infrastructure.

galibert:
I mean not supported in address maps, of course.

PR #13337, opened 2025-02-07
Update namcos12.cpp - truckk can't be a Japanese set

mamehaze:
marked it as ASIA/US? instead but it's most likely a US set, due to Parental Advisory warning.

The Japanese language in-game is hardcoded to disabled (windyfairy did some research on it before)

A real Japanese set would likely be hardcoded to Japanese in-game, or have the option to choose enabled.  That would most likely have the code TKK1, rather than TKK2 (although we can't be sure as the number is just release order, not region)

### DIFF
--- a/src/mame/namco/namcos12.cpp
+++ b/src/mame/namco/namcos12.cpp
@@ -3642,6 +3642,7 @@ GAME( 1999, aquarush,   0,        coh700,   lbgrande, namcos12_state,         em
 GAME( 1999, golgo13,    0,        golgo13,  golgo13,  golgo13_state,          empty_init, ROT0, "Eighting / Raizing / Namco", "Golgo 13 (Japan, GLG1/VER.A)", 0 ) /* KC054 */
 GAME( 2000, sws2000,    0,        coh700b,  namcos12, namcos12_state,         empty_init, ROT0, "Namco",           "Super World Stadium 2000 (Japan, SS01/VER.A)", MACHINE_NOT_WORKING ) /* KC055 */
 // truckk can't be a Japanese set despite the game title still being in Japanese.  The option for in-game Japanese text is disabled in this set, and it has a Parental Advisory screen usually associated with US releases
+// TKK1 is probably the Japanese release, although a TKK1 CD contains identical data to a TKK2 CD, so any difference is in the ROM data rather than the CD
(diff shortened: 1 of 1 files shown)


happppp:
Since it has the US warning screen, why think it might be the Asia version?

mamehaze:
Vas and RB seemed to think that some regions of Asia might also display that warning, since it doesn't also have the WDUD, only the warning.

cuavas:
It was in the shout box.

It isn’t a US rating screen – if it was it would use ESRB terminology (“Everyone”).  It isn’t a Hong Kong/Taiwan/Southeast Asia GSRR rating either, as that would be “General Public”.  It isn’t a PEGI rating, as it would get the 3+ due to being too difficult for a toddler to play.  It isn’t an Australian OFLC rating – that would be “G” for general audiences.

It’s close to the wording for a Korean GSRB rating, but IIRC the media laws there at the time still would have required completely scrubbing Japanese text.

We’ve seen Korean and English language flyers for the game with Asia/Pacific distributors listed, suggesting that releases in the Asia/Pacific region were at least planned.

mamehaze:
It is the format Namco typically used for US releases though, even if the wording isn't standards compliant

FoxhackDN:
I think it's an Asia region release (or possibly a general overseas one) based on this Korean flyer:

https://flyers.arcade-museum.com/videogames/show/2171

The game here is shown in Japanese (no edited on screen text) and the flyer is in English but it has Korean in some places.

Not to mention Bandai Namco has an English product page for the game... https://bandainamco-am.co.jp/am/english/aa/truckkyosokyoku/

happppp:
Isn't the warning this thing?
CN67fJw

FoxhackDN:
image
Taken from a Youtube video for reference

happppp:
Soul Calibur USA version with warning screen number 5 from flyer above

mamehaze:
Those are certainly the style of warnings and text Namco used for US releases, Ehrgeiz, Mr Driller, Tekken Tag...

ehrgeizua
drill
tektag

happppp:
mslug2 if you set BIOS to USA region (no such warning for Asia region). I'm not aware of Asian region arcade games having the AAMA warning.


With a funny spelling error.

mamehaze:
Namco also used it on the US releases of Point Blank 2
point_Blank_2

mamehaze:
So yeah, I still think US is most likely.

Hammy1986:
FYI i dumped CD of TKK1 and it matches TKK2 CDrom already in mame. So the region on these is only dictated with the flashroms.
Probably worth a note in the driver the CD's match if it's not noted already.

PR #13339, opened 2025-02-07
mame.lst: update primrag2 string

mikehyrus:
Forgot this one by accident.


happppp:
To be honest, I don't care about all the comments in this file, I'd rather they be removed. It's a setname index for the compile script, not something users are meant to browse/read.

mamehaze:
ideally the file needs to go away at some point too, but it seems there are still issues with making that step

cuavas:
> To be honest, I don't care about all the comments in this file, I'd rather they be removed. It's a setname index for the compile script, not something users are meant to browse/read.

Want me to do that with a script?

mikehyrus:
Yeah, sure thing Vas.

happppp:
I already have a local commit removing them for tiny.lst, give me a second, I'll push that one first.
And yeah, if noone minds it, I think the ones in mame.lst can be removed.

PR #13340, opened 2025-02-07
New machines marked as NOT_WORKING: MIT Cadr

wilbertpol:
New machines marked as NOT_WORKING
--------------------------------
MIT Cadr [Bitsavers, Wilbert Pol]

cadr.xml: Added 6 not working items.

System 78.48 [ams]
System 100.0 [ams]
System 300.0 [ams]
System 301.0 [ams]
System 302.0 [ams]
System 303.0 [ams]

AI disclosure: Claude Sonnet 5 assisted with debugging and fixing some small issues..


src/devices/cpu/cadr/cadr_dasm.h
```diff
@@ -0,0 +1,40 @@
...
+private:
+	static constexpr u64 NOP_MASK = u64(0x7fffffffeffff);
+
+	static const char *const bool_op[0x10];
+	static const char *const arith_op[0x10];
+	static const char *const arith_op_c[0x10];
+	static const char *const mult_div_op[0x20];
+	static const char *const output_bus_control[0x04];
+	static const char *const q_control[0x04];
+	static const char *const rp[0x04];
```
cuavas, line 31: Do these really need to be in the header?  It’s just exposed implementation details that slow down compiling downstream stuff and increase the chance of causing a recompile if the implementation changes.  Throw them in an anonymous namespace in the .cpp file.

src/mame/mame.lst
```diff
@@ -32779,6 +32779,9 @@ tx0_64kw                        // April 1956 MIT TX-0 (64kw RAM)
 tx0_8kwo                        // 1959 MIT TX-0 (8kw RAM)
 tx0_8kw                         // 1962 MIT TX-0 (8kw RAM)
 
+@source:mit/cadr.cpp
+cadr
+
 @source:mits/altair.cpp
```
cuavas, line 32785: Please keep this file sorted – “c” comes before “t”.  Also I think you’ll get merge conflicts from the comments being stripped.

src/devices/cpu/cadr/cadr_dasm.cpp
```diff
@@ -0,0 +1,600 @@
...
+	return 1;
+}
+
+void cadr_disassembler::a_source(std::ostream &stream, u64 op)
+{
+	util::stream_format(stream, "a[%o]", ((op >> 32) & 0x3ff));
+}
+
+
+void cadr_disassembler::m_source(std::ostream &stream, u64 op)
```
cuavas, line 73: These helper functions that don’t depend on the disassembler object state at all can be turned into free functions in an anonymous namespace to keep implementation details out of the header.

src/devices/cpu/cadr/cadr.h
```diff
@@ -0,0 +1,105 @@
...
+	u32 m_ic; // interrupt control
+	u32 m_lc; // 32 bits, 26 bits + interrupt enable bits
+	u8 m_diag_mode;
+	bool m_access_fault;
+	bool m_write_fault;
+	bool m_page_fault;
+	bool m_popj;
+	bool m_interrupt_pending;
+
+	void program_map(address_map &map);
```
cuavas, line 83: Don’t forget `ATTR_COLD` on this.

src/devices/cpu/cadr/cadr.cpp
```diff
@@ -0,0 +1,1007 @@
...
+			LOGMASKED(LOG_TRACE, "MD <- %x(%o), start-write, MD=%x(%o)\n", m_md, m_md, m_md, m_md);
+			break;
+		case 0x1b: // MD, write map
+			m_md = output;
+			write_vma_map();
+			break;
+
+		case 0x19: // MD, start-read
+		default:
+		   fatalerror("%x(%o): output %02x not implemented", m_prev_pc, m_prev_pc, (m_ir >> 19) & 0x1f); break;
```
cuavas, line 618: There isn’t any point putting a `break` after a noreturn function (there are a few cases in this file).  If you want to make it more explicit that it’s leaving via an exception, you can change `fatalerror` to `throw emu_fatalerror`.

src/mame/mit/cadr_tv_control.h
```diff
@@ -0,0 +1,57 @@
...
+	u32 video_ram_read(offs_t offset);
+	void video_ram_write(offs_t offset, u32 data);
+	u32 tv_control_r(offs_t offset);
+	void tv_control_w(offs_t offset, u32 data);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_add_mconfig(machine_config &config) override;
```
cuavas, line 32: Please stop copy-pasting the “device-level overrides” comment – it makes no sense.  Also, `device_add_mconfig` should be marked cold as well.

src/mame/mit/cadr_tv_control.h
```diff
@@ -0,0 +1,57 @@
...
+	TIMER_CALLBACK_MEMBER(tv_60hz_callback);
+
+	required_device<screen_device> m_screen;
+	devcb_write_line m_irq_cb;
+	emu_timer *m_60hz_timer;
+	u32 m_status;
+	u32 m_sync_csr;
+	u32 m_sync_address;
+	std::unique_ptr<u32[]> m_video_ram;
+	std::unique_ptr<u8[]> m_sync_ram;
```
cuavas, line 54: Would it be more appropriate to use `memory_share_creator` for these memory areas?

src/mame/mit/cadr_tv_control.cpp
```diff
@@ -0,0 +1,161 @@
...
+60MHz and 64Mhz clocks are mentioned in the docs, as well as 64.69Hz default refresh rate.
+
+Implementation based on description of the operation, not schematics.
+
+
+**********************************************************************************/
+#include "emu.h"
+#include "cadr_tv_control.h"
+
+#include "screen.h"
```
cuavas, line 27: You already included `screen.h` in the corresponding header – there’s no need to have it here as well.

src/mame/mit/cadr_tv_control.cpp
```diff
@@ -0,0 +1,161 @@
...
+		{
+			const u32 d = m_video_ram[line_start + x];
+			const u32 xs = x * 32;
+
+			for (int i = 0; i < 32; i++)
+			{
+				bitmap.pix(y, xs + i) = BIT(d, i) ? white : black;
+			}
+		}
+	}
```
cuavas, line 74: Please restrict the update to the clipping rectangle, at least in the vertical direction.

src/mame/mit/cadr_iob.h
```diff
@@ -0,0 +1,115 @@
...
+	u16 read(offs_t offset);
+	DECLARE_INPUT_CHANGED_MEMBER(mouse_changed);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_add_mconfig(machine_config &config) override;
+	virtual const tiny_rom_entry *device_rom_region() const override;
+	virtual ioport_constructor device_input_ports() const override;
```
cuavas, line 36: Same as the other file – don’t copy-paste comments that don’t make sense, and there are several cold lifecycle member functions here that should be marked cold.

src/mame/mit/cadr_iob.h
```diff
@@ -0,0 +1,115 @@
...
+	static constexpr int CHAOSNET_TRANSMIT_DONE_BIT = 7;
+	static constexpr u16 CHAOSNET_TRANSMIT_DONE = 1 << CHAOSNET_TRANSMIT_DONE_BIT;
+	static constexpr int CHAOSNET_RESET_TRANSMIT_BIT = 8;
+	static constexpr u16 CHAOSNET_LOST_COUNT = 0x1e00;
+	static constexpr int CHAOSNET_RESET_BIT = 13;
+	static constexpr u16 CHAOSNET_RESET = 1 << CHAOSNET_RESET_BIT;
+	static constexpr int CHAOSNET_RECEIVE_DONE_BIT = 15;
+	static constexpr u16 CHAOSNET_RECEIVE_DONE = 1 << CHAOSNET_RECEIVE_DONE_BIT;
+
+	static constexpr u16 CHAOS_BUFFER_SIZE = 512 / 2;
```
cuavas, line 79: Please move constants that don’t need to be here into an anonymous namespace.

src/mame/mit/cadr_iob.cpp
```diff
@@ -0,0 +1,644 @@
...
+	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_R) PORT_CHAR('r') PORT_CHAR('R') // R f2 02 14
+	PORT_BIT(0x08, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_F) PORT_CHAR('f') PORT_CHAR('F') // F f2 02 16
+	PORT_BIT(0x10, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_V) PORT_CHAR('v') PORT_CHAR('V') // V f2 02 18
+	PORT_BIT(0x20, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("ALT LOCK") // f2 02 1a
+	PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_UNKNOWN) // f2 02 1c
+	PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_RIGHT) PORT_CHAR(UCHAR_MAMEKEY(RIGHT)) // f2 02 1e
+
+	PORT_START("KEY.2")
+	PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_LCONTROL) PORT_CHAR(UCHAR_SHIFT_2) // f2 02 20
+	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME(": ±") // f2 02 22
```
cuavas, line 108: Doe this actually produce colon and ± characters, and if so, is there a good reason not to use `PORT_CHAR` for this?

src/mame/mit/cadr_disk.h
```diff
@@ -0,0 +1,59 @@
...
+	auto irq_callback() { return m_irq_cb.bind(); }
+
+	u32 read(offs_t offset);
+	void write(offs_t offset, u32 data);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_add_mconfig(machine_config &config) override;
```
cuavas, line 30: Same comment as other device headers.

src/mame/mit/cadr_disk.cpp
```diff
@@ -0,0 +1,361 @@
...
+		command_list_w(data);
+		break;
+	case 0x06:
+		disk_address_w(data);
+		break;
+	case 0x07:
+		start_w(data);
+		break;
+	}
+}
```
cuavas, line 361: Would you be better off with a device address map function rather than these dispatch functions?

src/mame/mit/cadr_disk.cpp
```diff
@@ -0,0 +1,361 @@
...
+	LOG("Disk Controller status read\n");
+	return m_status;
+}
+
+
+u32 cadr_disk_device::command_list_r()
+{
+	LOG("Disk Controller command list pointer read\n");
+	return m_clp;
+}
```
cuavas, line 123: You should check for disabled side effects before logging in read handlers to avoid spam when debugger windows are open.

wilbertpol:
Will re-open when I have had time to look into this again.

cuavas:
Can we mark it as a draft rather than closing it?

wilbertpol:
> Can we mark it as a draft rather than closing it?

Ok, converted to draft.

src/mame/mit/cadr_tv_control.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+This currently emulates a 768x939 monochrome display.
+
+Other information about displays:
+
+Color screen: 454x576, 16 colors, RGB.
+Grayscale monitor: 454x576, 16 scales of gray.
+'CPT' monitor: 768x896, monochrome?
+
+60MHz and 64Mhz clocks are mentioned in the docs, as well as 64.69Hz default refresh rate.
```
cuavas, line 18: Is this referring to 60 MHz and 64 MHz input clocks, or is it supposed to be 60 Hz and 64 Hz? The juxtaposition of numbers in the megahertz with a typical screen refresh rate seems odd with no further explanation.

src/mame/mit/cadr.cpp
```diff
@@ -0,0 +1,271 @@
...
+	u8 *maincpu = memregion("maincpu")->base();
+	u8 *prom = memregion("proms")->base();
+
+	for (int i = 0; i < 0x200; i++)
+	{
+		for (int b = 0; b < 8; b++)
+		{
+			maincpu[((0x1ff - i) * 8) + b] = prom[(i * 8) + b];
+		}
+	}
```
cuavas, line 194: Could you include a comment explaining the purpose of copying the PROM over the CPU ROM?

src/devices/cpu/cadr/cadr.cpp
```diff
@@ -0,0 +1,1008 @@
...
+
+
+void cadr_cpu_device::execute_byte()
+{
+	// TODO Misc functions
+	u8 rotation = m_ir & 0x1f;
+	if (((m_ir >> 10) & 0x03) == 0x03) {
+		if (BIT(m_ic, 29))
+		{
+			rotation = rotation ^ ((BIT(m_lc, 1) ^ BIT(m_lc, 0)) << 4);
```
cuavas, line 871: Mixed K&R and Allman snuck in here.

cuavas (commented):
Good to see this coming along.  It looks like it’s getting to the point where it would make a decent “not working but not really a skeleton” driver to include in MAME.  Will you be marking it ready to consider merging soon?

src/mame/mit/cadr_iob.cpp
```diff
@@ -517,7 +517,7 @@ void cadr_iob_device::map(address_map &map)
 	map(0x04 << 2, 0x04 << 2).lrw16(NAME([this] {
 		m_speaker_data ^= 1;
 		m_speaker->level_w(m_speaker_data);
-		return u16(0xffff);
+		return 0xffff;
```
cuavas: You do want to either explicitly cast the result to `u16` or use syntax like `[this] -> u16 { ...` or you end up with an extra shim to adapt the `int`-return function to the `u16` return it needs.  Also, you’ll need a side effects check, since this is a read handler that has side effects.

src/devices/cpu/cadr/cadr_dasm.h
wilbertpol: Header cleaned up, the disassembler was moved to an anonymous namespace.

src/mame/mame.lst
wilbertpol: Moved to keep the file sorted.

src/devices/cpu/cadr/cadr_dasm.cpp
wilbertpol: Changed helper functions into normal functions

src/mame/mit/cadr_iob.h
wilbertpol: Moved to anonymous namespace.

src/mame/mit/cadr_tv_control.cpp
wilbertpol: Those 60MHz and 64MHz references are mentions of input clocks that I found. And also a separate mention of a 64.9Hz refresh rate. I had not found schematics of the video hardware.

src/mame/mit/cadr_disk.cpp
wilbertpol: Added side effects checks to the read handlers of this device.

src/mame/mit/cadr_disk.cpp
wilbertpol: Converted to a device address map

src/mame/mit/cadr_disk.h
wilbertpol: Removed comment and marked more member functions as cold.

src/mame/mit/cadr_iob.h
wilbertpol: Removed comment and marked more member functions as cold.

src/mame/mit/cadr_tv_control.cpp
wilbertpol: Limited the update to the cliprect vertical direction.

src/mame/mit/cadr_tv_control.cpp
wilbertpol: screen.h include removed from card_tv_control.cpp

src/mame/mit/cadr_tv_control.h
wilbertpol: Removed comment and marked more member functions as cold.

src/devices/cpu/cadr/cadr.h
wilbertpol: member function marked cold.

cuavas:
The current CI failure on this PR is due to needing to be updated due to @galibert separating the different screen device classes.

src/mame/mit/cadr_tv_control.h
wilbertpol: Updated to use memory _share_creator for m_video_ram and m_sync_ram.

src/mame/mit/cadr_iob.cpp
wilbertpol: Yes, see the key to left of the 1 key in picture 4 here: https://mitmuseum.mit.edu/collections/object/1991.003.003.1-.3
and there is also a key at the regular place that produces ; and :.

wilbertpol:
Before marking as for review I want to take another look at the disassembler to make it output something that looks more like the original microcode assembly.

The driver is showing some signs of life again:

wilbertpol:
> The current CI failure on this PR is due to needing to be updated due to @galibert separating the different screen device classes.

And also the retirement of rotl_32 in favor of std::rotl. Both have now been taken care of.

src/devices/cpu/cadr/cadr.cpp
```diff
@@ -0,0 +1,1016 @@
...
+		// -x------ PROG.RESET
+		// --x----- PROMDISABLE
+		// ---x---- TRAPENB
+		// ----x--- STATHENB
+		// -----x-- ERRSTOP
+		// ------xx SPEED 00 - extra slow, 01 - slow, 10 - normal, 11 - fast
+		m_diag_mode = data;
+		m_inst_view.select(BIT(m_diag_mode, 5));
+	})).umask32(0xffff);
+}
```
cuavas: This function’s brace positioning doesn’t match the rest of the file.

You should put a side effects check on log messages in read handlers to avoid spam when debugger windows are open.

Unit masks are generally easier to understand if you use enough digits for the whole width, e.g. `.umask32(0x0000'ffff)`.

It’s easier to read memory maps if all the addressing stuff is grouped, rather than split before/after handlers, e.g. `map(0x00 << 4, 0x07 << 4).umask32(0x0000'ffff).lrw16(...` rather than putting the unit mask at the end.

You should make the read handler explicitly return `u16` to avoid an extra adapter shim being instantiated (`return 0;` makes it return `int`).  Either `return u16(0);` or use `-> u16` syntax after the lambda parameter list.

src/devices/cpu/cadr/cadr.cpp
```diff
@@ -0,0 +1,1016 @@
...
+		fatalerror("%x(%o): dispatch misc function %d not implemented", m_prev_pc, m_prev_pc, (m_ir >> 10) & 0x03);
+	}
+
+	const u32 m = std::rotl(m_m, rotation) & dispatch_mask[(m_ir >> 5) & 0x07];
+	u32 index = ((m_ir >> 12) & 0x7ff) | m;
+
+	if ((m_ir >> 8) & 0x03) {
+		const u8 l1 = m_vma_map_l1[(m_md >> 13) & 0x7ff];
+		const u16 l2_index = (l1 << 5) | ((m_md >> 8) & 0x1f);
+		const u32 l2 = m_vma_map_l2[l2_index] & 0xffffff;
```
cuavas, line 813: More inconsistent brace positioning – a K&R snuck in here, too.

src/devices/cpu/cadr/cadr_dasm.cpp
```diff
@@ -0,0 +1,303 @@
...
+	"", " MAP-14", " MAP-15", " MAP-BOTH-14-AND-15"
+};
+
+
+void output(std::ostream &stream, bool &need_sp, const char * const string)
+{
+	if (need_sp) stream << " ";
+	stream << string;
+	need_sp = true;
+}
```
cuavas, line 83: It’s trivial and won’t get in the way of merging, but you can optimise streaming out a single character like `if (need_sp) stream << ' ';` (character rather than string) so it doesn’t need to do the length check, etc.  There are a number of those in this file.

src/mame/mit/cadr_iob.cpp
```diff
@@ -0,0 +1,709 @@
...
+{
+	return ROM_NAME(cadr_keyboard);
+}
+
+
+static INPUT_PORTS_START(keyboard)
+	PORT_START("KEY.0")
+	PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_UNKNOWN) // f2 02 00
+	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("Ⅱ")// f2 02 02
+	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("Ⅳ") // f2 02 04
```
cuavas: srcclean won’t like these – you need to do it like:
```c++
	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME(u8"\u2161") // f2 02 02 (Ⅱ)
	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME(u8"\u2163") // f2 02 04 (Ⅳ)
```

src/mame/mit/cadr_iob.cpp
```diff
@@ -0,0 +1,709 @@
...
+	PORT_BIT(0x08, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_L) PORT_CHAR('l') PORT_CHAR('L') // f2 02 76
+	PORT_BIT(0x10, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_STOP) PORT_CHAR('.') PORT_CHAR('>') // f2 02 78
+	PORT_BIT(0x20, IP_ACTIVE_HIGH, IPT_UNKNOWN) // f2 02 7a
+	PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_UNKNOWN) // f2 02 7c
+	PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CODE(KEYCODE_BACKSLASH) PORT_CHAR('`') PORT_CHAR('~') // f2 02 7e
+
+	PORT_START("KEY.8")
+	PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("MACRO") PORT_CODE(KEYCODE_F1) PORT_CHAR(UCHAR_MAMEKEY(F1)) // f2 02 80
+	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("Ⅰ") // f2 02 82
+	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME("Ⅲ") // f2 02 84
```
cuavas: Same with these, you need to change it to:
```c++
	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME(u8"\u2160") // f2 02 82 (Ⅰ)
	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_NAME(u8"\u2162") // f2 02 84 (Ⅲ)
```

src/mame/mit/cadr_iob.cpp
```diff
@@ -0,0 +1,709 @@
...
+	m_csr = 0;
+	m_microsecond_clock_buffer = 0;
+	m_clock = 0;
+	m_chaos_csr = CHAOSNET_TRANSMIT_DONE;
+	m_speaker_data = 0;
+	m_speaker->level_w(m_speaker_data);
+	m_chaos_transmit_pointer = 0;
+	m_chaos_receive_pointer = 0;
+	m_chaos_receive_size = 0;
+	m_chaos_receive_bit_count = 0xffff;
```
cuavas: You shouldn’t change `m_p1` and any other things that represent MCU outputs or inputs to the IOB device here.  The MCU etc. will push out updated values for them them on reset if they change.  Set the initial values in `device_resolve_objects` or `device_start`.

src/mame/mit/cadr_iob.cpp
```diff
@@ -0,0 +1,709 @@
...
+			return m_chaos_receive_bit_count;
+		}
+		return u16(0xfff);
+	})).umask32(0xffff);
+	map(0x15 << 4, 0x15 << 4).lr16(NAME([this] {
+		 // host number of this interface
+		chaos_transmit_start();
+		return u16(m_my_chaos_address->read());
+	})).umask32(0xffff);
+}
```
cuavas: Same comment about keeping addressing together applies to this address map.

src/mame/mit/cadr_disk.cpp
```diff
@@ -0,0 +1,339 @@
...
+		LOG("Start recalibrate\n");
+		m_status |= 5;
+		break;
+
+	case 0x008: // 0010 - Read and compare
+	case 0x004: // 0004 - Seek
+	default:
+		fatalerror("Unknown disk controller command %03x initiated\n", m_command & 0x3ff);
+		break;
+	}
```
cuavas, line 337: Break after a `noreturn` function is unreachable and may cause a warning with some compilers.  Please remove it.

src/mame/mit/cadr.cpp
```diff
@@ -0,0 +1,265 @@
...
+		, m_disk_controller(*this, "disk_controller")
+		, m_maincpu(*this, "maincpu")
+		, m_mainirq(*this, "mainirq")
+		, m_xbusirq(*this, "xbusirq")
+		, m_iob(*this, "iob")
+		, m_tv_control(*this, "display")
+	{ }
+
+	void cadr(machine_config &config);
+
```
cuavas, line 47: You can mark this function cold.

src/mame/mit/cadr.cpp
```diff
@@ -0,0 +1,265 @@
...
+	map(0766042 << 1, 0766042 << 1).lrw16(
+		NAME([this] () {
+			return m_interrupt_status;
+		}),
+		NAME([this] (u16 data) {
+			m_interrupt_status = (m_interrupt_status & ~0x83fc) | (data & 0x83fc);
+			if (!BIT(m_interrupt_status, 15))
+				m_mainirq->in_w<IRQ_SOURCE_UNIBUS>(CLEAR_LINE);
+		})
+	).umask32(0xffff);
```
cuavas: As with the other maps, please keep the addressing stuff together (`map`, `umask32`, `cswidth`) rather than split before/after handlers.

src/mame/mit/cadr.cpp
```diff
@@ -0,0 +1,265 @@
...
+	u8 *prom = memregion("proms")->base();
+
+	// The address signals to the PROMs are inverted.
+	for (int i = 0; i < 0x200; i++)
+	{
+		for (int b = 0; b < 8; b++)
+		{
+			maincpu[((0x1ff - i) * 8) + b] = prom[(i * 8) + b];
+		}
+	}
```
cuavas, line 194: The usual place for ROM decoding/unscrambling is in a driver init function rather than in start.  Probably best to follow the convention.

cuavas (changes requested):
This is looking good.  It’ll be great to have this finally merged.  Just a few relatively minor things I noticed reading through the code.

PR #13343, opened 2025-02-08
bus/rs232: Add the Cricket external sound box for the Apple IIc.

robjustice:
Emulation of the Street Electronics the Cricket for the Apple IIc. Disks and rom dump is pending to be added to the A2DP. Can be grabbed from here as well:  https://archive.org/download/cricket-6801-03-801-00/Cricket_6801_03-801-00.bin

### DIFF
--- a/src/devices/bus/rs232/cricket.cpp
+++ b/src/devices/bus/rs232/cricket.cpp
@@ -1,0 +1,244 @@
+// license:BSD-3-Clause
+// copyright-holders:R. Belmont, R Justice
+
+/*
+    cricket.cpp - Street Electronics The Cricket enulation
+    based on mboardd.cpp
+
+    Similiar to the Mockingboard D, this device was a little box that attached to the IIc's modem port.
+    It included:
+    - Music via 2 x AY 8913
+    - Speech via TMS5220
+    - Clock support. This is somewhat limited, The cricket will maintain the time once set, and is not powered off.
+
+    The microcontroller used in the Cricket is a MC6801 with a custom mask programmed rom
+    Some information available here:
+    http://mirrors.apple2.org.za/Apple%20II%20Documentation%20Project/Peripherals/Speech/Street%20Electronics%20The%20Cricket/
+
+
+    Ports hooked up as follows:
+
+    Port 1:
+    x-------  x  nc
+    -x------  x  nc
+    --x-----  x  nc
+    ---x----  x  nc
+    ----x---  x  nc
+    -----x--  x  nc
+    ------x-  O  LED
+    -------x  O  *RESET (To both AY's)
+
+    Port 2:
+    ---x----  O  UART TX
+    ----x---  I  UART RX
+    -----x--  I  pullup to +5v \
+    ------x-  I  pullup to +5v  MC6801 Mode 7
+    -------x  I  pullup to +5v /
+
+    Port 3:
+    x-------  O  DSR (to IIc)
+    -x------  I  DTR (from IIc)
+    --x-----  O  *CS2 (AY-3-8913 U8)
+    ---x----  O  *CS1 (AY-3-8913 U7)
+    ----x---  O  BC1 (Both AY-3-8913's)
+    -----x--  O  BDIR (Both AY-3-8913's)
+    ------x-  O  *WS (TMS5220)
+    -------x  O  *RS (TMS5220)
+	
+	SC1 (IS3) I  120Hz generated from the power pack input. Relies on the filtering being poor enough to extract the mains frequency. 
+
+    Port 4:
+    x-------  I/O  D7
+    -x------  I/O  D6
+    --x-----  I/O  D5
+    ---x----  I/O  D4
+    ----x---  I/O  D3
+    -----x--  I/O  D2
+    ------x-  I/O  D1
+    -------x  I/O  D0
+
+*/
+
+#include "emu.h"
+#include "cricket.h"
+
+#include "cpu/m6800/m6801.h"
+#include "sound/ay8910.h"
+#include "sound/tms5220.h"
+#include "speaker.h"
+#include "machine/timer.h"
+
+
+namespace {
+
+ROM_START(cricket)
+	ROM_REGION(0x800, "mcu", 0)
+	ROM_LOAD("03-801-00.bin", 0x000000, 0x000800, CRC(0a61e208) SHA1(9977a21522e28ae485009d3c31440fb4f7b9054b))
+ROM_END
+
+class cricket_device : public device_t, public device_rs232_port_interface
+{
+public:
+	cricket_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+	virtual void input_txd(int state) override;
+    virtual void input_dtr(int state) override;
+
+	required_device<m6801_cpu_device> m_mcu;
+	required_device<ay8913_device> m_ay1;
+	required_device<ay8913_device> m_ay2;
+	required_device<tms5220c_device> m_tms;
+
+protected:
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_add_mconfig(machine_config &config) override ATTR_COLD;
+	virtual const tiny_rom_entry *device_rom_region() const override ATTR_COLD;
+
+private:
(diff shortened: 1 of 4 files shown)


rb6502:
You'll need to change the short device name to fix "mame -validate" since "cricket" is already taken.  Maybe "secricket" or something similar.

I don't know how the clock works (looks like it's stored in the 6801's internal RAM?) but it might be interesting to also inherit device_rtc_interface so it can be auto-set to the host machine's time/date on startup.  I won't hold up a merge for that though.

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+*/
+
+#include "emu.h"
+#include "cricket.h"
+
+#include "cpu/m6800/m6801.h"
+#include "sound/ay8910.h"
+#include "sound/tms5220.h"
+#include "speaker.h"
+#include "machine/timer.h"
```
cuavas, line 69: Please put `speaker.h` after the stuff from src/devices, and sort the stuff from src/devices (`machine` before `sound`).

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+public:
+	cricket_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+	virtual void input_txd(int state) override;
+    virtual void input_dtr(int state) override;
+
+	required_device<m6801_cpu_device> m_mcu;
+	required_device<ay8913_device> m_ay1;
+	required_device<ay8913_device> m_ay2;
+	required_device<tms5220c_device> m_tms;
```
cuavas, line 90: These almost definitely shouldn’t be `public`.

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+	virtual const tiny_rom_entry *device_rom_region() const override ATTR_COLD;
+
+private:
+	u8 p2_r() { return m_rx_state<<3 | M6801_MODE_7; };
+	void p1_w(u8 data);
+	u8 p3_r(void);
+	void p3_w(u8 data);
+	u8 p4_r(void);
+	void p4_w(u8 data);
+	void ser_tx_w(int state) { output_rxd(state); }
```
cuavas, line 105: Do you need this trampoline function `set_tx_w`? Can you not use `output_rxd` directly?

Also, MAME convention in C++ is to use empty parentheses for functions (you aren’t even being consistent about that within this block).

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+	int m_rx_state;
+    int m_dtr_state;
+    u8 m_data;
+
+    TIMER_DEVICE_CALLBACK_MEMBER(clock_interrupt);
```
cuavas, line 111: Please keep member functions together, don’t put member variables between member functions.

Also, use explicitly sized types for things that need to be registered for save states, and initialise signals some time before `device_start` (it should be OK to initialise these three signals in the constructor or in `device_resolve_objects`).

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+const tiny_rom_entry *cricket_device::device_rom_region() const
+{
+	return ROM_NAME(cricket);
+}
+
+void cricket_device::device_start()
+{
+}
```
cuavas, line 155: Don’t forget to register data members for save states.

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+void cricket_device::input_txd(int state)
+{
+	m_rx_state = (state & 1);
+}
+
+void cricket_device::input_dtr(int state)
+{
+	m_dtr_state = (state & 1);
+}
```
cuavas, line 171: Convention for these is `state ? 1 : 0`.

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    
+    if (!BIT(data, 0)) //5220 read
+        m_data = m_tms->status_r();
+    
+    if (!BIT(data, 1)) //5220 write
+        m_tms->data_w(m_data);
+
+    output_dsr(!BIT(data, 7));
+}
```
cuavas, line 223: Should this be checking for edges on bits 5, 4, 1 and 0?  Also, remember the spaces after commas, and think about what to do if multiple devices are read simultaneously.

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+u8 cricket_device::p4_r()
+{
+	return m_data;
+}
+
+void cricket_device::p4_w(u8 data)
+{
+	m_data = data;
+}
```
cuavas, line 233: You seem to be using `m_data` for two things – the value being driven onto P4, and the value seen on P4.  This inevitably leads to issues.

cuavas (changes requested):

src/devices/bus/rs232/cricket.cpp
robjustice: It has the data bus for the three devices connected to port4. Its setting the direction/ddr appropriately for the port and either reading or writing the data bus value before then driving the device control lines via the port3. I had it like this as this is how it would be working for the real hardware. Is this something that needs separating more for how mame handles things?

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
robjustice: updated

src/devices/bus/rs232/cricket.cpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+        else if(!BIT(data, 3))
+            m_data = m_ay2->data_r();
+    }
+    
+    if (!BIT(data, 0)) //5220 read
+        m_data = m_tms->status_r();
+    
+    if (!BIT(data, 1)) //5220 write
+        m_tms->data_w(m_data);
```
cuavas, line 225: You really need to detect *active edges* on these data bits.  If the software writes the same value to the bit twice in a row, it won’t cause an additional write.  For example, just looking at the TMS5220:

```c++
    ...

    if (BIT(~data & m_p3_out, 0)) // 5220 read
        m_tms_data = m_tms->status_r();
    else if (BIT(data, 0)) // 5220 deselected
        m_tms_data = 0xff;

    if (BIT(~data & m_p3_out, 1)) //5220 write
        m_tms->data_w(/* calculate actual levels from all devices driving the lines here */);

    m_p3_out = data;
```

Also, what exactly are bits 4 and 5 connected to?  The AY-3-891x chips don’t have a chip select input as such.  They have BC1, BC2 and BDIR inputs, but the only time the chip doesn’t do anything is when BC1 and BDIR are both low.  If either BC1 *or* BDIR is high, the chip responds.  So I don’t see how the logic here can be correct, unless there are additional AND gates to force the corresponding BDIR low if bit 5 or bit 4 is low.

src/devices/bus/rs232/cricket.cpp
rb6502: I understand what you're saying but there is one piece of firmware running this thing that doesn't change and it works right now.  And what it's trying to do (read and write the AY registers) is discrete and well defined.  There is no need for the whole AND thing.

robjustice: Just wondering what we need to do to resolve this?

PR #13347, opened 2025-02-09
psx.xml: Added 12 working items + 4 redumped items

ArcadeShadow:
16 Tales 1 (USA) [Redump]
2Xtreme (Europe) [Redump]
3Xtreme (USA, demo) [Redump]
3D Lemmings (Europe) [Redump]
3D Lemmings (Japan) [Redump]
4-4-2 Soccer (Europe) [Redump]
40 Winks (Europe) [Redump]
40 Winks (Europe, demo) [Redump]
40 Winks (Italy) [Redump]
40 Winks (Spain) [Redump]
Cheesy (Europe) [Redump]
Cheesy (Japan) [Redump]

Redumped software list items
2Xtreme (USA) [Redump]
3Xtreme (USA) [Redump]
3D Lemmings (USA) [Redump]
40 Winks (USA) [Redump]


hash/psx.xml
```diff
@@ -260,8 +260,9 @@ For an overview of US PS1 discs that have not been dumped and added to redump's
...
 	</software>
 
-	<software name="2xtreme">
-		<!-- Original images (Redump)
+	<software name="2xtreme_us" cloneof="2xtreme">
+		<!--
+		http://redump.org/disc/2094/
 		<rom name="2Xtreme (USA).cue" size="700" crc="7e95393e" sha1="ed526fe2905f428530cc001038dd72c413bbe291"/>
 		<rom name="2Xtreme (USA) (Track 1).bin" size="45715824" crc="aad124a6" sha1="3f25e5d1366e60e70a416109711e109d48c9f19b"/>
 		<rom name="2Xtreme (USA) (Track 2).bin" size="56064624" crc="6f8d371b" sha1="9b986d87e8314c72757f629f6d7201bdeb0190b0"/>
```
angelosa, line 268: Let's open a discussion about this:
as part of their clearance http://redump.org/ gives this information already in the given URL. I legit don't think parsing XML comments as a pseudo-field is trivial, plus the verbosity of these track lines that don't even report back the msf to begin with, so it needs a roundtrip to the website anyway.
What if we just use an `<info name="redump.org" value="2094" />`, where of course 2094 is the ID that needs parsing thru clients including MAME UI itself?

ArcadeShadow: That's fair enough for me.
All the hash information is irrelevant, with the source and ID indicated.

etabeta78:
I had started adding these hashes to psx protos and segacd/megacd mostly to
make easier to go back and re-create CHDs if necessity arose... with redump
dumps, the ID is sufficient to recover the necessary data :-)
As such I think the site ID is enough, but I think it is mostly appropriate
as a comment field

my 2c
    Fabio

hash/psx.xml
```diff
@@ -370,36 +376,65 @@ For an overview of US PS1 discs that have not been dumped and added to redump's
...
 		-->
 		<description>3Xtreme (USA)</description>
 		<year>1999</year>
-		<publisher>989 Studios</publisher>
+		<publisher>989 Sports</publisher>
+		<notes>Compatibility set as PAL since it didn't work on NTSC system</notes>
+		<info name="developer" value="989 Sports" />
+		<info name="language" value="English" />
 		<info name="serial" value="SCUS-94231"/>
+		<sharedfeat name="compatibility" value="PAL-E"/>
```
angelosa, line 393: This doesn't really make sense. At worst you should just drop the `sharedfeat` instead, it clearly sounds as something that needs investigating (and inclined to `supported="no"` here).

angelosa:
> As such I think the site ID is enough, but I think it is mostly appropriate
as a comment field

I think it's appropriate in the need of not looking at a XML file for something that a frontend can optionally parse and/or allow user to click on it. That's valid for good external source of truth as well, not necessarily redump.org alone.

hash/psx.xml
ArcadeShadow: Changed flag to supported="no" and change to compatibility to NTSC-U

angelosa (approved):

PR #13349, opened 2025-02-09
Added support for Roland PRO-E Arranger

giulioz:
Kinda works, timers are flaky, requires external MT-32 emulation (MUNT works).
Test mode for RAM and ROM passes.

Screenshot 2025-02-09 at 22 12 55

### DIFF
--- a/src/devices/cpu/mc68hc11/mc68hc11.cpp
+++ b/src/devices/cpu/mc68hc11/mc68hc11.cpp
@@ -408,20 +408,26 @@ void mc68hc11_cpu_device::serial_w(u8 val)
 // unimplemented
 uint8_t mc68hc11_cpu_device::baud_r()
 {
+	if (!machine().side_effects_disabled())
+		logerror("HC11: BAUD serial register read\n");
 	return m_baud;
 }
 void mc68hc11_cpu_device::baud_w(uint8_t data)
 {
+	logerror("HC11: BAUD serial register write %02x\n", data);
 	m_baud = data;
 }
 
 // unimplemented
 uint8_t mc68hc11_cpu_device::sccr1_r()
 {
+	if (!machine().side_effects_disabled())
+		logerror("HC11: SCCR1 serial register read\n");
 	return m_sccr1;
 }
 void mc68hc11_cpu_device::sccr1_w(uint8_t data)
 {
+	logerror("HC11: SCCR1 serial register write %02x\n", data);
 	m_sccr1 = data;
 }
 
@@ -449,13 +455,15 @@ void mc68hc11_cpu_device::scsr_w(uint8_t data)
 
 uint8_t mc68hc11_cpu_device::scdr_r()
 {
-	set_irq_state(0x14, false);
-	m_scsr |= 0b10000000;
+	if (!machine().side_effects_disabled())
+		set_irq_state(0x14, false);
+
+	m_scsr |= 0x80;
 	return m_scdr;
 }
 void mc68hc11_cpu_device::scdr_w(uint8_t data)
 {
-	m_scsr &= 0b01111111;
+	m_scsr &= 0x7f;
 	if ((m_sccr2 & 0x80) != 0)
 		set_irq_state(0x14, true);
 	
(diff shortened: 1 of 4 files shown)


happppp:
You forgot to initialize m_baud. If the CPU manual says it's unmodified at reset, do m_baud = 0 at device_start.
Also, the new variables each need a save_item.

giulioz:
@happppp good point, fixed it now

happppp:
Same thing in the driver by the way: If you want savestates to work, add a save_item to machine_start for keyscan and lcd mode.
Also, MAME drivers add m_ prefix to class variables.

giulioz:
@happppp 
Should be all done, took note for my future contributions, thanks!

src/devices/cpu/mc68hc11/mc68hc11.cpp
```diff
@@ -384,6 +397,71 @@ void mc68hc11_cpu_device::tmsk2_w(uint8_t data)
...
+}
+void mc68hc11_cpu_device::scsr_w(uint8_t data)
+{
+	m_scsr = data;
+}
+
+uint8_t mc68hc11_cpu_device::scdr_r()
+{
+	set_irq_state(0x14, false);
+	m_scsr |= 0b10000000;
```
angelosa, line 453: This needs to be protected against `!machine().side_effects_disabled()`, being a read access.

src/devices/cpu/mc68hc11/mc68hc11.cpp
```diff
@@ -384,6 +397,71 @@ void mc68hc11_cpu_device::tmsk2_w(uint8_t data)
...
+	return m_scdr;
+}
+void mc68hc11_cpu_device::scdr_w(uint8_t data)
+{
+	m_scsr &= 0b01111111;
```
angelosa, line 458: I definitely don't like using the binary notation for this kind of manipulation, it's imo unreadable and bound to mistake.

src/devices/cpu/mc68hc11/mc68hc11.cpp
```diff
@@ -384,6 +397,71 @@ void mc68hc11_cpu_device::tmsk2_w(uint8_t data)
...
+uint8_t mc68hc11_cpu_device::baud_r()
+{
+	return m_baud;
+}
+void mc68hc11_cpu_device::baud_w(uint8_t data)
+{
+	m_baud = data;
+}
+
+// unimplemented
```
angelosa, line 418: If it's "unimplemented" then it needs logging in the handler, not being concealed like this (nobody will see it essentially).

giulioz:
@angelosa good point, just actioned your feedback. Thanks!

src/devices/cpu/mc68hc11/mc68hc11.cpp
```diff
@@ -384,6 +397,79 @@ void mc68hc11_cpu_device::tmsk2_w(uint8_t data)
...
+	m_scsr = data;
+}
+
+uint8_t mc68hc11_cpu_device::scdr_r()
+{
+	if (!machine().side_effects_disabled())
+		set_irq_state(0x14, false);
+
+	m_scsr |= 0x80;
+	return m_scdr;
```
galibert, line 462: You shouldn't change the register on read.  If I understand what you're doing correctly, just return m_scsr | 0x80.

src/mame/roland/roland_proe.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+void roland_proe_state::proe(machine_config &config)
+{
+	XTAL main_clock = 16_MHz_XTAL;
+
+	MC68HC11A0(config, m_cpu, main_clock / 2); // MC68HC11A0P
+	m_cpu->set_addrmap(AS_PROGRAM, &roland_proe_state::proe_map_a);
+	m_cpu->in_pa_callback().set(FUNC(roland_proe_state::pa_r));
+	m_cpu->out_pa_callback().set(FUNC(roland_proe_state::pa_w));
+	m_cpu->serial_tx_cb().set(FUNC(roland_proe_state::int_midi_w));
```
galibert, line 323: We really, really don't want to hint that serial-faked-as-parallel is acceptable in new code.  Since you don't actually do anything in int_midi_w, please remove the callback until real serial is implemented.

PR #13350, opened 2025-02-10
sony/news_68k_iop.cpp: Implement NWS-831 driver

briceonk:
Let me know if any of this should be split out into a separate PR and I will do so.

Highlights:
- Working dual-processor system that can boot both NEWS-OS 2 and NEWS-OS 4
- Basic peripherals work (serial, SCSI, networking kinda)
- Lots of custom stuff that has been partially reverse-engineered like the front panel and the custom MMU

Lowlights:
- Networking stability is bad
- The amount of custom stuff means that this driver has a lot more educated guesses in it than the NEWS r3k, r4k, and single-processor 68k drivers do.
- NEWS-OS 4 installer uses unemulated tape drive commands, so the OS install process doesn't work right now. Instead, I used a script I wrote a while back to generate an image from the bitsavers dump of NEWS-OS 4.1C: https://github.com/briceonk/news-os/tree/master/src/news-inst
- The SCSI DMA setup is really hacky (just like the Macintosh stuff) because `iopboot` uses the features of the 68k that allow it to use narrower buses to do longword transactions

Front panel traced schematic:
image


happppp:
As you can see, MAME doesn't have other brands of chips unless there are differences to emulate.
For example, see there's no HD68000 or TMP68000
Or for exact types, no MC68000P8 or MC68000P12

If you want to document exact brands/types of chips used, you could add hardware notes or a PCB diagram to the driver.

briceonk:
> As you can see, MAME doesn't have other brands of chips unless there are differences to emulate. For example, see there's no HD68000 or TMP68000 Or for exact types, no MC68000P8 or MC68000P12
> 
> If you want to document exact brands/types of chips used, you could add hardware notes or a PCB diagram to the driver.

Should be fixed now.

src/mame/mame.lst
```diff
@@ -43378,6 +43378,9 @@ librie
 @source:sony/news_38xx.cpp
 nws3860
 
+@source:sony/news_68k_iop.cpp
+nws831
+
 @source:sony/news_68k.cpp
 nws1580
```
cuavas, line 43385: news_68k_iop.cpp should come after news_68k.cpp

src/mame/sony/news_020_mmu.h
```diff
@@ -0,0 +1,108 @@
...
+
+protected:
+	virtual void device_start() override;
+	virtual space_config_vector memory_space_config() const override;
+
+private:
+	// Helper class for reading page table entries
+	struct news_020_pte
+	{
+		news_020_pte(uint32_t pte_bits) : pte(pte_bits) {}
```
cuavas, line 38: This doesn’t need to be in the header – you can throw it in an anonymous namespace in the .cpp file so it isn’t unnecessarily compiled downstream.

src/mame/sony/news_020_mmu.h
```diff
@@ -0,0 +1,108 @@
...
+	bool m_enabled = false;
+	bool m_romdis = false; // ROMDIS bit controls if ROM is mapped to the start of address space or not
+
+	std::unique_ptr<u32[]> m_mmu_user_ram;
+	std::unique_ptr<u32[]> m_mmu_system_ram;
+	std::unique_ptr<u32[]> m_mmu_user_tag_ram;
+	std::unique_ptr<u32[]> m_mmu_system_tag_ram;
+
+	void clear_user_entries();
+	void clear_kernel_entries();
```
cuavas, line 103: Please keep data members and member functions grouped.  Don’t mix member functions and data members.

Please don’t put member initialisations in the header when the constructor is defined in a different file.  I’ve already had to clean up multiple cases where there were conflicting initialisations in the class declaration and the constructor.

src/mame/sony/news_020_mmu.cpp
```diff
@@ -0,0 +1,298 @@
...
+			  tag,
+			  modified_entry.pfnum(),
+			  modified_entry.valid(),
+			  modified_entry.modified(),
+			  modified_entry.fill_on_demand(),
+			  modified_entry.kernel_writable(),
+			  modified_entry.kernel_readable(),
+			  modified_entry.user_writable(),
+			  modified_entry.user_readable());
+#endif
```
cuavas, line 163: Do you really need the `#if` here?  The compiler optimises away `LOGMASKED` to nothing if it’s disabled, and the only side effect of construction of the `news_020_pte` is a RAM access, which can be eliminated as the pointers aren’t volatile or atomic or anything like that.  Using the `#if` means the code is liable to rot because the compiler never sees it when it’s disabled.

src/mame/sony/news_020_mmu.cpp
```diff
@@ -0,0 +1,298 @@
...
+		// TODO: What is the actual physical address width? The 0x1 part might also be an MMU control bit
+		offset = offset & (0x1fffffff >> 2);
+		if (!m_romdis && offset < ((0xffff - 0x100) >> 2))
+		{
+			result = this->space(0).read_dword((0x03000000 + 0x100) + (offset << 2), mem_mask);
+		}
+		else
+		{
+			result = this->space(0).read_dword(offset << 2, mem_mask);
+		}
```
cuavas, line 180: Can you at least use a `memory_access::specific<...>` to avoid going through the slowest path for accessing the address space?

src/mame/sony/news_iop_scsi.h
```diff
@@ -0,0 +1,84 @@
...
+	// read/write handlers
+	u8 read_wrapper(bool pseudo_dma, offs_t offset);
+	void write_wrapper(bool pseudo_dma, offs_t offset, u8 data);
+
+	void drq_w(int state);
+	void irq_w(int state);
+
+	// Constants for consumers of the bus error details
+	static const u8 READ_ERROR = 1;
+	static const u8 WRITE_ERROR = 0;
```
cuavas, line 36: Please use `static inline constexpr` for these kinds of things.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+#include "machine/am79c90.h"
+#include "machine/msm58321.h"
+#include "machine/ncr5380.h"
+#include "machine/nscsi_bus.h"
+#include "machine/pit8253.h"
+#include "machine/ram.h"
+#include "machine/timekpr.h"
+#include "machine/upd765.h"
+#include "machine/z80scc.h"
+#include "machine/bankdev.h"
```
cuavas, line 88: The stuff from `formats` is supposed to come after the stuff from `devices`.  Also, please sort the headers within each block.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+        virtual void machine_reset() override;
+
+        // address maps
+        void iop_map(address_map &map);
+        void iop_autovector_map(address_map &map);
+        void mmu_map(address_map &map);
+        void cpu_map(address_map &map);
+
+        // machine config
+        void common(machine_config &config);
```
cuavas, line 150: Please put `ATTR_COLD` on cold lifecycle functions (here and in devices, too).

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+        uint32_t m_cpu_inten = 0;
+        std::map<int, bool> m_cpu_int_state = {
+            {INPUT_LINE_IRQ1, false},
+            {INPUT_LINE_IRQ2, false},
+            {INPUT_LINE_IRQ3, false},
+            {INPUT_LINE_IRQ4, false},
+            {INPUT_LINE_IRQ5, false},
+            {INPUT_LINE_IRQ6, false},
+            {INPUT_LINE_IRQ7, false}
+        };
```
cuavas, line 285: Do you really need `std::map` for these?  It’s going to make save state support a pain.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+
+    const std::map<int, news_iop_state::cpu_irq_number> news_iop_state::cpu_irq_line_map = {
+        // AST is excluded from this check
+        {INPUT_LINE_IRQ2, CPIRQ1},
+        {INPUT_LINE_IRQ3, IOPIRQ3},
+        {INPUT_LINE_IRQ4, CPIRQ3},
+        {INPUT_LINE_IRQ5, IOPIRQ5},
+        {INPUT_LINE_IRQ6, TIMER},
+        {INPUT_LINE_IRQ7, PERR}
+    };
```
cuavas, line 322: Please don’t use static variables that require runtime allocation.  It slows down startup for *all* situations.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+    void news_iop_state::rtcsel_w(uint8_t data)
+    {
+        LOGMASKED(LOG_RTC, "rtc sel w 0x%x\n", data);
+        m_rtc->cs1_w(1);
+        m_rtc->cs2_w(1);
+
+        m_rtc->d0_w((data & 0x1) > 0);
+        m_rtc->d1_w((data & 0x2) > 0);
+        m_rtc->d2_w((data & 0x4) > 0);
+        m_rtc->d3_w((data & 0x8) > 0);
```
cuavas, line 469: Please use the `BIT` helper for extracting bits.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+            // Update input pin status if it has changed
+            const bool state = active_irq & (1 << irq.second);
+            if (m_cpu_int_state[irq.first] != state) {
+                if (irq.first != INPUT_LINE_IRQ6)
+                {
+                    LOGMASKED(LOG_INTERRUPT, "Setting CPU input line %d to %d\n", irq.first, state ? 1 : 0);
+                }
+                m_cpu_int_state[irq.first] = state;
+                m_cpu->set_input_line(irq.first, state ? 1 : 0);
+            }
```
cuavas, line 807: Please be consistent with K&R vs Allman formatting.

src/mame/sony/news_68k_iop.cpp
```diff
@@ -0,0 +1,1150 @@
...
+        int_check_cpu();
+    }
+
+    void news_iop_state::int_check_cpu()
+    {
+        const int active_irq = m_cpu_intst & m_cpu_inten;
+        for (auto irq : cpu_irq_line_map)
+        {
+            // Update input pin status if it has changed
+            const bool state = active_irq & (1 << irq.second);
```
cuavas, line 799: This is another place to use the `BIT` helper.

cuavas (changes requested):

briceonk:
I figured out the cause of the monitor ROM hang after I synchronized with master and fixed it, so I will work on your code review feedback next. Thanks for the review! :)

src/mame/sony/news_68k_iop.cpp
briceonk: No, I wrote this in a way that made sense to me when I was untangling the messy code I originally had from experimenting with the IRQ setup for the IOP. But I didn't think about save state support, so I will go back to the drawing board on this one.

src/mame/sony/news_020_mmu.h
briceonk: Should be fixed now - if the grouping for functions should be completely separate (i.e. have its own `private:` block along with the functions in `protected` and `public`), let me know and I'll change it.

src/mame/sony/news_020_mmu.cpp
briceonk: It wasn't needed, I wasn't thinking of the compiler optimizations. I removed it.

src/mame/sony/news_68k_iop.cpp
briceonk: I added it in a bunch of places, if there is anything else that should have it let me know :)

briceonk:
I will address the remaining comments this week. Didn't have a chance to get to all of them this weekend.

src/mame/sony/news_68k_iop.cpp
briceonk: Eliminated this part entirely, I don't know what I was thinking. Guess I missed that `input_state` existed :D

src/mame/sony/news_68k_iop.cpp
briceonk: Changed to initialization in the ctor instead

briceonk:
@cuavas  @pmackinlay I think I resolved all of the feedback, so this should be ready for re-review.

src/mame/sony/news_68k_iop.cpp
cuavas: As an optimisation, you might be able to get away with turning them into some kind of static `std::array` variables and then using `std::lower_bound` to find things.

PR #13351, opened 2025-02-10
oberheim/xpander.cpp: Added driver for Xpander.

m1macrophage:
New system marked not working: Oberheim Xpander.

Digital and digital-analog interface almost fully implemented. No audio. The layout is minimal, just includes the displays and a few buttons.

### DIFF
--- a/src/mame/oberheim/xpander.cpp
+++ b/src/mame/oberheim/xpander.cpp
@@ -15,11 +15,11 @@ low), and a control input ("chain advance", active low).
 
 The Xpander has two 6809-based computers. The main computer scans for button
 presses, drives the VFDs, interprets the digital and analog inputs (MIDI, CVs,
-triggers, etc)., and sends parameters to the voice computer.
+triggers, etc.), and sends parameters to the voice computer.
 
 The voice board consists of 6 analog voices controlled by the voice computer.
 
-Each voice is built around a CEM3347 (dual oscillator) and a CEM3372 (VCF and
+Each voice is built around a CEM3374 (dual oscillator) and a CEM3372 (VCF and
 VCA). VCO2 can modulate either VCO1 or the VCF to produce FM effects. There is
 also circuitry to generate pulse waves out of the saw ones, and to mix in noise.
 The noise source is shared for all voices. The CEM3372 is combined with a 4051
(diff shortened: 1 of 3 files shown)


src/mame/oberheim/xpander.cpp
```diff
@@ -0,0 +1,1151 @@
...
+inputs, 6 gate and 1 trigger inputs (user-configurable as active high or active
+low), and a control input ("chain advance", active low).
+
+The Xpander has two 6809-based computers. The main computer scans for button
+presses, drives the VFDs, interprets the digital and analog inputs (MIDI, CVs,
+triggers, etc)., and sends parameters to the voice computer.
+
+The voice board consists of 6 analog voices controlled by the voice computer.
+
+Each voice is built around a CEM3347 (dual oscillator) and a CEM3372 (VCF and
```
galibert, line 22: CEM 3374, not 3347 :-)

src/mame/oberheim/xpander.cpp
```diff
@@ -0,0 +1,1151 @@
...
+
+	static constexpr const int WRAP_BUFFER = 3;
+	const bool overflowed = newval <= WRAP_BUFFER &&
+	           oldval >= NUM_ENCODER_POSITIONS - WRAP_BUFFER;
+	const bool underflowed = newval >= NUM_ENCODER_POSITIONS - WRAP_BUFFER &&
+	           oldval <= WRAP_BUFFER;
+	m_encoder_dir[encoder] = ((newval > oldval) || overflowed) && !underflowed;
+
+	LOGMASKED(LOG_ENCODERS, "Encoder %d changed from: %d to: %d (o: %d, u: %d), dir: %d\n",
+	          param, oldval, newval, overflowed, underflowed, m_encoder_dir[encoder]);
```
galibert, line 957: vector<bool> is stupidly weird in C++ (it's essentially required to be a bitvector) and that breaks logerror.  You probably need to go bool(m_encoder_dir[encoder]) to pacify the CI.

src/mame/oberheim/xpander.cpp
m1macrophage: Oops, fixed.

src/mame/oberheim/xpander.cpp
m1macrophage: Thanks, fixed. Seems to only affect some setups (built fine on mine).

But I ran into a related issue when trying to implement save states (this driver uses a lot of vector<bool>), which is why save states are deferred to a future PR.

cuavas: Do you really need to use vectors all the time?  Can you not use `std::array<bool, N>` or a plain C array?

m1macrophage: Done in #13426.

PR #13353, opened 2025-02-10
ddenlovr.cpp: added DIPs' definitions for hnrose

Osso13:


Osso13:
Please don't merge until Vas has time to check this

cuavas:
I had a look at this, and the game definitely isn’t working.  If you win a game, it won’t recognise the double-up or take score button, even though these buttons are recognised in test mode.  It will time out waiting for one of these buttons, and then start sounding an alarm with a “CALL MASTER” error displayed.  The only way to clear the error is to clear memory.  Medal mode doesn’t seem to work properly either – it doesn’t seem to be trying to drive the hopper.  Possibly protection checks failing so it’s intentionally misbehaving?

Some of the DIP switches definitely don’t match what the manual says, it must be for a different version of the game.  Also, it definitely needs the fake “auto bet” input to be in the right state to behave properly.

cuavas:
I updated the inputs as best as I could to match what the game actually does, leaving the ones from the manual for things that can’t easily be verified.  I changed several to unknown, particularly on DIP-SW3 as it’s pretty clear the manual doesn’t match this version of the game.

I don’t know if you want to try fixing the game of if you just want to leave it not working.  The symptoms are kind of similar to Hanafuda Hana Gokou and Hanafuda Gokou.

Osso13:
I don't think I stand much of a chance to fix it, so I guess for now better inputs / dips definitions will have to do. Thanks!

Gently28:
[NIGHT ROSE.pdf](https://github.com/user-attachments/files/18775235/NIGHT.ROSE.pdf)

cuavas:
> [NIGHT ROSE.pdf](https://github.com/user-attachments/files/18775235/NIGHT.ROSE.pdf)

Thanks, we already have a copy of that manual.  Unfortunately it’s for a different version of the game.  It also won’t help fix the emulation.

PR #13356, opened 2025-02-11
Adding machine IBM System/23 Datamaster (system23.cpp, non-working); its keyboard; corrections to Intel 8275

RetroAND:
Good afternoon,

I have written the System/23 Datamaster driver. It is non working because there are unimplemented parts of the computer (DMA banking, character ROM banking, etc.) and also has some bugs (randomly crashing at test 19 or giving sporadical ROM errors for uninstalled memories). Other than that, the emulated machine works enough to have BASIC running correctly.

To make it work, some corrections were made to the Intel 8275 (8275.h, 8275.cpp) in order to emulate both its behaviour and also to simulate the light pen, which this computer uses to test the component.

If there is any issue with this driver, please let me know.

Thank you for your attention


src/devices/video/i8275.cpp
```diff
@@ -36,7 +35,7 @@
 
 #include "screen.h"
 
-//#define VERBOSE 1
+#define VERBOSE 1
```
angelosa, line 38: Please disable logging in production.
(should really use logmacro better but for now just disabling it is enough).

RetroAND: Done.

goldnchild:
Just wanted to mention in this thread that I did a rewrite of the keyboard to make it look more like the pcxt83 keyboard.  They have many similarities.

https://github.com/goldnchild/mame/tree/rewrite_sys23_keyboard

RetroAND:
Hello,

Do not accept this pull up yet. I have found a difference of behaviour between the stock 8275 and the CRTC the Datamaster uses. Since I modified the i8275 to accomodate changes to make the Datamaster work, I have to roll them back and correct them to add this new variant of the component.

Sorry for the hassle

angelosa:
Well, you just put the PR in draft if you don't want it to be merged just yet. Also it doesn't compile anyway.

PR #13367, opened 2025-02-15
oberheim/dmx.cpp: Added metronome and other audio improvements.

m1macrophage:
* Emulating metronome, including mixing with the rest of the voices.
* Emulating the RC HPFs for each voice and for the metronome.
* Corrected (inverted) phase of mixed audio.
* Using component values from electrongate.com, they seem to be more accurate than the '81 schematics.

Almost every component (resistors, capacitors) in the voice (but not the metronome) audio path is now accounted for. In other words, almost all components show up in some equation or filter configuration.

### DIFF
--- a/src/mame/oberheim/dmx.cpp
+++ b/src/mame/oberheim/dmx.cpp
@@ -44,10 +44,8 @@ PCBoards:
 - 2 x cymbal cards: A single voice that occupies two card slots.
 
 Known audio inaccuracies (reasons for MACHINE_IMPERFECT_SOUND):
-- Closed and Accent hi-hat volume variations might be wrong. See comments in
-  HIHAT_CONFIG (to be researched soon).
-- Uncertainty on component values for PERC2 (see comments in PERC_CONFIG).
-- No metronome yet.
+- Some uncertainty on component values for HIHAT and PERC2 (see comments in
+  HIHAT_CONFIG and PERC_CONFIG).
 - Linear- instead of audio-taper faders.
 - Envelope decay ignores diodes in capacitor discharge path. Given the quick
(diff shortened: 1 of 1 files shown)


galibert:
When will you consider it's not IMPERFECT_SOUND anymore? :-)

m1macrophage:
Looking for guidance on that.

On one hand, there is no longer any missing audio functionality [*]. Unless there are bugs, optimizations or new information that comes up, I don't expect any changes. So from this perspective, it should not be IMPERFECT_SOUND.

On the other hand, and depending on one's audiophile level : ) , I don't know if this could ever sound "identical" to the real machine. So it might always be imperfect from this perspective.

Let me know how I should proceed.


[*] Newer versions of the DMX shipped with newer voice cards, and those are not yet emulated. But that would only be an imperfection for the Memory Expansion and MIDI versions, which are not yet emulated.

angelosa:
The guidance is: mark a driver as `MACHINE_IMPERFECT_<something>` in cases where you _don't_ want bugs being reported in that specific area at MAMEtesters or here with GH issues. 
I gather this will need an experienced audiophile user to tell you if something is wrong so it should be removed, but obviously the decision is yours.

angelosa:
Also: bus slot options have their own flag system in the form of `static constexpr feature_type unemulated_features()` and `imperfect_features()`, the list of value is in [emu/device.h](https://github.com/mamedev/mame/blob/84c8c566b1e9db1028219dca4a6b2c80e1e01fc1/src/emu/device.h#L66)

m1macrophage:
Thanks for the info, followed up in #13370.

PR #13371, opened 2025-02-16
New working clone added

algestam:
Chequered Flag (Konami handheld) [algestam]

The game is presumed to have the same ROM as NASCAR racing. Haven't marked this driver a clone though, please advise if that is better and I'll update the driver.


happppp:
Yeah, just make it a clone like Mickey Mouse & Egg.
And could you rename it to kchqflag to be consistent with the arcade set?

algestam:
Thanks for the feedback! I'll address this and update the PR.

PR #13376, opened 2025-02-16
pc/sis630.cpp: Add notes and ROM placeholder for GameCristal JAMMA interface PCB

clawgrip:


src/mame/pc/sis630.cpp
```diff
@@ -17,14 +17,14 @@ TODO (main):
 - PS/2 mouse is unstable, worked around by disabling and using a serial mouse instead.
 
 TODO (usability, to be moved in a SW list):
-- windows xp sp3: tests HW then does an ACPI devtrap write ($48), will eventually BSoD with
+- Windows XP SP3: tests HW then does an ACPI devtrap write ($48), will eventually BSoD with
```
angelosa, line 20: Lowercase was intentional: all these entries needs to be ported over to SW list anyway ...

src/mame/pc/sis630.cpp
```diff
@@ -400,8 +400,24 @@ ROM_START(zidav630e)
...
+  |                                     | MCU           |  |  |  |
+  |                                     |_______________|  |__|  |
+  |                                      _________   Xtal        |
+  |                                     |DIPS x 8|   5 MHz       |
+  |                                                              |
+  |    Test sw->(o)         _________   _________   _________    |
+  | Service sw->(o)         SN74HC245N  SN74HC245N  SN74HC245N   |
+  |_________                                          ___________|
+           |_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|
+                          JAMMA
```
angelosa, line 420: Should be at the bottom of the specific header section, not here.

src/mame/pc/sis630.cpp
clawgrip: Fixed!

src/mame/pc/sis630.cpp
clawgrip: Fixed!

PR #13379, opened 2025-02-17
taito/taitojc.cpp: Fix character RAM size

cam900:
This PR fixes regression since https://github.com/mamedev/mame/commit/1c139c5a60440db98b38e7eb11b21137ada9a6bf .


happppp:
Is this a workaround for an array OOB bug somewhere?
Before and after your cleanups, m_char_ram was 16KB, and you're increasing it to 64KB.

cam900:
It's fixes glitchy text character, and fixes wrong size of character RAM. (12KB at before this PR, but actually 16KB and top 4KB area is unused)

happppp:
Ah right, memory_share_creator param is bytes, not length.
Isn't the correct fix this then?:

taitojc_v.cpp
line 18: change 0x80 to 0x60
line 116: change if to else if

cam900:
In POST screen, some area uses tile 0x60 for 'empty' tile in tilemap, so it's possibly OOB issue may occur if total tile count is changed.

happppp:
Ok I'll take your word for it.

PR #13380, opened 2025-02-17
k053936.cpp: invert blending, reduce code duplication

FredYeye:
The quest to improve blending on Konami systems continue.

A rather simple change once again: blending appears to be performed backwards in k053936.cpp's `K053936GP_copyroz32clip()`. I tried a few games that use this chip, but I only observed transparency in two games: `mmaulers` (planet fade-in in the intro) and `gaiapols` (water, glass floor). The visual changes are pretty minor from what I've seen, but it's a step forward nonetheless.

I also combined solid / alpha drawing paths to reduce duplicate code.

Another thing I tried was to convert the do loops to more conventional for loops, along with the usual nailing everything down with const and more appropriate types. This lost some performance, so decided to only commit the bare minimum for now.

### DIFF
--- a/src/devices/video/k053936.cpp
+++ b/src/devices/video/k053936.cpp
@@ -605,95 +605,48 @@ static inline void K053936GP_copyroz32clip( running_machine &machine,
 	cy = starty;
 	cx = startx;
 
-	if (blend > 0)
-	{
-		dst_ptr += dst_pitch;      // draw blended
-		starty += incyy;
-		startx += incyx;
+	dst_ptr += dst_pitch;
+	starty += incyy;
+	startx += incyx;
 
+	do {
 		do {
-			do {
-				int srcx = (cx >> 16) & 0x1fff;
-				int srcy = (cy >> 16) & 0x1fff;
-				int pixel;
-				uint32_t offs;
-				offs = srcy * src_pitch + srcx;
+			int srcx = (cx >> 16) & 0x1fff;
+			int srcy = (cy >> 16) & 0x1fff;
+			int pixel;
+			uint32_t offs;
+			offs = srcy * src_pitch + srcx;
 
-				cx += incxx;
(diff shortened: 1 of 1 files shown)


TheCoolPup:
nice to see more progress, whats strange on the previous pr i mentioned that the water on the dragon's stage during the 2nd intro is gone, but when going over to fbneo seems to be the exact opposite 
FinalBurn Neo v1 0 0 03 • Monster Maulers (ver EAA) 2_17_2025 4_58_29 PM

TheCoolPup:
figured i would post this here as well, noticed some issues with the water 
Mystic Warriors (ver JAA)  mystwarrj  - MAME 0 274 (LLP64) 2_17_2025 5_21_02 PM
Mystic Warriors (ver JAA)  mystwarrj  - MAME 0 274 (LLP64) 2_17_2025 5_21_37 PM

FredYeye:
Good to know! I made a thread for Konami Blending™ [here](https://github.com/orgs/mamedev/discussions/109) in case you're interested in checking it out, btw. It's mostly me stumbling around and master Galibert showing me how things work. I try to post various findings / comparisons there.

cuavas:
That actually looks like a straightforward bug with the person not knowing the source/destination parameter order for `alpha_blend_r32`.

TheCoolPup:
> That actually looks like a straightforward bug with the person not knowing the source/destination parameter order for `alpha_blend_r32`.

seems like it didnt do anything to fix the issue, wonder if its a underlying bug causing it

cuavas:
> > That actually looks like a straightforward bug with the person not knowing the source/destination parameter order for `alpha_blend_r32`.
> 
> seems like it didnt do anything to fix the issue, wonder if its a underlying bug causing it

That’s because you aren’t looking at the issue it’s supposed to fix.

PR #13384, opened 2025-02-19
Add a skeleton driver for Cisco CSS11501S-K9 Content Services Switch

clawgrip:
CSS11501S-K9 [ArcadeHacker]

### DIFF
--- a/src/mame/skeleton/css11501sk9.cpp
+++ b/src/mame/skeleton/css11501sk9.cpp
@@ -1,0 +1,136 @@
+// license:BSD-3-Clause
+// copyright-holders:
+
+/**********************************************************************************************************************
+
+Skeleton driver for Cisco CSS11501S-K9 Content Services Switch with SSL.
+
+The appliance has eight 10/100 Ethernet ports, one 10Base-T port for management, one RJ45 serial port for console, two
+PCMCIA slots for storage and one Gigabit Ethernet port (GBIC).
+
+Hardware components:
+
+Main board.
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/css11501sk9.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+void css11501sk9_state::machine_start()
+{
+}
+
+void css11501sk9_state::css11501sk9(machine_config &config)
+{
+	// basic machine hardware
+	RM7000BE(config, m_maincpu, 400'000'000); // Main CPU
+	RM7000BE(config, m_subcpu, 400'000'000); // SSL CPU
```
angelosa, line 102: No idea if it's different, may as well add to the note that both uses `RM7000A` really.

src/mame/skeleton/css11501sk9.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+namespace {
+
+class css11501sk9_state : public driver_device
+{
+public:
+	css11501sk9_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_maincpu(*this, "maincpu")
+		, m_subcpu(*this, "subcpu")
```
angelosa, line 71: Given this is coming from a SSL board this may as well be named `sslcpu` instead.

angelosa (approved):
Assuming I don't see a problem about `css11501sk9` being 11 characters long, and making it shorter doesn't seem a smart idea.

PR #13385, opened 2025-02-19
sinclair/spec128.cpp: added DMA mod

holub:
alternative to #12658 without UA858D


src/mame/sinclair/spec128.cpp
```diff
@@ -288,31 +288,106 @@ void spectrum_128_state::spectrum_128_fetch(address_map &map)
...
 	m_port_7ffd_data = 0;
 	spectrum_128_update_memory();
+
+	const u8 mod_dma = m_mod_dma.read_safe(0);
+	if (mod_dma & 1)
+	{
+		// mod_dma & 2 ? ZILOG : UA858D
+		const u8 port = mod_dma & 4 ? 0x6b : 0x0b;
+		m_maincpu->space(AS_IO).install_readwrite_handler(port, port, 0, 0xff00, 0, read8smo_delegate(*m_dma, FUNC(z80dma_device::read)), write8smo_delegate(*m_dma, FUNC(z80dma_device::write)));
+	}
```
cuavas, line 390: What’s going to happen if the user changes the configuration setting and then soft resets the emulated system?  Will this end up installed in two places?  What if they disable it and reset?  Will it still be installed from the previous start?

If there’s no clean way to uninstall this if the configuration changes before a soft reset, you need to only install it on the first reset.  A number of ISA cards, etc. do this check.

src/mame/sinclair/spec128.cpp
```diff
@@ -358,26 +433,35 @@ void spectrum_128_state::spectrum_128(machine_config &config)
...
 	config.set_maximum_quantum(attotime::from_hz(60));
 
-	/* video hardware */
+	Z80DMA(config, m_dma, X1_128_SINCLAIR / 10);
+	m_dma->out_busreq_callback().set_inputline(m_maincpu, Z80_INPUT_LINE_BUSRQ);
+	m_dma->out_int_callback().set_inputline(m_maincpu, INPUT_LINE_IRQ0);
+	m_dma->in_mreq_callback().set([this](offs_t offset) { return m_maincpu->space(AS_PROGRAM).read_byte(offset); });
+	m_dma->out_mreq_callback().set([this](offs_t offset, u8 data) { m_maincpu->space(AS_PROGRAM).write_byte(offset, data); });
+	m_dma->in_iorq_callback().set([this](offs_t offset) { return m_maincpu->space(AS_IO).read_byte(offset); });
+	m_dma->out_iorq_callback().set([this](offs_t offset, u8 data) { m_maincpu->space(AS_IO).write_byte(offset, data); });
```
cuavas, line 446: Should this really be here and not connected by default, or should the modified system with the DMA controller be a clone?

Also, you need an input merger when multiple things are trying to drive the same input (`INPUT_LINE_IRQ0` in this case).

holub: I don't think clone approach is feasible. In case we have multiple custom MODs: TurboFM, DMA, and potentially something else - clone approach is inachievable.
I better prefer to add these MODs as exp devices which they are not - problem with exp slot is... not all the clone have it.

cuavas: Would fake slots to represent modifications be feasible?  Just throwing ideas around.

holub: @cuavas I agree with all your concerns. Unfortunately I don't see straightforward way to do it now.
1. Input Merger required to revisit all spectrum's children which use inheritance.
2. Moving installation to startup: I don't know the way to access config params in these phases.

Putting this on my TODO list:
1. Apply input_merger for INPUT_LINE_IRQ0 in spectrum hierarchy
2. Investigate possibility of lightway slots for custom mods

PR #13386, opened 2025-02-19
[kn5000] Initial portions of dumps of the subcpu boot rom

felipesanches:


felipesanches:
Even though this is not yet the full dump, it is enough to perform the initial subcpu boot sequence and it removes a hack that was placing a few hardcoded instructions with an infinite loop to keep the subcpu quiet (not touching arbitrary addresses while looping through the entire address map).

galibert:
That's a little meh.  You don't really want pieces of the rom to end up in people's collections.  Can you rebuild the rom with ff fills and just tag it BAD_DUMP please?  And if your hypothesis that the fills are correct end up verified, you'll just remove the tag.

src/mame/matsushita/kn5000.cpp
```diff
@@ -836,15 +833,8 @@ ROM_START(kn5000)
...
-	ROM_REGION16_LE(0x20000, "mask", 0) // subcpu boot rom
-	ROM_LOAD("kn5000_mask_rom.ic30", 0x00000, 0x20000, NO_DUMP)
-	// hack to keep the CPU from touching SFRs arbitrarily while we do not have a proper ROM dump:
-	ROM_FILL(0x000000, 1, 0x68) // 68 fe = infinite loop
-	ROM_FILL(0x000001, 1, 0xfe)
-	ROM_FILL(0x01ff00, 1, 0x00) // RESET vector = 0x00fe0000
-	ROM_FILL(0x01ff01, 1, 0x00)
-	ROM_FILL(0x01ff02, 1, 0xfe)
-	ROM_FILL(0x01ff03, 1, 0x00)
+	ROM_REGION16_LE(0x20000, "mask", 0)
```
galibert, line 836: Also the tradition is to name a rom region with what it's used for.  So here "subcpu" would make more sense than the horribly generic "mask".

felipesanches:
Anything else needed here? This one is pretty straight-forward...

PR #13387, opened 2025-02-19
taito/qix.cpp: Cleanups/Updates

cam900:
- Use palette_device for palette feature
- Separate driver state class for kram3 due to encryption
- Use logmacro.h for MCU logging
- Reduce preprocessor defines
- Reduce literal tag usage
- Reduce duplicates
- Fix spacings
- Fix namings
- Add notes

### DIFF
--- a/src/mame/taito/qix.cpp
+++ b/src/mame/taito/qix.cpp
@@ -240,52 +240,84 @@ Interrupts:
 
 /*************************************
  *
+ *  Slither SN76489 I/O
+ *
+ *************************************/
+
+template <unsigned Which>
+void slither_state::sn76489_ctrl_w(int state)
+{
+	// write to the sound chip
+	if (!state && m_sn76489_ctrl[Which])
+		m_sn[Which]->write(m_pia[Which + 1]->b_output());
+
+	m_sn76489_ctrl[Which] = bool(state);
+}
+
+
+
+/*************************************
+ *
+ *  Slither trackball I/O
+ *
+ *************************************/
+
+template <unsigned Which>
+uint8_t slither_state::trak_r()
+{
+	return m_trak[Which + (m_flip ? 2 : 0)]->read();
+}
+
+
+
+/*************************************
+ *
  *  Data CPU memory handlers
  *
  *************************************/
 
 void qix_state::main_map(address_map &map)
 {
-	map(0x8000, 0x83ff).ram().share("share1");
+	map(0x8000, 0x83ff).ram().share("sharedram");
 	map(0x8400, 0x87ff).ram();
 	map(0x8800, 0x8bff).nopr();   /* 6850 ACIA */
-	map(0x8c00, 0x8c00).mirror(0x3fe).rw(FUNC(qix_state::qix_video_firq_r), FUNC(qix_state::qix_video_firq_w));
-	map(0x8c01, 0x8c01).mirror(0x3fe).rw(FUNC(qix_state::qix_data_firq_ack_r), FUNC(qix_state::qix_data_firq_ack_w));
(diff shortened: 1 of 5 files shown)


src/mame/taito/qix.h
```diff
@@ -39,140 +34,148 @@ class qix_state : public driver_device
...
+	void video(machine_config &config);
+
+	void init_kram3();
+
+private:
+	void lic_maincpu_changed(int state);
+	void lic_videocpu_changed(int state);
+
+	void main_map(address_map &map) ATTR_COLD;
+	void video_map(address_map &map) ATTR_COLD;
```
angelosa, line 134: Isn't this overshadowing what's in base class, requiring a `virtual override`?

src/mame/taito/qix.cpp
```diff
@@ -1370,7 +1403,7 @@ int qix_state::kram3_permut2(int tbl_index, int idx, const uint8_t *xor_table)
 	return xorval;
 }
 
-int qix_state::kram3_decrypt(int address, int value)
+static inline int decrypt(int address, int value)
```
cuavas, line 1406: This is only relevant to the encrypted Kram game – it’s better the make the name reflect that.

src/mame/taito/qix.cpp
```diff
@@ -1337,7 +1370,7 @@ static const uint8_t xor2_table[] =
 	99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,99,11, 6,99,
 };
 
-int qix_state::kram3_permut1(int idx, int value)
+static inline int permut1(int idx, int value)
```
cuavas, line 1373: Same for these permutation functions – they’re only relevant for one system, and it’s better to reflect that in the name.

src/mame/taito/qix.h
```diff
@@ -39,140 +34,149 @@ class qix_state : public driver_device
...
-		m_screen(*this, "screen")
+		m_videobank(*this, "videobank")
 	{ }
 
 	void qix_base(machine_config &config);
 	void qix(machine_config &config);
 	void qix_video(machine_config &config);
 	void qix_audio(machine_config &config);
-	void kram3(machine_config &config);
-	void kram3_video(machine_config &config);
```
cuavas, line 64: The ones that aren’t complete systems (`qix_video` and `qix_audio`) should be in the `protected` section.  Also, you can put `ATTR_COLD` on machine configuration functions.

src/mame/taito/qix.h
```diff
@@ -39,140 +34,149 @@ class qix_state : public driver_device
...
+	{ }
+
+	void kram3(machine_config &config);
+	void video(machine_config &config);
+
+	void init_kram3();
+
+protected:
+	virtual void main_map(address_map &map) override ATTR_COLD;
+	virtual void video_map(address_map &map) override ATTR_COLD;
```
cuavas, line 131: Since you’ve got a different machine configuration function anyway, is there really any point making these virtual and adding vtable slots?  Why not just give them non-ambiguous names and set the address maps in the machine configuration functions for this class?

src/mame/taito/qix.h
```diff
@@ -39,140 +34,149 @@ class qix_state : public driver_device
...
 	void slither(machine_config &config);
-	void slither_video(machine_config &config);
-	void slither_audio(machine_config &config);
+	void video(machine_config &config);
+	void audio(machine_config &config);
 
 protected:
 	virtual void machine_start() override ATTR_COLD;
 
+	virtual void video_map(address_map &map) override ATTR_COLD;
```
cuavas, line 161: Same here – you’ve got a separate machine configuration function, so you may as well give this a distinct name and set it in machine configuration.

PR #13388, opened 2025-02-19
mfabfz.cpp: Preliminary work to get the CP/M extension included

rfka01:
CP/M ROMs added - CP/M bootdisk softlist - information to add the floppy interface and memory banking

### DIFF
--- a/hash/mfacpm.xml
+++ b/hash/mfacpm.xml
@@ -3,10 +3,13 @@
 
 <!-- license:CC0-1.0 -->
 
-<softwarelist name="mfacpm" description="MFA Mikrocomputer fuer Ausbildung CP/M disks">
+<softwarelist name="mfacpm" description="MFA Mikrocomputer für Ausbildung CP/M disks">
+
+<!-- marked bad dump because it's the only known copy, but it contains utilities added by the copying service.
+It's unclear how it was originally distributed -->
 
 	<software name="mfacpm">
-		<description>MFA Mikrocomputer fuer Ausbildung original CP/M</description>
+		<description>MFA CP/M V 2.2 48tpi, 40 track DS System Disk</description>
 		<year>1987</year>
 		<publisher>BFZ Essen</publisher>
 		<!--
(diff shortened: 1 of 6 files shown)


hash/mfacpm.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+
+<!-- license:CC0-1.0 -->
+
+<softwarelist name="mfacpm" description="MFA Mikrocomputer fuer Ausbildung CP/M disks">
+
+	<software name="mfacpm">
+		<description>MFA Mikrocomputer fuer Ausbildung original CP/M</description>
```
cuavas, line 9: Did they actually use the spelling “fuer”, or did they use “für”?

src/mame/skeleton/mfabfz.cpp
```diff
@@ -87,6 +141,29 @@ class mfabfz_state : public driver_device
...
+private:
+	void mfacpm_io(address_map &map) ATTR_COLD;
+	void mfacpm_mem(address_map &map) ATTR_COLD;
+	void machine_reset() override ATTR_COLD;
+	void machine_start() override ATTR_COLD;
+	required_device<cpu_device> m_maincpu;
+	required_shared_ptr<uint8_t> m_shared_ram;
+	memory_view m_bootview;
+	required_device<i8251_device> m_uart;
+};
```
cuavas, line 166: Is it not possible to share a base class for these, or are they completely unrelated systems?  If they’re unrelated, do they belong in the same file as each other?

RetroAND: The system is modular, and I think the OP wants to be able to deliver a machine with the native, low level monitor and another with the optional but higher level CP/M extension.

hash/mfacpm.xml
rfka01: Good thing you opened that can of worms, I wasn't aware of the multitude of names they used for the system over time.
MFA is the abbreviation for "Mikrocomputer für Ausbildung", it's listed as such on Wiki, and this is how it's usually referred to. The system is ASCII only, so no Umlauts are on the the original keyboard. This is why I automatically replaced the "ü" with "ue".
Earlier systems are labelled "MFA Mikrocomputer", later ones "MFA Mediensystem" on the power supply.
The MAT ROMs that are already in MAME sign on as "BFZ/MFA Mikrocomputer".
Earlier documentation is for "MFA Mikrocomputer", later one is labelled as "MFA Mediensystem Mikrocomputertechnik".
I'm using the Umlaut now, but I'm keeping the "MFA Mikrocomputer für Ausbildung" as this is what it's known best for.
The "original" CP/M is on a copied disk with a hand printed label, as it came from a sort of distribution service (think shareware houses, but not shareware), with added utilities, so I've marked it as a bad dump.
I'll upload a revision soon.

src/mame/skeleton/mfabfz.cpp
rfka01: RetroAND's interpretation is correct. The system is on an ECB bus like backplane, and, you should ideally be able to pick and match in MAME. Until then it would be great to have CP/M going. In real hardware you modify some cards by setting system parameters via solder bridges, so I think it belongs to the same driver.

PR #13390, opened 2025-02-20
Add a skeleton for Microtek Powerpack In-Circuit Emulator PP-SW+ PIII

clawgrip:
Powerpack In-Circuit Emulator PP-SW+ PIII [ArcadeHacker]

### DIFF
--- a/src/mame/skeleton/powerpack.cpp
+++ b/src/mame/skeleton/powerpack.cpp
@@ -1,0 +1,80 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/****************************************************************************
+
+Skeleton driver for Microtek Powerpack In-Circuit Emulator model PP-SW+ PIII
+
+Main POD with two PCBs connected with a small backplane (also connects 
+with the pass-through CPU adapter using a flat cable):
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/powerpack.cpp
```diff
@@ -0,0 +1,80 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/****************************************************************************
+
+Skeleton driver for Microtek Powerpack In-Circuit Emulator model PP-SW+ PIII
+
+Main POD with two PCBs connected with a small backplane (also connects 
+with the pass-through CPU adapter using a flat cable):
```
angelosa, line 8: Details on the CPU adapter cable? How it connects to host?

clawgrip: cpupod-2
[Installation Guide FC_370.pdf](https://github.com/user-attachments/files/19085134/Installation.Guide.FC_370.pdf)

clawgrip: cpupod-4

PR #13391, opened 2025-02-20
add rx80 skeleton

goldnchild:
Here's a skeleton for the epson rx80.


cuavas (changes requested):
* Please turn this into a “private implementation” device in an anonymous namespace, like e.g. all the newer Apple II cards.
* Please just use `.set_constant(...)` for the inputs that need to see something other than the default value and leave the outputs unconnected for now, since this is really just a skeleton to allow the ROM to be debugged.

goldnchild:
ok, I made an anonymous namespace and converted it to DEFINE_DEVICE_TYPE_PRIVATE.  I  followed the model of the a2 4play device so I hope that's correct.

goldnchild:
Thanks for the edits and the merge!

PR #13395, opened 2025-02-22
apple2_flop_clcracked: Removed 2 dumps and updated metadata for Locksmith

A-Noid33:
Updated metadata for Locksmith

Removed (apple2_flop_clcracked.xml)
Locksmith (version 5.0 Revision C) (imperfect clean crack)
Locksmith (version 5.0 Revision F) (imperfect clean crack)

### DIFF
--- a/hash/apple2_flop_clcracked.xml
+++ b/hash/apple2_flop_clcracked.xml
@@ -9631,76 +9631,6 @@ license:CC0-1.0
 		</part>
 	</software>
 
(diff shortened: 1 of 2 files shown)


cuavas:
I’ll merge this as-is, but I think I’m going to make the major versions of Locksmith separate parent/clone groups, as it had some pretty major overhauls, e.g. the the addition of the disk editing features, and the new scripting system in 6.0.

PR #13397, opened 2025-02-22
taito/minivadr.cpp: Updates/Cleanups:

cam900:
- Restrict screen update function into cliprect
- Reduce literal tag usage
- Fix namings

### DIFF
--- a/src/mame/taito/minivadr.cpp
+++ b/src/mame/taito/minivadr.cpp
@@ -45,18 +45,20 @@ class minivadr_state : public driver_device
 {
 public:
 	minivadr_state(const machine_config &mconfig, device_type type, const char *tag)
-		: driver_device(mconfig, type, tag),
-		m_videoram(*this, "videoram"),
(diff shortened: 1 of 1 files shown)


src/mame/taito/minivadr.cpp
```diff
@@ -65,32 +67,26 @@ class minivadr_state : public driver_device
...
+		uint8_t const *const src = &m_videoram[y << 5];
+		uint32_t *const dst = &bitmap.pix(y);
+		for (int x = cliprect.left(); x <= cliprect.right(); x++)
 		{
-			pen_t pen = (data & 0x80) ? rgb_t::white() : rgb_t::black();
-			bitmap.pix(y, x) = pen;
-
-			data <<= 1;
-			x++;
+			dst[x] = BIT(src[x >> 3], 7 - (x & 7)) ? rgb_t::white() : rgb_t::black();
```
cuavas, line 78: `7 - (x & 7)` is equivalent to `~x & 7` – think about how binary maths works.

PR #13399, opened 2025-02-22
tek410x_kbd.cpp support for tek440x

Elektraglide:
Added Tek440x scan codes which differ from Tek410x slightly
Fixed some keycode mappings to allow working on Mac and PC.

@ajrhacker  would be interested in your thoughts


ajrhacker:
I disagree with the backslash key rebinding, and it isn't even done correctly here. Moving | to BACKSLASH2 seems OK in principle, and so I've done that.

PR #13404, opened 2025-02-23
linn/linndrum.cpp: Added interactive layout for the LinnDrum.

m1macrophage:
The slider script is copy-pasted without any changes from oberheim_dmx.lay.
Knobs can't be manipulated directly yet. Only through the Slider Control menu. Everything else is interactive.


m1macrophage:
Oops, fixed.

PR #13408, opened 2025-02-23
taito/flstory.cpp, taito/nycaptor.cpp: Correct sound CPU interrupt frequency

jotego:
Here's the circuit that generates the sound interrupt. The actual frequency is 122.07Hz, not 120Hz. This is from a PCB of _The FairyLand Story_ but I think the assumption of nycaptor being the same makes more sense than keeping 120Hz in that driver.

image


happppp:
Instead of force pushing/overwriting, please check when someone else does a commit after you submitted a PR.
The commit I did is gone, among some minor cleanups, it added: const attotime audio_irq_period = attotime::from_ticks(0x10000, 8_MHz_XTAL);

jotego:
> Instead of force pushing/overwriting, please rebase when someone else does a commit after you submitted a PR. The commit I did is gone (erased from history even, due to force push), among some minor cleanups, it added: const attotime audio_irq_period = attotime::from_ticks(0x10000, 8_MHz_XTAL);

I apologize. I was confused when git rejected the push from the terminal and I instictively just forced it. I did not realise of what was going on.

jotego:
Could you please do it again, @happppp ?

happppp:
> Could you please do it again, @happppp ?

I don't have the commit here, it was done via github. You could add it if you want, see the line I mentioned;
const attotime audio_irq_period = attotime::from_ticks(0x10000, 8_MHz_XTAL);
for nycaptor, since the XTAL apparently isn't certain(?) I'd suggest 8'000'000 instead of 8_MHz_XTAL.

The random cleanups I did don't matter, it was just the usual stuff I do when I browse a driver, like changing XTAL(x) to x_XTAL and source code spacing, or eg. removing the irrelevant comment that NMIs are from maincpu.

jotego:
I have not been able to recover the lost commit, but I have redone the changes you mentioned.

NYCAPTOR driver looks very similar to [FLSTORY schematics](https://github.com/jotego/jtbin/blob/master/sch/fairyland.pdf). Particularly, there is a whole subsystem for a second CPU that is not populated on the FLSTORY board but seems to match quite well the driver information for NYCAPTOR. So I think it should be used to fill the driver questions. I have replaced the guessed CPU frequencies for what the FLSTORY PCB did.

happppp:
nycaptor hw main Z80 is more likely 8MHz/2, see PCB notes, although both of them are bootlegs.

cuavas:
Do we have a consensus on what to do about the clock frequencies?  What’s your gut feel, @happppp?

PR #13410, opened 2025-02-23
add driver for sega sp400, put in anonymous namespace and DEFINE_DEVI…

goldnchild:
…CE_TYPE_PRIVATE

I removed all of the code that has to do with the actual printing, so this is kind of a skeleton.
That can easily be added later.

This isn't actually my code, it's fabiodl's and I'm just rewriting it to have an anonymous namespace and use DEFINE_DEVICE_TYPE_PRIVATE.

There is an existing pull request from a few years ago that has stalled.  Hopefully this "skeleton" may be acceptable.

Even more can be removed to make it a true skeleton.

### DIFF
--- a/src/devices/bus/sg1000_exp/sp400.cpp
+++ b/src/devices/bus/sg1000_exp/sp400.cpp
@@ -1,0 +1,241 @@
+// license: BSD-3-Clause
+// copyright-holders: Charles MacDonald, Devin Hill, Fabio Dalla Libera
+
+/*
+    license:BSD-3-Clause
+    Charles MacDonald
+    Devin Hill
+    Fabio Dalla Libera
+
+    Sega SP-400 Plotter
+
+    6805 dumped using https://github.com/charlesmacd/HD6805_Reader
+
+    Plotter uses ALPS Micrographic Printer DPG 1302 mechanism.
+
+    IC list:
+
+    * M5L8035LP (Mitsubishi clone of I8035)
+
+    P1 is the deserialized output data
+    P2.4 is the input serial data
+    P2.5 is busy (output to the computer)
+    P2.6 is the output strobe
+    P2.7 is busy (input from the 6805)
+
+    * HD6805V (Hitachi chip, pin compatible with MC6805P2, with 4K of ROM)
+
+    port A connected to the panel buttons
+        "By analyzing the disassembly I found out that three pins (PA5,6,7 of the 6805) are used to set
+        different configurations. The resolution can be changed, probably to support different mechanics.
+        The baud rate is 4800."
+    port B connected to the motors
+    port C connected to the pen and misc signals
+    port D connected to M5L8035LP parallel data
+*/
+
+// this is currently incomplete: missing paper output and pen sensor, everything else is working
+
+#include "emu.h"
+#include "sp400.h"
+
+#include "cpu/mcs48/mcs48.h"
+#include "cpu/m6805/m68705.h"
+
+namespace {
+
+class sp400_printer_device : public device_t,
+	public device_sk1100_printer_port_interface
+{
+public:
+	sp400_printer_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
+
+	static constexpr feature_type unemulated_features() { return feature::PRINTER; }
+
+	virtual void device_add_mconfig(machine_config &config) override ATTR_COLD;
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_reset_after_children() override ATTR_COLD;
+
+	const tiny_rom_entry * device_rom_region() const override ATTR_COLD;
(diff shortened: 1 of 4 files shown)


src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
...
+};
+
+sp400_printer_device::sp400_printer_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock):
+	device_t(mconfig, SP400_PRINTER, tag, owner, clock),
+	device_sk1100_printer_port_interface(mconfig, *this),
+	m_dsercpu(*this, "dsercpu"),
+	m_motcpu(*this, "motcpu"),
+	m_data(0),
+	m_busy(1),
+	m_dserdataout(0xFF),
```
cuavas, line 98: Please use lowercase hex digits.

src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
...
+	void update_pen_state(u8 data);
+	u8 pen_sensor() { return 0; }  // not currently hooked up
+
+	required_device<i8035_device> m_dsercpu;     // "deserializer cpu"
+	required_device<m6805_hmos_device> m_motcpu; // "motor cpu"
+
+	int m_data, m_busy;
+	int m_dserdataout, m_dserstrobe;
+	int m_motbusy, m_motPenUp, m_motPenDown;
+	int m_pendown;
```
cuavas, line 86: Please use `snake_case` consistently rather than this mix of `snakecase` and `llamaCase`.

Also, please use explicitly sized integers for things that need to be registered for save states.

src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+	return ROM_NAME( sp400 );
+}
+
+static INPUT_PORTS_START( sp400 )
+	PORT_START("BUTTONS")
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Color Select") PORT_CODE(KEYCODE_2_PAD)
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Line Feed")    PORT_CODE(KEYCODE_1_PAD)
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_KEYBOARD) PORT_NAME("Pen Change")   PORT_CODE(KEYCODE_3_PAD)
+INPUT_PORTS_END
```
cuavas, line 239: You don’t actually need `static` to indicate that you don’t want something visible from outside the translation unit from inside an anonymous namespace.

Also, these probably shouldn’t be `IPT_KEYBOARD`.  Maybe `IPT_KEYPAD` or `IPT_OTHER`?

src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+ioport_constructor sp400_printer_device::device_input_ports() const
+{
+	return INPUT_PORTS_NAME(sp400);
+}
+
+} // anonymous namespace
+
+DEFINE_DEVICE_TYPE_PRIVATE(SP400_PRINTER, device_sk1100_printer_port_interface, sp400_printer_device, "sega_sp400", "SP-400 Plotter")
```
cuavas, line 248: Can probably do to make the description “Sega SP-400 Plotter”.

src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
...
+	save_item(NAME(m_pendown));
+}
+
+void sp400_printer_device::device_reset()
+{
+}
+
+void sp400_printer_device::device_reset_after_children()
+{
+}
```
cuavas, line 126: You can leave out these – it’s not like reset propagates over a serial port anyway.

cuavas (changes requested):

src/devices/bus/sg1000_exp/sp400.cpp
```diff
@@ -0,0 +1,249 @@
+// license: BSD-3-Clause
+// copyright-holders: Charles MacDonald, Devin Hill, Fabio Dalla Libera
```
cuavas, line 2: How much actual code that qualifies as a creative work by any of these people is left?  There’s very little left at all that isn’t “facts” (address maps, ROM definitions) where only the formatting is eligible for copyright protection.  I’d also hazard a guess that Charles MacDonald didn’t contribute much to the actual code in the first place.

goldnchild: ok, I have reduced it to just fabiodl.  I have sent him an email asking if that is ok with him.

PR #13412, opened 2025-02-24
wing/pinkiri8.cpp: Updates/Cleanups:

cam900:
- Move video function into janshi_vdp_device
- Use device_gfx_interface for janshi_vdp_device gfx decoding feature
- Demote pinkiri8 to MACHINE_NO_SOUND because sound ROM not dumped (or not exists in real hardware?)
- Use set_format for palette format, Fix palette entry matches to palette RAM size(*2 for 16(15 actually) bit color)
- Add items into save state
- Fix input matrix reading behavior
- Reduce duplicates
- Reduce unnecessary lines
- Convert background tilemap to tilemap_t
- Fix namings
- Fix spacings

### DIFF
--- a/src/mame/wing/pinkiri8.cpp
+++ b/src/mame/wing/pinkiri8.cpp
@@ -39,11 +39,15 @@ Dumped by Chackn
 ***************************************************************************/
 
 #include "emu.h"
+
 #include "cpu/z180/hd647180x.h"
+
 #include "sound/okim6295.h"
+
 #include "emupal.h"
 #include "screen.h"
 #include "speaker.h"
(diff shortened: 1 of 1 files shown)


angelosa:
> Demote pinkiri8 to MACHINE_NO_SOUND because sound ROM not dumped (or not exists in real hardware?)

https://github.com/mamedev/mame/blob/f4e792e8179154861056883463586ab8cacf9cd0/src/mame/wing/pinkiri8.cpp#L474-L478

I presume the OKI socket is just unpopulated but needs the PCB pictures and I don't have those handy.

PR #13414, opened 2025-02-25
plugins/layout: Proposal for sharing slider functionality.

m1macrophage:
This PR implements one possible way to share slider (and other widget) functionality between layouts.  It also adapts the DMX layout to use the shared library, as a proof of concept.

The `sliders.lua` implementation is basically the implementation in the DMX and LinnDrum layouts, except that scanning the layout for sliders was replaced with `add_slider()` calls from the layout script. I expect that `sliders.lua` will be extended to support horizontal sliders, knobs, etc.

Open to feedback and suggestions from the experts.

### DIFF
--- a/plugins/layout/init.lua
+++ b/plugins/layout/init.lua
@@ -48,7 +48,8 @@ function layout.startplugin()
 			string = string,
 			tonumber = tonumber,
 			tostring = tostring,
-			table = table }
+			table = table,
+			slider_manager = require("layout/sliders") }
 		local script, err = load(script, script, "t", env)
 		if not script then
 			emu.print_warning("error loading layout script " .. err)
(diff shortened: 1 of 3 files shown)


cuavas:
Sorry, I don't want to do it this way. The current approach for "library" plug-ins is flawed. I have an idea for how to improve that, and at the same time provide better support for plug-ins that need to communicate with each other. When that's done, the stuff for doing this (as well as other stuff for other common layout tasks) can go in a library. For now, can you stick with copy/paste in layouts?

As usual, we're being held back by short-sighted implementation decisions from the past.

m1macrophage:
Closing  this PR. If there are ideas on how to achieve reuse in a forward-compatible way, happy to consider them.

PR #13416, opened 2025-02-25

tabudz:
**Description**
This PR fixes a security vulnerability in addBinding() that was cloned from libexpat but did not receive the security patch. The original issue was reported and fixed under https://github.com/libexpat/libexpat/commit/a2fe525e660badd64b6c557c2b1ec26ddc07f6e4.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2022-25236
https://github.com/libexpat/libexpat/commit/a2fe525e660badd64b6c557c2b1ec26ddc07f6e4


cuavas:
I’ve updated expat to 2.6.4 which includes this as well as various othr changes.

PR #13421, opened 2025-02-26
update to add network for Sega X-Board, Y-Board, Konami K056230, Namco C139

SailorSat:
This includes updated socket handling - accepting an incoming socket could be mistaken for end of stream; this has been changed so that accepting returns a would_block error.

Sega X-Board comms make 'Super Monaco GP', and 'Last Survivor' playable in LAN.

Sega Y-Board comms make 'Power Drift - Link Version' playable in LAN (setup dips correctly!)

Konami K056230 makes 'GTI Club', 'Jet Wave', 'Midnight Run', 'Polygonet Commanders', 'Poly-Net Warriors' and 'Winding Heat' playable in LAN.

Namco C139 makes 'Driver's Eyes' working
other C139 games are working but the network is highly timing critical (unstable).

--- image spam below ---

Konami K056230
gticlub_linkplay4
gticlub_linktest4
jetwave_linkplay4
jetwave_linktest4
midnrun_linkplay4
midnrun_linktest4
plygonet_linkplay4
plygonet_linktest4
polynetw_linkplay4
polynetw_linktest4
windheat_linkplay4
windheat_linktest4

Namco C139
acedrive-4pl-ingame
acedrive-4pl-teaser
adillor-4pl-teaser
cybrcomm-2pl-ingame
cybrcycc-4pl-ingame
cybrcycc-4pl-teaser
cybsled-2pl-ingame
raverace-4pl-feature
ridgera2_4pl_ingame
ridgera2_4pl_linkfeature
ridgeracf_attract
suzuka8h-8pl-ingame
suzuka8h-8pl-teaser
driveyes-ingame


SailorSat:
pdriftl - https://www.youtube.com/watch?v=R8OlbqSLZhQ
smgp - https://www.youtube.com/watch?v=0uVJrce4-es
midnrun / windheat - https://www.youtube.com/watch?v=0LWOftPXWNE
gticlub - https://www.youtube.com/watch?v=Jmj_X1fdC2o
ridgera2 - https://www.youtube.com/watch?v=tid2uVlX68k
raverace - https://www.youtube.com/watch?v=nx0NL4GTbWk

src/devices/machine/mb89372.cpp
```diff
@@ -0,0 +1,566 @@
...
+#include "emu.h"
+#include "mb89372.h"
+#include "emuopts.h"
+
+//**************************************************************************
+//  LIVE DEVICE
+//**************************************************************************
+
+// device type definition
+DEFINE_DEVICE_TYPE(MB89372, mb89372_device, "mb89372", "MB89372 Multi-Protocol Controller")
```
angelosa, line 34: Long name needs `Fujitsu` prefixed.

SailorSat: Alright, updated MB89372 and MB89374 accordingly.

MooglyGuy:
All or nearly all of my feedback points have been addressed, it doesn't appear that I can mark them as resolved on my own, though.

SailorSat:
> All or nearly all of my feedback points have been addressed, it doesn't appear that I can mark them as resolved on my own, though.

To be honest I can't even see any of those on github.
From my point of view, your comment was the first interaction on that topic.
github sometimes seems weird.

Hydreigon223:
I'm okaying decision to remove "NODEVICE_LAN" flag from thunderh/thunderhu because of its LANC usage for something other than cabinet linking.

src/mame/sega/m1comm.cpp
```diff
@@ -581,25 +593,22 @@ int m1comm_device::read_frame(int dataSize)
 					togo -= recv;
 					offset += recv;
 				}
-				else if (!filerr && recv == 0)
+				if ((!filerr && recv == 0) || (filerr && std::errc::operation_would_block != filerr))
 				{
 					togo = 0;
 				}
```
cuavas, line 599: You know `recv` can’t be negative, and it doesn’t return an error on a partial read, so why not just make it:

```c++
				}
				else if (std::errc::operation_would_block != filerr)
				{
```

That will catch all the condition’s you’re treating as errors here.

src/mame/sega/m1comm.cpp
```diff
@@ -581,25 +593,22 @@ int m1comm_device::read_frame(int dataSize)
...
 			}
 		}
 	}
-	else if (!filerr && recv == 0)
+	if ((!filerr && recv == 0) || (filerr && std::errc::operation_would_block != filerr))
 	{
+		osd_printf_verbose("M1COMM: rx connection error\n");
```
cuavas, line 605: It would be helpful to log the actual error.

src/mame/sega/m1comm.cpp
```diff
@@ -619,22 +628,17 @@ void m1comm_device::send_frame(int dataSize){
...
 
-	std::error_condition filerr;
-	std::uint32_t written;
-
-	filerr = m_line_tx->write(&m_buffer0, 0, dataSize, written);
+	uint32_t written = 0;
+	std::error_condition filerr = m_line_tx->write(&m_buffer0, 0, dataSize, written);
 	if (filerr)
 	{
+		osd_printf_verbose("M1COMM: tx connection error\n");
```
cuavas, line 635: Since you’ve got a non-blocking socket, there’s a very real possibility that you’ll end up with a partial write here.  You should be checking `written`.  Blocking really isn’t an option, as you’ll hold up the emulation.  Have you thought about what to do in that situation?  (An asynchronous signal can also cause a partial write to happen.)

Also remember that you won’t get an error from a partial write, you need to attempt to complete the write to see the error.  That’s the reason for those read/write loop helpers in src/lib/util/ioprocs.cpp

src/mame/sega/m1comm.cpp
```diff
@@ -315,6 +316,9 @@ void m1comm_device::comm_tick()
 {
 	if (m_linkenable == 0x01)
 	{
+		std::error_condition filerr;
+		uint64_t filesize; // unused
```
cuavas, line 320: Is there a good reason for lifting the unused file size here?  It makes the compiler’s job harder, because it needs to ensure the address is fixed every time a reference to it is passed to a function, and that the value isn’t nuked between potential uses.  Keeping them in the smallest scope possible lets the compiler generate more efficient code.

cuavas (commented):
I’ve only looked at a few of the files, but I’ve already found a bunch of stuff.  In addition to the in-place comments, quite a few files seem to be missing the terminating newline – you might want to run MAME’s `srcclean` tool over the files you’ve changed to clean that up as well as other formatting issues.

I expect at least some of the things I’ve noticed in the files I looked at already will be applicable to multiple files, so if you could apply fixes to other applicable places preemptively it might expedite things a bit.

cuavas:
> All or nearly all of my feedback points have been addressed, it doesn't appear that I can mark them as resolved on my own, though.

Did you remember to hit the “complete review” button or whatever it’s called?

scripts/src/machine.lua
```diff
@@ -2201,6 +2201,18 @@ if (MACHINES["MB89371"]~=null) then
...
 
+---------------------------------------------------
+--
+--@src/devices/machine/mb89371.h,MACHINES["mb89372"] = true
+---------------------------------------------------
+
+if (MACHINES["mb89372"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/machine/mb89372.cpp",
+		MAME_DIR .. "src/devices/machine/mb89372.h",
```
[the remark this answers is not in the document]
cuavas: It tells the build script which flag to turn on if it discovers that header in use.  It’s *very* important for `SOURCES=` builds.

src/mame/sega/m1comm.cpp
SailorSat: Are you sure we are using non-blocking sockets here?

As far as posixsocket.cpp and winsocket.cpp go, sockets are opened with TCP_NODELAY (which disables nagle's algorythm)

posixsocket.cpp sets O_NONBLOCK for "domains" only

src/mame/sega/m1comm.cpp
SailorSat: We discussed that a while back in PR#12904.

'no error and recv == 0' IS an actual error that needs to be handled, or we'll cause an infinite loop.

modusc896d352: github doesn't recognize "PR#12904", so when referencing PRs from github just type #12904 instead.

src/mame/namco/namcos22.h
```diff
@@ -16,6 +16,7 @@
 #include "namcomcu.h"
 #include "machine/timer.h"
 #include "sound/c352.h"
+#include "namco_c139.h"
 #include "video/rgbutil.h"
 #include "video/poly.h"
```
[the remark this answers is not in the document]
SailorSat: gave it a spin, hope this is fine now.

src/mame/sega/ybdcomm.cpp
```diff
@@ -0,0 +1,704 @@
...
+
+		case 2:
+			// status register?
+			LOG("ybdcomm-ex_r: %02x %02x\n", bank, m_ybd_stat);
+			return m_ybd_stat;
+
+		default:
+			logerror("ybdcomm-ex_r: %02x\n", offset);
+			return 0xff;
+}	}
```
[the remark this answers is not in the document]
SailorSat: checking side_effects_disabled now.
did a srcclean run and fixed formating.

src/mame/sega/ybdcomm.cpp
```diff
@@ -0,0 +1,704 @@
...
+
+	// prepare remotehost "filename"
+	m_remotehost[0] = 0;
+	strcat(m_remotehost, "socket.");
+	strcat(m_remotehost, mconfig.options().comm_remotehost());
+	strcat(m_remotehost, ":");
+	strcat(m_remotehost, mconfig.options().comm_remoteport());
+
+	m_framesync = mconfig.options().comm_framesync() ? 0x01 : 0x00;
+#endif
```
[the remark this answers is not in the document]
SailorSat: using std::string and util::string_format now.
i still prefer having those as members rather than building them every time.

src/mame/sega/ybdcomm.cpp
```diff
@@ -0,0 +1,704 @@
...
+	m_cpu->set_memory_map(&ybdcomm_device::ybdcomm_mem);
+	m_cpu->set_io_map(&ybdcomm_device::ybdcomm_io);
+
+	MB8421(config, m_dpram).intl_callback().set(FUNC(ybdcomm_device::dpram_int5_w));
+
+	MB89372(config, m_mpc, 8000000); // 16 MHz / 2
+	m_mpc->out_hreq_callback().set(FUNC(ybdcomm_device::mpc_hreq_w));
+	m_mpc->out_irq_callback().set(FUNC(ybdcomm_device::mpc_int7_w));
+	m_mpc->in_memr_callback().set(FUNC(ybdcomm_device::mpc_mem_r));
+	m_mpc->out_memw_callback().set(FUNC(ybdcomm_device::mpc_mem_w));
```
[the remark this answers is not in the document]
SailorSat: alright, used the XTALs.

src/mame/sega/ybdcomm.cpp
```diff
@@ -0,0 +1,704 @@
...
+	PORT_DIPSETTING(    0x07, "7" )
+	PORT_DIPSETTING(    0x08, "8" )
+	// enabled for debugging
+	PORT_DIPSETTING(    0x09, "9 (invalid)" )
+	PORT_DIPSETTING(    0x0a, "10 (invalid)" )
+	PORT_DIPSETTING(    0x0b, "11 (invalid)" )
+	PORT_DIPSETTING(    0x0c, "12 (invalid)" )
+	PORT_DIPSETTING(    0x0d, "13 (invalid)" )
+	PORT_DIPSETTING(    0x0e, "14 (invalid)" )
+	PORT_DIPSETTING(    0x0f, "15 (invalid)" )
```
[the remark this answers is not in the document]
SailorSat: removed all "invalid" options, added dip locations.

src/mame/sega/m1comm.cpp
SailorSat: should be fixed now.

src/mame/sega/m1comm.cpp
SailorSat: I now log the errors.

src/mame/sega/ybdcomm.cpp
```diff
@@ -0,0 +1,704 @@
...
+					m_linkenable = 0x01;
+					m_linkid = 0x00;
+					m_linkalive = 0x00;
+					m_linkcount = 0x00;
+					m_linktimer = 0x003a;
+
+					// read dips and write to shared memory
+					uint8_t sw1 = ioport("Link_SW1")->read();
+					mpc_mem_w(0x4000, sw1 & 0x0f);
+					mpc_mem_w(0x4001, (sw1 >> 4) & 0x0f);
```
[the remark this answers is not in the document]
SailorSat: changed osd_printf_verbose to LOG where applicate.
connection related messages should be verbose imho.

src/mame/sega/ybdcomm.h
```diff
@@ -0,0 +1,96 @@
...
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+	virtual void device_reset_after_children() override;
+
+	virtual ioport_constructor device_input_ports() const override ATTR_COLD;
+	virtual const tiny_rom_entry *device_rom_region() const override ATTR_COLD;
+
+	// optional information overrides
+	virtual void device_add_mconfig(machine_config &config) override ATTR_COLD;
```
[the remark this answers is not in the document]
SailorSat: removed those comments.

src/mame/sega/ybdcomm.h
```diff
@@ -0,0 +1,96 @@
...
+#define MAME_SEGA_YBDCOMM_H
+
+#pragma once
+
+#define YBDCOMM_SIMULATION
+
+#include "osdfile.h"
+#include "cpu/z80/z80.h"
+#include "machine/mb8421.h"
+#include "machine/mb89372.h"
```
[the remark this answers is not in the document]
SailorSat: i tried to update all files accordingly

src/mame/sega/ybdcomm.h
```diff
@@ -0,0 +1,96 @@
...
+public:
+	// construction/destruction
+	ybdcomm_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+
+	// ex-bus connection to host
+	uint8_t ex_r(offs_t offset);
+	void ex_w(offs_t offset, uint8_t data);
+
+	void ybdcomm_mem(address_map &map) ATTR_COLD;
+	void ybdcomm_io(address_map &map) ATTR_COLD;
```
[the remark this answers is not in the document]
SailorSat: you are right. corrected.

src/mame/sega/ybdcomm.h
```diff
@@ -0,0 +1,96 @@
...
+
+	int comm_frameOffset(uint8_t cabIdx);
+	int comm_frameSize(uint8_t cabIdx);
+	void comm_tick();
+	int read_frame(int dataSize);
+	void send_data(uint8_t frameType, int frameOffset, int frameSize, int dataSize);
+	void send_frame(int dataSize);
+
+	emu_timer *m_tick_timer;
+#endif
```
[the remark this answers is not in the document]
SailorSat: i moved some things around. but not sure if correct now.

src/mame/sega/m1comm.cpp
cuavas: Oh, I’m getting confused again.  I’m thinking of the interfaces in ioprocs.h.  It’s actually worse than that with `osd_file`.  Look at the doxygen comment for `read` in src/osd/osdfile.h:
```c++
	/// \param [out] actual Receives the number of bytes read if the
	///   operation succeeds.  Not valid if the operation fails.
```

`actual` is not guaranteed to be valid if the result is set.  You need to check `filerr` *before* you check `recv`, because if `filerr` is set, the value of `recv` is is not guaranteed to be valid at all.

You need to do something like this:
```c++
				if (filerr)
				{
					recv = 0;
					if (std::errc::operation_would_block != filerr)
						togo = 0;
				}
				else if (recv > 0)
				{
					for (int i = 0 ; i < recv ; i++)
					{
						m_buffer0[offset + i] = m_buffer1[i];
					}
					togo -= recv;
					offset += recv;
				}
				else
				{
					togo = 0;
				}
```

cuavas:
Sorry, almost 3AM, too tired to look at the full diff again tonight.  I’ll get back to it some time tomorrow.

src/mame/sega/m1comm.cpp
SailorSat: should be fine now.

src/mame/sega/m1comm.cpp
SailorSat: should be fine now.

src/devices/machine/k056230.cpp
```diff
@@ -207,119 +200,113 @@ void k056230_device::comm_tick()
...
 		{
-			osd_printf_verbose("k056230: listen on %s\n", m_localhost);
-			filerr = osd_file::open(m_localhost, OPEN_FLAG_CREATE, m_line_rx, filesize);
+			osd_printf_verbose("k056230: rx listen on %s\n", m_localhost);
+			u64 filesize; // unused
+			std::error_condition filerr = osd_file::open(m_localhost, OPEN_FLAG_CREATE, m_line_rx, filesize);
 			if (filerr.value() != 0)
 			{
-				osd_printf_verbose("k056230: rx connection failed\n");
+				osd_printf_verbose("k056230: rx connection failed - %02x, %s\n", filerr.value(), filerr.message());
```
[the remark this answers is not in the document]
SailorSat: I added the category, although any errors are not very meaningful right now at all.

having closed one of two instances, the 'other' shows this:
`M1COMM: rx connection failed - generic:0 No error
M1COMM: link lost
M1COMM: tx connection failed - generic:108 Unknown error`

MooglyGuy:
> > All or nearly all of my feedback points have been addressed, it doesn't appear that I can mark them as resolved on my own, though.
> 
> Did you remember to hit the “complete review” button or whatever it’s called?

I don't even see a "Complete Review" button, which is bizarre as I certainly seemed to be able to open one.

src/mame/sega/ybdcomm.h
```diff
@@ -0,0 +1,96 @@
...
+};
+
+// device type definition
+DECLARE_DEVICE_TYPE(YBDCOMM, ybdcomm_device)
```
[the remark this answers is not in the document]
MooglyGuy: Other things being done badly, previously, is because they pre-date the project effectively having any standards. It's usually better to ask Vas what the preferred guidelines are than to infer it from existing examples.

I was working on doing bulk cleanups along those lines at one point, but I've hit a spot of depression for the past few months.

MooglyGuy (approved):

src/devices/machine/mb89372.cpp
```diff
@@ -0,0 +1,566 @@
...
+			{
+				m_current_channel = channel;
+				return;
+			}
+			int reg = 0x20 + (channel * 4);
+			if ((m_channel[channel].m_state == 0) && (m_reg[reg + 3] & 0x80))
+			{
+				m_channel[channel].m_state = 1;
+				m_current_channel = channel;
+				logerror("meep... %01x\n", channel);
```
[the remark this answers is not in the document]
SailorSat: to be honest,
I don't even remember why I added that output in the first place.

I wasn't able to source a comm board with that chip to do tests on the real chip.
I'll remove that output.

src/devices/machine/k056230.cpp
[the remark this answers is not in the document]
cuavas: But you know that the `generic` category values are equivalent to `errno` values, so you can look up 108 in the definitions for `errno` values on your platform.  On Linux, 108 would be `ESHUTDOWN` – Cannot send after transport endpoint shutdown.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +139,210 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+		{
+			osd_printf_verbose("k056230: tx connect to %s\n", m_remotehost);
+			u64 filesize; // unused
+			std::error_condition filerr = osd_file::open(m_remotehost, 0, m_line_tx, filesize);
+			if (filerr.value() != 0)
+			{
+				osd_printf_verbose("k056230: tx connection failed - %s:%d %s\n", filerr.category().name(), filerr.value(), filerr.message());
+				m_line_tx.reset();
+			}
+		}
```
cuavas, line 227: Wouldn’t it make sense to start listening on start rather than in an address space write handler?

I’d question whether it’s a good idea to try opening sockets in a write handler at all:
* Opening a socket calls `gethostbyname` which blocks on name service lookup.  This can be time-consuming, and you’re holding up the emulation with this.
* At a quick glance, it appears to use a synchronous `connect` call, which is also going to hold up emulation, particularly if there’s a timeout (this could cause the UI to become unresponsive for over a minute – the default timeout is 72 seconds on Windows).
* You aren’t keeping track of the wall clock time on a failed connection attempt, and you aren’t implementing any kind of back-off on failed attempts.

The net result is that this approach is liable to cause MAME to get into a state where it appears to be frozen, and a user is unlikely to be holding the key/button assigned to UI Cancel in the brief interval between the long blocking connection attempts, so it will look like you can’t exit.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +139,210 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+				return 0;
+		}
+		if ((!filerr && bytes_read == 0) || (filerr && std::errc::operation_would_block != filerr))
+		{
+			osd_printf_verbose("k056230: rx connection failed - %s:%d %s\n", filerr.category().name(), filerr.value(), filerr.message());
+			m_line_rx.reset();
+			return 0;
+		}
+		bytes_total += bytes_read;
+	}
```
cuavas, line 320: Won’t this spin and hold up emulation if it receives a partial frame?  Spinning and making MAME unresponsive waiting for network data that may not arrive is not going to provide a pleasant user experience.  If you get a “would block” error, you need to let the emulation continue and check for data again later.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +139,210 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+	{
+		std::error_condition filerr = m_line_tx->write(&m_buffer0[bytes_total], 0, data_size - bytes_total, bytes_sent);
+		if (filerr)
+		{
+			osd_printf_verbose("k056230: tx connection failed - %s:%d %s\n", filerr.category().name(), filerr.value(), filerr.message());
+			m_line_tx.reset();
+			return;
+		}
+		bytes_total += bytes_sent;
+	}
```
cuavas, line 342: As I said on a previous iteration, you really don’t want to be holding up emulation waiting to send data.  You want to defer sending the remaining data until a point where you won’t get blocked.

cuavas (commented):
I’m really not comfortable with stuff that can easily get into a state where it makes MAME unresponsive to user input.  No-one likes software that freezes and needs to be killed when something goes wrong.  In this case, something going wrong could be as simple as misconfiguration causing connections to time out, loss of network connectivity, or the name service becoming unresponsive.

Now taking a step back, you’re up against framework issues:
* Saying the `osd_file` interface was “designed” is giving it too much credit.  Why is getting the length (which may not be applicable) part of opening a file?  Why do reads and writes always have an implicit seek (which also may not be applicable)?  Sockets hit both of those.
* The “magic” file names that create sockets were really just intended to do things like connecting a serial or parallel port to a socket rather than a file for dumping output.  They connect or listen on open (which happens on start or when you change media), and they don’t give much control (e.g. can’t cancel blocking operations, read will block if data is available but smaller than buffer size passed in, etc.).

You can mitigate the issues somewhat by moving socket operations (or at least opening socket) to a worker thread.  But even if you do that, you have no way to cancel a connection attempt, which means the user could end up having MAME freeze for over a minute waiting for a connection to time out so you can clean up your worker thread if they happen to try to exit just after a connection attempt begins.

Have you considered using ASIO directly for your networking code like the CPS2 comm board (for SF2: The Tournament Battle)?  That gives you asynchronous operations (emulated using threads if necessary), and you get a lot more control as you don’t have sockets hiding behind an interface that’s barely adequate for random access files.

MooglyGuy:
I want to throw my two cents in on this whole "OSD vs. ASIO" thing: Given the fact that @SailorSat here isn't just the person developing these networking-board implementations, but is the person actively *using* them, she knows better than anyone whether @cuavas's assertions of doom and gloom about main-thread blocking are likely to happen given the workflow of a typical user of this feature.

This is a feature intended not for playing these games across the Internet, but across multiple instances of MAME on either the same machine or on a LAN, for local parties and such.

The concerns being expressed about "what happens if a connection drops out" are broadly irrelevant, because the feature already self-selects for users that know what they're doing. Average users are almost certainly going to *try* to use it across the Internet, and they'll end up disappointed no matter what Ariane here does - because a user trying to do that is one of those "Doctor, it hurts when I hit myself with this hammer" situations.

The more likely scenario is *temporary* stutter or blocking of the main thread, if it can be blocked at all. When MAME already becomes unusable on Windows if the user erroneously fullscreens MAME while running with `-debug` and hits a breakpoint - which results in a locked-up MAME that's almost impossible to even get Task Manager to kill due to the fullscreen output taking priority over the desktop window manager - there's no value or point to focusing on the molehill that already has a mountain next to it.

Stop holding up contributors' progress in other areas of MAME just because there are some sins in the core that could and should be rectified. With the nebulous-to-nonexistent timeline for addressing those issues, it comes off as hypocrisy given the history of giving grief to @smf- for purportedly doing the same.

rb6502:
@cuavas Do you have any objections to this besides theoretical ones about MAME hanging?  Because I'm gonna merge it if not, and then SailorSat will have a solid base to rework it for ASIO.

cuavas:
> @cuavas Do you have any objections to this besides theoretical ones about MAME hanging? Because I'm gonna merge it if not, and then SailorSat will have a solid base to rework it for ASIO.

It isn't theoretical. Having MAME hang because some other random program is listening on a particular TCP port, or having it lock up if it loses connectivity is not an acceptable use experience.

I'm going to revert it if it's merged in this state, because if it *is* merged like this, it's inevitably going to become one more thing I have to fix myself. (But in this case, it's lots more things to fix.)

If you're feeling under pressure because Haze announced that this will be in the next release, meeting it in a state like this isn't a solution.

rb6502:
People would be opting in to running the game linked, at which point there is a whole universe of things out of our control.  And I'm fine with that.  I'll add the "you are now on your own" popmessage myself if it helps.  SailorSat's previous networking stuff has many of the same theoretical issues and has been in mainline for years without complaints, so this just feels malicious.

MooglyGuy:
I'm with RB on this one. With all of your criticisms for smf, this just makes you look like a massive hypocrite for doing exactly the same thing, Vas.

MooglyGuy:
"It's inevitably going to become one more thing I have to fix", get down off your cross, dude. Ariane has already said she's willing to look into a rewrite with enough time, so no, it isn't.

You perpetually bemoan things not being done to your liking, yet you never actually document what "to your liking" entails.

You perpetually go on about how things need to be better, yet you never document what "better" actually is.

Let's not forget that ultimately the current coding guidelines that exist on https://docs.mamedev.org/ weren't written by you, they were written by me, out of frustration for your perpetual lamentations about how nobody could get the coding style right.

At this point it seems like you genuinely enjoy being able to portray yourself as this set-upon victim of circumstance. Quit your bullshit.

SailorSat:
So I tried some asio rewrite for m1comm ( see [sega.zip](https://github.com/user-attachments/files/19238644/sega.zip) )

However I run into some compiler error. Not sure if I did something wrong.

`GCC 11.2.0 detected
Compiling src/mame/sega/model1.cpp...
In file included from ../../../../../3rdparty/softfloat/milieu.h:36,
                 from ../../../../../src/devices/cpu/i386/i386.h:11,
                 from ../../../../../src/mame/sega/model1.cpp:598:
../../../../../3rdparty/softfloat/mamesf.h:5: error: "LITTLEENDIAN" redefined [-Werror]
    5 | #define LITTLEENDIAN
      |
In file included from ../../../../../3rdparty/asio/include/asio/detail/socket_types.hpp:33,
                 from ../../../../../3rdparty/asio/include/asio/detail/win_tss_ptr.hpp:23,
                 from ../../../../../3rdparty/asio/include/asio/detail/tss_ptr.hpp:25,
                 from ../../../../../3rdparty/asio/include/asio/detail/call_stack.hpp:20,
                 from ../../../../../3rdparty/asio/include/asio/detail/thread_context.hpp:20,
                 from ../../../../../3rdparty/asio/include/asio/detail/recycling_allocator.hpp:20,
                 from ../../../../../3rdparty/asio/include/asio/detail/handler_alloc_helpers.hpp:21,
                 from ../../../../../3rdparty/asio/include/asio/detail/executor_function.hpp:19,
                 from ../../../../../3rdparty/asio/include/asio/execution/any_executor.hpp:24,
                 from ../../../../../3rdparty/asio/include/asio/execution.hpp:19,
                 from ../../../../../3rdparty/asio/include/asio/any_completion_executor.hpp:22,
                 from ../../../../../3rdparty/asio/include/asio.hpp:18,
                 from ../../../../../src/osd/asio.h:28,
                 from ../../../../../src/mame/sega/m1comm.h:14,
                 from ../../../../../src/mame/sega/model1.h:9,
                 from ../../../../../src/mame/sega/model1.cpp:596:
D:/msys64/mingw64/x86_64-w64-mingw32/include/winsock2.h:511: note: this is the location of the previous definition
  511 | #define LITTLEENDIAN 0x0001
      |
cc1plus.exe: all warnings being treated as errors
`

SailorSat:
So... I did successfully rewrite _some_ of the comm implementations to ASIO.

However ASIO seems to clash with both the SHARC DRC as well as the M68000 code.
I won't even try mb89372 and mb89374 for now...

Model2 fails
[m2comm.txt](https://github.com/user-attachments/files/19268208/m2comm.txt)


Sega Y-Board, Namco C139 and Konami k056230 fail
[k056230.txt](https://github.com/user-attachments/files/19268209/k056230.txt)
[namco_c139.txt](https://github.com/user-attachments/files/19268210/namco_c139.txt)
[ybdcomm.txt](https://github.com/user-attachments/files/19268211/ybdcomm.txt)

happppp:
And those drivers compile fine if you don't include asio.h?

SailorSat:
> And those drivers compile fine if you don't include asio.h? Which would be strange, since the errors you posted look like GCC 11.2 bugs or the version is outdated.

it does compile fine with the "old" code that uses osdfile.h rather than asio.h

SailorSat:
Converted all my comm implementations to ASIO.
Thanks to @happppp for getting those compile issues solved.

My ASIO based comms do:
- actually "resolve" localhost/port and remotehost/port on startup once.
- connect and accept "async".
- use a timer to keep track of connection timeouts.
- keep track of the sockets status in a simple variable (0 disconnected / 1 connecting or waiting for accept / 2 connected)
- as long as either of the sockets is in status 1, poll the io_context to process any async event that might have happened (without blocking).
- before "reading", check if enough data is in the buffer by "peeking"
- tread a partial write as "error" and disconnect.
- set TCP_NO_DELAY for the outgoing socket to keep latency down.
- setup largert buffers on sockets - 64k for TX, 512k for RX (should be more than enough)

that way the code cannot block.
partial writes should never happen, as long as the connection is healthy.

@cuavas : I hope this is fine now.

mamehaze:
> > @cuavas Do you have any objections to this besides theoretical ones about MAME hanging? Because I'm gonna merge it if not, and then SailorSat will have a solid base to rework it for ASIO.
> 
> It isn't theoretical. Having MAME hang because some other random program is listening on a particular TCP port, or having it lock up if it loses connectivity is not an acceptable use experience.
> 
> I'm going to revert it if it's merged in this state, because if it _is_ merged like this, it's inevitably going to become one more thing I have to fix myself. (But in this case, it's lots more things to fix.)
> 
> If you're feeling under pressure because Haze announced that this will be in the next release, meeting it in a state like this isn't a solution.

This is more of your authoritarian nonsense that I find so rude.  "Rule by fear" in this case the threat of revert.  For 99% of people this would have been fine, and SailorSat would have made the changes either before or after.  You're moaning about something that the people using it have been perfectly happy with for other systems for years.

Also I'm not 'putting pressure on' SailorSat by covering this, I'm covering cool stuff that happens in MAME.  I can just not bother instead, and people will go back to thinking nothing of note has happened in MAME for years.

Any reasonable project management would have just merged it and let the kinks be ironed out, not continued to load more and more demands on a PR by moving goalposts.  It followed how things were being done in MAME, it worked, it could be built upon.  It only becomes *your* problem when you piss the people working on stuff so much they don't actually come back.

dukeeeey:
We have similar code in supermodel for networking to sailorsat's original code and no one has ever complained. It's never been an issue.

MrShooter42:
Funny how when SailorSat makes the requested changes, Cuavas goes radio silent. What gives?

MooglyGuy:
> Funny how when SailorSat makes the requested changes, Cuavas goes radio silent. What gives?

It's the Thursday before code-freeze weekend release. It's not that deep, and you're not as smart as you believe you are.

MrShooter42:
> > Funny how when SailorSat makes the requested changes, Cuavas goes radio silent. What gives?
> 
> It's the Thursday before code-freeze weekend release. It's not that deep, and you're not as smart as you believe you are.

Fair enough, I got too passionate about this thread and left a stupid comment. My apologies. That said, I still agree with you and MameHaze on this issue.

SailorSat:
So... almost 3 weeks of radio silence now?

scripts/src/machine.lua
```diff
@@ -2201,6 +2201,18 @@ if (MACHINES["MB89371"]~=null) then
 	}
 end
 
+---------------------------------------------------
+--
+--@src/devices/machine/mb89372.h,MACHINES["mb89372"] = true
+---------------------------------------------------
+
+if (MACHINES["mb89372"]~=null) then
```
happppp, line 2209: MACHINES["XXX"] is uppercase (3 lines above as well)

happppp:
@cuavas is it still in a state where you'd revert it if someone else merged it?

happppp:
Anyone else have feedback? If no, let's merge it 24hrs from now.

happppp:
@SailorSat BTW could you update your 1st post? Mention ASIO?
It'll be the commit msg (whatsnew.txt)

SailorSat:
to my knowledge, yes. everything is working as expected.

cuavas:
I’ll look at this again tonight.  Sorry, I’m already under a mountain of broken stuff with the recompilers.

src/devices/machine/k056230.h
```diff
@@ -11,6 +11,8 @@
 
 #pragma once
 
+#include "asio.h"
+
```
cuavas, line 15: Before I get to anything else, there’s a very good reason this shouldn’t be included from a device header.  If you look at the actual header (src/osd/asio.h) it has to suppress several useful diagnostics, including the very important `-Winconsistent-missing-override` for clang and `-Wsuggest-override` and `-Wunused-variable` for GCC.  We really *don’t* want that getting suppressed for everything downstream from a device.  It will inevitably lead to rot setting in, and then it will be the person trying to clean it up who has to fix it everywhere downstream.

That’s why in capcom/cps2comm.h I forward-declare `cps2_comm_device::context` and only `#include "asio.h"` in capcom/cps2comm.cpp.  It’s not just to be clever.  It also saves a bit of compile time, as ASIO is fairly heavy.

FWIW I’ve tried opening pull requests against ASIO to at least get it to use `override` qualifiers for the common networking parts, but I get no response whatsoever to pull requests there.

Anyway, please structure the code in a way that doesn’t require pulling in a header that suppresses diagnostics in device headers.  It’s too much of a liability.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +152,316 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+
+		default:
+			logerror("k056230-set_mode: %02x\n", data);
+			break;
+	}
+}
+
+void k056230_device::set_ctrl(u8 data)
+{
+	m_linkenable = data && 0x10;
```
cuavas, line 173: I’m pretty sure this isn’t what you want to do.  You’re testing that `data` is non-zero and the literal constant 0x10 is non-zero (which is isn’t).

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +152,316 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+								else
+								{
+									LOG("k056230: RX connection from %s\n", sock.remote_endpoint());
+									std::error_code e;
+									m_acceptor.close(e);
+									m_sock_rx = std::move(sock);
+									m_sock_rx.non_blocking(true);
+									m_sock_rx.set_option(asio::socket_base::receive_buffer_size(524288));
+									m_sock_rx.set_option(asio::socket_base::keep_alive(true));
+									m_rx_state = 2;
```
cuavas, line 250: You don’t want to set the non-blocking option.  You want to use the async read function so you can get notifications on your I/O service when data is available using `async_read_some`.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +152,316 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+		osd_printf_verbose("k056230: RX error receiving - %d %s\n", err.value(), err.message());
+		m_sock_rx.close(err);
+		m_rx_state = 0;
+		return 0;
+	}
+
+	if (bytes_read == data_size)
+		return bytes_read;
+	return 0;
+}
```
cuavas, line 447: You don’t want to be dealing with socket operations on the critical path.  You want to use `async_receive_some` to read into your buffer asynchronously, so you can just examine it here and see if you have enough to make a complete frame.  If you do, update the tail pointer of the buffer to make more room.

You always want to have an `async_receive_some` in flight as long as you have space in your receive buffer.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +152,316 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+
+	std::error_code err;
+	std::size_t bytes_sent = m_sock_tx.send(asio::buffer(&m_buffer0[0], data_size), 0, err);
+	if (err || bytes_sent != data_size)
+	{
+		osd_printf_verbose("k056230: TX error sending - %d %s\n", err.value(), err.message());
+		m_sock_tx.close(err);
+		m_tx_state = 0;
+	}
+}
```
cuavas, line 462: Using `send` here is problematic for the same reasons that MAME’s sockets are problematic.  If you use `send` with a non-blocking socket, you still need to keep a local buffer and retry writes when the socket has buffer space available, which is the part that ASIO is supposed to help you deal with.

You want to use `async_transmit_some` to get as much on the wire as possible, then when you get the callback, call it again as long as you still have buffered data to transmit.  All you should be doing here is buffering the data and starting an `async_transmit_some` if you don’t already have one in flight.

src/devices/machine/k056230.cpp
```diff
@@ -119,6 +152,316 @@ void k056230_device::ram_w(offs_t offset, u32 data, u32 mem_mask)
...
+
+							for (unsigned j = 0x00; j < frame_size; j += 4)
+							{
+								frame_offset = (frame_start + j) / 4;
+								frame_data = m_buffer0[4 + j] |  m_buffer0[3 + j] << 8 |  m_buffer0[2 + j] << 16 |  m_buffer0[1 + j] << 24;
+								ram_w(frame_offset, frame_data, 0xffffffff);
+							}
+
+							// forward message to other nodes
+							send_frame(data_size);
```
cuavas, line 388: Is it really a good idea to be doing the forwarding on the emulation thread?

Remember the only time the emulation thread is synchronised with the wall clock is at the end of a frame interval on the first emulated screen in the system.  This means that if you fire a timer to try and check for and forward frames multiple times per frame, that will all be compressed into the time while the emulation is running, then you’ll get a period while MAME waits for real time to catch up when your timer doesn’t fire at all.  The faster the emulation speed is, the more pronounced the effect is.

This is why for the `cps2_comm_device` I do the cut-through forwarding on a separate thread with its own ASIO I/O context.  That allows the data to be forwarded as soon as it arrives, *even if that happens while the instance in question is sleeping between emulating frames*.  Given that the CPS-2 is 68k-based and gets emulated wicked fast (MAME is sleeping a lot more of the time than it’s actually running the emulation), this is fairly important to avoid the high likelihood of having a relatively long wait between a frame arriving and being forwarded.

Given the X-board and Y-board are also 68k-based systems that modern PCs have no trouble emulating way faster than real time, you’ll be seeing the same effects for those systems.

src/devices/machine/k056230.h
SailorSat: I'll give that context-thingie a spin then.

cuavas (commented):
Look, to be honest, this is showing the issues that tend to come up from trying to do something like this all at once:

* It’s huge and touches a dozens of files, so trying to review it all at once is difficult and requires a lot of time at once from the reviewer.
* There’s a lot of it, so it’s easy for what should be obvious errors that should be caught *before* review to slip though (e.g. use of Boolean `&&` where it should almost definitely be bitwise `&` what I saw on a cursory glance at machine/mb89372.cpp).
* A bunch of the changes do similar things, so if there’s a fundamental flaw in one, all of them need to change.  This compounds if there are multiple rounds of changes needed.

It would be a lot easier to try and get one of these right, then follow the pattern for the rest of them than trying to do it all at once.

But putting that aside, I gave the example of the cps2comm.cpp because it’s a fairly simple case, but I’ve put a lot of effort into doing it right:

* ASIO is kept out of the device header.
* All the socket operations are asynchronous to let ASIO deal with the complexities of worrying about when a socket is actually ready.
* Cut-through forwarding is done on a separate worker thread with its own I/O context to avoid having it blocked while the main emulation thread sleeps between emulated frames.
* The basic synchronisation and asynchronous task management operations are all encapsulated in small functions to make it easier to debug and understand.

It feels like this change tries to avoid actually using asynchronous operations as much as possible and tries to force the ASIO sockets to behave like MAME’s sockets-pretending-to-be-files, which defeats the purpose of using ASIO in the first place.

Taking a step back, do you understand the basic premise of the reactor pattern?

* You have one or more I/O contexts.
* You add event sources to the I/O contexts.
* If one of those event sources has something to tell you, it will generate a callback.
* The callback will only happen while the I/O context is running (i.e. from inside a `run` or `poll` call).
* You can use `post` to have an arbitrary function called on an I/O context when it’s ready.

You don’t poll your sockets for available data, you start an asynchronous read, and let them tell you when data is available.

src/devices/machine/k056230.cpp
SailorSat: well... good catch - although it worked ;)
it more likely is meant to be `m_linkenable = (data & 0x10) ? 1 : 0;`

SailorSat:
> Look, to be honest, this is showing the issues that tend to come up from trying to do something like this all at once:
Yes. that happens when you refuse to accept code that has been working (in some form) in the codebase for years and **force** new code to conform to some _arbritary_ standard you have set.
You wanted me to use ASIO in the first place - now you complain I am using it "wrong".

> It feels like this change tries to avoid actually using asynchronous operations as much as possible and tries to force the ASIO sockets to behave like MAME’s sockets-pretending-to-be-files, which defeats the purpose of using ASIO in the first place.
Why? You stated you don't want blocking operations, which I complied to.
Just because I don't use ASIO the way _you_ did, does not make the code invalid.

> Taking a step back, do you understand the basic premise of the reactor pattern?
No, I do not. I already stated that I'm no C++ programmer.
 
> You don’t poll your sockets for available data, you start an asynchronous read, and let them tell you when data is available.
Why is peeking the OS provided receive buffer (via ASIO) worse than having to deal with thread-safety and checking the threaded async-receive buffer?

I don't see the point in providing double buffering - one in MAME itself, and the one the OS provides.
Like previously stated, the code has been working with the dreaded "sockets-pretending-to-be-files" for years without issues.

I will try to put that stuff into a context to avoid having asio.h in the device headers though.

galibert:
SailorSat, if your reaction to a detailed review explaining how things should be done is "nah, I don't feel like it", we can close the PR right now you know?

SailorSat:
That is not what I said :)

I just don't understand why a socket has to be double buffered, multithreaded and asynchronous, when the initial issue was that the could possibly block (and freeze the emulation) - which i solved.

I should be able to get the asio stuff in a "context" - however threading is way above my skill level for sure.

SailorSat:
[m1comm_asio_context.zip](https://github.com/user-attachments/files/19624858/m1comm_asio_context.zip)

So I tried moving the stuff into a context as a first step...
and I'm running into compile errors again - which I do not understand.

`[MINGW64] D:\Privat\cabmame>make
GCC 14.2.0 detected
Compiling src/mame/sega/model1.cpp...
In file included from D:/msys64/mingw64/include/c++/14.2.0/memory:78,
                 from ../../../../../src/emu/emu.h:27:
D:/msys64/mingw64/include/c++/14.2.0/bits/unique_ptr.h: In instantiation of 'void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = sega_m1comm_device::context]':
D:/msys64/mingw64/include/c++/14.2.0/bits/unique_ptr.h:398:17:   required from 'std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = sega_m1comm_device::context; _Dp = std::default_delete<sega_m1comm_device::context>]'
  398 |           get_deleter()(std::move(__ptr));
      |           ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
../../../../../src/mame/sega/m1comm.h:20:7:   required from 'void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = sega_m1comm_device]'
   20 | class sega_m1comm_device : public device_t
      |       ^~~~~~~~~~~~~~~~~~
D:/msys64/mingw64/include/c++/14.2.0/bits/unique_ptr.h:398:17:   required from 'std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = sega_m1comm_device; _Dp = std::default_delete<sega_m1comm_device>]'
  398 |           get_deleter()(std::move(__ptr));
      |           ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
../../../../../src/emu/mconfig.h:193:26:   required from 'auto machine_config::device_add(const char*, Creator&&, Params&& ...) [with Creator = const emu::detail::device_type_impl<sega_m1comm_device>&; Params = {int}]'
  193 |                 auto device(type.create(*this, owner.first, owner.second, std::forward<Params>(args)...));
      |                             ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../../src/emu/device.ipp:36:56:   required from 'DeviceClass& emu::detail::device_type_impl<DeviceClass>::operator()(machine_config&, const char*, Params&& ...) const [with Params = {int}; DeviceClass = sega_m1comm_device]'
   36 |         return dynamic_cast<DeviceClass &>(*mconfig.device_add(tag, *this, std::forward<Params>(args)...));
      |                                             ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../../src/mame/sega/model1.cpp:1787:18:   required from here
 1787 |         SEGA_MODEL1_COMM(config, "m1comm", 0).set_default_bios_tag("epr15112");
      |         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
D:/msys64/mingw64/include/c++/14.2.0/bits/unique_ptr.h:91:23: error: invalid application of 'sizeof' to incomplete type 'sega_m1comm_device::context'
   91 |         static_assert(sizeof(_Tp)>0,
      |                       ^~~~~~~~~~~
sega.make:1662: recipe for target '../../../../mingw-gcc/obj/x64/Release/src/mame/sega/model1.o' failed
make[2]: *** [../../../../mingw-gcc/obj/x64/Release/src/mame/sega/model1.o] Error 1
makefile:874: recipe for target 'sega' failed
make[1]: *** [sega] Error 2
makefile:1120: recipe for target 'windows_x64' failed
make: *** [windows_x64] Error 2`

SailorSat:
@cuavas : I contextified the ASIO stuff as requested.

As previously stated - Yes, my code can be optimised. But does it need to be "perfect" right now?
Having worked with (win)sock for almost 30 years, I still do not see the need for multithreaded doublebuffered asynchronous sockets.

That aside, I'd rather not try to thread them now, as I am way out of my comfort zone already.
I might give doing stuff in a thread a try someday - but I don't feel ready for that right now.

As previously stated - the primary concern was the emulation "blocking", which cannot happen anymore, thanks to ASIO handling connects.

having looked at cps2comm, I wonder how exactly it handles a full send buffer either.
calling send_in (or send_out) tries to append data to the buffer.
if the socket is not connected OR the send buffer is full, the data to send simply gets droped.

dukeeeey:
I think the suggestions are just over engineering. Even a blocking socket would almost never block for sending in this context because the amount of data being sent is so small it would never fill the TCP socket buffer.

But on the other end, if you can't block or pause the emulation waiting for data, can emulation even continue without said data?

In supermodel originally I wrote a heavily threaded solution with async send / receive wrapped around the SDL network classes. But in the end that complexity simply wasn't required.

MooglyGuy:
I get it, but between Vas and OG, I guess I would say they're ostensibly the respective equivalents of Tech Director and CTO for the project.

Like I said, I know from experience that Vas is phenomenal when explaining things 1:1 and when he's not under a ton of stress.

While this wasn't the source of his stress, I get the impression that his hand felt forced due to this PR being erroneously reported as being "in the next version" when that was far from a guarantee. This only compounded any stress, and resulted in soem miscommunication among all parties involved - the original mention of the CPS-2 comms board emulation could have been phrased as less of a suggestion and more of a requirement, for example.

And the end of the day, this is obviously a highly desirable feature to have in, but it could use a bit of refining. I don't always agree with every bit of code review that I get at my day job, but sometimes I have to take it on faith that the reviewer knows something I don't.

If I ask the person for an explanation, they take the time to give it, but in that case we're both on the clock. With MAME being a hobby - albeit a passionate one, certainly - for all involved, that timeliness of an explanation can't always be guaranteed.

I hope nobody is in a life and death situation depending on whether this PR goes in this month, the next month, or the next month. The overall magnitude of it _does_ merit particular care in examining it, and it also merits careful explanation in why certain things are required to be the way they are. But MAME isn't the Therac-25. Nobody's going to live or die. We can all take the time to breathe a bit, leave the PR open (or keep working on refining it, if the motivation exists), and reconvene when all parties involved have more availability.

SailorSat:
Updated K056230.
master/slave-mode, size-bits, id mask and interupts working.
still missing the "sub-id" features, although no game seems to use those.

Racin' Force links with up to 8 players, however it does not feature flying deloreans... we NEED roads :(

plygonet_linktest4_debug
racinfrc-linkplay8
racinfrc-linktest8

cuavas:
Sorry, haven’t forgotten this, just trying to get other stuff under control.

The number of recompiler bugs was *way* higher than I was expecting (mixture of “never actually worked”, “years old regression”, and “I wish I caught that *before* merging the AArch64 back-end PR”).  But I *think* that’s in at least half-decent shape at the moment, and I haven’t heard anyone saying that a pile of games have regressed in 0.277.

There are a couple of things I need to do that kind of relate to the new sound stuff before I can get on to other stuff.

I’m sure there are ways we can move forward with this PR, but the size of it and number of things it touches make it a bit unwieldy.

SailorSat:
I can break this up into several stand-alone PRs, if that makes things more easy.

SailorSat:
I didn't realize, there is a spectator mode for Suzuka 8hrs too.

image
https://www.youtube.com/watch?v=sG1zyNSWWW4

SailorSat:
@cuavas : something like this?

https://github.com/SailorSat/cabmame/blob/m1comm_0277/src/mame/sega/m1comm.cpp

SailorSat:
So I got to a point where the ASIO stuff is fully asynchronous on a dedicated thread.

however, I have to note that the whole thing now feels way more unstable than before.
framesync on "virtua formula" froze the emulation several times for unknown reasons.
i need several tries to get "ridge racer 2" to link up at all now, and it looses network after some time.
some games don't link up at all anymore.

why is that? because the code in the those games was never supposed to handle different speeds.
at last with the 3d games, the time emulation requires depends on the frame/viewport.
without the network "blocking" the emulation at some point, it goes out of sync pretty quickly.

so right now I have 4 "ways of doing it".

1. old skool osd_file sockets.
worked for years without issues.

2. "wrong use" of ASIO (synched "non-blocking" sockets)
seems to work for me.

3. async, polled ASIO.
to be honest, I didn't fully test every game. seemed to work.

4. async, threaded ASIO.
it works, but it causes all sorts of issues in practice.

so, how am I supposed to handle this?
I still don't see the need for async operations and a dedicated thread.
Even more so as it turns out to make communication even more unstable as it is.

**EDIT**
linking multiple instances of MAME for network play IS a very niche use-case.

-kinda off-topic-
there are some cases like "drivers eyes", where one MAME instance emulated three whole boardsets, because it was "sold as one unit".
there are other cases, like "ridge racer full-scale", where MAME currently only emulates one boardset, and you have to use 3 instances to get it running.

in those "multi screen" cases, I'm still unsure how MAME should handle those in the first place. given that we have multicore CPUs nowadays, I'd rather run 3 instances of mame in full speed, than having one instance that might slow down.

actual multiplayer games... while running a twin virtua racing on one machine with dual screens (and two sets of inputs) is perfectly possible, i don't think that is a common use-case.
it seems more likely to have two dedicated machines, running one instance of the game by itself.

I guess there are no plans for any "sync multiple instances"-features in the works?

SailorSat:
I've split this up into several standalone PRs. - this PR can be considered "obsolete" now.
(mb89374 and mb89372 will be done at a later point)

I just want to get these comms 'in' in a usable state by now, as they have been lying around in my fork for quite some time - and constantly jumping branches is a P.I.T.A. to be honest.
yes, the code is not perfect, and probably large parts will be rewritten soon - but the current code is WORKING at least.

I have a sneaking suspicion my multithreaded code has all sorts of race condition issues.
which is to be expected after all - i have zero experience with multithreading in that regards.

I've spent way to much time, energy and last-but-not-least money on that kind of stuff.
And having a serious depression is not helping either.

--

as all I have to do to "switch operation mode" is replace the context class, I have done some intense testing...

running multithreaded mode breaks framesync for unknown reasons in m1comm/m2comm/s32comm.
it works fine for a while, then all instances suddenly freeze, as if a packet was "droped" - which should not be possible as the same code in polled-mode works fine.
i will have to do some more debugging on that at a later point.

likewise, SMGP seems to "hang" (without freezing) in multithreaded mode, but works fine in polled mode.
also, running SMGP with a timer, rather than "on status register access" makes the game very jittery - reverted back to register access ticks.


for now I'm, trying to source an actual X-Board or Y-Board comm to test a real MB89372, as I did with the MB89374...

IMG_20250321_093149


I have two C139s on the table for testing. - I really need to pin down what causes a transmission to start in each mode.
I might swap in a C422 at a later point, as some of my system 22s have C422 instead of C139.

IMG_20250331_192129

IMG_20250331_210944

PR #13422, opened 2025-02-27
Multiple synth layouts: updated to the latest slider script.

m1macrophage:
* Explicit configuration: Using add_* functions, instead of scanning through the layout for sliders.
* Separated the copy-pasteable portion of the script.
* A no-op for the DMX and LinnDrum layouts.
* Fixes multitouch and slider "drift", and adds support for relative adjustments to the D70 sliders (they were using an older version of the script).


m1macrophage:
@felipesanches FYI.

PR #13423, opened 2025-02-27

tabudz:
**Description**
This PR fixes a security vulnerability in start_input_tga() that was cloned from libjpeg-turbo but did not receive the security patch. The original issue was reported and fixed under https://github.com/libjpeg-turbo/libjpeg-turbo/commit/82923eb93a2eacf4a593e00e3e672bbb86a8a3a0.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2018-11212
https://github.com/libjpeg-turbo/libjpeg-turbo/commit/82923eb93a2eacf4a593e00e3e672bbb86a8a3a0

### DIFF
--- a/3rdparty/libjpeg/rdtarga.c
+++ b/3rdparty/libjpeg/rdtarga.c
@@ -365,7 +365,8 @@ start_input_tga (j_compress_ptr cinfo, cjpeg_source_ptr sinfo)
       width <= 0 || height <= 0 ||
       source->pixel_size < 1 || source->pixel_size > 4 ||
       (UCH(targaheader[16]) & 7) != 0 || /* bits/pixel must be multiple of 8 */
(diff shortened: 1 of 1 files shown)


ajrhacker:
I'm not sure how this would fix any real problem, given that `width == 0 || height == 0` is redundant with `width <= 0 || height <= 0` only a few lines above in the same conditional.

JimCarlTay:
The person behind this says it's to fix a security flaw.

cracyc:
The upstream fix linked doesn't include the `width <= 0 || height <= 0` line so it's fixing something that's already been fixed.  Plus I don't think anything in mame uses targa image files.

cuavas:
MAME doesn’t load TARGA images, and this change looks wrong considering the rest of the `if` statement.  No point merging it, we’ll get whatever comes next time we update libjpeg.

PR #13424, opened 2025-02-27

tabudz:
**Description**
This PR fixes a security vulnerability in singlevar() that was cloned from lua but did not receive the security patch. The original issue was reported and fixed under https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2022-28805
https://github.com/lua/lua/commit/1f3c6f4534c6411313361697d98d1145a1f030fa


cuavas:
I’ll update to Lua 5.4.7 at some point and we’ll get the fix then.  We don’t use constant `_ENV` in any of our plugins or layouts.  If a user tries using constant `_ENV` in their own script (e.g. an autoboot script) it just won’t work.

PR #13430, opened 2025-02-28
Gfx save

Deedolith:
Implemented the possibility to save the currently displayed gfx set using the snapshot hotkey (default F12).
The file (PNG) will be saved under the gfxsave directory wich will be created if needed).

### DIFF
--- a/src/frontend/mame/ui/gfx_writer.cpp
+++ b/src/frontend/mame/ui/gfx_writer.cpp
@@ -11,12 +11,12 @@
 #include "emu.h"
 #include "gfx_writer.h"
 
-bitmap_rgb32 gfxWriter::getBitmap(int xCells, int yCells, gfx_viewer::gfxset::setinfo& set, gfx_element& gfx) const
+bitmap_argb32 gfxWriter::getBitmap(int xCells, int yCells, gfx_viewer::gfxset::setinfo& set, gfx_element& gfx) const
 {
 	auto cellXpix{ (set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.height() : gfx.width() };
 	auto cellYpix{ (set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.width() : gfx.height() };
 
-	bitmap_rgb32 bitmap;
+	bitmap_argb32 bitmap;
 	bitmap.reset();
 	bitmap.allocate(cellXpix * xCells, cellYpix * yCells);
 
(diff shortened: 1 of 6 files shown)


Deedolith:
Hello,
It's been over 2 weeks since I posted this request, and nothing happened since then.
What's going on ?
Is there something missing ? Wrong ?
Is it rejected ?

Please, let me know.

JimCarlTay:
It looks like workflows are still awaiting approval.

src/frontend/mame/ui/gfx_writer.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "main.h"
+#include "png.h"
+#include "video.h"
+#include "ui.h"
+#include "emu.h"
+#include "gfx_writer.h"
+
+bitmap_argb32 gfx_writer::getBitmap(int xCells, int yCells, gfx_viewer::gfxset::setinfo& set, gfx_element& gfx) const
+{
+	auto cellXpix{ (set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.height() : gfx.width() };
```
galibert, line 16: You seem to like that method of initialization for scalar variables.  I know it's correct, but it's extremely unusual.  We try to avoid unusual when possible, so it would be nice to switch back to the standard '='.

src/frontend/mame/ui/viewgfx.cpp
```diff
@@ -2,1605 +2,1478 @@
...
+	m_pointer_inside = false;
+	m_bitmap_dirty = true;
+	return mame_ui_manager::HANDLER_CANCEL;
+}
 
-		bool indirect() const noexcept
-		{
-			return subset::INDIRECT == m_which;
-		}
+uint32_t gfx_viewer::handle_palette(mame_ui_manager& mui, render_container& container, bool uistate)
```
galibert, line 77: You changed the position on & and * w.r.t the type and variable name everywhere.  People could write books about where it makes sense, but at least mame is almost consistently putting & and * on the variable, so please don't change that, and please do it that way in new code (e.g. gfx_writer.cpp).  Yes, we know Mame is not perfectly consistent everywhere (there are two main styles for a start, K&R and Allman), and we will not do gratuitous code churn to align things everywhere, but in that area we almost are.

src/frontend/mame/ui/gfx_writer.cpp
Deedolith: Noted.
I updated te source code accordingly.

Deedolith:
Hello,

After over a month, still no approval or rejection.
Sad to say, but that's disapointing ...

Deedolith:
Closing this pull request since it is now conflicting with current version.

PR #13441, opened 2025-03-04
Add a Skeleton driver for Honeywell CAMIR-F1 motion sensor

clawgrip:
CAMIR-F1 [ArcadeHacker]

### DIFF
--- a/src/mame/skeleton/camirf1.cpp
+++ b/src/mame/skeleton/camirf1.cpp
@@ -1,0 +1,76 @@
+// license: BSD-3-Clause
+// copyright-holders:
+
+/*************************************************************
+
+Skeleton driver for Honeywell CAMIR-F1 wireless passive
+infrared motion sensor with built-in colour camera.
+
+Hardware based on a PIC18LF26K22 as main CPU with a Conexant
+CX93610 (DIFT JPEG Encoder with a 656 Camera Interface and
+Optional Microphone Input) and a IS25LQ040 Serial Flash.
+
+PCB:
+  _________    _    _________
+ /         |__| |__|         \
+|                             |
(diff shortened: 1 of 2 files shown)


src/mame/skeleton/camirf1.cpp
```diff
@@ -0,0 +1,76 @@
+// license: BSD-3-Clause
+// copyright-holders:
+
+/*************************************************************
+
+Skeleton driver for Honeywell CAMIR-F1 wireless passive
+infrared motion sensor with built-in colour camera.
+
+Hardware based on a PIC18LF26K22 as main CPU with a Conexant
+CX93610 (DIFT JPEG Encoder with a 656 Camera Interface and
```
angelosa, line 10: `BT.656`

src/mame/skeleton/camirf1.cpp
```diff
@@ -0,0 +1,76 @@
...
+	void camirf1(machine_config &config) ATTR_COLD;
+};
+
+INPUT_PORTS_START( camirf1 )
+INPUT_PORTS_END
+
+void camirf1_state::camirf1(machine_config &config)
+{
+	// PIC18LF26K22
+}
```
angelosa, line 61: Is this core unique enough to have no CPU hooked up?

src/mame/skeleton/camirf1.cpp
clawgrip: Fixed

src/mame/skeleton/camirf1.cpp
clawgrip: Fixed, using PIC16 as the nearest choice.

Thx!

PR #13442, opened 2025-03-04
snk/munchmo.cpp: Updates/Cleanups

cam900:
- Cleanup gfx decode layouts, Use GFXDECODE_SCALE for internally scaled gfxs
- Split tile index LUT ROM and tile GFX ROM region for cleanup gfx decode layout and Fix ROM region naming confusion
- Fix save state support
- Fix typename for boolean flag
- Fix naming

### DIFF
--- a/src/mame/snk/munchmo.cpp
+++ b/src/mame/snk/munchmo.cpp
@@ -132,10 +132,12 @@ void munchmo_state::palette(palette_device &palette) const
 
 	for (int i = 0; i < palette.entries(); i++)
 	{
+		int bit0, bit1, bit2;
+
 		// red component
-		int bit0 = BIT(color_prom[i], 0);
-		int bit1 = BIT(color_prom[i], 1);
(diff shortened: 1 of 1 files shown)


src/mame/snk/munchmo.cpp
```diff
@@ -132,12 +132,10 @@ void munchmo_state::palette(palette_device &palette) const
...
-		int bit0, bit1, bit2;
-
 		// red component
-		bit0 = BIT(color_prom[i], 0);
-		bit1 = BIT(color_prom[i], 1);
-		bit2 = BIT(color_prom[i], 2);
+		int bit0 = BIT(color_prom[i], 0);
+		int bit1 = BIT(color_prom[i], 1);
+		int bit2 = BIT(color_prom[i], 2);
 		int const r = 0x21 * bit0 + 0x47 * bit1 + 0x97 * bit2;
```
cuavas, line 139: These locals are declared separately because they’re reused for essentially different things throughout the function.

PR #13446, opened 2025-03-05
Update rmhaihai.cpp

Gently28:


cuavas:
I added the DIP switch locations and comments in 773a4c8f6486394cd50c8c24b3baabaf17aa4a26.

PR #13451, opened 2025-03-07
z80/z80dasm.cpp: Switched *R (LDIR, OTIR, etc) debugger flag to STEP_OVER

holub:
STEP_COND on repeatable instructions doesn't allow to skip execution with F10
STEP_OVER is more reasonable for this instructions as allowed to take advantage from F10/F11

Per my understanding that should close match expectation:
`static constexpr u32 STEP_OVER       = 0x20000000;   // this instruction should be stepped over by setting a breakpoint afterwards`

### DIFF
--- a/src/devices/cpu/z80/z80dasm.cpp
+++ b/src/devices/cpu/z80/z80dasm.cpp
@@ -33,10 +33,10 @@ static const char *const s_mnemonic[] =
 const u32 z80_disassembler::s_flags[] =
 {
 	0        ,0        ,0        ,0        ,STEP_OVER,0    ,0        ,0        ,
-	STEP_COND,0        ,STEP_COND,0        ,0        ,0    ,0        ,0        ,
+	STEP_OVER,0        ,STEP_OVER,0        ,0        ,0    ,0        ,0        ,
 	STEP_COND,0        ,0        ,0        ,STEP_OVER,0    ,0        ,0        ,
-	0        ,STEP_COND,0        ,STEP_COND,0        ,0    ,0        ,0        ,
-	STEP_COND,0        ,STEP_COND,0        ,0        ,0    ,STEP_COND,STEP_COND,
+	0        ,STEP_OVER,0        ,STEP_OVER,0        ,0    ,0        ,0        ,
+	STEP_OVER,0        ,STEP_OVER,0        ,0        ,0    ,STEP_OVER,STEP_OVER,
 	0        ,0        ,0        ,0        ,0        ,0    ,STEP_OUT ,STEP_OUT ,
 	STEP_OUT ,0        ,0        ,0        ,0        ,0    ,0        ,0        ,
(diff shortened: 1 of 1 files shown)


cuavas:
I think at some point in the past, at least `djnz` *did* have the “step over” – it apparently changed in c7597225b4f0a3531d53dd806048b136956eb850.  @ajrhacker and @galibert can you explain how the flags are supposed to be used for instructions like this?  Is it supposed to be possible to skip over these things when they use the `STEP_COND` flag?

cuavas:
Is the `gni` debugger command on a looping instruction supposed to run until the instruction falls through?

holub:
The problem with djnz it can used either as loop with back reference or as "conditional jp" and jump forward.

ajrhacker:
`gni` is precisely equivalent to `go X`, where X is the address of the next instruction in the disassembly, and works regardless of what the instruction does. You can also do `gni N`, though I set an arbitrary limit on N because it has to look up each instruction successively.

The `over` command works sort of like `gni` when it encounters an instruction flagged with `STEP_OVER`, and does ordinary stepping otherwise. There's some tricky logic to handle delay slots (which Z80 doesn't have) and conditional calls (which the Z80 does have). It's a bit _over_-complicated, but MAME has had it for a while longer than `gni`.

There are a few calling conventions which neither `gni` nor `over` can handle properly. One is where the subroutine call instruction is followed by an arbitrary number of inline operands. Another is subroutines with multiple returns.

As for the Z80 instructions in question:

DJNZ is now treated as the conditional branch instruction it is. It doesn't really make sense to flag it `STEP_OVER`, given that it doesn't necessarily form a closed loop.

LDIR, INIR, OTIR, etc. are executed by the Z80 much like a combination of their seldom-used single-shot counterparts (LDI, INI, OUTI) with a DJNZ back to the instruction itself (the conditional timing penalties are similar). `STEP_OVER` would work more reliably for these instructions, though it's still not logically correct since they don't perform subroutine calls.

One idea I've entertained is adding a live branch predictor for MAME CPU cores to implement through `device_disasm_interface`. It would take the address of a conditional instruction, preferably that of the current program counter or something very close to it, and predict if a branch would be taken or not.

ajrhacker:
@holub: It was not actually my intention to lend wholehearted approval to this. That said, I don't intend to revert it quickly.

At least there aren't too many CPUs that really handle specific repeated operations by refetching the instruction each time. Most of them bump the PC back only if an interrupt is taken, if the operation is interruptible at all.

holub:
@ajrhacker I'm really sorry. I read following as a confirmation that intention is right.

> LDIR, INIR, OTIR, etc. are executed by the Z80 much like a combination of their seldom-used single-shot counterparts (LDI, INI, OUTI) with a DJNZ back to the instruction itself (the conditional timing penalties are similar). STEP_OVER would work more reliably for these instructions, though it's still not logically correct since they don't perform subroutine calls.

I tested the change and think this way debbugger behave better. If there are any concerns, please feel free to revert.

PR #13456, opened 2025-03-09
psx.xml: Added 11 working items + 1 not working + 4 redumped

ArcadeShadow:
2002 FIFA World Cup (Europe) [Redump]
2002 FIFA World Cup (France) [Redump]
2002 FIFA World Cup (Germany) [Redump]
2002 FIFA World Cup (Italy) [Redump]
2002 FIFA World Cup (Spain) [Redump]
3, 2, 1, Smurf! My First Racing Game (Europe) [Redump]
Ace Combat (Japan) [Redump]
Air Combat (Europe) [Redump]
Air Combat (Europe) (EDC) [Redump]
Darkseed (Japan) [Redump]
Transport Tycoon 3D: SL Kara Hajimeyou!! (Japan) [Redump]

New NOT working software list addition
Bedlam (Europe) [Redump]

Redumped software list items
2002 FIFA World Cup (USA) [Redump]
Air Combat (USA) [Redump]
Smurf Racer! (USA) [Redump]
Darkseed II (Japan) [Redump]

### DIFF
--- a/hash/psx.xml
+++ b/hash/psx.xml
@@ -31,12 +31,16 @@ For an overview of US PS1 discs that have not been dumped and added to redump's 
 
 
 	<software name="007rcng">
+		<!--
+		redump.org/disc/633
+		<rom name="007 Racing (USA).cue" size="82" crc="5c58d484" sha1="f74c52098f149c91f066b17e9000d5a059ec4405"/>
+		<rom name="007 Racing (USA).bin" size="565011552" crc="5a39ae64" sha1="91c0c4b5c888f074a777f891be5432064678070e"/>
+		-->
 		<description>007 Racing (USA)</description>
 		<year>2000</year>
 		<publisher>Electronic Arts</publisher>
 		<info name="developer" value="Eutechnyx"/>
 		<info name="language" value="English"/>
-		<info name="redump.org" value="633"/>
 		<info name="serial" value="SLUS-01300, SLUS-01300CE"/>
 		<sharedfeat name="compatibility" value="NTSC-U"/>
 		<part name="cdrom" interface="cdrom">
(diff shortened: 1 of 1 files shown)


hash/psx.xml
```diff
@@ -60214,6 +60119,24 @@ The entries in this section are intended to replace the existing "low-grade" Jap
 		</part>
 	</software>
 
+	<software name="bedlam" supported="no">
+		<description>Bedlam (Europe)</description>
+		<year>1997</year>
+		<publisher>GT Interactive</publisher>
+		<notes>Scrambed graphics</notes>
```
angelosa, line 60126: Scrambled, even?

ArcadeShadow: Thanks. Corrected the typo.

cuavas:
Can we get some discussion on the major change that this PR makes that isn’t mentioned in the description?  It adds metadata items with redump.org numbers in them.  Is this really where we want to go?  There’s no guarantee these will be stable or available long term.  There’s already a history of people putting all their eggs in one basked only for it to disappear (e.g. undumped wiki, where the promised database dump was never delivered).

ArcadeShadow:
I understand your perspective. It was proposed in the PR #13347 that this approach could serve as a means to avoid directly including and identifying the tracks contained within the CHD, especially considering that such information is not accessible through MAME itself but only via the XML file.
It is also worth highlighting that this information is important to know the source that was used for the creation of the CHD.
I acknowledge the concern regarding the inclusion of third-party identifiers, which may not be reliable or available in the long term. Would there be an alternative solution to address this matter? Or, for the time being, would it be more prudent to retain this information in the form of comments?

FoxhackDN:
I have hesitated to add Redump ID numbers to my recent software list additions; they're the only source of some metadata that isn't saved in CHDs (the Track index stuff) and it is important to have a reference handy for whenever that's added.

Regardless, I *heavily prefer* using comments instead of an XML tag. Some Redump items *have* been removed in the past, or replaced with correct dumps. And sometimes the Redump pages have info about the dump itself. I know there's concerns about the reliability / longevity of the site, but people have been working to mirror the data hosted there, so it's not going to be completely lost.

angelosa:
> Is this really where we want to go?

The alternative would be scraping redump.org for good and add all the info we don't have in these comments, something I won't be doing because I already loathe over how these half baked sections already litters our CD based XMLs, imagine adding extra verbosity on my own.

ArcadeShadow:
Well, in that case I will remove the redump.org id in metadata item info, and add the tracks info in comments.
Maybe in the future, there could be a complementary file with that information. That way, it wouldn't overload the current file.

ArcadeShadow:
Changes made. @cuavas check if it's ready to go.

PR #13459, opened 2025-03-10
nintendo/gba.cpp: Updates/Cleanups:

cam900:
- Implement master volume for PSG and directsound
- use required_device_array for DACs
- Fix FIFO sound emulations
- Use struct for FIFO sound
- Fixes typename of member variable for fix save state support

reference: https://problemkaputt.de/gbatek.htm#gbasoundcontroller


src/mame/nintendo/gba.cpp
```diff
@@ -865,22 +865,28 @@ void gba_state::gba_io_w(offs_t offset, uint32_t data, uint32_t mem_mask)
 
 			if (ACCESSING_BITS_16_31)
 			{
+				// master volume, ((data >> 16) & 3) == 3 is prohibited?
+				m_gbsound->set_output_gain(ALL_OUTPUTS, float(1 << ((data >> 16) & 3)) / 4.0f);
+				m_ldac[0]->set_output_gain(ALL_OUTPUTS, float(1 + BIT(data, 18)) / 2.0f);
+				m_rdac[0]->set_output_gain(ALL_OUTPUTS, float(1 + BIT(data, 18)) / 2.0f);
+				m_ldac[1]->set_output_gain(ALL_OUTPUTS, float(1 + BIT(data, 19)) / 2.0f);
+				m_rdac[1]->set_output_gain(ALL_OUTPUTS, float(1 + BIT(data, 19)) / 2.0f);
```
angelosa, line 873: Not sure I like the idea of 5 floating point divisions back-to-back for something that just seems simple `{ 0.5, 1.0 }` tables for the DACs, `{ 0.25, 0.5, 0.75, 1.0 }` (as your code) or `{ 0.3(3), 0.66(6), 1.0, 1.0? }` (I guess) for the GBA sound.

Also: if 3 is "prohibited" then it's worth a `popmessage` or a `logerror`, depending if behaviour is actually known or not.

PR #13462, opened 2025-03-10
Add CUBE EuroBEEB (BBC Micro variant) as not working

clawgrip:
New clones marked not working
-----------------------------
CUBE EuroBEEB [Lee Davison]

### DIFF
--- a/src/mame/acorn/bbc.cpp
+++ b/src/mame/acorn/bbc.cpp
@@ -2134,6 +2134,39 @@ void bbcm_state::autoc15(machine_config &config)
 }
 
 
+/***************************************************************************
+
+    CUBE EuroBEEB Single Board Computer PCB from Control Universal Ltd
+
+****************************************************************************/
+
+void bbc_state::eurobeeb(machine_config &config)
+{
+	bbca(config);
+
+	/* TODO: CUBE EuroBEEB hardware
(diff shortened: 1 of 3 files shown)


Pernod70:
This doesn't belong in bbc.cpp, it's a simple 6502 based board that runs BBC BASIC. Even the VIA and ACIA don't match the BBC memory map. I've been working on a eurocube.cpp that is similar to acrnsys.cpp in that it emulates the individual boards, but not quite ready yet.
Having said that I don't mind the ROMs being left here so they can be credited, and I'll move them over to my WIP when ready.

clawgrip:
> This doesn't belong in bbc.cpp, it's a simple 6502 based board that runs BBC BASIC. Even the VIA and ACIA don't match the BBC memory map. I've been working on a eurocube.cpp that is similar to acrnsys.cpp in that it emulates the individual boards, but not quite ready yet. Having said that I don't mind the ROMs being left here so they can be credited, and I'll move them over to my WIP when ready.

Ah! That's great! I'll close this PR then.

Can you please have a look at PR #13460  ?? It's for Optomax V, an Image Analyser that uses an EuroBEEB card for OSD on the integrated screen (it's usually used with two monitors, one external and one integrated).

PR #13464, opened 2025-03-11
cpu/drcbearm64.cpp: Emit optimised memory access code

cuavas:
This is more of the stuff recommended by @galibert – I need help testing this on a 64-bit ARMv8 system.

The basic premise is that:
* The address line count, global address mask and dispatch table pointers for an address space don’t change after setup.  This means you can treat them as immediate constants/PC-relative addresses and avoid accessing the address space object altogether.
* In practice, the global address mask is always a contiguous sequence of set bits.  This means the address can be masked with an AND immediate instruction.  The dispatch table index can also be extracted with a UBFX instruction before the mask is applied for slightly better pipeline utilisation.

I’ve used some functions to reduce repetition, and I’ve commented to code to explain what it’s doing (not just the instructions it emits).  I’ve read the code multiple times and hand-simulated it, but I may well have missed something.  Use of functions should also make fixing potential screw-ups simpler.

* If I’ve screwed up native width reads or writes, basically everything using the DRC should break.
* If I’ve screwed up writes narrower than native width for little Endian CPUs, `sf2049` should fail to boot.
* If I’ve screwed up writes narrower than native width for big Endian CPUs, `pmac6100 mac750` should show graphical corruption during the boot process.

Some good things to test:
* `sf2049` boots to attract mode (little Endian CPU, 32-bit data bus)
* `pmac6100 mac750` boots to the desktop without graphical corruption during the boot process (big Endian CPU, 64-bit data bus)
* `finfurl` boots to attract mode (tests that global address mask is being applied correctly)
* `coolmini` boots to attract mode (16-bit data bus)
* `fiveside` boots to attract mode (less than 32 address lines)

Also, I’d appreciate it if someone could get me the logged generated code from `-bench 1 -drc_log_native` for the following so I can sanity check it:
* `-bench 1 -drc_log_native sf2049`
* `-bench 1 -drc_log_native pmac6100`
* `-bench 1 -drc_log_native finfurl`
* `-bench 1 -drc_log_native coolmini`
* `-bench 1 -drc_log_native fiveside`

These changes are more likely to improve performance on systems with weaker memory interfaces and/or narrower decode/issue stages (e.g. ARM Cortex CPUs are more likely to benefit than Apple M series CPUs).  If anything’s going to show a pronounced improvement it’s probably `fiveside`.

### DIFF
--- a/src/devices/cpu/drcbearm64.cpp
+++ b/src/devices/cpu/drcbearm64.cpp
@@ -541,6 +541,7 @@ private:
 	void emit_float_str_mem(asmjit::a64::Assembler &a, const asmjit::a64::Vec &reg, const void *ptr) const;
 
 	void emit_memaccess_setup(asmjit::a64::Assembler &a, const be_parameter &addrp, const memory_accessors &accessors, const address_space::specific_access_info::side &side) const;
+	void emit_narrow_memwrite(asmjit::a64::Assembler &a, const be_parameter &addrp, const parameter &spacesizep, const memory_accessors &accessors) const;
 
 	void get_carry(asmjit::a64::Assembler &a, const asmjit::a64::Gp &reg, bool inverted = false) const;
 	void load_carry(asmjit::a64::Assembler &a, bool inverted = false) const;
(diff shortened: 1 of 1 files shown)


grant2258:
> * `sf2049` boots to attract mode (little Endian CPU, 32-bit data bus)
> 
> * `pmac6100 mac750` boots to the desktop without graphical corruption during the boot process (big Endian CPU, 64-bit data bus)
> 
> * `finfurl` boots to attract mode (tests that global address mask is being applied correctly)
> 
> * `coolmini` boots to attract mode (16-bit data bus)
> 
> * `fiveside` boots to attract mode (less than 32 address lines)

* I couldn't test sf2049,  if no one else pipes in ill find it and test tomorrow. All the tests went to attract 

* pmac6100 mac750 never noticed any corruption and it booted fine

* finfurl had an odd blue image it could be fine just runs slow on the pi never played the game before just pointing that out.

* coolmini boot fine the attract was fine, the title screen at the start looked like the gfx where lagging could and it was running 100%

I tested on a pi5 cross compiled with clang 


here are the bench results you asked for.
[bench.zip](https://github.com/user-attachments/files/19198242/bench.zip)

rb6502:
sf2049 runs fine with the changes.  coolmini was the big speed gain, 100% faster (2400% vs 2500%, mind you).  fiveside was consistently 10-15% faster, and everything else was a wash.

PR #13467, opened 2025-03-12
Update dynax.cpp

Gently28:


cuavas:
Besides not building, this removes the note indicating that the DIP-SW4:9 settings are apparently reversed compared to what the manual says.

PR #13471, opened 2025-03-13
language: Belarusian

holub:


holub:
I found someone who agreed to provide a second opinion.
We can merge as is and I'll update later or wait a bit (hopefully that doesn't take much time)

holub:
Good overall

Issue #13475, opened 2025-03-14
64-bit ARMv8 recompiler back-end will crash for some forms of ADDC/SUBB


cuavas:
The 64-bit ARMv8 back-end uses the function template `op_add` to generate code for UML `ADD` and `ADDC`, and the function template `op_sub` to generate code for UML `SUB` and `SUBB`.  The asmjit instruction ID to use is passed as a template argument (`kIdAdds`/`kIdAdcs` for `ADD`/`ADDC`, and `kIdSubs`/`kIdSbcs` for `SUB`/`SUBB`).

Both `op_add` and `op_sub` detect situations where immediate forms of the emitted add/subtract instruction should be used.  This works for `adds` and `subs`.  However, AArch64 does not have immediate forms for `adc`/`adcs` and `sbc`/`sbcs`.  This means it will blow up if the either source operand for `ADDC` is a suitable immediate value for `add`/`addc`/`sub`/`subc`, or if the second source operand for `SUBB` is a suitable immediate value for `add`/`addc`/`sub`/`subc`.

We seem to be getting very lucky at the moment in that none of the recompiler front-ends seem to actually emit `ADDC` or `SUBB` with immediate operands.  However, I’m very uncomfortable with stuff that can blow up on valid inputs, and UML itself does not place any restrictions on using immediate source operands for `ADDC`/`SUBB`.

Either `op_add` and `op_sub` need to be reworked to suppress using immediate forms for `ADDC`/`SUBB`, or they should be split into separate functions.

PR #13476, opened 2025-03-14
cpu/drcbearm64.cpp: Fix issues with and optimise ADD[C]/SUB[B] code generation

cuavas:
This attempts to fix potential crashes if a recompiler front-end uses immediate operands with `ADDC`/`SUBB` as described in GitHub #13475, as well as optimising more cases where register loads can be avoided.

I’d appreciate review and/or testing.


cuavas:
I got @galibert to check that the code looks sane.  I'm going to merge this.  There's a bigger performance sink in the ARMv8 recompiler that's going to keep me awake until I deal with it.

Issue #586, opened 2016-01-25
Game metadata should be in the original language.


trap15:
Using romanizations for game metadata is fairly lossy and at the very least it's poor documentation. I propose there should be a secondary title string for the original language, keeping the romanized version.

Ideally, the developer could also use UTF-8 to write it in, instead of using escaped unicode. For this, srcclean needs to stop horribly destroying unicode for no reason.

felipesanches:
Some proposals:

This works, but it terribly unreadable.
```
GAME( 1992, wwmarine, 0, segac2, wwmarine, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Marine \u308f\u304f\u308f\u304f\u30de\u30ea\u30f3", 0 )
GAME( 1991, soniccar, 0, segac2, soniccar, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Sonic Patrol Car \u308f\u304f\u308f\u304f\u30bd\u30cb\u30c3\u30af\u30d1\u30c8\u30ab\u30fc", 0 )
```

This seems much better!
```
GAME( 1992, wwmarine, 0, segac2, wwmarine, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Marine わくわくマリン", 0 )
GAME( 1991, soniccar, 0, segac2, soniccar, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Sonic Patrol Car わくわくソニックパトカー", 0 )
```

trap15:
I'd like to put it in two different fields, so something like:

```
GAME( 1992, wwmarine, 0, segac2, wwmarine, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Marine", "わくわくマリン", 0 )
```

Or for fields where they're the same, maybe something like 

```
GAME( 1992, wwmarine, 0, segac2, wwmarine, segac2_state, bloxeedc, ROT0, "Sega", "Waku Waku Marine", "", 0 )
```

felipesanches:
What about in the case of accented latin text?

Could it be like this?

```
COMP( 1972, patinho,  0,        0,      patinho_feio,  patinho_feio, patinho_feio_state, patinho_feio, "Escola Politécnica - Universidade de São Paulo", "", "Patinho Feio", "", MACHINE_NO_SOUND_HW | MACHINE_NOT_WORKING)
```

Bear in mind that both the "company" and the "full name" fields may have unicode characters...
[...]

PR #13477, opened 2025-03-14
konami/mystwarr.cpp - Add subtitle to mystwarr

mikehyrus:
This PR adds the subtitle to Mystic Warriors: Wrath of the Ninjas.

### DIFF
--- a/src/mame/konami/mystwarr.cpp
+++ b/src/mame/konami/mystwarr.cpp
@@ -1,7 +1,7 @@
 // license:BSD-3-Clause
 // copyright-holders:R. Belmont, Phil Stroffolino, Acho A. Tang, Nicola Salmoria
 /**************************************************************************
- * Mystic Warrior (c) 1993 Konami
+ * Mystic Warriors (c) 1993 Konami
  * Metamorphic Force (c) 1993 Konami
  * Violent Storm (c) 1993 Konami
  * Martial Champion (c) 1993 Konami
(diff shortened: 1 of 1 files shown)


mikehyrus:
Oh, out of curiosity should I change the Japanese subtitle to English for the JXX sets or leave it as is?

cuavas:
Also not that if you have the info box set to show screenshots, history.dat or something, as many people do, you also have to switch to the info view (potentially twice – image tab to info tab, then cycle through info sources to general info) before you can see which version you have highlighted if the title overflows the box.

happppp:
Googling around, this subtitle looks correct to me. Maybe not as prominent as eg. "Judgement Day" for the Terminator 2 movie, but yes, official subtitle.

That MAME's search fails on long titles is a longstanding issue with MAME's search algorithm, not the title's fault. The search algo could be made to support chopping up the title string into main title, subtitle, meta data like region of revision. https://mametesters.org/view.php?id=7313

cuavas:
> That MAME's search fails on long titles is a longstanding issue with MAME's search algorithm, not the title's fault. The search algo could be made to support chopping up the title string into main title, subtitle, meta data like region of revision. https://mametesters.org/view.php?id=7313

No-one seems to be able to actually produce an algorithm that doesn’t either perform too poorly for live searching or excessively increase startup time for indexing.

Also, that doesn’t change the fact that excessively long titles produce usability issues because you can’t distinguish between items in a reasonably sized list box.

happppp:
On small displays, yes there's not much room for the game title list.
On PC? (my screenshot), not as bad, you'd need a crazy long title like sfrush to not fit the panel. I think It'd look nicer if all the font size here was same as the text in the left/right panels. Plus you'd get more vertical room.

Screenshot (749)

cuavas:
The UI needs to be usable from TV-viewing distance, though.  It’s a compromise catering to everyone or no-one, but it is what it is.

Still, your screenshot illustrates that is we find more versions of Samurai Shodown 64: Warriors Rage, it would overflow the box even with the wider aspect ratio and smaller font size you’re using because of the inclusion of all the transliterated regional titles in the description.

happppp:
Anyway, I'm fine with having sub-titles in MAME game title strings. If you're suddenly against it, you'll need a lot more people with the same opinion before such a rule gets enforced.
And when removing them from MAME, it's going to be a lot of manual work. eg. the sub-title for sf2 (World Warrior) is not vital, but for SF Rush The Rock, or the Samurai Shodown 64 sequel you mentioned, it is.

happppp:
Well, regarding this PR, do our opinions matter, or what's already the custom in MAME? such as sf2 having the sub-title.
At the same time, people with direct commit access can merrily add them without objection (eg. I added a sub-title to wgp yesterday).

angelosa:
Frankly, "Bubble Bobble: Push Start Button" is a very old meme (iirc Gridle said it, in an old MW post). 

Also we already been there that `GAME` and friends macros don't have enough granularity for alt titles, modifications, optional fields, common branding (e.g. `Deal or no Deal` capitalization) etc. etc. etc.
cfr. #586 and probably other discussions scattered here and there

rb6502:
I side with Vas here.  Nobody in history has ever called this game "Mystic Warriors Colon Wrath Of The Ninjas", given there was never a sequel, and the point of MAME is being able to find and run the software, not show how pedantic you are.  You can subtitle all you want in the comments in the driver, but let's keep the user-interactable names reasonable.

happppp:
I believe the main point of MAME is preservation and documentation, and with that comes accurate naming of software. It's on our site too under "What is MAME?", not just to reject ourselves from being warez software: https://www.mamedev.org/
Taking a driver of yours for example, the same goes for the odd way of naming Apple ][ and Apple ///, that is how Apple called their computers, correct? Not the more user-digestable Apple II or Apple III.

rb6502:
We're talking about even being able to distinguish the different sets for users on the built-in menu.  ][ vs II is a wash on screen space in the menu in most fonts so using the actual typography from the case doesn't hurt anything.  Including the unnecessary subtitle on Mystic Warriors destroys preservation because people using the menu won't be able to read what the sets even are.

ICEknigh7:
In addition to appearing on the title screen, the subtitle here seems to also be included in the trademark itself, so I'd say it's historically relevant enough to have documented as part of the full title: https://furm.com/trademarks/mystic-warriors-wrath-of-the-ninjas-74307816

rb6502:
Feel free to add it to the comment at the top of the file with the game's title.  Then we are documenting it.  As the MAME introduction says, the source code is the documentation.

mikehyrus:
Thanks to FredYeye's WIP the rest of the title is more visible.

happppp:
I'd understand more if the reason was "I don't like it", although this makes it much more subjective and harder to 'enforce'. I don't really care much about this PR/this specific game in particular, I'm more worried that omitting sub-titles will become a 'rule' now because several devs are against it.

Our UI is not so great, and I hope it will improve over time. Knowing MAME, that time can be many years. But still, I don't think the relatively small UI gamelist panel size should take precedence over accurate naming of titles.

The search algo isn't working well. The main issue here is not the sub-titles but the meta data, like (Japan, ver ABC, set 1). There may be better ways to solve it, one I'd have in mind is to split up the strings between title and meta data, and give the 2nd part a much lesser score. While at it, the sub-title could also be made a substring and give a lesser score than the main part of the title.

PR #13478, opened 2025-03-15
snk/ngp.cpp, snk/k1ge.cpp: Updates/Cleanups:

cam900:
Common:
- Fix save state support
- Move screen update function into snk/k1ge.cpp

snk/k1ge.cpp:
- Use indirect palette for color palette
- Implement full screen update function
- Fix typename consistency
- Reduce duplicates
- Fix spacings

snk/ngp.cpp:
- Remove unnecessary screen update function
- Reduce literal tag usages
- Fix namings


src/mame/snk/k1ge.cpp
```diff
@@ -59,77 +61,130 @@ uint8_t k1ge_device::read(offs_t offset)
...
+		set_pen_indirect(offset - 0x100 + MONO_COLOR, MONO_COLOR + 0x18 + data);
+		break;
+	case 0x110: case 0x111: case 0x112: case 0x113:
+		data &= 0x07;
+		set_pen_indirect(offset - 0x100 + MONO_COLOR, MONO_COLOR + 0x20 + data);
+		break;
+	case 0x114: case 0x115: case 0x116: case 0x117:
+		data &= 0x07;
+		set_pen_indirect(offset - 0x100 + MONO_COLOR, MONO_COLOR + 0x28 + data);
+		break;
```
cuavas, line 131: Can you simplify this?  Is the code for all these cases equivalent to:
```c++
		data &= 0x07;
		set_pen_indirect(offset - 0x100 + MONO_COLOR, MONO_COLOR + ((offset & 0x1c) << 1) + data);
```

src/mame/snk/k1ge.cpp
```diff
@@ -59,77 +61,130 @@ uint8_t k1ge_device::read(offs_t offset)
...
+	int offset_x = (scroll_x >> 3) * 2;
 	int px = scroll_x & 0x07;
-	uint16_t map_data;
-	uint16_t hflip;
-	uint16_t pcode;
-	uint16_t tile_addr;
-	uint16_t tile_data;
 
-	base += ( ( ( ( scroll_y + line ) >> 3 ) * 0x0040 ) & 0x7ff );
+	base += ((((scroll_y + line) >> 3) * 0x0040) & 0x7ff);
```
cuavas, line 165: The 0x7ff mask doesn’t really represent the actual values you can get because the left shift means the low six bits will be zero.  This would be clearer if the mask happened before the scaling:
```c++
	base += (((scroll_y + line) >> 3) & 0x1f) * 0x0040;
```

src/mame/snk/k1ge.cpp
```diff
@@ -232,116 +283,108 @@ void k1ge_device::draw_sprite_plane( uint16_t *p, uint16_t priority, int line, i
...
```
cuavas, line 365: Same code again here.

src/mame/snk/k1ge.cpp
```diff
@@ -446,49 +485,43 @@ void k2ge_device::draw_sprite_plane( uint16_t *p, uint16_t priority, int line, i
...
```
cuavas, line 502: And here as well.

cuavas (commented):
This code has a lot of copy/paste.  I still think it could probably be cleaned up by making member function templates for the loops and supplying lambdas to do the actual work that needs to be different for the different models.

Do you think it’s worth doing just the palette cleanup and tidying the code first before trying to reduce the duplication, or would it be easier to do it all at once?

cam900:
> This code has a lot of copy/paste. I still think it could probably be cleaned up by making member function templates for the loops and supplying lambdas to do the actual work that needs to be different for the different models.
> 
> Do you think it’s worth doing just the palette cleanup and tidying the code first before trying to reduce the duplication, or would it be easier to do it all at once?

I'm worked both in code, with reducing code using boolean flags and functions

cuavas:
OK, this is definitely better than before.  I don’t really like having an “is colour” flag member variable, but that can possibly be cleaned up in the future.

PR #13483, opened 2025-03-16
sound/gb.cpp: Updates/Cleanups:

cam900:
- Fix wave RAM corruption function
- Fix wave RAM read/write behaviour for GBA
- Fix wave channel skip and output emulation
- Fix sweep emulation
- Remove outdated comment
- Use single integer for pulse duty table
- Fix master volume
- Reduce duplicates
- Fix typenames for boolean flags
- Fix outputs
- Fix docs, Fix namings

reference: https://gbdev.io/pandocs/


src/devices/sound/gb.cpp
```diff
@@ -411,10 +401,10 @@ void gameboy_sound_device::update_square_channel(SOUND &snd, uint64_t cycles)
 		{
 			cycles -= distance;
 			distance = 0x800 - snd.frequency;
-			uint64_t counter = 1 + cycles / distance;
+			const uint64_t counter = 1 + cycles / distance;
 
 			snd.duty_count = (snd.duty_count + counter) & 0x07;
-			snd.signal = wave_duty_table[snd.duty][snd.duty_count];
+			snd.signal = BIT(wave_duty_table[snd.duty], snd.duty_count);
```
cuavas, line 407: This changes the output for these channels from balanced -1/+1 to unbalanced 0/+1 levels.  Is that correct?  Can you provide a reference for this change in behaviour?

cam900: That is digital sample, and converted at DAC output.

src/devices/sound/gb.cpp
```diff
@@ -1323,50 +1317,49 @@ void gameboy_sound_device::sound_stream_update(sound_stream &stream, std::vector
...
 		}
 
 		/* Adjust for master volume */
-		left *= m_snd_control.vol_left;
-		right *= m_snd_control.vol_right;
+		left *= 1 + m_snd_control.vol_left;
+		right *= 1 + m_snd_control.vol_right;
 
 		/* Update the buffers */
 		outputl.put_int(sampindex, left, 32768 / 64);
```
cuavas, line 1365: I’m not sure this will be giving the correct scaled values now.

* The range of the native channel outputs is 0x0..0x0f, corresponding to normalised 1..-1.
* `convert_output` uses the formula `0xf - (2 * val)` which will give values in the range 15..-15.
* There are four channels summed together, giving a range of 60..-60.
* This is multiplied by the sound volume control values of 1..8, giving a range of 480..-480.
* You’re scaling against `32768 / 64`, i.e. 512, so you won’t get the full output range.  Wouldn’t it be better to use `outputl.put_int(sampindex, left, 15 * 4 * (1 + 7));` or something so the maximum output values will map to the full output range?

PR #13486, opened 2025-03-17
mb86233: write the results of an ALU operation before writing to register/memory, and only test the d register after ALU operations

gm-matthew:
Sega Rally currently has a bug where the cars will glitch into certain parts of the track and fly into the air:

https://youtu.be/-QdM_98HiYA?t=170

I've discovered that the cause of this bug is a particular TGP instruction in a subroutine that calculates square roots, which tries to add the result of a multiplication to the D accumulator register at the same time as loading a value into the same register:

`01B3: 1D9DB24E fsmd : mov $0x4e, d`

Before the TGP emulation rewrite in MAME 0.197, the ALU operation would be performed first and the result in the D register would be overwritten by the MOV operation. Since the rewrite the ALU operation is split into two parts: the calculation is performed first and the result stored in an intermediate register, and the D register is updated at the end overwriting the value loaded by the MOV operation.

With the ALU operation result overwritten by the MOV operation the maximum error of the square root result is 0.17%, while vice versa the maximum error is a whopping 21.1%, which is enough to cause the above bug.

With this edit I've moved the alu_post() call to before write_mem_internal_1(), write_mem_io_1() and write_reg() when performing a MOV operation; I haven't bothered for LAB operations since these never affect the D register. I've also removed testdz() since it seems that loading a value into the D register does not overwrite the flags set by an ALU operation.

The result is that the bug no longer occurs, however the opponent cars are now slower than they are on real hardware (similar to MAME 0.196 and earlier). Their increased speed after the TGP rewrite was a side effect of this emulation bug; the game commonly calls a TGP function that calculates the distance between two vectors, and the result was inflated by the incorrect square root calculation; larger values lead to the opponent cars moving faster.

I suspect the real reason why the opponent cars move faster on real hardware is because the fdvd operation is only an approximation (possibly using an internal lookup table?), and that it tends to approximate slightly too high. Fudging the calculation by multiplying the result of a divide by 1.01 leads to the cars being much closer to real hardware, though obviously this would not be an acceptable solution for MAME. To actually make the speed of opponent cars match that of real hardware, we would need to figure out exactly how the TGP performs the fdvd operation.

I wonder if it would be a good idea to merge the alu_pre() and alu_post() functions into a single alu() function which is called after reading from registers/memory (and also checking that we don't need to stall) and before writing to registers/memory; there is no longer any real need to separate the two functions.

### DIFF
--- a/src/devices/cpu/mb86233/mb86233.cpp
+++ b/src/devices/cpu/mb86233/mb86233.cpp
@@ -256,18 +256,6 @@ void mb86233_device::pcs_pop()
 		m_pcs[i] = m_pcs[i+1];
 }
 
-void mb86233_device::testdz()
-{
-	if(m_d)
-		m_st &= ~F_ZRD;
-	else
-		m_st |= F_ZRD;
-	if(m_d & 0x80000000)
-		m_st |= F_SGD;
-	else
-		m_st &= ~F_SGD;
-}
-
 void mb86233_device::stset_set_sz_int(u32 val)
 {
 	m_alu_stset = val ? (val & 0x80000000 ? F_SGD : 0) : F_ZRD;
@@ -324,7 +312,7 @@ void mb86233_device::alu_pre(u32 alu)
 	}
 
 	case 0x06: {
-		// fmad
+		// fadd
 		m_alu_stmask = F_ZRD|F_SGD|F_CPD|F_OVD|F_DVZD;
 		m_alu_r1 = f2u(u2f(m_d) + u2f(m_a));
 		stset_set_sz_fp(m_alu_r1);
@@ -676,9 +664,9 @@ void mb86233_device::write_reg(u32 r, u32 v)
 	case 0x14: m_b = set_exp(m_b, v); break;
 	case 0x15: m_b = set_mant(m_b, v); break;
 		/* c */
-	case 0x19: m_d = v; testdz(); break;
-	case 0x1a: m_d = set_exp(m_d, v); testdz(); break;
-	case 0x1b: m_d = set_mant(m_d, v); testdz(); break;
+	case 0x19: m_d = v; break;
+	case 0x1a: m_d = set_exp(m_d, v); break;
+	case 0x1b: m_d = set_mant(m_d, v); break;
 	case 0x1c: m_p = v; break;
 	case 0x1d: m_p = set_exp(m_p, v); break;
 	case 0x1e: m_p = set_mant(m_p, v); break;
@@ -812,6 +800,7 @@ void mb86233_device::execute_run()
 				u32 v = m_data.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_io_1(r2, v);
 				break;
 			}
@@ -822,6 +811,7 @@ void mb86233_device::execute_run()
 				u32 v = m_data.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_io_1(r2, v);
 				break;
 			}
@@ -832,6 +822,7 @@ void mb86233_device::execute_run()
 				u32 v = m_io.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_internal_1(r2, v, false);
 				break;
 			}
@@ -842,6 +833,7 @@ void mb86233_device::execute_run()
 				u32 v = m_data.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_internal_1(r2, v, true);
 				break;
 			}
@@ -852,6 +844,7 @@ void mb86233_device::execute_run()
 				u32 v = m_data.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_internal_1(r2, v, false);
 				break;
 			}
@@ -862,6 +855,7 @@ void mb86233_device::execute_run()
 				u32 v = m_program.read_dword(ea);
 				if(m_stall) goto do_stall;
 				ea_post_0(r1);
+				alu_post(alu);
 				write_mem_internal_1(r2, v, false);
 				break;
 			}
@@ -872,6 +866,7 @@ void mb86233_device::execute_run()
 					// mov reg, mem
 					u32 v = read_reg(r2);
 					if(m_stall) goto do_stall;
+					alu_post(alu);
 					write_mem_internal_1(r1, v, false);
 					break;
 				}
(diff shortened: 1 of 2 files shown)


angelosa:
I presume this fixes [vf2](https://mametesters.org/view.php?id=8676) and [motoraid](https://mametesters.org/view.php?id=7264) , latter been incorrectly put as "invalid report" (game is NOT marked MNW)

cuavas:
@galibert can you please take a look at this?

TheLostNinja:
> I presume this also fixes [vf2](https://mametesters.org/view.php?id=8676) and [motoraid](https://mametesters.org/view.php?id=7264) , latter been incorrectly put as "invalid report" (game is NOT marked MNW)

Manx TT Laxey Coast map crash after one of mid-track checkpoints is fixed too.

angelosa:
Yeah, vaguely remember that was a thing. For motoraid iirc you either submerge or be skyrocketed in stage ... 3?

TheLostNinja:
> I presume this also fixes [vf2](https://mametesters.org/view.php?id=8676) and [motoraid](https://mametesters.org/view.php?id=7264) , latter been incorrectly put as "invalid report" (game is NOT marked MNW)

The broken fighter's avatar in VF2 still present.

angelosa:
> The broken fighter's avatar in VF2 still present.

Yeah, shouldn't be fixed by this PR (it's scissor test + focal distance afaik). What should is the hitbox left-right bogus crossups, as per steps to reproduce in the attached report.

gm-matthew:
> Manx TT Laxey Coast map crash after one of mid-track checkpoints is fixed too.

Makes sense, the game uses almost exactly the same copro code as Sega Rally just with a few extra functions added to the end.

> The broken fighter's avatar in VF2 still present.

This is because the emulated renderer adds 1.0 to the z-value of every polygon, which is not correct for 2/2A games at least. In any case this is not a TGP bug.

Gurudumps:
This isn't widely known, but certain Model 2a/2b games will not run on certain boards because Sega updated some of the TGP chips to a later rev and as far as I know those were not decapped. I have access to over 100 dead 2a/2b boards if someone can decap and extract the internal datas. Manx TT and Motoraid are at least some of the games that use the later TGPs, hence why there are TGP bugs in the emulation.

gm-matthew:
> This isn't widely known, but certain Model 2a/2b games will not run on certain boards because Sega updated some of the TGP chips to a later rev and as far as I know those were not decapped. I have access to over 100 dead 2a/2b boards if someone can decap and extract the internal datas. Manx TT and Motoraid are at least some of the games that use the later TGPs, hence why there are TGP bugs in the emulation.

AFAIK all Model 2/2A boards use the 315-5677 as the copro DSP; it was the geometry DSPs (315-5678, 315-5679) that were updated with A and/or B revisions, probably with modifications to the internal program ROMs.

Model 2B uses the ADSP-21062 (SHARC) instead of the TGP which may of course have emulation bugs of its own.

Gurudumps:
I own boards I know what's on them. There is a rev C.......

TheLostNinja:
Also physichs of shot enemies in Virtual Cop 2 became a more adequate, but after collusion of 2 patrol cars with Virtual Coffee van hijacked by bandits in the middle of File 1 Scene 2, road still turns out to be fully free (Unlike real hardware captures. Here is the reference videos: [1](https://www.youtube.com/watch?v=0HUF-ENKrtI), [2](https://www.youtube.com/watch?v=ZNoDZsjEWCI).).

TheLostNinja:
In Wing War, the aircrafts became invincible for machine guns and collisions with enviorenmental elements.

angelosa:
> In Wing War, the aircrafts became invincible for machine guns and collisions with enviorenmental elements.

Regression with this PR?

TheLostNinja:
> > In Wing War, the aircrafts became invincible for machine guns and collisions with enviorenmental elements.
> 
> Regression with this PR?

Looks like this problem [is older](https://mametesters.org/view.php?id=7563) - althought I found it only now, during the cursory test of various games using TGP.

gm-matthew:
> You quoted before I edited it. Of course 2B doesn't apply here. I own boards I know what's on them. You totally missed the point. There is a rev C chip on 2A (315-5679C). More importantly it's used on Manx TT and Motoraid.....

You are claiming that the 315-5679C revision may be responsible for some of the TGP emulation issues for Manx TT and Motor Raid in MAME, but this is impossible since 315-5679C is a geometry DSP and MAME only emulates the copro TGP currently.

Do you have any Model 2/2A boards with a later revision of the copro DSP (315-5677)?

Gurudumps:
I don't care what chips MAME emulates nor do I care what causes the emulation issues (that's not my problem).
I'm not 'claiming' anything. I'm telling you (from a hardware point-of-view) that Manx TT and Motoraid ***REQUIRES*** the board with the C rev chips. 
This same board has 2x 315-5678A and 2x 315-5679C and 315-5677A.

gm-matthew:
> I don't care what chips MAME emulates nor do I care what causes the emulation issues (that's not my problem). I'm not 'claiming' anything. I'm telling you (from a hardware point-of-view) that Manx TT and Motoraid _**REQUIRES**_ the board with the C rev chips. This same board has 2x 315-5678A and 2x 315-5679C and 315-5677A.

I had presumed that there was no updated revision of 315-5677 because [this page](https://gurudumps.otenko.com/m2status/index.html) on your site didn't mention one, but if you're saying that 315-5677A does indeed exist then I stand corrected.

According to the [Manx TT manual](https://segaretro.org/images/0/00/ManxTT_Model2_Manual_Twin.pdf) the CPU board has part number 837-10848-01-91, so I presume this must be the one with 315-5677A, 315-5678A and 315-5679C. Interestingly the [Sega Rally manual](https://segaretro.org/images/9/9e/SegaRally_Model2_Manual_Twin.pdf) says that the CPU board may be either 837-10848-91 or 837-10848-01-91, so I'm guessing perhaps Sega Rally can use either one but Manx TT cannot? I don't know, I'm just looking to improve emulation of Model 2 games in MAME :-)

Gurudumps:
You can't go by numbers in the manual. Sega uses a different number for every game, also shown on a sticker on each board. The -91 etc is on a sticker.
No changes to the PCB were ever done. All 2A boards are "837-10848-01 MODEL 2 A-CRX CPU BD"
The chips are the very same pin-compatible chips but with different internal ROM so board changes were not required.

angelosa:
Gonna use this rather than opening an issue/discussion for convenience:
If this fixes `vf2` hurt boxes then why the same fix doesn't work for the original `vf`? Like you right-left punch with Akira (the most simple reproduction, there's a lot more involved), only the first punch connects?
What's the catch here?

gm-matthew:
> Gonna use this rather than opening an issue/discussion for convenience: If this fixes `vf2` hurt boxes then why the same fix doesn't work for the original `vf`? Like you right-left punch with Akira (the most simple reproduction, there's a lot more involved), only the first punch connects? What's the catch here?

The copro TGP ROM for `vf` is a bad dump so it could potentially be the reason why hit boxes aren't working properly.

angelosa:
Right, iirc the Virtua Racing ROM adapted to Virtua Fighter now that you mentioned it.

PR #13488, opened 2025-03-17
sound/s_dsp.cpp: Updates/Cleanups

cam900:
- Fix pitch modulation emulation
- Fix input clock
- Fix save state support
- Fix indent inconsistency
- Use std::clamp for clamping functions
- Use reference for voice state
- Use logmacro.h for logging

reference: https://snes.nesdev.org/wiki/SNESdev_Wiki

### DIFF
--- a/src/devices/sound/s_dsp.cpp
+++ b/src/devices/sound/s_dsp.cpp
@@ -30,6 +30,8 @@
 #include "emu.h"
 #include "s_dsp.h"
 
+#include <algorithm>
+
 /***************************************************************************
  CONSTANTS AND MACROS
 ***************************************************************************/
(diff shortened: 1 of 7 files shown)


src/devices/sound/s_dsp.cpp
```diff
@@ -696,7 +696,7 @@ void s_dsp_device::dsp_update( s16 *sound_ptr )
 			sound_ptr++;
 
 			if (outr > 32767)
-					*sound_ptr = 32767;
+				*sound_ptr = 32767;
```
angelosa, line 699: This core should probably use `put_int_clamp` / `std::clamp` against `s16`, everywhere, while not using `int` internal storages.
(`put_int` checks only against `max` value)

PR #13505, opened 2025-03-21
3rdparty/zlib: Updated to 1.3.1.

holub:


ajrhacker:
This update seems fairly innocuous, and the 0.276 release branch was forked just recently, so I think it's fine to merge it now.

PR #13506, opened 2025-03-21
v60: implement correct rounding mode when converting float to integer

gm-matthew:
Source: https://multimedia.cx/NEC_V60pgmRef.pdf, page 53

Hopefully this eliminates invisible walls in Virtua Racing


happppp:
Can be reduced to switch (TKCW & 7) and default: trunc()?

PR #13507, opened 2025-03-21
i960: don't preserve sign bit when shifting integer left

gm-matthew:
Commit 4cce14e changed the behavior of the shli instruction to preserve the sign bit, but I don't know of any other CPUs that do this nor is there any evidence in the [programmers reference manual](http://bitsavers.informatik.uni-stuttgart.de/components/intel/i960/80960KB_Programmers_Reference_Manual_Mar88.pdf) to suggest the i960 actually does this.

The i960 manual says that shifting left is supposed to be equivalent to multiplying by a power of 2, but if you multiply the signed integer 2,000,000,000 by 2 it overflows to negative, so the sign bit is not being preserved in this case.

If there is actual evidence that the i960 preserves the sign bit when shifting integers left I would be interested to see it, but until then I can only assume that commit 4cce14e is a regression; the opponent cars in Daytona USA have weird orientation issues that did not occur before this commit.

### DIFF
--- a/src/devices/cpu/i960/i960.cpp
+++ b/src/devices/cpu/i960/i960.cpp
@@ -1157,7 +1157,7 @@ void i960_cpu_device::execute_op(uint32_t opcode)
 				m_icount--;
 				t1 = get_1_ri(opcode);
 				t2 = get_2_ri(opcode);
-				set_ri(opcode, t1 >= 32 ? 0 : ((int32_t)t2) << t1);	// sign is not preserved
+				set_ri(opcode, t1 >= 32 ? 0 : ((int32_t)t2) << t1);
 				break;
 
 			default:
--- a/src/devices/cpu/i960/i960.cpp
+++ b/src/devices/cpu/i960/i960.cpp
@@ -1157,7 +1157,7 @@ void i960_cpu_device::execute_op(uint32_t opcode)
 				m_icount--;
 				t1 = get_1_ri(opcode);
 				t2 = get_2_ri(opcode);
-				set_ri(opcode, (t2 & 0x80000000) | (t1 >= 32 ? 0 : (t2<<t1) & 0x7fffffff)); // sign is preserved
+				set_ri(opcode, t1 >= 32 ? 0 : ((int32_t)t2) << t1);	// sign is not preserved
 				break;
 
 			default:


happppp:
AFAIK shifting negative numbers (from the int32_t cast) is undefined in C++.

ajrhacker:
"The shli instruction shifts zeros in from the least-significant bit; if the bits shifted out are not the same as the sign bit, an overflow fault is generated. If overflow occurs, *the sign of the result is the same as the sign of the src operand*." (Intel 80960MC Programmer's Reference Manual [1988], p. 17-131; emphasis added.)

happppp:
Ok, feel free to revert.

ajrhacker:
> I don't know of any other CPUs that do this

Many historic CPU architectures included "arithmetic" left shifts that preserve the sign even in case of overflow. Preserving the sign during left shifts is an obvious necessity with sign-magnitude arithmetic, but it was fairly common even on architectures that used two's complement arithmetic. These architectures generally provided overflow flags but not carry flags for ordinary arithmetic operations, which meant that double-precision arithmetic was limited to 2N-1 bits rather than 2N bits.

gm-matthew:
Unfortunately implementing the above behavior did not fix the orientation issues with the opponent cars in Daytona USA, so I will not be submitting it as a pull request.

Here is the bit of code where the integer overflow takes place:

`00011688: 59B58183          subi    r3,g6,g6`
**`0001168C: 59B58F10          shli    16,g6,g6`**
`00011690: 59B58D90          shri    16,g6,g6`
`00011694: 74B58588          divi    r8,g6,g6`
`00011698: 5918C096          addi    g6,r3,r3`

The integer overflow mask is set in the arithmetic control register so a fault does not occur, nor is the integer overflow flag checked.

Oddly enough there are three other instances in this subroutine of a register being shifted 16 bits left only to be immediately shifted back to the right again, except that the **shlo** instruction is used to shift left. Here's one example:

`00011630: 59290183          subi    r3,r4,r5`
**`00011634: 59294E10          shlo    16,r5,r5`**
`00011638: 59294D90          shri    16,r5,r5`
`0001163C: 74314588          divi    r8,r5,r6`
`00011640: 5938C086          addi    r6,r3,r7`

The programmers are shifting the register left by 16 bits only to immediately shift it back right 16 bits again, effectively converting the value to a 16-bit signed integer.  It looks like the programmers meant to use **shlo** for the instruction at 1168C but accidentally used **shli** instead, and yet the car orientation bug does not occur on a real Model 2 board.

The only explanation I can come up with is that the special integer overflow behavior somehow does not apply to the 80960KB specifically; the 80960KB manual does not mention the sign bit being preserved in the case of integer overflow at all.

PR #13519, opened 2025-03-25
x68k_flop.xml: added/replaced 17 dumps from original disks

rzero9:
A new update for the X68000 software list. I finally figured out how to properly convert those Konami disks with the "track full of overlapped sectors" protection (turns out the track had to be aligned correctly with the index signal), so the Namachuukei 68 dump that rockleevk did a long time ago now works and can be added to the list.

I also included the new disks that krugman has been dumping lately, all of them are working as far as I have tested.

As usual, I have compared everything with the existing images. Protected disks that already had a cracked version have been added preserving the existing ones, and disks where the decoded data matches the existing image have been replaced, since the MFM format represents the disk more accurately at a lower level.

Domino Soldier Albion [krugman]
Guerriere Lyeward [krugman]
Hanafuda Tengoku [krugman]
Horror of Cridewell (Revised Edition) [krugman]
Hyper UD (v1.00) [krugman]
Jigen no Tabi [krugman]
Kindan no Paradise [krugman]
King's Dungeon [krugman]
Magic Knight [krugman]
Momotarou II Ohimesama ga Ippai (v1.11) [krugman]
Namachuukei 68 [rockleevk]
Rouge - Manatsu no Kuchibeni [krugman]
Sadistic Gamers Part-4 - Rumi no Crime Play [krugman]

Replaced software list items
----------------------------
Baruusa no Fukushuu [krugman]
Jane [krugman]
Nicoll [krugman]
Quintia Road [krugman]

### DIFF
--- a/hash/x68k_flop.xml
+++ b/hash/x68k_flop.xml
@@ -301,10 +301,6 @@ MicroProse
 ==========
 F15 Strike Eagle II Sabaku no Arashi Sakusen Scenario Disk
 
-NIC
-===
-Magic Knight
-
 New
 ===
 Princess Maker
@@ -359,7 +355,6 @@ QUIZ Ikasete...
 
 Soft Plan
 =========
-King's Dungeon
 Dark Odyssey
 
 Techno Brain
@@ -380,7 +375,6 @@ Wrestle Angels 2 Top Eventer
 Zenryutsu
 =========
 Endan Rekishi Emaki Nukata no Ookimi
-Kindan no Paradise
 Sailor fuku Ikenai Taiken Kokuhaku Shuu Vol.1 Lost Virgin
 Sailor fuku Ikenai Taiken Kokuhaku Shuu Vol.2 Temptation
 Sailor fuku Ikenai Taiken Kokuhaku Shuu Vol.3 Ecstasy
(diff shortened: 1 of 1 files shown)


DopefishJustin:
I don't know if I like the idea of replacing everything with .mfm even for standard disks, it makes the data harder to examine and work with in other software for basically no benefit?

rzero9:
I don't feel super strongly about it either way (mostly because i hate the whole "format wars" thing with a passion), but I can think of at least one benefit: having the full MFM data in there allows you to verify that the image has been dumped from an undamaged disk by checking the header/data CRC fields. There is still a 1 in 65536 chance of the CRC matching by pure coincidence, of course, but it's better than nothing.

Also, there is always the possibility of software with subtle protection schemes that depend on things like non-standard gap data. I don't think I've found anything like that yet, but if they exist, they should look exactly like standard disks at first glance.

Now, whether those benefits outweigh the benefit of having the decoded data directly available and usable in other emulators (and the question of "should MAME care about what other emulators want?")... that's a matter of opinion. If MAMEdev collectively decides this is not the way to go, I don't mind reverting those changes, but I think there should be some sort of agreed-upon standard (even if it's a loose one) so we can avoid these discussions.

Lord-Nightmare:
can the MFM format properly represent some of the wackier 'overlapping sectors' type (upd765 compatible) protections properly? or would we be better off with a flux representation like MFI for those?

rzero9:
It can, in fact some of the disks added in this PR have overlapping sectors. AFAIK the only thing it can't do is protections that rely on "weak bits", but I don't know if there is a supported alternative format that can handle that. I have tried converting some of those to MFI with Applesauce (which is the only thing I know of that can convert flux dumps to MFI) but they never worked.

DopefishJustin:
HxCFloppyEmulator has added MFI output as of v2.16.12.2, I dunno if it supports weak bits though.

ICEknigh7:
I think the benefit here is, since this format holds more information, it can be converted to any others while the others can't be converted back to an MFM dump without loss.

PR #13525, opened 2025-03-27
Add a skeleton for an Oldsmobile Driver Information System

devinacker:
Random eBay salvage that I used to test 87AD internal ROM dumping (I don't have the actual unit, just the MCU).

New NOT_WORKING machine
----------
Oldsmobile Driver Information System (version 2.2 CH) [Devin Acker]


devinacker:
After some sleuthing (thanks @Sneakernets) this appears to be from an Oldsmobile "Driver Information System". I'll update the patch accordingly later.

PR #13527, opened 2025-03-28
src/mame/sinclair/spec128.cpp: Memory access specifics moved to base driver; Cleaned up comments style

holub:


src/mame/sinclair/spec128.cpp
```diff
@@ -221,18 +221,18 @@ void spectrum_128_state::spectrum_128_port_7ffd_w(offs_t offset, uint8_t data)
...
 	content_early(1);
 
-	/* D0-D2: RAM page located at 0x0c000-0x0ffff */
-	/* D3 - Screen select (screen 0 in ram page 5, screen 1 in ram page 7 */
-	/* D4 - ROM select - which rom paged into 0x0000-0x03fff */
-	/* D5 - Disable paging */
+	// D0-D2: RAM page located at 0x0c000-0x0ffff
+	// D3 - Screen select (screen 0 in ram page 5, screen 1 in ram page 7
+	// D4 - ROM select - which rom paged into 0x0000-0x03fff
+	// D5 - Disable paging
```
angelosa, line 227: I feel like comments about what a r/w handler does should be before the function body not inside it (i.e. this block moved right before `void spectrum_128_port_7ffd_w`).

This is more of a guideline than a request obviously.

PR #13529, opened 2025-03-29
adp/stella8085.cpp: add 11 nonworking romsets

stonedDiscord:
a WIP schematic for 4087-000-101 can be found at
https://github.com/stonedDiscord/weisse_cpu
The release dates are largely taken from
https://www.goldserie.de/index.php/galerien/3217-excellent

### DIFF
--- a/src/mame/adp/stella8085.cpp
+++ b/src/mame/adp/stella8085.cpp
@@ -3,8 +3,8 @@
 
 /*
 
-Stella 8085 based hardware
-Electromechanical fruit machines
+Elektronische Steuereinheit
+8085 based hardware
 Lots of lamps and 4 7-segment LEDs
 
 Main components:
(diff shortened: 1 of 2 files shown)


Osso13:
rn_image_picker_lib_temp_9b7425ad-d42a-4fa5-a4c5-59b2af6e062e
PCB pic for the set for which I created the skeleton clearly shows the RTC. The XTAL was verified 10.240, so please don't remove those things. If you are sure the other sets don't have the RTC and have the different XTAL value, please create a different machine config. If the sets are actually on a significantly different kind of PCB, please create a different driver.

stonedDiscord:
Those are the same PCBs, you can have a look at the board view in your browser
https://kicanvas.org/?github=https%3A%2F%2Fgithub.com%2FstonedDiscord%2Fweisse_cpu
What board revision is this, it's unreadable in the picture
image
How did you verify the crystal, it looks bent the wrong way in the picture.
The frequency is marked as 6.144MHz on the 4087-000-101 I have.

I will add a machine config in this case, the RTCs seem different enough.

Hammy1986:
Here's some pics of dice master... Thanks to @ Trol 

rn_image_picker_lib_temp_172a57b
rn_image_picker_lib_temp_5fae369
rn_image_picker_lib_temp_4055ed9

stonedDiscord:
This is why I assumed the crystal to be 6MHz.



These are some of the most common game machines in Germany around 1990.

ajrhacker:
All program ROMs for excellent appear to be dumped at half their proper size.

stonedDiscord:
image

HN462732 is ...... 4kb each.
Sadly I don't have access to that board anymore, I'll try to get a redump.

ajrhacker:
The ROMs as received were 2,048 bytes each. A mistake has occurred somewhere.

cuavas:
This also doesn’t have credits for additions.  I’m running out of time in the month very fast.  I’m just going to comment out things that don’t line up.

cuavas:
Also, are you using parent/clone relationships correctly?  Are these actually different versions of the same games, or, as the comment suggests, are you making all the games on similar boards clones?

stonedDiscord:
They are all different games on the same boards. Is this not how the parent/clone relation works?
The wiki says
```Note that parent/clone relationships mainly exist to group different versions of the same system and are somewhat arbitrary.```
In the end all of these are gambling machines that mainly change in name and decoration.
As for the credits, I dumped these myself.

rb6502:
Parent/clone is for different revisions of the same game that share ROM data.  Games that are different games and only share a boot ROM or something should not be clones.

stonedDiscord:
in that case they are all parents. my mistake.

PR #13530, opened 2025-03-30
cpu/upd7725.cpp: Fix naming for host interface function, Suppress side effects for debugger reads

cam900:
NEC UPD7725/UPD96050 DSPs are not just for SNES cartridges, but also used at SSV arcade system and Prose 2000/2020 card.


Lord-Nightmare:
I would suggest instead splitting the 'host_r' and 'host_w' functions into 'data_r' and 'data_w' plus 'status_r' (there isn't a 'status_w' on the real device except for some unimportant test mode stuff). (I already have a local change to do this, as part of a much larger retooling of the upd7725)

cam900:
> I would suggest instead splitting the 'host_r' and 'host_w' functions into 'data_r' and 'data_w' plus 'status_r' (there isn't a 'status_w' on the real device except for some unimportant test mode stuff). (I already have a local change to do this, as part of a much larger retooling of the upd7725)

Done since https://github.com/mamedev/mame/pull/13530/commits/1d45fa74139ea37e308ae520629a0338c6d593fd .

src/devices/bus/snes/event.cpp
```diff
@@ -152,7 +155,8 @@ void sns_pfest94_device::chip_write(offs_t offset, uint8_t data)
 	{
 		// DSP access
 		offset &= 0x1fff;
-		m_upd7725->snesdsp_write(offset < 0x1000, data);
+		if (BIT(~offset, 12))
+			m_upd7725->data_w(data);
```
cuavas, line 159: Will bit 12 being clear actually prevent the write from happening, or will the chip still be selected and treat it as a data write in both cases?

Also, you can get rid of the `offset &= 0x1fff` in these functions since you’re only testing one bit anyway.

src/devices/bus/snes/upd.cpp
```diff
@@ -142,14 +142,18 @@ void sns_rom20_necdsp_device::device_add_mconfig(machine_config &config)
...
 }
 
 
 void sns_rom20_necdsp_device::chip_write(offs_t offset, uint8_t data)
 {
 	offset &= 0x7fff;
-	m_upd7725->snesdsp_write(offset < 0x4000, data);
+	if (BIT(~offset, 14))
+		m_upd7725->data_w(data);
 }
```
cuavas, line 157: Same here, you can get rid of the `offset &= 0x7fff`, and I’d like someone to double-check whether bit 14 really does stop it from writing.

src/devices/bus/snes/event.cpp
cam900: Offset bit 12 is data/status mode select at both read/write accessing, and status write is ignored at write case.

src/mame/seta/ssv.cpp
```diff
@@ -438,8 +406,8 @@ void drifto94_state::drifto94_map(address_map &map)
 	ssv_map(map, 0xc00000);
 //  map(0x210002, 0x210003).nopw();                                      // ? 1 at the start
 	map(0x400000, 0x47ffff).nopw();                                       // ?
-	map(0x480000, 0x480001).rw(FUNC(drifto94_state::dsp_dr_r), FUNC(drifto94_state::dsp_dr_w));
-	map(0x482000, 0x482fff).rw(FUNC(drifto94_state::dsp_r), FUNC(drifto94_state::dsp_w));
+	map(0x480000, 0x480001).rw(m_dsp, FUNC(upd96050_device::data_r), FUNC(upd96050_device::data_w));
+	map(0x482000, 0x482fff).rw(FUNC(drifto94_state::dsp_r), FUNC(drifto94_state::dsp_w)).umask16(0x00ff);
```
cuavas, line 410: This doesn’t seem quite right – you’re installing a 16-bit handler with a unit mask that has 8 bits set.  Shouldn’t it be an 8-bit handler?

PR #13531, opened 2025-03-30
machine/rc, paia/fatman: Generalizing RC circuit device.

m1macrophage:
Generalized and moved the RC circuit device from `paia/fatman.cpp` to `machines/rc.cpp`.

This will be needed in `moog/source.cpp`. It will also be the core building block for the envelope generators in the LinnDrum and DMX.

Filename and class name suggestions welcome. I originally planned to add this in` rescap.h/.cpp` (the device is literally a resistor and a capacitor, after all :)), but `rescap.h` is included in a lot of places. Maybe the new device can move there after it matures? Open to suggestions, of course.

### DIFF
--- a/scripts/src/machine.lua
+++ b/scripts/src/machine.lua
@@ -3012,6 +3012,18 @@ end
 
 ---------------------------------------------------
 --
+--@src/devices/machine/rescap.h,MACHINES["RC"] = true
+---------------------------------------------------
+
+if (MACHINES["RC"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/machine/rc.cpp",
+		MAME_DIR .. "src/devices/machine/rc.h",
+	}
+end
+
+---------------------------------------------------
+--
 --@src/devices/machine/rf5c296.h,MACHINES["RF5C296"] = true
 ---------------------------------------------------
 
--- a/src/devices/machine/rc.cpp
+++ b/src/devices/machine/rc.cpp
@@ -1,0 +1,85 @@
+// license:BSD-3-Clause
+// copyright-holders:m1macrophage
+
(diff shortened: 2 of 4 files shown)


galibert:
I'd like to start regrouping virtual-analog related classes by having them have a VA_ prefix on them, if you don't mind.  I hope to do some VA_VCO, VA_VCF and friends in the hopefully not too far future.

m1macrophage:
Note that this device is more generally applicable, and not only relevant to audio applications.

For context, I am also working on the streaming version of this class (a `device_sound_interface` in `devices/sound`), to use as an Envelope Generator for the DMX and LinnDrum. That one will uses this RC_CIRCUIT class at its core. Based on the guidance here, that one should be named something like VA_RC_EG.

Given the context above, should I proceed with the rename? Happy to do it if that's preferred. Or if you prefer that I augment this PR when the changes above are done, so that the bigger picture is more evident, that's fine as well. Let me know.

galibert:
Honestly I don't really have a clear picture outside of that we have to create the building blocks to emulate the sound on all these analog synths we're supporting.  And it will be easier for all of us if they are readily identifiable.

Note that we should keep performance in mind.  I'm not sure we want to blow up an ADSR into a bunch of devices, that adds a lot of calls and copying around that will end up detrimental at some point.  What do you think?

m1macrophage:
Yes, makes sense. I will integrate the RC circuit logic into the envelope generator. It makes things simpler too.

Marking this PR as draft for now.

m1macrophage:
Superseded by #13545. That PR incorporates the naming and design suggestions above, and includes the VCA implementation to showcase how the EG and VCA can be combined. Closing this PR.

PR #13532, opened 2025-03-30
add io map for soundcpu huangyeh_m68k.cpp

flama12333:
should add io maps?
since I see in other driver that uses io map 
att.cpp and thepit.cpp


happppp:
Only if they're used by the emulated game.

angelosa:
SW don't access I/O space.

PR #13534, opened 2025-03-30
Layout for tek4404

Elektraglide:
Has 4 status LEDs (inside display on 4404) as well as HD activity LED

### DIFF
--- a/src/mame/layout/tek4404.lay
+++ b/src/mame/layout/tek4404.lay
@@ -1,0 +1,52 @@
+<?xml version="1.0"?>
+<!--
+license:CC0-1.0
+authors: Elektraglide
+
+Tektronix 4404 Artificial Intelligence System.
+-->
+<mamelayout version="2">
+	<element name="static_black"><rect><color red="0.0" green="0.0" blue="0.0" /></rect></element>
+	<element name="led_round" defstate="1">
+		<disk state="1">
+			<color red="0.2" green="0.0" blue="0.0" />
+		</disk>
+		<disk state="0">
+			<color red="1.0" green="0.0" blue="0.0" />
+		</disk>
+	</element>
+	<element name="led_rect" defstate="1">
+		<rect state="1">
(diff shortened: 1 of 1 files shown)


src/mame/layout/tek4404.lay
```diff
@@ -0,0 +1,52 @@
...
+	</element>
+	<view name="Graphics">
+		<bounds x="0" y="0" width="640" height="496" />
+		
+		<screen index="0">
+			<bounds x="0" y="0" width="640" height="480" />
+		</screen>
+		<element name="spacer" ref="static_black">
+			<bounds x="0" y="480" width="640" height="16" />
+		</element>
```
cuavas, line 34: You don’t really need this thing as you’ve got explicit bounds for the view, and the LEDs will draw on top of the black background anyway.  But if you insist on using it, you’re better off putting it behind the screen rather than on top to ensure it doesn’t slightly crop the bottom of the screen when the floating point view coordinates are quantised to whole pixel values.

src/mame/layout/tek4404.lay
```diff
@@ -0,0 +1,52 @@
...
+		<element name="led_4" ref="led_round">
+			<bounds x="22" y="484" width="8" height="8" />
+		</element>
+		<element name="led_8" ref="led_round">
+			<bounds x="10" y="484" width="8" height="8" />
+		</element>
+		<element name="led_disk" ref="led_rect">
+			<bounds x="614" y="485" width="16" height="6" />
+		</element>
```
cuavas, line 50: Would it be a good idea to add text labels for the LEDs?  As it is, there are three identical-looking round LEDs, the function of which won’t be clear without examining the source.  Also, the outputs corresponding to these don’t appear to actually be hooked up?

felipesanches: I'm guessing these are the LEDs it represents:
image

Full picture at:
http://www.bitsavers.org/pdf/tektronix/44xx/pictures/4404_CPU_T.JPG

PR #13543, opened 2025-03-31
model2: wait state when writing to copro TGP input FIFO; also adjust TGP frequencies on Model 1/2

gm-matthew:
The Manx TT course select rotation bug was caused by the i960 running too far ahead of the copro TGP and overwriting 0x918000 before the TGP got a chance to read it; adding a wait state when writing to the input FIFO prevents this.

Also increase frequency of TGP to 16.666 MHz (50 MHz / 3) on Model 2 and decrease to 13.333 MHz (40 MHz / 3) on Model 1; the [MB86232 datasheet](https://bitsavers.trailing-edge.com/components/fujitsu/_dataSheets/MB86232_DSP_Sep89.pdf) specifies that a machine cycle equates to three master clock cycles so presumably the same applies to the MB86233 and MB86234. 13.333 MHz for the MB86233 would match the 75ns cycle time listed for the MB86232, and it is likely that the [MB86234 process shrink](https://caps0ff.blogspot.com/2017/11/tgp-status-update.html) allows it to run at a higher frequency.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -31,8 +31,6 @@
     - manxtt: no escape from "active motion slider" tutorial (needs analog inputs),
               bypass it by entering then exiting service mode;
     - manxtt: no bikes are visible (not a z-sort issue!);
-    - manxtt: course select island map doesn't rotate properly:
-              timing issue, i960 sends read/write geo addresses to bufferram 0x918000/4 which TGP never reads in time.
     - sgt24h: first turn in easy reverse course has ugly rendered mountain in background;
     - skytargt: really slow during gameplay (fixed?);
     - skytargt: short draw distance (might be down to z-sort);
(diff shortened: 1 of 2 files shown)


src/mame/sega/model2.cpp
```diff
@@ -31,8 +31,6 @@
     - manxtt: no escape from "active motion slider" tutorial (needs analog inputs),
               bypass it by entering then exiting service mode;
     - manxtt: no bikes are visible (not a z-sort issue!);
-    - manxtt: course select island map doesn't rotate properly:
-              timing issue, i960 sends read/write geo addresses to bufferram 0x918000/4 which TGP never reads in time.
```
angelosa, line 35: The issue just above is also fixed since the TGP rewrite fwiw.

gm-matthew: Indeed it is. There's probably other issues that can be removed from the list; I need to do some testing

happppp:
Was the TGP clock never measured for model1/model2?
There's a Guru style readme in model1.cpp, but no info on TGP clock.

galibert:
The tgp clock inputs are indeed 40MHz on model1 and 50MHz on model 2.

PR #13545, opened 2025-04-01
Implemented VA EG (Envelope Generator) and VCA (Voltage Controlled Amplifier) sound devices.

m1macrophage:
* Implemented VA_RC_EG (in `sound/va_eg`). Emulates an RC-circuit-based envelope generator. Values are published to a sound stream and can also be accessed directly.
* Implemented VA_VCA (in `sound/va_vca`). Emulates a voltage-controlled amplifier. 
* Replaced the custom EG and VCA implementations in `paia/fatman`, `linn/linndrum` and `oberheim/dmx` with the ones above.

This also required moving initialization and triggering code from the custom VCA devices in the `linndrum` and `dmx` into the core audio device in each driver.

Followed the naming and EG structure suggestions in  #13531. Open to additional suggestions of course.

### DIFF
--- a/src/devices/sound/va_eg.h
+++ b/src/devices/sound/va_eg.h
@@ -15,7 +15,7 @@
 // - Machine configuration: rc_eg.set_c(C);
 // - Start attack:          rc_eg.set_r(attack_r).set_target_v(max_v);
 // - Start decay:           rc_eg.set_r(decay_r).set_target_v(sustain_v);
-// - Start release:         rc_eg.set_r(realse_r).set_target_v(0);
+// - Start release:         rc_eg.set_r(release_r).set_target_v(0);
 //
 class va_rc_eg_device : public device_t, public device_sound_interface
 {
--- a/src/devices/sound/va_eg.h
+++ b/src/devices/sound/va_eg.h
@@ -1,0 +1,47 @@
+// license:BSD-3-Clause
+// copyright-holders:m1macrophage
+
+#ifndef MAME_SOUND_VA_EG_H
+#define MAME_SOUND_VA_EG_H
+
+#pragma once
+
+// Building block for emulating envelope generators (EGs) based on RC circuits.
+// The voltage is published to as a sound stream, and is also available by
+// calling `get_v()`.
+class va_rc_eg_device : public device_t, public device_sound_interface
+{
+public:
+	va_rc_eg_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock = 0) ATTR_COLD;
+
+	// These setters can be used during both: configuration and normal operation.
+	// `r` and `c` must be > 0. To instantly set the capacitor's voltage to
+	// a specific value, use set_instant_v().
+	va_rc_eg_device &set_r(float r);
+	va_rc_eg_device &set_c(float v);
+	// Sets target voltage to (dis)charge towards.
+	va_rc_eg_device &set_target_v(float v);
+	// Sets the voltage to the given value, instantly.
+	va_rc_eg_device &set_instant_v(float v);
+
+	float get_v(const attotime &t) const;
+	float get_v() const;  // Get voltage at the machine's current time.
+
+protected:
+	void device_start() override ATTR_COLD;
+	void sound_stream_update(sound_stream &stream, const std::vector<read_stream_view> &inputs, std::vector<write_stream_view> &outputs) override;
+
+private:
+	sound_stream *m_stream;
+
+	float m_r;
+	float m_c;
+	float m_rc_inv;
+	float m_v_start;
+	float m_v_end;
+	attotime m_t_start;
+};
+
+DECLARE_DEVICE_TYPE(VA_RC_EG, va_rc_eg_device)
+
+#endif  // MAME_SOUND_VA_EG_H
--- a/src/devices/sound/va_eg.h
+++ b/src/devices/sound/va_eg.h
@@ -7,7 +7,7 @@
 #pragma once
 
 // Building block for emulating envelope generators (EGs) based on RC circuits.
-// The voltage is published to as a sound stream, and is also available by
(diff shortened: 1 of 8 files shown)


src/devices/sound/va_vca.h
```diff
@@ -0,0 +1,54 @@
...
+
+	// By default, the CV will be treated as a typical gain (output = cv * input)
+	// The configure_*() functions below might change this.
+
+	// CEM3360 (or AS3360) VCA in linear CV configuration: CV connected to pin
+	// Vc, and pins Vo and Ve connected to each other.
+	va_vca_device &configure_cem3360_linear_cv();
+
+	// Ignored when streaming CVs are enabled.
+	void set_fixed_cv(float cv);
```
galibert, line 32: Isn't a cpu-controlled "fixed-cv" usually interpolated through some RC though?  We don't really have jumps in an analog synth (except for VCO sync, but that's another story).

src/devices/sound/va_vca.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+DEFINE_DEVICE_TYPE(VA_VCA, va_vca_device, "va_vca", "Voltage Controlled Amplifier")
+
+va_vca_device::va_vca_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock)
+	: device_t(mconfig, VA_VCA, tag, owner, clock)
+	, device_sound_interface(mconfig, *this)
+	, m_stream(nullptr)
+	, m_has_cv_stream(false)
+	, m_min_cv(-FLT_MAX)
+	, m_max_cv(FLT_MAX)
```
galibert, line 17: maybe 0, 5 or 0, 1 could be a better default?  It wouldn't go into nan/inf at least.

src/devices/sound/va_eg.h
```diff
@@ -0,0 +1,47 @@
+// license:BSD-3-Clause
+// copyright-holders:m1macrophage
+
+#ifndef MAME_SOUND_VA_EG_H
+#define MAME_SOUND_VA_EG_H
+
+#pragma once
+
+// Building block for emulating envelope generators (EGs) based on RC circuits.
```
galibert, line 9: Maybe something like (to contrast with things like the SSM2056):

Building block for emulating envelope generators (EGs) based on a single RC circuit.  The controlling source sets a target voltage and the device interpolates up or down to it through a RC circuit.

src/devices/sound/va_vca.h
```diff
@@ -0,0 +1,54 @@
...
+#define MAME_SOUND_VA_VCA_H
+
+#pragma once
+
+// Emulates a voltage-controled amplifier (VCA). The control value (CV) can
+// be set directly (set_fixed_cv()), or it can be provided in a sound stream
+// (input 1), by a device in va_eg.h, for example. The behavior of specific VCAs
+// can be emulated by using the respective configure_* functions. Note that "CV"
+// ("control value") could either refer to a control voltage, or a control
+// current, depending on the device.
```
galibert, line 14: When the cv is provided via a stream, this is also a ring modulator.

src/devices/sound/va_vca.h
galibert: Oh no that's dumb, the RC is the device, duh.  Even if the 4051 (or whatever) has some impedance, it's probably pretty sharp anyway.

m1macrophage: Nothing is really instant. But yeah, for some cases, fixed might work out OK.

One example is the resistor switching on the DMX, where there is no capacitor. There's likely parasitic capacitance, but probably negligible.

Similarly for the 4051-based S&H example you called out. Typical 4051 resistance is a few hundred Ohm. Combined with the typical 10nF cap, the time constant is 1-10 usec, so pretty sharp indeed (though it might be slower in practice, due to other non-ideal analog effects).

However, there are cases where designers intentionally interpolate CVs (add an extra R and C to the S&H circuits), to avoid "step" artifacts. Emulating those S&Hs accurately would require using an RC EG for each S&H.

src/devices/sound/va_vca.cpp
m1macrophage: Went with 100.  The expectation is that the VCA device should be able to process voltage streams. Though an even smaller limit could be used (50? 20?).

src/devices/sound/va_eg.h
m1macrophage: Yeah, that reads better. Added. Also included an example.

src/devices/sound/va_eg.h
```diff
@@ -6,9 +6,17 @@
...
+// Building block for emulating envelope generators (EGs) based on a single RC
+// circuit. The controlling source sets a target voltage and the device
+// interpolates up or down to it through a RC circuit. The voltage is published
+// as a sound stream and also available from `get_v()`.
+//
+// For example, emulating a CPU-controlled ADSR EG will look something like:
+// - Machine configuration: rc_eg.set_c(C);
+// - Start attack:          rc_eg.set_r(attack_r).set_target_v(max_v);
+// - Start decay:           rc_eg.set_r(decay_r).set_target_v(sustain_v);
+// - Start release:         rc_eg.set_r(realse_r).set_target_v(0);
```
galibert, line 18: Typo "realse" :-)

src/devices/sound/va_vca.h
galibert: Actually I'm wrong, a ring modulator allows for negative values, sorry.

src/devices/sound/va_eg.h
m1macrophage: 😭

src/devices/sound/va_vca.h
galibert: ...and I'm wrong when I say I'm wrong, tat's not the EQ, that's the vca.  Need more sleep.

PR #10714, opened 2022-12-21
Victor9k sound ii

pauldevine:

PR #13549, opened 2025-04-03
Get audio working on Victor 9000 (victor9k)

donohoe00:
Audio on the Victor 9000 was not working, mainly due to missing functionality on the transmit path of the mc6852 SSDA.

Summary of changes:
- mc6852: Add code to initiate serial output of transmit data.
- mc6852: Implement transfer complete callback, which keeps sending data from FIFO, and sends sync or mark codes upon FIFO underrun.
- mc6852: fix a couple of small coding errors.
- mc6852/m68sfdc: Add a function to allow an existing (unorthodox?) transmit mechanism to continue working, to avoid breaking m68sfdc
- 6522via: software running on the victor9k seems to rely on a hardware behavior that wasn't implemented in the emulator: setting the counter latch value to 0 produces a high-frequency output on PB7 in continuous mode, but software expects this not to output a clock signal on PB7 (symptom is a high-pitch squeal when software wants to produce no sound).
- Implement a low-pass to filter the codec output.

### DIFF
--- a/src/devices/machine/6522via.cpp
+++ b/src/devices/machine/6522via.cpp
@@ -535,7 +535,14 @@ TIMER_CALLBACK_MEMBER(via6522_device::t1_tick)
 	if (T1_CONTINUOUS (m_acr))
 	{
 		m_t1_pb7 = !m_t1_pb7;
-		m_t1->adjust(clocks_to_attotime(TIMER1_VALUE + IFR_DELAY));
+		if (TIMER1_VALUE > 0)
+		{
+			m_t1->adjust(clocks_to_attotime(TIMER1_VALUE + IFR_DELAY));
+		}
+		else
+		{
+			m_t1_active = 0;
+		}
 	}
 	else
 	{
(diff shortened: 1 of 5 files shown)


simzy39:
Great PR. Maybe there is something worthwhile in reading another attempt: https://github.com/mamedev/mame/pull/10714

donohoe00:
Thanks for the pointer to the earlier CR!  I took the lowpass filtering from that CR, and it seemed to improve the sound quality.  In the process, I found and fixed a couple of things I missed in mc6852.  I think I'm fairly confident wrt. the mc6852 changes - in fact I'd say there's a reasonable chance it will cause some other devices to start functioning!

donohoe00:
Is there anything I need to do to progress this?

rb6502:
Sorry about that.  It's just that anything touching the 6522 VIA sets off alarm bells because it has a lot of users in MAME and some of them are fairly picky about the behavior.  I'll try out your changes on a bunch of drivers later and if nothing breaks the rest of it looks clean.

PR #13553, opened 2025-04-03
bus/snes/st018.cpp: Add primary support of ST018 coprocessor

cam900:
moritas2 is now boots, and in-game AI is works.

### DIFF
--- a/src/devices/bus/snes/snes_carts.cpp
+++ b/src/devices/bus/snes/snes_carts.cpp
@@ -20,6 +20,7 @@
 #include "sfx.h"
 #include "sgb.h"
 #include "spc7110.h"
+#include "st018.h"
 #include "sufami.h"
 #include "upd.h"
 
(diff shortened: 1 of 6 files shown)


src/devices/bus/snes/snes_carts.cpp
```diff
@@ -40,7 +41,7 @@ void snes_cart(device_slot_interface &device)
 	device.option_add_internal("lorom_sgb2",    SNS_LOROM_SUPERGB2);    // SuperGB2 base cart - unsupported
 	device.option_add_internal("lorom_st010",   SNS_LOROM_SETA10);
 	device.option_add_internal("lorom_st011",   SNS_LOROM_SETA11);
-	device.option_add_internal("lorom_st018",   SNS_LOROM);             // Cart + ST018 - unsupported
+	device.option_add_internal("lorom_st018",   SNS_LOROM_ST018);       // Cart + ST018 - unsupported
```
angelosa, line 44: Comment here is outdated, probably worth stripping entirely (should be obvious by the device name anyway, that's valid for the other entries too).

src/devices/bus/snes/st018.cpp
```diff
@@ -0,0 +1,161 @@
+// license:BSD-3-Clause
+// copyright-holders:Fabio Priuli
```
angelosa, line 2: If you did this then it's legit not an @etabeta78 attribution

src/mame/nintendo/snes.cpp
```diff
@@ -1200,7 +1200,10 @@ void snes_console_state::machine_start()
 			case SNES_CX4:      // this still uses the old simulation instead of emulating the CPU
 			case SNES_ST010:    // this requires two diff kinds of chip access, so we handle it in snes20_lo/hi_r/w
 			case SNES_ST011:    // this requires two diff kinds of chip access, so we handle it in snes20_lo/hi_r/w
-			case SNES_ST018:    // still unemulated
+				break;
+			case SNES_ST018:      // skeleton support
+				m_maincpu->space(AS_PROGRAM).install_read_handler(0x003800, 0x0038ff, 0, 0xbf0000, 0, read8sm_delegate(*m_cartslot, FUNC(base_sns_cart_slot_device::chip_read)));
+				m_maincpu->space(AS_PROGRAM).install_write_handler(0x003800, 0x0038ff, 0, 0xbf0000, 0, write8sm_delegate(*m_cartslot, FUNC(base_sns_cart_slot_device::chip_write)));
 				break;
```
angelosa, line 1207: That is anything but `skeleton`.

src/devices/bus/snes/st018.cpp
cam900: It's based on bus/snes/upd.cpp, that's why i unchanged this.

PR #13559, opened 2025-04-05
emu/sound.h: Improved performance for sound ~3%-4%

holub:
`u32 samples()` is heavily used and has significant impact on execution, possible because `s32 -> u32`
In this PR, I'm precalculating m_samples value which gives some boost to performance.

`m_start`, `m_end` have been renamed to `m_startt`, `m_endd` in order to prove scope and validate compilation. They must be renamed back when confirmed.

before:
```
» ./mame docastle -bench 300
Average speed: 1411.73% (299 seconds)
```
after:
```
» ./mame docastle -bench 300
Average speed: 1460.54% (299 seconds)
```

### DIFF
--- a/src/emu/sound.h
+++ b/src/emu/sound.h
@@ -234,7 +234,7 @@ protected:
 	{
 		// start has to be set after end, since end can expand the buffer and
 		// potentially invalidate start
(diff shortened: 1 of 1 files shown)


galibert:
That code is going to be entirely changed in less than a week, sorry.

PR #13568, opened 2025-04-07
Msx: adding Video80 homebrew card

fulivi:
Hi,

with this PR I'd like to add the video80 card I built for my MSX1 system back in 1988, please.
I also fixed the ROM image for the Philips VY0010 floppy interface.
Thanks.
--F.Ulivi

### DIFF
--- a/src/devices/bus/msx/cart/video80.cpp
+++ b/src/devices/bus/msx/cart/video80.cpp
@@ -28,7 +28,6 @@
 #include "screen.h"
 
 // Debugging
-#undef VERBOSE
 #define VERBOSE 0
 #include "logmacro.h"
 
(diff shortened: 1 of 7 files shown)


src/devices/bus/msx/cart/video80.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "cpu/z80/z80.h"
+#include "video/mc6845.h"
+
+#include "emupal.h"
+#include "screen.h"
+
+// Debugging
+#undef VERBOSE
+#define VERBOSE 0
+#include "logmacro.h"
```
angelosa, line 33: Why you need to `#undef VERBOSE` like at all?

fulivi: I don't know, it's been carried over by copy-paste since forever...

PR #13571, opened 2025-04-09
machine/dp83932c.cpp: Advance CRBA on packet RX

### LINKED ISSUE #12491: NEWS-OS 4.2.1aR networking issues with the NWS-5000X and SONIC drivers
### MAME version

0.266 (and latest master)

### System information

Linux (Fedora Workstation 40)
i9-12900K, 64GB RAM
US English locale and display language

### INI configuration details

_No response_

### Emulated system/software

nws5000x

### Incorrect behaviour

XDMCP sessions do not function on versions of MAME built after commit d98993d7ff92cf07b9378bd79f358e37dc8e2084 (`git bisect` pointed to this commit)

Telnet and some other basic things still function. The `sxdm` login screen shows up, but typing just results in nothing being echoed back in the username field and the system doesn't respond to the packets in the expected way (shown in the Wireshark capture below):
image

Good capture from the commit before:
image

### Expected behaviour

XDMCP sessions should function normally, and allow the user to type in their login credentials.

### Steps to reproduce

1. On an affected version, boot NEWS-OS 4.2.1aR
2. Enable networking, and start `sxdm` on the emulated NWS-5000X
3. Attempt to start an X session from the host system using the Xephyr program: `Xephyr -query 192.168.2.5 -screen 1920x1080x8 -fp <font path with NEWS-OS fonts> +byteswappedclients :2`
4. Try to type in a username into the `sxdm` login prompt

### Additional details

I suspect this is related to how the SONIC emulation interacts with MAME's networking layer, given the commit in question didn't touch the SONIC driver but did change some other networking stuff, and the fact that the `nws3260` driver, which uses the LANCE chip instead, is not affected on NEWS-OS 4.1R. I also would not be surprised if the NWS-5000X driver's configuration of the SONIC/networking is causing this somehow.

I'm not very familiar with the internals of MAME's networking stack, but I will try to debug this when I can. Maybe someone else already knows what could have happened here :) I can work with anyone on this if needed (OS images, further testing, etc) as well.

briceonk:
Resolves #12491

The commit I identified in the issue appears to have revealed a fun race condition in the SONIC driver. If another packet is received before the emulated host acknowledges it and touches the RBA, the driver was overwriting the previous packet with the new one rather than advancing to the next open space in the RBA. This would cause a lot of duplicate ACKs that would cause the connection to be really slow at best, or totally non-functional at worst. With this change, the NWS-5000X driver is fully restored to an operational state.

Screenshot From 2025-04-08 23-41-21

There are still some spurious TCP retransmissions, but these do not cause any functional issues as far as I can tell (and showed up in MAME 0.261 as well when I was testing this, so is not a regression related to this change).

Oddly, the datasheet does not explicitly mention CRBA increment, but there is the following snippet that shows that it should be advancing:
image

I also sanity-checked this change by checking the QEMU SONIC emulation, and it has the same CRBA increment functionality: https://github.com/qemu/qemu/blob/dfaecc04c46d298e9ee81bd0ca96d8754f1c27ed/hw/net/dp8393x.c#L804, so this seems legit. 

@pmackinlay Let me know if there is something else I should do to fix this instead (or if you'd like me to test this with any other drivers before merging).


rb6502:
This sounds VERY similar to a SONIC failure I was having with 68k MacOS.  Will check it out.

PR #13575, opened 2025-04-10
model2: improve z-sort and clipping

gm-matthew:
Polygons in later windows are always drawn after polygons from earlier windows, regardless of z value. This appears to fix z-sorting in all games that I have tested. At the moment I'm leaving the "bucket sort" implementation intact but I wonder if there's a better way to do it. I would guess that the real hardware is performing an insertion sort since it can z-sort the polygons as they are being received from the geometry DSPs.

Four planes representing the visible frustum are now used for polygon clipping, rather than a single plane for z-clipping. It is no longer necessary to add 1.0 to z during projection to prevent texture corruption, and so objects very close to the camera (such as the course map in Daytona, which has a z value of 0.00009) are now visible at the correct size and position.

I have removed z-clip culling as it was only resulting in polygons close to the camera being erroneously culled. The master z-clip is set to 0xFF by all games that I have seen, with a few setting it to 0 when not actively rendering; it looks to me more like a "render enable" register than a z-clip value.

model2_v.cpp mentions the existence of a Model 2B-CRX manual which was used in writing the Model 2 renderer emulation; if anyone has access to this manual I would greatly appreciate a copy or link.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -10,15 +10,13 @@
     MAME driver by R. Belmont, Olivier Galibert, ElSemi and Angelo Salese.
 
     TODO:
-    - z-sort, focal distance, color gamma and Mip Mapping still needs to be properly sorted in the renderer;
+    - color gamma and Mip Mapping still needs to be properly sorted in the renderer;
     - sound comms still needs some work (sometimes m68k doesn't get some commands or play them with a delay);
     - 2C games needs TGPx4 emulation;
     - outputs and artwork (for gearbox indicators);
     - clean-ups;
 
     TODO (per-game issues)
-    - daytona: crashes when coining it up with master network active
-               culprit is a wrong command parameter in geo_parse texture data opcode;
     - daytona: car glasses doesn't get loaded during gameplay;
     - doa, doaa: corrupted sound, eventually becomes silent;
     - dynamcopc: corrupts palette for 2d (most likely unrelated with the lack of DSP);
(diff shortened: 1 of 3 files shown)


TheLostNinja:
Some z-sorting issues with enemy bullets still left in Zero Gunner.

angelosa:
> model2_v.cpp mentions the existence of a Model 2B-CRX manual which was used in writing the Model 2 renderer emulation; if anyone has access to this manual I would greatly appreciate a copy or link.

Any contact method?

> I have removed z-clip culling as it was only resulting in polygons close to the camera being erroneously culled. The master z-clip is set to 0xFF by all games that I have seen, with a few setting it to 0 when not actively rendering; it looks to me more like a "render enable" register than a z-clip value.

However that's what the documentation claims, I presume the formula used is wrong then.

> The index value of the reciprocal of the smallest Z value among all the vertices that form the polygon (i.e., 1/Z) when expressed in IEEE single-precision floating-point decimal format is compared with the PCR; if the value is greater than the PCR (the vertex is close to the viewport), the polygon is erased (not displayed).
When the system is reset, the initial value of the PCR is "0", so all polygons are erased (not displayed).
In order to display polygons, it is necessary to set a suitable value in the PCR. To completely disable this function, set the value 255 in the PCR.

gm-matthew:
> Any contact method?

You can contact me by e-mail at robot_wars_master@yahoo.co.uk.

> > The index value of the reciprocal of the smallest Z value among all the vertices that form the polygon (i.e., 1/Z) when expressed in IEEE single-precision floating-point decimal format is compared with the PCR; if the value is greater than the PCR (the vertex is close to the viewport), the polygon is erased (not displayed).
> > When the system is reset, the initial value of the PCR is "0", so all polygons are erased (not displayed).
> > In order to display polygons, it is necessary to set a suitable value in the PCR. To completely disable this function, set the value 255 in the PCR.

Okay, I think I can figure out how to make this work.

gm-matthew:
> Some z-sorting issues with enemy bullets still left in Zero Gunner.

I've identified the cause of this issue: the enemy bullets are set to use the z value from the previous polygon in the model, but there is no previous polygon because each bullet is a single quad. The previous z value is reset to 1.0e10 for each object, so the enemy bullets all end up underneath everything except the background.

I already have a preliminary fix where the previous z value is only reset at the start of each frame which seems to work, though more testing will be needed to make sure the fix doesn't break any other games. Also I'd like to see if the Model 2B manual has anything to say about previous z values.

PR #13577, opened 2025-04-10
sound/dac76.cpp: Emulating multiplying capability and using it in dmx and linndrum.

m1macrophage:
* Added support for reference current (multiplying capability).
* Added support for voltage output.
* Added support for streaming reference current.
* Used those capabilities in `oberheim/dmx.cpp` and `linn/linndrum.cpp`.

I went with this approach because it better matches the real device. The DAC76 is a multiplying DAC. But I can think of some alternative implementations. @galibert or others, let me know if one of those is preferable:
1. Add a `configure_dac76_voltage_output()` in `sound/va_vca`, and have drivers that need multiplying capability always follow the `dac76_device` with a `va_vca_device`. This is similar to the implementation before this PR, but the dac76-specific logic will move from individual drivers to `va_vca`.
1. Add an `mdac76_device` subclass in `sound/dac76`, so as not to clutter the `dac76_device` interface.

### DIFF
--- a/src/devices/sound/dac76.cpp
+++ b/src/devices/sound/dac76.cpp
@@ -38,10 +38,36 @@ dac76_device::dac76_device(const machine_config &mconfig, const char *tag, devic
 	device_t(mconfig, DAC76, tag, owner, clock),
 	device_sound_interface(mconfig, *this),
 	m_stream(nullptr),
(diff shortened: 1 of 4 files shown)


galibert:
I like keeping close to the reality of the device, and I don't see the point with an external vca.

PR #13578, opened 2025-04-10
cpu/avr8/avr8.cpp: Updates:

cam900:
- Fix address mask (word(2 byte) unit)
- Fix boot size
- Fix PC shifting in state_string_export
- verified from their datasheets

handheld/pensebem.cpp: Fix internal ROM region size

makerbot/replicator.cpp: Use lowercase hexadecimal values

skeleton/venteta.cpp: Fix CPU type for fix PC size (16 bits - also word unit) and internal RAM size (4000 bytes)

ultimachine/rambo.cpp: Fix internal ROM region size, Use lowercase hexadecimal values

### DIFF
--- a/src/mame/skeleton/venteta.cpp
+++ b/src/mame/skeleton/venteta.cpp
@@ -105,7 +105,7 @@ void venteta_state::venteta(machine_config &config)
 {
 	ATMEGA1280(config, m_maincpu, 24_MHz_XTAL); // TODO: actually ATMEGA103
 	m_maincpu->set_addrmap(AS_PROGRAM, &venteta_state::program_map);
-	m_maincpu->set_addrmap(AS_PROGRAM, &venteta_state::data_map);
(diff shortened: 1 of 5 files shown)


src/mame/skeleton/venteta.cpp
```diff
@@ -96,8 +103,9 @@ GFXDECODE_END
 
 void venteta_state::venteta(machine_config &config)
 {
-	ATMEGA168(config, m_maincpu, 24_MHz_XTAL); // TODO: actually ATMEGA103
+	ATMEGA1280(config, m_maincpu, 24_MHz_XTAL); // TODO: actually ATMEGA103
 	m_maincpu->set_addrmap(AS_PROGRAM, &venteta_state::program_map);
+	m_maincpu->set_addrmap(AS_PROGRAM, &venteta_state::data_map);
```
ajrhacker, line 108: This map should be installed for `AS_DATA` rather than overriding the previous `AS_PROGRAM` map.

PR #13579, opened 2025-04-11
update info. Actually is ym2413 koftball.cpp

flama12333:
not a mcu.


flama12333:
my friend had database about of sound chip clone now added recently. 
https://wiki.agiri.ninja/sound_chip_clones:indexa

flama12333:
also my question is has protection? becauses does not has mcu. in the rom program yes?

PR #13582, opened 2025-04-11
model2: fix error in clipping plane calculation

gm-matthew:
Panoramic views during attract sequence of various games in linked mode should now work properly


TheLostNinja:
Maybe an offtop, but is there a way to use a multicab for link-featured games in a single emu session, like in a case of Sega GP Rider or Cool Riders?

rb6502:
@TheLostNinja This is not a discussion forum.  But the answer is "no".

PR #13583, opened 2025-04-11
victor9k: Fix very slow floppy disk performance

donohoe00:
- Removed support for the 8048 SCP for the sake of simplicity and maintainability (further rationale in the commit msg)
- Mimic the SCP reset behaviour, now that the SCP is no longer present
- Fixes around the reading of signals which can be output by both the SCP and the VIAs

Result is a big speedup: retries/timeouts that were happening previously do not occur, and the system now seems to perform as expected.


cuavas:
I think we’d rather have the 8048 emulation fixed properly rather than removed altogether.  We prefer to emulate stuff in more detail where possible, and this should be within MAME’s current capabilities.

rb6502:
I agree, it oughta be possible to get it to work while actually running the 8048.  I can probably be convinced to merge this as-is if you'd agree to make another attempt with the 8048 though.  We can help you with the MAME-specific details there.  I have way more experience syncing things with 8-bit microcontrollers in MAME than anyone ought to, and Vas is good at that sort of thing too.

donohoe00:
I've pushed a new, more minimal set of changes, which are functionally equivalent to the old ones, but do not remove any 8048 related code.  This leaves the door open to debugging and fixing the drive control algorithm in the future.

donohoe00:
Converted to draft since I'm making progress on the 8048-related issues.

IntergalacticMicrosystems:
This has been working great for me - tested with lots of file copying with DOS and DOS utilities, making bootable disks, and general use with single and double-sided images.
Much more useful with the speed improvements.

pauldevine:
Agree this current iteration mimics the timing on the vintage hardware more accurately. I've also tested it over several days and is working reliably for me.

rb6502:
I completely missed that this was ready for review, sorry.  Could you sync up and fix the conflict?

PR #13584, opened 2025-04-12
emu/emumem.h: Optimized cache line detection

holub:
That just 1-2% sometimes but still...

### DIFF
--- a/src/emu/emumem.h
+++ b/src/emu/emumem.h
@@ -1877,13 +1877,15 @@ public:
 
 	// see if an address is within bounds, update it if not
 	void check_address_r(offs_t address) {
-		if(address > m_addrend_r || address < m_addrstart_r)
-			m_root_read->lookup(address, m_addrstart_r, m_addrend_r, m_cache_r);
+		if(EXPECTED(address >= m_addrstart_r && address <= m_addrend_r))
+			return;
+		m_root_read->lookup(address, m_addrstart_r, m_addrend_r, m_cache_r);
 	}
 
 	void check_address_w(offs_t address) {
-		if(address > m_addrend_w || address < m_addrstart_w)
-			m_root_write->lookup(address, m_addrstart_w, m_addrend_w, m_cache_w);
+		if(EXPECTED(address >= m_addrstart_w && address <= m_addrend_w))
+			return;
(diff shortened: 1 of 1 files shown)


cuavas:
Does this actually change the compiled code when optimisations are on?  Can you show the before/after compiler output?

If you’re trying to hint to the compiler that a particular condition is likely/unlikely to be met, the `EXPECTED`/`UNEXPECTED` helper macros (from src/osd/osdcomm.h) expand to the relevant compiler helpers (see [here](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) for GCC documentation).

cuavas:
To support my feel for what the compiler does, consider this: https://godbolt.org/z/WcqPr9c88

It’s a slightly contrived example, but you can see that GCC produces equivalent structure in either case.  In both cases, it does two comparisons, short-circuiting one if the other is true, and does a forward branch past the code to call the function.

The one change that this makes is that it tests the upper bound before testing the lower bound (the order matters because it has to avoid the memory access for the comparison that’s short-circuited), but I wouldn’t expect that to make a difference considering that in the vast majority of cases, you’d hope that the address is within the current range which requires both comparisons to be performed.

holub:
I feel same way, that compiler should make its own decision. The bench of pacman shows this code more often up from overage than original. But definitely for such numbers they can't be relyable. I'll try to grab output later.
The real reason I made this change because I see somethung strange in profiler: pacman never exceeded boundaries here. It always use 'return' in this check.

cuavas:
I mean, using the `UNEXPECTED` or `EXPECTED` helper will encourage GCC and clang to optimise for one case or the other, but I wouldn’t expect just changing the expressions to reliably have the desired effect.

holub:
That is awesome:

> In general, you should prefer to use actual profile feedback for this (-fprofile-arcs), as programmers are notoriously bad at predicting how their programs actually perform.

cuavas:
> That is awesome:
> 
> > In general, you should prefer to use actual profile feedback for this (-fprofile-arcs), as programmers are notoriously bad at predicting how their programs actually perform.

Yeah, feel free to ignore that.

holub:
Honestly I can't be sure pacman's case is average or edge case. Let's stash it for now.

holub:
Alright. docastle shows noticeable +2%

PR #13587, opened 2025-04-13
Update strings.po

Gently28:
The revision refers to the Windows 11 Chinese Simplified menu buttons and option titles, many places are still not perfect, modify it again and again!Characteristic:
(1) Corrected the traditional and English characters that are currently left in MAME, such as 統 (統), 將 (will), (INI) initialization, etc., Figure 1
(2) Depending on the region, the name will be based on the mainland rather than Hong Kong and Taiwan, such as screen (screen), program (program), export (export), etc., Figure 2
(3) Delete duplicate redundant or unobstructed notes, etc., Figure 3
(4) Added the complete plug-in items that have not yet been integrated, Figure 4.
Left Official Version Right Modified style

### DIFF
--- a/language/Chinese_Simplified/strings.po
+++ b/language/Chinese_Simplified/strings.po
@@ -1,5 +1,5 @@
 # Chinese Simplified translations for MAME package
-# MAME 套件的简体中文翻译
+# MAME 中文简体语言包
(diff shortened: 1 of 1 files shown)


MooglyGuy:
Having heard no objections from any party which would be in a position to understand Chinese (either simplified or otherwise), merging.

PR #13588, opened 2025-04-13
machine/rescap.h: Implemented audio potentiometer law.

m1macrophage:
Used it in oberheim/dmx.cpp and linn/linndrum.cpp.


m1macrophage:
It built locally, but looks like pow() cannot be used in a constexpr in a lot of setups. Marking as draft.

m1macrophage:
Should be good now.

PR #13591, opened 2025-04-13
Add skeleton for Casio CPS-2000

devinacker:
Also some fixes to the uPD934 and MSM6200.

New NOT_WORKING machine
----------
Casio CPS-2000 [BCM, Devin Acker]

### DIFF
--- a/src/mame/casio/cz1.cpp
+++ b/src/mame/casio/cz1.cpp
@@ -910,7 +910,7 @@ void cz1_state::cz1(machine_config &config)
 	m_mcu->t0_in_cb().set(FUNC(cz1_state::sync49_r));
 	m_mcu->t1_in_cb().set([this] () { return BIT(m_main_port[2], 7); });
 
-	MSM6200(config, "kbd").irq_cb().set_inputline(m_mcu, MCS48_INPUT_IRQ);
+	MSM6200(config, "kbd", 2'470'000).irq_cb().set_inputline(m_mcu, MCS48_INPUT_IRQ);
 
 	config.set_default_layout(layout_cz1);
 }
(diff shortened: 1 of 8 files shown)


src/mame/casio/cz1.cpp
```diff
@@ -910,7 +910,7 @@ void cz1_state::cz1(machine_config &config)
 	m_mcu->t0_in_cb().set(FUNC(cz1_state::sync49_r));
 	m_mcu->t1_in_cb().set([this] () { return BIT(m_main_port[2], 7); });
 
-	MSM6200(config, "kbd").irq_cb().set_inputline(m_mcu, MCS48_INPUT_IRQ);
+	MSM6200(config, "kbd", 2'470'000).irq_cb().set_inputline(m_mcu, MCS48_INPUT_IRQ);
```
cuavas, line 913: Where’s this frequency coming from?

cuavas:
The code looked fine after a quick skim.  I’d just like clarification on where that MSM6200 clock frequency comes from.

src/mame/casio/cz1.cpp
devinacker: There's an oscillator on the main board that goes out to the keyboard PCB (it's hard to make out in the schematic because of the poor scan quality, but the parts list makes it more clear)
image
image

cuavas: You’re supposed to use `_XTAL` markup for those.  I’ll fix it.

cuavas: Yeah, if it’s a crystal/resonator, use `_XTAL` markup.  If it’s an RC/LC circuit or something, just put in the frequency.

happppp: Yes xtal.cpp includes ceramic oscillators.

devinacker: Good to know, thanks.

PR #13594, opened 2025-04-15
multibus: fixed regression on isbc202

fulivi:
Hi,
this commit fixes a regression on isbc202 emulation caused by cf679363963, which removed ownership of multibus slots by multibus bus. By doing so it removed the propagation of clock down to cards.
iSBC202 was affected because it (probably) is the only multibus card in MAME that relies on clock to come from the bus.
Thanks.
--F.Ulivi

### DIFF
--- a/src/mame/intel/imds2.cpp
+++ b/src/mame/intel/imds2.cpp
@@ -97,7 +97,7 @@ namespace {
 #define IPC_XTAL_Y2     8_MHz_XTAL
 
 // Y1 oscillator of IPC board: 19.6608 MHz
-constexpr auto IPC_XTAL_Y1 = 19.6608_MHz_XTAL;
+static constexpr auto IPC_XTAL_Y1 = 19.6608_MHz_XTAL;
 
 class imds2_state : public driver_device
 {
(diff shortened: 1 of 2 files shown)


src/mame/intel/imds2.cpp
```diff
@@ -97,7 +97,7 @@ namespace {
 #define IPC_XTAL_Y2     8_MHz_XTAL
 
 // Y1 oscillator of IPC board: 19.6608 MHz
-#define IPC_XTAL_Y1     19.6608_MHz_XTAL
+constexpr auto IPC_XTAL_Y1 = 19.6608_MHz_XTAL;
```
cuavas, line 100: If you turn this into a global variable, it needs to have a constant address across an invocation and hence increases static data size.  If it’s only used in a single function, you can make it a local variable.

cuavas (commented):
Do you think this is the best approach?  Would it be better to change the slot constructor to allow a clock frequency to be supplied rather than always using zero?

src/mame/intel/imds2.cpp
ajrhacker: Aren't constexpr variables implicitly inline since C++17?

fulivi:
Better like this?

fulivi:
@cuavas May I ask you if you think this can be merged, please? Thanks

src/mame/intel/imds2.cpp
cuavas: Only `constexpr` `static` data members are implicitly `inline`.  A scope `constexpr` variable implicitly `static`.

But if the address of a `static inline` data member leaks out of a function (e.g. by passing it to `logerror` or something else that implicitly passes by reference), the compiler will instantiate an out-of-line copy, and the linker needs to coalesce the out-of-line copies at link time.  It doesn’t actually prevent the increase in static data size.

cuavas:
Well there are two issues with it now:
* The slot clock used at construction isn’t used, because it gets the bus clock at start anyway.  It’s kind of pointless passing the clock value to slots.
* You still can’t configure slots with different clock frequencies without a separate bus for each group (I don’t know if this can ever happen in real life for multibus, but e.g. PCI systems sometimes have a mixture of 33 MHz and 66 MHz slots).
* If the clock frequency of the bus is changed after start, it doesn’t propagate to the slots.

PR #13596, opened 2025-04-16
src/devices/bus/a2bus/a2frob.cpp: Add Apple II frob device

goldnchild:
Adding a driver for the FrobCo Frob Atari 2600 ROM emulator for the Apple II.

https://forums.atariage.com/topic/380267-the-frob-livesagain/

### DIFF
--- a/src/devices/bus/a2bus/a2frob.cpp
+++ b/src/devices/bus/a2bus/a2frob.cpp
@@ -1,13 +1,22 @@
 // license:BSD-3-Clause
-// copyright-holders:Nathan Woods, Wilbert Pol, David Shah
+// copyright-holders:Nathan Woods, Wilbert Pol, David Shah, Golden Child
 /***************************************************************************
 
-  Atari VCS 2600 driver / Apple 2 Frob Card by Frobco
+  Apple 2 Frob Card by Frobco
 
-********************************/
-//
-// Notes: Once the Frob memory is loaded, you can reset the 2600 with Numpad Minus
-//
+  The Frob Card is a ROM emulator device for the Atari 2600 that operates
+  under the control of an Apple II.  It also allows the Apple II and Atari 2600
+  to exchange information with a bidirectional port.
+
+  The Atari 2600 code comes directly from a2600.cpp and has been converted
+  to be a device instead of a driver.
+
+  The slot interface for the a2600 has been removed for simplicity as well as
+  the PAL a2600 variant.
+
+  Note: Once the Frob memory is loaded, you can reset the 2600 with Numpad Minus
+
+***************************************************************************/
 
 
 #include "emu.h"
@@ -36,32 +45,17 @@ namespace {
 
 static const uint16_t supported_screen_heights[4] = { 262, 312, 328, 342 };
 
-
-// convert it from a driver_device to a
-//  public device_t,
-//  public device_a2bus_card_interface
-// machine_start -> device_start
-// machine_reset -> device_reset
-
-// remove the slot interface for simplicity
-
-
-
-//class a2600_frob_base_device : public driver_device
 class a2600_frob_base_device :
 	public device_t,
 	public device_a2bus_card_interface
 {
 
 public:
-
 	a2600_frob_base_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
 
 	INPUT_CHANGED_MEMBER(reset_a2600);
 
-
 protected:
-
 	a2600_frob_base_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, u32 clock) :
 	device_t(mconfig, type, tag, owner, clock),
 		device_a2bus_card_interface(mconfig, *this),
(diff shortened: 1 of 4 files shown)


src/devices/bus/a2bus/a2frob.cpp
```diff
@@ -0,0 +1,430 @@
...
+	u8 a2600_read_fff2(offs_t offset)
+	{
+		if (!machine().side_effects_disabled()) m_byte_waiting_flag_vcs = 0;
+		return m_byte_for_vcs;
+	}
+
+	u8 a2600_read_fff1(offs_t offset)
+	{
+		return (m_byte_waiting_flag_apple ? 0 : 1) << 7 | (m_byte_waiting_flag_vcs << 6);
+	}
```
cuavas, line 113: Why have you made this stuff `public`?

Also, you need to use `synchronize` barriers to get data from one CPU to the other to avoid things being seen in the wrong order.  The card CPU should see the flag indicating that the mailbox for the host CPU is full as soon as it writes it, but the host CPU should not see the flag or the updated value until your on the other side of a sync barrier.  The same applies for communication in the opposite direction.

src/devices/bus/a2bus/a2frob.cpp
```diff
@@ -0,0 +1,430 @@
...
+	switch (offset & 0x1)
+	{
+		case 0:
+			m_frob_control_reg = data;
+			m_frob_apple_control = BIT(data, 4);
+			m_bidirectional_active = BIT(data, 5);
+			m_frob_page = BIT(data, 0, 4);
+//          printf("m_frob_apple_control = %x  m_frob_page = %x\n",m_frob_apple_control, m_frob_page);
+			break;
+		case 1:
```
cuavas, line 187: Use logmacro.h helpers for configurable logging, not commented prints.  Commented code always rots.

cuavas (changes requested):

src/devices/bus/a2bus/a2frob.cpp
happppp: synchronize() doesn't work the opposite direction*, some drivers do trickery with machine().scheduler().perfect_quantum(x) to make sure the 'opposite' side comms answer timing is correct.

*:
scheduler runs CPUs in serial, CPU A, then CPU B.
let's say CPU A: 1ms.
scheduler runs CPU B.
At 0.5ms point, CPU B wants to write to a latch that shared with CPU A. CPU A however, is already in the future (from CPU B perspective)

cuavas: And to solve that properly (since you don’t know in advance whether a CPU is going to attempt to communicate across scheduling domains), you need an efficient way to roll back and retry a smaller timeslice until you get an “acceptable” delay.  But since the CPUs themselves don’t know who’s ahead at any given time, putting both directions through `synchronize` at least means that whichever CPU is ahead will try to end its timeslice early when possible.

happppp: Is it possible for CPU B to be ahead of CPU A? And I don't mean slight deviations with icount < 0.

eg. you have 3 CPUs in a driver. Then the scheduler runs the CPUs in this order A,B.C.A,B,C,A,B,C, etc. (or A,B,C,B,C,B,C if CPU B does a synchronize()). It doesn't eenie meenie miney mo them. CPU B is never ahead of CPU A. CPU C is never ahead of CPU A or B.
The order is the same as the machine config order. It's why for arcade drivers with a 1-way soundlatch, it's important to list the audiocpu after maincpu.

src/devices/bus/a2bus/a2frob.cpp
```diff
@@ -0,0 +1,430 @@
...
+#include "sound/tiaintf.h"
+#include "speaker.h"
+#include "../../mame/atari/tia.h"
+
+#include "machine/mos6530.h"
+
+//#define VERBOSE (LOG_GENERAL)
+#include "logmacro.h"
+
+#include "a2frob.h"
```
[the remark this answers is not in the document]
goldnchild: ok, I've resorted the header, I hope it's properly ordered now.

ok, I've copied the tia from src/mame/atari/tia.cpp into a file a2frobtia.cpp so there will be no cross dependency.

src/devices/bus/a2bus/a2frob.cpp
```diff
@@ -0,0 +1,430 @@
...
+	virtual void write_cnxx(u8 offset, u8 data) override
+	{ if (!m_frob_apple_control) m_frob_mem[(offset & 0xff)|(m_frob_page << 8)] = data; }
+
+	virtual bool take_c800() override { return false; }
+
+	u8 m_frob_mem[0x1000] = { 0 };  // 4k frob memory
+
+	u8 read_frob_mem(offs_t offset) { return !m_frob_apple_control ? 0 : m_frob_mem[offset & 0xfff]; }
+
+	void write_frob_mem(offs_t offset, u8 data) { m_frob_mem[offset & 0xfff] = data; }
```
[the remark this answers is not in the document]
goldnchild: ok

src/devices/bus/a2bus/a2frob.cpp
```diff
@@ -0,0 +1,430 @@
...
+		case 1:
+			if (!machine().side_effects_disabled())
+			{
+				m_byte_waiting_flag_apple = 0; // clear byte waiting for apple
+			}
+			return m_byte_for_apple;
+		default:
+			return 0;
+	}
+}
```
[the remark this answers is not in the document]
goldnchild: ok, I have put in some synchronize calls to do the sync, I hope that that will work.

PR #13597, opened 2025-04-16
Add Windows on ARM build to CI

hmartinez82:
Starts addressing https://github.com/orgs/mamedev/discussions/112

Now that Github has Windows on ARM runners, one can use them for native builds under the MSYS2 CLANGARM64 environment. 

The `mingw-w64-clang-aarch64-gcc-compat` package just adds an alias where `gcc` and `g++` are actually `clang` and `clang++`. There is some internal Makefile that is probably not honoring the `OVERRIDE_` variables, so an aliased `gcc` did the trick.

image

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -68,3 +68,48 @@ jobs:
           chdman.exe
           unidasm.exe
         if-no-files-found: error
+
+  build-windows-arm64:
+    runs-on: windows-11-arm
+    defaults:
+      run:
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        subtarget: [mame, tiny]
+        include:
+          - subtarget: mame
+            executable: mame
+          - subtarget: tiny
+            executable: mametiny
+    steps:
+    - uses: msys2/setup-msys2@v2
+      with:
+        msystem: CLANGARM64
+        install: git make mingw-w64-clang-aarch64-cc mingw-w64-clang-aarch64-python mingw-w64-clang-aarch64-libc++ mingw-w64-clang-aarch64-gcc-compat
+    - uses: actions/checkout@main
+      with:
+        fetch-depth: 0
+    - name: Build
+      env:
+        MINGW64: "/clangarm64"
+        OVERRIDE_AR: "llvm-ar"
+        OVERRIDE_CC: clang
+        OVERRIDE_CXX: clang++
+        ARCHOPTS: "-fuse-ld=lld"
+        SUBTARGET: ${{ matrix.subtarget }}
+        ARCHOPTS_C: " -Wno-nontrivial-memcall -march=armv8.2-a"
+        ARCHOPTS_CXX: " -Wno-nontrivial-memcall -march=armv8.2-a"
+        TOOLS: 1
+        PLATFORM: arm64
+      run: make -j3
+    - name: Validate
+      run: ./${{ matrix.executable }}.exe -validate
+    - uses: actions/upload-artifact@main
+      with:
+        name: ${{ matrix.executable }}-windows-clang-arm64-${{ github.sha }}
+        path: |
+          ${{ matrix.executable }}.exe
+          chdman.exe
+          unidasm.exe
+        if-no-files-found: error
--- a/3rdparty/bimg/3rdparty/astc-encoder/source/astcenc_mathlib.h
+++ b/3rdparty/bimg/3rdparty/astc-encoder/source/astcenc_mathlib.h
@@ -70,6 +70,9 @@
 #ifndef ASTCENC_NEON
   #if defined(__aarch64__)
     #define ASTCENC_NEON 1
+    #undef ASTCENC_SSE
+    #undef ASTCENC_AVX
+    #undef ASTCENC_POPCNT
   #else
     #define ASTCENC_NEON 0
   #endif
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-windows.yml
```diff
@@ -68,3 +68,48 @@ jobs:
...
+      run:
+        shell: msys2 {0}
+    strategy:
+      matrix:
+        subtarget: [mame, tiny]
+        include:
+          - subtarget: mame
+            executable: mame
+          - subtarget: tiny
+            executable: mametiny
```
cuavas, line 84: This should be integrated into the matrix with the other builds.

There’s also no point building the tiny subtarget with the same configuration as the full build.  We build the tiny subtarget with clang for x86-64 to test that the Windows-specific parts of MAME can be built with clang without generating too much load on the CI servers.

.github/workflows/ci-windows.yml
```diff
@@ -68,3 +68,48 @@ jobs:
...
+    - name: Build
+      env:
+        MINGW64: "/clangarm64"
+        OVERRIDE_AR: "llvm-ar"
+        OVERRIDE_CC: clang
+        OVERRIDE_CXX: clang++
+        ARCHOPTS: "-fuse-ld=lld"
+        SUBTARGET: ${{ matrix.subtarget }}
+        ARCHOPTS_C: " -Wno-nontrivial-memcall -march=armv8.2-a"
+        ARCHOPTS_CXX: " -Wno-nontrivial-memcall -march=armv8.2-a"
```
cuavas, line 102: What’s triggering the `nontrivial-memcall` warning?  We don’t do these kinds of workarounds in CI files:
* In general we try to address warnings if it makes sense and the compiler isn’t just being stupid.
* We demote warning-as-error to warning in the Lua build scripts when it’s something we want to clean up.
* We suppress problematic warnings that we can’t work around in the Lua build scripts for third-party libraries.

Why do you need to specify `-march` options?  What does it default to?

cuavas:
I added it to the matrix here just to see what happens: cuavas/mame@1466eb986000936246752f358239bf07acb8456a

.github/workflows/ci-windows.yml
hmartinez82: - I can remove  `-Wno-nontrivial-memcall` because only Clang 20 will fail to build it, but the mamedev runners are using Clang 19 still which I think will be fine.
 
- `-march` is necessary to bring the instruction set that 7z is expecting. 8.2 is totally fine because even the Snapdragon 8cx (the one from 2019) supports it.

hmartinez82:
Ah, understood. I tried to be as orthogonal as I could with my changes to not mess with the existing code. But that should work. Although you are missing one more entry for `mametiny` built for aarch64

.github/workflows/ci-windows.yml
hmartinez82: Ahhhhhhhh ok . I had no idea.

makefile
```diff
@@ -142,7 +142,7 @@ MAKEPARAMS := -R
 ifeq ($(OS),Windows_NT)
 OS := windows
 GENIEOS := windows
-PLATFORM := x86
+PLATFORM ?= x86
```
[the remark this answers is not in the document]
hmartinez82: I have not researched that, but since the MSYS2 shell is emulated, `PROCESSOR_ARCHITECTURE` will always be AMD64 even on an ARM64 machine.

3rdparty/bimg/3rdparty/astc-encoder/source/astcenc_mathlib.h
```diff
@@ -70,6 +70,9 @@
 #ifndef ASTCENC_NEON
   #if defined(__aarch64__)
     #define ASTCENC_NEON 1
+    #undef ASTCENC_SSE
+    #undef ASTCENC_AVX
+    #undef ASTCENC_POPCNT
```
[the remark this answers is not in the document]
hmartinez82: Of course course about the upstream, but not undefining these is what's causing this to fail: https://github.com/cuavas/mame/actions/runs/14523675557/job/40750274562

I wasn't going to hang this PR on waiting for upstream. If you prefer that I can try pushing it up.

.github/workflows/ci-windows.yml
cuavas: I’d still rather leave it with the default profile for now.  We aren’t supplying `-march` for x86-64 builds either, even though our binary releases require x86-64-v2 features.

cuavas:
I added it myself in aa3dc30a1838d3f0983b744b20834de4dd4ef836.

3rdparty/bimg/3rdparty/astc-encoder/source/astcenc_mathlib.h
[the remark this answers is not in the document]
cuavas: No it isn’t.  What caused that to fail was having `PLATFORM` set to `x86` for an ARM build.  There’s absolutely no need for this hack.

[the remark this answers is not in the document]
hmartinez82: Ah awesome. Ok, that is perfect.  Indeed, the last thing I fixed was setting the PLATFORM to arm64.  So I didn't think to go back to my previous fixes and see if that solved everything. Awesome finding!

PR #13607, opened 2025-04-19
cpu/z80/z80.cpp: Get rid of large flag LUTs

holub:
@happppp yeah, this one is better than #13605 
Not sure what I tested before...


src/devices/cpu/z80/z80.cpp
```diff
@@ -40,8 +40,6 @@ u8 z80_device::SZ_BIT[] = {};   // zero, sign and parity/overflow (=zero) flags
 u8 z80_device::SZP[] = {};      // zero, sign and parity flags
 u8 z80_device::SZHV_inc[] = {}; // zero, sign, half carry and overflow flags INC r8
```
holub, line 41: @cuavas do you remember what was your consern about efficiency of such vars?

cuavas: I’m not sure exactly what the context is, but in many cases, calculating values is faster than using tables due to better locality and less TLB pressure.  We’ve benchmarked it with parity calculations – even without a population count instruction, the Hamming weight calculation ended up faster than using lookup tables.

holub: Thanks. We are on the same page then. I just didn't remember if I had missed something else.

PR #13610, opened 2025-04-21
linn/linndrum.cpp: Modeled click filter, improved hihat decay emulation, and did cleanups.

m1macrophage:
Primary changes:
* Hihat decay knob will have an effect even if changed during playback.
* Modeling the "click" filter. Validated with a [simulation](https://falstad.com/circuit/circuitjs.html?ctz=CQAgjCAMB0l3BWEAmAbAZlqgHKsBOSZffMdVEBSSy6hAUwFowwAoAdxABY4Ufuu2PtUisAxgKHJ+CMMmHdwTITFQJUcAkQDs6rsm3oosSBBhtOs+dLr5r-UZbsKEzg-NETX17fO8h0ZGoubmhtZRBGGFMiQnUwSG1IfDxIBCNo0yhWAHNKZ0DbeXRyKGyne2pUXwVHEGrK+pr3bIAnJp95BoDS6mkwVFyOnopu9AQPcuGbEF8+h1YwJO5eceKJgI2IIJR4aggLYfVqOcoNbLzutVGa7pFWdu7-K-UyuUhBzifzp+c6-2OR3OojyzxkzmuZU8+T8zjBwVC4UYQiiJgm6mO2kMXHQSS4SEyZg4MLORRJ0K4glJlDk1JCYAiqlQeGZ2kpS30pgohOMhx4dHOlKEgNE7Ss1IBwJ6umJ-M28jlMzqcrWKwF9wAHrN9HwkAkqdIhPSQgAJAAKADEADoAZwAFOwABb0AB2toAbgB7AA2AFcALb0W0270ASwAJvRWrbQzbbQBDAAutoSkAApABKaC2gAy9AAZsmbTaxJ6AA70aCsLWGfBnCgJUrqOvGkAAIUttodzrdNq9fsDwbDkejNoAPCm4JnszaAEqhnKOosl8uV6vcAa7CgIV5BFvgeQAYTAtoA9LbDwB9Q85gCSh4A0peACKHts5gDyj-XnMoyD85zoootLHmeF7Xnej4vm+n7flq+JGOgBCULoAT-sBR7IGBNpXje95PgA6reZoAKKzj+O71LiKHaPU-D0ph2G4ZBhHEWR67aGAQjoMkICENxtKtq0l7sKGFbCTkLrhraXZWgAjvJtrCV6oZiPQl6SeGGbrggOKULglCUQg2D7mAIStKINY7KgxCzPgtEYAq4DmSJYlRhpUkcRgG7yNgkBGmASCtv2Ab0Ou2CcRutHYNgRi-kJll8cgQVmeAcAKkswEuaJ4keeGXlBfiIB+cEVKtqmaYccsXA1No+BGqhra6JVNbVfZdnCrwrZUC1fFcBAlJIPZCplc5fADOFWICNF2AhPiCKpYlEUDWyfFENwU2tiUMpavgOo1UNhgbUaY3vIMWrLbQ1CEH4kBxWN82JXtc3tSktFUAtc0TOuJABAUfl-UFY0WmIMZxja9DxjaoZRraiaeh6UahvmACetpOq6cPOsG7D0PQZZg7aYYupDrSsJ6aUHioKy-TASHJcYez2dcDN9JT6DkweVNlDwtPQIFNkEiYcDM+oDPbFzHNAA) (see changes in `update_volume_and_pan`).
  * Required fixing the `HIGHPASS1P` mode in `flt_biquad` (see below).

Other minor changes:
* Renamed input for hihat decay pot.
* Updated "mux drums" section comments.
* Updated snare_w comments.
* Other minor comment changes.

**flt_biquad changes**

The `HIGHPASS1P` mode was not being used anywhere AFAICT. That mode did not seem to work: when tested with a sweep, it produced a flat response.

The implementation in this PR is based on the bilinear transform for an RC filter, including frequency unwrap (following the  relevant [wikipedia info](https://en.wikipedia.org/wiki/Bilinear_transform)) . The parameters it produces match those in the book *Designing Audio Effect Plugins in C++*  (*proving* they are the same requires some tedious algebra). The implementation was also validated with sine sweeps and sine waves of specific frequencies.

I was tempted to change `LOWPASS1P` in a similar way. I validated the new implementation as above. But the existing implementation works, and it is used in multiple places (indirectly, via `opamp_mfb_lowpass`). So I saw no reason to change it. Not in this PR anyway.

### DIFF
--- a/src/devices/sound/flt_biquad.cpp
+++ b/src/devices/sound/flt_biquad.cpp
@@ -527,6 +527,7 @@ void filter_biquad_device::recalc()
 				break;
 			// For highpass and friends, block the entire signal.
 			case biquad_type::HIGHPASS1P:
+			case biquad_type::HIGHPASS1P1Z:
 			case biquad_type::HIGHPASS:
 			case biquad_type::BANDPASS:
 			case biquad_type::PEAK:
@@ -558,9 +559,16 @@ void filter_biquad_device::recalc()
 				m_b1 = m_b2 = m_a2 = 0.0;
 				break;
 			case biquad_type::HIGHPASS1P:
-				m_b0 = 1.0 / (K + 1.0);
-				m_b1 = -m_b0;
-				m_a1 = (K - 1.0) / (K + 1.0);
+				m_a1 = -exp(-2.0 * M_PI * (0.5 - m_fc / m_stream->sample_rate()));
+				m_b0 = 1.0 + m_a1;
+				m_a1 = -m_a1;
+				m_b1 = m_b2 = m_a2 = 0.0;
+				break;
+			case biquad_type::HIGHPASS1P1Z:
+				normal = 1.0 / (K + 1.0);
+				m_b0 = normal;
+				m_b1 = -normal;
+				m_a1 = (K - 1.0) * normal;
 				m_b2 = m_a2 = 0.0;
 				break;
 			case biquad_type::LOWPASS:
--- a/src/devices/sound/flt_biquad.cpp
+++ b/src/devices/sound/flt_biquad.cpp
@@ -558,10 +558,10 @@ void filter_biquad_device::recalc()
 				m_b1 = m_b2 = m_a2 = 0.0;
 				break;
 			case biquad_type::HIGHPASS1P:
-				m_a1 = -exp(-2.0 * M_PI * (0.5 - m_fc / m_stream->sample_rate()));
-				m_b0 = 1.0 + m_a1;
-				m_a1 = -m_a1;
-				m_b1 = m_b2 = m_a2 = 0.0;
+				m_b0 = 1.0 / (K + 1.0);
+				m_b1 = -m_b0;
+				m_a1 = (K - 1.0) / (K + 1.0);
+				m_b2 = m_a2 = 0.0;
 				break;
 			case biquad_type::LOWPASS:
 				m_b0 = Ksquared * normal;
--- a/src/devices/sound/flt_biquad.h
+++ b/src/devices/sound/flt_biquad.h
@@ -25,7 +25,8 @@ public:
 		PEAK,
 		LOWSHELF,
 		HIGHSHELF,
-		RAWPARAMS
+		RAWPARAMS,
+		HIGHPASS1P1Z
 	};
 
 	struct biquad_params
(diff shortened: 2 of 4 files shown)


galibert:
You'll probably be interested in https://www.native-instruments.com/fileadmin/ni_media/downloads/pdf/VAFilterDesign_2.1.0.pdf

The time discretisation methods make me wonder if I should work on the sc01 again..

src/devices/sound/flt_biquad.cpp
Lord-Nightmare, line 1: This is actually a different filter entirely; Highpass1P is 1 pole no zeroes; there's a different filter I have yet to add but have locally which is Highpass1P1Z which is 1 pole 1 zero, which is equivalent to what you just changed here.

I.e.
```
			case biquad_type::LOWPASS1P:
				m_a1 = exp(-2.0 * M_PI * (m_fc / m_stream->sample_rate()));
				m_b0 = 1.0 - m_a1;
				m_a1 = -m_a1;
				m_b1 = m_b2 = m_a2 = 0.0;
				break;
			case biquad_type::HIGHPASS1P:
				m_a1 = -exp(-2.0 * M_PI * (0.5 - m_fc / m_stream->sample_rate()));
				m_b0 = 1.0 + m_a1;
				m_a1 = -m_a1;
				m_b1 = m_b2 = m_a2 = 0.0;
				break;
			case biquad_type::LOWPASS1P1Z:
				normal = 1.0 / (1.0 / K + 1.0);
				m_b0 = m_b1 = normal;
				m_a1 = (1.0 - 1.0 / K) * normal;
				m_b2 = m_a2 = 0.0;
				break;
			case biquad_type::HIGHPASS1P1Z:
				normal = 1.0 / K + 1.0;
				m_b0 = normal;
				m_b1 = -normal;
				m_a1 = (K - 1.0) * normal;
				m_b2 = m_a2 = 0.0;
				break;
```

m1macrophage: Thanks for the info! I added and used the `HIGHPASS1P1Z` implementation, and reverted `HIGHPASS1P` to the previous version.

I added `HIGHPASS1P1Z` at the end of the `biquad_type` enum, to avoid messing up existing save states. But let me know if you prefer this to be declared below `HIGHPASS1P`.

Note that there seems to be a missing parenthesis in `normal = ...` of the posted implementation. Also, I couldn't get `HIGHPASS1P` to work (neither on the "click" in the linddrum, nor with sine sweeps). Not sure what's going on there.

m1macrophage:
Yeah, I discovered that book a couple of months back. Though still a lot of reading to do.

I also bumped into the sc01 implementation, when searching for info on filters in MAME. Cool tutorial on the bilinear transform, in the comments! :)

src/devices/sound/flt_biquad.cpp
Lord-Nightmare: You're absolutely right about the missing parenthesis, I hadn't tested that locally and it was wrong, thanks for catcing that!

The HIGHPASS1P does work as far as I know, but its response curve isn't extremely useful (its mostly flat unless the cutoff is very close to the nyquist frequency), it was more added since its one of the 'general' biquad equations I found documented online.

m1macrophage: Ah, I see. That explains it.

PR #13618, opened 2025-04-23
raizing_batrider.cpp: added note about undumped bbakraid set

sergiopolog:
Added note about undumped bbakraid set: (Hong Kong) (Tue May 25 1999),  based on the non-unlimited version of the game

[https://www.youtube.com/watch?v=1Fm6kpPTZkM](https://www.youtube.com/watch?v=1Fm6kpPTZkM)


sergiopolog:
Are these kinds of notes allowed on source code comments?

PR #13622, opened 2025-04-25
cpu/z80/nsc800.cpp: fix typo in name

holub:


scripts/src/cpu.lua
```diff
@@ -3048,20 +3048,18 @@ if CPUS["Z80"] or CPUS["KC80"] or CPUS["Z80N"] or CPUS["T6A84"] then
 		MAME_DIR .. "src/devices/cpu/z80/r800.h",
 		MAME_DIR .. "src/devices/cpu/z80/z84c015.cpp",
 		MAME_DIR .. "src/devices/cpu/z80/z84c015.h",
-		MAME_DIR .. "src/devices/cpu/z80/t6a84.cpp",
-		MAME_DIR .. "src/devices/cpu/z80/t6a84.h",
 	}
```
cuavas, line 3051: Why does this also remove these from the build?

happppp: duplicate entry

cuavas: Oh never mind, I see it’s duplicated from further up.

PR #13624, opened 2025-04-26
CD-i: Add Cursor Blink

### LINKED ISSUE #13522: CD-i: Cursor does not blink
### MAME version

2.74

### System information

N/A

### INI configuration details

```INI

```

### Emulated system/software

CD-i

### Incorrect behaviour

The cursor does not support the Blink feature.

### Expected behaviour

The cursor will blink when the blinking feature is used.

### Steps to reproduce

1. You can use the Validation Disc (EU).
2. Basecase UCM
3. Graphics Cursor
4. GC_Blnk

You will notice the cursor does not blink or change colors.

### Additional details

I have a patch which adds this feature. I will submit it later when I have some time.

Vincent-Halver:
Intends to fix #13522.

This also fixes an error in mcd212.h which dropped the middle bit of the Cursor's ON timer. I changed it to use bitshifted binary which more clearly is readable as a correctly shifted mask.

The time calculation was produced algebraically by following the description on the page in the comment. It should work for NTSC as well as PAL, however NTSC games have not been tested.

To test, I recommend the Validation Disc (Europe). Use the Cursor Blink test, and verify the timing of the blinks correctly matches the description for each test.

This also rewrites the cursor's draw logic to be slightly cleaner.

### DIFF
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -740,39 +740,45 @@ void mcd212_device::mix_lines(uint32_t *plane_a, bool *transparent_a, uint32_t *
 void mcd212_device::draw_cursor(uint32_t *scanline)
 {
 	if (!(m_cursor_control & CURCNT_EN))
-	{
+	{ // Cursor is Disabled
 		return;
 	}
 
-	uint16_t y = (uint16_t)screen().vpos();
-	const uint16_t cursor_x =  m_cursor_position & 0x3ff;
-	const uint16_t cursor_y = ((m_cursor_position >> 12) & 0x3ff) + m_ica_height;
-	if (y >= cursor_y && y < (cursor_y + 16))
+	const bool invert = BIT(m_cursor_control, CURCNT_BLKC_SHIFT);
+	uint8_t color_index = (m_cursor_control & CURCNT_COLOR);
+
+	if (m_blink_active)
 	{
-		const int width = get_screen_width();
-		uint32_t color = s_4bpp_color[m_cursor_control & CURCNT_COLOR];
-		y -= cursor_y;
-		if (m_cursor_control & CURCNT_CUW)
-		{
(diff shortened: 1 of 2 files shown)


src/mame/philips/mcd212.cpp
```diff
@@ -740,39 +740,45 @@ void mcd212_device::mix_lines(uint32_t *plane_a, bool *transparent_a, uint32_t *
...
-	{
+	{ // Cursor is Disabled
 		return;
 	}
 
-	uint16_t y = (uint16_t)screen().vpos();
-	const uint16_t cursor_x =  m_cursor_position & 0x3ff;
-	const uint16_t cursor_y = ((m_cursor_position >> 12) & 0x3ff) + m_ica_height;
-	if (y >= cursor_y && y < (cursor_y + 16))
+	const bool invert = BIT(m_cursor_control, CURCNT_BLKC_SHIFT);
```
ajrhacker, line 747: This constant isn't defined.

rb6502:
@ajrhacker do you have any further issues with this?

src/mame/philips/mcd212.h
```diff
@@ -235,6 +236,10 @@ class mcd212_device : public device_t,
 	emu_timer *m_ica_timer = nullptr;
 	emu_timer *m_dca_timer = nullptr;
 
+	// Cursor State
+	uint16_t m_blink_time; // Counter that tracks how long since the last m_blink_active last changed.
+	bool m_blink_active = false;
```
cuavas, line 241: You haven’t registered these new data members for saved states in `device_start`.  Please check that all data members that hold live state are registered for saved states.

cuavas (changes requested):

src/mame/philips/mcd212.h
Vincent-Halver: Thanks, yes that makes sense. It should be fixed now.

Vincent-Halver:
@cuavas Any further concerns?

PR #13625, opened 2025-04-26
add in game music info in the dip switch for jxzh and koftball koftball.cpp

flama12333:
the sw1 number 2 could be auto-play? in kaimenhu but unused in jxzh due to different.
since will start automatically


cuavas:
I don’t think there’s much value to this PR – the driver is already in considerably better shape now with more DIP switches identifie.

PR #13629, opened 2025-04-27
ef9345: apply same double-height scaling as the real chip

fabio-d:
Before this commit, characters with the double-height attribute (H) were scaled by equally repeating all their scanlines twice.

In the real chip, the first scanline is repeated three times and the last one only once.

---
Image comparisons (source code [here](https://github.com/fabio-d/minitel-ef9345-testsuite/blob/512d2378df58b25683431e9b077d3fa9b13f88c8/tests/test_size.py#L103)):
 * real output from both EF9345 and TS9347 chips (captured via a [custom board](https://github.com/fabio-d/minitel-ef9345-testsuite/tree/main/hw_devboard)):
real_chip
* current MAME output:
mame

The four boxes show the four different scaling modes supported by the chip:
1. Regular size (i.e. no scaling)
1. Double width
1. Double height
1. Double width and height

As you can see, in the last two boxes (those involving double height) the first three lines are different and all the subsequent lines are shifted. This PR makes the output of MAME match the real chips.

### DIFF
--- a/src/devices/video/ef9345.cpp
+++ b/src/devices/video/ef9345.cpp
@@ -352,25 +352,25 @@ uint8_t ef9345_device::get_dial(uint8_t x, uint8_t attrib)
 void ef9345_device::zoom(uint8_t *pix, uint16_t n)
 {
 	uint8_t i, j;
-	if ((n & 0x0a) == 0)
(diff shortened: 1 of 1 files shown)


fabio-d:
Thanks! cc @jfdelnero

PR #13630, opened 2025-04-28
cpu/z80: Optimized detection of attention needed in service step

holub:


holub:
Here is a bench for clang. For gcc seams like there is no change.
```
~/workspace/mame (z80-gen-test) » repeat 3 ./mame-clang-master pacman -bench 100 && sleep 1
Average speed: 4163.01% (99 seconds)
Average speed: 4078.89% (99 seconds)
Average speed: 4025.29% (99 seconds)
------------------------------------------------------------------------------------------------------------------------------------
~/workspace/mame (z80-gen-test) » repeat 3 ./mame pacman -bench 100 && sleep 1
Average speed: 4130.63% (99 seconds)
Average speed: 4110.27% (99 seconds)
Average speed: 4106.35% (99 seconds)
-----------------------------------------------------------------------------------------------------------------------------------
~/workspace/mame (z80-gen-test) » repeat 3 ./mame-clang-master docastle -bench 100 && sleep 1
Average speed: 800.43% (99 seconds)
Average speed: 799.59% (99 seconds)
Average speed: 803.15% (99 seconds)
-------------------------------------------------------------------------------------------------------------------------------------
~/workspace/mame (z80-gen-test) » repeat 3 ./mame docastle -bench 100 && sleep 1
Average speed: 806.14% (99 seconds)
Average speed: 812.21% (99 seconds)
Average speed: 811.35% (99 seconds)
--------------------------------------------------------------------------------------------------------------------------------------
~/workspace/mame (z80-gen-test) » repeat 3 ./mame-clang-master zexall -bench 100 && sleep 1
Average speed: 3785.92% (99 seconds)
Average speed: 3791.36% (99 seconds)
Average speed: 3729.37% (99 seconds)
--------------------------------------------------------------------------------------------------------------------------------------
~/workspace/mame (z80-gen-test) » repeat 3 ./mame zexall -bench 100 && sleep 1
Average speed: 3771.44% (99 seconds)
Average speed: 3848.18% (99 seconds)
Average speed: 3840.11% (99 seconds)
```

src/devices/cpu/z80/z80.cpp
```diff
@@ -621,34 +619,35 @@ void z80_device::device_start()
...
+	state_add(Z80_DE2,              "DE2",       m_de2.w);
+	state_add(Z80_HL2,              "HL2",       m_hl2.w);
+	state_add(Z80_WZ,               "WZ",        WZ);
+	state_add(Z80_R,                "R",         m_rtemp).callimport().callexport();
+	state_add(Z80_I,                "I",         m_i);
+	state_add(Z80_IM,               "IM",        m_im).mask(0x3);
+	state_add(Z80_IFF1,             "IFF1",      m_iff1).mask(0x1);
+	state_add(Z80_IFF2,             "IFF2",      m_iff2).mask(0x1);
+	state_add(Z80_HALT,             "HALT",      m_halt).mask(0x1);
+	state_add(Z80_INPUT_LINE_BUSRQ, "BUSRQ",     m_busrq_state).mask(0x1);
```
happppp, line 650: I'm not sure about adding input pins to state_add. That aside, there's a conflict here:
Z80_INPUT_LINE_BUSRQ = 2
Z80_A = 2

holub: Ah, I'll remove then

cuavas (commented):
You need to be careful with turning things into bit fields.

It turns updates into a read/modify/write operation, because it now has to read the current value, set or clear a bit, and then write it back to memory since there are no memory operations narrower than a byte.  If you use individual `bool` variables, updates only require a write operation to set the new value (except on things like early Alpha AXP processors that lack memory operations narrower than a word).

It’s only more efficient if it’s tested a lot more frequently than it’s updated as it can reduce the number of memory operations required to test all the flags at once.  If it’s updated frequently, the read/modify/write operations will more than cancel out what you save when testing it.

(You also need to be careful turning things into bitfields in threaded code as the updates are definitely not atomic, but that doesn’t apply here.)

holub:
> You need to be careful with turning things into bit fields.
> 
> It turns updates into a read/modify/write operation, because it now has to read the current value, set or clear a bit, and then write it back to memory since there are no memory operations narrower than a byte. If you use individual `bool` variables, updates only require a write operation to set the new value (except on things like early Alpha AXP processors that lack memory operations narrower than a word).
> 
> It’s only more efficient if it’s tested a lot more frequently than it’s updated as it can reduce the number of memory operations required to test all the flags at once. If it’s updated frequently, the read/modify/write operations will more than cancel out what you save when testing it.
> 
> (You also need to be careful turning things into bitfields in threaded code as the updates are definitely not atomic, but that doesn’t apply here.)

We are good here. Save/update is rare for those. The goal is to avoid multiple IFs in service step for each opcode. Ultimately we manually guarantee single IF instead of a chain unless any of these operations really required attention.

PR #13632, opened 2025-04-28
Workarounds for Linux build with older SDL version

ajrhacker:
This is a rather crude workaround for the fact that `SDL_GetDefaultAudioInfo` was first added in SDL2 version 2.24.0. The Linux distro I have only supports 2.0.20, and the instructions in `docs/source/initialsetup/compilingmame.rst` (which still specifies "2.0.14 or later") do not work for updating to a later version.

The reverse enumeration order, which puts "sof-hda-dsp Speaker + Headphones" first instead of last on my setup, is an even grosser hack that may be worse for others.

### DIFF
--- a/src/osd/modules/sound/sdl_sound.cpp
+++ b/src/osd/modules/sound/sdl_sound.cpp
@@ -97,13 +97,18 @@ int sound_sdl::init(osd_interface &osd, const osd_options &options)
 	// Capture is not implemented in SDL2, and the enumeration
 	// interface is different in SDL3
 	int dev_count = SDL_GetNumAudioDevices(0);
+#if 0
 	for(int i=0; i != dev_count; i++) {
+#else
+	for(int i=dev_count-1; i != -1; i--) {
+#endif
 		SDL_AudioSpec spec;
 		const char *name = SDL_GetAudioDeviceName(i, 0);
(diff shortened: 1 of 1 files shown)


src/osd/modules/sound/sdl_sound.cpp
```diff
@@ -97,13 +97,18 @@ int sound_sdl::init(osd_interface &osd, const osd_options &options)
...
+#else
+	for(int i=dev_count-1; i != -1; i--) {
+#endif
 		SDL_AudioSpec spec;
 		const char *name = SDL_GetAudioDeviceName(i, 0);
 		int err = SDL_GetAudioDeviceSpec(i, 0, &spec);
 		if(!err)
 			m_devices.emplace_back(name, spec.freq, spec.channels);
 	}
+#if SDL_MAJOR_VERSION > 2 || SDL_MINOR_VERSION >= 24
```
cuavas, line 111: You’re supposed to use `#if SDL_VERSION_ATLEAST(x, y, z)` for these kinds of things.

cuavas:
I cribbed the part to get MAME building with older versions of SDL here: d2bfb8c3f4fa8d8f6e39c20c55d20ede8c4cd950

PR #13633, opened 2025-04-28
build: update vs/msbuild settings to reflect current clangcl

pmackinlay:
* separate msvc, intel and clang compiler options

### DIFF
--- a/scripts/genie.lua
+++ b/scripts/genie.lua
@@ -1230,9 +1230,11 @@ configuration { "vs20*" }
 if _OPTIONS["vs"]==nil then
 		buildoptions {
 			"/bigobj",
-			"/Zc:preprocessor",
-			"/utf-8",
 			"/permissive-",
+			"/utf-8",
+			"/Zc:enumTypes",
+			"/Zc:preprocessor",
+			"/Zc:templateScope",
 		}
 
 		buildoptions {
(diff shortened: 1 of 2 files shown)


pmackinlay:
This allows current master to build with clangcl (either in Visual Studio or MSBUILD) with some very minor source changes. I'll submit the latter soon pending conversations with other contributors. MSVC still builds as well/badly as before, but I have not tested the Intel compiler; I've attempted to leave its settings as-is, and don't know whether they are still valid or not.

pmackinlay:
Longer term, I'd like to align the gcc/clang/clangcl warnings but may require replacing the current "gmake or ninja" test to something more explicitly based on the compiler rather than the build environment.

scripts/genie.lua
```diff
@@ -1237,14 +1227,18 @@ configuration { "vs20*" }
...
 			"/permissive-",
-			"/w45038", -- warning C5038: data member 'member1' will be initialized after data member 'member2'
+			"/utf-8",
+			"/Zc:enumTypes",
+			"/Zc:preprocessor",
+			"/Zc:templateScope",
 		}
 
 		buildoptions {
+			"/w45038", -- warning C5038: data member 'member1' will be initialized after data member 'member2'
```
cuavas, line 1241: Where’s this actually being hit?  We have this error enabled for other compilers, so it shouldn’t be happening in our own code.

pmackinlay: This is actually enabling the warning, not disabling it; I don't know why Microsoft made it off by default.

PR #13636, opened 2025-04-29
moog/source.cpp: emulating envelope generators and LFO timing.

m1macrophage:
Used netlists for the simulation of the voltage-to-current converters. The resistor networks in these circuits are non-trivial requiring solving a system of equations, and the transistors are used in their exponential region. Netlist use is confined to DC calculations.

### DIFF
--- a/src/devices/sound/va_eg.h
+++ b/src/devices/sound/va_eg.h
@@ -17,6 +17,16 @@
 // - Start decay:           rc_eg.set_r(decay_r).set_target_v(sustain_v);
 // - Start release:         rc_eg.set_r(release_r).set_target_v(0);
 //
+// The set_*() methods expect a monotonically increasing machine time, and will
+// assert()-fail if that's not the case.
+// If a single instance is used by multiple CPUs, this requirement can be met by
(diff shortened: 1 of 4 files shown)


Lord-Nightmare:
There's a small issue here: the way nl_source.cpp is linked sort of works by accident, since there's no nl_source.h included by source.cpp to let the compiler know that it needs to link in nl_source.cpp (this will break single-driver builds).

The solution, as far as I can tell, is to remove the line

`NETLIST_EXTERNAL(moogsource)`

from source.cpp and instead put it in its own file, nl_source.h, which will be #include'd from source.cpp.
nl_source.h would be something like:
```
// license:BSD-3-Clause
// copyright-holders:m1macrophage

NETLIST_EXTERNAL(moogsource)

```
and that's it.

m1macrophage:
Sent #13680

I tried a single-driver build, and the presence of the .h file didn't seem to make a difference. In both cases, I had to include `nl_source.cpp` in the `files {...}` section for linking to work. But maybe I missed some step, or this is just an issue with different setups.

Regardless, your suggestion matches the standard way to do things, so better to go with that anyway.

PR #13643, opened 2025-04-30
3rdparty/expat: Updated expat to version 2.7.1 (fixes compilation with emscripten 3.1.35).

holub:


sairuk:
thanks for this, I was about to open a discussion about this and some other emscripten compilation issue i've run into recently

holub:
I think newer versions of emscripten with higher clang can still be problematic. But 3.1.35 at least must be fine after this.

sairuk:
Anything after 3.1.35 throws errors. 

I managed to build with 4.0.7 after patching dear-imgui, upstream just ignore the error that was problematic. The resulting output from a 4.0.7 build behaves as per 3.1.36 and above anyway, so not much use atm

PR #13646, opened 2025-04-30
namco/mappy.cpp: Pac'n'Pal settings fix - [Lives != 5][Bonus 0x28] = "30k, 80k & Every…

codefrog2002:
image

Pac'n'Pal settings fix - [Lives != 5][Bonus 0x28] = "30k, 80k & Every 80k" not 30/100/80.  

Behavior observations are backed up by in-game messaging and the bonus tables in ROM at 0xB75A
[b75a:03 00 30 00 00 80 00 ff:	dc.b	0x03,0x00,0x30,0x00,0x00,0x80,0x00,0xff]
The structure of this table is 
0x03 for "3 levels of bonus"
0x00,0x30,0x00 for 30k
0x00,0x80,0x00 for 80k


mamehaze:
why are you testing with a set that clearly isn't the one supported in MAME?

codefrog2002:
Sorry, the relevant data is the same in the BNEI version as in the MAME 'pacnpal' set.

image

codefrog2002:
To answer your question as to "why":  I work on a product line which uses BNEI-modified ROMs with my own proprietary emulator. I compare the behavior of my emulator against MAME from time to time and discovered a discrepancy between the original game's settings and the MAME setting strings.  I did check to make sure the difference wasn't caused by BNEI modifications but then I thoughtlessly used a screen shot which clearly showed the BNEI trademark instead of the original NAMCO message.

mamehaze:
Ok, just thought I'd better ask, because sometimes modifications to such settings are made in modern rereleases (I've modified some in the past before for such releases)

As long as you've confirmed they're the same that's all good though.

PR #13648, opened 2025-04-30
saturn.xml - Bad dumps removed

FoxhackDN:
- Removed darksavja, darksavjb, darksavjc sets
-- darksavja, darksavjb: both data tracks match redump, but audio tracks are incomplete and shifted;
-- darksavjc: Hacked set containing a modified US build of the game

### DIFF
--- a/hash/saturn.xml
+++ b/hash/saturn.xml
@@ -10226,18 +10226,18 @@ Sega logo animation draws offset [VDP2]
 		</part>
 	</software>
 
-	<!-- Identifying Dark Savior (JAP) (DW0579)... -->
-	<software name="darksavja" cloneof="darksav" supported="no">
-		<description>Dark Savior (Japan, alt)</description>
-		<year>1996</year>
-		<publisher>Climax</publisher>
-		<info name="serial" value="T-22101G"/>
-		<info name="release" value="19960830"/>
-		<info name="alt_title" value="ダークセイバー"/>
+	<!-- Identifying Dark Seed (JAP) (DW0616)... -->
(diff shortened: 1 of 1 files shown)


FoxhackDN:
After some testing for my previous commit, I realized that a lot of the extra stuff in the software list might be bad dumps. In this case, darksavja and darjsavjb (Darkwater and linlhutz sets respectively) have matching data tracks to Redump's confirmed dump, but the audio tracks are bad, so I removed them.

darksavjc's data track appears to be a hacked US English version, so that's definitely not supposed to be here.

cuavas:
Remind me because I forgot – did bad dumps end up getting removed in the end?

FoxhackDN:
> Remind me because I forgot – did bad dumps end up getting removed in the end?

Yeah, I did some of them in this pull.

https://github.com/mamedev/mame/pull/13658

I didn't continue examining the other discs because of a lack of space to nab everything and test things. I'm not 100% convinced some of these aren't hacked builds... I wish there was more info about these dumps to begin with. Ugh.

angelosa:
We have several `status="baddump"`, a couple are for indices as per #10308 , the rest is iso-mp3 sourced stuff that needs to be substituted (pretty sure redump did all of them by this point)

FoxhackDN:
Yeah, I was going through their stuff and comparing the data tracks to make sure they were the same game. There's been a bunch of alternate dumps revealed with slightly different audio tracks in the past couple of years.

I just wish I had enough time to do all this stuff properly. And energy.

PR #13652, opened 2025-05-01
Disable assembly including DRC on unsupported platforms

### LINKED ISSUE #13649: 0.277 does not build on ppc64 or s390x Linux
Hi,

it looks like 8c28d3ff7eb420452a8f3499d77b3957ca7edf58 has broken building on platforms on which asmjit is not available. Building fails with
```
/usr/bin/ld: cannot find -lasmjit: No such file or directory
```

belegdol:
Disable all assembly on s390x and DRC assembly on ppc.

This should fix #13649.

I am waiting for the test build to complete hence the draft status.

### DIFF
--- a/makefile
+++ b/makefile
@@ -396,6 +396,18 @@ ifeq ($(findstring aarch64,$(UNAME)),aarch64)
 ARCHITECTURE :=
 endif
 
+ifneq (,$(findstring ppc,$(UNAME)))
+ifndef FORCE_DRC_C_BACKEND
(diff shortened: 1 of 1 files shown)


cuavas:
b4e1f5f1b50fd3402232dd0385cac021ba488db6 should fix this.

belegdol:
> [b4e1f5f](https://github.com/mamedev/mame/commit/b4e1f5f1b50fd3402232dd0385cac021ba488db6) should fix this.

Thanks, will test it later.

PR #13653, opened 2025-05-02
(Proposal) emu/sound.cpp: Have update() also compute the sample spanning the current time.

m1macrophage:
Guarantees that: if update() is called at time `t`, any subsequent call to `sound_stream_update()` will receive a stream with `.start_time() > t`.
<br>

This is a followup to the discussion in #13637.  It is a proposal for what the semantics of `update()` could be. This seems to also match the old behavior, and fixes the assertion failure (in debug builds) in `va_eg.cpp`.

The `+ 1` change ensures that `update()` will also process the sample at `idx` , instead of stopping right before that.

The `+ 500'000'000` change ensures the result of the division rounds up, instead of down.

### DIFF
--- a/docs/source/techspecs/device_sound_interface.rst
+++ b/docs/source/techspecs/device_sound_interface.rst
@@ -272,7 +272,7 @@ multiplies the current gain by the given value.
 The method ``set_sample_rate`` allows to change the sample rate of the
 stream.  The method ``update`` triggers a call of
 ``sound_stream_update`` on the stream and the ones it depends on to
-compute all samples up to and including the one spanning the current time.
+compute all samples whose ``sample_time()`` is smaller than the update time.
 
 
 4. Devices using device_mixer_interface
--- a/docs/source/techspecs/device_sound_interface.rst
+++ b/docs/source/techspecs/device_sound_interface.rst
@@ -272,7 +272,7 @@ multiplies the current gain by the given value.
 The method ``set_sample_rate`` allows to change the sample rate of the
 stream.  The method ``update`` triggers a call of
 ``sound_stream_update`` on the stream and the ones it depends on to
-reach the current time in terms of samples.
+compute all samples up to and including the one spanning the current time.
 
 
 4. Devices using device_mixer_interface
--- a/docs/source/techspecs/device_sound_interface.rst
+++ b/docs/source/techspecs/device_sound_interface.rst
@@ -272,7 +272,8 @@ multiplies the current gain by the given value.
 The method ``set_sample_rate`` allows to change the sample rate of the
 stream.  The method ``update`` triggers a call of
 ``sound_stream_update`` on the stream and the ones it depends on to
-compute all samples whose ``sample_time()`` is smaller than the update time.
+compute all samples whose ``sample_time()`` is smaller or equal to the update
+time.
 
 
 4. Devices using device_mixer_interface
--- a/src/emu/sound.cpp
+++ b/src/emu/sound.cpp
@@ -491,7 +491,8 @@ void sound_stream::init()
 u64 sound_stream::get_current_sample_index() const
 {
 	attotime now = m_device.machine().time();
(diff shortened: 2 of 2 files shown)


happppp:
This appears to fix netlist buffer overflows, and also fixes hunchbak crash.
I only tested a handful of games. I see new crashes in dkong and invaders (SYMBOLS are off so I can't provide a log).

cuavas:
Is that philosophically sound, or should it only calculate samples that are in the past from its point of view?  @galibert?

m1macrophage:
> [...] should it only calculate samples that are in the past from its point of view?

One thing to note: if we consider a sample's "time" to be the value of `sound_stream.sample_time(idx)`, then the above statement will be true even after this PR. The intention of this PR is to have  `update()` compute *all* samples whose `sample_time()` is <= the update time (I will update the doc file accordingly).

But that statement might not be true if we consider a sample as representing a time range, rather than an instance in time. So this all depends a bit on the exact definition of `sample_time`.

<br>
BTW If the preference is to go with the current implementation, to avoid causing other problems, that's totally fine. I can drop this PR and update `va_eg` accordingly. I just raised this because it was an apparent change in behavior. I also know this is a nuanced topic, and I assume people are busy, so not expecting quick answers or decisions.

galibert:
I think you're just hitting a misunderstanding of mine on what end_time did.  Hopefully my next commit will fix it (in a little while, that hit emu.h and thus is forcing a full recompile before I can test).

m1macrophage:
I tested the latest change. Unfortunately, the same issue still exists. Concrete example:

* `update()` gets called at time `4:20330403639246424`  (seconds:attoseconds).
  * Forces a `sound_stream_update` with `stream.start_time() == 4:20000000000000000` and `stream.end_time() == 4:20328125000000000`
* The next call to `sound_stream_update` has a `stream.start_time() == 4:2032812500000000` and `stream.end_time() == 4.20338541000000000`

Note that the start_time on the last `sound_stream_update` invocation is less than the update time.

This can be reproduced in non-debug builds by changing the following assert to a hard failure.
https://github.com/mamedev/mame/blob/8d3207ce92d6d217a51881cc657df5b795acbdbc/src/devices/sound/va_eg.cpp#L78-L81

galibert:
Thought about it, the +1 is needed because you want to include the sample at idx.  But you do not want the rounding, you want the samples to happen at the start of the interval, not in the middle.

galibert:
See b1bc1d310af979fc57a3692ca349b9fa09c34e36

m1macrophage:
Thanks for addressing that!

It is still possible, but harder, to trigger that assertion. Example:

* `update()` called at `8:968971354:151971904`  (seconds:nanoseconds:remaining attoseconds)
  * Invokes `sound_stream_update()` with `stream.start_time() == 8:968968750:0 and stream.end_time() == 8:968971354:0`.
* Subsequent `sound_stream_update()` gets `stream.start_time() == 8:968971354:0 and stream.end_time() == 8:969005208:0`.

Note that the `start_time` in the last step is earlier than the `update()` time. This is what the "rounding" change was addressing, but maybe there is a better way?

Or is this considered OK, and I should just compensate  in `va_eg`, by adding half the sample interval time to `start_time`?

I guess the answer might depend on whether samples are considered points in time or time intervals.

galibert:
Hmmm, that's a precision issue.  What a pain...

m1macrophage:
Yeah... If it ends up being more practical to just compensate in va_eg, I can do that.

galibert:
No, I just need to find how to do 128 bits computations without a 128 bits type...

galibert:
I fail at reproducing the issue, tried with linndumo abd obdmx, how do you do it?

m1macrophage:
It is a bit tricky. I run obdmx (debug build) and after it boots, start spamming the the 'z', 'c', 'v' and 'b' keys as quickly as possible (those are the voices that use the EG). It usually takes a few seconds of doing that for the assertion to trigger, though it can take longer.

galibert:
Ok, my latest push as a fair chance to never trigger that again :-)

PR #13654, opened 2025-05-02
cpu/z80 Improved core source - internal TADR var removed

holub:


src/devices/cpu/z80/z80.lst
```diff
@@ -142,16 +138,16 @@ macro   jr_cond
 
 macro   arg16_call
 	call arg16
-	m_ea = TDAT; TADR = PC-1;
-	call nomreq_addr
+	m_ea = TDAT;
+	call nomreq_addr PC-1
```
holub, line 142: I'm thinking of making macro call like `@nomreq_addr PC-1` (sometimes late)
Speak up if anybody against it.

src/devices/cpu/z80/z80.lst
```diff
@@ -289,254 +281,224 @@ macro   r800:sbc_hl
...
 	if (BC != 0 && !(F & ZF)) {
-		TADR = HL;
-		5 * call nomreq_addr
+		5 * call nomreq_addr HL
 		PC -= 2; WZ = PC + 1; F &= ~(YF | XF); F |= (PC >> 8) & (YF | XF);
 	}
 
 macro   r800:cpdr
 	call cpd
-	call nomreq_addr
```
holub, line 502: @wilbertpol FYI

holub:
The change is mosetly machanical replacment. I've validated with reliable timings/flags tests from my toolset bit apprechiate if someone perform visual diff validation.

src/devices/cpu/z80/z80.lst
```diff
@@ -8817,15 +8403,14 @@ z80n:ed36	# add bc,**
 	BC += TDAT;
 
 z80n:ed8a	# push **
-	call arg16
-	using std::swap; swap(TDAT_H, TDAT_L);
-	call push
+	call arg
+	TDAT_H = TDAT_L;
+	call arg
+	call push TDAT
```
cuavas, line 8409: This looks a little odd in the diff, but I’ll trust you on it.

src/devices/cpu/z80/z80.lst
```diff
@@ -6763,9 +6359,8 @@ edff	# DB   ED
 
 00db	# IN   A,(n)
 	call arg
-	TADR = TDAT8 | (A << 8);
-	call in
-	A = TDAT8; WZ = TADR + 1;
+	call in TDAT8 | (A << 8)
+	A = TDAT8; WZ = (TDAT8 | (A << 8)) + 1; // Check!
```
cuavas, line 6363: Is this going to work the same way?  The WZ update will use the updated A value now while previously it used the original A value.

src/devices/cpu/z80/z80.lst
holub: That's ok. arg16 swaps the values and here we had to swap them again. Just inlined without mirrored swaps

src/devices/cpu/z80/z80.lst
holub: Good catch. I remember I put this comment somwhere but did't find where. Tests work fine so possibly both are "correct". Let me recheck and confirm

holub: Updated. Now it matches original.

src/devices/cpu/z80/z80.lst
```diff
@@ -6763,9 +6359,9 @@ edff	# DB   ED
 
 00db	# IN   A,(n)
 	call arg
-	TADR = TDAT8 | (A << 8);
-	call in
-	A = TDAT8; WZ = TADR + 1;
+	TDAT2 = TDAT8 | (A << 8);
+	call in TDAT2
+	A = TDAT8; WZ = (TDAT2 | (A << 8)) + 1;
```
cuavas, line 6364: OK tell me what I’m missing.  To me, this looks like:
* Get n in TDAT8
* Store full 16-bit address `(A << 8) | TDAT8` in TDAT2
* Do I/O read into TDAT8
* Store value read into A
* Update WZ with full address OR'd with new A value plus one

The old code doesn’t appear to OR with the new A value.  I would’ve expected:
```
00db	# IN   A,(n)
	call arg
	TADR = TDAT8 | (A << 8);
	call in
	A = TDAT8; WZ = TADR + 1;
	TDAT2 = TDAT8 | (A << 8);
	call in TDAT2
	A = TDAT8; WZ = TDAT2 + 1;
```

cuavas: Sorry, copy/paste error, updated comment to include snippet:
```
00db	# IN   A,(n)
	call arg
	TDAT2 = TDAT8 | (A << 8);
	call in TDAT2
	A = TDAT8; WZ = TDAT2 + 1;
```

holub: Sorry for been inattentive. You're right and thank you for been such!

PR #13655, opened 2025-05-02
New working systems ------------------- Flashback [James Wallace, BuyItFixIt]

james-wallace-ghub:


james-wallace-ghub:
If anyone wants Manuals, PCB data etc, the dumper has given me them to share. struggling to think of a 14 pin audio IC, though.

cuavas:
Sorry, not a fan of overridden default assignments when it isn't absolutely necessary – in particular it doesn’t play nicely with controller configuration files.

The inputs in this driver could do with cleanup in general.  The hold buttons should probably be the “hold” or “stop” input types, and there are more things that should be “service” or “memory reset” types.

cuavas:
Also, it helps if you’re consistent with saying whether something is working or not (PR title versus source).

PR #13661, opened 2025-05-03
misc/flower.cpp  Added Flower (Japan)

UED-Wayder:
New working clones
Flower (Japan)

Change Old 'Flower (Japan)' to 'Flower (Japan, bootleg)'

Flower (Japan)
main cpu ROM(1.5j) is same as Flower (US)

sound cpu ROM Label fix

flowerj
Original (US and Japan)

flowerjb
bootleg


happppp:
I see you closed it because of a mistake with cps2.cpp, but I have a question anyway: Are you able to measure the Z80 frequency on the PCB?

UED-Wayder:
I asked the dumper and he said it was dumped in the past and the PCB is currently missing, so unfortunately we were unable to find out.
Sorry I couldn't help you.

PR #13665, opened 2025-05-03
victor9k: Add hard drive support

donohoe00:
Adds support for expansion bus card which interfaces to the Xebec s1410 via SASI.  Adds support for a single controller + hard drive, although a second controller plus hard drive should be possible.  Using FACTORYF and HDSETUP, I was able to create a new hard drive image, create partitions, and get working c: etc. to appear.  HDTEST utility is passing.

### DIFF
--- a/src/mame/act/v9kdmaib.cpp
+++ b/src/mame/act/v9kdmaib.cpp
@@ -1,0 +1,292 @@
+// license:BSD-3-Clause
+
+#include "emu.h"
+#include "v9kdmaib.h"
+
+#include "logmacro.h"
+
+/*
+ * Emulation of the Victor 9000 DMA Interface Board, which interfaces to the
+ * Xebec S1410 Hard Disk controller.
+ *
+ * Source: https://bitsavers.org/pdf/victor/victor9000/Victor_9000_Sirius_1_Hard_Disk_Subsystem.pdf
+ *
+ * Compared to the typical SCSI/SASI controller device, the DMA Interface Board
+ * acts more like a passthrough, and the usual state machine is primarily
+ * implemented on the main CPU.
+ */
+
+DEFINE_DEVICE_TYPE(V9KDMAIB, v9kdmaib_device, "v9kdmaib", "Victor 9000 DMA Interface Board")
+
+v9kdmaib_device::v9kdmaib_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+	v9kdmaib_device(mconfig, V9KDMAIB, tag, owner, clock)
+{
+}
+
+v9kdmaib_device::v9kdmaib_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock) :
+	nscsi_device(mconfig, type, tag, owner, clock),
+	nscsi_slot_card_interface(mconfig, *this, DEVICE_SELF),
+	m_dma_r(*this, 0xff),
+	m_dma_w(*this),
+	m_irq_handler(*this),
+	m_dma_on(false),
+	m_dma_write(false),
+	m_dma_addr(0),
+	m_asserting_ack(false),
+	m_non_dma_req(false),
+	m_ctrl(0),
+	m_data(0),
+	m_bus_ctrl(0),
+	m_irq_state(false)
+{
+}
+
+uint8_t v9kdmaib_device::read(offs_t offset)
+{
+	// Lower address bits are ignored
+	if (offset >= 0x80)
+	{
+		offset &= ~0x1f;
+	}
+	else
+	{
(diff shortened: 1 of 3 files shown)


src/mame/act/v9kdmaib.cpp
```diff
@@ -0,0 +1,292 @@
...
+			// Write only
+			break;
+		case R_DATA:
+			if (m_non_dma_req)
+			{
+				m_non_dma_req = false;
+				scsi_bus->data_w(scsi_refid, 0);
+				m_data = scsi_bus->data_r();
+				m_asserting_ack = true;
+				scsi_bus->ctrl_w(scsi_refid, S_ACK, S_ACK);
```
cuavas, line 69: You need to check `machine().side_effects_disabled()` before doing stuff that affects state on reads.  If you don’t, it causes chaos when debugger memory windows are open.

src/mame/act/v9kdmaib.cpp
```diff
@@ -0,0 +1,292 @@
...
+		else
+		{
+			// ACK will be generated upon direct data register access
+			m_non_dma_req = true;
+		}
+	} else if (m_asserting_ack) {
+		scsi_bus->ctrl_w(scsi_refid, 0, S_ACK);
+		m_asserting_ack = false;
+	}
+
```
cuavas, line 245: Please use Allman style consistently, not random bits of K&R in between.

src/mame/act/victor9k.cpp
```diff
@@ -711,6 +733,11 @@ void victor9k_state::machine_reset()
 //  machine_config( victor9k )
 //-------------------------------------------------
 
+void victor9k_state::scsi_devices(device_slot_interface &device)
+{
+	device.option_add("harddisk", NSCSI_S1410);
+}
```
cuavas, line 739: Why not allow the user to connect any SCSI device they want?  Part of the fun of MAME is that you can use it to develop software that uses peripherals that weren’t commonly used with the system in its heyday.

src/mame/act/victor9k.cpp
```diff
@@ -829,6 +856,19 @@ void victor9k_state::victor9k(machine_config &config)
...
+	NSCSI_CONNECTOR(config, "scsi:7").option_set("v9kdmaib", V9KDMAIB).machine_config(
+		[this](device_t *device)
+		{
+			v9kdmaib_device &v9kdmaib(downcast<v9kdmaib_device &>(*device));
+
+			device->set_clock(15_MHz_XTAL / 3);
+			v9kdmaib.irq_handler().append(m_pic, FUNC(pic8259_device::ir4_w));
+			v9kdmaib.dma_read().set(*this, FUNC(victor9k_state::hd_dma_r));
+			v9kdmaib.dma_write().set(*this, FUNC(victor9k_state::hd_dma_w));
+		});
```
cuavas, line 870: Why not allow more SCSI devices?  Did the system not physically allow it somehow?

cuavas (changes requested):

src/mame/act/victor9k.cpp
donohoe00: The bus interface is actually SASI (pre-cursor to SCSI), and there was no connector external to the machine.  I did enable up to 7 targets, as per your other feedback.  I could add options for CD and/or tape, if you think it makes sense.  I'm not sure if any SASI targets other than hard disks were ever created.

pauldevine:
I've tested this PR on various sized hd's for the Victor 9000 and can report it's working quite reliably. I've been able to do low-level formats of the disk as well as partition, high-level format, install, and boot DOS.

IntergalacticMicrosystems:
This PR code had been working well for me - starting out from scratch with the low-level formatf, hdsetup, and booting DOS from the drive.
Lots of copying images and utilities, it's been 100% robust for me.

src/mame/act/victor9k.cpp
cuavas: There were SASI and M-SASI magneto-optical drives as well, but MAME doesn’t emulate them.

src/mame/act/victor9k.cpp
```diff
@@ -621,6 +631,18 @@ void victor9k_state::fdc_irq_w(int state)
...
+{
+	address_space &program = m_maincpu->space(AS_PROGRAM);
+	return program.read_byte(offset);
+}
+
+void victor9k_state::hd_dma_w(offs_t offset, uint8_t data)
+{
+	address_space &program = m_maincpu->space(AS_PROGRAM);
+	program.write_byte(offset, data);
+}
```
cuavas, line 644: As an optimisation, you should use a `memory_access::specific` rather than going through the generic address space handlers every time, but that can be done later.

PR #13671, opened 2025-05-05
cpu/z80: Changed macro call syntax in generator input

holub:
I found this way more readable.


cuavas:
Haha, now I have to remember *another* use of @ in a programming language.  It’s been:

* Dereference operator in standard Pascal (equivalent to ^), e.g. `foo@.bar := 5;`
* Address of operator in extended Pascal, e.g. `foo.next = @bar;`
* Extended keyword prefix in Objective-C, e.g. `@interface`, `@implementation`, etc.
* Fixed address markup in extended C dialects for embedded systems, e.g. `uint8_t TDR@0x0008, TCR@0x0009;`

I don’t think I’ve seen it used as syntax to invoke a subprogram before.

angelosa:
`@` is also for decorators in [Python](https://peps.python.org/pep-0318/)

holub:
Just because the parser is very dumb here it requires something to represent invocation.
Ideally it shouldn't use any special prefixes.
IMHO this way syntax becomes catchier than before. Feel free to pick any symbol you like - I'm not attached to @ in any way.

holub:
btw logically here it's inline, not invoke

cuavas:
Logically it invokes a subprogram – the fact that it’s inlined rather than called is an implementation detail.

PR #13673, opened 2025-05-05
cocoa debugger: debug view scroll workarounds

biappi:
The Cocoa debugger has an annoying bug where the first time the console "overflows" the bottom, the view will not scroll.

This patch includes a workaround to skip frame size notifications in the middle of the text update, and this seems to fix the issue.

### Before
https://github.com/user-attachments/assets/13edad06-76cc-4f65-bcbb-98f936225fb9 

### After
https://github.com/user-attachments/assets/b3a9a118-748b-4984-a2da-5d82f40237db

### DIFF
--- a/src/osd/modules/debugger/osx/debugview.mm
+++ b/src/osd/modules/debugger/osx/debugview.mm
@@ -49,6 +49,11 @@ static void debugwin_view_update(debug_view &view, void *osdprivate)
 
 
 @implementation MAMEDebugView
+{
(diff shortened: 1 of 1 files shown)


rb6502:
Thanks, this is the best the Cocoa debugger's ever worked!  At some point making our native Cocoa layer (make OSD=mac) work properly so we don't have SDL interfering will be the best solution but that's a project in itself.

PR #13676, opened 2025-05-05
cpu/z80: Flatten 16b rw operations in order to avoid vars manipulations

holub:


holub:
I really want to get rid of TDAT at all but don't see it feasible now. Hopefully graceful refactoring will help to achieve this at some point.

PR #13678, opened 2025-05-06
a2bus/ieee488.cpp: add Apple II IEEE-488 card

goldnchild:
I saw where someone had hacked an Apple II ieee488 card to boot from a Commodore drive so I got interested in the ieee 488 card.

I don't know of anything that it can tested with but the card seems to respond to simple commands.

https://www.applefritter.com/content/new-rom-apple-gpib-card
https://www.youtube.com/watch?v=0wE3RrGb8cc
https://git.applefritter.com/hexsane/AMIGODOS/


rb6502:
Handled in https://github.com/mamedev/mame/commit/858987da831e2105cd28cd160ca8bd0f8fd65122 .

PR #13680, opened 2025-05-07
moog/source.cpp: Moved netlist declaration to nl_source.h.

m1macrophage:


cuavas:
The corresponding netlist source file has no copyright headers.

Due to the fact that the actual information conveyed in a netlist isn’t eligible for copyright protection (it’s considered “purely functional”), copyright protection only applies to the formatting of the text.  As such, we generally like to keep our netlists CC0 licensed to make life simple.

Can we add the CC0 license headers to this netlist?

m1macrophage:
Ah, got it. Sent #13887.

PR #13683, opened 2025-05-07
namcos12: Fix Tekken 3 release date

AnthonyRyuki:
Despite the copyright year, there are many sources that the game was released in 1997, rather than 1996.


happppp:
March 20, 1997

PR #13684, opened 2025-05-07
Add PLD dumps to Gaelco games and verify bruteforced ones [Josep Quingles, Recreativas.org]

clawgrip:


clawgrip:
> Should the remaining comment say "Bruteforced but not verified"? (and should the removed comments have said that?)

Yes, only three PLDs were confirmed with an unprotected dump.

Allofich:
Right now it says "bruteforced but verified," so that looks like a typo.

clawgrip:
> Right now it says "bruteforced but verified," so that looks like a typo.

This PLD was bruteforced, and the bruteforced dump was verified to work on the real hardware. It's the usual comment on this cases.

Allofich:
Maybe I'm missing something, but I interpreted this PR as that you verified dumps for this PR that were previously not verified (allowing removal of the BAD_DUMP flag), but the removed comments already said "Bruteforced but verified," suggesting they were already verified. So I thought there must just be have been a "not" missing in those comments, and the one remaining BAD_DUMP, is also missing a "not" there, but is that incorrect?

PR #13689, opened 2025-05-08
added a derived m6502 type for the VT3xx series sound CPU with a few basic opcode changes + preliminary hook up so that denv150 boots to something

mamehaze:


mamehaze:
this is currently based on the denv150 set, which uploads a sound program as it doesn't rely on any kind of internal ROM from the main CPU side

other sets using the sound CPU need the internal ROM on the main CPU mapping so that they upload an initial program (2 versions of this ROM, probably for different VT models, are currently loaded in some of the ROMsets, and will likely be moved to a device ROM as it seems to be standard, one for CPUs with scrambled opcodes, one for CPUs without)

mamehaze:
I'm away for the weekend, so won't be able to address anything, but if this gets merged as a checkpoint then I'll continue work on it when I resume.  There's going to be a lot of shuffling and refactoring of this code as I develop a better understanding so I'm not really worried if there's any short term breakage.

Some code could probably be better, but that will come as the bigger picture becomes more clear with how these platforms work, as these VT3xx models are basically completely undocumented so it's a case of learning as I go.

mamehaze:
Various VT3xx sets are now showing signs of life

new_250
new_barbie
new_cars
new_denv
new_dp
new_fz
new_paw
new_pj
new_sp
new_sw
denver_menu
fusion
lexi30
seal

PR #13692, opened 2025-05-10
dataeast/brkthru.cpp: Remove brkthrubl

clawgrip:
Has the same ROMs as the original World version


happppp:
In other words, that set wasn't a bootleg?
this comment: 3002A + 3002B PCBs <-- Data East PCBs

clawgrip:
> In other words, that set wasn't a bootleg? this comment: 3002A + 3002B PCBs <-- Data East PCBs

Yesss

Issue #6655, opened 2020-05-06
Debugger: Proposed new behavior for `symlist` command


Stevie-O:
The `symlist` command's current behavior makes symbol-discoverability more difficult than necessary (it certainly tripped me up). 

Currently, `symlist` with no parameters displays *only* global symbols; to view the symbols for the current CPU you have to use `symlist <cpu>`.

I propose the following new behavior:
* `symlist` with no arguments displays all currently-available symbols: globals and those associated with the current CPU.
* `symlist global` (or somesuch, I'm open to the syntax) displays global symbols only.
* `symlist -1` would also display global symbols; since `symlist` works with debugger variables (`temp9=1; symlist temp9` will display symbols for CPU 1) this would preserve that property (it would  actually make the command even more flexible, I'd think)

This offers the following benefits:
* It makes the behavior of `symlist` more convenient for users unfamiliar with the nuances of MAME's debugger -- one command to view all of the symbols available to them at once.
* It brings the behavior of `symlist` with no arguments closer to that of other commands that take an optional CPU specifier (e.g. `trace`): specifically, the behavior when no CPU is specified is to assume the current CPU/context

## Some background:

When I was trying to understand the MAME debugger's behavior, I ran into some severe under-documentation issues.  For example, there do not appear to be any symbols formally documented anywhere.

I got confused because, technically, the word "symbol" technically refers to an object's _name_ -- it's called a "symbol" because it's an identifier used to represent the actual object (here, a storage location in memory).

With that in mind, the term "global symbols" becomes ambiguous, or at least misleading.  I interpreted "global symbols" to mean "names available in all contexts" and "per-CPU symbols" to mean "names specific to a particular CPU".

I was surprised to note that there wasn't a "number of cycles executed" symbol for CPUs, when there was a "number of frames displayed" symbol.  Eventually I happened to try "print cycles" and got an output (by the way, perhaps `print` should probably output a `$` prefix, since it outputs in hex.  It was really confusing when the sidebar says "cycles = 52" and `print cycles` outputs 34.)

Anyway, when `cycles` worked but didn't show up in `symlist`, I got confused.  I located the place in the source code where the symbols were registered and got _really_ confused: https://github.com/mamedev/mame/blob/dc83375dcaaecddc3302cc9eddb0e39b17cb3f04/src/emu/debug/debugcpu.cpp#L61
[...]

PR #13694, opened 2025-05-10
debugger: symlist usability + symbol table extensibility

dave-br:
This addresses issue #6655 (symlist command usability), plus adds a bit of plumbing for future extensibility.

### Symlist command usability

The distinction between global and CPU symbols is unclear to most who don't read the source, so this makes `symlist` more natural:
- `symlist` with no arguments displays all global *and* :maincpu symbols, with clear header text for each list.  At the bottom, prints helper text to make user aware of the cpu form
- `symlist <cpu>` works as it has before, only displays symbols for specified CPU

### Symbol table extensibility

This pr includes some changes in anticipation of features that add new kinds of symbol tables.  Example: the source-level debugging pr (#13444) adds symbol tables for local and global variables that appear in the debugged source code.  The following were originally part of the source-level debugging PR, so merging this will shrink that PR somewhat.
- Add an enum field to symbol table for its 'type', for prettier printing from symlist.
- symlist now traverses symbol table chain completely.  No noticeable difference from before, but this will make extensions to the chain work automatically
- Eliminate hard-coded limit (1000) on number of symbols to print per symbol table

### Sample session

A CoCo with Speech-Sound cartridge (thus, two valid CPUs).

```
>symlist

**** CPU ':maincpu' symbols ****
a = 0
b = B
cc = 90
curflags = 90
curpc = 10C
cycles = 24  (read-only)
d = B
dp = 0
lastinstructioncycles = 18  (read-only)
logunmap = 1
pc = 10C
s = 7F1E
totalcycles = 24E077  (read-only)
u = ABEE
x = 402
y = A00E

**** Global symbols ****
beamx = 85  (read-only)
beamy = D9  (read-only)
cpunum = 0  (read-only)
frame = A1  (read-only)
temp0 = 0
temp1 = 0
temp2 = 0
temp3 = 0
temp4 = 0
temp5 = 0
temp6 = 0
temp7 = 0
temp8 = 0
temp9 = 0
wpaddr = 0  (read-only)
wpdata = 0  (read-only)
wpsize = 0  (read-only)

To view the symbols for a particular CPU, try symlist <cpu>,
where <cpu> is the ID number or tag for a CPU.


>symlist :maincpu

**** CPU ':maincpu' symbols ****
a = 0
b = B
cc = 90
curflags = 90
curpc = 10C
cycles = 24  (read-only)
d = B
dp = 0
lastinstructioncycles = 18  (read-only)
logunmap = 1
pc = 10C
s = 7F1E
totalcycles = 24E077  (read-only)
u = ABEE
x = 402
y = A00E


>symlist 0

**** CPU ':maincpu' symbols ****
a = 0
b = B
cc = 90
curflags = 90
curpc = 10C
cycles = 24  (read-only)
d = B
dp = 0
lastinstructioncycles = 18  (read-only)
logunmap = 1
pc = 10C
s = 7F1E
totalcycles = 24E077  (read-only)
u = ABEE
x = 402
y = A00E


>symlist 1

**** CPU ':ext:multi:slot3:ssc:pic7040' symbols ****
a = FF
b = 0
curflags = D0
curpc = F2D1
cycles = 0  (read-only)
genpc = F2D1
lastinstructioncycles = 9  (read-only)
logunmap = 1
pc = F2D1
sp = 4C
st = D0
totalcycles = 24E062  (read-only)


>symlist 2
Invalid CPU index 2
```

### DIFF
--- a/src/emu/debug/debugcmd.cpp
+++ b/src/emu/debug/debugcmd.cpp
@@ -3909,58 +3909,80 @@ void debugger_commands::execute_memdump(const std::vector<std::string_view> &par
 
 void debugger_commands::execute_symlist(const std::vector<std::string_view> &params)
 {
-	const char *namelist[1000];
+	// Default to CPU "0" if none specified
+	const char * cpuname = (params.empty()) ? "0" : params[0].cbegin();
+	device_t *cpu = nullptr;
 	symbol_table *symtable;
-	int count = 0;
-
-	if (!params.empty())
+	if (!m_console.validate_cpu_parameter(cpuname, cpu))
 	{
-		// validate parameters
-		device_t *cpu;
-		if (!m_console.validate_cpu_parameter(params[0], cpu))
-			return;
-		symtable = &cpu->debug()->symtable();
-		m_console.printf("CPU '%s' symbols:\n", cpu->tag());
+		if (!params.empty())
+			return;     // Explicitly specified cpu is invalid
+
+		// Somehow cpu "0" is invalid, so just stick with global symbol table
+		symtable = &m_machine.debugger().cpu().global_symtable();
 	}
 	else
 	{
-		symtable = &m_machine.debugger().cpu().global_symtable();
-		m_console.printf("Global symbols:\n");
+		symtable = &cpu->debug()->symtable();
 	}
 
-	// gather names for all symbols
-	for (auto &entry : symtable->entries())
+	// Traverse symbol_table parent chain, printing each table's symbols in its own block
+	for (; symtable != nullptr; symtable = symtable->parent())
 	{
-		// only display "register" type symbols
-		if (!entry.second->is_function())
+		// Skip globals if user explicitly requested CPU
+		if (symtable->type() == symbol_table::BUILTIN_GLOBALS && !params.empty())
+			continue;
+
+		if (symtable->entries().size() == 0)
+			continue;
+
+		std::vector<const char *> namelist;
+
+		// Print heading for table
+		switch (symtable->type())
 		{
-			namelist[count++] = entry.second->name();
-			if (count >= std::size(namelist))
-				break;
+		case symbol_table::CPU_STATE:
+			m_console.printf("\n**** CPU '%s' symbols ****\n", cpu->tag());
+			break;
+		case symbol_table::BUILTIN_GLOBALS:
+			m_console.printf("\n**** Global symbols ****\n");
+			break;
+		default:
+			assert (!"Unrecognized symbol table type");
+		}
+
+		// gather names for all relevant symbols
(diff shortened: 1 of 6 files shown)


dave-br:
Just a gentle bump... merging this fixes an issue (i.e., improves symlist usability), and helps to shed some more code off of my larger source-debugging PR.  Happy to answer any questions.

src/emu/debug/debugcmd.cpp
```diff
@@ -3909,58 +3909,80 @@ void debugger_commands::execute_memdump(const std::vector<std::string_view> &par
 
 void debugger_commands::execute_symlist(const std::vector<std::string_view> &params)
 {
-	const char *namelist[1000];
+	// Default to CPU "0" if none specified
+	const char * cpuname = (params.empty()) ? "0" : params[0].cbegin();
```
cuavas, line 3913: This makes no sense at all.  Why wouldn’t you default to the CPU that’s **currently focused** in the debugger rather than the first CPU?  That’s what all the other commands do (`bpset`, `wpset`, etc.).  Always defaulting to the first CPU a very poor choice.

Also, as I’ve said elsewhere, the purpose a symbol table is being used for is not a characteristic of the symbol table, and shouldn’t be a property of the symbol table.  Requiring a model object to know all the situations in which it will be used is broken design.

dave-br: @cuavas:  Sorry about that CPU thing, and thanks for catching it.  I opened a PR to fix (https://github.com/mamedev/mame/pull/13886).

Regarding the symbol table type enum, I mentioned a follow-up suggestion on https://github.com/mamedev/mame/pull/13733, but hadn't heard back yet.  The gist was to allow the symlist command to provide helpful category headers in its output as future symbol tables get added to the parent chain, such as local and global variables from the source code for source-level debugging. The goal is to categorize **what** the symbol table contains, but not exactly where or how it's used.

This allows symlist to be more data-driven, which leads to simple code even when there are 4 symbol tables chained together...

```
	// Traverse symbol_table parent chain, printing each table's symbols in its own block
	for (; symtable != nullptr; symtable = symtable->parent())
	{
		switch (symtable->type())
		{
			// Print symbol table heading like "**** CPU '%s' symbols ****"
			// or "**** Source-level local variables ****", etc.
		}

		// Now print contents of symbol table
	}
```

Alternatively, if symbol tables cannot know anything about their contents, then rather than looping through the symbol table parent chain, symlist would need to be hard-coded to obtain each symbol table from a known accessor / owner object so it could reliably print the descriptive header before it.  This would require ongoing maintenance to the symlist code if new symbol tables are added to the parent chain in the future.  I was hoping to avoid this, but if you think it's cleaner this way, I can do it.

What do you think?

PR #13705, opened 2025-05-15
various VT3xx (mostly VT369) work + some other VT refactoring

mamehaze:
- added a derived 6502 core for the VT3xx Sound CPU (built into the SoC), it has some customized opcodes
- added enough emulation of the Sound CPU to get denv150 to pass initial checks
- added internal SoC ROM to a number of sets that copy the sound program from it, allowing those to pass startup checks too
- added preliminary support for some of the extra VT3xx background video features, note even the way the hardware implements these is ugly, they really are hacked on top of the standard NES modes
- added preliminary support for extra VT3xx sprite modes
- the above also allow various Lexibook sets to show boot screens eg. lxcmcysw, lxcmcyfz, lxcmcydp, lxcmcysp, lxcmcycr, lxcmcypj, lxcmcyba, lxcmcypp
- paired some scrambling/encryption to specific SoC classes, as it's likely these represent slightly customized models, added modes for nubsupmf, red5mam, dgun2593, gcs2mgp, 240in1ar which also now show something
- started looking at the new DMA modes needed by 240in1ar etc. (disabled for now as it causes issues)
- denv150 basically now works, but leaving it marked as NOT WORKING until the rest of the sound CPU logic is hooked up
- flagged a few sets as potentially bad dumps
- verified / corrected metadata for a number of sets now that they boot and can be better identified
- identified some cases where there seem to be additional protection devices present, supplying either code or data
- misc other VT related tweaks


mamehaze:
gcs2mgp is one of the ones driving a lower resolution vertically orientated LCD, required more address bits

mamehaze:
With preliminary V3xx sprite modes and some other fixes many of these now render proper menus.  They still fail to go into the games (which I'll need to investigate) but it's progress.

PR #13707, opened 2025-05-16
Fix tdv2100 display edge-cases

Frodevan:
I had a closer look at the schematics and a few of the jumper-settings in the manual, and realized a few things:

1. A handful of jumpers had their polarity wrong, leading to wrong default settings
2. The latching of the attribute happens on the rising edge of the character-clock, i.e. active with the character immediately following the change

### DIFF
--- a/src/mame/tandberg/tdv2100_disp_logic.cpp
+++ b/src/mame/tandberg/tdv2100_disp_logic.cpp
@@ -489,15 +489,14 @@ void tandberg_tdv2100_disp_logic_device::update_attribute(int at_row, int at_col
 				{
 					m_attribute = 0;
 				}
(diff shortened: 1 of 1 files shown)


Frodevan:
Update:
I did check on original hardware, and there was a line I did not catch in the schematics which does indeed blank attribute effects for attribute changes. So while the code changes related to point 2 is more accurate here timing-wise, there should be no visible differences.

cuavas:
Please put spaces around binary operators (e.g. `(foo & 0x03)` rather than `(foo&0x03)`) to make stuff more readable, and use the `BIT` helper when testing individual bits.

Frodevan:
Spacing has been adjusted

PR #13708, opened 2025-05-17
namcos21: clear screen to palette index 0 instead of 0xff

FlykeSpice:
This makes the sky background from Winning Run games to blend more seamless with the fading horizon (specially winrungp and winrun91), this is just a duct-tape fix and the actual clear screen behavior from the hardware needs to be studied.


angelosa:
image

I tested `solvalou` / `winrungp` / `cybsled`, there are some extra flags in the unused section of palette RAM (that is RGB888). I don't think it's alpha, more like bit selects for vblank / framebuffer erase stuff?

angelosa (approved):
Won't break anything else because `winrun` and friends uses their own `screen_update` compared to the other variants. Adding a note for unknown bits 2-1 in unused paletteram is a nice to have.

PR #13709, opened 2025-05-17
Adjust PV-1000 screen stretch to show full game area by default

KashtPR:
Set the PV-1000 default screen scaling to:

    set_default_position(1.0, 0, 6283.0/5724.0, 0)

Previously, the default scaling caused games to appear cropped on both the left and right edges, requiring users to manually adjust stretch settings to see the full image. I understand that the change was done this way to simulate how it would look on an old CRT TV. With modern displays and monitors, real hardware can, and most of the time will show the full game area.

This change ensures the entire game area is visible by default, resulting in a much better out-of-the-box user experience while preserving visual accuracy.

The vertical stretch of ~1.097659 matches the original PAR/DAR correction derived from:
    (216 / 206.0) / (244 / 212.0) = 6283 / 5724

This maintains a 4:3 display shape with a pixel aspect ratio of 48:35, per lidnariq's analysis.

Attached are visuals of a couple games with current settings and with fixed default settings.
Guntus_MAME0277
Guntus_withFix
Turpin_MAME0277
Turpin_withFix
        
As noted earlier, this fix not only preserves emulation accuracy 100%, but also provides everyone trying the Casio PV-1000 in MAME the best out-of-the-box experience in the modern TVs and monitors today, with absolutely no changes to sliders or configuration files.

### DIFF
--- a/src/mame/casio/pv1000.cpp
+++ b/src/mame/casio/pv1000.cpp
@@ -554,10 +554,11 @@ void pv1000_state::pv1000(machine_config &config)
 	// Display aspect is MAME's 4:3 default.
 
 	// Note that this value is overridden by the user's pv1000.cfg, if present.
-	// 206px x 48/35(PAR) / 4/3(DAR) = 212sl
+	// To allow full game display by default while preserving pixel aspect and display aspect, setting Screen Horizontal Stretch to 1.0, Screen Vertical Display to 1.097659
+	// This matches previous ratio established by lidnariq based on 206px x 48/35(PAR) / 4/3(DAR) = 212sl, or m_screen->set_default_position(216/206.0, 0, 244/212.0, 0)
 	m_screen->set_default_position(
-			216/206.0, 0, //216 px in storage aspect; cropped to 206 px
-			244/212.0, 0); //244 sl in storage aspect; cropped to 212 sl
+			1.0, 0,
+			6283.0/5724.0, 0); 
 	m_screen->set_screen_update(FUNC(pv1000_state::screen_update_pv1000));
 	m_screen->set_palette(m_palette);
 


cuavas:
The software you’re demonstrating this with was developed decades after the machine was obsolete, likely using emulators.

The current raster position values were set in #12570 by @lidnariq, supposedly based on the part of the display that’s hidden in the overscan area when using real hardware with a correctly adjusted NTSC-J monitor.

Do you have any evidence that @lidnariq was wrong about how the actual system behaves?  My gut feel is that @lidnariq is correct, and these games would be cropped if you ran them on real hardware.

KashtPR:
> The software you’re demonstrating this with was developed decades after the machine was obsolete, likely using emulators.
> 
> The current raster position values were set in #12570 by @lidnariq, supposedly based on the part of the display that’s hidden in the overscan area when using real hardware with a correctly adjusted NTSC-J monitor.
> 
> Do you have any evidence that @lidnariq was wrong about how the actual system behaves? My gut feel is that @lidnariq is correct, and these games would be cropped if you ran them on real hardware.

Let me clarify, the software I used is MAME 0277, the latest and greatest version. I did not use the ePV1000 emulator or anything else.

Also, I am not saying @lidnariq is wrong. The total opposite, he is 100% correct. However, different monitors behave differently, some show more of that overscan area than others. In fact, all of Inufuto's homebrew games, which are tested in real hardware and captured with NTSC-J AV, show the entire games as evidenced in his Youtube channel. Here is an example of his latest homebrew for the PV-1000: https://www.youtube.com/watch?v=Kue9yF4rsjc

I don't think the assumption that every monitor used to play these games with real hardware will behave exactly the same. Hence, this fix basically sets the horizontal stretch to 1, so everything is displayed, while adjusting the vertical stretch at the exact ratio.

Another way to achieve the same thing is to add this in the pv1000.cfg:

<screen index="0" vstretch="1.097659" />
     
This will do exactly the same. With the current MAME code, the pv1000.cfg has this:

<screen index="0" hstretch="1.048544" vstretch="1.150943" />

If you divide those numbers, 1.150943/1.048544, you get the 1.097659, which is exactly (216 / 206.0) / (244 / 212.0). The way it is written in my change is preserving perfect geometry.

So to reiterate, the change I proposed was mostly to provide a better out of the box experience to MAME users wanting to play Casio PV-1000 games. I went through the experience of seeing all of Inufuto's games played in real hardware shared in his Youtube videos and trying to get them to show up the exact same way as in the videos in MAME was impossible. I had to go and really figure out this ratio in order to get the right setting that would display the entire game while preserving 100% the screen's geometry and pixel ratio. Anyone who loads any of the PV-1000 games into MAME today and wants to get the entire game area to show up, as soon as they move a slider, there is no guarantee that they get it right, as you cannot set it back to exactly 1.0 for either horizontal or vertical stretch, and definitely cant set it to 1.097659. Unless people know about this specific ratio, they can adjust it to where they think its perfect, but it is not.

I learned a lot about the PV-1000 after having some back and forth chats with @lidnariq , and yes, his code is 100% correct in every way. Where I disagree is with the assumption that ALL monitors and TVs behave the same, therefore recommended this change.

KashtPR:
This is easily reproduceable. 

1. Download latest version of MAME.
2. Run: mame.exe pv1000 (select whichever game you want)  I recommend Turpin and Guntus, so you can do a direct comparison.
3. Close the game.
4. Go to the cfg directory and open the pv1000.cfg. You will see: screen index ="0" hstretch="1.048544" vstretch="1.150943". These numbers are the ratios defined by 216 / 206 and 244 / 212.0, respectively.

You will see the games exactly the same way in those screenshots.

I am fine with keeping the code unchanged, I now know the exact ratio I need to be able to see the entire game area while preserving exact geometry, so I can make the config change and call it a day. Other users trying these games for the first time may wonder exactly the same thing I did, why can't I see Inufuto's games like he is showing them on his Youtube channel in real hardware. The benefit of the change is to provide a better out of the box experience for other people trying these games out. I know I struggled to get the ratio correct, and without looking at the code, I wouldn't have figured it out.

rb6502:
Going on precedent, I think this could be done as a configuration switch, with the current cropped behavior as the default.  While televisions all had overscan, consumer NTSC-J and NTSC-U/C monitors were widely available which could easily be adjusted to show the full frame.  Many developers used Commodore (rebranded Phillips OEM) 1084 or similar monitors to develop console games until the end of NTSC for that reason.

We arguably should do something similar for NES/Famicom as well.  Commercial software for those systems relied heavily on overscan to hide flaws near the edges.  But the standard for emulators has always been to show the full frame for whatever reason.

KashtPR:
> But the standard for emulators has always been to show the full frame for whatever reason.

That is what I thought. My experience with other systems within MAME was that the full frame is displayed for pretty much all I have checked out. There are very few that I found where they aren't displayed in full frame by default, haven't really checked a lot of the drivers' code but I wonder if it is common for other drivers to hardcode these stretch values.

Since there were initial doubts regarding what I used to take the first screenshots, here is a full video capture of me installing the latest MAME, running both games with latest MAME, then running them with the changed code. Here is the link: https://youtu.be/kKQP2TP_znI

I think the expectation from many, especially the new generation of MAME users, is that they want the full frame. The newer generations have no idea what a CRT is, yet alone overscan, unfortunately.

cuavas:
> Going on precedent, I think this could be done as a configuration switch, with the current cropped behavior as the default.

You can already adjust it in the sliders menu.

simzy39:
From what I understand, lidnariq said to you, "I stand by the MAME defaults as being accurate. I don't want it changed."
https://forums.nesdev.org/viewtopic.php?t=24671&sid=db28f889e58fb0ddd9625d8cd49f42a3&start=15

KashtPR:
> From what I understand, lidnariq said to you, "I stand by the MAME defaults as being accurate. I don't want it changed." https://forums.nesdev.org/viewtopic.php?t=24671&sid=db28f889e58fb0ddd9625d8cd49f42a3&start=15

Yes, and I am pretty sure he was referring to our discussion about the 28 tiles vs 30 that I was referring to, which I was basing on wrong information. I later rechecked all the info in his wiki and the driver and everything is correct. As I have mentioned several times in here, lidnariq's code for the driver is correct, but I disagree with forcing some horizontal and vertical stretching values based on an assumption that it is not 100% accurate in all TVs and monitors.

Try playing any of the games for the PV-1000, the area cropped due to the default stretching in the driver is real game data, not garbage data as it happens with other games in other systems.

As I mentioned earlier, I now know exactly what value should go in the pv1000.cfg in order to get the entire frame while preserving pixel and display aspect ratios. The change I proposed here is for the default value to show the entire frame while leaving everything intact, a change that can be altered via a config file.

If the MAME standard is to show the entire frame for games in their respective systems, then the default should do the same, and people wanting to cut the screen to simulate the overscan can use the sliders to do this. If not, then leave as is, but no matter how much you manually play with the sliders you will not be able to preserve perfect geometry and show the entire frame, so new users who have no idea about any of this will always experience games cropped.

lidnariq:
> Yes, and I am pretty sure he was referring to our discussion about the 28 tiles vs 30 that I was referring to

Absolutely not.

> I disagree with forcing some horizontal and vertical stretching values based on an assumption that it is not 100% accurate in all TVs and monitors.

I guarantee that exactly 0% of all consumer TVs, explicitly not including studio monitors, explicitly excluding people who modified their TVs in their service menus, could display the so-called "96% NTSC" field that you're asking for here.

The last manufactured SDTV CRT TVs basically ended at 92%, cropping roughly 10 native pixels in total divided between the left and right. TVs in 1983, when the PV-1000 came out, very likely cropped more. Any TV that could display the entire 224 pixel PV-1000 output would have an enormous highly-visible border on the NES & SMS (13 extra pixels!), and Megadrive (16).

Note that we are _not_ talking about arcade systems here. There are no calibration screens to help the the arcade owner to adjust. Everyone had the TV that they had, and game designers had just darn well better not put any thing in the sides and corners. This continues to be a problem with homebrew for all systems, where people are focusing on the emulator experience and pretending that the display technology of the time isn't a factor.

KashtPR:
Thanks @lidnariq for replying. I did not think this was going to cause this much drama. I do not think it is fair to isolate the issue to only homebrew development as it is clear that whoever wrote Turpin back in 1983 wanted to use a bit more of the screen than what the driver is cropping. The game has a square stage with red, white, and blue lines, and the cropping in the current driver hides the white and red lines on the left side. Why would the developer back then not position the stage better to account for that cropping? There are other examples in the stock games that are less evident but show exactly this, the games developed for this system during 1983 wanted to use as much of the screen as possible, and yes, the technology back then probably worked against them. 

So yes, I understand what you are doing in the stretching parameters in the driver. You are getting it as close to what the consumer TVs would allow back then, forcing that cropping at the driver level. And I am not saying it is wrong. 

Yes, I was focusing on emulator experience, it is 2025 and technology has advanced tremendously. I thought that is what MAME was about, achieving perfect emulation for all these very old consoles while bringing a great experience to its users. Then why not have the games display at their best by default without really messing any of the emulation code or geometry that you worked very hard to figure out? The change I propose is not impacting any of the hardware emulation, it only impacts how a user first sees the game when loaded. 

My apologies if I offended any of the participants in here. I will give this PR a rest now.

cuavas:
> I disagree with forcing some horizontal and vertical stretching values based on an assumption that it is not 100% accurate in all TVs and monitors.

There are only two kinds of displays that matter here, though: NTSC-J TV sets used by typical users and software developers, and video monitors used by hardware designers.  There’s a standard adjustment for overscan.  You don’t have to adjust horizontal and vertical size when switching programme sources.  Any TV service centre would adjust the horizontal and vertical size using a test pattern generator.

> If the MAME standard is to show the entire frame for games in their respective systems, then the default should do the same, and people wanting to cut the screen to simulate the overscan can use the sliders to do this.

Not really – consider Cidelsa arcade hardware Midway Astrocade, Sharp X1, and Sony SMC777.  In all those cases, the default is to stretch the overscan area out of the frame.

> no matter how much you manually play with the sliders you will not be able to preserve perfect geometry and show the entire frame, so new users who have no idea about any of this will always experience games cropped, or distorted, or both.

Using the sliders in MAME is no worse than twiddling knobs on a real TV.  You still won’t get a “perfect” image without using a test pattern.

PR #13710, opened 2025-05-17
roland_tr707.cpp: Extended driver to encompass most hardware other than audio and LCD.

m1macrophage:


happppp:
FYI pwm_display_device has separate functions for both sides of the matrix, so you usually don't need to keep track of those 2 variables, like this:
https://github.com/mamedev/mame/commit/f6f8f93830e712601691568df895dc04366ac615

A lot of my old drivers still unnecessarily call matrix(), it's because I added write_my/mx to the device much later.

Is the va_eg change a workaround for @galibert sound update sync timing issue?
Or do you(OG) plan to look further into it to see if there is an underlying issue that can be fixed?

galibert:
Yes, I want to fix it.

  OG.

m1macrophage:
Ah, I see. I fixed the matrix thing. I will send a PR to fix that in my other drivers.

The `va_eg` change is not related to the update timing issue. It is new functionality to predict when a voltage will reach a certain value. This is useful for emulating comparators connected to the RC network, such as in this driver.

PR #13714, opened 2025-05-17
cirsa/missbamby.cpp: Add Arfyc 3 Mini

clawgrip:
3 Mini [Javier Blanco, Recreativas.org]


cuavas:
Why `3mini` rather than `mini3` like the name seems to be?  Also the list at the top says “3 Mini” while the description says “Mini 3”.

clawgrip:
> Why `3mini` rather than `mini3` like the name seems to be? Also the list at the top says “3 Mini” while the description says “Mini 3”.

When looking on the legal registry for the year, we found it registered as 3 Mini. They registered three machines on the same batch, Arfyc 3, Arfyc 3 Deluxe and Arfyc 3 Mini.

PR #13715, opened 2025-05-17
cpu/mb86235: Improve TGPx4 interpreter

gm-matthew:
Most Sega Model 2C-CRX games are now working with 3D graphics. Top Skater still doesn't display properly because it uses a secondary transformation matrix which is not implemented in the HLE geometry engine; proper emulation of the geometry DSP is required.


gm-matthew:
Is it possible to restart the checks? If not, I'll close this pull request and open a new one

Gurudumps:
Just close and re-submit. It's faster than waiting. In other words, why wait for some random to fix something when you can fix it yourself immediately ;-)

gm-matthew:
Closing this pull request so I can resubmit

PR #13718, opened 2025-05-18
cpu/mb86235: Improve TGPx4 interpreter

gm-matthew:
Most Sega Model 2C-CRX games are now working with 3D graphics. Top Skater still doesn't display properly because it uses a secondary transformation matrix which is not implemented in the HLE geometry engine; proper emulation of the geometry DSP is required.

Resubmitting because the previous pull request failed validation due to unrelated commits that have since been fixed. Hopefully it should pass this time

### DIFF
--- a/src/devices/cpu/mb86235/mb86235.cpp
+++ b/src/devices/cpu/mb86235/mb86235.cpp
@@ -1,22 +1,17 @@
 // license:BSD-3-Clause
-// copyright-holders:Angelo Salese, ElSemi, Ville Linde
+// copyright-holders:Angelo Salese, ElSemi, Ville Linde, Matthew Daniels
 /********************************************************************************
  *
  * MB86235 "TGPx4" (c) Fujitsu
  *
- * Written by Angelo Salese & ElSemi
+ * Written by Angelo Salese, ElSemi & Matthew Daniels
  *
  * TODO:
  * - rewrite ALU integer/floating point functions, use templates etc;
- * - move post-opcodes outside of the execute_op() function, like increment_prp()
- *    (needed if opcode uses fifo in/out);
  * - rewrite fifo hookups, remove them from the actual opcodes.
- * - use a phase system for the execution, like do_alu() being separated
- *    from control();
  * - illegal delay slots unsupported, and no idea about what is supposed to happen;
- * - externalize PDR / DDR (error LED flags on Model 2);
- * - instruction cycles;
- * - pipeline (four instruction executed per cycle!)
+ * - externalize PDR (error LED flags on Model 2);
+ * - perform external memory access if AALU-A is 0x400 or higher
  *
  ********************************************************************************/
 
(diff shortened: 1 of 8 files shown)


angelosa:
Failure is unrelated to this, I'm on it.

src/devices/cpu/mb86235/mb86235.cpp
```diff
@@ -1,22 +1,17 @@
 // license:BSD-3-Clause
-// copyright-holders:Angelo Salese, ElSemi, Ville Linde
+// copyright-holders:Angelo Salese, ElSemi, Ville Linde, Matthew Daniels
 /********************************************************************************
  *
  * MB86235 "TGPx4" (c) Fujitsu
  *
- * Written by Angelo Salese & ElSemi
+ * Written by Angelo Salese, ElSemi & Matthew Daniels
```
angelosa, line 7: It's suggested to just use `copyright-holders`, no need to repeat in header (this is a pre-license remnant, where we didn't have SPDX identifiers)

src/devices/cpu/mb86235/mb86235d.cpp
```diff
@@ -261,32 +261,32 @@ offs_t mb86235_disassembler::dasm_control(std::ostream &stream, uint32_t pc, uin
...
-			util::stream_format(stream, "SETM #%01X, CBSB", (ef2 >> 8) & 7);
+			util::stream_format(stream, "SETM #%Xh, CBSB", (ef2 >> 8) & 7);
 			break;
 		case 0x0b:
-			util::stream_format(stream, "SETM #%d, RF", (ef2 >> 7) & 1);
+			util::stream_format(stream, "SETM #%Xh, RF", (ef2 >> 7) & 1);
 			break;
 		case 0x0c:
-			util::stream_format(stream, "SETM #%d, RDY", (ef2 >> 4) & 1);
+			util::stream_format(stream, "SETM #%Xh, RDY", (ef2 >> 4) & 1);
```
cuavas, line 276: Why the inconsistent `H` and `h` suffixes?  Is there actually a difference?

angelosa: image

happppp: That picture is related to "h" vs uppercase "H"? It's probably just a typo.

gm-matthew: Chapter 8 of the MB86235 documentation lists the assembly example for `SETM` with an uppercase `H` suffix when setting all fields, but when setting only one specific field the examples are listed with lowercase `h` suffixes; I do not know the reason for this difference.

If consistency is preferred we could certainly use uppercase `H` suffixes for all uses of `SETM`.

cuavas: Oh, so it’s actually preserved from the original documentation.

PR #13719, opened 2025-05-18
Link to a to-do list / problems in Mame's pv1000 emulation

simzy39:
This just links to a list of what's wrong with Mame's emulation

### DIFF
--- a/src/mame/casio/pv1000.cpp
+++ b/src/mame/casio/pv1000.cpp
@@ -4,6 +4,9 @@
 
     Driver for Casio PV-1000
 
+TODO: This lists problems with Mame's emulation
+      https://obscure.nesdev.org/wiki/Casio_PV-1000/Emulation
+
 ***************************************************************************/
 
 #include "emu.h"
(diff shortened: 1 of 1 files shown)


angelosa:
I don't think @lidnariq needs a reminder himself wrote ...

simzy39:
It's for the next person in the future so they know what it's missing

Gurudumps:
It's still written as if it's external by someone with a grudge against MAME.
Re-write it from the MAME driver point of view....

TODO:
- fix joystick responding incorrectly
- correct RAM allocation [RAM size should be xxxx at memory locations 0xyyyy, 0xzzzz etc]
- raster effects are missing
- support more than two pages ($0000 and $2000) of pattern ROM
- support moving the tilemap address
- support moving pattern ram
- support banking the pattern data over the entire memory region not just over the entire cartridge

cuavas:
It’s not helpful for people who don’t really understand what’s going on properly to try and maintain source TODO lists.

PR #13727, opened 2025-05-20
various VT3xx (mostly VT369) work + some other VT refactoring, many sets now boot + games selectable

mamehaze:
(squashed version of https://github.com/mamedev/mame/pull/13705 to use as  a new base)

- added a derived 6502 core for the VT3xx Sound CPU (built into the SoC), it has some customized opcodes
- added enough emulation of the Sound CPU to get denv150 to pass initial checks
- added internal SoC ROM to a number of sets that copy the sound program from it, allowing those to pass startup checks too
- added preliminary support for some of the extra VT3xx background video features, note even the way the hardware implements these is ugly, they really are hacked on top of the standard NES modes
- added preliminary support for extra VT3xx sprite modes
- the above also allow various Lexibook sets to show boot screens eg. lxcmcysw, lxcmcyfz, lxcmcydp, lxcmcysp, lxcmcycr, lxcmcypj, lxcmcyba, lxcmcypp
- paired some scrambling/encryption to specific SoC classes, as it's likely these represent slightly customized models, added modes for nubsupmf, red5mam, dgun2593, gcs2mgp, 240in1ar which also now show something
- started looking at the new DMA modes needed by 240in1ar etc. 
- denv150 basically now works, but leaving it marked as NOT WORKING until the rest of the sound CPU logic is hooked up
- flagged a few sets as potentially bad dumps
- verified / corrected metadata for a number of sets now that they boot and can be better identified
- identified some cases where there seem to be additional protection devices present, supplying either code or data
- replaced vt1682 ALU in vt3xx emulation with its own implementation as it isn't the same
- added relative offset support (a kind of banking internal to the VT3xx) allowing the majority of games in the multi-game menus to go ingame
- added external banking support to a number of the larger games (anything larger than 32Mbytes has to be externally banked)
- corrected ROM loading for a number of sets
- fixed additional opcodes on sound CPU, and added the remaining peripherals
- misc other VT related tweaks


mamehaze:
https://www.youtube.com/watch?v=b3ZdMv4q5Bs  is an example of what now boots

mamehaze:
There's more work / research I'd like to do here, and definitely some larger scale refactoring + renaming I'd like to do, but I'm going to make this a checkpoint before it becomes unmanageable in a single pull request.

I'm not aware of any regressions with the current code (if anything some older bugs likely got fixed, eg menu navigation in unkra200 which broke in a much older refactor.)

PR #13728, opened 2025-05-21
cpu/z80: Renamed LUT fields

holub:


holub:
doesn't make sense with #13729

PR #13729, opened 2025-05-21
cpu/z80: Handled flags lazyly

holub:
Credits to @smf- for the idea!

Before:
```
» repeat 3 ./mame zexall -bench 100 && sleep 1
Average speed: 9008.28% (99 seconds)
Average speed: 9315.16% (99 seconds)
Average speed: 9412.46% (99 seconds)
-------------------------------------------------------------------------
» repeat 3 ./mame docastle -bench 100 && sleep 1
Average speed: 1529.40% (99 seconds)
Average speed: 1539.67% (99 seconds)
Average speed: 1538.97% (99 seconds)
-------------------------------------------------------------------------
» repeat 3 ./mame pacman -bench 100 && sleep 1
Average speed: 14265.06% (99 seconds)
Average speed: 13265.54% (99 seconds)
Average speed: 13061.40% (99 seconds)
```
After:
```
» repeat 3 ./mame zexall -bench 100 && sleep 1
Average speed: 10282.77% (99 seconds)
Average speed: 10319.55% (99 seconds)
Average speed: 9790.69% (99 seconds)
-------------------------------------------------------------------------
» repeat 3 ./mame docastle -bench 100 && sleep 1
Average speed: 1556.15% (99 seconds)
Average speed: 1537.15% (99 seconds)
Average speed: 1557.02% (99 seconds)
-------------------------------------------------------------------------
» repeat 3 ./mame pacman -bench 100 && sleep 1
Average speed: 14055.56% (99 seconds)
Average speed: 13918.92% (99 seconds)
Average speed: 14550.54% (99 seconds)
```

_P.S. I feel I'm getting blind while trying to review it visually. Will do couple more attempts later myself.
Main flags tests are passing `Timing_Tests-48k_v1.0` and Raxsoft's `z80full`_

### DIFF
--- a/src/devices/cpu/z80/r800.cpp
+++ b/src/devices/cpu/z80/r800.cpp
@@ -69,7 +69,7 @@ u8 r800_device::r800_sll(u8 value)
 {
 	const u8 c = (value & 0x80) ? CF : 0;
 	const u8 res = u8(value << 1);
-	set_f(SZP[res] | c);
+	set_f(SZYXP[res] | c);
 	return res;
 }
 
--- a/src/devices/cpu/z80/r800.cpp
+++ b/src/devices/cpu/z80/r800.cpp
@@ -67,19 +67,25 @@ void r800_device::device_validity_check(validity_checker &valid) const
  ***************************************************************/
 u8 r800_device::r800_sll(u8 value)
 {
-	const u8 c = (value & 0x80) ? CF : 0;
 	const u8 res = u8(value << 1);
-	set_f(SZYXP[res] | c);
+	{
+		set_f(SZYXP[res], SF | ZF | YXF | PF);
+		set_0_f(HF | NF);
+		m_f.carry = value & 0x80;
+	}
+
 	return res;
 }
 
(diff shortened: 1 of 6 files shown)


holub:
Please check if I properly undestand m_af state handling for import/export/debugger purpose.
With this change F part of it stays unsynced with lazy state unless it's really necessary.

Do I need to add lazy flags to save state or import/export implies it's properly handled through m_af updates?

happppp:
It's faster even though there are 7 ifs for each set_f?
Is the mask param always known at compile time? If so, what if you do template<u8 Mask> set_f(u8 f)

holub:
> It's faster even though there are 7 ifs for each set_f? Is the mask param always known at compile time? If so, what if you do `template<u8 Mask> set_f(u8 f)`

I've tried templates here. Saw no advantage. Presumably as we use constants anyway compiler is smart enough to optimize it properly and get rid of unnecessary branches.

holub:
The further optimization would be not to set flags, but values which can be calculated on get. Similar to yx. That could give real improvement.
But that's on todo.

PR #13733, opened 2025-05-23
debugger: more granular naming for symbol table types

dave-br:
This is to address the naming feedback from https://github.com/mamedev/mame/commit/b39e6846221f958ef10d02cb16f5c35221f0fb2b.  That commit added an enum to identify the type of each symbol table.  This PR adds a few enum values so that symbol tables can be more granularly typed.  With the PR, the list is:

```
        CPU_STATE,         // CPU registers, etc.
        DEBUGGER_GLOBALS,  // Debugger global symbol table (e.g., beamx, beamy, frame, etc.)
        CHEAT_ENTRY,       // symbols used in cheat entry actions (argindex, temp variables)
        CHEAT_MANAGER,     // symbols from the cheat manager (frame value, from/tobcd functions)
        LUA_SCRIPT,        // custom symbols added by a LUA script for use from the LUA script
```

This reduces the set of symbols confusingly identified as global.  But the debugger's global symbol table is still named as such (for now) because that’s how the public symlist docs, preexisting function names, and code comments have been referring to it.  If you’d like those changed, that’s totally fine, let me know what you’d like.  A few candidates: DEBUGGER_GENERAL, DEBUGGER_COMMON, DEBUGGER_BUILTIN.

### DIFF
--- a/src/emu/debug/debugcmd.cpp
+++ b/src/emu/debug/debugcmd.cpp
@@ -3930,7 +3930,7 @@ void debugger_commands::execute_symlist(const std::vector<std::string_view> &par
 	for (; symtable != nullptr; symtable = symtable->parent())
 	{
 		// Skip globals if user explicitly requested CPU
-		if (symtable->type() == symbol_table::BUILTIN_GLOBALS && !params.empty())
+		if (symtable->type() == symbol_table::DEBUGGER_GLOBALS && !params.empty())
 			continue;
 
 		if (symtable->entries().size() == 0)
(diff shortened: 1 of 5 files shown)


dave-br:
@cuavas  I've modified this PR with a different approach to consider.  Still want to allow the symlist command to provide helpful category headers in its output in a way that easily allows for adding future symbol table instances into the parent chain (e.g., local and global variables from the source code for source-level debugging):
* Remove table_type enum & field altogether
* Replace with description (metadata) string, but decoupled from symbol_table type
* Add described_symbol_table subclass of symbol_table to contain description string.  (I think a wrapper class would be cleaner, but is incompatible with the symbol table parent chain unless we did parallel lists of symbol_tables & descriptions.)

Is this better or worse?  Any changes you'd like to see?

PR #13734, opened 2025-05-23
Fix dip switches for Mahjong Crystal Gal

buffis:
Some fixes for dips for Mahjong Crystal Gal.
I have the papers for "crystalg" and the DIP definition matches the ones of "crystal2" which is pretty much a clone of that game.

Papers: https://imgur.com/a/TIoM8DT

- Added correct DIPs for crystalg
- Replaced "Unknown" to "Unused" where papers state they are unused
- Changed default difficulty to 7 (specified in papers)
- Changed Demo Sound default to off (specified in papers)


angelosa:
> I'll defer to Vas about merging this, but I'll note that in MAME traditionally attract mode sounds are set on even if the manual states the default is off. This to make sound regressions testing more easy.

Yeah, that's the reason. Please add a note stating that manual suggests using default off but we need the on state by default.

cuavas:
Please don’t use “unused” unless you’ve confirmed from the code that they’re unused.  Many games actually use these DIP switches for undocumented purposes (e.g. incomplete features, debug leftovers, etc.).

cuavas:
I had a look at this and found a few things:
* `crystal2` wants the ANALYZER input to be a toggle
* `crystalg` doesn’t implement attract mode sound at all (static attract screen)
* I couldn’t confirm whether `crystalg` actually implements the Sangen Rush feature
* The last chance count settings are 0/3/5/10 as stated in the manual, MAME’s 0/1/3/10 is wrong

PR #13742, opened 2025-05-24
hh_sm510.cpp: added new game Uchitel' arifmetiki

MilanBA:
New working systems
-------------------
Uchitel' arifmetiki (Elektronika) [Milan Galcik, Igor Nys]

### DIFF
--- a/src/mame/handheld/hh_sm510.cpp
+++ b/src/mame/handheld/hh_sm510.cpp
@@ -35,7 +35,7 @@ TODO:
   gnw_climbern, gnw_dkcirc, gnw_dkhockey, gnw_dkjrp, gnw_dkong3, gnw_gcliff,
   gnw_mariocmt, gnw_mariocmta, gnw_mariotj, gnw_mbaway, gnw_mmousep,
   gnw_pinball, gnw_popeyep, gnw_sbuster, gnw_snoopyp, gnw_zelda, trtreisl,
-  trspacadv, amusarit, uchitari
+  trspacadv, vesarif, uchitari
 
 ================================================================================
 
@@ -4526,7 +4526,7 @@ ROM_END
   * KB1013VK1-2 MCU
   * lcd screen with custom segments, 1-bit sound
 
-  In 1984, Электроника (Elektronika, USSR) released an unlicensed clone of
+  In 1983, Электроника (Elektronika, USSR) released an unlicensed clone of
   Nintendo G&W Egg: Ну, погоди! (Nu, pogodi!). This was followed by several other
   titles that were the same under the hood, only differing in graphics. They also
   made a slightly modified version, adding a new game mode (by pressing A+B)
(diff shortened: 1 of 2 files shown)


happppp:
Unlike Amusing Arithmetics, no export version exists, right?
So, use the transliterated title, which would be "Uchitel' arifmetiki", and the setname/rom "uchitela" or something.

Are the (?) because this was dumped from a loose chip and not an intact handheld? If so, how did you create the SVG?

happppp:
If it's a PCB from the same factories as the other handhelds, I think it's clear enough that it would be an Elektronika brand product, so is the (?) necessary there?
I'd also name it "title (prototype?)" in the game title string.

In uchitari's comment block, could you write some notes about why it's considered a prototype, and that the game title is not fully determined?
See for example nupogodi and auslalom comment blocks where we wrote some observations.

MilanBA:
You are right, Uchitel' arifmetiki was most probably intended to be released under the Elektronika brand.
I remove the (?) and modify the game title string as well.

Re comment block... Could be something like this?
---
So far, there is no evidence of a real unit. Only two hardware parts are available at the moment:
* PCB with chip (in fact, there must have been two PCBs; the other one was for the keypad, battery compartment and speaker)
* LCD
Both parts were found in ruins of the factory in Ukraine formerly producing LCD games.
As the device did not come into mass production, it's considered to be just a prototype.
The game title is not fully determined. There is a reference to a game called "Учитель арифметики" in the Микропроцессорные средства и системы magazine (year 1990, number 4, page 38), and this reference was taken as the most probable name for this game.
---

By the way, Elektronika released Nu pogodi already in 1983. Don't know if it's important and worth to be changed in my next commit.
Please advise.

happppp:
Yeah something like that.
Where did you get that Nu pogodi is from 1983? When I google around, it says 1984 everywhere.

BTW while I was browsing the driver game list, I remembered that we use transliterated titles for all the Elektronika games, not the English export titles. So Amusing Arithmetic is Vesolaya arifmetika, could you rename it in this same PR?

And what is Time↔Alarm for? The main reason I'm asking is because ↔ is not a valid character in MAME strings and you have to do u8"Time\u????Alarm" ?=hex utf8 index for this char.

MilanBA:
Sure, I can rename Amusing Arithmetic to Vesolaya arifmetika.
Should I also change amusarit to vesarif?

I used ↔ just to indicate the key loops through the time-alarm_on-time-alarm_off sequence. I can change the port name to something else, for example simple "Alarm" ?

Re 1983: check the following pictures...

NuPogodi_1983_1
NuPogodi_1983_2
NuPogodi_1983_3
NuPogodi_1983_4

happppp:
Yup, the Vesolaya arifmetika rom can be renamed too.
Button name: sure, name it Alarm.
And Nu pogodi 1983, also ok, go ahead and change it (no need to make a new PR)

MilanBA:
All done.

PR #13748, opened 2025-05-27
vt369 - allow rtvgc300/rtvgc300fz to display menus / select games, vt32 - preliminary support for new video mode used by lxpcsp 

mamehaze:
- added basic support for 'high resolution' mode on VT369 (currently renders at standard resolution, but the tiles used are now correct)
- added another new sprite mode, used in high res mode
- menus are now visible in rtvgc300, rtvgc300fz, background is visible on image match in lxcmcysp is visible
- moved some VT32 specific code to the VT32 device as it was preventing aero engine in the rtvgc300 sets from working
- split VT32 PPU into its own derived device
- added preliminary support for a new VT32 (or maybe a newer variant of it) specific(?) video mode used by lxpcsp and maybe a few others
- added VT32 specific support for a bitswap on PPU character data pulls used by some titles (some of the bonus games in myaasa etc.)
- improved a few notes

### DIFF
--- a/src/devices/video/ppu2c0x.h
+++ b/src/devices/video/ppu2c0x.h
@@ -78,6 +78,8 @@ public:
 	auto int_callback() { return m_int_callback.bind(); }
 
 	void spriteram_dma(address_space &space, const u8 page);
+	void set_spriteram_value(offs_t offset, u8 data) { m_spriteram[offset] = data; }
+
 	void render(bitmap_rgb32 &bitmap, bool flipx, bool flipy, int sx, int sy, const rectangle &cliprect);
 	u32 screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect);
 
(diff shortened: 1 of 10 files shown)


mamehaze:
lx1 lx2

lx3 lx4

The menus can now be used, and games selected, so while they still render in a lower resolution than they should, the sets can now be used for further testing of other aspects of the emulation.

mamehaze:
going to mark this as mergable because it improves some understanding and we've got 2 more versions of these on the way (hence why I wanted to make them boot, so those can be verified once dumped)

mamehaze:
The extra VT32 video mode (at least I'm assuming it's VT32 due to other factors) allows the Lexibook Power Console to show something, so you can select the unlicensed bootleg of Nintendo's NES Pinball game.  Why that thing is so prevalent in so many of these plug and play games I don't know.

mamehaze:
@ajrhacker as you dealt with the first part of these changes could you take a look at this part.  I'd rather not make any more changes (as they carry a higher risk) before this part is in.

PR #13755, opened 2025-05-29
phc25: Invert cassette polarity.

ovalenti:
### Description

When trying to load the WAV file produced by castool on a real phc-25, we noticed that the program is not detected and the loading fails.

We compared with a working WAV file and noticed that the baud polarity is inverted. If we invert the polarity in the WAV file produced by castool, then it loads correctly on real hardware.

### Change

This PR proposes to invert the cassette polarity both in the machine and in the format ; this is a noop for people loading a .phc file in MAME, but should fix using castool for the real target.

- [ ] ~~Tested the castool binary from CI with real hardware~~ Can't do because castool is not bundled in artifacts.

### DIFF
--- a/src/lib/formats/phc25_cas.cpp
+++ b/src/lib/formats/phc25_cas.cpp
@@ -57,15 +57,15 @@ static int phc25_output_bit(int16_t *buffer, int sample_pos, bool bit)
 
 	if (bit)
 	{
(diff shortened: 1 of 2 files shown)


ovalenti:
@ICEknigh7, it definitely looks like it.

Then this PR is not the actual solution, and the cassette device should be added an "inverts" attribute that would be set only for dataset recorder, I guess ?

PR #13756, opened 2025-05-29
fidelity/vsc.cpp: add printer port output

goldnchild:
Adds serial output to vsc, use -rs232 terminal and set the terminal RX in Machine Configuration to 600 baud, 7N1.

RE will reset the unit, printing "CHESS CHALLENGER"

clicking on a piece button in the right sidebar buttons, then clicking on ST will print the board, CL to exit

Each move gets sent to the printer.

### DIFF
--- a/src/mame/fidelity/vsc.cpp
+++ b/src/mame/fidelity/vsc.cpp
@@ -145,10 +145,17 @@ chess computers also have support for it. Two models were released:
 FP: Challenger Printer - thermal printer, MCU=D8048C243
 IFP: Impact Printer - also compatible with C64 apparently.
 
+The printer data outputs at 600 baud, with 7 data bits and 1 stop bit.
+
+To use the printer with the serial terminal, add -rs232 terminal to the command line.
+For the terminal to work properly, in the Machine Configuration menu,
+set the RX data rate to 600, 7 data bits, no parity, and 1 stop bit.
+
 *******************************************************************************/
 
 #include "emu.h"
 
+#include "bus/rs232/rs232.h"
 #include "cpu/z80/z80.h"
 #include "machine/i8255.h"
 #include "machine/clock.h"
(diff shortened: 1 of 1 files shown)


happppp:
Hi! I saw your forum post and am working on adding rs232 to some Fidelity chess drivers.
I've tried it once before, I think the reason I failed back then is because I hadn't tried to set data bits to 7.

@holub yes it's possible, but I think it's kind of silly to have to do this (see below). And I don't really like drivers having 'knowledge' of which rs232 devices MAME supports.
```
m_rs232->set_option_device_input_defaults("null_modem", DEVICE_INPUT_DEFAULTS_NAME(rs232));
m_rs232->set_option_device_input_defaults("printer", DEVICE_INPUT_DEFAULTS_NAME(rs232));
m_rs232->set_option_device_input_defaults("pty", DEVICE_INPUT_DEFAULTS_NAME(rs232));
m_rs232->set_option_device_input_defaults("terminal", DEVICE_INPUT_DEFAULTS_NAME(rs232));
```

goldnchild:
> Hi! I saw your forum post and am working on adding rs232 to some Fidelity chess drivers. I've tried it once before, I think the reason I failed back then is because I hadn't tried to set data bits to 7.
> 
> @holub yes it's possible, but I think it's kind of silly to have to do this (see below). And I don't really like drivers having 'knowledge' of which rs232 devices MAME supports.
> 
> ```
> m_rs232->set_option_device_input_defaults("null_modem", DEVICE_INPUT_DEFAULTS_NAME(rs232));
> m_rs232->set_option_device_input_defaults("printer", DEVICE_INPUT_DEFAULTS_NAME(rs232));
> m_rs232->set_option_device_input_defaults("pty", DEVICE_INPUT_DEFAULTS_NAME(rs232));
> m_rs232->set_option_device_input_defaults("terminal", DEVICE_INPUT_DEFAULTS_NAME(rs232));
> ```

ohhhh, so that's how it works, I tried setting some DEVICE_INPUT_DEFAULTS but it didn't work.

I was able to get it working with some other fidelity units, I can add them as well.

happppp:
> I was able to get it working with some other fidelity units, I can add them as well.

Thanks but it's not necessary, as I said I'm working on it.

happppp:
Here we go https://github.com/mamedev/mame/commit/ead01ec36726a27b340954c35f034e076b48776e

PR #13764, opened 2025-05-31
trs/coco3.cpp: Change order of GIME initialization.

tlindner:
This fixes a hang on start that many people experienced.

### DIFF
--- a/src/mame/trs/coco3.cpp
+++ b/src/mame/trs/coco3.cpp
@@ -264,6 +264,15 @@ void coco3_state::coco3(machine_config &config)
 	INPUT_MERGER_ANY_HIGH(config, m_irqs).output_handler().set_inputline(m_maincpu, M6809_IRQ_LINE);
 	INPUT_MERGER_ANY_HIGH(config, m_firqs).output_handler().set_inputline(m_maincpu, M6809_FIRQ_LINE);
 
+	// memory and video hardware
+	GIME_NTSC(config, m_gime, XTAL(28'636'363), MAINCPU_TAG, RAM_TAG, m_cococart, MAINCPU_TAG);
+	m_gime->set_screen("screen");
+	m_gime->hsync_wr_callback().set(m_pia_0, FUNC(pia6821_device::ca1_w));
+	m_gime->fsync_wr_callback().set(m_pia_0, FUNC(pia6821_device::cb1_w));
+	m_gime->irq_wr_callback().set(m_irqs, FUNC(input_merger_device::in_w<2>));
+	m_gime->firq_wr_callback().set(m_firqs, FUNC(input_merger_device::in_w<2>));
+	m_gime->floating_bus_rd_callback().set(FUNC(coco3_state::floating_bus_r));
+
 	PIA6821(config, m_pia_0);
 	m_pia_0->writepa_handler().set(FUNC(coco_state::pia0_pa_w));
 	m_pia_0->writepb_handler().set(FUNC(coco_state::pia0_pb_w));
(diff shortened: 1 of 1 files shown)


happppp:
Is this a race condtion issue like so?:

- device_a::device_reset()
- device A writes with a devcb to device B in its device_reset
- device_b::device_reset()

changed to:
- device_b::device_reset()
- device_a::device_reset()
- device A writes with a devcb to device B in its device_reset

Since the lockup bug is random ("many people"), the underlying issue could be an uninitialized variable.

cuavas:
This looks pretty dubious.  Are you sure this isn’t just working by coincidence now?  Where was it actually hanging?

tlindner:
I've been getting reports over the years of hanging green screens in the CoCo 3 driver. But I've never experienced it personally.
Then someone suggested to add `memset(...,0xFF...)` to the GIME CTOR. Setting all of the internal state to FFs. This caused it to hang on startup every time. It still had the memset that set everything to zero in device_reset. That made me think there was a race condition between the GIME and PIA. 
This patch fixed the issue, even with the malicious CTOR.

tlindner:
The reports always said the hang was after emulation start, but before BASIC came up. Just a blank green screen.

tlindner:
The reports always said the hang was after emulation start, but before BASIC came up. Just a blank green screen.

cuavas:
Did you get a stack trace of the “hang”?  Was the emulator hung, or was it just the emulated system not responding?

cuavas:
> I've been getting reports over the years of hanging green screens in the CoCo 3 driver. But I've never experienced it personally. Then someone suggested to add `memset(...,0xFF...)` to the GIME CTOR. Setting all of the internal state to FFs. This caused it to hang on startup every time. It still had the memset that set everything to zero in device_reset.

So what is it that happens at start or reset time that causes the “hang”, and does the GIME need to be ready before reset time to avoid it?

Remember there’s a clearly-defined start sequence:
* Call machine configuration handlers – this is where devices are constructed.
* Call config complete handlers, order can be a bit weird when slot devices are involved.
* Resolve object finders.
* Call resolve object handlers in simple depth-first tree walk order.
* (This is the point where you need to be prepared to receive calls to “live signal” handler member functions.)
* Call start handlers.  This starts as a depth-first tree walk, but devices can throw a missing dependencies exception if they want to get another start call after something else is started.
* Call reset handlers in simple depth-first tree walk order.

It sounds to me like you’re either setting something to zero in `device_reset` that you shouldn’t, or you aren’t preparing something early enough.  Remember you mustn’t change the remembered state of inputs in `device_reset` – they should only change when the device the input comes from updates them.

> That made me think there was a race condition between the GIME and PIA.

There can’t be a race condition as it isn’t multi-threaded.

> This patch fixed the issue, even with the malicious CTOR.

No it doesn’t.  It just hides the issue.  The root issue is still there.  Please fix it properly.

happppp:
If the device_reset order breaks eg. an irq or a devcb r/w because it relied on another device being reset before it, then the mcfg order *is* the root issue.
For example, let's say on the hw, device B has a delay circuit before its reset pin that makes sure it's reset after device A. You can't do that properly in MAME, and reordering the mcfg is the only way to make sure devices are reset in the order you want to.

It's either an uninitialized variable and/or actually relying on device_reset order.

cuavas:
If something holds one of the devices in reset longer than the other, then the devices are not being reset at the same time.  The solution is to actually emulate that, e.g. use a timer to hold that device in reset for however long it is in real life.

happppp:
I mean power-on reset. MAME has no control over timing for that.

cuavas:
You can start a timer in `machine_reset` or wherever to bring “other” device out of reset at the correct time.

happppp:
Is it possible? diexec devices maybe (haven't tried*), but something like a PIA? Sure, you could call device_reset a 2nd time manually, but MAME's reset routine would still call device_reset with everyone else at reset.

*: set_input_line (for RESET pin), or set_disable() or suspend() are all diexec.

cuavas:
Yeah, you can’t avoid the initial `device_reset` (and `device_reset_after_children`), but if the device implements a reset input (which a lot of devices currently don’t, as they’re based on the flawed assumption that reset is always instantaneous), you can keep it reset for as long as it should be.  There are some devices that do provide a proper reset input, e.g. the 82586, the E05xx things, the PSION ASIC2, and a few others.

PR #13767, opened 2025-06-02
emu/debug: Replaced saver/loadr with more generic savem/loadm which handels both region and share memory

holub:


Osso13:
Please remember to update debughlp.cpp, too.

holub:
Sure, right after I see others have no concerns regarding this.

ajrhacker:
I don't think this is quite justified. There may be memory regions and memory shares with the same names.

holub:
I was thinking about it. It's quite immoral to expect the user to know about internal representation.
Maybe I can check if this conflict exists and request to supply an additional parameter which clarifies the type.

holub:
...or alternatively we can have savem which successes in case no conflicts only + samer, saves, etc

galibert:
Or an alternative alternative, assuming the parser is not insane, having a savem s:whatever/r:whatever/etc where the non-prefixed DWIM when it's not ambiguous and errors otherwise.

holub:
Making it draft and working on @galibert 's version.

cuavas:
Is it really worth making the syntax more complicated to avoid an extra command?  I already need to take a trip to the documentation when I want to do anything interesting with expressions ([this section](https://docs.mamedev.org/debugger/#memory-accesses)).

Also you’re outright removing a pair of commands which is going to break a pile of scripts, and also any ROM-patching cheats that depend on them.

On top of that, in the existing expression syntax, `m` is used for memory regions, not shares (e.g. `monitor.mb@78` means “the byte at address 78 in the memory region with the tag `monitor`).  If you introduce syntax in another place where `m` represents a memory share, that’s going to confuse people.

holub:
I can keep loadr and add loads. My concern is the user not expected to know what she really wants - she seeing the name in dropdown and uses the name. How do you expect her to know if it's share or region or banked?

cuavas:
If you really want to avoid adding extra commands, you could probably extend `save` and `load` to allow a region or share to be specified with a suffix similar to the ones used in expressions.  (Right now they only operate on address spaces.)

For example, use the `.m` suffix to access a region, as we already do in expressions.  So this:
```
save data.bin,200:monitor.m,100
```

Would be equivalent to this:
```
saver data.bin,200,100,:monitor
```

You’d need to choose another suffix for shares that doesn’t clash with current expression syntax (perhaps `.s`).

Also other things could be extended to support suffixes for region/share:
* Expression syntax
* `fill`, `dump`, `strdump` and `find` commands

The thing to be aware of if you do this is that then the `<addr>:<spec>` syntax will have two variants: one that only supports address spaces (e.g. for `wpset`) and one that also supports other things with suffixes.

But I think getting rid of `loadr`/`saver` or changing them in an incompatible way would be a mistake.

cuavas:
> I can keep loadr and add loads. My concern is the user not expected to know what she really wants - she seeing the name in dropdown and uses the name. How do you expect her to know if it's share or region or banked?

I mean, it tells you:
image

See there it says “Region” for memory regions.

image

And it says “memory” for shares.

They already need to know what a region is to use the `.m` syntax in expressions.

holub:
So... the plan is:
* Keep `saver`
* Add `savee` without exposing
* Allow to specify `.m` `.e` for `save` without auto-magical selection

"e" corresponds to m(e)mory in dropdown or shar(e) in internal representation.

same for `load`/`fill`/`dump`/...

cuavas:
If you’re extending the syntax for one of the `load`, `save`, `fill`, `dump`, etc. commands, it needs to be done for all of them.  If you add a suffix for accessing a memory share, it needs to be supported in expressions as well as commands.  It would be confusing otherwise.

The source argument parsing is already encapsulated for things that take the `<addr>:<spec>` syntax for address/space.  It would be a matter of adding a variant that can yield a region or share instead of an address space.  There’s already some kind of support for using regions as well as address spaces in the expression handling code.

holub:
Understood. I'll be working on all variants then.

Any concerns about "e" ?

cuavas:
I’m not sure.  I would’ve gone for `s` since people using the MAME debugger are kind of assumed to have at least glanced at the MAME source.  Any opinion @galibert?

holub:
The plan is clear and it will take some time.
You have time to pick the letter.

holub:
I've implemented save/load but don't really like how is the code structured. If you have any ideas how to make it better let me know while I'm implementing the rest.

holub:
That must be it

holub:
Thoughts?

holub:
gents?

PR #13768, opened 2025-06-02
ADP: Servicetastatur

stonedDiscord:
This is the service keyboard for all the machines in the adp folder.
It starts up but complains about the i2c eeprom being wrong/missing.

Picture, schematic and ROM (not by me):
https://archive.org/details/schaltplan-forum

### DIFF
--- a/src/mame/adp/servicetastatur.cpp
+++ b/src/mame/adp/servicetastatur.cpp
@@ -7,7 +7,7 @@ Profitech 3000 Servicetastatur
 Hardware:
 - CPU: 80C52 @ 11.0592MHz
 - LCD: LCD4002A
-- Memory: 27C256 EPROM (32KB), 24C16 I2C EEPROM (2KB)
+- Memory: 27C256 EPROM (32KB), 24CS16 I2C EEPROM (2KB)
 
 Key Matrix Layout:
 Col 0 (P1.0): OK, F4, UP
@@ -42,14 +42,14 @@ enum
 
 enum
 {
-	PORT_3_RXD     = 1 << 0,
-	PORT_3_TXD     = 1 << 1,
-	PORT_3_INT0    = 1 << 2,
-	PORT_3_INT1    = 1 << 3,
-	PORT_3_SDA     = 1 << 4,
-	PORT_3_SCL     = 1 << 5,
-	PORT_3_WR      = 1 << 6,
-	PORT_3_RD      = 1 << 7,
+	PORT_3_RXD  = 1 << 0,
+	PORT_3_TXD  = 1 << 1,
+	PORT_3_INT0 = 1 << 2,
+	PORT_3_INT1 = 1 << 3,
+	PORT_3_SDA  = 1 << 4,
+	PORT_3_SCL  = 1 << 5,
+	PORT_3_WR   = 1 << 6,
+	PORT_3_RD   = 1 << 7,
 };
 
 class servicet_state : public driver_device
@@ -87,13 +87,6 @@ private:
 	uint8_t m_port3 = 0xff;  // Port 3 output latch
 	uint8_t m_lcd_data = 0;      // LCD data bus
 
-	// I2C debugging state
-	bool m_last_scl = true;
-	bool m_last_sda = true;
-	uint32_t m_i2c_bit_count = 0;
-	uint8_t m_i2c_byte = 0;
-	bool m_in_i2c_transaction = false;
-	
 	void servicet_io(address_map &map) ATTR_COLD;
 	void servicet_map(address_map &map) ATTR_COLD;
 };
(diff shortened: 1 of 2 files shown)


stonedDiscord:
This is the service keyboard for all the machines in the adp folder.
It starts up but complains about the i2c eeprom being wrong/missing.

Picture, schematic and ROM (not by me):
https://archive.org/details/schaltplan-forum

rb6502:
Can it be made to write out a valid EEPROM?  If so, we'd want to include that in the ROM set.

Also, how does this interact with the actual machines?

stonedDiscord:
It plugs into the machine with a 6pin connector on a cable.
Wiping the EEPROM on a real one doesn't show the error message so I assume it can't detect it.
The EEPROM is used to save settings from one machine and apply them to others for when you have to set up an entire arcade hall.
If there is a way to write it out it's probably some key combination during startup, I tried to press some keys but I don't think they registered correctly.

angelosa:
Perhaps the `eeprom` is expected to be 1 filled to make it initialize?

stonedDiscord:
```cpp
	ROM_REGION(0x800, "eeprom", 0)
	ROM_FILL(0x000, 0x7ff, 0xff )
```
sadly didn't change anything

angelosa:
Just to be sure:
1. you need to delete your existing eeprom file, in `nvram` directory.
2. `ROM_REGION(0x800, "eeprom", ROMREGION_ERASEFF)` has same effect.

stonedDiscord:
same result

rb6502:
It's likely the I2C comms aren't right - it might need a pull-up resistor on SCL and/or SDA or something of that nature.

stonedDiscord:
There are no resistors (or other components) on the lines between the CPU and the EEPROM.
My suspicion is that I did something wrong in `servicet_state::port3_r()` but I can't find out what and would like to submit it in this state anyway.

src/mame/adp/servicetastatur.cpp
```diff
@@ -0,0 +1,300 @@
...
+		m_maincpu(*this, "maincpu"),
+		m_i2cmem(*this, "eeprom"),
+		m_lcd(*this, "hd44780"),
+		m_io_keys(*this, "IN%u", 0U)
+	{ }
+
+	void servicet(machine_config &config);
+
+private:
+	virtual void machine_start() override ATTR_COLD;
```
startaq, line 89: Those two should be `protected`.

src/mame/adp/servicetastatur.cpp
```diff
@@ -0,0 +1,300 @@
...
+	required_device<mcs51_cpu_device> m_maincpu;
+	required_device<i2cmem_device> m_i2cmem;
+	required_device<hd44780_device> m_lcd;
+	required_ioport_array<3> m_io_keys;
+	
+	uint8_t m_port1 = 0xff;
+	uint8_t m_port3 = 0xff;
+	uint8_t m_lcd_data = 0;
+
+	void servicet_io(address_map &map) ATTR_COLD;
```
startaq: Move these two up please, to not mix variables/functions.

src/mame/adp/servicetastatur.cpp
```diff
@@ -0,0 +1,300 @@
...
+ROM_START( servicet )
+	ROM_REGION( 0x8000, "maincpu", 0 )
+	ROM_LOAD( "service_tastatur_v3.3.u3", 0x0000, 0x8000, CRC(8eb161c4) SHA1(d44f3b38e75e1095487893d8b30c4e3212c1a143) )
+
+	ROM_REGION(0x800, "eeprom", ROMREGION_ERASEFF)
+ROM_END
+
+} // anonymous namespace
+
+GAME( 1992, servicet, 0, servicet, servicet, servicet_state, empty_init, ROT0, "ADP", u8"Merkur Service Testgerät", MACHINE_NOT_WORKING | MACHINE_NO_SOUND_HW )
```
startaq: Missing newline at end of file.

src/mame/adp/servicetastatur.cpp
```diff
@@ -0,0 +1,300 @@
...
+			m_lcd_data = data;
+			
+			m_lcd->rs_w(rs);
+			m_lcd->rw_w(0);
+			m_lcd->db_w(data);
+			
+			m_lcd->e_w(1);
+			m_lcd->e_w(0);
+		}
+	} else if (offset == 0x4000){
```
startaq: Bracing style inconsistent here (and below).

stonedDiscord:
The display flickers in the emulation and I wasn't able to find a good probe point with my scope on it. The real device also flickers when I look at the LCD from an angle.

startaq (approved):

PR #13770, opened 2025-06-02
cpu/sharc: correctly align 32-bit data transfers using PM bus

gm-matthew:
Both Fighting Vipers and Sonic the Fighters will occasionally write values to internal memory using the DM (data memory) bus, and then try to read them back using the PM (program memory) bus; these were not being aligned correctly, resulting in corrupted values and ultimately leading to the "glitchy polygons" bug for certain characters.

According to the [ADSP-21062 manual](https://www.analog.com/media/en/dsp-documentation/processor-manuals/50836807228561adsp2106xsharcprocessorusersmanual_revision2_1.pdf) (page 5-5), 32-bit memory transfers using the PM bus are aligned to the upper 32 bits of the bus, whereas the existing code aligns it to the lower 32 bits; my fix correctly aligns data to the upper 32-bits of the PM bus for both 32-bit reads and writes by performing a QWORD access, as is used for 48-bit transfers, and shifting left or right by 16 bits accordingly.

Without having intricate knowledge of MAME's memory mapping code, I am uncertain if there is a way to set up the memory map so that 32-bit transfers automatically align to the upper 32 bits of the PM bus rather than having to do it manually as my implementation does; if there is, I welcome an alternative implementation.

### DIFF
--- a/src/devices/cpu/sharc/sharcmem.hxx
+++ b/src/devices/cpu/sharc/sharcmem.hxx
@@ -10,8 +10,7 @@ uint32_t adsp21062_device::pm_read32(uint32_t address)
 
 void adsp21062_device::pm_write32(uint32_t address, uint32_t data)
 {
-	u64 res = ((u64)data << 16) | (m_program->read_qword(address) & 0xffff);
-	m_program->write_qword(address, res);
+	m_program->write_qword(address, (u64)data << 16, 0xffffffff0000);
 }
 
 uint64_t adsp21062_device::pm_read48(uint32_t address)


angelosa:
fwiw `m_program` is set as:
`	, m_program_config("program", ENDIANNESS_LITTLE, 64, 24, -3, address_map_constructor(FUNC(adsp21062_device::internal_pgm), this))`

I don't think there's a way to align memory accesses like you describe here, OG?

gm-matthew:
I had a feeling there wasn't a way to set up the memory map to do the alignment automatically, but's it nice to get it more-or-less confirmed by a MAME dev more experienced than myself :-)

ajrhacker:
> Without having intricate knowledge of MAME's memory mapping code, I am uncertain if there is a way to set up the memory map so that 32-bit transfers automatically align to the upper 32 bits of the PM bus rather than having to do it manually as my implementation does; if there is, I welcome an alternative implementation.

One variation that MAME's memory system allows and which might be appropriate here is using the mem_mask parameter in the calls to `read_qword` and `write_qword`.

galibert:
I don't see any better way than using the mask.  There's an open question though, the documentation not being 100% clear there, are the bottom 16 bits kept as-is or cleared?  Any evidence anywhere?

gm-matthew:
I would presume that the bottom 16 bits are kept as-is.

The DM bus is actually 40 bits wide to accommodate extended precision floating point data (currently not emulated), and once again 32-bit transfers are aligned to the upper 32 bits of the DM bus. If the lower 8 bits were cleared, every 32-bit write would have the side effect of clearing 8 bits elsewhere which could easily lead to data being corrupted.

I believe that the bottom 8 bits of the DM bus are ignored for 32-bit transfers, and that the same applies to the bottom 16 bits of the PM bus.

rb6502:
The ADSP-21062 manual doesn't seem to address that detail, but the easy way to check would be to make the emulation actually stomp the bottom bits and see if the games still run.  Likely there'll be an immediate obvious failure and that would work as proof.

rb6502:
`If an ADSP-2106x program attempts to write 40-bit data (in a 48-bit word) to a memory block configured for 32-bit data, the lower 16 bits (of the 48-bit word) are truncated. If a 40-bit data read is attempted, the lower 8 bits will be zeros. `

I can't tell if this means writes zero out the bottom bits or they're just read back as zero in that mode.  So testing if the Sega code puts up with it is probably the easiest way to tell.

gm-matthew:
I've checked the copro SHARC code for every single Model 2B game and most of them only use the PM bus for loading instructions and never for reading or writing data. The only exceptions are Fighting Vipers and Sonic the Fighters; in both cases the PM bus is used for reading and writing data to Block 0 (which is never accessed by the DM bus) but only reading data from Block 1.

Since the PM bus only writes to a block that is set up to be 48 bits wide and is never accessed by the DM bus, it shouldn't make any difference whether the lower 16 bits are cleared or not since they are never actually used.

I've tested most of the 2B games with the mask for 32-bit writes over the PM bus removed (so the lower 16 bits are cleared) and they appear to work just fine, but that doesn't confirm whether 32-bit writes using the PM bus clear the lower 16 bits or not.

I haven't yet tested any other machines that use the ADSP-21062; I shall do so later this evening.

gm-matthew:
I've looked at several of the Konami games that use the ADSP-21062 and they appear to use the PM bus only for dynamically loading program code from shared RAM, and they do so using the PX register which is 48 bits wide.

It seems that it does not matter whether we clear the lower 16 bits or not when writing a 32-bit word using the PM bus, though I'd personally be in favour of leaving the lower 16 bits as-is because it seems more logical to me.

rb6502:
Thanks for checking!

PR #13774, opened 2025-06-03
nintendo: Move Plug & Play SoC drivers (mostly VTxx & SH6578) to tvgames directory

ajrhacker:
@mamehaze, do you approve of this reorganization? (I mentioned it on Bannister but it scrolled off.)


mamehaze:
I don't really mind.

They were originally put there because in many of the early ones there are a *lot* of straight-up NES games in the multigame units, and even a defined NES 'mapper' to represent them, but MAME is never going to support them in that way anyway (because it doesn't actually make sense, they're not NES compatible)

tvgames was meant more of a place to put things without just further clogging up misc/skeleton folders though, so they had somewhere to go before moving them to more appropriate places, so in that sense I'd probably just make a 'vrt' folder and put them there instead.

I'd also make a sunplus (or spg, or generalplus) folder and move the sunplus/generalplus stuff to it, and likewise probably a xavix one, as once we start emulating all the custom controls and refactor things a bit that's going to end up in a lot more driver files too.

(if you decide to move the sunplus stuff please make sure any outstanding PRs adding SunPlus sets are merged first to avoid conflicts though)

ajrhacker:
I don't think organizing drivers by IC manufacturer is the way to go here. I believe that was tried for some of the handheld drivers and rejected (partly because those MCU manufacturers also had their own branded products).

mamehaze:
in this case I think it defines the products a bit more though.

the SoCs are systems, they were made by VRT/SunPlus etc. and while each unit was typically it's own thing rather than a single system with carts, that doesn't change much.  The completeness of each chip, the fact that they handle all the graphics / sounds / IO etc. as well as the line of progression between each platform makes them much more like a manufacturer/console structure.

For XaviX especially, nearly every game (all but 2) carries the SSD company copyright and startup logos (I suspect it was a license requirement to use the SDK, just as you see on consoles, and those not showing it were bending the rules)

I don't really see a difference between these SoCs and a platform provider, and things are very much already grouped by platform provider.

ajrhacker:
I'll mothball this for now.

mamehaze:
I think it's worth dropping the nes_ prefixes on the files at least, which was on my 'TODO' list, and I saw you'd done with this, but yeah either the move needs doing or it doesn't, having an open PR which moves everything just kinda blocks progress until a decision is made.

PR #13786, opened 2025-06-06
sega/segahang.cpp: Fix PCM sample loading for endurobl

cam900:
I think it's copy-paste error or confused with 2xYM2203/YM2151 sound set?

### DIFF
--- a/src/mame/sega/segahang.cpp
+++ b/src/mame/sega/segahang.cpp
@@ -2154,7 +2154,7 @@ ROM_START( endurobl )
 
 	ROM_REGION( 0x20000, "pcm", 0 ) // Sega PCM sound data
 	ROM_LOAD( "epr-7762.ic5", 0x00000, 0x8000, CRC(bc0c4d12) SHA1(3de71bde4c23e3c31984f20fc4bc7e221354c56f) )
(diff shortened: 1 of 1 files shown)


angelosa (approved):

src/mame/sega/segahang.cpp
```diff
@@ -2154,7 +2154,7 @@ ROM_START( endurobl )
 
 	ROM_REGION( 0x20000, "pcm", 0 ) // Sega PCM sound data
 	ROM_LOAD( "epr-7762.ic5", 0x00000, 0x8000, CRC(bc0c4d12) SHA1(3de71bde4c23e3c31984f20fc4bc7e221354c56f) )
-	ROM_LOAD( "epr-7763.ic6", 0x10000, 0x8000, CRC(627b3c8c) SHA1(806fe7dce619ad19c09178061be4607d2beba14d) )
+	ROM_LOAD( "epr-7763.ic6", 0x08000, 0x8000, CRC(627b3c8c) SHA1(806fe7dce619ad19c09178061be4607d2beba14d) )
```
cuavas, line 2157: Perhaps the region should have a fill flag to fill the unused space with either all high or all low, but none of the other games that don’t use all the `pcm` region do this, so it’s no worse than before.

PR #13788, opened 2025-06-07
heathzenith/z100.cpp: Confirm/update PROM images

mgarlanger:
PROMs from my Z100 were imaged and compared with the existing "BAD_DUMP" images that were typed in from the manual. Only 1 image 444-127 was different, and comparing the 2, it appears the type one may have had a typo.

Code has been updated (BAD_DUMP removed, and 444-127 hash updated).

The code is not currently using the PROMs, so no differences in behavior is expected, machine is still not working.


ajrhacker:
The Video RAM Mapping Module PROM is in fact currently used by the emulation.

PR #13790, opened 2025-06-07
hash/snes: New dump that will never give you up

chungy:
MAME needed some more Rick Astley, and this ROM exists ;)

The ROM was downloaded from "vizzed.net" and the time stamp inside it might not reflect the origin, but I haven't been able to track down a more primary source ("rick roll" ain't the easiest thing to search for with narrowing things down). If the date should be revised later, so be it, but 2008 is around the right time frame for Rick Rolling's hey day.

### DIFF
--- a/hash/snes.xml
+++ b/hash/snes.xml
@@ -54962,6 +54962,19 @@ to ensure nothing has been touched in the Retro Quest cart production in 2013/20
 		</part>
 	</software>
 
+	<software name="rickroll">
+		<description>Rickrolled!</description>
+		<year>2008</year>
+		<publisher>Smallhacker &amp; smkdan</publisher>
+		<info name="release" value="20081204" /> <!-- File date in zip archive -->
+		<part name="cart" interface="snes_cart">
+			<feature name="slot" value="lorom" />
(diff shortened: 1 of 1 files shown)


ajrhacker:
Unfortunately, the "!" in the file name will fail validation and let you down.

ajrhacker:
This was apparently originally released as a ROM hack of Super Mario World under the title Xka Shack Demo. The 2008 date seems to be correct.

hash/snes.xml
```diff
@@ -66892,6 +66892,26 @@ to ensure nothing has been touched in the Retro Quest cart production in 2013/20
...
+	<!--
+		Initially released as a patch from Super Mario World, as a
+		form of trolling consistent with the “Rick Roll” meme.
+
+		https://www.smwcentral.net/?p=section&a=details&id=17482
+	-->
+	<software name="xkashack">
+		<description>Xka Shack</description>
+		<year>2008</year>
+		<publisher>Smallhacker &amp; smkdan</publisher>
```
angelosa, line 66904: Is there an official cart manufactured for this?

PR #13792, opened 2025-06-08
Various GNU/Hurd fixes

pinotree:
This PR provides various fixes to help MAME build on GNU/Hurd; there are still few things to fix (e.g. `PATH_MAX`). In particular:
- cherry-pick an upstream fix in bx (see https://github.com/bkaradzic/bx/pull/346) to use `BX_PLATFORM_POSIX` more, and drop support for very old GNU libc versions
- fix a local addition in bx for Hurd, i.e. use include the standard `<pthread.h>` instead of the non-existing `<pthread/pthread.h>`
- use `dirent::d_type` and `<pty.h>` on any GNU libc platform, as they are implemented by that C library
- fix the usage of `dirent::d_name`, so it is used properly no matter its actual implementation type

See the message of the commits for longer explanations.

### DIFF
--- a/src/osd/modules/file/posixdir.cpp
+++ b/src/osd/modules/file/posixdir.cpp
@@ -157,7 +157,7 @@ const osd::directory::entry *posix_directory::read()
 	m_entry.name = m_data->d_name;
 
 	sdl_stat st;
-	bool stat_err(0 > sdl_stat_fn(util::string_format("%s%c%s", m_path, PATHSEPCH, m_data->d_name).c_str(), &st));
+	bool stat_err(0 > sdl_stat_fn(util::string_format("%s%c%s", m_path, PATHSEPCH, static_cast<const char *>(m_data->d_name)).c_str(), &st));
 
 #if HAS_DT_XXX
 	switch (m_data->d_type)


src/osd/modules/file/posixdir.cpp
```diff
@@ -157,7 +157,7 @@ const osd::directory::entry *posix_directory::read()
 	m_entry.name = m_data->d_name;
 
 	sdl_stat st;
-	bool stat_err(0 > sdl_stat_fn(util::string_format("%s%c%s", m_path, PATHSEPCH, m_data->d_name).c_str(), &st));
+	bool stat_err(0 > sdl_stat_fn(util::string_format("%s%c%s", m_path, PATHSEPCH, static_cast<const char *>(m_data->d_name)).c_str(), &st));
```
cuavas, line 160: What kind of screwed-up version of dirent.h has a `dirent` or `stat` structure where `d_name` doesn’t look like a C string and needs to be cast?  If anything I’d say this increases the likelihood of something going wrong if someone screws up and end up with a `wchar_t` version of the structure when they shoudn’t.

src/osd/modules/file/posixptty.cpp
```diff
@@ -25,7 +25,7 @@
 #elif defined(__NetBSD__) || defined(__OpenBSD__) || defined(__APPLE__)
 #include <termios.h>
 #include <util.h>
-#elif defined(__linux__) || defined(__EMSCRIPTEN__)
+#elif defined(__linux__) || defined(__EMSCRIPTEN__) || defined(__GLIBC__)
```
cuavas, line 28: I don’t think detecting the use of glibc here is the right thing to do.  glibc can be used on platforms that don’t use pty.h.

galibert: Isn't glibc the one providing pty.h?

src/osd/modules/file/posixdir.cpp
galibert: One that uses a end-of-structure, "variable" size C char array.  E.g. char mumble[0]; or worse char mumble[1];.

src/osd/modules/file/posixptty.cpp
cuavas: I thought it was possible to build GLIBC for kernels other than Linux, and if they don’t provide Linux-style PTY, you won’t get a pty.h staged for distribution (e.g. a system that uses STREAMS).

src/osd/modules/file/posixdir.cpp
pinotree: As I wrote, in my commit message:

> According to POSIX [1], the type of dirent::d_name is loosely defined
as "char d_name[]", as array with an undefined size. In particular, few
ways are seen in the wild:
(a) "char d_name[size]", i.e. as proper array with a full size
(b) "char d_name[1]"/"char d_name[0]", i.e. as C flexible arrays
Regardless of its type, dirent::d_name is used as if it was a classic
const char *, i.e. as pointer to a null-terminated string.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/dirent.h.html

src/osd/modules/file/posixptty.cpp
pinotree: > I thought it was possible to build GLIBC for kernels other than Linux

glibc supports also GNU/Hurd currently, and there used to be a port for the GNU/kFreeBSD Debian distribution.

> if they don’t provide Linux-style PTY

`openpty()` is not a "Linux-style PTY", it's rather a BSD-derived API that it not standard in any way; see the documentation for it:
- the glibc documentation: https://sourceware.org/glibc/manual/latest/html_mono/libc.html#Opening-a-Pseudo_002dTerminal-Pair
   - `pty.h` is explicitly mentioned here
- Linux man page: https://linux.die.net/man/3/openpty

cuavas: https://www.gnu.org/software/gnulib/manual/html_node/Supported-Platforms.html

> * Native Windows:
>   * mingw is occasionally tested. Only the latest version of mingw is tested; older versions are not supported.
>   * MSVC 14 (Microsoft Visual Studio 2015 14.0) is occasionally tested. Only “release” builds (compiler option ‘-MD’) are supported, not “debug” builds (compiler option ‘-MDd’).
>
> Note that some modules are currently unsupported on native Windows: `mgetgroups`, `getugroups`, `idcache`, `userspec`, `openpty`, `login_tty`, `forkpty`, `pt_chown`, `grantpt`, `pty-h`, `savewd`, `mkancesdirs`, `mkdir-p`, `euidaccess`, `faccessat`. The versions of Windows that are supported are Windows 10 and newer.

My memory was correct.  When glibc is built for a target that doesn’t support similar enough PTYs to Linux, you don’t get a staged `pty.h` and `openpty` (along with the other PTY-related functions) is not available.  Testing for glibc is not sufficient for determining that `pty.h` is available.

Issue #7957, opened 2021-04-15
Leftover sprites after continue countdown in contra


furrtek:
Sprites forming the digits of the countdown timer aren't cleared properly after pressing start when it shows 10.
Shown here: https://mametesters.org/view.php?id=401

I tried simplifying how the sprite RAM is handled to reflect how the real chip works (no buffer copying involved) but it didn't solve the issue.

The CPU maintains a sprite list starting at 0x1200 in RAM, which is used to generate the sprite RAM data for k007121_2 each new frame. The first few entries seem to always be used for P1's sprite and bullets.
When the countdown timer starts, those entries are overwritten by several 8x8 sprites to form the digits.

If the player presses start when the counter is below 10, P1's sprites overwrite enough of the entries to clear the single digit.
However if the player presses start at 10, more sprites are used so P1's sprites aren't enough to clear both digits.
Firing or jumping causes P1's sprite list to expand for a moment, effectively clearing the remaining sprites.

It seems that the CPU is rewriting the garbage sprites, since they reappear even if the k007121_2's sprite RAM (0x5000~0x5FFF) is cleared manually. So the bug might not be related to the k007121.

galibert:
I don't know that hardware (I know the following ones), but Konami tends to
follow a line of design.  In the later ones there can only be one sprite
per priority level because that level tells the chip where to write the
sprite data in its internal sprite ram (we don't emulate that correctly,
partially because we don't emulate that part of the tbyahhoo protection
correctly), and sprites with the lowest priority level (0 or ff, depending
on a register) are ignored.  Could there be something similar going on?

  OG.

furrtek:
If I understood correctly, the k007121 only has global priority settings and individual priority bits for scroll layer tiles only.
The sprites seem to be rendered one after the other in the order they are in sprite RAM.

Contra clears unused sprites by setting their Y position to 0xF0. For a moment I thought that could be a sprite list end code, since stopping sprite rendering with that condition solved the countdown issue, but that also caused some blinking when sprites were moved so it must not be the right solution. I also can't find how a sprite attribute byte could stop the parsing counter in the chip's schematics.

About the sprite RAM double buffering, it's actually very simple: reg 3 bit 3 selects which buffer is used for rendering (0000~07FF or 0800~0FFF). The bit is toggled each frame, so the CPU can load the free buffer while the other is being used.
I'm not 100% sure if the CPU can always access both buffers, or only the free one.

smf-:
[...]

PR #13795, opened 2025-06-08
k007121: sprite list and related fixes

jwestfall69:
I was recently doing some testing on a contra board around the 007121 IC and found the following

- There is no sprite limit, but instead a 264 8x8 sprite block limit
- There is no buffering of sprite data on the attached SRAM

From this the follow changes have been made

- Use the 264 8x8 sprite block limit to determine number of sprites that should be rendered
- Remove sprite ram buffering, and move logic for picking sprite ram bank into k007121.cpp
- Comment updates related to the limit and non-buffering

Additionally the high bit of tiles are stored in one of the control registers, if this bit changes all tiles should be marked as dirty.

These changes fix a number of sprite related issues on games using this chip

mametesters [00185](https://mametesters.org/view.php?id=185) (combatsc: missing sprites)
mametesters [00401](https://mametesters.org/view.php?id=401) + issue #7957 (contra: left over sprites)
etc

### DIFF
--- a/src/mame/konami/k007121.cpp
+++ b/src/mame/konami/k007121.cpp
@@ -225,8 +225,8 @@ void k007121_device::sprites_draw(bitmap_ind16 &bitmap, const rectangle &cliprec
 
 	assert(MAX_SPRITE_BLOCKS < 0x199); // floor(0x800 / SPRITE_FORMAT_SIZE)
 
-	// There is 0x1000 sprite ram, which is broken up into 2 0x800 chunks.
-	// The follow control bit determine which chunk is used.
+	// There is 0x1000 sprite ram, which is broken up into 2 0x800 banks.
+	// The following control bit determines which bank is used.
 	if (BIT(m_ctrlram[3], 3))
 		source += 0x800;
 


happppp:
1x2 and 2x1 sprites are 2 blocks, not 3.

Could you reword this? Mention your 264 measurement. And if the Combat School note is nonsense, remove it.
```
 * There are 0x40 sprites, each one using 5 bytes. However the number of
 * sprites can be increased to 0x80 with a control register (Combat School
 * sets it on and off during the game).
```

src/mame/konami/k007121.cpp
```diff
@@ -210,22 +215,57 @@ void k007121_device::ctrl_w(offs_t offset, uint8_t data)
  *
  */
 
+#define SPRITE_FORMAT_SIZE 5
+#define MAX_SPRITE_BLOCKS 264		// Maximum number of 8x8 sprite blocks that can be drawn
+#define MAX_SPRITES 0x199		// floor(0x800 / SPRITE_FORMAT_SIZE)
```
happppp: Put these inside the function, like so:
constexpr unsigned BLAH = val;

And instead of checking it in the loop,
how about assert(MAX_SPRITE_BLOCKS < 0x199); // floor(0x800 / SPRITE_FORMAT_SIZE)

happppp:
Also, if I see it correctly, can the sprite ram bank handling be removed from the MAME drivers?
So, inside the sprites_draw function:

```
if (BIT(m_ctrlram[3], 3)
    source += 0x800;
```

jwestfall69:
Thanks for the feedback, I will look into the requested changes.  I was also going to write a quick test to run on contra to see what the maximum number of sprites the 007121 will draw.  The 0x199 number was there to verify I didn't go past the end of sprite ram when counting up the number of 8x8 blocks.

The code related to the 264 limit is found here (main/src/tests/sprite.asm, but is disable from the builds)

https://github.com/jwestfall69/mad/tree/main/machine/contra

Is there a preferred way to submit the updates?  Should I re-do the existing commits so there will be a clean history or is adding commits on top of the existing ones ok?  I didn't see anything on the wiki talking about it.

happppp:
0x199 is 409 decimal, and the minimum blocks per sprite is 1, so there's no way max blocks of 264 it will exceed it and it doesn't need to be checked for in the loop.
Yes you can append new commits to existing PRs. Maybe find a tutorial on Google on how? I would tell if I knew. To be honest I only made a handful of PRs on github myself. Master branch commit rights made me lazy.

edit: We usually squash PRs when merging them, so they end up as 1 commit.  Of course, the individual commits are still visible in the PR itself, which is ok.

angelosa:
> Is there a preferred way to submit the updates? Should I re-do the existing commits so there will be a clean history or is adding commits on top of the existing ones ok? I didn't see anything on the wiki talking about it.

Pushing new commits to a fork will automatically update this PR (assuming you are over the same branch obviously). There's no wiki article because it's standard git knowledge, specifically "feature branch" and (after this gets merged) [syncing a fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)

cuavas:
> And if the Combat School note is nonsense, remove it.

It isn’t nonsense – it’s saying that Combat School changes that control register during the game, using both 0x40 and 0x80 sprite modes.

src/mame/konami/combatsc_v.cpp
```diff
@@ -334,10 +323,7 @@ void combatsc_state::pf_control_w(offs_t offset, uint8_t data)
 	}
 	if (offset == 3)
 	{
-		if (data & 0x08)
-			memcpy(m_spriteram[m_video_circuit].get(), m_videoram[m_video_circuit] + 0x1000, 0x800);
-		else
-			memcpy(m_spriteram[m_video_circuit].get(), m_videoram[m_video_circuit] + 0x1800, 0x800);
+		m_spriterambank[m_video_circuit] = (data & 0x8) >> 3;
```
cuavas: You should use the `BIT` helper for extracting a single bit.

src/mame/konami/contra.cpp
```diff
@@ -366,10 +367,7 @@ void contra_state::K007121_ctrl_w(offs_t offset, uint8_t data)
 
 	if (offset == 3)
 	{
-		if ((data & 0x8) == 0)
-			m_buffered_spriteram[Which]->copy(0x800, 0x800);
-		else
-			m_buffered_spriteram[Which]->copy(0x000, 0x800);
+		m_spriterambank[Which] = (data & 0x8) >> 3;
```
cuavas: Same here, use case for the `BIT` helper.

src/mame/konami/k007121.cpp
```diff
@@ -210,22 +215,57 @@ void k007121_device::ctrl_w(offs_t offset, uint8_t data)
...
-	// TODO: sprite limit is supposed to be per-line! (check MT #00185)
-	int num = 0x40;
-	//num = (m_ctrlram[0x03] & 0x40) ? 0x80 : 0x40; // WRONG!!! (needed by combatsc)
 
-	int inc = 5;
+	// determine number of sprites that will be drawn
+	int num_sprites = 0;
+	int sprite_blocks = 0;
+	for (int i = 0;i < MAX_SPRITES && sprite_blocks < MAX_SPRITE_BLOCKS; i+= SPRITE_FORMAT_SIZE) {
+		num_sprites++;
```
cuavas: Please use Allman formatting consistently – the rest of the file puts openng braces on their own lines.  Also, space after each semicolon in the `for` control statement, and spaces around binary operators.

happppp:
> It isn’t nonsense – it’s saying that Combat School changes that control register during the game, using both 0x40 and 0x80 sprite modes.

The comment is saying that it draws 64 sprites, and a flag used in Combat School increases it to 128. Both of which are false according to current tests.
Higher up, there's also a comment about this flag, saying that Labyrinth Runner disagrees with the Combat School assumption.

cuavas:
That just means it’s incorrect.  “Nonsense” would be meaningless or self-contradictory.  But sure, there was already a comment further down stating that it was “WRONG!” in all caps, so it should be removed.  I’m not saying it should stay, just that it’s incorrect and poorly worded, rather than nonsensical.

jwestfall69:
> edit: We usually squash PRs when merging them, so they end up as 1 commit. Of course, the individual commits are still visible in the PR itself, which is ok.

Thanks that is kinda what I was asking about.  The original 3 commits are somewhat distinct changes, each with their own lengthy comment.  I will beef of the comment on the PR since its going to get squashed.

jwestfall69:
This is the result of the max sprite test on contra hardware.

IMG_1390

Each green square is an 8x8 sprite. It just ends up bumping into the 264 max 8x8 sprite block limit.

angelosa:
fwiw is normal to have a max sprite limit and a per-scanline limit.

happppp:
Cool. Did you try other sprite sizes with that test?
So: 8x16 and 16x8 = max 132 sprites, and 16x16 = max 66 sprites?

Mixed case: Try 132 8x8 sprites and the rest of the spriteram filled with 16x16 sprites, does it draw 33 16x16 sprites?

Then for edge cases:
How many 32x32 sprites can it draw? 264/16 = 16.5, so does it draw 16 sprites, 17 sprites? Or 16 sprites, and 4 blocks of the last sprite?
What if you make it draw 263 8x8 sprites and 1 larger sprite?

@angelosa this seems more like a framebuffer, so no per-scanline sprite limit. But yeah, good to double check if it has a scanline limit too.

jwestfall69:
> fwiw is normal to have a max sprite limit and a per-scanline limit.

There doesn't seem to be a limit in the x or y axis (not which would apply to a scanline limit since the screen is rotated 90)

Each of the below tests are drawing 128 sprites 1 pixel apart alternating between green/red colored sprites.
IMG_1391
IMG_1392

jwestfall69:
> Cool. Did you try other sprite sizes with that test? So: 8x16 and 16x8 = max 132 sprites, and 16x16 = max 66 sprites?
> 
> Mixed case: Try 132 8x8 sprites and the rest of the spriteram filled with 16x16 sprites, does it draw 33 16x16 sprites?
> 

I can try some of these tonight

> Then for edge cases: How many 32x32 sprites can it draw? 264/16 = 16.5, so does it draw 16 sprites, 17 sprites? Or 16 sprites, and 4 blocks of the last sprite? What if you make it draw 263 8x8 sprites and 1 larger sprite?

I ran into this when I was originally debugging and confused me before I realized there was the 264 limit.  If a large sprite were to straddle the 264 boundary it would only draw the available 8x8 sprite blocks.  Soon as it hit the limit it stops drawing the rest of the sprite.

happppp:
> I ran into this when I was originally debugging and confused me before I realized there was the 264 limit. If a large sprite were to straddle the 264 boundary it would only draw the available 8x8 sprite blocks. Soon as it hit the limit it stops drawing the rest of the sprite.

I see, then that raises the question: which 8x8 block(s) does it draw? But I think that's too much to ask and I won't ask you to test that. It'd mean testing the block drawing order for all sprite sizes.
Knowing that it's for sure maximum 264 blocks, no matter the sprite sizes, is the most important fact.

jwestfall69:
Here are the results.  Nothing really interesting.  They all cut off exactly at 264.

IMG_1393
IMG_1394
IMG_1395
IMG_1396
IMG_1397

happppp:
Ok thanks for testing, the 264 block limit is exactly as you expected.

angelosa:
> @angelosa this seems more like a framebuffer, so no per-scanline sprite limit. But yeah, good to double check if it has a scanline limit too.

264 is a weird number: it's same as vtotal, which several Konami HWs actually uses (shaolins.cpp, simpsons.cpp, twin16.cpp, probably more concealed under `screen.set_size`). That obviously may be a coincidence, but regardless I wonder about how and when the chip operates to fetch this data.

happppp:
Yeah, perhaps it processes 1 sprite (8x8 block) per scanline.

@jwestfall69 there's a 1 frame delay on hardware between sprite processing and them appearing on the screen. I added it here: https://github.com/mamedev/mame/commit/ca106b63db5a3f2f5d4939add37bef94f5f085da
Although it's not the sprite list that's buffered, this should be sufficient. Actually buffering the rendered sprites will get complicated quickly, due to the mixing with priority mask.

jwestfall69:
@happppp here is a video the drawing order for each sprite size.  Its not drawing in 8x8 blocks for larger sprites.  Looks like it draws 64 pixels, filling up the entire length of the sprite as it goes.
 
https://github.com/user-attachments/assets/dd50c797-9d0f-4290-a74b-9cdd6cd903c3

happppp:
Thanks again for testing, I updated the notes here: https://github.com/mamedev/mame/commit/7410d18d46547f429afc9d5d4ec9208bf7c6b009

PR #13796, opened 2025-06-08
sega/model2: renderer improvements

gm-matthew:
Texture mirroring is now implemented correctly (carpet in Sonic the Fighters and pitch markings in Virtua Striker).

Fixed broken textures for trees in Sega Rally and mountain in Super GT 24h. The texture sheets are mapped as 2048x1024, with certain textures spread across both 1024x1024 halves, but stored as 1024x2048 in texture RAM.

I'll probably rewrite the texturing code when I eventually implement trilinear filtering and mipmapping, but this will do for now.

### DIFF
--- a/src/mame/sega/model2_v.cpp
+++ b/src/mame/sega/model2_v.cpp
@@ -1320,10 +1320,9 @@ void model2_state::geo_parse_np_ns( geo_state *geo, u32 *input, u32 count )
 			else luminance = fabs( dotl );
 
 			luminance = (luminance * texparam->diffuse) + texparam->ambient;
-			luma = (int32_t)luminance;
+			luminance = std::clamp(luminance, 0.0f, 255.0f);
 
-			if ( luma > 255 ) luma = 255;
-			if ( luma < 0 ) luma = 0;
+			luma = (int32_t)luminance;
 
 			/* add the face bit to the luma */
 			luma += face;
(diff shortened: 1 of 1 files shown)


src/mame/sega/model2_v.cpp
```diff
@@ -1481,10 +1477,11 @@ void model2_state::geo_parse_np_s( geo_state *geo, u32 *input, u32 count )
 			specular *= texparam->specular_scale;
 
 			luminance = (luminance * texparam->diffuse) + texparam->ambient + specular;
-			luma = (int32_t)luminance;
 
-			if ( luma > 255 ) luma = 255;
-			if ( luma < 0 ) luma = 0;
+			if (luminance > 255.0f) luminance = 255.0f;
+			if (luminance < 0.0f) luminance = 0.0;
```
ajrhacker: This would be a good place to use std::clamp.

src/mame/sega/model2_v.cpp
```diff
@@ -1839,10 +1836,11 @@ void model2_state::geo_parse_nn_s( geo_state *geo, u32 *input, u32 count )
...
```
ajrhacker: Same here.

vmicka72:
Hello matthew. I'm following model 2 progress with much interest (Sega Rally Arcade is my favorite game of all time :-). Do you plan to implement texture filtering and sound/music fix in futur?

PR #13798, opened 2025-06-09
roland_tr707.cpp, video/hd61602.cpp: LCD emulation.

m1macrophage:
* Emulated HD61602 device.
* Incorporated it into the 707 driver and layout.
* Corrected tempo potentiometer curve.
* Made dinsync testable.
* Marked systems as supporting save.
* Minor layout cleanup.


happppp:
Could you move HD61602 to video/hd61602?
I don't think much if anything at all can be shared between MAME's implementation of *02 and *03. I don't specifically mean your implementation of it, but the chip itself. If you glance over the datasheet, you might think they're similar enough, but study it more (or emulate it like you did), then you realize the access mode and duty cycle stuff makes it very different.

m1macrophage:
Agreed. Moved to video/hd61602.

src/devices/video/hd61602.cpp
```diff
@@ -0,0 +1,188 @@
...
+	// 1-byte nop command
+	if (m_count == 0 && (data & 0xc0) == 0xc0)
+		return;
+
+	m_data = (m_data << 8) | data;
+	m_count = (m_count + 1) & 1;
+	if (m_count != 0)  // Waiting for the next byte.
+		return;
+
+	switch (BIT(m_data, 14, 2))
```
angelosa, line 122: I guess `case 3` is undocumented/reserved? May still be a good idea to log it out.

m1macrophage: I added a comment. Case 3 is handled at the top of the function, because it requires skipping the rest of the logic.

happppp:
Where are you getting the hardcoded framerates from? It should be determined by the OSC pin(s), no?

m1macrophage:
They are taken from the datasheet, for a 100KHz clock.  100KHz is the nominal frequency of the internal oscillator, and the recommended frequency for an external one. The datasheet only provides frame rates for a 100KHz clock. An external clock is not emulated right now.

I updated comments in the code to reflect the above, and I documented functionality that is not emulated.

m1macrophage:
One thing to add: the datasheet gives the impression that this controller should always run at 100KHz. So I just lifted the values reported there, to make the correspondence with the datasheet obvious. But I can "derive" those frame rates by dividing the 100KHz clock (which will be a const for now), if that's preferable compared to the added comments.

happppp:
TODO note is ok.
I don't think the new CI errors are yours, it's the one that was fixed here: https://github.com/mamedev/mame/commit/9f810075abba80d63e06b79e6da8b487e3354ec6

PR #13800, opened 2025-06-09
Add two Xor World sets (different hardware, not working)

clawgrip:
Xor World (different hardware, ver 1.1, checksum 3333BA, prototype) [Josep Quingles, Recreativas.org, David Haywood]

New clones marked not working
-----------------------------
Xor World (different hardware, ver 1.1, checksum 333462, prototype) [Josep Quingles, Recreativas.org, David Haywood]

### DIFF
--- a/src/mame/gaelco/gaelco.cpp
+++ b/src/mame/gaelco/gaelco.cpp
@@ -10,6 +10,7 @@ Games supported:
 
 Year   Game                PCB                       NOTES
 ------------------------------------------------------------------------
+1990   Xor World           ?                         Prototype
 1991   Big Karnak          REF 901112-1              Unprotected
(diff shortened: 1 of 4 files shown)


mamehaze:
(I made a few slight modernisations in here at the same time as I looked at this for Clawgrip, eg replacing the VIDEO_START override stuff with video_start and some additional classes)

I've left it as not a clone, because it seems like the existing 'xorworld' set was pretty much a complete rebuild targetting very different hardware so it doesn't feel natural to call them clones of each other.

An argument could be made for splitting gaelco.cpp into separate drivers anyway as they are all different PCBs, but it would be a lot of duplicated code for boards that are otherwise very close to each other.

PR #13802, opened 2025-06-09
improvements to the modular system version of XorWorld, promote to working

mamehaze:
(also fixed what seem to be some copy+paste bugs in the inputs of Splash Modular system)

machines promoted to working / new working machines
--------------------
Xor World (Modular System, prototype, set 1) [Josep Quingles, Recreativas.org, David Haywood]

clones promoted to working / new working clones
--------------------
Xor World (Modular System, prototype, set 2) [Josep Quingles, Recreativas.org, David Haywood]


mamehaze:
@Osso13 I've done this to assist with Clawgrip in sorting out this Gaelco stuff.

I don't know if you want to take this and try adding sound / other bits, probably similar to Splash Modular System.

mamehaze:
much like the 'flat' PCB versions of Xor World added in https://github.com/mamedev/mame/pull/13800 I've decided to not have this set as a clone.

it really seems like these were 3 different projects on 3 different pieces of hardware; 3 different attempts to create a game that ultimately never made it to market in any form.

mamehaze:
after discussion with Osso over time availability I've just added the Splash sound hardware to this.  Music tempo seems a bit fast at least, but it's in the ballpark of how this should work.

PR #13803, opened 2025-06-10
patinhofeio internal layout: fix encoding of extended ascii strings

felipesanches:
Some of these Brazilian Portuguese strings were broken probably due to automated source cleanup. Hopefully, by using markup that references the unicode codepoints of these characters explicitly, the strings won't break again.


happppp:
I see, yes that got broken by automatic source clean a long time ago. It's not restricted anymore nowadays, so you can use Ó Ç Ã these days. Makes it more readable too, see for example Russian text here: https://github.com/mamedev/mame/blob/master/src/mame/layout/intellect02.lay#L29

PR #13810, opened 2025-06-11
sinclair/timex: Batched Timex's drivers

holub:
I see it this way...


ajrhacker:
I don't think this makes much sense. Timex doesn't seem to have had anything to do with either the QL or Z88.

holub:
> I don't think this makes much sense. Timex doesn't seem to have had anything to do with either the QL or Z88.
That's the way I see it: https://www.timexsinclair.com/company/sinclair-research-ltd/

ajrhacker:
"Timex Sinclair" is the name of that website and what appears to be a short-lived joint venture between the two companies which doesn't seem to have produced many computers other than a rebadged ZX81 and the ZX Spectrum variants in the timex.cpp driver.

holub:
What's your suggestion? Keep everything as is, just move zx88 to common pile?

holub:
v2: #13812

PR #13814, opened 2025-06-12
model1/model 2: TGP sin/cos lookup tables

gm-matthew:
Calculating `index ^= 0x3fff` when `ang & 0x4000` was leading to results that were not always completely accurate; for example, calculating sin when ang = 0x2000 (equivalent to 45 degrees) gives a result of 0.70710677 which is accurate to single precision, but ang = 0x6000 (equivalent to 135 degrees) gives a result of 0.707039 which is very slightly off. This tiny error wasn't making any discernible difference in most games, but for some reason it was messing up Daytona; at very specific angles the player car would move sideways during a frame instead of forwards, and the camera would rotate to follow the car moving sideways:

image

Changing the calculation to `index = std::min(0x4000 - (int)index, 0x3fff)` makes it so the results are always accurate, and the above bug no longer occurs. Clamping the index to 0x3fff is required because 0x4000 would end up reading from the atan2 lookup table.

I have made the change for both Model 1 and Model 2, since I presume the logic on each is the same.

### DIFF
--- a/src/mame/sega/model1_m.cpp
+++ b/src/mame/sega/model1_m.cpp
@@ -183,8 +183,8 @@ u32 model1_state::copro_sincos_r(offs_t offset)
 {
 	offs_t ang = m_copro_sincos_base + offset * 0x4000;
 	offs_t index = ang & 0x3fff;
-	if(ang & 0x4000)
-		index ^= 0x3fff;
+	if (ang & 0x4000)
+		index = std::min(0x4000 - (int)index, 0x3fff);
 	u32 result = m_copro_tables[index];
 	if(ang & 0x8000)
 		result ^= 0x80000000;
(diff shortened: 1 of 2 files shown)


galibert:
I honestly wouldn't be surprised if the bug happened on the real arcade though.

gm-matthew:
> I honestly wouldn't be surprised if the bug happened on the real arcade though.

I've played this game a lot on real hardware and have never experienced the bug, nor have I ever observed it in any videos of the game running on real hardware.

mamehaze:
is this related to the weird 'wheelie down the hill' thing the cars do too, or is that something else?

gm-matthew:
> is this related to the weird 'wheelie down the hill' thing the cars do too, or is that something else?

That's due to an i960 bug that I actually submitted a fix for but it ended up being reverted.

angelosa:
https://github.com/mamedev/mame/pull/13507 ?

gm-matthew:
> #13507 ?

Yes, that's it

MooglyGuy:
Why not create an i960 subclass for the 80960KB specifically? As you point out in #13507, the manual for the 80960KB in particular does not mention the sign-bit preservation logic, so it stands to reason that the change was correct, albeit only for the model of i960 used on Model 2.

gm-matthew:
> Why not create an i960 subclass for the 80960KB specifically?

Looks like ajrhacker has just done this: 85c3477 :-)

PR #13817, opened 2025-06-13
contra: add missing watchdog

jwestfall69:
Confirmed on hardware that both reads and writes will ping the watchdog.  Additionally reading the watchdog address always returns 0xff.


jwestfall69:
hrm borked something left me fix

PR #13818, opened 2025-06-13
contra: add missing watchdog

jwestfall69:
Confirmed on hardware that both reads and writes will ping the watchdog.  Additionally reading the watchdog address always returns 0xff.

### DIFF
--- a/src/mame/konami/contra.cpp
+++ b/src/mame/konami/contra.cpp
@@ -147,6 +147,7 @@ Notes:
 #include "cpu/m6809/hd6309.h"
 #include "cpu/m6809/m6809.h"
 #include "machine/gen_latch.h"
+#include "machine/watchdog.h"
 #include "sound/ymopm.h"
 #include "video/bufsprite.h"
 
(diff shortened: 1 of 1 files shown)


jwestfall69:
I also did some measurements with a logic analyzer to see how long it takes the watchdog to trigger.  From the last write to the watchdog address to the reset line going low I got these results.

314.76 ms
315.79 ms
316.31 ms
317.21 ms
317.33 ms
317.63 ms

happppp:
Instead of measuring the time from manual watchdog reset to watchdog trigger, could you measure the frequency of watchdog resets if you don't r/w from watchdog at all?

Screenshot (770)
furrtek's schematics.
This is what, 16 vblanks? It does not match your measurement.

Since you have a logic analyzer could you also measure vblank and hblank frequency? What MAME currently has is an educated guess:
hblank: 24000000 / 3 / 512 = 15625
vblank = hblank / 264 = ~59.18

jwestfall69:
If I run the board without a rom I get this for the cpu reset line
contra_watchdog

hblank
contra_hblank

vblank
contra_vblank

happppp:
Ok thanks for measuring that, it looks like vblank/hblank timing agrees with MAME.
The watchdog period still isn't a multiple of vblanks though, which is odd.

galibert:
It's very close to 524288 cycles (2**19) of the 3Mhz clock though.  The reset time seems to be around 409600 cycles, which is a little stranger but not insane.

jwestfall69:
I've been reviewing the 007121 and contra schematics.  The watchdog reset comes from the 005924 component. Its a vertical board surrounded by a rubber coating.  It appears to have a number of small IC's on it.  I can see it getting my watchdog pings as well as it triggering resets if I don't ping.  I'm not seeing any clock signal going into it, so maybe one of its small ICs is a clock?  Perhaps caius has more info since he made a reproduction of the part.

The section of the schematic in your screen shot appears to just be a delay for now long the 007121 will hold the cpu RESET low after its receives a reset on its external reset pin (NEXR) from the 005924.  I believe CPU reset gets held low for 8 vblanks + however long the 005924 held it low.  This matches up very close for how long the reset is low (136ms) in my measurement.  Course how long before a watchdog triggers would depend on the 005924.

happppp:
Ah, I misunderstood, so you're saying it's this guy:
AFE = 0x1e in contra
no clock input, so watchdog timing is internal to the chip.

Screenshot (773)

jwestfall69:
Yes exactly, but 005924 is a daughter board and not a chip.

IMG_1400
IMG_1401

PR #13819, opened 2025-06-13
lgtnfght: removed NMI connection as it does not exist on schematics.

jotego:
Trigon schematics show a pull-up resistor at the NMI pin. However, the driver connects that signal.

image

This PR removes the connection in the driver. The sound is not affected after the change.

I created a separate branch in my fork for this, so feel free to push other changes to it.


cuavas:
Just as a sanity check, have you checked continuity on a real board to ensure it isn’t an error in the schematic?  It could have a pullup resistor as well as a connection if it’s being driven from an open collector output, for example.

happppp:
No NMI routine at 0x66 in the Z80 ROM.
If the hw did have NMI hooked up and not used it, I'd expect at least a retn safety measure like some of the other games in this driver.

PR #13823, opened 2025-06-13
sinclair/tsconf.cpp: Changed default ram data to 0x00; cram full init

holub:


src/mame/sinclair/tsconf.cpp
```diff
@@ -231,6 +231,27 @@ void tsconf_state::machine_reset()
...
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x0000, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x2108, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x0000, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x2108, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x0000, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x2108, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318,
+		0x0000, 0x0010, 0x4000, 0x4010, 0x0200, 0x0210, 0x4200, 0x4210, 0x0000, 0x0018, 0x6000, 0x6018, 0x0300, 0x0318, 0x6300, 0x6318
+	};
+	for (auto i = 0; i < 0x100; i++)
+		cram_write16(i << 1, (cram_init[i] >> 8) | (cram_init[i] << 8)); // init color RAM
```
cuavas: Where does this come from?  Can you load it as a “ROM” rather than having it in an array here?

holub: https://github.com/tslabs/zx-evo/blob/master/pentevo/fpga/current/video/mem/video_cram.mif

I'm planning to load it as a ROM eventually, but that requires entire cram refactoring and change it to proper 16bit data size. For now it just patches what we have.

cuavas: You don’t need to change the RAM width to load it as a ROM, you could load it as a 16-bit ROM region, and then just do the same thing here, like:
```c++
	u16 const *const cram_init = &memregion("cram_init")->as_u16();
	for (auto i = 0; i < 0x100; i++)
		cram_write16(i << 1, (cram_init[i] >> 8) | (cram_init[i] << 8)); // init color RAM
```
or something similar.

cuavas: Edited the code fragment in the previous comment.  Commenting on three PRs at once is a bad idea.

holub: It's fine, I'll figure out. It is only required for spg. For regular boot BIOS update palette.

holub:
here you go

PR #13827, opened 2025-06-14
heathzenith/z100.cpp: Fix 8085 speed and minor cleanup

mgarlanger:
Fixes:
- Correct 8085 CPU speed to 5 MHz
- add save state (save_item) calls.

Cleanup:

- Add more background comments
- Expand TODO list
- Consistent spacing in existing comments
- Document motherboard jumpers
- Code had mixture of uint8_t/u8 declarations, standardize to u8. 
- Convert most of the bit masking to use `BIT()`
- Convert single line comments to `//` style
- Add variables to avoid continually reading config ports
- Update dip switch name(to match schematics)/update settings
- Create member variables for devices used in multiple places.

### DIFF
--- a/src/mame/heathzenith/z100.cpp
+++ b/src/mame/heathzenith/z100.cpp
@@ -4,129 +4,156 @@
 
     Zenith Z-100
 
+    The Z-100 was the assembled version of the Heath H-100, which came
+    in kit form. There were two main versions of the computer, a low-profile,
+    without a CRT monitor, and the "all-in-one", with a monochome CRT.
+
+    Initially model numbers had standardized numbering with a two or three
+    letter prefix, a 3 digit base, and a two digit suffix. The first letter
+    of the prefix was either Z for Zenith (assembled) or H for Heath (kit).
+    The second letter specified the disk drives: F for 48 tpi DS floppy or
+    W for Winchester (hard-drive). An early version of the service manual
+    specified 3 more letters, but not sure if they were ever sold: X for
+    Customer OEM System, H for 96 tpi DS floppy, and E for 48 tpi SS Floppy.
+    There are 4 base numbers: 100 for low-profile with monochome video board,
+    110 for low-profile with color video board, 120 for all-in-one with
+    monochome video, and 130 for all-in-one with color video board (for
+    use with an external color monitor). For the suffix, the first number
+    is the number of 64k memory banks, and the second number is the number
+    of disk drives. Once new versions of the motherboard, which supported
+    higher CPU speeds(8 MHz for the 8088), and higher capacity memory chips,
+    model numbers deviated from the format specified above.
+
     15/07/2011 Skeleton driver.
 
     Commands:
     Press DELETE to abort boot, then press H to list all the commands.
 
     TODO:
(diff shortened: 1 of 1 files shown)


src/mame/heathzenith/z100.cpp
```diff
@@ -243,126 +326,144 @@ class z100_state : public driver_device
...
+	save_item(NAME(m_gbank));
+	save_item(NAME(m_display_mask));
+	save_item(NAME(m_flash));
+	save_item(NAME(m_clr_val));
+	save_item(NAME(m_tmr_status));
+	save_item(NAME(m_start_addr));
+	save_item(NAME(m_vidint_enable));
+	save_item(NAME(m_memory_ctrl));
+
+	m_parity = make_unique_clear<u32[]>(m_ram.bytes() / 32);
```
cuavas: Shouldn’t the contents of parity RAM be saved as well?

src/mame/heathzenith/z100.cpp
```diff
@@ -243,126 +326,144 @@ class z100_state : public driver_device
...
 }
 
 void z100_state::video_start()
 {
-	m_gvram = make_unique_clear<uint8_t[]>(0x30000);
+	m_gvram = make_unique_clear<u8[]>(0x30000);
```
cuavas, line 376: Same with this – shouldn’t it be included in saved states?

src/mame/heathzenith/z100.cpp
```diff
@@ -649,31 +766,31 @@ INPUT_PORTS_START( z100 )
...
-	PORT_DIPSETTING(    0x80, "PAL 50 Hz" )
+	PORT_DIPNAME( 0x10, 0x00, "Not used" )
+	PORT_DIPSETTING(    0x00, DEF_STR( Off ) )
+	PORT_DIPSETTING(    0x10, DEF_STR( On ) )
+	PORT_DIPNAME( 0x20, 0x00, "Not used" )
+	PORT_DIPSETTING(    0x00, DEF_STR( Off ) )
+	PORT_DIPSETTING(    0x20, DEF_STR( On ) )
+	PORT_DIPNAME( 0x40, 0x00, "Not used" )
+	PORT_DIPSETTING(    0x00, DEF_STR( Off ) )
+	PORT_DIPSETTING(    0x40, DEF_STR( On ) )
```
cuavas: For unused DIP switches, you should use `DEF_STR( Unused )` for consistency.

src/mame/heathzenith/z100.cpp
```diff
@@ -363,8 +368,8 @@ void z100_state::machine_start()
 	save_item(NAME(m_start_addr));
 	save_item(NAME(m_vidint_enable));
 	save_item(NAME(m_memory_ctrl));
-
-	m_parity = make_unique_clear<u32[]>(m_ram.bytes() / 32);
+	save_pointer(NAME(m_gvram), 0x30000);
```
cuavas: You should register this for saving in the same function where it’s allocated.  If you don’t need to separately override `video_start` in different derived classes, you can get rid of it altogether and allocate everything here in `machine_start`.

Issue #13829, opened 2025-06-15
ghostb nmi behavior (+ 6x09 cpu bug?)


jwestfall69:
I've been writing some code that runs on the ghostb hardware and came across an issue with NMI behavior in MAME.  On hardware if I enabled NMI it will pulse on its own ever vblank, while in MAME it only triggers once.  MAME seems to be treating the NMI enable/disable as an ack to the NMI, which it isn't on hardware.

https://github.com/mamedev/mame/blob/f3abbc8145f3d6850c9e8ab1e5d06c03a54de1bd/src/mame/dataeast/dec8.cpp#L192

I couldn't find any schematics for ghostb board, but worked backwards from the NMI pin of the 6309 CPU.  I comes from the 74LS00 @ 13c (dual input NAND gate).  Input 1 is vblank, input 2 comes from the 74LS273 @ 3a (latch).  This latch gets updated when a write happens to 0x3840, which is what you write to to enable/disable NMI.  So there is no hardware logic to ack NMIs.

I tried updating the dec8.cpp to have ghostb mimic the same latch behavior as gondo with its m_nmigate.  This fixes the behavior on my code, but causes breakage in ghostb.  Looking into the breakage in ghostb I think its a bug in the 6x09 cpu code.  For whatever reason ghostb disables NMI when it enters its NMI handler and then will re-enabled it (bp 89c9) a little before it exits the NMI.  On the instruction after re-enabling NMIs (89fc) the cpu seems to be re-triggering NMI even though its already in NMI.  This happens over and over causing the stack to fill up ram.

jwestfall69:
Here is a diff of the ghostb nmi changes

[mame-ghostb-nmi.txt](https://github.com/user-attachments/files/20743614/mame-ghostb-nmi.txt)

happppp:
NMI on most CPUs (AFAIK including 6809 family) really is non-maskable. So if NMI pin triggers again while the software hasn't done an RTI yet, yes it jumps to the NMI vector again.

jwestfall69:
Interesting, I was not aware of that.  Seems like the same thing should be happening on hardware if it's re-enabling the NMI while still in vblank.   I will poke the hardware and try to figure out whats up.

jwestfall69:
Seems like its a timing thing.  On hardware the NMI gets re-enabled a little after VBLANK is over so it doesn't re-trigger another NMI.

[...]

PR #13832, opened 2025-06-15
ghostb: there is no ACK for NMIs

jwestfall69:
NMI on the 6309 comes from the 74LS00 @ 13C (dual input NAND gate) where input 1 is vblank and input 2 comes from a latch (74LS273 @ 3A). The latch gets updated with writes to 0x3840, which is what you write to to enable/disable NMI.  So there is no hardware logic to ACK NMIs.


happppp:
ref: https://github.com/mamedev/mame/issues/13829

PR #13833, opened 2025-06-15
New Working Clone

Hydreigon223:
Gauntlet Legends (version 1.4)

Changed warfa and nbanfl to use the different bootroms as BIOS options (in the process, this removes set warfac since the only the bootrom differs from its parent set). I myself had to make changes in 2.7.u27 to reflect the changes made in its former driver_init so it could be used as a BIOS option.

### DIFF
--- a/src/mame/midway/vegas.cpp
+++ b/src/mame/midway/vegas.cpp
@@ -2362,6 +2362,7 @@ void vegas_state::cartfury(machine_config &config)
 
 // there is a socket next to the main bios roms for updates, this is what the update region is.
 
+// TODO: Bootroms for gauntleg should be configured as BIOS options at some point while the disk update roms should be configured as a software list instead.
 
 ROM_START( gauntleg )
 	ROM_REGION32_LE( 0x80000, PCI_ID_NILE":rom", 0 )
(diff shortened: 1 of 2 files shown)


src/mame/midway/vegas.cpp
```diff
@@ -2741,14 +2766,6 @@ void vegas_state::init_nbashowt()
...
 {
 	m_maincpu->mips3drc_set_options(MIPS3DRC_FASTEST_OPTIONS | MIPS3DRC_STRICT_VERIFY | MIPS3DRC_EXTRA_INSTR_CHECK);
-	// The first three bytes of the blitz00_nov30_1999.u27 ROM are FF's which breaks the reset vector.
-	// These bytes are from blitz00_sep22_1999.u27 which allows the other ROM to start.
-	// The last byte which is part of the checksum is also FF. By changing it to 0x01 the 4 byte checksum matches with the other 3 changes.
-	uint8_t *romPtr = memregion(PCI_ID_NILE":rom")->base();
-	romPtr[0x0] = 0xe2;
-	romPtr[0x1] = 0x00;
-	romPtr[0x2] = 0xf0;
-	romPtr[0x7ffff] = 0x01;
```
angelosa, line 2751: Is `nbanfl` working by removing this patch?

Hydreigon223: BIOS version 2.5 and 2.6 work. BIOS version 2.7 normally wouldn't work without the patch but I myself edited the rom file to reflect changes from the driver init so it works now.

Deleted my last comment as I didn't mean to start a review in a weird fashion.

PR #13838, opened 2025-06-16
Ensure error.log is flushed on Windows OSD

MooglyGuy:
The comment in `win_osd_file::flush`:
`// shouldn't be any userspace buffers on the file handle`
is not accurate.

Per Microsoft's documentation for Fileapi.h and the `FlushFileBuffers` function:
> Typically the [WriteFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-writefile) and [WriteFileEx](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-writefileex) functions write data to an internal buffer that the operating system writes to a disk or communication pipe on a regular basis. The FlushFileBuffers function writes all the buffered information for a specified file to the device or pipe.

This tracks with the behavior observed when MAME crashes while running with `-log`: The log file is often truncated.

The buffered amount can be significant; when debugging a crash in a Taito Type-Zero game, MAME could run for upwards of a minute prior to crashing, with nothing written to error.log due to even several kilobytes of text data being insufficient to result in an automatic flush to disk.

While this has a performance cost when hitting the error log with extreme frequency, being able to have a complete view of what was written to the log in the event of a crash can, itself, be useful in diagnosing said crash.

### DIFF
--- a/src/osd/modules/file/winfile.cpp
+++ b/src/osd/modules/file/winfile.cpp
@@ -107,8 +107,10 @@ public:
 
 	virtual std::error_condition flush() noexcept override
 	{
-		// shouldn't be any userspace buffers on the file handle
-		return std::error_condition();
+		if (!FlushFileBuffers(m_handle))
+			return win_error_to_error_condition(GetLastError());
+		else
+			return std::error_condition();
 	}
 
 private:


cuavas:
Does this only flush userspace buffers, or does it cause an actual filesytem sync (i.e. flushing the kernel’s cache to storage)?  If it does the latter, it’s probably going to cause performance to tank for anything that sends a significant volume of messages to `logerror`.  (Putting a `syncfs` call at the equivalent place in posixfile.cpp made it pretty much unusable.)

galibert:
Due to disk caching interactions within the system, the FlushFileBuffers function can be inefficient when used after every write to a disk drive device when many writes are being performed separately. If an application is performing multiple writes to disk and also needs to ensure critical data is written to persistent media, the application should use unbuffered I/O instead of frequently calling FlushFileBuffers. To open a file for unbuffered I/O, call the [CreateFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea) function with the FILE_FLAG_NO_BUFFERING and FILE_FLAG_WRITE_THROUGH flags. This prevents the file contents from being cached and flushes the metadata to disk with each write. For more information, see [CreateFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea).

Yeah, wrong way to do it.

MooglyGuy:
Agreed, this was the wrong way to go. Closing.

A correct solution will be to add `OPEN_FLAG_UNBUFFERED` to the osdfile.h file-open flags. However, I'll need assistance from a dev on Linux and a dev on macOS to help with those respective platforms.

cuavas:
We can’t use `FILE_FLAG_NO_BUFFERING` without a comprehensive rewrite of that class, because it requires you to read/write whole sectors at a time, and the I/O buffers you provide must also be aligned to the sector size.  It doesn’t cause the metadata cache to be flushed after every write, and IIRC doesn’t stop the disk cache from being used.

`FILE_FLAG_WRITE_THROUGH` causes an implicit data *and metadata* flush after every write, but still caches data for reads.

I guess the ideal solution would be to rewrite the class the use `FILE_FLAG_NO_BUFFERING` and manage its own sector cache, but that would be a lot of work.  That makes the question whether this is a good stop-gap measure, or if it’s going to hurt performance, too much.

MooglyGuy:
Depends on how much performance impact is too much. Performance was perceptibly impacted when I was logging upwards of ~128-160 characters for every vertex written by the Taito Type-Zero driver, but is that the common case?

Generally my expectation is that when I flip every single possible `LOG_*` on in a modern 3D driver that I'll probably have a performance slog, file flushing or not. In fact, I often end up eating this performance cost by default due to peppering additional `printf("blah"); fflush(stdout);` throughout my code when working on a problem.

If a driver is spamming `logerror()` enough that there's a noticeable performance loss outside the scope of active development (i.e. just running stock MAME with `-log`), then that is probably something that should be addressed on a per-device or per-driver basis.

cuavas:
> Agreed, this was the wrong way to go. Closing.
> 
> A correct solution will be to add `OPEN_FLAG_UNBUFFERED` to the osdfile.h file-open flags. However, I'll need assistance from a dev on Linux and a dev on macOS to help with those respective platforms.

The POSIX implementation uses the unistd.h functions which do no userspace buffering at all – the only buffering you get is what `core_osd_file` (from src/lib/util/corefile.cpp) provides on top of it.

The trouble is, from a quick look, there doesn’t seem to be a simple way to bypass the buffering.  The options are:
* Calling `FlushFileBuffers` which also flushes data and metadata caches
* Using `FILE_FLAG_WRITE_THROUGH` which causes an implicit data and metadata flush after every write
* Using `FILE_FLAG_NO_BUFFERING` which requires you to do your own userspace caching to ensure you only read/write whole sectors

cuavas:
Previous comment should say “…on Windows, there doesn’t seem to be…”

MooglyGuy:
I don't feel strongly about it either way, I just happened to sidetrack into working out why my error logs were being cut off while iterating on taitotz.cpp's 3D support and ending up with a crash.

Either way, the comment should probably be removed from flush() in src/osd/modules/file/winfile.cpp, as it's inaccurate.

cuavas:
Sure, I’ll change the comment next time I push something.

MooglyGuy:
Cheers, much appreciated.

PR #13845, opened 2025-06-18
posixfile.cpp: remove redundant realpath() call

pinotree:
Commit 1bd116c5c8f8f3d28e4db7925efcb87279ba860b makes use of the allocating mode of `realpath()` by duplicating the `realpath()` invocation already done (i.e. with the same input path). In practice, if the `realpath()` invocation with `nullptr` fails, the `realpath()` invocation with a destination buffer is very likely to fail in the same way.

Hence drop the `realpath()` invocation with a destination buffer, and move the `realpath()` invocation with `nullptr` in the existing `if`/`else if` chain to maintain the existing `behaviour`.

Since `path_buffer` now is needed only for the fallback `getcwd()` invocation, move it in the inner scope where it is used. Since the usage of `getcwd()` already has a growing buffer logic, make the initial buffer size shorter, which should be enough to get current working directory already at the first attempt in most of the cases.

### DIFF
--- a/src/osd/modules/file/posixfile.cpp
+++ b/src/osd/modules/file/posixfile.cpp
@@ -425,13 +425,6 @@ std::error_condition osd_get_full_path(std::string &dst, std::string const &path
 			dst = canonical.get();
 			return std::error_condition();
 		}
-
-		std::vector<char> path_buffer(PATH_MAX);
-		if (::realpath(path.c_str(), &path_buffer[0]))
-		{
-			dst = &path_buffer[0];
-			return std::error_condition();
-		}
 		else if (path[0] == PATHSEPCH)
 		{
 			dst = path;
@@ -439,6 +432,7 @@ std::error_condition osd_get_full_path(std::string &dst, std::string const &path
 		}
 		else
 		{
+			std::vector<char> path_buffer(128);
 			while (!::getcwd(&path_buffer[0], path_buffer.size()))
 			{
 				if (errno != ERANGE)


src/osd/modules/file/posixfile.cpp
```diff
@@ -425,20 +425,14 @@ std::error_condition osd_get_full_path(std::string &dst, std::string const &path
...
 			return std::error_condition();
 		}
-
-		std::vector<char> path_buffer(PATH_MAX);
-		if (::realpath(path.c_str(), &path_buffer[0]))
-		{
-			dst = &path_buffer[0];
-			return std::error_condition();
-		}
 		else if (path[0] == PATHSEPCH)
```
rb6502, line 428: Should this still have the else, which belongs to a now-removed if ()?

pinotree: This is a chain like
```c++
if (cond1) {
   // code for cond1
   return ...
} else if (cond2) {
   // code for cond2
   return ...
} else {
   // code for else
   return ...
}
```
which is equivalent to
```c++
if (cond1) {
   // code for cond1
   return ...
}
if (cond2) {
   // code for cond2
   return ...
}
// code for else
return ...
```
I replaced the now-removed `if` called `realpath()` with a buffer with the `if` that calls `realpath(..., nullptr)`, as it does the same thing, keeping the `if` chain as it was for consistency.

cuavas:
The real purpose of this PR seems to be to get rid of `PATH_MAX`, which as far as I can tell is just to build on GNU Hurd.  I don’t think this is beneficial.  After 35 years of development, GNU Hurd still has no meaningful presence as a desktop OS (or in any other OS segment for that matter).  I don’t think we should be pandering to an OS that no-one uses.  I mean, we’ve dropped support for Windows XP and older versions of macOS that have orders of magnitude more users than GNU Hurd.

PR #13851, opened 2025-06-20
sega/model2.cpp: sound comms

gm-matthew:
Most revisions of Model 2 feature an NEC D71051C (Intel 8251 compatible) serial controller which is used to communicate with the Model 1 sound board and/or the SCSP; this is accessed over the 16-bit secondary bus at address 0x1C80000.

But the Model 2B-CRX does not feature a standalone D71051C. Instead it has an 8251-compatible serial controller built into the 315-5797 buffer controller (Model 2B manual chapter 6), which is accessed over the 32-bit primary bus at address 0x9C0000.

All Model 2 games have been tested and have sound working.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -964,41 +964,20 @@ void model2_state::model2_check_irqack_state(u32 data)
 	}
 }
 
-/* TODO: rewrite this part. It's a 8251-compatible chip */
-u32 model2_state::model2_serial_r(offs_t offset, u32 mem_mask)
+u8 model2_state::model2_serial_r(offs_t offset)
 {
-	if (offset == 0)
-	{
-		u32 result = 0;
-		if (ACCESSING_BITS_0_7 && (offset == 0))
-			result |= m_uart->data_r();
-		if (ACCESSING_BITS_16_23 && (offset == 0))
-			result |= m_uart->status_r() << 16;
-		return result;
-	}
-
-	return 0xffffffff;
+	return m_uart->data_r();
 }
 
 
-void model2_state::model2_serial_w(offs_t offset, u32 data, u32 mem_mask)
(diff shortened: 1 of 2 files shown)


angelosa:
Virtua Striker have few SFXs that keeps repeating (i.e. GK rejection on shoots, start a two player game and shoot straight at full power around 3/4 of field to quickly repro), is it unrelated to this?

gm-matthew:
> Virtua Striker have few SFXs that keeps repeating (i.e. GK rejection on shoots, start a two player game and shoot straight at full power around 3/4 of field to quickly repro), is it unrelated to this?

This occurs in the official MAME 0.277 build so it seems to be unrelated to the changes in this commit.

rb6502:
I believe @angelosa is asking if this fixes that problem.  I assume from your response that answer is no, though?

gm-matthew:
> I believe @angelosa is asking if this fixes that problem. I assume from your response that answer is no, though?

Ah okay, this commit does not fix that problem. I suspect it is most likely an SCSP emulation bug.

gm-matthew:
The sound that keeps repeating also (sometimes) does so in the sound test, where it is called SD_PUNCHING. Definitely an SCSP issue.

vmicka72:
Does this corrrect sounds in sega rally?

gm-matthew:
> Does this corrrect sounds in sega rally?

No. That requires fixes to the SCSP emulation which I am not currently working on.

If one of my commits fixes a particular issue, I will usually say so in the description.

vmicka72:
> > Does this corrrect sounds in sega rally?
> 
> No. That requires fixes to the SCSP emulation which I am not currently working on.
> 
> If one of my commits fixes a particular issue, I will usually say so in the description.

Thanks ! Do you plane to work on texture filtering? It's the big visually feature that miss actually in my opinion :-)

PR #13852, opened 2025-06-20
add some inputs and outputs as well as osd_file socket fix

SailorSat:
- added "seat" input as mappable control to harddriv/racedriv.
- added drive command output for gticlub hardware.
- added outputs to namcos22
- made osd_socket accepts to return the same as reading with 0 bytes available, instead of returning no-error 0, which *might* be confused with end-of-stream.

( this is a spin-off of #13421 )


src/mame/konami/gticlub.cpp
```diff
@@ -440,6 +440,11 @@ void gticlub_base_state::sysreg_w(offs_t offset, uint8_t data)
 			m_pcb_digit[offset] = bitswap<7>(~data,0,1,2,3,4,5,6);
 			break;
 
+		case 2:
+			// GTI club drive commands
+			m_pcb_digit[offset] = data;
+			break;
```
cuavas, line 448: This shouldn’t be a “digit” output if it isn’t one of the digital displays.

src/mame/namco/namcos22.cpp
```diff
@@ -2816,9 +2816,19 @@ void namcos22_state::handle_driving_io()
...
 }
 
+/* TODO: REMOVE (THIS IS HANDLED BY "IOMCU") */
+void namcos22_state::handle_output_io()
+{
+	// outputs
+	m_mcu_out[0] = m_shareram[0x20/2] & 0xff; // lamps, coin counters
+	m_mcu_out[1] = m_shareram[0x40/2] & 0xff; // drive commands (raverace, acedrive, victlap)
+	m_mcu_out[2] = m_shareram[0x42/2] & 0xff; // led outputs (acedrive, victlap)
+}
```
cuavas: If you name like `handle_x_io` it’s going to be misleading because it looks like another fake MCU interrupt handler like `handle_driving_io` or `handle_cybrcom_io`.

Also, the `m_mcu_out` are used as individual bit outputs for the Super System 22 games that use them.  It’s not a good idea to reuse them as bit fields like this.

src/osd/modules/file/winrtsocket.cpp
```diff
@@ -95,7 +95,7 @@ class win_osd_socket : public osd_file
 					m_listening = false;
 					actual = 0;
 
-					return error::NONE;
+					return error::FAILURE;
```
cuavas, line 98: I don’t think this will compile at all – note that `osd_file::error` no longer exists at all, and this class no longer implements `osd_file` properly.  It’s probably better not to change it at all unless you’re going to test that it actually still builds for WinRT partition.

src/mame/konami/gticlub.cpp
SailorSat: alright. it shall be a seperate `m_pcb_output` then.

src/mame/namco/namcos22.cpp
SailorSat: okay, it is part of `handle_driving_io` now, and uses a seperate `m_mcu_output`.

src/osd/modules/file/winrtsocket.cpp
SailorSat: Still the behaviour should be identical to the other osd_file socket implementations, right?
It it does not build right now, the change won't hurt anyway.

PR #13853, opened 2025-06-20
k056230: working implementation (ASIO, async, threaded).

SailorSat:
- this makes 'GTI Club', 'Jet Wave', 'Midnight Run', 'Polygonet Commanders', 'Poly-Net Warriors' and 'Winding Heat' playable in LAN.
- "Racin' Force" links up fine, but has other emulation issues.

( this is a spin-off of https://github.com/mamedev/mame/pull/13421 )

### DIFF
--- a/src/devices/machine/k056230.cpp
+++ b/src/devices/machine/k056230.cpp
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:Fabio Priuli
+// copyright-holders:Fabio Priuli, Ariane Fugmann
 /**************************************************************************************************
 
 Konami IC 056230 (LANC)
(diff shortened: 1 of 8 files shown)


angelosa:
Note: LANC implementation for `konami/viper.cpp` is just glue logic to make an handful games to go further, `xtrial` being one of them. Is this tested against it?

SailorSat:
I did not change anything on the k056230_viper-device.
I can't test `xtrial` myself because I don't have that chd file.

angelosa:
`k056230_viper_device` depends on `k056230_device`, so you technically touched it. As long it can generate irqs it should be good to go.

SailorSat:
- changed ram access handlers to 8-bit, as the real hardware (according to racin' force shematics and ks40011 documentation uses 8-bit data bus)
- i also checked `xtrial`, and it starts up fine.

PR #13860, opened 2025-06-21
moog/source.cpp, paia/fatman.cpp: Disable streaming for EGs that don't need it.

m1macrophage:


galibert:
Couldn't you just use get_sound_requested_outputs() to automatically create
the stream when something is connected to it?

  OG.

m1macrophage:
Ah, didn't realize I could do that. I'll look into it.

PR #13862, opened 2025-06-22
m2comm: ASIO rewrite (async polled)

SailorSat:
- this changes m2comm from osd_file sockets (blocking) to ASIO sockets (async).
- for some reason I have yet to determine, comm_framesync causes issues in multithreaded mode, hence this one uses a polled context instead.
- to be multithreaded at a later point.

( this is a spin-off of https://github.com/mamedev/mame/pull/13421 )

### DIFF
--- a/src/mame/sega/m2comm.cpp
+++ b/src/mame/sega/m2comm.cpp
@@ -3,7 +3,7 @@
 
 /*
 Sega MODEL2 COMMUNICATION BOARD 837-10537
-( http://images.arianchen.de/sega-comm/daytona_comm.jpg )
+( http://images.arianchen.de/sega-comm/model2-front.jpg / http://images.arianchen.de/sega-comm/model2-back.jpg )
 |-----------------------------------------------------------------------------|
 | |-------------------|                             |-------------------|     |
 | |-------------------|                             |-------------------|     |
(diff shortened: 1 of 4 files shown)


src/mame/sega/m2comm.cpp
```diff
@@ -165,24 +166,400 @@ Sega PC BD MODEL2 C-CRX COMMUNICATION 837-12839
...
+		if (!sending)
+			start_send_tx();
+
+		m_ioctx.poll();
+
+		return data_size;
+	}
+
+private:
+	class fifo
```
angelosa, line 308: Is there any real reason about doing an own version of a FIFO instead of reusing `util::fifo`?

src/mame/sega/m2comm.h
```diff
@@ -45,43 +43,42 @@ class m2comm_device : public device_t
...
-	uint8_t m_buffer1[0x1000]{};
-	uint8_t m_framesync;
+	uint8_t m_shared[0x4000];  // 16k shared memory
+	uint8_t m_zfg;             // z80 flip gate - bit 0 switches memory banks, bit7 is connected to FG bit 0
+	uint8_t m_cn;              // bit0 is used to enable/disable the comm board
+	uint8_t m_fg;              // i960 flip gate - bit0 is stored, bit7 is connected to ZFG bit 0
+
 	uint16_t m_frameoffset;
 
 #ifdef M2COMM_SIMULATION
```
angelosa, line 59: What's the catch in having this macro define?

src/mame/sega/m2comm.cpp
SailorSat: short answer - because cuavas pointed me to cps2comm as example for a rewrite
there is no particular reason for my own version.

src/mame/sega/m2comm.h
SailorSat: Consider it a leftover from the initial implementation - I was working on that NEC chip and hence needed a switch for running the Z80 vs. running my simulation code.

PR #13874, opened 2025-06-24
current Plug and Play PR - added mouse kids

mamehaze:
new NOT WORKING machines
--------------------
Mouse Kids (Japan) [David Haywood, TeamEurope]


mamehaze:
probably needs a better title.  needs IO studying to emulate the mouse etc. at least.

mamehaze:
did this in the wrong tree, sorry

PR #13878, opened 2025-06-27
arkanoid.cpp: corrected Hexa release year

AugiteSoul:
Added the year for hexa and clone hexaa according to the entry in the GRB archived scans: https://web.archive.org/web/20160508161306/http://www.grb.or.kr/download/GameImage/1999/%EC%A0%9C769%ED%98%B8.jpg


cuavas:
FWIW that isn’t necessarily the release year.  That document is from when the game was submitted for age rating, compliance with media laws, etc.  The date should not be any later than the release date, but given it was in September, the actual release could have been in 1991.

PR #13879, opened 2025-06-27
z80dma.cpp: Added UA858D variant; sinclair/spectrum.cpp: Added configurable DMA slot

holub:


src/devices/machine/z80dma.cpp
```diff
@@ -885,3 +885,28 @@ void z80dma_device::bai_w(int state)
...
+		if (data & 0x08)
+			m_regs_follow[m_num_follow++] = GET_REGNUM(MASK_BYTE);
+		if (data & 0x10)
+			m_regs_follow[m_num_follow++] = GET_REGNUM(MATCH_BYTE);
+	}
+	else
+	{
+		z80dma_device::write(data);
+	}
+}
```
angelosa: Missing EOF

scripts/src/bus.lua
```diff
@@ -4996,6 +4996,20 @@ if (BUSES["AY_SLOT"]~=null) then
...
+
+if (BUSES["DMA"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/bus/spectrum/dma/slot.cpp",
+		MAME_DIR .. "src/devices/bus/spectrum/dma/slot.h",
+		MAME_DIR .. "src/devices/bus/spectrum/dma/cards.cpp",
+		MAME_DIR .. "src/devices/bus/spectrum/dma/cards.h",
+	}
+end
+
```
angelosa: Isn't `DMA` a very common name that may cause clashing in the long run?

holub: For Spectrum unlikely. I can do DMA_SLOT, similar to AY.
I used _SLOT for AY only because it had modem mods which were attached to AY slot. Here it is always DMA, that's why I dropped SLOT part.

Pernod70: Personally I'd suggest SPECTRUM_DMA, SPECTRUM_AY, SPECTRUM_ZXBUS, name them according to your directory structure.

angelosa (approved):

PR #13880, opened 2025-06-29
sega/model2: implement proper lighting; improve "gamma" table

gm-matthew:
With the implementation of lighting, stadium ads in Virtua Striker now have the correct colors.

Model 2 games set a "gamma curve" by using the color translate RAM. It is most likely that each cabinet has its monitor factory-calibrated to produce a desirable picture for the specific game it was designed to play. Virtua Fighter 2, Fighting Vipers and Sonic the Fighters feature adjustable bias and gain settings, presumably to allow software calibration when installing the game board into an Astro City or similar multi-purpose cabinet.

For now I have settled on the translation `raw_value = std::max(((double)i - 64.0) * 255.0 / 191.0, 0.0)` since this produces decent results for most games without requiring any game-specific adjustments. For the above titles which allow software calibration, setting the bias to 64 and gain to 51 gives optimal results.

The 2D tilemap layers are currently not affected by color translate RAM or the "gamma table"; this is on my to-do list.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -10,13 +10,12 @@
     MAME driver by R. Belmont, Olivier Galibert, ElSemi and Angelo Salese.
 
     TODO:
-    - color gamma and Mip Mapping still needs to be properly sorted in the renderer;
+    - Mip Mapping still needs to be properly sorted in the renderer;
     - sound comms still needs some work (sometimes m68k doesn't get some commands or play them with a delay);
     - outputs and artwork (for gearbox indicators);
     - clean-ups;
 
     TODO (per-game issues)
-    - daytona: car glasses doesn't get loaded during gameplay;
     - doa, doaa: corrupted sound, eventually becomes silent;
     - dynamcopc: corrupts palette for 2d;
     - fvipers: enables timers, but then irq register is empty, hence it crashes with an "interrupt halt" at POST (regression);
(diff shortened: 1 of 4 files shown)


angelosa (approved):
Feel like that (in theory) could use a `PORT_ADJUSTER`, however where it reads is an hot path so I'm fine with current handling.

angelosa:
`daytona` now consistently black out textures with `-nothrottle -frameskip 10`, I guess the culprit is:

```c++
	if(m_render_test_mode == true)
		draw_framebuffer( bitmap, cliprect );
	else
	{
		model2_3d_frame_start();

		/* let the geometry engine do it's thing */
		// TODO: move it from here
		geo_parse();

		/* have the rasterizer output the frame */
		model2_3d_frame_end( bitmap, cliprect );
	}
```

TheCoolPup:
daytona93 doesnt seem to suffer from this issue
Daytona USA  daytona93  - MAME 0 277 (LLP64) 6_30_2025 7_04_02 PM

gm-matthew:
> `daytona` now consistently black out textures with `-nothrottle -frameskip 10`, I guess the culprit is:
> 
> ```c++
> 	if(m_render_test_mode == true)
> 		draw_framebuffer( bitmap, cliprect );
> 	else
> 	{
> 		model2_3d_frame_start();
> 
> 		/* let the geometry engine do it's thing */
> 		// TODO: move it from here
> 		geo_parse();
> 
> 		/* have the rasterizer output the frame */
> 		model2_3d_frame_end( bitmap, cliprect );
> 	}
> ```
> 
> 

The issue is due to `geo_parse()` only being run every 10 frames, resulting in certain critical geometrizer update commands being missed; in this particular case I expect it is a `geo_texture_parameters()` call being missed because the luma values for all polygons are 0, hence why they are all blacked out.

It's probably time to move the `geo_parse()` call over to model2.cpp where it belongs so it can be run every frame (or every other frame if running at 30 Hz), regardless of frameskip settings.

Update: moving both `model2_3d_frame_start()` and `geo_parse()` over to `model2_state::model2_interrupt` seems to fix the issue, though further refinement will be needed.

pizzzza19:
My personal dump "daytona" romset also reproduce black textures for the first boot but works fine after several boots. Not sure somewhere cache or memory organized automatically.
My dump was able to boot without "model1io" ROM till ver 0.278. For this ver 0.279 we must need actual dump for my reference.

PR #13883, opened 2025-06-30
skeleton/terminal.cpp: Add Blaupunkt CDS 32_ID terminal

clawgrip:
CDS 32-ID [OldComputers ES]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -43658,6 +43658,7 @@ t4490
 @source:skeleton/terminal.cpp
 7951om
 alcat258
+blaucds32
 loewe715
 loewed
 teleguide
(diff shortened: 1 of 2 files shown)


Pernod70:
Can you split this into it's own blaucds32.cpp file, it really has nothing in common with anything else in terminal.cpp.

clawgrip:
> Can you split this into it's own blaucds32.cpp file, it really has nothing in common with anything else in terminal.cpp.

Yes!!

angelosa:
I believe @Pernod70 meant to split to own driver, rather than having a different base machine_config (and given what we discussed on Discord about it that's for the best).

clawgrip:
Yes, I’ll do it tonight. Thx!

clawgrip:
> Can you split this into it's own blaucds32.cpp file, it really has nothing in common with anything else in terminal.cpp.

Done, thanks a lot!!!!!

PR #13901, opened 2025-07-01
new NOT WORKING machine

mamehaze:
new NOT WORKING machine
---------
Songjiangyanyi Final [Little0]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -40836,6 +40836,7 @@ jingling
 jzth
 puckpkmn
 puckpkmnb
+songjang
 
 @source:sega/saturn.cpp
 hisaturn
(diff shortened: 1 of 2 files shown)


mamehaze:
also maybe the title could be better
jang_1

Hammy1986:
> set as draft because this seems to show that this clone hardware _does_ have the Genesis Z80 even if the other games aren't using it, so it would probably be best to just move all this to one of the Genesis derived drivers rather than copying all the sound Z80 hookup.

TV16b (SOC) http://водолей-нск.рф/price/sega_md2/?parts&d=TV16B
TA06SD (VDP CLONE) http://водолей-нск.рф/price/mikroshemy/?parts&d=TA-06SD

"BOX" Chip is only known on this hardware
Along with the SOC's internal YM2612 there is a UMC 6612 on all of these boards (a YM Clone)
Why it's duplicated externally is what i cannot understand...

mamehaze:
Ok, moved it to the other driver, it has music now.

mamehaze:
2nd game is this

PR #13909, opened 2025-07-04
debugger/qt/debuggerview.cpp: Used DCA_SELECTED color friendly for Light/Dark theme

holub:
Existing color selection doesn't work well when switching from Light to Dark theme

image
image

I'm suggesting to use hardcoded colors friendly for both. see: https://ethanschoonover.com/solarized/

image
image


angelosa (approved):
Ok for now, eventually need exposing these palette setups to a separate entity, need more colors than Qt actually allows us to use ...

PR #13911, opened 2025-07-04
Add Y-Unit Test ROMs *as much of the dump we have)

james-wallace-ghub:


james-wallace-ghub:
This is a strange one, it can be configured for all the video DMA types, or can autodetect based on behaviour. I've given it the config it was found in, with Super High Impact as inputs to trigger as much of the test as is possible.

If any of this is unsuitable, let me know, I'd like to preserve this in a way that works as well as it can.

angelosa:
I assume this will drop unintelligible crap if run due of the lack of GFX data? Is it possible to fake an ASCII ROM taking the other games as reference?

james-wallace-ghub:
> I assume this will drop unintelligible crap if run due of the lack of GFX data? Is it possible to fake an ASCII ROM taking the other games as reference?

The Logo test fails, as well as some of the GFX and DMA specific stuff but there's enough code there to do the menu, the auto detection test and the colour test. It doesn't expose my weaknesses in our setup as far as I can see, but clearly those with more of a 34010 understanding may want to see if they can prove how it's working out which variant it's under at any given time.

Issue #13910, opened 2025-07-04
[GNU/Linux] SDL2 audio (ALSA) needs manual device selection to play audio (was fine in 0.277)


vanfanel:
### MAME version

Latest GIT code, commit hash: dd9d1ae0e78c1fa41504bf8c753979da9d2f6ffb

### System information

GNU/Linux kernel version 6.12.32
MESA stable 25.1.4
CPU/APU AMD Ryzen 7 8700G with integrated audio, using monitor HDMI audio .

### INI configuration details

```INI
#
# CORE CONFIGURATION OPTIONS
#
readconfig                1
writeconfig               0

#
# CORE SEARCH PATH OPTIONS
#
homepath                  .
rompath                   roms
hashpath                  hash
samplepath                samples
artpath                   artwork
ctrlrpath                 ctrlr
inipath                   $HOME/.mame;.;ini
fontpath                  .
cheatpath                 cheat
crosshairpath             crosshair
pluginspath               plugins
languagepath              language
swpath                    software

#
# CORE OUTPUT DIRECTORY OPTIONS
#
cfg_directory             cfg
nvram_directory           nvram
input_directory           inp
state_directory           sta
snapshot_directory        snap
diff_directory            diff
comment_directory         comments
share_directory           share

#
# CORE STATE/PLAYBACK OPTIONS
#
state                     
autosave                  0
rewind                    0
rewind_capacity           100
playback                  
record                    
exit_after_playback       0
mngwrite                  
aviwrite                  
wavwrite                  
snapname                  %g/%i
snapsize                  auto
snapview                  auto
snapbilinear              1
statename                 %g
burnin                    0

#
# CORE PERFORMANCE OPTIONS
#
autoframeskip             0
frameskip                 0
seconds_to_run            0
throttle                  1
sleep                     1
speed                     1.0
refreshspeed              0
lowlatency                0

#
# CORE RENDER OPTIONS
#
keepaspect                1
unevenstretch             1
unevenstretchx            0
unevenstretchy            0
autostretchxy             0
intoverscan               0
intscalex                 0
intscaley                 0

#
# CORE ROTATION OPTIONS
#
rotate                    1
ror                       0
rol                       0
autoror                   0
autorol                   0
flipx                     0
flipy                     0

#
# CORE ARTWORK OPTIONS
#
artwork_crop              0
fallback_artwork          
override_artwork          

#
# CORE SCREEN OPTIONS
#
brightness                1.0
contrast                  1.0
[...]

PR #13914, opened 2025-07-05
sdl_sound.cpp: Correct default device selection logic.

vanfanel:
The logic on the `SDL_OpenAudioDevice()` call was inverted.
This:
`SDL_OpenAudioDevice(dev.m_def ? nullptr : dev.m_name.c_str(), 0, &dspec, &ospec, 0);`
...would mean that if `dev.m_def` has non-zero value, `nullptr` would be used instead of `dev.m_name.c_str()`, which makes no sense: we want to use `dev.m_name.c_str()` when `dev.m_def` is non-zero, and `nullptr` otherwise.

As a side-effect, this works around SDL2's ALSA backend inability to report a default audio device with `SDL_GetDefaultAudioInfo()` and thus is a workaround for issue https://github.com/mamedev/mame/issues/13910
That 's because, when no default device is reported by  `SDL_GetDefaultAudioInfo()`, nullptr is properly passed to `SDL_OpenAudioDevice()` now and the default device is chosen by SDL2 as most SDL2 programs/games do.

### DIFF
--- a/src/osd/modules/sound/sdl_sound.cpp
+++ b/src/osd/modules/sound/sdl_sound.cpp
@@ -208,7 +208,7 @@ uint32_t sound_sdl::stream_sink_open(uint32_t node, std::string name, uint32_t r
 	dspec.callback = sink_callback;
 	dspec.userdata = stream.get();
 
-	stream->m_sdl_id = SDL_OpenAudioDevice(dev.m_def ? nullptr : dev.m_name.c_str(), 0, &dspec, &ospec, 0);
+	stream->m_sdl_id = SDL_OpenAudioDevice(dev.m_def ? dev.m_name.c_str() : nullptr, 0, &dspec, &ospec, 0);
 	if(!stream->m_sdl_id)
 		return 0;
 	SDL_PauseAudioDevice(stream->m_sdl_id, 0);


ajrhacker:
I strongly suspect this fix is technically incorrect and just masks the problem. It forces SDL to select the default device for any device *not* found through SDL_GetDefaultAudioInfo.

vanfanel:
> I strongly suspect this fix is technically incorrect and just masks the problem. It forces SDL to select the default device for any device _not_ found through SDL_GetDefaultAudioInfo.

Can you please look closely at this part?

`dev.m_def ? nullptr : dev.m_name.c_str()`

This is on the original code: it's wrong. If `dev.m_def` is non-zero, you want `dev.m_name.c_str()`, not `nullptr`.
I am not masking anything, there's a logic problem here, unless I am totally wrong on what `dev.m_def ? nullptr : dev.m_name.c_str()` means...

ajrhacker:
Passing a null pointer to `SDL_OpenAudioDevice` as the first parameter requests "the most reasonable default device" (https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice). `dev.m_def` exists because the structure isn't capable of holding a null pointer. Reversing the conditional test will result in all named devices becoming equivalent to the default device.

vanfanel:
> Passing a null pointer to `SDL_OpenAudioDevice` as the first parameter requests "the most reasonable default device" (https://wiki.libsdl.org/SDL2/SDL_OpenAudioDevice). `dev.m_def` exists because the structure isn't capable of holding a null pointer. Reversing the conditional test will result in all named devices becoming equivalent to the default device.

But:

`dev.m_def ? nullptr : dev.m_name.c_str()`

..means:
-If `dev.m_def` is not null, then the parameter passed is `nullpr`
-If `dev.m_def` is null, then the parameter passed is `dev.m_name.c_str()`

That makes no sense: `nullpr` (which means "most reasonable default device" as you said) should be passed ONLY when `dev.m_def` is "empty".
That's how the logic works unless there's something I don't understand.

galibert:
m_def is true when there is a default device, but it was not found among the list of existing devices.  We create an extra device entry with that boolean set. So it does mean "use the default as seen by sdl", e.g. nullptr as the name, when we target it.

vanfanel:
> m_def is true when there is a default device, but it was not found among the list of existing devices. We create an extra device entry with that boolean set. So it does mean "use the default as seen by sdl", e.g. nullptr as the name, when we target it.

Ah! I see now. Sorry for the noise, and thanks for the patience explaining it.
I've fixed the problem from the SDL2 side of things anyway, so I am closing this. Thanks for your valuable time discussing this! :)

PR #13916, opened 2025-07-07
Show Toaplan test switch as input instead of dip switch, remove default keybind

### LINKED ISSUE #13882: Toaplan test switch keybinding cannot be changed
### MAME version

0.277

### System information

Windows 10 x64 (official binary) and Fedora 42 (built from source)

### INI configuration details

```INI

```

### Emulated system/software

Batsugun (batsugun), Dogyuun (dogyuun), etc.

### Incorrect behaviour

In various Toaplan games and games based on Toaplan hardware, there's a "test switch" that gives cheats such as invincibility when activated. In MAME this shows up as a dipswitch that can be toggled on/off, but it's also bound to F1, and as far as I can tell there's no way to change this, nor is there any indication that this binding even exists. This bit me for a long time until I found out this exists, because I had reset bound to F1 and had no idea why the test switch kept turning on.

### Expected behaviour

The F1 binding for the test switch comes from this bit of code at the bottom of toaplipt.h:
https://github.com/mamedev/mame/blob/b6d7c8d6dfb0e12addae9f40d01e078391d40738/src/mame/toaplan/toaplipt.h#L170-L188
Simply changing the SHOW_TEST_AS_DIP #define from 1 to 0 would be an improvement in my book, since that makes the test switch visible and rebindable from the input settings menu. I of course can change (and have changed) it in my own build, but in my opinion this would also be a better default for the official builds. Open to other ideas of what to do with this test switch, though.

### Steps to reproduce

_No response_

### Additional details

_No response_

jackrjli:
Fixes #13882


angelosa (approved):
My only doubt: is `Test Switch` for the label the right expression for this?

jackrjli:
> My only doubt: is `Test Switch` for the label the right expression for this?

It's labeled as "TEST SW" in the [manual for Truxton II](https://www.arcade-museum.com/manuals-videogames/T/Truxton-II-pin--DIP.pdf) (see row 15 in the table on page 2) as well as for [Batsugun](https://www.arcade-museum.com/manuals-videogames/B/Batsugun.man.pdf) and [Dogyuun](https://www.arcade-museum.com/manuals-videogames/D/Dogyuun-Jap_00004.pdf), so keeping that name makes sense to me.

happppp:
I removed the hidden F1 keybinding here: https://github.com/mamedev/mame/commit/d1aeddf321145bb7fcbd183c701604e47530aeb7

It doesn't look like having it as a normal input instead of an on/off switch makes sense.

PR #13929, opened 2025-07-08
nes_vt369_vtunknown: Add GB-50

gatecat:
This is interesting as it's a VT369 (or possibly VT389) system that supports loading games from an SD card. Marked as NOT WORKING because the SD card support isn't emulated - the regular 150 built in games seem okay, although they're nothing special.


ajrhacker:
What are the credits for this set?

gatecat:
It was dumped by me (Myrtle Shah)

PR #13930, opened 2025-07-09
machine/i2cds1307.cpp: Added I2C DS1307; Used the device in Spectrum Next

holub:


src/devices/machine/i2cds1307.cpp
```diff
@@ -0,0 +1,66 @@
...
+/**********************************************************************
+    DS1307 64x8, Serial, I2C Real-Time Clock (Maxim Integrated Products)
+**********************************************************************/
+
+#include "emu.h"
+
+#include "i2cds1307.h"
+
+
+DEFINE_DEVICE_TYPE(I2C_DS1307, i2c_ds1307_device, "i2c_ds1307", "I2C DS1307 RTC/SRAM")
```
angelosa, line 12: I find the i2c_ prefix for filename/device class to be quite dull tbh.

rb6502: It's not what I would've done because the DS1307 is I2C by definition, but it does fit in with existing usage like i2cmem so I can't really object (and I can't say that's setting a bad example because nobody would know what machine/mem.cpp is).  In the not-too-far future I plan to create machine/i2c (or devices/i2c on the same level as machine) and revisit the naming of I2C components.  But now is not that time.

holub: I'm dropping the prefix as part of #13932. Let me know if you want it back.

PR #13935, opened 2025-07-10
formats/all.cpp: Include IPF, 86F disk formats in floptool

invertego:
IPF is included unconditionally by scripts/formats.lua, so the HAS_FORMATS condition was always false.

86F is also included unconditionally but simply wasn't added to formats/all.cpp when it was introduced.

I added 86F to the "Container FM/MFM" category for consistency with the grouping in devices/imagedev/floppy.cpp and moved IPF to the "PC" category for the same reason.

cf. 3b13ae6d7121b3e1de8e5aa48918f9f9b6567f7a, add6b4515206d5973ec7e3e30012a469fecc49bb

### DIFF
--- a/src/lib/formats/all.cpp
+++ b/src/lib/formats/all.cpp
@@ -9,6 +9,7 @@
 
 #include "cqm_dsk.h"
 #include "dsk_dsk.h"
+#include "ipf_dsk.h"
 #include "td0_dsk.h"
 #include "hxchfe_dsk.h"
 #include "hxcmfm_dsk.h"
@@ -18,6 +19,7 @@
 #include "pc_dsk.h"
 #include "d88_dsk.h"
 #include "dfi_dsk.h"
+#include "86f_dsk.h"
 #include "fs_unformatted.h"
 
 #ifdef HAS_FORMATS_2D_DSK
@@ -532,10 +534,6 @@
 #include "pc98_dsk.h"
 #endif
 
-#ifdef HAS_FORMATS_IPF_DSK
-#include "ipf_dsk.h"
-#endif
-
 #ifdef HAS_FORMATS_PHC25_CAS
 #include "phc25_cas.h"
 #endif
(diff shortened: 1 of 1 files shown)


invertego:
While I was doing this, I ran some regexes and found other formats that are missing from all.cpp. However, I don't know much about these additional formats or have samples to test, so I left them out of this change. If anyone want to take that up, here is the missing code.

```cpp
#ifdef HAS_FORMATS_ABC1600_DSK
#include "abc1600_dsk.h"
#endif
#ifdef HAS_FORMATS_BK0010_DSK
#include "bk0010_dsk.h"
#endif
#ifdef HAS_FORMATS_IDPART_DSK
#include "idpart_dsk.h"
#endif
#ifdef HAS_FORMATS_LW30_DSK
#include "lw30_dsk.h"
#endif
#ifdef HAS_FORMATS_TIBDD001_DSK
#include "tibdd001_dsk.h"
#endif

#ifdef HAS_FORMATS_ABC1600_DSK
	en.add(FLOPPY_ABC1600_FORMAT); // abc1600_dsk.h
#endif
#ifdef HAS_FORMATS_BK0010_DSK
	en.add(FLOPPY_BK0010_FORMAT); // bk0010_dsk.h
#endif
#ifdef HAS_FORMATS_IDPART_DSK
	en.add(FLOPPY_IDPART_FORMAT); // idpart_dsk.h
#endif
#ifdef HAS_FORMATS_LW30_DSK
	en.add(FLOPPY_LW30_FORMAT); // lw30_dsk.h
#endif
#ifdef HAS_FORMATS_TIBDD001_DSK
	en.add(FLOPPY_TIB_DD_001_FORMAT); // tibdd001_dsk.h
#endif
#ifdef HAS_FORMATS_CP68_DSK
	en.add(FLOPPY_CP68_FORMAT); // cp68_dsk.h
#endif
#ifdef HAS_FORMATS_FDOS_DSK
	en.add(FLOPPY_FDOS_FORMAT); // fdos_dsk.h
#endif
#ifdef HAS_FORMATS_MFM_HD
	en.add(MFMHD_GEN_FORMAT); // mfm_hd.h
#endif
#ifdef HAS_FORMATS_ACORN_DSK
	en.add(FLOPPY_CUMANA_DFS_FORMAT); // acorn_dsk.h
#endif
```

angelosa:
Is it okay to have IPF and 86F enabled everywhere? @galibert

galibert:
IPF has a decent header, shouldn't be a problem.  86F I don't know, never
saw that format.

invertego:
To be clear, this change doesn't enable these formats for any new devices. It adds two formats to mame_formats_full_list(), which is only utilized by floptool. It will affect the auto-detect feature of that tool, but both formats have unique header identifiers.

PR #13936, opened 2025-07-10
nes_vt32_soc.cpp - added typo240

mamehaze:
new NOT WORKING machines
------------
Vintage Gamer 240-in-1 [David Haywood, TeamEurope]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -35470,6 +35470,7 @@ myaass
 retro400
 rfcp168
 rminitv
+typo240
 
 @source:nintendo/nes_vt369_vtunknown.cpp
 36pcase
--- a/src/mame/nintendo/nes_vt32.cpp
+++ b/src/mame/nintendo/nes_vt32.cpp
@@ -663,6 +663,11 @@ ROM_START( mc_9x6sa )
 	ROM_LOAD( "999999-in-1, 8 bit slim station, newpxp-dvt22-a pcb.bin", 0x00000, 0x200000, CRC(6a47c6a0) SHA1(b4dd376167a57dbee3dea70eb16f1a38e16bcdaa) )
 ROM_END
 
+ROM_START( typo240 )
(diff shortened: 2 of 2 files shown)


mamehaze:
fairly sure this is a VT32 as it boots in the driver, has the VT32 encryption and has Curly Monkey 2 on it
 

however, all sub-menus show the same menu instead of the one they're meant to
@ajrhacker if you want to take a quick look at this one I won't add anything else to this PR for now

All the game names are in the start of the ROM, in the first bank however, so maybe it's some kind of register readback or protection (or math chip?)

it does seem to access what would be the ALU on VT369
	map(0x4130, 0x4137).rw(FUNC(vt3xx_soc_base_device::alu_r), FUNC(vt3xx_soc_base_device::alu_w));
	map(0x4138, 0x413d).r(FUNC(vt3xx_soc_base_device::alu_r)); // mirror?
addresses, so could be that.  some VT32 models apparently do have it?

(unless of course they've been misidentified, and some VT36x models have the VT32 encryption etc. instead?)

if you don't want to look at it, I'll try moving the ALU handler to a device, and adding it to both VT32 and VT369

ajrhacker:
I don't think the accesses are ALU-like. That said, the code clearly expects that bytes stored in $4132 and $4134 can be read back. I've just fixed that.

mamehaze:
Could be they're just using the ALU registers as extra storage without triggering anything.  Sometimes to watch out for in the future I guess!  Thanks for your opinion on it anyway.

I've sent a bunch more VT based things to TeamEurope, although most are just crappy multigame bootlegs, and I suspect many will be identical to each other just in different cases with different model numbers.  Some seem to be censored versions with specific games (especially SMB) removed without changing the menu entry.

mamehaze:
with the changes in master this seems to work as well as the other VT32 sets, all games that I can see are playable, just a few have slightly broken title screens and such (which are the same in other units with those games, so probably a rendering issue with the extended modes, or timing issue)

PR #13942, opened 2025-07-11
sinclair/glukrs.cpp: Updated clock to schematics defined; sinclair/pentevo.cpp: Removed hacks around RTC registers

holub:


src/mame/sinclair/glukrs.cpp
```diff
@@ -14,18 +14,9 @@
 #include "glukrs.h"
 
 glukrs_device::glukrs_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
-		: mc146818_device(mconfig, GLUKRS, tag, owner, clock)
+		: mc146818_device(mconfig, GLUKRS, tag, owner, 32'768)
```
cuavas, line 17: Ignoring the clock frequency is bad form.  Wouldn’t this use a 32768 Hz watch crystal for timing in practice?  It should still be passed in from machine configuration.

happppp: I presume the OSC is internal here? Then the driver mcfg should not 'know' the clock and should not pass any value to it. (in other words, what holub did is correct in that case)

rb6502: I doubt it's internal, but every clock chip I've read the datasheet for wants a 32768 Hz crystal and several of them say any other value is not guaranteed to work.

holub: This device is specific to ZX Evolution usage and have clock "defined". I would like to avoid ability to modify this clock value as it implies additional configuration in RTC registers to work correctly.
As a option I can add validation to device which will check if supplied clock is desired.

cuavas: We allow changing the “standard” clock on other real-time clocks and chips that have standard clocks defined.  There are systems that abuse chips by using non-standard clock frequencies.  An example is the Zaccaria arcade games that overclock Signetics video/sound chips to get higher horizontal resolutions than they were designed for.  If someone runs an RTC with a non-standard clock input, they should get non-standard seconds.

PR #13944, opened 2025-07-11
sinclair/glukrs.cpp: Added expected clock value validation

holub:


holub:
as per #13942 comments

holub:
> I still think people should be able to change the clock frequency and get non-standard time if they want to. Being able to emulate “bad behaviour” is part of what MAME does. Can it be demoted to a verbose message rather than a hard error?

They can change if they want to. Isn't it just validation message they can ignore? I'm confused how can I achieve this anyway...

PR #13955, opened 2025-07-14
Juicebox - allow loading of raw ROM dumps from Software List, add 6 new dumps to the SL

mamehaze:
new software list entries
------------
juicebox.xml
Ed, Edd n Eddy - Episode 39 (H2679) [Sean Riddle]
Dexter's Laboratory - Episode 51 (H6247) [Sean Riddle]
Vans Triple Crown Volume 1 (BMX) (H2688) [Sean Riddle]
Vans Triple Crown Volume 2 (Skateboarding) (H2689) [Sean Riddle]
Juicebox Music Video - Avril Lavigne, The Calling, Clay Aiken (H2681) [Sean Riddle]
Juicebox Music Video - Liz Flair, Stacie Orrico, Skye Sweetnam, Zoegirl (H6254) [Sean Riddle]

- change softlist entries that are functional from partial to simply supported, as the ones that boot don't really have any major issues.

### DIFF
--- a/hash/juicebox.xml
+++ b/hash/juicebox.xml
@@ -22,7 +22,7 @@ license:CC0-1.0
 
 *  H2679 - Ed, Edd n Eddy Episode #39
    H2680
-*  H2681 - Juice Box Music Video (Avril Lavine, The Calling, Clay Miken)
+*  H2681 - Juice Box Music Video (Avril Lavigne, The Calling, Clay Aiken)
    H2682 - Juiceware Animated Assortment
 *V H2683 - Codename: Kids Next Door Volume 1
    H2684 - Ed, Edd n Eddy Volume 1
@@ -55,7 +55,7 @@ license:CC0-1.0
    H62xx carts
 
 *  H6226 - Juice Box Music Video (Ashlee Simpson, The Cure, Vanessa Carlton, fan_3, Ashlee Simpson)
-*  H6228 - Juice Box Music Video (Avril Lavine, The Calling, Clay Miken, Ruben Studdard)
+*  H6228 - Juice Box Music Video (Avril Lavigne, The Calling, Clay Aiken, Ruben Studdard)
 *  H6230A- Juice Box Music Video (Baha Men, Jump5, Jump5, Beu Sisters, ZOEgirl)
 *  H6232 - Juice Box Music Video (Zoegirl, Jump5, Jump5, Baha Men, Atomic Kitten)
    H6233 - Kenny Chesney - 'Live Those Songs', 'No Shoes, No Shirt, No Problems', 'I Go Back'  Sara Evans - 'Suds in the Bucket, Martina McBride - 'This One's For The Girls'


mamehaze:
I've been sitting on these dumps for over a year now because the awkwardness of having to tack headers on just for testing was a little annoying.

Sadly the carts for this system seem to be rapidly failing, probably due to low grade flash.  In a batch of over 20 carts only about half of them gave good dumps, with 3 reading as completely blank and the others clearly corrupt.  The same was found for the previous batch too.

I've opted not to add the sets that don't boot due to the dumps being obviously bad for now.

hash/juicebox.xml
```diff
@@ -19,51 +18,72 @@ license:CC0-1.0
    The flash ROMs on these cartridges look to be deteriorating, each will likely need verifying against multiple cartridges to verify the integrity of
    the data.
 
-   H2679 - Ed, Edd n Eddy Episode #39
+   H26xx/H27xx coded carts
+
+*  H2679 - Ed, Edd n Eddy Episode #39
+   H2680
+*  H2681 - Juice Box Music Video (Avril Lavine, The Calling, Clay Miken)
```
FoxhackDN: Sorry to be a bother, but the correct spelling for these two is Avril Lavigne and Clay Eiken. The names are also misspelled further ahead in the other cart that has their music.

mamehaze: seems to be Clay Aiken, not Clay Eiken, but I've fixed the comments

FoxhackDN: Thanks and sorry, I kant spel gud

cuavas:
This breaks loading all the GamePark GP32 (gp32.xml) software list items, and possibly other things as well.

mamehaze:
ah, looks like I pushed the wrong version of the code

mamehaze:
actually, no I didn't, GP32 still works fine, the intended code is in there

mame gp32 -bios 157e dunguarde
mame gp32 -bios 156k dunguard
mame gp32 -bios 156k astonish

etc. all work

the code falls through to the old identification code using the fake headers unless it finds the extra info in the softlist, nothing is broken.

this is something I tested when adding these, as initially it *did* break.

mamehaze:
some games are locked to specific BIOS versions (or malfunction if used with the wrong region machines) the software list has no way of knowing this

mame gp32 -bios 166m tearsast

works for example, but that doesn't get recognised with other BIOS options

mamehaze:
anyway it all works as it did before, nothing is regressed with this, just GP32 driver is awkward to use because of some kind of region locking or firmware revision locking.

might be better to treat the older GP32 units as different machines (or simply use the newest firmware as the primary one) rather than handling things with BIOS options, then the softlist can be filtered.

mamehaze:
the BIOS requirements are noted in 'usage' in the GP32 software list
info name="usage" value="Requires firmware version 1.6.6"

PR #13957, opened 2025-07-15
Add point-ROM auto-increment to Namco System 23

MooglyGuy:
This fixes point-ROM verification in Final Furlong's service menu.

In addition, it significantly improves graphics in Downhill Bikers and Race On!, as both games look for the expected walking-bit patterns from offset 0x100 to 0x1FF in the point-ROM data, with the expectation of the current point-ROM peephole address to auto-increment after each LSW read. If the walking-bit patterns aren't found, it zeroes out a global model-index offset, otherwise it sets that offset to 0x80.

Downhill Bikers won't fail to POST if these bit patterns aren't present, so this might be some sort of development leftover where the first 512 bytes of data were only added to the point-ROM data as part of the retail build process.

### DIFF
--- a/src/mame/namco/namcos23.cpp
+++ b/src/mame/namco/namcos23.cpp
@@ -4810,9 +4810,8 @@ u16 namcos23_state::c417_ptrom_lsw_r()
 		LOGMASKED(LOG_C417_REG, "%s: c417 point rom (over-limit) lsw read: %04x\n", machine().describe_context(), 0xffff);
 		return 0xffff;
 	}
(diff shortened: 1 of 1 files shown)


happppp:
Don't forget the side effects check, so something like:
```
const u16 data = rom[x];
if (!side_effects...)
	x++; // does it need & or % to make sure it won't go OOB? - edit: i see LOGMASK about a limit check, so I guess that's covered
return data;
```

MooglyGuy:
There are other handlers that need a pass for that, too, the check for that specific function in particular has now been added.

PR #13958, opened 2025-07-16
sinclair/specnext_dma.cpp: Added prescaler register

holub:
Now we comply with the factory test
image


src/mame/sinclair/specnext_dma.cpp
```diff
@@ -39,10 +42,9 @@ specnext_dma_device::specnext_dma_device(const machine_config &mconfig, const ch
 
 void specnext_dma_device::write(u8 data)
 {
-	z80dma_device::write(data);
-
 	if (num_follow() == 0)
 	{
+		z80dma_device::write(data);
```
angelosa, line 47: Shouldn't need this move of `z80dma_device::write`, it's repeated in both branches anyway?

holub: There is a "if" inside which changes internal state. I'll not be able to track nreg in "else" overwise

PR #13959, opened 2025-07-16
m62_bkungfu.cpp - protection notes

mamehaze:
m62_bkungfu.cpp - added notes about other features of the protection

### DIFF
--- a/src/mame/irem/m62.h
+++ b/src/mame/irem/m62.h
@@ -54,6 +54,9 @@ public:
 	void init_battroad();
 
 protected:
+	virtual void machine_start() override ATTR_COLD;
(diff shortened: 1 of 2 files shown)


mamehaze:
this transfers some of my notes on the protection into the driver (as it handles everything to do with the tilemap, not just the encrypted backgrounds)

I still think we'll need an MCU dump from somewhere before this ever works properly, as it seems likely some elements are coming from internal ROM - some commands draw HUD elements in given states with nothing more than a simple parameter (eg. energy) where the actual tiles used are a bar, not even simple numbers.

It's also possible some of the animated elements have layout definitions coming from internal ROM, as again they're not requesting pointers into the encrypted data, but instead simple 'element number' commands.

PR #13961, opened 2025-07-17
sbasketb: Fixed default score dip based on manual

FoxhackDN:


FoxhackDN:
The service manual - https://www.arcade-museum.com/manuals-videogames/S/Super-Basketball.pdf - shows that the default setting is 100-115, not 70-78.

PR #13963, opened 2025-07-17
sinclair/spectrum_v.cpp: Added ULA snow effect support [holub, spectramine]

holub:
~~There is an issue with precise timings. Will try to sort that out next time.~~

### DIFF
--- a/src/devices/cpu/z80/z80.lst
+++ b/src/devices/cpu/z80/z80.lst
@@ -46,6 +46,7 @@ macro   wm16_sp %dd
 
 macro   rop
 	m_m1_cycles-2 !! TDAT8 = opcode_read();
+	// TODO based on ULA contention analyzes refresh must send already incremented R.
 	if (refresh_en)
 		m_refresh_cb((I << 8) | (R2 & 0x80) | (R & 0x7f), 0x00, 0xff);
 	+ 2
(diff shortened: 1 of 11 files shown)


holub:
I'm done with 48 machine. ~~128/+2 works somehow but not precise...~~

holub:
Let it snow!

ICEknigh7:
Can confirm that the snow effect seen after Strider crashes looks identical as in real hardware:
image 
image

The screen glitches in The Ninja Warriors look a bit different but those may depend on something else than the known snow effect:
image 
image
image 
image
image
image 
image

Allofich:
> Can confirm that the snow effect seen after Strider crashes looks identical as in real hardware:

Judging from the screenshots, there is still a small difference. There are a few dots of the effect on the left or right of the columns in the top image not present in the bottom one.
image
image

holub:
> > Can confirm that the snow effect seen after Strider crashes looks identical as in real hardware:
> 
> Judging from the screenshots, there is still a small difference. There are a few dots of the effect on the left or right of the columns in the top image not present in the bottom one. image image

That are live data from the memory. It indeed can be different/dynamic on the both machines.
What are more confusing - screenshots form NW. But atm I have no inspiration to digg into this problem deeper

ICEknigh7:
> > Can confirm that the snow effect seen after Strider crashes looks identical as in real hardware:
> 
> Judging from the screenshots, there is still a small difference. There are a few dots of the effect on the left or right of the columns in the top image not present in the bottom one. image image

I've re-checked the real hardware footage and those dots are not present in other cycles, which does indeed make current emulation look identical:
image

So not sure if that was a problem with the video encoding or if other few random dots just also happen to appear sometimes, here's how that specific frame looks in other points of the footage:
image
image
image

PR #13966, opened 2025-07-18
add kim-1 cassette tapes

Dagarman:
add kim-1 cassette tapes from http://retro.hansotten.nl/

https://www.dropbox.com/scl/fo/l0l8itidgzihtyw14jltq/ABiVkk3_tnyBFgVl9MD_PFs?rlkey=1h87rbfjkjk24qntnw79f42c4&st=p90y9pkz&dl=0

### DIFF
--- a/hash/kim1_cass.xml
+++ b/hash/kim1_cass.xml
@@ -18,6 +18,78 @@ kimtape -m -a0 -da1 a1.bin a1.wav
 
 <softwarelist name="kim1_cass" description="KIM-1 cassettes">
 
+	<software name="baseball">
+		<description>Baseball</description>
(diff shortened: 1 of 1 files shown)


Dagarman:
Thanks. I'll try to figure out how to do the diffs again for next time.

cuavas:
This adds checked-in DOS line endings.  If the whole file shows as a difference, it’s usually a warning sign that something is very wrong.  Can people please check when red flags like this?

Unfortunately this won’t be fixed for the 0.279 release.

Dagarman:
OK. I forgot how to compile so I just used put it in the github folder.
I’ll work on fixing it

PR #13973, opened 2025-07-20
m68010 ctor accepts type param 

Elektraglide:
added a couple of additional ctors that allow passing in derived class type


galibert:
The constructors supposed to be called from a derivative should be protected, not public.

Elektraglide:
fixed

PR #13978, opened 2025-07-20
CD-i: Add Undefined Matte Behavior

### LINKED ISSUE #13976: CD-i: Matte Behavior slightly different than console
### MAME version

0.278 (mame0278-259-gce9d7e7c0af-dirty)

### System information

N/A

### INI configuration details

```INI

```

### Emulated system/software

_No response_

### Incorrect behaviour

Image

### Expected behaviour

Image

### Steps to reproduce

1. Validation Disc (EU)
2. Video Representation > More > Mattes
3 Watch the red/blue test until it switches to the green/blue test.

Notice that in the final frame while it's switching tests, the output doesn't match exactly.

### Additional details

When the "end" coordinate of the first matte is beyond the start coordinate of the second matte, the manual is unclear about the expected behavior. The console verification shows the desired behavior.

Vincent-Halver:
Fixes #13976

Greenbook says not to change the MF flag within a line. However, if you do, the behavior is undefined.

Console verification shows that the real console will use these values. The non-spec behavior is explained in a comment.

Additionally, this tidies up the Matte function.

Previous Behavior
{0C366E68-17F7-4B26-AC48-FB056B3B1381}

Console Verified Behavior
{73492324-2344-4C51-A7E3-2FD111502E4B}

New Behavior
{D2F117E2-E375-4B1B-BC33-751EE76493CE}

### DIFF
--- a/src/mame/philips/mcd212.cpp
+++ b/src/mame/philips/mcd212.cpp
@@ -54,17 +54,15 @@ inline ATTR_FORCE_INLINE uint8_t mcd212_device::get_matte_op(const uint32_t matt
 
 void mcd212_device::update_matte_arrays()
 {
-	bool latched_mf[2]{ false, false };
-	uint8_t latched_wfa = m_weight_factor[0][0];
-	uint8_t latched_wfb = m_weight_factor[1][0];
 	const int width = get_screen_width();
-
 	const int num_mattes = BIT(m_image_coding_method, ICM_NM_BIT) ? 2 : 1;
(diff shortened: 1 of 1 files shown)


Vincent-Halver:
Additionally, this fixes the incorrect RGB transparency mode 2. #13979
image

Previously this would check the high bit instead of the low bit. When checking the "false" condition, a high bit of 0 would match 0. But a highbit of 1 (0x80) would not match a 1. This is now fixed by using the BIT command correctly.

PR #13992, opened 2025-07-26
sinclair/specnext.cpp: Added video output configuration (VGA/HDMI); refresh rate 50/60Hz; timings for all reference machines

holub:
see:
https://wiki.specnext.dev/Reference_machines
https://wiki.specnext.dev/Video_Timing_Register


src/mame/sinclair/specnext.cpp
```diff
@@ -807,11 +835,87 @@ void specnext_state::memory_change(u16 port, u8 data) // port_memory_change_dly
...
+		, HZ_TO_ATTOSECONDS(28_MHz_XTAL / 2) * width * height);
+	m_ula->set_raster_offset(left, top);
+	m_lores->set_raster_offset(left, top);
+	m_tiles->set_raster_offset(left, top);
+	m_layer2->set_raster_offset(left, top);
+	m_sprites->set_raster_offset(left, top);
+
+	m_eff_nr_03_machine_timing = m_nr_03_machine_timing;
+	m_eff_nr_05_5060 = m_nr_05_5060;
+	popmessage("%s: %s %dHz", machine_name, is_hdmi ? "HDMI" : "VGA", m_nr_05_5060 ? 60 : 50);
```
rb6502: This should be LOGMASKED() or similar rather than just shown to the user.

holub: IMHO this way it gives good user experience. Similar how your TV set shows such message.

holub:
I'll postpone till resolve bgfx issue. It crashes unconditionally after configure() for other values too.

holub:
The issue is related to bgfx. I've tried other -video and there it works like a charm

rb6502:
There is a lingering issue in BGFX where it's  mixing up the stride with the width or something along those lines.  macOS prints diagnostics to the terminal about it in some cases.

src/mame/sinclair/specnext.cpp
```diff
@@ -57,10 +57,31 @@ namespace {
...
+	u16 max_hsync;
+	u16 max_hblank;
+	u16 min_hactive; // 256x192 area
+	u16 max_hc;
+
+	u16 min_vblank;
+	u16 int_v;
+	u16 min_vsync; // displays don't like vsync = vblank
+	u16 max_vsync;
+	u16 max_vblank;
```
angelosa, line 74: They are two separate things indeed, blank = back porch + sync + front porch 

upd7220_blank_ref

src/mame/sinclair/specnext.cpp
angelosa: If we want to pursue this road then `popmessage` isn't the right tool positioning-wise. 
Should be a specific user opt-in function, background color should be more transparent, and be between the [action-safe and title-safe area](https://en.wikipedia.org/wiki/Safe_area_(television)). 
Consider also that `popmessage` has this tendency of sticking right inside in the title-safe area for too much (specifically for `temp.length() / 40 + 2`, in seconds), and this isn't a message that should stick for this long ...

holub: I've switched to general LOG

galibert:
There are some conflicts, can you fix them to your taste?

holub:
must be it

> There are some conflicts, can you fix them to your taste?

holub:
> The issue is related to bgfx. I've tried other -video and there it works like a charm

After some experiments I realized the offender was vulkan driver but not bgfx. bgfx been blamed for no reason.

Issue #2561, opened 2017-08-13
Philips CD-i [cdimono1] - Zelda CD-i can't play in-game cutscenes properly


LuigiBlood:
Basically, to play cutscenes in-game, you have to leave the button pressed until the sprites disappear.
If you just tap the button, the cutscene will not play once the sprites disappear (it does not softlock the game, you can tap a button again to skip the cutscene... except it's not played at all).

I've tried to look into it in the source and it sounds like an interrupt (timing?) problem. But I cannot seem to pinpoint the cause of it.

MooglyGuy:
There are a bunch of these sorts of issues. A long-term correct fix will be to emulate the now-dumped SERVO and SLAVE microcontrollers in order to eliminate problems with the HLE-based simulation of their behaviors, but that's going to be quite a lot of work in and of itself.

angelosa:
Last time I checked there are 3 different Zelda on CD-i.
Which one?

LuigiBlood:
> Last time I checked there were 3 different Zelda on CD-i.
> Which one is the winner here?

Wand of Gamelon & Faces of Evil

Vincent-Halver:
I can't reproduce. When I go to the shop, and stab the person, the animation plays even though I'm not holding anything down.

1. Wand of Gamelon
2. Go to the town Samarkand
3. Go all the way right
4. Hold down and press button 2 in front of the door to enter
5. Stab the woman in the green dress.
6. Don't hold down any buttons

The cutscene plays. Is it hard to reproduce, or maybe it was fixed in the last few years?

<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/b53857b7-40d3-4bf8-80f5-2314437a608e" />

Vincent-Halver:
0. Wand of Gamelon
1. Go to the town that starts with S.
2. Go into the first building.
3. Stab the guy 10 times.
4. Each time he says "Of course I'm on your side", skip the cutscene.
5. On the 10th time, the game will freeze.

Next, repeat the same.
4. Instead of skipping, don't skip. Watch it 10 times all the way through.
5. On the 10th time, the game won't freeze.
6. Now stab him 5 more times, this time skip as before.
7. On the 5th (15th total) time, the game will freeze.

I still can't reproduce the "non-soft lock" condition discussed here. The game-freezing glitch is definitely a well known problem with the CD-i emulator, but that does not seem to be what this bug is about. 

[...]

Issue #13231, opened 2025-01-14
CD-i: Unclear correct configuration for 625 scanlines


Vincent-Halver:
### MAME version

0.272 (mame0272-397-ge2e6629e89b)

### System information

N/A

### Emulated system/software

CD-i

### Incorrect behaviour

High Res patterns don't work.

1. Use the Validation Disc (Europe)
2. Phillips Tests > 1st option > Drawing > Clut 4 High Res
3. Click through the examples. Take special notice of the "Outline" (Example 3)
![image](https://github.com/user-attachments/assets/3a6953fc-d86f-49db-bffc-06a8eeaa6867)
4. Compare to how the outline is supposed to look (e.g. one of the other draw modes). Notice how many horizontal 1 pixel tall features are missing.

If you look at cdi.cpp
```c++
	screen.set_raw(14976000, 960, 0, 768, 312, 32, 312);
```

Then you will see the vertical height is only 312, with a 32 line vblank. Though this is to spec, the screen needs 625 lines to support High Res.

### Expected behaviour

All CD-i players are expected to support the "Base Case". Clut4 High Res mode is in the base case.

High Res mode renders interlaced. According to the Green Book Spec, page V-38:

> The video decoder scan timing is such as to drive a standard 525 line (typically 60Hz)
or 625 line (typically 50Hz) television display, with two fields per frame.

> In the extended case, High Resolution images may be displayed with two interlaced
fields per frame, or line sequentially. 

Given there is only 312 lines of output, it's not obvious where to draw the additional lines. So the screen must be adjusted.

```c++
screen.set_raw(14976000, 960, 0, 768, 625, 64, 625);
```
This was my attempt, however I'm looking for guidance on whether this is a correct setup of the Screen object. 

The minimal patch to fix this issue will be to increase the screen resolution, and then update the `screen_update` to
1. Skip all processing after line 312
2. Preferably have each line draw two output lines (N and N+1) to avoid regression of the visuals.

Only after the resolution is corrected can support for High Res image decoding be added.

Vincent-Halver:
Example diagram I found. Yes, each "Field" runs for 312 lines. It seems likely to me that the screen must be 625 lines tall. But this also then leads to the image appearing squished. To properly interlace, some tricks can be done to correct the image, but before I get too deep doing that, I wanted some feedback on the expected idiomatic way that it should be done.
[...]

PR #13997, opened 2025-07-27
CD-i: Add Interlace Graphics

Vincent-Halver:
Minimal code to double the vertical resolution for #13231 . 

I would appreciate feedback, especially as I don't feel confident of the proper expected way to do this within MAME's screen object.

Old version
image
New Version (crisper, due to pixels being less stretched out).
image

Difference map showing the only difference is the "blurriness" due to resolution upscaling.

{BA07974D-8D1E-4411-AF66-1E86B39B581A}

This was tested minimally and seems stable. 

Something closer to a proper interlaced image can be accomplished using the _PA bit, however current screen blanks between frames preventing the parity bit from being able to produce a correct interlace image yet.

### DIFF
--- a/src/mame/philips/cdi.cpp
+++ b/src/mame/philips/cdi.cpp
@@ -434,7 +434,7 @@ void cdi_state::cdimono1_base(machine_config &config)
 	m_mcd212->int_callback().set(m_maincpu, FUNC(scc68070_device::int1_w));
 
 	screen_device &screen(SCREEN(config, "screen", SCREEN_TYPE_RASTER));
-	screen.set_raw(14976000*2, 960, 0, 768, 312*2-32, 32, 312*2-32);
+	screen.set_raw(960*(312*2-32)*50, 960, 0, 768, 312*2-32, 32, 312*2-32);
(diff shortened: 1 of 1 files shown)


ICEknigh7:
If you're just looking for sharpness and the actual system's resolution is not really higher, that's what the `prescale` option in mame.ini does.

Vincent-Halver:
You can read more information in the linked bug. This does not yet fix the bug, but is a necessary precondition.

Vincent-Halver:
image
Interlaced graphics are added.

Photo CDs now show double resolution as before.

Seems like it *helps reduce the risk of* but does not fully fix #2561 .

Vincent-Halver:
This is now ready for review. I fixed all the known gaps in the original change.

**Notice:** This breaks save states.
This is because the interlaced pixels from the previous frame are now part of the save state.

**Games this reportedly helps:**
The Flowers of Robert Mapplethorpe - Makes use of interlace graphics for higher resolution photos.
Photo CD Sampler - Same, as with many other art or photo discs.
Zelda Wand of Gamelon - Appears to partially fix one of the known freeze glitches caused by skipping dialog.

PR #14003, opened 2025-07-29
added first pass of a new internal ROM to some st2302u_bbl_spi.cpp sets, move to handheld folder + new working system

mamehaze:
new WORKING machine
---
Throwback Pocket Video Game Console 150+ 8-Bit Games [Sterophonick]

- added different internal ROM to some st2302u_bbl_spi.cpp sets [David Haywood]


mamehaze:
- this was hand typed, needs double checking, games boot but with significant graphical problems
- none of these have any kind of TV out, so I've moved them to the handheld folder instead of tvgames

mamehaze:
the ROM should now be verified, but there are still issues.

also the music seems correct on the menu at 24mhz here, rather than 32mhz, so maybe retro150/retro150a are an older SoC, or at least slower clocked.  The trojan ran slower on those than dphh8633 too.

mamehaze:
@ajrhacker could you have a look over this and merge if possible (I'm a bit stumped on the gfx issues now though, the trojan data was double checked, the internal ROM should be good, it still loads code etc. properly via the SPI but whatever's going on with the graphics on the sets here is strange)

mamehaze:
I've also checked enough of dphh8633 to be 99% sure it's the same internal ROM at this point.

src/mame/handheld/st2302u_bbl_spi.cpp
```diff
@@ -260,32 +264,40 @@ static INPUT_PORTS_START(bbl380)
...
 	m_maincpu->add_route(0, "mono", 1.00);
 	m_maincpu->add_route(1, "mono", 1.00);
 	m_maincpu->add_route(2, "mono", 1.00);
 	m_maincpu->add_route(3, "mono", 1.00);
+}
 
+void bbl380_state::bbl380(machine_config &config)
+{
+	ST2302U(config, m_maincpu, 32000000);  // 32Mhz clock correct for music tempo. SoC type not confirmed
+	SPEAKER(config, "mono").front_center();
```
cuavas: MHz, not Mhz.  Also please use digit grouping like `32'000'000` so order of magnitude is easy to see at a glance.

src/mame/handheld/st2302u_bbl_spi.cpp
```diff
@@ -302,108 +314,132 @@ void bbl380_state::bbl380(machine_config &config)
...
+void bbl380_state::bbl380_menuprot(machine_config &config)
+{
+	bbl380(config);
+	m_maincpu->out_pb_callback().set(FUNC(bbl380_state::output2_w));
+}
+
+void bbl380_state::bbl380_24mhz(machine_config &config)
+{
+	bbl380(config);
+	ST2302U(config.replace(), m_maincpu, 24000000); // 24Mhz clock correct for music tempo. SoC type not confirmed
```
cuavas: Same here – MHz and digit grouping.

src/mame/handheld/st2302u_bbl_spi.cpp
```diff
@@ -302,108 +314,132 @@ void bbl380_state::bbl380(machine_config &config)
...
+
+// internal OTPROM BIOS, dumped from dgun2953 PCB, 6000-7fff range
+#define INTERNAL_ROM_TYPE1 \
+	ROM_REGION(0x800000, "maincpu", ROMREGION_ERASEFF) \
+	ROM_LOAD("st2x_internal_type1.bin", 0x002000, 0x002000, CRC(f4dc1fc2) SHA1(bbc11539c48eb612ebae50da45e03b6fde440941))
+
+// internal OTPROM BIOS, dumped from retro150a PCB, 6000-7fff range
+#define INTERNAL_ROM_TYPE2 \
+	ROM_REGION(0x800000, "maincpu", ROMREGION_ERASEFF) \
+	ROM_LOAD("st2x_internal_type2.bin", 0x002000, 0x002000, CRC(32d96794) SHA1(9d7e3e284f1656d8b2f7dae754cab1f82b3a1d61))
```
cuavas: Should these really have such a large CPU ROM region?  Only half this size is actually mapped in the address map, and all the systems seem to have much smaller content than this.

mamehaze:
feedback has been addressed

PR #14014, opened 2025-08-01
Skeleton driver for Criterium 75 TTL pinball from Petaco

clawgrip:
Criterium 75 [Orlando STRPinball, Recreativas.org]

### DIFF
--- a/src/mame/pinball/criter75.cpp
+++ b/src/mame/pinball/criter75.cpp
@@ -1,0 +1,133 @@
+// license:BSD-3-Clause
+// copyright-holders:
+/****************************************************************************************************
+
+  Skeleton driver for Petaco Criterium 75 pinball.
+
+  There's an older (1975) versión of this pinball, from Recel, but it's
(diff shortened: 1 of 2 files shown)


src/mame/pinball/criter75.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+void criter75_state::criter75(machine_config &config)
+{
+	// basic machine hardware
+	NETLIST_CPU(config, m_maincpu, netlist::config::DEFAULT_CLOCK()).set_source(netlist_criter75);
+}
+
+ROM_START( criter75 )
+	ROM_REGION( 0x10000, "maincpu", ROMREGION_ERASE00 )
+	// There's no CPU ROMs
```
cuavas: Why do you have a CPU region for a system that has no CPU?

clawgrip: Fixed, thanks!!!

PR #14015, opened 2025-08-01
integraph/interpro.cpp: Add missing PLDs/PROMs and fix IDPROM for ip2700 [ArcadeHacker]

clawgrip:


src/mame/intergraph/interpro.cpp
```diff
@@ -1643,7 +1643,7 @@ ROM_END
 ROM_START(ip2700)
 	// feature[0] & 0x04: supports RETRY if clear
 	ROM_REGION32_LE(0x80, "idprom", 0)
-	ROM_LOAD32_BYTE("msmt1280.bin", 0x00, 0x20, CRC(32d833af) SHA1(7225c5f5670fe49d86556a2cb453ae6fe09e3e19))
+	ROM_LOAD("mprgz260f_9310270_82s123.u118", 0x000, 0x020, CRC(bef9a02b) SHA1(5a50f8922fc0a7ae5f7934ccc1e178f3c8ec8489))
```
cuavas, line 1646: What does this change in the IDPROM?  Does it change the host ID or similar?  If it does, was the previous one actually bad, or was it just from a different unit?

@pmackinlay was there anything special about how the IDPROM content was chosen?

pmackinlay: The previous file was hand crafted not dumped (due to lack of hardware availability), and should have been marked a BAD_DUMP. I don't expect any issue replacing this one (or any others) with a real dump.

PR #14019, opened 2025-08-03
chdman: fix file permissions for verify --fix

landfillbaby:
gives EBADF on sha1 write currently due to opening in read-only mode. is the variable `fix_sha1` named ok?


landfillbaby:
well nevermind, now it throws FILE_NOT_WRITEABLE on open with --fix instead. although that might be better?

landfillbaby:
ok it works! i duplicated some error handling code though, could refactor

PR #14020, opened 2025-08-03
adp/stella8085.cpp: lamps, layout and 5 more games

stonedDiscord:
# new NOT WORKING software list entries

Bahia [stonedDiscord]
Disc [stonedDiscord]
Doppelstart [stonedDiscord]
Macao [stonedDiscord]
Rasant [stonedDiscord]


stonedDiscord:
Steps to start the machines:
Assign inputs to `Hoch 1,-` and `Initialisieren` (i use H and I)
Press them both at the same time to make the machine initialize its NVRAM.
The error LED will stop blinking and the machine will do a self-test which fails.
It will display the error code `E000 0002` on the service display.

The non gambling ones have code to self initialize when the battery runs out and go into the error condition directly.

stonedDiscord:
This also adds the Intel 8256 with timers and some IO, it's serial part isn't finished but not required to start the machine.

stonedDiscord:
is there anything wrong?
should i split this up and submit the I8256 separately?
I'm stuck on this for now and would like to submit it in this state.
The most complete machine is disc2000 (the one I own at home) with a nice layout.
I don't really have a TODO to write into this because I'm not sure what exactly the problem is, my suspicion was that it expects the discs to spin and see flashing on the light sensors but it never attempts to spin up the wheels. The error code says that only motor 2 failed and *something's* wrong with the coin counter which doesn't make sense.
I think I can continue working on this when I finish my reverse engineered schematic of the board.

PR #14021, opened 2025-08-03
glfgreat: uses k053260's TIM2 output as interrupt source.

jotego:
The sound interrupt for the Z80 is wrong on glfgreat. It should be the TIM2 output of the K053260. This output produces a pulse with a 2ms period, hence the pin name. Furrtek's schematics for this chip show the function.

The M68000 does not produce an interrupt on the Z80 on this board.

I have modified the files to implement this, but given my poor understanding of the MAME framework, something might be wrong. Particularly, the K053260 should set the interrupt request and it should not be cleared until the Z80 starts the interrupt execution. I do not know if my changes implement that correctly.

There is a note about wrong audio before but I have not really listened to the audio from the old driver as it must be wrong if a 500Hz interrupt becomes a random interrupt from the M68000. 

This is a picture from the original schematics. I have a board and I have measured these pins and the schematic is accurate.

Please push to the same branch on my repo or ask me if you need further changes on this. I am not sure whether this is enough to remove the "BAD SOUND" warning from the driver.

glfgreat

### DIFF
--- a/src/devices/sound/k053260.cpp
+++ b/src/devices/sound/k053260.cpp
@@ -184,7 +184,8 @@ TIMER_CALLBACK_MEMBER(k053260_device::update_state_outputs)
 // least. These counters together produce a count that is 112 times slower than
 // the SH1 pin, taking it down from 56kHz to 500Hz. The period of 500Hz is 2ms,
 // hence the pin name TIM2 and also the signal name 2MS seen in glfgreat schematics.
-void k053260_device::tim2_count() {
+void k053260_device::tim2_count()
+{
 	bool over = m_tim2 >= 111;
 	if (over)
 	{
--- a/src/devices/sound/k053260.cpp
+++ b/src/devices/sound/k053260.cpp
@@ -179,8 +179,13 @@ TIMER_CALLBACK_MEMBER(k053260_device::update_state_outputs)
 	m_timer_state = (m_timer_state+1) & 3;
 }
 
+// TIM2 is implemented on the original chip as two counters in series, where the
+// first one may be bypassed with a test bit, which is not used by glfgreat at
+// least. These counters together produce a count that is 112 times slower than
+// the SH1 pin, taking it down from 56kHz to 500Hz. The period of 500Hz is 2ms,
+// hence the pin name TIM2 and also the signal name 2MS seen in glfgreat schematics.
 void k053260_device::tim2_count() {
-	bool over = m_tim2>=111;
+	bool over = m_tim2 >= 111;
 	if (over)
 	{
 		m_tim2 = 0;
--- a/src/devices/sound/k053260.cpp
+++ b/src/devices/sound/k053260.cpp
@@ -95,11 +95,13 @@ k053260_device::k053260_device(const machine_config &mconfig, const char *tag, d
 	, device_rom_interface(mconfig, *this)
 	, m_sh1_cb(*this)
 	, m_sh2_cb(*this)
+	, m_tim2_cb(*this)
 	, m_stream(nullptr)
 	, m_timer(nullptr)
 	, m_keyon(0)
 	, m_mode(0)
 	, m_timer_state(0)
+	, m_tim2(0)
 	, m_voice{ { *this }, { *this }, { *this }, { *this } }
 {
 	std::fill(std::begin(m_portdata), std::end(m_portdata), 0);
@@ -119,6 +121,7 @@ void k053260_device::device_start()
 	save_item(NAME(m_keyon));
 	save_item(NAME(m_mode));
 	save_item(NAME(m_timer_state));
+	save_item(NAME(m_tim2));
 
 	for (int i = 0; i < 4; i++)
 		m_voice[i].voice_start(i);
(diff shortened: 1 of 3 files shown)


src/devices/sound/k053260.cpp
```diff
@@ -169,13 +172,23 @@ TIMER_CALLBACK_MEMBER(k053260_device::update_state_outputs)
...
+void k053260_device::tim2_count() {
+	bool over = m_tim2>=111;
+	m_tim2 = over ? 0 : m_tim2+1;
+	if(over) {
+		m_tim2_cb(ASSERT_LINE);
+	} else {
+		m_tim2_cb(CLEAR_LINE);
+	}
+}
+
```
happppp, line 201: Can be simplified:

```
void k053260_device::tim2_count()
{
	m_tim = (m_tim + 1) % 111;
	m_tim2_cb((m_tim2 == 0) ? ASSERT_LINE : CLEAR_LINE);
}
```

Is this 111 value from Furrtek's schematics of this chip?

src/mame/konami/tmnt2.cpp
```diff
@@ -2499,6 +2501,7 @@ void glfgreat_state::glfgreat(machine_config &config)
 	m_k053260->add_route(0, "speaker", 1.0, 1);
 	m_k053260->add_route(1, "speaker", 1.0, 0);
 	m_k053260->sh1_cb().set(FUNC(glfgreat_state::z80_nmi_w));
+	m_k053260->tim2_cb().set(FUNC(glfgreat_state::z80_int_w));
```
happppp: No trampoline needed, I suggest:
m_k053260->tim2_cb().set_inputline(m_audiocpu, 0, HOLD_LINE);

FYI what you have at z80_int_w will write it much more often than 2ms, and will never clear IRQ.
The correct way over there would be: if (state) m_audiocpu->set_input_line(0, HOLD_LINE);

src/devices/sound/k053260.cpp
jotego: I have made it a bit more consistent:
``` C++
void k053260_device::tim2_count() {
	bool over = m_tim2>=111;
	if (over)
	{
		m_tim2 = 0;
		m_tim2_cb(ASSERT_LINE);
	}
	else
	{
		m_tim2++;
		m_tim2_cb(CLEAR_LINE);
	}
}
```

It might take more lines, but I think it is easier to read. Also, the modulus operator is likely to be less efficient once compiled.

TIM2 is implemented on the original chip as two counters in series, where the first one may be bypassed with a test bit, which is not used by glfgreat at least.

These counters together produce a count that is 112 times slower than the SH1 pin, taking it down from 56kHz to 500Hz. The period of 500Hz is 2ms, hence the pin name TIM2 and also the signal name 2MS in the schematics.

src/mame/konami/tmnt2.cpp
jotego: I have modified it as per your suggestion:
```C++
m_k053260->tim2_cb().set_inputline(m_audiocpu, 0, HOLD_LINE);
```
I have deleted the `z80_int_w` helper function. Should this produce the intended effect?

src/devices/sound/k053260.cpp
happppp: Could you add  the description of the TIM2 counter to a comment in tim2_count()? (aka what you just wrote)
Also, please keep similar source code style as the rest of the file: spacing and braces.

if(x) -> if (x)
blah { -> blah [enter] {
x>=111 -> x >= 111

src/mame/konami/tmnt2.cpp
happppp: Yes.

src/devices/sound/k053260.cpp
jotego: Thanks. I have added the comment and matched the style of the rest of the file.

```C++
// TIM2 is implemented on the original chip as two counters in series, where the
// first one may be bypassed with a test bit, which is not used by glfgreat at
// least. These counters together produce a count that is 112 times slower than
// the SH1 pin, taking it down from 56kHz to 500Hz. The period of 500Hz is 2ms,
// hence the pin name TIM2 and also the signal name 2MS seen in glfgreat schematics.
void k053260_device::tim2_count() {
	bool over = m_tim2 >= 111;
	if (over)
	{
		m_tim2 = 0;
		m_tim2_cb(ASSERT_LINE);
	}
	else
	{
		m_tim2++;
		m_tim2_cb(CLEAR_LINE);
	}
}
```

happppp: Cool.
I still see 1 mismatched brace at the top ;)

jotego: Right! I completely missed that one. Fixed:

```C++
void k053260_device::tim2_count()
{
	bool over = m_tim2 >= 111;
	if (over)
	{
		m_tim2 = 0;
		m_tim2_cb(ASSERT_LINE);
	}
	else
	{
		m_tim2++;
		m_tim2_cb(CLEAR_LINE);
	}
}
```

src/mame/konami/tmnt2.cpp
angelosa: Wait, isn't it supposed to clear the line with state == 0?

happppp: jotego said nope. Only when Z80 takes the IRQ.

jotego: There is a flip flop that latches the interrupt request until the Z80 serves it. See device **2E**, the CLR pin is controlled by the Z80's IORQ pin only.

happppp:
I tested it locally before merging and yeah, the sound issues are gone.

src/devices/sound/k053260.cpp
```diff
@@ -169,13 +172,33 @@ TIMER_CALLBACK_MEMBER(k053260_device::update_state_outputs)
...
+	bool over = m_tim2 >= 111;
+	if (over)
+	{
+		m_tim2 = 0;
+		m_tim2_cb(ASSERT_LINE);
+	}
+	else
+	{
+		m_tim2++;
+		m_tim2_cb(CLEAR_LINE);
```
cuavas, line 198: You really shouldn’t be calling the devcb every time, only when it changes – that’s going to be a drain on performance.

happppp: From what I've seen, yeah it's a performance drain, when it's in the MHz range (eg. a CPU with a devcb after every opcode).

happppp: https://github.com/mamedev/mame/commit/2a3290a8647a08d42bf9a37a4776e35f46998b0e

PR #14029, opened 2025-08-05
Commit for Foenix F256K retro-computer.

dtremblay:
More information about the hardware is available here: https://wiki.f256foenix.com/

For the list of supported features, see: https://wiki.f256foenix.com/index.php?title=Emulation#MAME.

For ROMs, see the release page: https://github.com/dtremblay/mame/releases
The ROM is called f256k.zip - there is also a SD Card image available there.


.github/workflows/ci-linux.yml
angelosa, line 1: Can you please revert this CI addition?

src/mame/f256/f256.cpp
angelosa, line 1: For every file added, needs the SPDX license you want these files and your handle at top, in the form:

```c++
// license: <license_spdx>
// copyright-holders: <your_handle>
```

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2002 @@
...
+    m_maincpu->set_addrmap(AS_PROGRAM, &f256_state::program_map);
+
+    m_screen->set_refresh_hz(60); // Refresh rate (e.g., 60Hz)
+    m_screen->set_size(800,525);
+    //m_screen->set_visarea(160, 799, 45, 524);  // this is how it should reall work, but the screen ends up offset
+    m_screen->set_visarea(0, 639, 0, 479);
+    m_screen->set_screen_update(m_video, FUNC(tiny_vicky_video_device::screen_update));
+    TINY_VICKY(config, m_video, MASTER_CLOCK);
+    m_video->sof_irq_handler().set(FUNC(f256_state::sof_interrtupt));
+    m_video->sol_irq_handler().set(FUNC(f256_state::sol_interrtupt));
```
angelosa: We generally put setters near their construction, like:

```c++
    W65C02(config, m_maincpu, MASTER_CLOCK/4);
    m_maincpu->set_addrmap(AS_PROGRAM, &f256_state::program_map);

    RAM(config, m_ram).set_default_size("512k").set_default_value(0x0);
    RAM(config, m_iopage0).set_default_size("8k").set_default_value(0x0);
    RAM(config, m_iopage1).set_default_size("8k").set_default_value(0x0);
    RAM(config, m_iopage2).set_default_size("8k").set_default_value(0x0);
    RAM(config, m_iopage3).set_default_size("8k").set_default_value(0x0);

    BQ4802(config, m_rtc, MASTER_CLOCK / 1000);  // RTC clock in kHz

    SCREEN(config, m_screen, SCREEN_TYPE_RASTER);
    m_screen->set_refresh_hz(60); // Refresh rate (e.g., 60Hz)
    m_screen->set_size(800,525);
    //m_screen->set_visarea(160, 799, 45, 524);  // this is how it should reall work, but the screen ends up offset
    m_screen->set_visarea(0, 639, 0, 479);
    m_screen->set_screen_update(m_video, FUNC(tiny_vicky_video_device::screen_update));

    TINY_VICKY(config, m_video, MASTER_CLOCK);
    m_video->sof_irq_handler().set(FUNC(f256_state::sof_interrtupt));
    m_video->sol_irq_handler().set(FUNC(f256_state::sol_interrtupt));
```

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2002 @@
...
+};
+
+void f256_state::data_map(address_map &map)
+{
+    map(0x0000, 0x1FFF).ram().share(IOPAGE0_TAG);
+    map(0x0000, 0x1FFF).ram().share(IOPAGE1_TAG);
+    map(0x0000, 0x1FFF).ram().share(IOPAGE2_TAG);
+    map(0x0000, 0x1FFF).ram().share(IOPAGE3_TAG);
+}
+
```
angelosa: Huh? This will just map `IOPAGE3_TAG` for the `address_space` rule of last one wins, how it's supposed to really work?

dtremblay: I agree this is confusing.  The F256K has an IO register in byte 0x01 which determines which IO devices to map.  This way more devices can be supported with only a 16-bit bus.

.github/workflows/ci-linux.yml
dtremblay: I didn't intend to share this in the PR.  I'll remove.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2008 @@
...
+    , m_sdcard(*this, "sdcard")
+	, m_spi_clock_state(false)
+	, m_spi_clock_sysclk(false)
+	, m_spi_clock_cycles(0)
+{
+}
+
+void f256_state::f256k(machine_config &config)
+{
+    W65C02(config, m_maincpu, MASTER_CLOCK/4);
```
ajrhacker: I believe this should be using the W65C02S variant.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2008 @@
...
+    
+    BQ4802(config, m_rtc, MASTER_CLOCK / 1000);  // RTC clock in kHz
+    //set interrupt handler for the RTC
+    m_rtc->int_handler().set(FUNC(f256_state::rtc_interrupt_handler));
+    
+    TINY_VICKY(config, m_video, MASTER_CLOCK);
+    m_video->sof_irq_handler().set(FUNC(f256_state::sof_interrtupt));
+    m_video->sol_irq_handler().set(FUNC(f256_state::sol_interrtupt));
+
+    MOS6522(config, m_via6522_0, MASTER_CLOCK / 4);  // Atari Joysticks
```
ajrhacker: I believe this should be using the W65C22S variant (though MAME does emulate the subtle differences between VIA variants at present).

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2008 @@
...
+    }
+}
+
+//-------------------------------------------------
+//  VIA0 - JOYSTICK
+//-------------------------------------------------
+u8 f256_state::via0_system_porta_r()
+{
+    //logerror("VIA #0 Port A Read ioport JOY2: %02X\n", data);
+    return ioport("JOY2")->read();
```
ajrhacker: These port reads should be using `required_ioport` or similar finder objects. (Actually, these should ultimately be refactored to use slot devices, which would make the full range of console-compatible controllers available.)

ajrhacker (changes requested):

src/mame/f256/f256.cpp
dtremblay: OK. I've updated this.

src/mame/f256/f256.cpp
dtremblay: I could not confirm this, but it is a surface mount chip, so a recent VIA chip.  So I've made the change.

src/mame/f256/f256.cpp
dtremblay: I am uncertain how to use "slot devices".  I have added two required_ioport devices for the joysticks (Atari VCS style) and re-tested.

dtremblay:
Addressed all comments.

ajrhacker (approved):

dtremblay:
So... am I supposed to do something else to get this PR accepted? It seems to be stuck.  The workflow hasn't been approved yet.

stonedDiscord:
It sometimes takes a few days or weeks until a maintainer has time to look at PRs that aren't just a few lines or a new dump.
The workflow did run and it seems it failed.
Click the red X next to your latest commit above.

src/devices/machine/bq4847.cpp
```diff
@@ -101,7 +102,11 @@ bq4845_device::bq4845_device(const machine_config& mconfig, const char* tag, dev
 	: bq4847_device(mconfig, BQ4845, tag, owner, clock)
 {
 }
-
+bq4802_device::bq4802_device(const machine_config& mconfig, const char* tag, device_t* owner, uint32_t clock)
+	: bq4847_device(mconfig, BQ4845, tag, owner, clock)
```
ajrhacker: There's a copy-and-paste error here that is causing validation to fail.

src/mame/f256/f256.h
```diff
@@ -0,0 +1,207 @@
...
+	required_memory_region m_rom;
+	required_memory_region m_font;
+	required_device<screen_device> m_screen;
+	required_device<bq4802_device> m_rtc;
+	required_ioport_array<8> m_keyboard; // the number 8 will require 8 COL
+	required_device<via6522_device> m_via6522_0, m_via6522_1;
+	required_device<sn76489_device> m_sn0, m_sn1;
+	required_device<ymf262_device> m_opl3;
+	required_device<mos6581_device> m_sid0, m_sid1;
+	required_ioport m_joy1, m_joy2;
```
angelosa: `m_iopage`, `m_via6522`, `m_sn` and `m_sid` should all be `required_device_array`
`m_joy` should be a `required_ioport_array`

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2010 @@
...
+    uint8_t bank = mmu_lut[mmu * 8 + slot];
+
+    // fslot < 0x40 is RAM, greater is FLASH/ROM
+    if (bank < 0x40)
+    {
+        // Slot 6 is where I/O devices are located, when IO_DISABLE is 0
+        if (slot == 6 && (m_ioreg & 0x4) == 0)
+        {
+            switch (m_ioreg & 0x3)
+            {
```
angelosa, line 262: Feel like this could be expressed with `address_map_bank` and use of `memory_view` specifically for the 0xc000-0xdfff area, to me it looks similar to a PC Engine or an OS-9 based machine like Hitachi S1.
Also: how much RAM this is really supposed to have?

dtremblay: The machine has 512K of flash, 512K of static RAM.  There's also an extension cartridge for 256K available as flash or RAM.

src/mame/f256/f256.h
dtremblay: So you want me to combine m_sn0 and m_sn1 has a required_device_array, and so on?

src/mame/f256/f256.cpp
dtremblay: I couldn't understand the memory bank documentation.  Could this be done as an update?  I intend to add other variants of the machine (with 65816 CPU, for example).

angelosa: https://wiki.f256foenix.com/index.php?title=Memory_Management

Need to point me where's the I/O explaination.

PS: yeah it _can_ run OS-9, with a special 6809 flavour if I'm getting it right?

src/mame/f256/f256.h
MooglyGuy: Yes. See for example src/mame/atari/klax.cpp, member `m_msm` for usage, or do a Find In Files for `required_device_array`.

MooglyGuy: See also, for example, member `m_inputs` in src/mame/atari/missle.cpp for use of `required_ioport_array`.

MooglyGuy:
One overarching bit of feedback, it's usually not ideal to suggest that feedback be done via a Discord server. Part of the point of a pull request is that the code review occurs within the pull request itself.

While there is certainly a particular generational shift towards the use of Discord as a substitute for websites, that isn't something that tends to be shared by the wider software development community, at least those old enough to have been doing professional software development for the past couple of decades.

dtremblay:
> One overarching bit of feedback, it's usually not ideal to suggest that feedback be done via a Discord server. Part of the point of a pull request is that the code review occurs within the pull request itself.
> 
> While there is certainly a particular generational shift towards the use of Discord as a substitute for websites, that isn't something that tends to be shared by the wider software development community, at least those old enough to have been doing professional software development for the past couple of decades.

In my commit message, I state to report "defects" in Discord.  I did not suggest to perform the PR work there.
About generational comment: I think this is not warranted in a PR.  And, I'm probably as old as you are if not older.

src/mame/f256/f256.h
dtremblay: I've made the requested modifications and retested.

angelosa:
Out of sheer curiosity: how this driver behaves when testing?
You are claiming that the driver is working, and by cursory glance I bet it really isn't.

src/devices/machine/bq4847.cpp
```diff
@@ -51,7 +52,7 @@ enum
 	reg_interrupts,         // 0  0  0  0 AIE  PIE   PWRIE ABE    0x00 on powerup
 	reg_flags,              // 0  0  0  0  AF   PF    PWRF BVF    0x00 after reading
 	reg_control,            // 0  0  0  0 UTI STOP* 24/12* DSE
-	reg_unused              // 0x00
+	reg_century             // 0x00-0x99
```
pmackinlay: Please add a trailing comma to match the existing style.

src/devices/machine/bq4847.cpp
```diff
@@ -101,7 +102,11 @@ bq4845_device::bq4845_device(const machine_config& mconfig, const char* tag, dev
 	: bq4847_device(mconfig, BQ4845, tag, owner, clock)
 {
 }
-
+bq4802_device::bq4802_device(const machine_config& mconfig, const char* tag, device_t* owner, uint32_t clock)
```
pmackinlay: Please use right justification for reference and pointer tokens to match existing style.

src/devices/machine/bq4847.h
```diff
@@ -84,14 +89,21 @@ class bq4847_device : public device_t,
 	int m_rst_state;
 	int m_wdi_state;
 	bool m_writing;
+	bool has_century = false;
```
pmackinlay: Please use `m_` prefix on class variables to match existing style. Also initialize this variable in the constructor not the declaration (see e.g., `m_writing`).

src/devices/machine/bq4847.h
```diff
@@ -84,14 +89,21 @@ class bq4847_device : public device_t,
...
 };
 
 class bq4845_device : public bq4847_device
 {
 public:
 	bq4845_device(const machine_config& mconfig, const char* tag, device_t* owner, uint32_t clock);
 };
-
```
pmackinlay, line 94: Please retain existing separating blank lines.

src/devices/machine/bq4847.h
```diff
@@ -84,14 +89,21 @@ class bq4847_device : public device_t,
...
+class bq4802_device : public bq4847_device
+{
+public:
+	bq4802_device(const machine_config& mconfig, const char* tag, device_t* owner, uint32_t clock);
```
pmackinlay: As before, please use right justification for references and pointers to match existing style.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+    $08:0000 - $0F:FFFF	Flash
+    $10:0000 - $13:FFFF	Expansion Memory - NOT IMPLEMENTED
+    $14:0000 - $1F:FFFF	Reserved
+*/
+void f256_state::program_map(address_map &map)
+{
+    // the address range 0:F
+    // 0: LUT Edit $80 allows writing to 8 to F, LUT Select 0 to 3
+    map(0x0000, 0x000F).rw(FUNC(f256_state::lut_r), FUNC(f256_state::lut_w));
+    map(0x0010, 0xFFFF).rw(FUNC(f256_state::mem_r), FUNC(f256_state::mem_w));
```
pmackinlay: The purpose of the address map is to avoid having to do all of the address decoding in the handlers. `mem_r` and `mem_w` should be replaced by adding entries to the address map here.

src/mame/f256/f256.h
```diff
@@ -0,0 +1,205 @@
...
+
+protected:
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+
+private:
+    required_device<cpu_device> m_maincpu;
+	required_device<ram_device> m_ram;
+	required_device_array<ram_device, 4> m_iopage;
```
pmackinlay, line 60: Using a `ram_device` doesn't make much sense for these. Since you're already allocating memory using the `ram()` in `data_map()`, you should change these to be a `required_shared_ptr` instead and remove the code in `machine_config`.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+    m_addition_result = acc1 + acc2;
+}
+
+//-------------------------------------------------
+//  DMA Methods
+//-------------------------------------------------
+uint8_t f256_state::get_random()
+{
+    uint8_t m_random = rand();
+    return m_random & 0xFF;
```
pmackinlay: The mask is unnecessary. Also please use lowercase for hex literals.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+}
+//-------------------------------------------------
+//  DMA Methods
+//-------------------------------------------------
+void f256_state::perform2DFillDMA()
+{
+
+    uint8_t fill_byte = m_iopage[0]->read(0xDF01 - 0xC000);
```
pmackinlay: Are these `m_iopage` addresses RAM, or memory-mapped I/O devices?

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+}
+
+//-------------------------------------------------
+//  device_reset
+//-------------------------------------------------
+void f256_state::device_reset()
+{
+	driver_device::device_reset();
+    reset_mmu();
+    m_via6522[0]->reset();
```
pmackinlay, line 1527: Subdevices will automatically be reset, don't reset them manually.

src/mame/f256/f256.h
```diff
@@ -0,0 +1,205 @@
...
+	uint8_t m_mmu_reg, m_ioreg;
+	uint8_t mmu_lut[32];
+	void reset_mmu();
+	u8   lut_r(offs_t offset);
+	void lut_w(offs_t offset, u8 data);
+	u8   mem_r(offs_t offset);
+	void mem_w(offs_t offset, u8 data);
+
+	// screen update
+	void sof_interrtupt(int state);
```
pmackinlay: Typo `interrtupt`.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+    m_timer1_load = 0;
+    m_timer1_val = 0;
+}
+
+//-------------------------------------------------
+//  Interrupts
+//-------------------------------------------------
+void f256_state::sof_interrtupt(int state)
+{
+    if (state) // && ((m_interrupt_masks[1] & 0x01) == 0))
```
pmackinlay, line 1557: This condition prevents the interrupting device from deasserting the interrupt to the CPU?

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+{
+    //logerror("VIA #0 Port A Read ioport JOY2: %02X\n", data);
+    return m_joy[1]->read();
+}
+u8 f256_state::via0_system_portb_r()
+{
+    //logerror("VIA #0 Port B Read ioport JOY1: %02X\n", m_via_joy1);
+    return m_via_joy1;
+}
+void f256_state::via0_system_porta_w(u8 data)
```
pmackinlay: These trampolines (through `via0_cb2_write()`) shouldn't be necessary.

src/mame/f256/f256.h
```diff
@@ -0,0 +1,205 @@
+// license:BSD-3-Clause
+// copyright-holders:Daniel Tremblay
+#ifndef MAME_F256_F256_H
+#define MAME_F256_F256_H
+
+#pragma once
+
+#include "machine/ram.h"
```
pmackinlay: Order headers from most to least specific, and in alphabetic order within groups.

src/mame/f256/f256.h
```diff
@@ -0,0 +1,205 @@
...
+#include "bus/pc_kbd/hle_mouse.h"
+#include "debug/debugcon.h"
+#include "debugger.h"
+//#include "bus/cbmiec/cbmiec.h"
+//#include "bus/cbmiec/c1581.h"
+#include "machine/ins8250.h"
+
+#define MASTER_CLOCK        (XTAL(25'175'000))
+#define MUSIC_CLOCK         (XTAL(14'318'181))
+#define MAINCPU_TAG                 "maincpu"
```
pmackinlay, line 35: In general, using finders should remove the need for these `TAG` preprocessor macros.

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+//-------------------------------------------------
+void f256_state::device_start()
+{
+	driver_device::device_start();
+    reset_mmu();
+    // TODO: Copy the font from file to IO Page 1
+    //memcpy(m_iopage[1], m_font, 0x800);
+    for (int i=0;i<0x800;i++)
+    {
+        m_iopage[1]->write(i, m_font->as_u8(i));
```
pmackinlay: Is this really what happens, or should the font ROM be decoded where `m_iopage[1]` is currently being decoded?

src/mame/f256/f256.cpp
```diff
@@ -0,0 +1,2011 @@
...
+        0xb9, 0xba, 0xbb, 0xbb, 0xbc, 0xbd, 0xbd, 0xbe, 0xbf, 0xbf, 0xc0, 0xc1, 0xc2, 0xc2, 0xc3, 0xc4,
+        0xc4, 0xc5, 0xc6, 0xc6, 0xc7, 0xc8, 0xc8, 0xc9, 0xca, 0xca, 0xcb, 0xcc, 0xcc, 0xcd, 0xce, 0xce,
+        0xcf, 0xd0, 0xd0, 0xd1, 0xd2, 0xd2, 0xd3, 0xd4, 0xd4, 0xd5, 0xd6, 0xd6, 0xd7, 0xd7, 0xd8, 0xd9,
+        0xd9, 0xda, 0xdb, 0xdb, 0xdc, 0xdc, 0xdd, 0xde, 0xde, 0xdf, 0xe0, 0xe0, 0xe1, 0xe1, 0xe2, 0xe3,
+        0xe3, 0xe4, 0xe4, 0xe5, 0xe6, 0xe6, 0xe7, 0xe7, 0xe8, 0xe9, 0xe9, 0xea, 0xea, 0xeb, 0xec, 0xec,
+        0xed, 0xed, 0xee, 0xef, 0xef, 0xf0, 0xf0, 0xf1, 0xf1, 0xf2, 0xf3, 0xf3, 0xf4, 0xf4, 0xf5, 0xf5,
+        0xf6, 0xf7, 0xf7, 0xf8, 0xf8, 0xf9, 0xf9, 0xfa, 0xfb, 0xfb, 0xfc, 0xfc, 0xfd, 0xfd, 0xfe, 0xff
+    };
+
+    memcpy(m_iopage[0]->pointer(), gamma_1_8, 256);
```
pmackinlay, line 1479: What is this gamma table in hardware?

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+}
+
+tiny_vicky_video_device::tiny_vicky_video_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock):
+    tiny_vicky_video_device(mconfig, TINY_VICKY, tag, owner, clock)
+{
+}
+
+rgb_t tiny_vicky_video_device::get_text_lut(uint8_t color_index, bool fg, bool gamma)
+{
+    uint8_t red =   iopage0_ptr[(fg ? 0x1800 : 0x1840) + color_index * 4 + 2];
```
pmackinlay: Is `iopage0_ptr` really pointing to a palette or RAMDAC device here?

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return m_sol_irq_handler.bind();
+    };
+protected:
+    // device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
+    // virtual void device_add_mconfig(machine_config &config) override;
+private:
+    bool running = false;
```
pmackinlay: Class variables should have `m_` prefix.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,81 @@
...
+    uint8_t *iopage3_ptr = nullptr;  // Pointer to IO Page 3
+
+    // cursor handling routines
+    bool cursor_visible = false;
+    bool enable_cursor_flash = true;
+    uint8_t cursor_counter = 0;
+    uint8_t cursor_flash_rate = 60;
+
+    // raster
+    rgb_t get_text_lut(uint8_t color_index, bool fg, bool gamma);
```
pmackinlay, line 55: Generally declare functions before variables

pmackinlay (changes requested):

dtremblay:
> Out of sheer curiosity: how this driver behaves when testing? You are claiming that the driver is working, and by cursory glance I bet it really isn't.

I'm not sure what you are basing your conclusion on.  I wouldn't submit the PR if the system was not working.  Maybe I misunderstand what you imply by "driver".  I can give you a demo, if you like, via MS Teams or I could upload a video to YouTube.  It works quite well on Windows, Linux and Mac - users in the community have tested this, not I.  

You do need to have the ROMs of course (provided: [here](https://github.com/dtremblay/mame/releases) ).  The ROMs include a disk operating system and SuperBASIC and other useful utilities.  If a hard disk image is provided, applications can be loaded.  The F256K machines have now been around for about two years and there are now new variants being manufactured now, for example with 65816 and 6809 processors (instead of 6502) and other sound chips.

angelosa:
> I'm not sure what you are basing your conclusion on. 

From your tendency of dismissing comments without actually doing anything. https://github.com/mamedev/mame/pull/14029#discussion_r2264050903
The driver is clearly not doing a logical to physical translation, and I can't see how a system that expects 512 KiB can work with a single page of 8 KiB, mapped on the same place on top.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+        cursor_counter++;
+        if (cursor_counter > cursor_flash_rate *2)
+        {
+            cursor_counter = 0;
+        }
+        // if MCR=0 or MCR bit 7 is set, then video is disabled
+        if (mcr != 0 && (mcr != 0x80))
+        {
+            // TODO: generate start of frame (SOF) interrupt
+            m_sof_irq_handler(1);
```
angelosa, line 63: No logic should be present in `screen_update`, that will break if user frameskips.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+            {
+                // Check the Sart of Line registers
+                uint8_t sol_reg = iopage0_ptr[0x1018];
+                // 12-bit line
+                uint16_t sol_line = iopage0_ptr[0x1018] + ((iopage0_ptr[0x101A] & 0xF) << 8);
+                uint32_t *row = topleft + y * 800;
+                if ((sol_reg & 1) != 0 && y == sol_line)
+                {
+                    // raise an interrupt
+                    m_sol_irq_handler(1);
```
angelosa, line 89: As above, this needs to be moved in an `emu_timer`, latching with screen positions.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+{
+    if (running && iopage0_ptr)
+    {
+        uint8_t mcr = iopage0_ptr[0x1000];
+        uint8_t mcr_h = iopage0_ptr[0x1001];
+        cursor_counter++;
+        if (cursor_counter > cursor_flash_rate *2)
+        {
+            cursor_counter = 0;
+        }
```
angelosa: Should use `frame_number()` instead of manually keeping track of the count. This will again break with frameskipping.

angelosa (changes requested):

src/devices/machine/bq4847.cpp
dtremblay: I don't understand what you mean by "right justification": I copied the constructor from a previous class, so my code is identical to the previous ones.

src/devices/machine/bq4847.h
dtremblay: All the code in this class is consistent.  I could modify the rest of the class to adhere to this request.

src/mame/f256/f256.cpp
dtremblay: In the machine, I/O are mapped to these addresses (0xC000 to 0xDFFF) in four pages.  These are the text and text color, the color Look Up Tables for text and graphics, the Gamma color correction table and the fonts.  This memory is not overlapping (or reusing) system RAM.

src/mame/f256/f256.h
dtremblay: I thought this was convention, as it is used more than 7000 times in 1200 files.  I misunderstood.

src/mame/f256/f256.cpp
dtremblay: The machine itself loads the font when the machine is booted.  The font area is R/W, so users can modify it.  I could not figure out a more elegant want to copy the font into the iopage area, hence the commented out memcpy.

src/mame/f256/f256.cpp
dtremblay: The gamma table is used to correct the display colors.  When gamma is enabled in the display register, all the colors displayed are adjusted according to this indexed table.

src/mame/f256/f256.cpp
dtremblay: The font is loaded at boot time.  Users can modify the font at run time.

src/mame/f256/tiny_vicky.cpp
dtremblay: It's a color Look Up Table (LUT), indeed like a palette.  But only 16 colors for foreground and 16 colors for background.  Each color is represented by 4 bytes: blue, green, red, and alpha (alpha is ignore by the machine so far).

src/mame/f256/tiny_vicky.h
dtremblay: I've modified all the private class variables to have the m_ prefix.

src/mame/f256/tiny_vicky.h
dtremblay: I've moved the functions before variables.

src/devices/machine/bq4847.h
dtremblay: I re-added the space.

src/mame/f256/f256.cpp
dtremblay: The I/O page information is not available in the wiki, unfortunately.  The PDF has the details:  [F256K PDF](https://github.com/pweingar/F256Manual/blob/main/tex/f256jr_ref.pdf).  See starting at page 14.

src/mame/f256/tiny_vicky.cpp
MooglyGuy: > It's a color Look Up Table (LUT), indeed like a palette. But only 16 colors for foreground and 16 colors for background. Each color is represented by 4 bytes: blue, green, red, and alpha (alpha is ignore by the machine so far).

Perfect use case for a `palette_device`. The final bitmap should be 32bpp, so you're solid in that regard, but iit's preferable to use a `palette_device` for even fairly baisc LUT remapping.

src/devices/machine/bq4847.cpp
pmackinlay: I'm referring to having spaces before the `*` and `&` tokens instead of after them. Most likely your editor is changing the justification of these without you noticing, but our convention is `const char *tag` not `const char* tag`, for example.

src/devices/machine/bq4847.h
pmackinlay: It's not consistent with respect to the spacing around `*` and `&` tokens.

src/mame/f256/f256.cpp
pmackinlay: This answer doesn't address the intent of the question. It's very unlikely that hardware performs copy of this data into RAM. What does the hardware do, if anything? What does "loaded at boot time" mean?

src/mame/f256/f256.cpp
pmackinlay: I understand what a gamma table is, I'm asking what hardware is involved in implementing it? Again, it's very unlikely that hardware is doing an arbitrary `memcpy()` from somewhere into RAM. Is it a RAMDAC or some other device with a fixed or programmable table that should be emulated?

src/mame/f256/tiny_vicky.cpp
pmackinlay: The intent of the question was to prompt you to explain what the underlying hardware is, such that it could be emulated in a more realistic/accurate fashion. Functionally it's a LUT; what hardware is used to implement it?

dtremblay: It's implemented via closed-source FPGA logic.

src/mame/f256/f256.cpp
dtremblay: This is implemented with closed-source FPGA logic.

src/mame/f256/f256.cpp
dtremblay: This is implement with closed-source FPGA logic.

src/devices/machine/bq4847.cpp
```diff
@@ -112,10 +119,15 @@ void bq4847_device::rtc_clock_updated(int year, int month, int day, int day_of_w
 			(((hour % 24) >= 12) ? 0x80 : 0x00) | convert_to_bcd((hour % 12) ? (hour % 12) : 12);
 		m_register[reg_minutes] = convert_to_bcd(minute);
 		m_register[reg_seconds] = convert_to_bcd(second);
-		m_register[reg_year] = convert_to_bcd(year);
+		m_register[reg_year] = convert_to_bcd(year%100);
```
cuavas: Please use spaces around binary operators, like `year % 100` (but not unary operators like `++x`).

src/devices/machine/bq4847.h
```diff
@@ -31,6 +31,11 @@ class bq4847_device : public device_t,
 
 	void write_wdi(int state); // watchdog disabled if wdi pin is left floating
 
+	void set_century(bool value)
+	{
+		m_century = value;
+	}
```
cuavas: If this is an intrinsic characteristic of the device as opposed to an input, it shouldn’t be part of the public interface.

src/devices/machine/bq4847.cpp
```diff
@@ -372,7 +398,7 @@ bool bq4847_device::is_clock_register(int regnum)
 {
 	return (regnum == reg_seconds || regnum == reg_minutes || regnum == reg_hours ||
 		regnum == reg_date || regnum == reg_days || regnum == reg_month
-		|| regnum == reg_year);
+		|| regnum == reg_year || regnum == reg_century);
 }
```
cuavas, line 402: This, combined with the change to `write`, will change the behaviour of *all* devices in this file to copy the contents of what’s the century register on some devices only on an update transfer.

src/devices/machine/bq4847.cpp
cuavas, line 1: In general, adding device type flags that need to be checked all over the place is an ugly way of doing things.  Can you think of a way of doing this without needing intrusive checks inserted throughout the code?

src/devices/machine/bq4847.h
cuavas: There’s still no space between the class declarations here.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
+// license:BSD-3-Clause
+// copyright-holders:Daniel Tremblay
+#ifndef MAME_F256_TINY_VICKY_H
+#define MAME_F256_TINY_VICKY_H
+
+#define CHAR_HEIGHT     8
+#define CHAR_WIDTH      8
```
cuavas: Please don’t pollute the global namespace, especially with macros with very generic names.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
...
+
+//#include <atomic>
+
+class tiny_vicky_video_device : public device_t
+{
+public:
+    tiny_vicky_video_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock);
+    tiny_vicky_video_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock = 0);
+    //tiny_vicky_video_device() = default;
```
cuavas: Internal constructors that take a device type for derived classes to use shouldn’t be public, and shouldn’t have default arguments.

You can’t have a defaulted constructor for a device – it makes no sense to have a commented one, either.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
...
+    uint32_t screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect);
+    void stop()
+    {
+        m_running = false;
+        m_videoram_ptr = nullptr;
+        m_iopage0_ptr = nullptr;
+        m_iopage1_ptr = nullptr;
+        m_iopage2_ptr = nullptr;
+        m_iopage3_ptr = nullptr;
+    }
```
cuavas, line 26: This will create confusion with `device_t::stop()`.  You should be using the regular device lifecycle functions.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
...
+        m_videoram_ptr = videoram;
+        m_iopage0_ptr = iopage0;
+        m_iopage1_ptr = iopage1;
+        m_iopage2_ptr = iopage2;
+        m_iopage3_ptr = iopage3;
+    }
+    void start()
+    {
+        m_running = true;
+    };
```
cuavas, line 38: Same with this – it’s a really bad idea to give it this name.

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
...
+        m_iopage3_ptr = nullptr;
+    }
+    void set_videoram(uint8_t *videoram, uint8_t *iopage0, uint8_t *iopage1, uint8_t *iopage2, uint8_t *iopage3)
+    {
+        m_videoram_ptr = videoram;
+        m_iopage0_ptr = iopage0;
+        m_iopage1_ptr = iopage1;
+        m_iopage2_ptr = iopage2;
+        m_iopage3_ptr = iopage3;
+    }
```
cuavas, line 34: Can you do this with object finders instead of passing pointers around?

src/mame/f256/tiny_vicky.h
```diff
@@ -0,0 +1,83 @@
...
+    uint8_t *m_iopage0_ptr = nullptr;  // Pointer to IO Page 0
+    uint8_t *m_iopage1_ptr = nullptr;  // Pointer to IO Page 1
+    uint8_t *m_iopage2_ptr = nullptr;  // Pointer to IO Page 2
+    uint8_t *m_iopage3_ptr = nullptr;  // Pointer to IO Page 3
+
+    // cursor handling routines
+    bool m_cursor_visible = false;
+    bool m_enable_cursor_flash = true;
+    uint8_t m_cursor_counter = 0;
+    uint8_t m_cursor_flash_rate = 60;
```
cuavas, line 75: Please don’t initialise members at declaration when the constructor is in a different file.  It’s too easy to end up with contradictory initialisations in the two files.

src/mame/mame.lst
```diff
@@ -18248,6 +18248,9 @@ topgunnr
 victorba
 victory
 
+@source:f256/f256.cpp
+f256k
+
```
cuavas: You could probably throw this in src/mame/homebrew, but even if you don’t, shouldn’t the project be foenixretro or something?  We avoid single-machine projects.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+            bool enable_graphics = (mcr & 0x4) > 0;
+            uint16_t lines = (mcr_h & 1) == 0 ? 480 : 400;
+
+            uint8_t border_x = 0;
+            uint8_t border_y = 0;
+            rgb_t border_color = rgb_t();
+            if (display_border) {
+                border_x = m_iopage0_ptr[0x1008];
+                border_y = m_iopage0_ptr[0x1009];
+            }
```
cuavas: Please be consistent with brace positioning.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+            uint32_t *topleft = (uint32_t *)bitmap.raw_pixptr(0);
+            for (int y = 0; y < lines; y++)
+            {
+                // Check the Sart of Line registers
+                uint8_t sol_reg = m_iopage0_ptr[0x1018];
+                // 12-bit line
+                uint16_t sol_line = m_iopage0_ptr[0x1018] + ((m_iopage0_ptr[0x101A] & 0xF) << 8);
+                uint32_t *row = topleft + y * 800;
```
cuavas, line 85: This isn’t safe use of bitmaps.  You should get a pointer to a row with `&bitmap.pix(y)`.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+                    }
+
+                    if (enable_graphics)
+                    {
+                        if (y % 2 == 0)
+                        {
+                            // Tiny Vicky Layers for Bitmaps, Tilemaps and sprites
+                            uint8_t LayerMgr0 = m_iopage0_ptr[0xD002 - 0xC000] & 0x7;
+                            uint8_t LayerMgr1 = m_iopage0_ptr[0xD002 - 0xC000] >> 4;
+                            uint8_t LayerMgr2 = m_iopage0_ptr[0xD003 - 0xC000] & 0x7;
```
cuavas: Please use `snake_case` for local variables.  We tend to reserve `LlamaCase` for template parameters.

Also, please consistently use lowercase hex digits.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+        return rgb_t(m_iopage0_ptr[0x800 + m_iopage1_ptr[lutAddress + 2]],
+                    m_iopage0_ptr[0x400 + m_iopage1_ptr[lutAddress + 1]],
+                    m_iopage0_ptr[m_iopage1_ptr[lutAddress]]);
+    }
+}
+
+uint32_t tiny_vicky_video_device::screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect)
+{
+    if (m_running && m_iopage0_ptr)
+    {
```
cuavas, line 55: In this function, you really should be making an effort to at least restrict the drawing to the clipping rectangle vertically so that raster effects have a chance of working.

src/mame/f256/tiny_vicky.cpp
```diff
@@ -0,0 +1,613 @@
...
+        if (x == maxX)
+        {
+            break;
+        }
+    }
+}
+
+void tiny_vicky_video_device::draw_mouse(uint32_t *row, bool enable_gamma, uint16_t line, uint16_t width, uint16_t height)
+{
+    uint8_t mouse_reg = m_iopage0_ptr[0xD6E0 - 0xC000];
```
cuavas: Is this “I/O page” actually shared RAM, or is it a bunch of memory-mapped registers?  If it’s memory-mapped registers, you should put proper read/write handlers in the address map rather than pretending it’s memory.

src/mame/f256/f256k.md
cuavas, line 1: Please don’t do this.  MAME has documentation, you don’t need a “how to use MAME” tutorial here.

src/mame/f256/f256.h
cuavas, line 1: You shouldn’t have a separate header for this if you aren’t realistically going to need derived classes in different translation units.  Just put it all in one file in an anonymous namespace.

src/mame/f256/f256.cpp
dtremblay: Fixed

src/mame/f256/f256.cpp
dtremblay: Done

src/devices/machine/bq4847.cpp
dtremblay: Understood.  Correct this file and applied to others.

src/devices/machine/bq4847.h
dtremblay: Done

src/devices/machine/bq4847.h
dtremblay: Fixed

src/mame/f256/f256.cpp
dtremblay: I have removed the loop and replaced with memcpy, using m_font->base() method.

src/devices/machine/bq4847.cpp
dtremblay: Fixed

src/devices/machine/bq4847.h
dtremblay: Moved to protected block.

src/mame/f256/tiny_vicky.h
dtremblay: Added **MAME_F256_** prefix to variable names.

src/mame/f256/tiny_vicky.h
dtremblay: I've removed the commented out code.
I've moved the offending constructor (with device type) to protected block.

src/mame/mame.lst
dtremblay: There are more machines coming, once I get over the hurdle of this first one.  There are F256K2, F256Jr and F256JrJr.  All different boards with slightly different capabilities such as sound chips or PS/2 keyboard and mouse, instead of matrix keyboard.  Also different processors.

Would committing to homebrew simplify the PR acceptance?

src/mame/f256/tiny_vicky.cpp
dtremblay: Fixed in three places.

src/mame/f256/tiny_vicky.h
MooglyGuy: @cuavas's point was to avoid #defines in headers in general. If you must use a header, make them static constexpr members of the driver class instead.

As an alternative, if you address his later feedback about using an anonymous namespace and keep everything to a single compile unit to omit the header at all, you can go hog-wild with whatever #defines you'd like.

src/mame/f256/tiny_vicky.cpp
dtremblay: Fixed.
Fixed all 0x values to lowercase.

src/mame/mame.lst
MooglyGuy: How different are the machines compared to the baseline? It could be viable to have derived machine classes within the same compile unit. Ideally, it would allow everything to stay within an anonymous namespace, and would also address the earlier feedback point by keeping #defines specifically within a compile unit rather than a header, although it could also be addressed via static constexpr members.

MooglyGuy: As for where the driver itself should go, putting it in src/mame/homebrew/ seems like the most straightforward path.

src/mame/f256/f256.cpp
dtremblay: The whole address range from 0x0000 to 0xffff is mapped in 8K blocks using a control register at address 0x0.
The address from 0xc000 to 0xdfff can be mapped to memory or to I/O using control register at address 0x1.
I'm re-reading the MAME documentation to see how I can implement memory_bank and memory_view concepts.

src/devices/machine/bq4847.cpp
dtremblay: Added trailing comma.

src/devices/machine/bq4847.cpp
dtremblay: You were correct, my editor was changing it for me.  Fixed.

src/mame/f256/f256k.md
dtremblay: OK. I'll remove the file.

src/mame/f256/f256.h
dtremblay: There will be more machines.  The next one is the same hardware with a 65816 processor and different address map for I/O.  I just wanted to get this one finished before I start something else.

src/mame/f256/tiny_vicky.cpp
dtremblay: I've removed the counter private variable and replaced with the screen.frame_number().
Verified that the cursor is respecting the register value.

src/mame/f256/f256.cpp
dtremblay: I'm looking into it.

src/mame/f256/f256.h
dtremblay: Fixed

src/mame/f256/f256.h
dtremblay: Re-ordered headers.

PR #14032, opened 2025-08-08
util/xmlfile.h: API update

ajrhacker:
- Convert most const char * parameters to std::string_view, cleaning up associated use of null pointers
- Rename get_name and get_value methods to name and value and make them return references to std::string

### DIFF
--- a/src/emu/audio_effects/eq.cpp
+++ b/src/emu/audio_effects/eq.cpp
@@ -112,19 +112,19 @@ void audio_effect_eq::config_load(util::xml::data_node const *ef_node)
 		reset_high_shelf();
 
 	for(u32 band = 0; band != BANDS; band++) {
-		if(ef_node->has_attribute(util::string_format("f%d", band+1).c_str())) {
+		if(ef_node->has_attribute(util::string_format("f%d", band+1))) {
 			m_f[band] = ef_node->get_attribute_int(util::string_format("f%d", band+1), 0);
 			m_isset_f[band] = true;
 		} else
 			reset_f(band);
 
-		if(ef_node->has_attribute(util::string_format("q%d", band+1).c_str())) {
+		if(ef_node->has_attribute(util::string_format("q%d", band+1))) {
 			m_q[band] = ef_node->get_attribute_float(util::string_format("q%d", band+1), 0);
 			m_isset_q[band] = true;
 		} else
 			reset_q(band);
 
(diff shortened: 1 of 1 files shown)


src/lib/util/xmlfile.h
```diff
@@ -132,34 +132,34 @@ class data_node
...
+	bool has_attribute(std::string_view attribute) const;
 
 	// return a pointer to the string value of an attribute, or nullptr if not present
-	std::string const *get_attribute_string_ptr(const char *attribute) const;
+	std::string const *get_attribute_string_ptr(std::string_view attribute) const;
 
-	// return the string value of an attribute, or the specified default if not present
-	const char *get_attribute_string(const char *attribute, const char *defvalue) const;
+	// return the string value of an attribute, or the specified default or empty string if not present
+	std::string_view get_attribute_string(std::string_view attribute, std::string_view defvalue = std::string_view()) const;
```
cuavas, line 141: I really dislike treating not present as an empty string by default.  It goes against the “XML way”.  Maybe you could make it return a `std::optional<std::string_view>` instead?  Then the default value could be `std::nullopt`.

src/lib/util/xmlfile.h
```diff
@@ -78,11 +78,11 @@ class data_node
...
 
-	char const *get_name() const { return m_name.empty() ? nullptr : m_name.c_str(); }
+	const std::string &name() const { return m_name; }
 
-	char const *get_value() const { return m_value.empty() ? nullptr : m_value.c_str(); }
-	void set_value(char const *value);
-	void append_value(char const *value, int length);
+	const std::string &value() const { return m_value; }
+	void set_value(std::string_view value) { m_value.assign(value); }
+	void append_value(std::string_view value) { m_value.append(value); }
```
cuavas, line 85: Exposing that the value is a `std::string` in the getter feels like a bit of a leaky abstraction.  I know you’re doing it (rather than returning `std::string_view`) for convenience because various things want to use the “value” for things that want a NUL-terminated string, and you can use `.c_str()` this way, but it feels dirtier before because it changes it from “value is internally stored as something that can be exposed as a NUL-terminated string` to “value is internally stored as a `std::string`”.

src/lib/util/xmlfile.h
ajrhacker: I recall a similar proposed use of `std::optional` in the options utility being shot down because of noexcept purism.

ajrhacker:
Closing this for now. It's just not that important.

PR #14033, opened 2025-08-08
misc/amaticmg: Add Fruit Poker (V.Swe 1.74)

FakeShemp:
I haven't really done any code for MAME before, so there probably are a lot of issues. Please point them out and I'll fix asap.

I don't really know much about the system so stuff like INIT is not based on anything, but a pure guess. 

FPoker

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -29381,6 +29381,7 @@ am_mg3a
 am_mg4sk
 am_mg4v
 am_mg5hu
+fpoker
 suprstar
 
 @source:misc/amerihok.cpp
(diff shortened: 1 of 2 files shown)


angelosa:
Looks fine by quick glance. Need whatsnew credit about who did the dump and perhaps a TODO note describing how it's behaving.

angelosa:
Sounds like CI is failing because roms are using uppercase characters (and extra `.`?). 

```c++
 	auto const valid_name_char = [] (char ch) { return ((ch >= '0') && (ch <= '9')) || ((ch >= 'a') && (ch <= 'z')) || (ch == '_'); };
```

You can test it locally by `./yourbuild -validate` after it compiled.

FakeShemp:
Thanks! I think those issues have been fixed now. I was the one who did the dump as well.

Osso13:
Please remember to submit the necessary data, like detailed here https://wiki.mamedev.org/index.php/Submitting_Source_Code

PR #14034, opened 2025-08-09
sequential/sixtrak.cpp: Preliminary driver for the Six-Trak.

m1macrophage:
--------------------
Sequential Circuits Six-Trak [DBWBP]


angelosa:
I'm missing something here: why it's not using [sound/cem3394.cpp](https://github.com/mamedev/mame/blob/master/src/devices/sound/cem3394.cpp)? Is it a technical reason?

m1macrophage:
> I'm missing something here: why it's not using [sound/cem3394.cpp](https://github.com/mamedev/mame/blob/master/src/devices/sound/cem3394.cpp)? Is it a technical reason?

Sound is not yet emulated. Sound emulation should indeed use sound/cem3394.

rb6502:
An important note here: a modified Six-Trak was used as the sound board in the Bally/Sente arcade hardware.  (Sente licensed it from Sequential).  That implementation is in sente6vb.cpp.

m1macrophage:
Ah, good to know. Looking at that implementation, there are similarities indeed. From a quick look, I didn't find anything that could be meaningfully abstracted out yet, but that might change as the Six-Trak driver evolves.

m1macrophage:
Applied suggestions from #14100: removed rom_ignore and fixed typos :)

PR #14035, opened 2025-08-09
Added default NVRAM for Downhill Bikers

MooglyGuy:


OctopusButtons:
I think there's a regression or problem with this, because sometime after 0.279 (which worked as expected) Downhill Bikers now gives an "nvram - not found" error in 0.285 and 0.284 and 0.283, regardless of whether I clear out the nvram folder or place the nvram file that was working in 0.279.

**Legendary work on Namco 3D by the way, it's fantastic to see.**

JimCarlTay:
> I think there's a regression or problem with this, because sometime after 0.279 (which worked as expected) Downhill Bikers now gives an "nvram not found" error in 0.285 and 0.284 and 0.283, regardless of whether I clear out the nvram folder or place the nvram file that was working in 0.279.
> 
> **Legendary work on Namco 3D by the way, it's fantastic to see.**

I'm not able to reproduce this on 0.285 or my nightly build. Meanwhile, the only System 23 game that still fails to boot on my host system (which runs Windows) is Race On (raceon), which stays at a gray screen after the self-tests.

OctopusButtons:
> I'm not able to reproduce this on 0.285 or my nightly build. Meanwhile, the only System 23 game that still fails to boot on my host system (which runs Windows) is Race On (raceon), which stays at a gray screen after the self-tests.

Oops, thanks for the info that yours was working, I found what I did wrong and it's fine now.

And yes, I see permanent grey screen on Race On! after POT.

PR #14041, opened 2025-08-11
megacd.xml: Six software items added

FoxhackDN:
New working software list items:
Loadstar - The Legend of Tully Bodine (USA, prototype) [redump.org]
Snatcher (USA, prototype 19941007) [redump.org]
Star Wars - Rebel Assault (USA, re-release) [redump.org]
The Secret of Monkey Island (USA, re-release) [redump.org]
WireHead (Europe, prototype 19950912) [redump.org]

New non-working software list items:
Heavy Nova (Japan, prototype 19911102) [redump.org]

### DIFF
--- a/hash/megacd.xml
+++ b/hash/megacd.xml
@@ -9676,7 +9676,7 @@ Glitchy tile in border area (verify)
 		<part name="cdrom" interface="cdrom">
 			<feature name="cd_matrix" value="MED2481   DISC MAKERS"/>
 			<diskarea name="cdrom">
-				<disk name="Secret of Monkey Island, The (USA) (Limited Run Games)" sha1="565cfd79331466d126d1b98048af79ee20a545b5"/>
+				<disk name="secret of monkey island, the (usa) (limited run games)" sha1="565cfd79331466d126d1b98048af79ee20a545b5"/>
 			</diskarea>
 		</part>
 	</software>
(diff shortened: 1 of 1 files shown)


FoxhackDN:
Turns out Snatcher has a US header, so it'll only boot in a US region system.

I'll edit the submission once I get more information from the dumper. Sorry about that.

FoxhackDN:
I have confirmed with the dumper that the Snatcher disc is actually a US build, and I'll amend the submission as soon as Redump updates their entry (so the CRCs and filenames match.)

hash/megacd.xml
```diff
@@ -136,6 +135,18 @@ Dumps that are known to exist, but are lost in the aether:
...
+|                            Title                            |                               Notes                               |
++=============================================================+===================================================================+
+| The Secret of Monkey Island                                 | Reissued by Limited Run in 2020. Only 2500 copies were available. |
+|                                                             | CRCs differ from original release but data should be identical.   |
++=============================================================+===================================================================+
+| Star Wars: Rebel Assault                                    | Reissued by Limited Run in 2020. Only 2500 copies were available. |
+|                                                             | Data track is identical to original release, but track 2 has a    |
+|                                                             | different CRC. Gameplay should be unchanged.                      |
++=============================================================+===================================================================+
```
cuavas: Are these placeholders for undumped discs, or is this just trivia?  If it’s just trivia, it feels like stuff that doesn’t belong here.

FoxhackDN: I figured both were still too new to be added (less than five years since release, despite them being OOP) but were worth mentioning. They have been dumped for a while.

If you want this info removed, then no problem.

cuavas: The general rule is three years since release.  These are re-issues of old games anyway, so I’d even be willing to bend the three-year rule a bit.

FoxhackDN: Allrighty. I'll look for the files and add them to the PR in the next few days.

PR #14043, opened 2025-08-11
psx.xml: Added 18 working items + 2 redumped items

ArcadeShadow:
Actua Pool (Europe) [Redump]
Adiboo & Paziral's Secret (Europe) [Redump]
Adibou et l'Ombre Verte (France) [Redump]
Alone in the Dark: Jack is Back (Europe) [Redump]
Alone in the Dark 2 (Japan) [Redump]
Amerzone: Das Testament des Forschungsreisenden (Germany) [Redump]
Amerzone: El Legado del Explorador (Spain) [Redump]
Amerzone: Il Testamento dell'Esploratore (Italy) [Redump]
Amerzone: O Testamento do Explorador (Portugal) [Redump]
Amerzone: The Explorer's Legacy (Europe) [Redump]
L'Amerzone (France) [Redump]
Pool Shark (Europe) [Redump]
TechnoMage: De Terugkeer der Eeuwigheid (Netherlands) [Redump]
TechnoMage: Die Rüeckkehr der Ewigkeit (Germany) [Redump]
Technomage: El Retorno de la Eternidad (Spain) [Redump]
TechnoMage: En Quête de L'Eternité (France) [Redump]
TechnoMage: Return of Eternity (Europe) [Redump]
TechnoMage: Ritorno all'Eternitá (Italy) [Redump]

Redumped software list items
Alone in the Dark: One-Eyed Jack's Revenge (USA) [Redump]
Ultimate 8 Ball (USA) [Redump]

### DIFF
--- a/hash/psx.xml
+++ b/hash/psx.xml
@@ -919,19 +919,23 @@ The game freezes when loading (after the intro)
 		</part>
 	</software>
 
-	<software name="aitd2">
-		<!-- Original images (Redump)
+	<software name="aitd2_us" cloneof="aitd2">
+		<!-- 
+		http://redump.org/disc/3083
 		<rom name="Alone in the Dark - One-Eyed Jack's Revenge (USA).cue" size="115" crc="e7e7f0de" sha1="4e282f1c2fe8f2252ce18601611673feceb3fb64"/>
 		<rom name="Alone in the Dark - One-Eyed Jack's Revenge (USA).bin" size="715309056" crc="83f2a2d1" sha1="d928e147ca1ef265929e1c51d8f8ae39de2a24f9"/>
 		-->
-		<description>Alone in the Dark - One-Eyed Jack's Revenge (USA)</description>
+		<description>Alone in the Dark: One-Eyed Jack's Revenge (USA)</description>
 		<year>1996</year>
 		<publisher>Kokopeli Digital Studios</publisher>
+		<info name="barcode" value="7 52919 47002 2"/>
+		<info name="developer" value="I-Motion"/>
+		<info name="language" value="English"/>
 		<info name="serial" value="SLUS-00239"/>
 		<sharedfeat name="compatibility" value="NTSC-U"/>
 		<part name="cdrom" interface="cdrom">
 			<diskarea name="cdrom">
-				<disk name="alone in the dark - one-eyed jack's revenge (usa)" sha1="66da4ab3ad0f385d48f89fa8a0f4b6b174a40469"/>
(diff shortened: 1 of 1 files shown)


hash/psx.xml
```diff
@@ -62244,6 +62600,138 @@ During gameplay, the commentary speech audio is playing too fast
...
+		<info name="developer" value="Sunflowers"/>
+		<info name="language" value="Spanish"/>
+		<info name="serial" value="SLES-03244"/>
+		<sharedfeat name="compatibility" value="PAL-E"/>
+		<part name="cdrom" interface="cdrom">
+			<diskarea name="cdrom">
+				<disk name="Technomage - El Retorno de la Eternidad (Spain)" sha1="ec8de298b3308891ac3705ab010b721c686661e3"/>
+			</diskarea>
+		</part>
+	</software>
```
cuavas, line 62689: LibCrypt depends on data stored in subchannel Q.  IIRC at one point, redump had this stored in a weird format in separate files that chdman doesn’t support.  Do all the CD images you added that depend on subchannel data have it correctly stored in the CHDs?  Thy need to be marked as bad dumps if they don’t.

ArcadeShadow: I wasn't aware of that detail. Yes, redump keep this information in a separated SBI file format. Those CHDs don't have this subchannel included. I didn't find a way to add the sub into de CHD (maybe you can help).
If not, I will mark them as bad dump (and add a note) in next PR.

cuavas: Any dumps missing data like that need to be marked bad dumps and `supported="no"`.  They will obviously fail protection checks if they’re missing data.

cuavas: This brings up a question – are you actually testing things at all?  LibCrypt-protected software won’t work unless you create the CHD with raw 2448 bytes-per-sector data including the important stuff in subchannel Q.  But you’re marking stuff as supported when it was apparently created with 2352 bytes-per-sector data which requires the software using it to implicitly assume the default subchannel content.  That suggests you’re marking everything as supported without even testing whether known protection checks pass.

ArcadeShadow: First of all I apologize for my lack of knowledge about the PSX protection system.
I tested all those dumps, and they work without any error. That's why they aren't marked as not supported.
It’s one of two things: either the protection is already included in the dumps, or that protection is not being validated.

angelosa: That depends on the protection scheme used. 
LibCrypt was originally used with Final Fantasy VIII, which afaik just checks once, during bootup. Later games are more creative, like LMA Manager 2001 and 2002 will hamper the game to be unwinnable later on in more or less subtle ways.

PR #14045, opened 2025-08-11
vsystem: hook up rs232 to f1gp, gstriker and tail2nos.

SailorSat:
tail2nos and f1gp use an MC6850 with dual DE9 ports to link up to 6 units in a ring.
gstriker has only one DE9 port to link 2 units.

the exact clock of the MC6850 is unknown.
6 players fails at 19200 baud, but works with 38400 baud (and up).


SailorSat:
Test-Footage.

https://youtu.be/jnX99usnKg0?t=40
https://youtu.be/M-ijC6e__qE?t=40
https://youtu.be/c4DM6tpDw_Q?t=8

big thanks to Kale and startaq for their input.

src/mame/vsystem/f1gp.cpp
```diff
@@ -120,6 +124,8 @@ class f1gp_state : public driver_device
 	required_device<palette_device> m_palette;
 	optional_device<generic_latch_8_device> m_soundlatch; // not f1gpbl
 	required_device<acia6850_device> m_acia;
+	optional_device<rs232_port_device> m_rs232a;
+	optional_device<rs232_port_device> m_rs232b;
```
startaq: Does the bootleg also have the same linking hardware? Then those could be changed to `required_device` and hooked up there too.

src/mame/vsystem/f1gp.cpp
```diff
@@ -826,9 +840,19 @@ void f1gp_state::f1gp(machine_config &config)
...
 	ACIA6850(config, m_acia, 0);
 	m_acia->irq_handler().set_inputline("sub", M68K_IRQ_3);
-	m_acia->txd_handler().set("acia", FUNC(acia6850_device::write_rxd)); // loopback for now
+	m_acia->txd_handler().set("rs232a", FUNC(rs232_port_device::write_txd));
 
-	clock_device &acia_clock(CLOCK(config, "acia_clock", 1'000'000)); // guessed
+	// dual DE-9 ports
+	// COM-IN (inner) and COM-OUT (outer) according to manual
+	rs232_port_device &rs232a(RS232_PORT(config, "rs232a", default_rs232_devices, "loopback"));
+	rs232a.rxd_handler().set("rs232b", FUNC(rs232_port_device::write_txd));
```
startaq: Instead of `rs232a` and `rs232b` a better name might be `com_in` and `com_out` then? I assume `rs232a` would be the COM-OUT, why does it have an `rxd` handler? Isn't it only used to transmit data?

SailorSat: that rxd handler is for the default "loopback" slot option to work properly.
*EDIT* now that I think of it, the loopback is no longer needed, right?

src/mame/vsystem/f1gp.cpp
SailorSat: good call. it appears to have similar hardware.

mamehaze: the bootleg software probably still supports it, although I'd be quite surprised if they actually have the connectors on the board (but that could vary from bootleg to bootleg)

ajrhacker:
The one small problem is that the actual baud rate is likely not 19200 or 38400 or any other standard rate.

Someone with a PCB of one of these games needs to measure what frequency is present on MC6850 pins 3 & 4.

SailorSat:
someone on discord offered to trace those on a "super formula" board. has no ways of measuring them though.
waiting for their reply.

angelosa (approved):

SailorSat:
according to the person on discord, pin 3 and 4 are connected together and fed by pin 6 of a LS393.
that LS393 has 20MHz on pin 1 and GND on pin 2.

based on that and the datasheet of a LS393 the divider should be 16.
20MHz / 16 (LS393) / 16 (in the 6850) equals a real baudrate of 78125.

given some tolerances, I'd setup the rs232 defaults with 76800 baud.
the null_modem device (or more likely rs232 device) does not offer a 76800 baud option however.

Issue #14030, opened 2025-08-06
Commit 6b3ba498edb4856f14d0de872934a9c112c7e009 broke ensoniq esq5505 keyboards


cbrunschen:
### MAME version

0.279 (any version that includes commit 6b3ba498edb4856f14d0de872934a9c112c7e009)

### System information

Compiling and running on Ubuntu Linux 25.04, on a Framework 16 laptop with an AMD 7940HS x86-64 CPU. 

### INI configuration details

```INI

```

### Emulated system/software

vfx, vfxsd, sd1, sd132

### Incorrect behaviour

esq5505-based keyboards, specifically the vfx, vfxsd, sd1 and sd132 keyboards, do not generate sound in response to MIDI input. Also, some front-panel button presses also cause system errors, when they should not.

### Expected behaviour

esq5505-based keyboards, specifically the vfx, vfxsd, sd1 and sd132 keyboards, *should* generate sound in response to MIDI input, and front panel button presses should not cause system errors

### Steps to reproduce

1. Note the first bad commit, `6b3ba498edb4856f14d0de872934a9c112c7e009` per `git bisect`
2. `git checkout 6b3ba498edb4856f14d0de872934a9c112c7e009^` - the commit just before the first bad one, i.e., the last good commit,
3. Compile mame, including at least SOURCES=src/mame/ensoniq/esq5505.cpp: `make -j16 SOURCES=src/mame/ensoniq/esq5505.cpp`
4. Run mame, with MIDI input - in this example, from my Korg microKEY keyboard, but substute your own: `./mame -w -rp ~/mame/roms/ sd1 -http -midiin "microKEY microKEY _ CTRL"` (where `sd1` can be any of `vfx`, `vfxsd`, `sd1` or `sd132`)
5. Connect a web browser to [`http://localhost:8080/esqpanel/FrontPanel.html`](http://localhost:8080/esqpanel/FrontPanel.html) to see the front panel for the emulated keyboard
6. Once the keyboard has finished starting it should display a screen:
    ````
    INT0 OMNIVERSE   FLASH-BACK    SD1-PAD
         SQUARE-PAD  NU-MEANING   ASCENSION
    ````
10. send MIDI key events from the MIDI device to the emulated Ensoniq keyboard8. Note that sound is being generated!
[...]

PR #14046, opened 2025-08-12
Reset the transmit register when requested by witing CR with bit 3 set.

cbrunschen:
This fixes https://github.com/mamedev/mame/issues/14030 .


pmackinlay:
This is not the correct fix. Proper fix merged here 9b198e4b0a5a3ba87b603704f09b630376343d70

pmackinlay:
For clarity, product documentation is quite explicit that disabling the transmitter does not prevent the transmission of any data still in the transmit shift or holding registers. The actual bug was that TxRDY was being set after this data was sent despite the transmitter being in a "disable pending" state (and this was confusing the system firmware).

The referenced commit should prevent this from happening by tracking the transmitter enable/disable state properly.

cbrunschen:
Thank you so much!

PR #14048, opened 2025-08-12
adp/adp.cpp: add nvram to funny land

stonedDiscord:
Function 188dc in funlddlx4 will initialize the NVRAM and make the machine boot.

There's still issues with the inputs so it stays as MACHINE_NOT_WORKING


cuavas (commented):
Is this default NVRAM taken from a real machine, or handcrafted to look like what the program expects?  For the systems where it says the NVRAM content is “taken from parent”, should it have the `BAD_DUMP` flag until it’s verified that the same content is OK for the clones?

stonedDiscord:
It's crafted by running the initialization routine in the emulated system.
In my tests the clones generated the same NVRAM.
The difference between the clones seem to be the switch in currency from DM to €

cuavas:
Can you note that it’s just default initialisation the program generates for convenience, and not required factory initialisation content?

stonedDiscord:
Where should I put the comment? In the header or next to the ROM loads?

The RAMs are blank in the factory and get initialized using the service keyboard (adp/servicet.cpp) which ends up calling 188dc

angelosa:
> Where should I put the comment? In the header or next to the ROM loads?

Header please, given the alternative will repeat itself for every declaration.

PR #14049, opened 2025-08-12
Adding a layout for Ensoniq ESQ-1.

felipesanches:
- It can play notes by clicking on the keyboard.
- Volume and Data Entry sliders are hooked up via Lua scripting.
- Volume slider still non-functional. Data Entry slider works fine.
- SVG snippets were kept to the absolute minimum (drawings of envelope and oscillators diagrams)
- The full unit was drawn, but only the relevant part is visible on the default layout. We may later add other views such as "Full keyboard".

## Before

Screenshot From 2025-08-12 19-01-09

## After

Screenshot From 2025-08-12 18-53-10


src/mame/ensoniq/esq1.cpp
```diff
@@ -665,32 +693,151 @@ void esq1_state::sq80(machine_config &config)
...
+	PORT_BIT(0x80, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(esq1_state::key_stroke), 0xb3) PORT_NAME("ENV 3")
+
+	PORT_START("COL6")
+	PORT_BIT(0x01, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(esq1_state::key_stroke), 0xb4) PORT_NAME("DCA 4")
+	PORT_BIT(0x02, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(esq1_state::key_stroke), 0xb5) PORT_NAME("Modes")
+	PORT_BIT(0x04, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(esq1_state::key_stroke), 0xb6) PORT_NAME("Split")
+	PORT_BIT(0xf8, IP_ACTIVE_HIGH, IPT_UNUSED)
+
+	PORT_START("INTERNAL_KBD0")
+	PORT_BIT(0x001, IP_ACTIVE_HIGH, IPT_KEYBOARD) PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(esq1_state::internal_kbd_key_stroke), 0x98) PORT_NAME("C0")
```
galibert: Can you use the PORT_GM_* stuff for the key names?  It's not adding anything right now, true, but I hope it will in the future.

felipesanches: sure!

felipesanches: image

When I play on my USB-MIDI controler, MAME plays the ESQ1 sounds but the keys do not change color just like when I click on the keys on the internal layout.

One obvious reason for this is that there's nothing in the driver specifying that those clickable elements should trigger MIDI events (or visually react to receiving them).

The second reason is that these USB-MIDI events are probably being received by the ESQ1 cpu via the serial port assigned to the MIDI IN rear connector and these keys drawn on the layout are scanned from the internal keybed by an MCU (high-level emulated in this case).

I am not sure if there's any difference in behaviour in the ESQ1 between key presses from the internal keybed versus MIDI events from the rear connectors. But I am aware of such cases for other synths. For instance, the KN5000 has some self-test modes that can only be activated by holding specific combinations of keys in the internal keyboard.

So, I believe there should be a way for users to specify where to send the MIDI events when using an external MIDI controller. Does the UI already have a way to specify that? And how should we declare that on the driver? Or is this usecase still not supported?

galibert: Using (usb-)midi as an input device is not implemented yet.  PORT_GM_* is one of the steps in that direction though, which is why I try to have it comprehensive in the drivers, even if right now it's only setting the name.

Right now when you play on your m-wave, it goes to the esq1 midi in indeed.  It's not seen as a piano keyboard press.

felipesanches: **EDIT:** This message was written before I reloaded the page, so I had not seen @galibert's message sent 9 minutes earlier.

----

What I am describing is basically the need to let the user decide whether to have the external MIDI controller play the role of emulating "notes arriving via MIDI" versus "notes being played with fingers".

As these things are handled by different code-paths on the synth firmware, they sometimes can have special handling just like in the case of Technics KN5000 self-tests.

felipesanches: ok, cool!
I'll annotate it all with `PORT_GM_*`

felipesanches: When adding `PORT_GM_*` should I omit the `PORT_NAME` with the musical key name? Or should I keep both?

felipesanches:
The octaves are wrong! I'll reassign the midi note values to fit the correct range.

src/mame/ensoniq/esq1.cpp
galibert: It replaces the PORT_NAME

felipesanches: OK, I've just removed those redundant `PORT_NAME` entries now.

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,708 @@
...
+			d="m 0,9 l 3,5 3,-5
+			   m 48,-1 l 1.5,6 -5.5,-2
+			   M 30,3 h 14 l 23,23 h 10
+			   m 52,6 h -9 l -25,25 h -65
+			   m 156,-25 h 99
+			   m 0,-24 h -57 l -18,18 h -24
+			   m -57,0 h -9 l -25,-25 h -92 l 0,28 h 176" />
+		</svg>
+		]]></data></image>
+	</element>
```
cuavas: Please don’t use CSS unnecessarily when attributes will work.  Also, please indent stuff to show structure.  Something like:
```
	<element name="oscillators_diagram">
		<image><data><![CDATA[
			<svg width="285" height="58">
				<path fill="none" stroke="#fecb24" stroke-width="1"
						d="
							m 0,9 l 3,5 3,-5
							m 48,-1 l 1.5,6 -5.5,-2
							M 30,3 h 14 l 23,23 h 10
							m 52,6 h -9 l -25,25 h -65
							m 156,-25 h 99
							m 0,-24 h -57 l -18,18 h -24
							m -57,0 h -9 l -25,-25 h -92 l 0,28 h 176" />
			</svg>
		]]></data></image>
	</element>
```

and the same for other embedded fragments.

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,708 @@
...
+		<rect><color red="0.05" green="0.05" blue="0.05"/></rect>
+	</element>
+
+	<element name="invisible-rect">
+		<rect><color red="0" green="0" blue="0" alpha="0"/></rect>
+	</element>
+
+	<element name="black-rect">
+		<rect><color red="0.0" green="0.0" blue="0.0"/></rect>
+	</element>
```
cuavas: You should avoid making separate elements for things that are the same thing in different colours.  It’s more efficient to use a single white rectangle and colour it when instantiating it.

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,708 @@
...
+		<element ref="white_rect"><bounds x="0" y="65" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="0" y="74" width="9" height="1"/></element>
+
+		<element ref="white_rect"><bounds x="38" y="20" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="38" y="29" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="38" y="38" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="38" y="47" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="39" y="56" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="38" y="65" width="9" height="1"/></element>
+		<element ref="white_rect"><bounds x="38" y="74" width="9" height="1"/></element>
```
cuavas: This isn’t a good idea – each of these will be composited as an additional textured quad.  It’s a lot more efficient to make a single element to represent these tick marks so it’s composited as a single textured quad for all of them.  In fact, you should put as much of the slider as you can in a single element to reduce the amount of geometry you’re pushing around.

Remember MAME still thinks immediate mode is cool, so this overhead adds up quickly.

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,708 @@
...
+	<element name="midi_out_text"><text string="MIDI OUT"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+	<element name="pedal_cv_text"><text string="PEDAL/CV"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+	<element name="left_mono_rear_text"><text string="LEFT/MONO"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+	<element name="right_rear_text"><text string="RIGHT"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+	<element name="audio_out_text"><text string="AUDIO OUT"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+
+
+	<!-- TODO: SVG images for the logos
+		<element name="ensoniq_logo"><text string="ensoniq"><color red="0.83" green="0.86" blue="0.83" /></text></element>
+	-->
```
cuavas: You aren’t going to be able to do this with reasonably human-readable SVG.  It’s not going to go in an internal layout.  There’s no point having a TODO for it.

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,708 @@
...
+	<bounds left="0" right="1280" y="-100" height="640"/>
+	<element ref="dark_grey_rect"><bounds x="-460" y="-87" width="1777" height="85"/></element> <!-- rear -->
+	<element ref="panel_sheet"><bounds x="-460" y="0" width="1777" height="188"/></element> <!-- main portion of control panel -->
+	<element ref="dark_grey_rect"><bounds x="-460" y="190" width="1777" height="55"/></element> <!-- middle -->
+	<!-- <element ref="grey_rect"><bounds x="-227" y="257" width="1507" height="263"/></element> --> <!-- under the keybed -->
+	<element ref="panel_sheet"><bounds x="-460" y="257" width="297" height="246"/></element>  <!-- pitch & mod side-panel -->
+	<element ref="dark_grey_rect"><bounds x="-490" y="-95" width="27" height="620"/></element> <!-- left border -->
+	<element ref="dark_grey_rect"><bounds x="1320" y="-95" width="27" height="620"/></element> <!-- right border -->
+	<element ref="dark_grey_rect"><bounds x="-390" y="310" width="24" height="145"/></element> <!-- pitch wheel -->
+	<element ref="dark_grey_rect"><bounds x="-315" y="310" width="24" height="145"/></element> <!-- mod wheel -->
```
cuavas: Please use indentation to show structure.  The content of the `<view>` element should be indented.

src/mame/layout/esq1.lay
happppp: Trademark issue anyway if MAME adds it in the source, and claims CC0.

src/mame/layout/esq1.lay
happppp: Maintainability over efficiency here, I'd say.
```
<element ref="descriptive_rect"><bounds x="0" y="0" width="10" height="10" /></element>
<element ref="rect"><bounds x="0" y="0" width="10" height="10" /><color red="0" green="0" blue="0" /></element> <!-- repeated hardcoded color everywhere this color rect is used -->
```

src/mame/layout/esq1.lay
felipesanches: Cool. But could we leave the comment as a tip to external layout authors?

src/mame/layout/esq1.lay
felipesanches: yeah, my rationale was exactly what @happppp described here. The handling of colors in the .lay schema is quite insane. In the near future I'd like to propose improvements for a version 3 with better ways to author layouts and better color declarations are one of the things I would include.

src/mame/layout/esq1.lay
felipesanches: Would using `<repeat>` here help? Or would it still be the same issue?

> Remember MAME still thinks immediate mode is cool, so this overhead adds up quickly.

Maybe it is clear to you because you know about some internal implementation detail, but I fail to comprehend the meaning of this sentence. What is the thing you refer to when you use this "immediate mode" expression?

felipesanches: Let me take the opportunity to mention that I think sliders should be supported natively by the layout system. The current approach of building our own slider all the time by copy-pasting the same lua script everywhere we need it is a mess. It is very error prone and it is not a joy for layout authors.

This demonstrates that there's a need for the feature. And also that it would be useful to have a way to include external scripts and then we could maintain a small library of useful reusable things (like the slider script and also external assets such as SVG files - as described by @rb6502 and others at https://github.com/mamedev/mame/pull/12735#issuecomment-2590291368).

galibert: Believe us, we know.  It is, as is often the case, more complicated that it should be to add.

src/mame/layout/esq1.lay
felipesanches: done

src/mame/layout/esq1.lay
felipesanches: done

src/mame/layout/esq1.lay
felipesanches: @cuavas, what do you think?

src/mame/layout/esq1.lay
m1macrophage: Some time ago, I sent a proposal for how to share slider functionality (#13414), based on @galibert 's suggestion [here](https://github.com/mamedev/mame/pull/13177#discussion_r1937398263). At the time, it was deemed better to just copy-and-pate, since the proposal was not compatible with the future plans for layouts.

This is up to the MAME team. But if there is appetite to revisit the approach above, since the number of layouts with sliders has increased, or if there are suggestions for how to make the proposal "less incompatible" with the future, let me know.

src/mame/layout/esq1.lay
rb6502: The version of the Ensoniq wordmark from this era expired long ago.  Creative has kept a different Ensoniq wordmark active (last renewed 2016), but it's visually different and the last Ensoniq branded anything was 2002.  So I'm not worried about trademark here.

That said, the human-readable SVG thing is real, so this isn't a TODO.  I don't have an issue with it being a comment, just having it marked TODO, which has a specific meaning in MAME.

src/mame/layout/esq1.lay
rb6502: For reference, "immediate mode" means you generate draw commands and send them to the GPU naively as you go.  This is bad among other reasons because you're potentially incurring DMA setup and teardown time for each command instead of just having it all-at-once. The alternative is "retained mode" where the commands are put into a scene graph and the CPU does a pass to generate an optimized display list from the results.

felipesanches: > For reference, "immediate mode" means

Thanks for the clarification!

src/mame/layout/esq1.lay
felipesanches: Cool. The `TODO` is not there anymore. It was removed on an amended commit pushed here a couple days ago.

felipesanches: > The version of the Ensoniq wordmark from this era expired long ago. Creative has kept a different Ensoniq wordmark active (last renewed 2016), but it's visually different and the last Ensoniq branded anything was 2002. So I'm not worried about trademark here.
> 
> That said, the human-readable SVG thing is real, so this isn't a TODO. I don't have an issue with it being a comment, just having it marked TODO, which has a specific meaning in MAME.

Well... I think I can make it fairly readable:

```xml
<svg width="286" height="70">
	<g fill="#e2dbce" stroke="none">
		<!-- E -->
		<path d="
			M 0,0 v 20 h 67 v -20 Z
			m 0,25 v 20 h 67 v -20 z
			m 0,25 v 20 h 67 v -20 z"/>

		<!-- S -->
		<path d="
			M 76,0
			v  45 h  47 v  5 h -47 v  20 h 67
			v -45 h -47 v -5 h  47 v -20 Z" />

		<!-- Q -->
		<path d="
			M 152,0
			v 70 h 86 v -20 h -19 V 0 Z
			m 20,20 h 27 v 30 h -27 z" />

		<!-- 1 -->
		<path d="
			M 266,0
			l -12,20 h 12 v 50 h 20 V 0 Z" />
	</g>
	
	<!-- The triangle -->
	<path
		fill="#36dbe9" stroke="none"
		d="m 222,0 h 39 l -19.5,32.5 z"/>
</svg>
```

Screenshot From 2025-08-15 15-33-11

felipesanches:
There are a few questions that I left for @cuavas above and I may address them when I get an answer, but I consider that I have put my best effort here and that this PR is ready for a final review & merge.

src/mame/layout/esq1.lay
felipesanches: > This isn’t a good idea – each of these will be composited as an additional textured quad. It’s a lot more efficient to make a single element to represent these tick marks so it’s composited as a single textured quad for all of them. In fact, you should put as much of the slider as you can in a single element to reduce the amount of geometry you’re pushing around.
> 
> Remember MAME still thinks immediate mode is cool, so this overhead adds up quickly.

@cuavas, I understand that you are worried about hypothetical performance issues, but how would you measure it? Do you have suggestions for any profiling / performance measurement tool that could give us a concrete evidence of this specific snippet in the ESQ1 layout causing performance issues?

I have run the driver with the layout as described in the PR and I do not perceive (subjectively) any performance issue.

There are roughly a bit more than 300 elements in this layout. Why would these elements from the slider be more problematic than all the other ones?

Also, the layout in this PR makes the driver extremely more useful than it is in its current state at git master. In my opinion, its usefulness outweighs any hypothetical performance concerns. Especially when such performance issue seems to not even be measurable.

felipesanches:
I've just fixed the handling of the underline display command and updated the layout to render it properly. Finally we have visual feedback for the user indicating which is the currently selected item in the ESQ-1 VFD.

Here, the selected voice is "HARP" on internal instrument bank number 3:
image

And here the waveform setting ("SQUARE") is selected for oscillator number 2:
Screenshot From 2025-08-19 03-48-32

The underlines are also correctly rendered on the VFD-only view:
image

felipesanches:
Adopted suggestion by @happppp to use SVG to draw the up & down arrows instead of relying on the font having glyphs for these very specific unicode codepoints.  

# Before/After:

Screenshot From 2025-08-19 12-58-52
Screenshot From 2025-08-19 12-59-37

felipesanches:
Also improved the rendering of arrows in the oscillators diagram, by having them filled, as in the original panel artwork.

# Before/After:

Screenshot From 2025-08-19 12-58-33
Screenshot From 2025-08-19 12-57-46

src/mame/layout/esq1.lay
felipesanches: @cuavas, does https://github.com/mamedev/mame/pull/14049/commits/b511c53957f006715ed3ac1a3b10155ce04e2ef3 properly address your concern?

felipesanches:
Very minor tweak here. Better shapes for the arrows in the oscillators diagram.

# Before / After
Screenshot From 2025-08-21 19-46-22
Screenshot From 2025-08-21 19-41-57

felipesanches:
OK. I think I am done here. I'm pretty happy about the result.

The only thing not included is the Ensoniq logo, which is indeed not possible to make look good with simple/small SVG snippet.

Here's a comparison of the "Full unit" view and a photo of a real ESQ-1:
Screenshot From 2025-08-21 20-24-15

felipesanches:
The layout offers a few views:

# Default Layout
(which I consider to be the ideal balance between legibility and usability)
Screenshot From 2025-08-21 20-31-38

# Full unit
Screenshot From 2025-08-22 02-17-40

# VFD close-up
Screenshot From 2025-08-21 20-31-19

# Sequencer & Voice
Screenshot From 2025-08-21 20-31-28

felipesanches:
In my opinion, there is one known issue that I still do not know how to fix. To make the pitch wheel auto reposition itself to the center position (resetting its value to default but also animating the movement in the layout) is there a specific syntax for the INPUT_PORTS declaration on the driver that would achieve that? Or do I need to employ some other technique such as Lua scripting to implement such spring behavior animation?

src/mame/layout/esq1.lay
```diff
@@ -0,0 +1,895 @@
...
+		<element id="slider_~slider_id~" ref="invisible-rect">
+			<bounds x="12" y="10" width="23" height="77"/>
+		</element>
+		<element ref="slider_ticks"><bounds x="0" y="0" width="46" height="94"/></element>
+		<element id="slider_knob_~slider_id~" ref="slider-knob">
+			<animate inputtag="~port_name~" inputmask="0x7f"/>
+			<bounds state="100" x="9" y="11" width="29" height="16"/>
+			<bounds state="0" x="9" y="66" width="29" height="16"/>
+		</element>
+	</group>
```
cuavas, line 163: You don’t necessarily need to use SVG for the ticks – you could use a bunch of `<rect>` components in the element.

But can you combine the channel, rail and ticks into a single element?  You should be able to just stack those components within the element.  That will give one quad rather three to composite.

cuavas (commented):
This is looking quite a bit better than last time.

After what @happppp said, I’m still not sure about including logos in internal artwork.

src/mame/layout/esq1.lay
felipesanches: It was a bunch of `<rect>` components before. I changed it to a single element [per your request](https://github.com/mamedev/mame/pull/14049#discussion_r2273917026) (and I decided to use a single `<svg>` element to make it nice and clean).

I'm having a hard time understanding your request here. Isn't stacking a group of elements what a `<group>` does?

In cases like this one when it is not yet clear what a request means even after you requested it twice, it would be helpful if you could actually type a snippet of code as an example of what you want to see.

Also, your wording was "to reduce the amount of geometry you’re pushing around", but the only thing that is dragged is the slider knob. All the rest remains static.

Have you missed [this question](https://github.com/mamedev/mame/pull/14049#discussion_r2279750841) about ways to measure these hypothetical performance issues?

Finally, this style of slider implementation that I am using here is almost identical to the reference code I copied from (`src/mame/layout/oberheim_dmx.lay`). Does that implementation also have an issue to be addressed, then?

And have you seen [this question](https://github.com/mamedev/mame/pull/14049#issuecomment-3212443497) about how to animate spring behaviour of a mod wheel?

felipesanches:
> This is looking quite a bit better than last time.
> 
> After what @happppp said, I’m still not sure about including logos in internal artwork.

The svg of the Ensoniq logo that I hand optimized in this PR came from here https://commons.wikimedia.org/wiki/File:Ensoniq_logo.svg where it's legal status is described as:

> The depicted text is ineligible for [copyright](https://en.wikipedia.org/wiki/Copyrights) and therefore in the [public domain](https://en.wikipedia.org/wiki/public_domain) because it is not a “literary work” or other protected type in sense of the local copyright law. Facts, data, and unoriginal information which is common property without sufficiently creative authorship in a general typeface or basic handwriting, and simple geometric shapes are not protected by copyright.

And regarding trademark law, @rb6502 said (at https://github.com/mamedev/mame/pull/14049#discussion_r2279117011) that "The version of the Ensoniq wordmark from this era expired long ago."

happppp:
Here is the trademark: https://trademarks.justia.com/735/24/ensoniq-73524524.html
Status: 800 - Registered And Renewed
Owner: CREATIVE TECHNOLOGY LTD.

And yes, trademark != copyright. But that doesn't mean the trademarks are public domain, they're still intellectual property.

felipesanches:
> Here is the trademark: https://trademarks.justia.com/735/24/ensoniq-73524524.html Status: 800 - Registered And Renewed Owner: CREATIVE TECHNOLOGY LTD.
> 
> And yes, trademark != copyright. But that doesn't mean the trademarks are public domain, they're still intellectual property.

OK. Thanks for the clarification. I will remove the commit that included the Ensoniq logo.

What about the ESQ-1 logo? Is it legally safe to keep? Or should it also be removed?

happppp:
No idea, I'd say better be safe than sorry and don't copy-paste brand logos.

rb6502:
That's not the same version of the wordmark that the ESQ-1 used.  That specific version is listed as abandoned.  (Specifically I mean without the lines breaking it up where it's just a font - the Creative version with the lines was never actually used on a product).

rb6502:
To be precise, this is the one that's on my real hardware ESQ-1 as well as the photos on Wikipedia: https://tsdr.uspto.gov/#caseNumber=75222275&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch

And it's listed as cancelled and abandoned.

felipesanches:
I have just force-pushed after removing the commits with the brand logos.

The fully branded version is now hosted as an external layout at my personal repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork for anyone who may be interested in using it.

cuavas:
> I have just force-pushed after removing the commits with the brand logos.
> 
> The fully branded version is now hosted as an external layout at my personal repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork for anyone who may be interested in using it.

Thanks.  I’d prefer to err on the side of not doing things that could be construed as implying endorsement.

felipesanches:
The repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork is completely mine, with no endorsement from MAME devs. Sorry if that gave a bad impression. I am personally completely responsible for what I do in that repo.

cuavas:
> The repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork is completely mine, with no endorsement from MAME devs. Sorry if that gave a bad impression. I am personally completely responsible for what I do in that repo.

I was thanking you for removing it from the PR, that’s all.

felipesanches:
> The repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork is completely mine, with no endorsement from MAME devs. Sorry if that gave a bad impression. I am personally completely responsible for what I do in that repo.

I have also, in good faith, now renamed that personal repo to https://github.com/ArqueologiaDigital/external_artwork_for_emulators

I hope that's better.

cuavas:
> > The repo at https://github.com/ArqueologiaDigital/MAME_synths_artwork is completely mine, with no endorsement from MAME devs. Sorry if that gave a bad impression. I am personally completely responsible for what I do in that repo.
> 
> I have also, in good faith, now renamed that personal repo to https://github.com/ArqueologiaDigital/external_artwork_for_emulators
> 
> I hope that's better.

I wasn’t complaining about your repo – that’s between you, Microsoft (owner of GitHub) and Creative.  You’re reading too much into it :)

felipesanches:
Thanks for approving and merging the PR! ;-)

If possible, please let me know the answer to the question about animating spring movement on a pitch wheel [here](https://github.com/mamedev/mame/pull/14049#issuecomment-3212443497).

PR #14053, opened 2025-08-15
Fix regressions in apple2 video

arekkusu42:
This PR fixes two regressions which broke various beam-racing demos.

Examples of lo-res and text errors:
	*French Touch, CC65c02 (part 2, Homer sinewave ragged edges)
	*French Touch, CCII (part 2, HGR/GR ragged rasterbars)
	*Deater, Fancy lo-res interlace (completely broken)
	*Deater, Megademo (split screen letter typing, fireworks ending)

Examples of page 1/2 errors:
	*French Touch, CC65c02 (part 1, bouncing box too far left)
	*Deater, Megademo (split screen letter typing)

Here's one screenshot, before/after this PR, compared to a real Apple IIe:
DeaterMegaDemo_0279_before
DeaterMegaDemo_0279_after
DeaterMegaDemo_HW_IIe

While investigating, I also wrote a small unit test, attached here:
[VaporTest_250814.zip](https://github.com/user-attachments/files/21789974/VaporTest_250814.zip)

This test is very similar to Sather's "Little Text Window" but instead of
relying on $C019 to sync, it uses Lancaster's HGR vaporlock, and has an
interactive BASIC shell to shift the lock phase around and toggle between
Text/GR/HGR.  Note that this assumes an NTSC display, and it must be run
at 1 MHz (due to other bugs in MAME with the Zip chip status registers
and read_floatingbus() cycle counting, which I plan to address in future PRs.)

Before/after this PR, and a real Apple IIe:
VaporlockCenter_0279_before
VaporlockCenter_0279_after
VaporlockCenter_HW_IIe

This test also demonstrates more bugs in MAME (broken color-killer, some
video softswitch latching is off by one cycle, and glitch artifacts aren't
emulated); this PR only fixes the regressions from old behavior.

TODO: IIgs $C068 probably also needs to participate in scr_w()?  Need to
verify hardware behavior...

### DIFF
--- a/src/mame/apple/apple2video.cpp
+++ b/src/mame/apple/apple2video.cpp
@@ -531,7 +531,6 @@ template <a2_video_device::model Model, bool Invert, bool Flip>
 void a2_video_device::text_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect, int beginrow, int endrow)
 {
 	uint8_t const *const aux_page = m_aux_ptr ? m_aux_ptr : m_ram_ptr;
-
 	uint32_t const start_address = use_page_2() ? 0x0800 : 0x0400;
 
 	beginrow = (std::max)(beginrow, cliprect.top());
@@ -592,6 +591,8 @@ template void a2_video_device::text_update<a2_video_device::model::IVEL_ULTRA, t
 
 void a2_video_device::hgr_update(screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect, int beginrow, int endrow)
 {
+	unsigned const start_address = use_page_2() ? 0x4000 : 0x2000;
+
 	beginrow = (std::max)(beginrow, cliprect.top());
 	endrow = (std::min)(endrow, cliprect.bottom());
 	int const startcol = (cliprect.left() / 14);
@@ -599,8 +600,6 @@ void a2_video_device::hgr_update(screen_device &screen, bitmap_ind16 &bitmap, co
 
 	// B&W/Green/Amber monitor, CEC mono HGR mode, or IIgs $C021 monochrome HGR
 	bool const monochrome = monochrome_monitor() || m_monohgr || (m_monochrome & 0x80);
-
-	unsigned const start_address = use_page_2() ? 0x4000 : 0x2000;
 
 	// verified on h/w: setting dhires w/o 80col emulates a rev. 0 Apple ][ with no orange/blue
 	uint8_t const bit7_mask = m_dhires ? 0 : 0x80;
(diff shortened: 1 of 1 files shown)


rb6502:
Thanks!  Once CI passes this is a no-brainer apply.  I actually wasn't aware that the Deater megademo had regressed that much on the printer typing screen, so it's good to see that nailed down.  I really appreciate the unit test as well.  That's so much nicer than running through these demos and waiting for the problem to happen.

Do cycle-counting vapor lock demos work on hardware with a Zip Chip at > 1 MHz speeds?  Obviously any floating bus issues with the Zip should be fixed but I'd be surprised.

$C068 likely does need scr_w() but IIgs raster timing is a known horror show at the moment.  When I wrote the current IIgs driver, I didn't understand that the actual raster structure is generated by the Mega II and therefore identical to the IIe.  Just the pixel clock and how many pixels per 1 MHz cycle the VGC outputs changes.  I've started rewriting how that all works but it's going to be a minute.

arekkusu42:
Thanks for the review & merge!

I'm not aware of any software that does it, but on the IIgs it
is at least feasible do beam-racing at 2.8 MHz, since there is
a common master clock divisor.  On the other hand, because the
Zip chip XTAL isn't a multiple of the 1 MHz clock, beam-racing
can never be sensible there.  However, I would very much
like to write unit tests which:
	1) draw complicated stuff (as fast as possible)
	2) query the current Zip state
	3) disable the Zip
	4) vaporlock, show hard-to-emulate patterns
	5) restore the old Zip state, goto 1)
Currently, 2) isn't possible in MAME due to incomplete $C05B emulation.

Additionally, read_floatingbus() using m_maincpu->total_cycles()
must be switched to the screen's hpos/vpos; the CRT shouldn't
magically scan faster when the CPU clock changes. ^_^;

I have WIP PRs for both of those, so my local build works well
(on apple2p/e/ee/c/cp, gs is another story...) but:
A) I need more hardware validation with a real Zip (my IIgs
is waiting on a PSU rebuild, and I need to locate a volunteer
with a IIc+), and
B) there are several timing interactions I need to detangle;
$C019, vblank/hpos/vpos are all subtly broken due to set_raw()
placing the HBL to the right of the video.  I have not yet
grokked the intended usage pattern, need to study other drivers.
And writing unit tests for the IOU scanner is time consuming...

With apple2gs, the first problem with raster timing is that there
are currently two different 1 MHz clocks defined. ^_^;  Fixing
that (and the above) should make it stable.

rb6502:
For set_raw(), look at amstrad.cpp, which does properly place hbend to the left of the active area.  The side effect is that you'll have to change screen_update() to draw at a different horizontal position, but that's relatively easy.

The same thing is possible with vbend, so we could just make vpos the actual Woz counter value on the 8-bits, which definitely is not zero-based.

arekkusu42:
Thanks for the tip.  That's the path I had started down, but it seemed silly to re-add the same constant offset in every partial update, instead of fixing the bitmap's m_base once during alloc/resize.  I'll keep poking at it.

PR #14062, opened 2025-08-17
pc/at.cpp: get rid of i386 macro

angelosa:


angelosa:
Unless I'm missing something there's no need for this macro at all, checking with CI aid as courtesy.

ajrhacker:
`#undef i386` seems to have been added to the old MESS driver in 2008. Even a cursory look at https://github.com/mamedev/mess-svn/blob/master/src/emu/cpu/i386/i386.h shows how it might have been useful back then. (The associated "legacy" macros, long since purged from the MAME source, were defined in https://github.com/mamedev/mess-svn/blob/master/src/emu/devcpu.h.)

PR #14065, opened 2025-08-17
ti/ti89.cpp: no sound hardware

stonedDiscord:
The 3.5mm TRS jack is the serial link port not a headphone out.


cuavas:
FWIW it’s a 2.5mm socket, not 3.5mm, and various games actually do use it for crude sound output.  MAME supports this abuse with a “slot card” option.

PR #14068, opened 2025-08-18
Extension bus for Technics KN5000 and its HD-AE5000 optional board.

felipesanches:
HD-AE5000 adds a hard-drive, a parallel port (for performing backups on a PC) and an additional audio output.

This initial commit declares the extension interface and the HD-AE5000 ROMs.

Additional work still needs to be done after more research to trace the signals on the HD-AE5000 PCB.

Screenshot From 2025-08-18 00-07-42

Screenshot From 2025-08-18 00-08-20


scripts/src/bus.lua
```diff
@@ -5520,6 +5520,20 @@ if (BUSES["TMC600"]~=null) then
 	}
 end
 
+---------------------------------------------------
+--
+--@src/devices/bus/technics/kn5000_extension.h,BUSES["TECHNICS"] = true
+---------------------------------------------------
+
+if (BUSES["TECHNICS"]~=null) then
```
rb6502: I'd call this bus something like "kn5000" unless all Technics keyboards used it.

rb6502: I do see the actual files are named how I'd expect.  Maybe a structure like bus/technics/kn5000/kn5000_extension.cpp so it's clear in the future what's going on.

felipesanches: I have just pushed a commit with the changes you proposed, but I think I did something wrong, not sure exactly what:

Screenshot From 2025-08-21 22-34-11

felipesanches: It looks like something went bad on my local build. I am doing make clean and then a new build from scratch. I suspect everything is actually fine with the commit.

felipesanches: @rb6502, yes, I can confirm everything is fine. Your suggested changes are implemented and in good shape.

felipesanches:
Please merge this.

PR #14070, opened 2025-08-19
new NOT WORKING machines (Sega 119)

mamehaze:
new NOT WORKING machines
----
119 [Hammy, David Haywood]

### DIFF
--- a/src/mame/galaxian/119.cpp
+++ b/src/mame/galaxian/119.cpp
@@ -198,30 +198,19 @@ static INPUT_PORTS_START( sega119 )
 	PORT_DIPSETTING(    0x00, DEF_STR( On ) )
 INPUT_PORTS_END
 
-static const gfx_layout charlayout =
-{
-	8,8,
-	RGN_FRAC(1,3),
-	3,
-	{ RGN_FRAC(0,3), RGN_FRAC(1,3), RGN_FRAC(2,3) },
-	{ STEP8(0,1) },
-	{ STEP8(0,8) },
-	8*8
-};
-
 static const gfx_layout spritelayout =
 {
 	16,16,
 	RGN_FRAC(1,3),
 	3,
-	{ RGN_FRAC(0,3), RGN_FRAC(1,3), RGN_FRAC(2,3) },
+	{ RGN_FRAC(2,3), RGN_FRAC(1,3), RGN_FRAC(0,3) },
 	{ STEP8(0,1), STEP8(8*8,1) },
 	{ STEP8(0,8), STEP8(16*8,8) },
 	16*16
 };
 
 static GFXDECODE_START( gfx_sega119 )
-	GFXDECODE_SCALE( "tiles", 0, charlayout, 0, 4, GALAXIAN_XSCALE, 1)
+	GFXDECODE_SCALE( "tiles", 0, gfx_8x8x3_planar, 0, 4, GALAXIAN_XSCALE, 1)
 	GFXDECODE_SCALE( "tiles", 0, spritelayout, 0, 4, GALAXIAN_XSCALE, 1)
 GFXDECODE_END
 
@@ -262,12 +251,12 @@ void sega119_state::extend_tile_info(u16 *code, u8 *color, u8 attrib, u8 x, u8 y
 void sega119_state::sega119(machine_config &config)
 {
 	// basic machine hardware
-	Z80(config, m_maincpu, GALAXIAN_PIXEL_CLOCK/3/2);
+	Z80(config, m_maincpu, GALAXIAN_PIXEL_CLOCK / 3 / 2);
 	m_maincpu->set_addrmap(AS_PROGRAM, &sega119_state::prg_map);
-	// nmi is unused (just returns) timing is done by polling vblank
+	// nmi is unused (just returns), timing is done by polling vblank
 
 	Z80(config, m_audiocpu, 8_MHz_XTAL/2);
-	m_audiocpu->set_periodic_int(FUNC(sega119_state::sound_nmi), attotime::from_hz(8_MHz_XTAL/0x800));
+	m_audiocpu->set_periodic_int(FUNC(sega119_state::sound_nmi), attotime::from_hz(8_MHz_XTAL / 0x800));
 	m_audiocpu->set_addrmap(AS_PROGRAM, &sega119_state::sound_map);
 	m_audiocpu->set_addrmap(AS_IO, &sega119_state::sound_io_map);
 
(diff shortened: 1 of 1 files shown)


src/mame/sega/119.cpp
```diff
@@ -0,0 +1,287 @@
...
+	ROM_REGION( 0x6000, "samples", 0 ) // what for? processed in SW or a playback chip?
+	ROM_LOAD( "119_9.bin",   0x0000, 0x2000, BAD_DUMP CRC(b917e2c2) SHA1(8acd598b898204e18a4cfccc40720d149f401b42) ) //  FIXED BITS (xxxx1xxx)
+
+	// colour PROMs?
+ROM_END
+
+} // anonymous namespace
+
+// ROT180 is unusual, but all tiles are flipped in ROM
+GAME( 1986, sega119, 0, sega119, sega119, sega119_state, empty_init, ROT180, "Sega / Coreland", "119", MACHINE_NOT_WORKING )
```
angelosa: This looks a "small" release for '86 standards, so probably intended to be installed in cocktail cabinets (where ROT180 would make sense).

mamehaze:
does anybody feel like finishing this off? (@happppp , @Osso13 ?) I haven't looked at the sound, and I'm not usually the best person to be hooking up sound anyway.

mamehaze:
maybe just a prototype, before protection was added, rather than a bootleg?  afaik PCB picture is the same as the one that's been floating around before.

not sure why bootleggers would copy the rest of the protection circuitry when it seems to be otherwise a unique board, if they were making their own board they'd just leave it out entirely?

happppp:
It looks like a European bootleg to me. SGS Thomson EPROMS, RAM chips from West Germany, and I see some from Portugal.
Most chips are from 1985, some 1986. A prototype would more likely be from 1983, if the assumption of the game having been created in 1983 is true.

mamehaze:
seems quite illogical to me, could have been it was produced under license in Italy like some of the Sidam boards, which also look very like bootlegs, but have original and official builds of the games on them.

lack of Sega markings I'd say because it's an Alpha board design, Sega don't typically have proper ROM labels on protos either.

mamehaze:
that said, a lot of recent prototypes that have surfaced have come from boards found in Italy / Germany etc. and sometimes those were more obvious bootlegs / conversions, so it is possible, it's just this being a unique/original PCB design rather than something on an existing platform that makes it seem less likely to me in this case.

mamehaze:
the Alpha MCU protected games do upload a program for the MCU (which is how they used to be emulated as a virtual processor, as the MCU implements its own virtual machine for the bytecode) so I guess seeing if there's any patched out code to upload a program (and a program that should be uploaded) would be one way to maybe be more certain.

happppp:
That's not the part I'm arguing against. Yes it's possible that the Alpha Denshi MCU is unpopulated on the genuine PCB as well, and that the ROMs were unmodified, therefore from MAME's POV not a bootleg.
Hence I've put (bootleg?) with the questionmark.

happppp:
Original PCB:
(can see SEGA with some kind of serial label, and "1983(?) made in Japan" and SEGA under that too)


Hammy PCB:

Hammy1986:
It's a bootleg, but looks to be a "perfect copy" from hardware point of view, and graphics. hopefully the program is also unmodified.

happppp:
I can confirm the "sometimes killed by something invisible?" bug, very rare.
I don't think anything's wrong with tile banking, these tiles (steel beams that can be climbed on) are also used on level 4 and look more more pretty there.
Screenshot (788)

mamehaze:
The 'killed by something invisible' thing happens quite consistently near the top of the first level, it doesn't feel like something that would have been left in the game (but if it never actually got released, all bets are off)

mamehaze:
btw, do you have any opinion on the sample ROM? it always read for Hammy with the 'fixed bits' problem, but he observed no problems on the pins.  could it just be 7-bit data?

mamehaze:
[aaa2.zip](https://github.com/user-attachments/files/21869185/aaa2.zip)
here's one inp where it happens (not the usual place I see it though)

mamehaze:
[moreinps.zip](https://github.com/user-attachments/files/21870153/moreinps.zip)

2 more in as many plays.  weirdly earlier it was making me fall at just above the ladder after this one.

either way it happens a lot on the first level (but I've not seen it on levels after the first one)

it's definitely not very rare though, it happens once in about every 3 games.

happppp:
Yeah, that's where I've seen it happen, upper-right of 1st stage. I'm not strict enough to consider the game not working due to that bug but I'll leave the choice to you.
The DAC, looks like 8-bit to me? It's more likely that the EPROM d3 pin has gone bad.

Hammy1986:
I will send the full pcb set to porchy to dump and double verify everything on his equipment - the other prom looks a bit suspect too...

RE: the death bug, was there any protection removed ?? maybe they missed it....
OR maybe it's a enemy / falling object sprite that is not displaying?

mamehaze:
If there's a good chance it's a prototype that might contain such bugs I'm willing to consider it working.

If it's a released game it seems too serious.

happppp:
Probably limited release, not prototype.
It's on Sega's history page (archived, I can't find it on their current website): https://web.archive.org/web/20181231195247/https://sega-interactive.co.jp/special/history/title/119.html

Sprite not displaying: I checked if it has extra sprites, but no it looks like it has 1 sprite set (8 sprites as default Galaxian hw)

mamehaze:
They do have that history page, but it seems to just use the same screenshots as everybody else.  Interestingly it states the game uses a 4-way stick, but it's clearly programmed to use diagonals to fire shots at an angle (which you have to use in places) so I doubt they actually know all that much about it themselves.

happppp:
It'd need to be tested on PCB (practice on MAME 1st). Cold boot or 1st life is not required, it can happen during any session.
None of the DIP switches appear to have effect on the bug. Forcing 4-way joystick does not have effect on it either.

mamehaze:
https://www.youtube.com/watch?v=8CqbTCtoyh0  (See around 6:18) not limited to 1st level either, happens on the 2nd when recording this.  It's surprisingly frequent.  In that case there was a bullet on the line earlier, and it does seem to happen more when bullets have been active (maybe the stray bullet sprites you get on the right of the 1st level are related somehow?)  Maybe bullet behavior is different to standard Galaxian?

I did have, during another session, a bug where some unwanted graphics got drawn on the status bar too, which makes me wonder if there's a bug in the game that corrupts memory.  The positions you can fall in seem pretty standard though, always one of a few locations.

Maybe this is why the game didn't get released in 1983, was forgotten about, went to testing again in 86, and maybe didn't even really see much of a release then?

Might be worth testing if it happens with flipscreen 'on' and ROT180 (the tiles are upside down in ROM) but that would be weird.

happppp:
None of the DIP switches I tried have effect on the bug, that includes flipscreen and the unknown ones.

mamehaze:
Ah yeah, thought you just meant the unknown ones, but yeah, while we have seen flipscreen bugs in games before, not so much on games that are designed for cocktail in the first place.

mamehaze:
Anyway, maybe just promote it.  I'm really starting to think it's just a bug in the game, which explains why it ended up so rare.

happppp:
Sure https://github.com/mamedev/mame/commit/3db576bdc991ed28cfe3ff6925bbfca99ab73884

PR #14071, opened 2025-08-19
Namco System 23: Clear all machine-class members on reset

MooglyGuy:
Hopefully this will fix the occasional no-polygons issue that some MAME developers have reported.

### DIFF
--- a/src/mame/namco/namcos23.cpp
+++ b/src/mame/namco/namcos23.cpp
@@ -6092,24 +6092,35 @@ void crszone_state::machine_start()
 
 void namcos23_state::machine_reset()
 {
-	m_c435.buffer_pos = 0;
-	m_subcpu->set_input_line(INPUT_LINE_RESET, ASSERT_LINE);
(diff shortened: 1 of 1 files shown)


ajrhacker:
I can confirm that this restores 3D to System 23, though some major graphical glitches remain in attract mode. The actual hardware probably only resets selected things, so I've also experimented with more selective clearing. The things that cause obvious problems when not zeroed out initially include `m_c435.pio_mode` (basically all polygons) and, more bafflingly, `m_c421.dram_a` ("Powered by NAMCO®" and other elements).

MooglyGuy:
I would encourage you to just merge the change if it improves things for you.

It's not like we're dealing with some Galaxian board where anything and everything can be easily verified. It's a late-90's 3D platform where pretty much every change involves firing a shotgun in the dark and hoping that it doesn't hit any existing, working, assumptions.

Just take the W that the improved clearing works for you and don't look too hard at it.

PR #14080, opened 2025-08-22
goldnpkr.cpp: jkrbonus is by Seta

soundsnow:
https://emulation.gametechwiki.com/index.php/GCA3A#Seta_.5BJP.5D_.281981.2F2002.29


happppp:
Any proof for this other than that wiki? https://emulation.gametechwiki.com/index.php?title=GCA3A&type=revision&diff=124641&oldid=124640 (this was you?)

soundsnow:
No. I can't find more information about Seta's early history on the Internet. Maybe only the PCB is from Seta or Allumer. The publisher should not be Seta. Seta did not publish games under its own name at the time. According to this wiki, Hana Awase 6 should already be in the Seta folder in MAME.

happppp:
Yeah, I don't think 1 person saying "is that you, seta?" and adding it to a wiki is enough for MAME to blindly take over that assumption.

angelosa:
There's just one thing less trustworthy than EmuGen wiki, and that's GCA3A on EmuGen wiki.

cuavas:
The name is generic enough that even if Seta made a game by that name, this may not be it.  We need more evidence that this was a Seta product than one of the most notoriously unreliable wikis.

PR #14083, opened 2025-08-24
namcos23: Add Race On! and Angler King Controls [sjy96525]

sjy555525:


sjy555525:
I tested in "I/O TEST" mode, they are working, but Angler King Start button doesn't Work.

Race ON!
Push View Change button


Push Handle Analog Dec button


Push Handle Analog Inc button


Push Gas Analog button


Push Brake Analog button



Angler King
Push Arm Yaw Analog Dec button


Push Arm Yaw Analog Inc button


Push Arm Pitch Analog Inc button


Push Rod Yaw Analog Dec button


Push Rod Yaw Analog Inc button


Push Rod Pitch Analog Inc button


Push Reel Analog Inc button

src/mame/namco/namcos23.cpp
```diff
@@ -6670,6 +6670,26 @@ INPUT_PORTS_END
...
 static INPUT_PORTS_START(aking)
 	PORT_INCLUDE(s23)
+
+	PORT_MODIFY("JVS_PLAYER1")
+	PORT_BIT(0x00000020, IP_ACTIVE_HIGH, IPT_JOYSTICK_UP) PORT_NAME("Up Select")
+	PORT_BIT(0x00000010, IP_ACTIVE_HIGH, IPT_JOYSTICK_DOWN) PORT_NAME("Down Select")
+	PORT_BIT(0x00000002, IP_ACTIVE_HIGH, IPT_BUTTON1) PORT_NAME("Enter") // BUTTON1
+
+	PORT_MODIFY("JVS_ANALOG_INPUT1")
+	PORT_BIT(0xffff, 0x8000, IPT_PADDLE) PORT_MINMAX(0x0100, 0xffff) PORT_SENSITIVITY(100) PORT_KEYDELTA(2560) PORT_NAME("Arm Yaw") PORT_REVERSE
```
angelosa, line 6680: 2560 is way beyond the max limit for `PORT_KEYDELTA` (255)

happppp: PORT_KEYDELTA is an s32, AFAIK it should work fine and this 2560 keydelta is already used a lot in namcos23 (not sure if this is a carefully chosen value or just taken out of a hat and copy pasted)

angelosa: Well it's surely clamped in UI `Analog Controls`

happppp: Works fine for me in UI? eg. it's 2560 in rapidrvr and it doesn't clamp to 255 if I change it.
edit: max val in UI is: std::max(defvalue * 4, 255) - defvalue in this case is the x value in PORT_KEYDELTA(x)

Hydreigon223:
I myself was reluctant to add this at some point after doing tests with Race On and Angler King's analog input ranges and was stumped in figuring out what port the start button is mapped to. Is that button assigned to a non-standard JVS bit?

ghost:
These input changes are fine. For Race On they match my real PCB. In the cabinet the pedals and other pots will have a hard stroke limit on them due to physical pedals and wheel not having the 100% motion of the pot because doing that would break the pot in 2 pieces when force is applied to the control. The full range of the real controls is to where the test shows OK.
For Angler King the IO PCB is the common ASCA4 or ASCA5. If controls are missing it will be one of the inputs on the IO chip that isn't used on other games that remains unmapped.

PR #14084, opened 2025-08-24
sc3000_cart.xml: New software item added, metadata cleanup

FoxhackDN:
New working software list item:
BASIC Level III (Japan, Version 1.1) [bsittler]

I also fixed the country names to be like other software lists, and added information Basic Level III's export variants.

### DIFF
--- a/hash/sc3000_cart.xml
+++ b/hash/sc3000_cart.xml
@@ -6,8 +6,22 @@ license:CC0-1.0
 
 <softwarelist name="sc3000_cart" description="Sega SC-3000 cartridges">
 
-	<software name="basic3">
-		<description>BASIC Level III (Jpn)</description>
+	<software name="basic311">
+		<description>BASIC Level III (Japan, Version 1.1)</description>
+		<year>1985</year>
+		<publisher>Sega</publisher>
+		<part name="cart" interface="sg1000_cart">
+			<feature name="slot" value="level3" />
+			<dataarea name="rom" size="32768">
+				<rom name="basic sk iii (japan).sg" size="32768" crc="55f47493" sha1="737a06b3ada62942514a2facfea3fb1b0ca9c346" offset="000000" />
+			</dataarea>
+			<dataarea name="ram" size="32768">
+			</dataarea>
+		</part>
+	</software>
+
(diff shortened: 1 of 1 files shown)


cuavas:
If you’re cleaning up metadata, please also use lowercase for descriptive text when it isn’t proper nouns, initialisms or literal quotes (e.g. “Japan, alt” and “version 1.1”).

On the topic of the addition itself, does the parent need a version suffix, or could it be `basic3`?  Would the older version be clearer if it was `basic3v10` or `basic3_10` to provide a clearer division between the numbers?

FoxhackDN:
> If you’re cleaning up metadata, please also use lowercase for descriptive text when it isn’t proper nouns, initialisms or literal quotes (e.g. “Japan, alt” and “version 1.1”).

Allrighty, I'll fix that up in a bit.

> On the topic of the addition itself, does the parent need a version suffix, or could it be `basic3`? Would the older version be clearer if it was `basic3v10` or `basic3_10` to provide a clearer division between the numbers?

Either is fine! I added the version to the parent to make the difference clearer, but I don't mind changing them again,

cuavas:
> > On the topic of the addition itself, does the parent need a version suffix, or could it be `basic3`? Would the older version be clearer if it was `basic3v10` or `basic3_10` to provide a clearer division between the numbers?
>
> Either is fine! I added the version to the parent to make the difference clearer, but I don't mind changing them again,

In a lot of cases, we go with “no suffix is parent/default”, but I still wanted to check whether there’s a good reason to include the version suffix on the parent with in this particular case,

FoxhackDN:
> > > On the topic of the addition itself, does the parent need a version suffix, or could it be `basic3`? Would the older version be clearer if it was `basic3v10` or `basic3_10` to provide a clearer division between the numbers?
> > 
> > 
> > Either is fine! I added the version to the parent to make the difference clearer, but I don't mind changing them again,
> 
> In a lot of cases, we go with “no suffix is parent/default”, but I still wanted to check whether there’s a good reason to include the version suffix on the parent with in this particular case,

I gotcha. No worries. I'll revert the set name.

FoxhackDN:
Whoops, sorry, forgot to fix the clones...

PR #14087, opened 2025-08-24
harddriv_m.cpp: Fixed "BAD POLY BUFF ERROR" in hdrivairp 

Madcadden:
If you run the "DS IV GRAPHICS PROGRAM AND DATA RAM TESTED BY 68010" in the hdrivairp test menu the 68010 seeds the ADSP DM 0x0000–0x1FFE with 0x5555 before writing the sync address. This code simulates the test behaviour so that the game can load normally without the "bad poly buff error"


happppp:
I see, you moved it to reset(), though it's still unlikely that the real hw replicates this test in hardware, so best discard that idea.

What is plausible though is that the hw boots with that RAM filled with non-0, like so: https://github.com/mamedev/mame/commit/213e325818dbc7ddcdb44362e75a6a6621b0c53b
I only added it to hdrivairp, since I'm not sure it'd apply to all games on this driver.

Madcadden:
Hi @happppp thanks for getting the code into the appropriate place and looks neater too.

I was wondering about something that i noticed coming back to the game on the newer version of MAME. I used to overclock some of the CPU's so the game would run at a higher frame rate and it didn't effect the sound negatively but on the newer version i notice if i apply the same overclocks it makes the sound of the engine cut in and out every couple of seconds like it's stuttering which is kind of annoying. Has the code in MAME for the sound processing changed since last year?

happppp:
I don't think the driver slowed down the last year, still, make sure to check it's running at 100% (press F11, see indicator at top-right).
And yes, there's a known recent sound stuttering bug in MAME, but it doesn't sound like what you described: https://mametesters.org/view.php?id=9203

PR #14092, opened 2025-08-26
philips/patgen.cpp: Fix V38 PAL size

stonedDiscord:
Made a mistake because I typed them out manually and assumed it was the same size as the other v38


angelosa:
While at it, could you please be consistent with these pal loaders? I see `pm5644l00` that doesn't separate by `0x2000` stride.

PR #14091, opened 2025-08-26
apple2_flop_orig.xml, apple2_flop_clcracked.xml added 39 working, 1 non-working, 3 removed, 5 promoted to working

A-Noid33:
Meta data cleanups focusing on The Learning Company

New working software list items (apple2_flop_orig.xml)
-------------------------------
Addition Magician [4am, A-Noid]
Bumble Games (version 1.3) [4am, A-Noid]
Bumble Games (version 1.4) [4am, A-Noid]
Bumble Plot (The Learning Company) (version 1.0) [4am, A-Noid]
Bumble Plot (The Learning Company) (version 1.1) [4am, A-Noid]
Bumble Plot (The Learning Company) (version 1.2) [4am, A-Noid]
Fat Jack [4am, A-Noid]
Gertrude's Puzzles (version 1.0) [4am, A-Noid]
Gertrude's Puzzles (version 1.1) [4am, A-Noid]
Gertrude's Puzzles (version 1.2) [4am, A-Noid]
Gertrude's Secrets (version 1.0) [4am, A-Noid]
Gertrude's Secrets (version 1.1) [4am, A-Noid]
Gertrude's Secrets (version 1.2) [4am, A-Noid]
Gertrude's Secrets (version 1.3) [4am, A-Noid]
Juggles' Rainbow (version 1.0) [4am, A-Noid]
Juggles' Rainbow (version 1.4) [4am, A-Noid]
Logic Tools (prototype) [4am, A-Noid]
Magic Spells (version 1.1) [4am, A-Noid]
Magic Spells (version 2.1x) [4am, A-Noid]
Magic Spells (version 2.2) (800K 3.5") [4am, A-Noid]
Math Rabbit (version 1.0) [4am, A-Noid]
Math Rabbit (version 1.1) [4am, A-Noid]
Moptown Hotel (version 1.2) [4am, A-Noid]
Moptown Hotel (version 1.3) [4am, A-Noid]
Moptown Parade (version 1.0) [4am, A-Noid]
Number Stumper [4am, A-Noid]
Reader Rabbit (version 1.0) [4am, A-Noid]
Reader Rabbit (version 1.1) [4am, A-Noid]
Reader Rabbit (version 1.3) [4am, A-Noid]
Reader Rabbit (version 1.4) [4am, A-Noid]
Robot Odyssey I (version 1.0) [4am, A-Noid]
Robot Odyssey I (version 2.0) [4am, A-Noid]
[...]

PR #14094, opened 2025-08-27
imagedev/floppy.cpp: Allow images to be write protected from software list

ajrhacker:
Since a feature like this seems to be commonly requested, and some of the new software in #14091 demands it, I'm showing how simple it is to implement. The intended usage is:

> `<feature name="write_protected" value="true" />`

to protect one disk image, or

> `<sharedfeat name="write_protected" value="true" />`

to protect all disk images. If other devs want to use a different name or value for this, please comment.

### DIFF
--- a/src/devices/imagedev/floppy.cpp
+++ b/src/devices/imagedev/floppy.cpp
@@ -672,6 +672,11 @@ std::pair<std::error_condition, std::string> floppy_image_device::call_load()
 		m_image.reset();
 		return std::make_pair(image_error::INVALIDIMAGE, "Incompatible image file format or corrupted data");
 	}
+
+	const char *wp = get_feature("write_protected");
+	if(wp && !std::strcmp(wp, "true"))
+		make_readonly();
+
 	m_output_format = is_readonly() ? nullptr : best_format;
 
 	m_image_dirty = false;
--- a/src/devices/imagedev/floppy.cpp
+++ b/src/devices/imagedev/floppy.cpp
@@ -656,9 +656,9 @@ std::pair<std::error_condition, std::string> floppy_image_device::call_load()
 	const floppy_image_format_t *best_format = nullptr;
 	for (const floppy_image_format_t *format : m_fif_list) {
 		int score = format->identify(*io, m_form_factor, m_variants);
-		if(score && format->extension_matches(filename()))
+		if (score && format->extension_matches(filename()))
 			score |= floppy_image_format_t::FIFID_EXT;
-		if(score > best) {
+		if (score > best) {
 			best = score;
 			best_format = format;
 		}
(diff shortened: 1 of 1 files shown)


A-Noid33:
@ajrhacker Thanks for the code changes.  Wouldn't we need to update the XML schema "softwarelists.dtd?   For disks (hard drive images), they used writeable.

			<!ELEMENT dataarea (rom*)>
				<!ATTLIST dataarea name CDATA #REQUIRED>
				<!ATTLIST dataarea size CDATA #REQUIRED>
				<!ATTLIST dataarea width (8|16|32|64) "8">
				<!ATTLIST dataarea endianness (big|little) "little">
				<!ELEMENT rom EMPTY>
					<!ATTLIST rom name CDATA #IMPLIED>
					<!ATTLIST rom size CDATA #IMPLIED>
					<!ATTLIST rom crc CDATA #IMPLIED>
					<!ATTLIST rom sha1 CDATA #IMPLIED>
					<!ATTLIST rom offset CDATA #IMPLIED>
					<!ATTLIST rom value CDATA #IMPLIED>
					<!ATTLIST rom status (baddump|nodump|good) "good">
					<!ATTLIST rom loadflag (load16_byte|load16_word|load16_word_swap|load32_byte|load32_word|load32_word_swap|load32_dword|load64_word|load64_word_swap|reload|fill|continue|reload_plain|ignore) #IMPLIED>
			<!ELEMENT diskarea (disk*)>
				<!ATTLIST diskarea name CDATA #REQUIRED>
				<!ELEMENT disk EMPTY>
					<!ATTLIST disk name CDATA #REQUIRED>
					<!ATTLIST disk sha1 CDATA #IMPLIED>
					<!ATTLIST disk status (baddump|nodump|good) "good">
					<!ATTLIST disk writeable (yes|no) "no">

A-Noid

ajrhacker:
`disk` is an entirely different XML element which is only used for CHD software images. The `feature` and `sharedfeat` elements already allow the name and value to be practically any string.

A-Noid33:
@ajrhacker You're right.  This will protect all disks.  We definitely have some data disks in Apple2, which we would want to be writable.   It would probably be better to have it as an attribute on the rom element, so that each disk can be write-protected independently.   Just my opinion.

A-Noid.

ajrhacker:
The `<feature>` tag goes inside the `<part>` tag and only applies to that part (floppy disk, cartridge ROM, whatever). The `<sharedfeat>` tag goes outside and is distributed to all software parts.

A-Noid33:
@ajrhacker  Thanks... I understand now.  That should work.

A-Noid

A-Noid33:
@ajrhacker I verified that this change does work for Apple II both as a feature or a sharedfeat.  I'll update my pull request #14091.
Thanks, 
A-Noid

Issue #7091, opened 2020-08-16
i recreated brdrline sounds from schematics - but file names in mame don't fit


beta-tester:
hello,
i recreated Borderline (brdrline) sound files according its (original) schematic in LTSpice.

i used this [Borderline_-_1981_-_Sega.pdf](https://www.gamesdatabase.org/Media/SYSTEM/Arcade/Manual/formated/Borderline_-_1981_-_Sega.pdf) schematic.
the schematic is a little bit buggy (e.g. transistors were swapped)
but the result sounds like I mean to know them from my memory...

now i rebuild the folowing sounds from schematic:
* **ANIMAL-SOUND**, loop + fades out, uses MB4391.
* **CRY-SOUND**, fades out, uses MB4391.
* **EMAR-SOUND**, fades out, uses MB4391.
* **GUN-SOUND**, fades out, uses MB4391.
* **HIT-SOUND**, fades out, uses MB4391.
* **JEEP-SOUND**, loop.
* **POINT-SOUND**, fades out, uses MB4391.
* **WALK-SOUND**, loop, unknown period.

note: all sounds that fades out uses the MB4391. i could not find any information about that IC on internet,
so i assumed it is an audio mixer, but i don't know its characteristics - linearity.
i simply multiplied the control signal (at pin CON) with the input signal (at pin IN).
because of that, maybe the timing of the fade is too long.

ok, now i have the sound files but the file names at [mame/src/mame/audio/vicdual.cpp](https://github.com/mamedev/mame/blob/master/src/mame/audio/vicdual.cpp#L580) do absolutely not correspond to the names from the schematic.
so i don't know, what file name is the counterpart to the sound trigger from the schematic.

please, could somebody help me to give the files the right filename - or nicer, if somebody could define the the file names according the schematics.

and can somebody tell my how i can see how long which trigger is pulled high/low,
to be able to shorten sound files to the correct length.
also for the WALK-SOUND i don't know how long a period is.

here the recreation in LTSpice files and its sound files:
[brdrline-LTSpice.zip](https://github.com/mamedev/mame/files/5088926/brdrline-LTSpice.zip)

stilett0:
Thanks.

We'd rather this be implemented as a MAME netlist than by adding more sample sets, especially because you've gone and made a netlist with LTSpice anyways.

We've been discussing the MB4391 ourselves lately.

beta-tester:
is there a documentation of how to recreate sounds with the MAME netlist?
[...]

PR #14099, opened 2025-08-30
added an url of real hardware sounds as reference

beta-tester:
just as reference a video of real hardware sounds: https://www.youtube.com/watch?v=-U8iGhnAVCI

### DIFF
--- a/src/mame/sega/nl_brdrline.cpp
+++ b/src/mame/sega/nl_brdrline.cpp
@@ -1590,6 +1590,9 @@ NETLIST_START(brdrline)
 	 *          or trigger?
 	 *          or were the acrade what i played in the past a modified bootleg?
 	 *          or is it only in my head?
+	 *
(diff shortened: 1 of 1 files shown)


happppp:
Thanks, but no need for this. People know youtube has video/audio recordings of arcade games and can do a simple search.

edit: Ah, you're the author of this netlist, I didn't realize.
Still, IMO it's better to update that note with factual info, instead of a link to youtube.

```
	 *   note 1: as far as i remember (from the early 1980'th) it was triggered more often while crawling through the field at sector2 & 3,
	 *    issue in schematic/netlist?
	 *          or trigger?
	 *          or were the acrade what i played in the past a modified bootleg?
	 *          or is it only in my head?
```

beta-tester:
no problem.
(it was not that easy to find that video, 4 years ago :D )

EDIT: the video is linked in that issue https://github.com/mamedev/mame/issues/7091

PR #14100, opened 2025-08-30
sequential/prophet5.cpp: Preliminary driver for the Prophet 5 Rev 3.0

m1macrophage:
--------------------
Sequential Circuits Prophet 5 Rev 3.0 [DBWBP]

The synth's computer and the digital-analog interface should be fully emulated, except for cassette I/O and autotune. Analog audio is not emulated in this PR.


cuavas:
If you’re missing cassette I/O you can flag the system with something like:
```c++
	static constexpr feature_type unemulated_features() { return feature::TAPE; }
```
as a public class member.  This will show up in the internal UI.

m1macrophage:
> If you’re missing cassette I/O you can flag the system with something like [...]

Ah, right, done.

src/mame/sequential/prophet5.cpp
```diff
@@ -0,0 +1,973 @@
...
+1-part multitimbrality. See update_sh() for details.
+
+There is no dedicated ADC chip. The knobs and external CV are scanned by routing
+the knob (or input) voltages to a window comparator. Those voltages are compared
+with the ADC reference, which is controlled by the firwmare. See update_vmux()
+and adc_r() for details.
+
+TODO: Outline of voice architecture.
+
+This driver is based on the Prophet 5 Rev 3.0 technicam manual, and is intended
```
galibert: Minor typo there

src/mame/sequential/prophet5.cpp
```diff
@@ -0,0 +1,973 @@
...
+	PORT_START("seq_cv_in")  // J2, external CV input.
+	PORT_ADJUSTER(50, "SEQ CV IN") PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(prophet5_state::cv_in_changed), 0)
+
+	PORT_START("trimmer_seq_offset")  // R385, 100K trimmer.
+	PORT_ADJUSTER(50, "TRIMMER: SEQ OFFSET") PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(prophet5_state::seq_trimmer_adjusted), 0)
+
+	PORT_START("trimmer_seq_scale")  // R386, 10K trimmer.
+	// Default value calibrated for ADC_CV_SEQ_IN = CV_SEQ_IN / 2, with some
+	// error due to adjuster resolution. Exact calibration works out to 47.3.
+	PORT_ADJUSTER(47, "TRIMMER: SEQ SCALE") PORT_CHANGED_MEMBER(DEVICE_SELF, FUNC(prophet5_state::seq_trimmer_adjusted), 1)
```
galibert, line 957: Do we really want those as inputs instead of fixing the post-calibration value?

src/mame/sequential/prophet5.cpp
```diff
@@ -0,0 +1,973 @@
...
+INPUT_PORTS_END
+
+ROM_START(prophet5rev30)
+	ROM_REGION(0xc00, "maincpu", 0)  // 3 x 2708 1Kbyte ROMS.
+	ROM_LOAD("0.v8.1.u312", 0x000000, 0x000400, CRC(7cb704f0) SHA1(915c46486a983d2cee140cf6e9f5c505d620b956))
+	ROM_IGNORE(0x400)  // The first 1K bytes are duplicated. But 2708 ROMs should only contain 1K.
```
galibert: If the rom dumps should be half-size due to the hardware, trim them.  It's not good to enshrine semi-bad dumps.

src/mame/sequential/prophet5.cpp
m1macrophage: Oops, fixed, along with a few other typos.

src/mame/sequential/prophet5.cpp
m1macrophage: Done.

src/mame/sequential/prophet5.cpp
m1macrophage: This and the other "trimmer" inputs are physical trimmers on the PCB. If there aren't any big concerns,I have a preference for (and enjoy) emulating these, when practical, to more accurately represent the hardware and allow for experimentation.

PR #14103, opened 2025-08-31
konami/djmain.cpp: Use generic gfx decode layout, Minor cleanups

cam900:
- Use std::swap for swap layer
- Fix namings
- add notes


happppp:
The layer sort in screen_update: now that I see this, isn't this konami_sortlayers3?

cam900:
> The layer sort in screen_update: now that I see this, isn't this konami_sortlayers3?

It's not actually same as konami_sortlayers3; but similar purpose.

PR #13937, opened 2025-07-10
Sega UFO Catcher 7 Second dump + 7 DIP decription

pizzzza19:
UFO 7 Second ROM dumped.

Also i added UFO 7 DIP decription which was blank.

PR #14107, opened 2025-09-01
scripts/build: add a python script for checking EoF markers across source tree

angelosa:


angelosa:
1. Windows is untested;
2. May as well chain together with `check_include_guards.py`, no point in setting two different CIs for the same base env imo.

scripts/build/check_eof.py
```diff
@@ -0,0 +1,27 @@
...
+			with open(item, mode="r", encoding="utf-8") as fh:
+				try:
+					fh.seek(0, os.SEEK_END)
+					fh.seek(fh.tell() - CHAR_DEPTH, os.SEEK_SET)
+					value = fh.read(CHAR_DEPTH)
+					if value != os.linesep:
+						raise ValueError(value, item)
+				except UnicodeDecodeError:
+					print(item, file=sys.stderr)
+					raise
```
cuavas, line 27: This misses the various .ipp, .hxx, .lst, .inc, etc. files around the source tree.

I wonder if it wouldn’t be better to also check for various other stuff, perhaps:
* Parse the whole file as UTF-8 and check for invalid UTF-8, overlong sequences, surrogates, etc.
* Check for lines with trailing whitespace.
* Check for non-native line endings (if you run the script on Linux, this would catch checked-in DOS line endings).
* Check for X bit set on C-language sources.
* Maybe some other simple stuff?

I do wonder what performance would be like if it was written in Python, although considering how well makedep.py performs these days, it might not be too bad.

I doubt it would do much to deter the people who check this stuff in most frequently, but it would be nice to get a flag on PRs at least.  But then #13937 was merged when it obviously couldn’t build due to the added `GAME` line having the same short name as an existing one.  This gives the impression that people aren’t even checking the most basic things before merging stuff, so I don’t know how effective anything will be.

angelosa: > I do wonder what performance would be like if it was written in Python, although considering how well makedep.py performs these days, it might not be too bad.

Roughly 0.17 seconds locally.

angelosa: > I doubt it would do much to deter the people who check this stuff in most frequently, but it would be nice to get a flag on PRs at least. But then https://github.com/mamedev/mame/pull/13937 was merged when it obviously couldn’t build due to the added GAME line having the same short name as an existing one. This gives the impression that people aren’t even checking the most basic things before merging stuff, so I don’t know how effective anything will be.

Sounds like a different script type, one that cross checks .lst with .cpp files in `src/mame` (don't think there are any other file extensions that use a `GAME` macro?)

cuavas: Yeah, but made the regular “compile MAME” jobs fail anyway due to duplicate symbols, so the person creating the PR and the person merging it already weren’t looking at build results.  Having another CI job fail won’t fix that.

firewave: Clang has a compiler warning for this:
```
/home/user/a.cpp:3837:2: warning: no newline at end of file [-Wnewline-eof]
 3837 | }
      |  ^
```

Automatic EOL handling in git is done via `.gitattributes` (updating an existing repo is fidgety though).

Invalid UTF-8 sequences in comments are caught by Clang with `-Winvalid-utf8`. Detetcing such sequences in actual code is only in clang-tidy IIRC.

cuavas: > Clang has a compiler warning for this:

That won’t address things that aren’t C++ source files (e.g. Python, Lua, XML, and various inputs to source generator tools).

> Automatic EOL handling in git is done via `.gitattributes` (updating an existing repo is fidgety though).

It’s [already set](https://github.com/mamedev/mame/blob/master/.gitattributes), but that doesn’t stop people from not setting `core.autocrlf=true` or otherwise messing things up and checking in DOS line endings as happened in 81a978429cb4d09855a0fc816123b840b6780266.

Also, none of that changes the fact that automated checks are useless if people just ignore the results.

angelosa: > Also, none of that changes the fact that automated checks are useless if people just ignore the results.

Can you protect master branch to avoid merging PRs if they fails for whatever reason? Needs admin rights in mamedev settings, around `Require status checks to pass before merging`

cuavas: I see @Osso13 likes this idea.  I guess it won’t be so bad as it would have been in the past now that CI runs don’t take as long as they used to (it used to be about three hours for macOS and over four hours for Windows, now it’s down to a bit over two hours for Windows).

What do other people think.  @galibert?

galibert: I wouldn't mind, I usually wait for CI when merging anything.

cuavas: OK, I tried enabling this rules that CI has to pass, but it won’t work at the moment.  The trouble is, not all jobs apply to all code changes, so e.g. if I enable checking the `build-linux (gcc)` status, it will prevent merging a PR that only affects the docs (and vice versa for `build-docs` status).

To make this workable, we need dummy jobs that set status results for the inverse conditions.

PR #14110, opened 2025-09-01
Better controls

Gorilli09:
SDL is better


happppp:
Is this a statement or a PR? What you have here won't work.
Although it hasn't been updated yet to 3.0, MAME already supports SDL via a compile option.

happppp:
Now you changed from trying to add SDL3 to fiddling with strings? What is your aim here?
A lot of the strings you changed, aren't supposed to, and the button offset in seattle.cpp (and other racing games) is intentional because of a conflict with digital pedal inputs.

happppp:
Closing, this won't work out.
If you want to practice coding MAME, start smaller. And set up a compile environment first to test your work.

PR #14111, opened 2025-09-01
scripts/build: add a python script for checking style rules

stonedDiscord:
I realize this is in direct competition with #14107, I started work on this on saturday when I posted the question if this could be automated in the Discord server.

The action will post code reviews, much like a human reviwer would.
An example can be seen here: https://github.com/stonedDiscord/mame/pull/5

I went through https://docs.mamedev.org/contributing/cxx.html to come up with these rules.
It will detect
- Missing newline at EOF
- Invalid UTF-8
- Wrong type of line ending
- Trailing whitespaces
- Uppercase hex literals
- Single line block comments
- Lowercase constants
- Uppercase functions
- Uppercase class names
- Uppercase Enum names

It mostly uses regex to detect those, I haven't run into any false positives yet, but I'm on the fence about dropping the constants check because it will complain about variables that someone has marked const.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -20,11 +20,15 @@ jobs:
       - name: Install reviewdog
         uses: reviewdog/action-setup@v1.3.2
 
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v46.0.5
+
       - name: Run style check
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          files=$(git diff --name-only --diff-filter=d "${{ github.event.base_ref }}~1" "${{ github.event.head_ref }}" | grep -E '\.(c|cpp|h|hpp)$' || true)
+          files=$(echo ${{ steps.changed-files.outputs.all_changed_files }} | grep -E '\.(cpp|h)$' || true)
           if [ -n "$files" ]; then
             python3 scripts/build/style.py $files \
               | reviewdog -efm="%f:%l: %m" \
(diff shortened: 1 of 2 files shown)


cuavas:
And ironically, it’s a Python script with no final newline…

stonedDiscord:
I'm aware that I get this constantly wrong, that's why I asked and then wrote this script.
I guess I'm simply not used to that style rule and I will probably get it wrong again.
The point of the script is to let newbies like myself deal with the review bot so maintainers have more time to review PRs that are ready for merging and don't get fed up or annoyed as much.
The CI doesn't block merging, same as with human reviews.

.github/workflows/style.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  stylecheck:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install reviewdog
+        uses: reviewdog/action-setup@v1.3.2
```
angelosa: I don't like the idea of using an external dependency without a discussion first.

stonedDiscord: The alternative is to do it inside the script using the API
https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#about-pull-request-review-comments
I can give it a go if you don't want to rely on an external action.

stonedDiscord:
I've changed it to use no external dependencies and to post a single review with multiple comments, fixing any rate limit issues before they appear.

I've tested the style checker against #14020 here:
https://github.com/stonedDiscord/mame/pull/7

stonedDiscord:
Added checks:
- SPDX header
- Include order
- ROM region

PR #14112, opened 2025-09-02
igs/goldstar.cpp: update the % Main Game Pay Rate for jkrmast

jbtx77gr:
update the % Main Game Pay Rate (jkrmast)


cuavas:
What are the updated values based on?  Did you confirm that the values are correct for all three versions using these I/O port definitions?

Also, where did the current (presumably incorrect) values actually come from?

jbtx77gr:
The original values were taken from the game manual --> https://pdfhost.io/v/JZ3as5f7RS_Joker_Master_2000_SE_8LS_Manual_PDF . Usually, the manuals cover multiple versions, and the differences between them are minimal.
For the two versions I have already checked (jkrmast & jkrmasta), the payout settings are 55%, 65%, 75%, and 85%.

I will check the third, newer version later, and if any adjustments are needed, we’ll apply them. We’ll also be uploading additional sets from this series soon, some of which support a lower payout option of 40%, as documented in the original manual.

For now, the recent updates only apply to the sets jkrmast and jkrmasta.

cuavas:
But your change here will also affect `jkrmastb` – can you check to see if it’s correct for that version as well?

jbtx77gr:
i have check for jkrmastb, it uses 40%, we commit later for it. also we commit the correct coin values for each version too soon. no panic , all are under control. I’ll try to fix this whole mess with the coin values, but it will take some time because I need to carefully check each version one by one. Most of the cloned games automatically inherit values from the main version, and many of those values are incorrect.

RobertoFresca:
Changes are now merged in the driver.

PR #14114, opened 2025-09-02
Am6012/tl084 DAC analog emulation

Madcadden:
Added AM6012/TL084 DAC analog emulation.

Emulating AM6012/TL084 DAC based off info from Atari schematics. Cleans up audio pops/clicks and distortion especialy during car crashes.

### DIFF
--- a/src/mame/atari/harddriv.h
+++ b/src/mame/atari/harddriv.h
@@ -571,6 +571,20 @@ private:
 	uint16_t                m_comram[0x400/2]{};
 	uint64_t                m_last_bio_cycles = 0;
 
+
+	// --- AM6012/TL084 DAC analog path emulation state ---
+	emu_timer*             m_am6012_timer = nullptr;   // mixer timer (e.g., 96 kHz)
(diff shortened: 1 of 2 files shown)


happppp:
I haven't looked at the schematics myself, but if there's a filter and the cause of the pops/clicks you're trying to prevent here aren't due to something else, we have src/devices/sound/flt_biquad.

Out of curiosity, did an AI write half of the source code here? Not that it's a bad thing, as long as you understand what it wrote and why.

Madcadden:
Hi, yeah i used A.I to help me code this after i found the PCB schematics and some notes from Jed Margolin. I tried to use the src/devices/sound/flt_biquad you mentioned and i was able to make it crackle and pop less but couldn't tune the variables in a way that got rid of it 100% and it sounded muffled maybe somebody else who knows how to use that properly can do it. 

i found out later that the defaults i got for this currently aren't in line with what Jed said about the sound output on the original hardware it would be closer to the real thing if the K12=220 and K3=160. I tried those out and it still worked good.

PR #14117, opened 2025-09-03
bus/megadrive/titan.cpp, bus/megadrive/rom.cpp: Add notes for mapper chip

cam900:
In Mega drive/Genesis Super Street Fighter 2 Cartridge, Sega 315-5709/315-5779 controls bankswitching and determines upper half of ROM space (Backup RAM or ROM).

Titan Overdrive 2 requires compatible mapper hardware as above, but ROM space is forced to ROM?

Reference (for mapper chip): https://segaretro.org/images/7/78/IC_BD_16M_42PIN_x_4_EPROM_32X_RD_User%27s_Manual.pdf

### DIFF
--- a/src/devices/bus/megadrive/rom.cpp
+++ b/src/devices/bus/megadrive/rom.cpp
@@ -546,7 +546,7 @@ uint16_t md_rom_fram_device::read_a13(offs_t offset)
 }
 
 /*-------------------------------------------------
- SUPER STREET FIGHTER 2
+ SUPER STREET FIGHTER 2 (with 315-5709/315-5779)
  -------------------------------------------------*/
 
 uint16_t md_rom_ssf2_device::read(offs_t offset)
@@ -573,6 +573,14 @@ void md_rom_ssf2_device::write_a13(offs_t offset, uint16_t data)
 				m_bank[offset] = data & 0xf;
 				memcpy(ROM + offset * 0x080000/2, ROM + 0x400000/2 + (m_bank[offset] * 0x080000)/2, 0x080000);
 			}
+			/*
+			else
+			{
+				D0: Determine upper half (0x200000 onward) of ROM space: Backup RAM(1) or ROM(0)
+				D1: Backup RAM Write protect
+				(No official licensed software used these features)
+			}
+			*/
 		}
 	}
 }
--- a/src/devices/bus/megadrive/titan.cpp
+++ b/src/devices/bus/megadrive/titan.cpp
@@ -6,6 +6,8 @@ Titan Overdrive 2 mapper
 
 https://plutiedev.com/beyond-4mb
 
+- Bankswitching controlled by SEGA 315-5709/315-5779 mapper or compatible hardwares
+
 TODO:
 - This is same as SSF2 mapper, merge at cart slot rewrite;
 
@@ -62,6 +64,14 @@ void md_rom_titan_device::write_a13(offs_t offset, uint16_t data)
 			// D7-D6 unconnected, sgdk_bap (and likely others) cares about D4
 			m_bank[offset] = data & 0x3f;
 		}
+		/*
+		else
+		{
+			D0: Determine upper half (0x200000 onward) of ROM space: Backup RAM(1) or ROM(0)
+			D1: Backup RAM Write protect
+			(No official licensed software used these features)
+		}
+		*/
 	}
 }
 


angelosa:
[Phantasy Star IV and Sonic 3 & Knuckles (thru lock-on?)](https://github.com/jsgroth/jgenesis/wiki/Tricky%E2%80%90to%E2%80%90Emulate-Games#phantasy-star-iv) actually depends on the backup RAM to be view selectable (those uses different mappers). 
Super Street Fighter 2 shouldn't have any SRAM, and it's the *only* official release known to need the ROM banks. The 32x development cartridge refers to a different configuration, very unlikely that is implemented here.
For homebrews the conventional form is to use `SEGA SSF` in the header at $100, cfr. [plutiedev reference](https://plutiedev.com/beyond-4mb), and the official [documentation](https://krikzz.com/pub/support/mega-everdrive/x3x5x7/dev/extended_ssf-v2.txt) actually states that bank 0 is mappable and SRAM overlay is on a different bit, something I haven't touched because Titan mapper is a one-shot PoC until I have the chance to rewrite this.

Further reading:
https://github.com/TASEmulators/BizHawk/issues/695

src/devices/bus/megadrive/rom.cpp
```diff
@@ -573,6 +573,14 @@ void md_rom_ssf2_device::write_a13(offs_t offset, uint16_t data)
 				m_bank[offset] = data & 0xf;
 				memcpy(ROM + offset * 0x080000/2, ROM + 0x400000/2 + (m_bank[offset] * 0x080000)/2, 0x080000);
 			}
+			/*
+			else
+			{
+				D0: Determine upper half (0x200000 onward) of ROM space: Backup RAM(1) or ROM(0)
+				D1: Backup RAM Write protect
+				(No official licensed software used these features)
```
cam900: "No official licensed software used these features" in here means "No official licensed software used SRAM overlay and write protect register in 315-5709/315-5779 mapper chip".
These registers are used in official licensed software without 315-5709/315-5779 mapper chip.

angelosa: Again: what's your rationale/reference that makes you think Super Street Fighter 2 uses a 32x dev cart with those numbers?

cam900: Reference is for mapper chip, and i think dev cart mentioned at reference and actual MD Super Street Fighter 2 cart PCB has same mapper chip. There's another dev cart has same mapper chip as reference/MD Super Street Fighter 2 cart PCB: https://segaretro.org/images/a/a0/IC_BD_4M_32PIN_x_8_EPROM_32X_RD_User%27s_Manual.pdf

angelosa: PCB picture, post it.

cam900: https://gendev.spritesmind.net/mirrors/www.nepereny.com/misc2/games/genesis_super_street_fighter_2.htm here is PCB for MD Super Street Fighter 2; it has 315-5779, but compatible as 315-5709.
Some PCB has 315-5709 instead 315-5779.
see also: megadriv.xml mentioned 315-5709 IC in Super Street Fighter 2.

PCB with 315-5709: https://storage.googleapis.com/images.pricecharting.com/6b41c02b48a6a12967bf17268e0b259a0dc1c94097afd863c1c6ebee0d85bec5/1600.jpg

angelosa: That doesn't mean the mapper has an internal SRAM / has the SRAM view select hooked up, does it?

cam900: Yeah: SRAM is external, and just overlays upper half of 68K cartridge space if access enabled.

cam900: > Yeah: SRAM is external, and just overlays upper half of 68K cartridge space if access enabled.

If SRAM space overlay in 0x200000 onward is enabled when SRAM is not connected, then memory space area in 0x200000 onward will be just empty or open-bus.

angelosa: Ok, and what's the end boundary for SSF2, assuming you tested this on an actual cart?

angelosa: SSF2 init goes:
$a130fd = 8
$a130ff = 9
[do some other stuff]

$a130f1 = 2 (unmap SRAM at $200000 to ?)
$a130f3 = 1
$a130f5 = 2
[remaining banks]
$a130ff = 7

cam900: It's each register is 6 bit, so bank boundary is up to 32MB(0 to 63). empty area is just open bus?

angelosa: I don't mean the ROM, I mean the overlay where SRAM should live.

```c++
void megadrive_rom_ssf2_device::cart_map(address_map &map)
{
	for (int i = 0; i < 8; i++)
	{
		const u32 page_size = 0x08'0000;

		map(0x000000 | (page_size * i), 0x07ffff | (page_size * i)).bankr(m_rom_bank[i]);
	}

	map(0x200000, <unknown_end_boundary>).view(m_sram_view);
	m_sram_view[0](0x200000, <unknown_end_boundary).rw(/* handlers_for_sram, omitted for simplicity */);
}
```

cam900: SRAM is mapped up to 0x200000 + its size, onward to 0x3fffff is open bus when SRAM accessing bit enabled.

angelosa: I doubt your mask is right, in the `[do some other stuff]` above there's ROM checksum for 0x300000-0x3fffff, throwing a red screen of death if the SRAM view would respond there.

cam900: so reference in above has wrong info maybe(it says 0x200000 to 0x3fffff is mapped to SRAM when SRAM access bit is enabled, but it's not true and actually lesser area is mapped to SRAM?) even its official documents from sega...

angelosa: /VRES does the trick, B27 connection prevents SRAM being visible at startup while mapping the banks linearly.
image

cam900: That means, 0xa130f1 is should be set to 0x02 at reset, right? (RAM accessing disabled and write protected)

angelosa: That's my understanding of it yeah. Notice that I've already implemented this in the upcoming rewrite.

angelosa:
Closing, don't have any use of this (and touched files will be removed at some point anyway).

PR #14118, opened 2025-09-04
add notes about 30 when press A and K key marywu.cpp

flama12333:


flama12333:
I Forgot something
 If hold Key A or K then press f3 to reset, will shown error 30 code.

PR #14121, opened 2025-09-04
i8256: calculate timer from external clock

stonedDiscord:
The datasheet said
`If FRQ=0, the timer input frequency is 16kHz.`
That is correct, but only if the input frequency is also correct and the divider (C0,C1) is set accordingly.


stonedDiscord:
included in #14478

PR #14126, opened 2025-09-05
sony/zn.cpp: Fix PCB info, SFEX needs two KM4132G271Q-12

clawgrip:


Gurudumps:
No, sorry. While the RAM addition is accurate for *some* games, DO NOT ADD A FULL STOP TO EVERY LINE ON MY README'S.
The . on some lines of my readme's is when I make a comment. For a basic parts list the full stop is not added. Please do not make random and pointless changes to my readme's.

ghost:
No, sorry. While the RAM addition is accurate for some games, DO NOT ADD A FULL STOP TO EVERY LINE ON MY README'S.
You kept the * on that line but deleted the line that explains what the * means, so as it is now the * is superfluous to requirements (but in reality is still required).
The . on some lines of my readme's is when I make a comment, as per standard writing punctuation rules. For a basic parts list the full stop is not added. Please do not make random and pointless changes to my readme's.
I will submit a fix myself for the RAM. Please close this.

felipesanches:
It is very fair to complain about the added full stops. But I wonder about the sentence "Usually unpopulated, but needed by a few games (like Street Fighter EX/EX Plus)." Is that valid info?

ghost:
This is known info. Check the src. There are 2 types of machines, one with 1MB VRAM and one with 2MB VRAM. Those 2MB machines are the ones with 2 of those RAMs populated.
It's not that important, it's just a doc not a design specification ;-)
Now 20+ years later with all games working it's really just something that makes the src more interesting to non-coders and arcade enthusiasts.
As I said I'll submit a proper fix for the doc soon.

Hammy1986:
SFEX is 100% a 2x vram game, but it will run with 1x vram with a few GFX glitches.

ghost:
Yes we know. There are many games that use 2 VRAMs (not only some Capcom games) and they will have graphics glitches when only 1 is populated. This is KNOWN. Do you think it's pure coincidence that all the games are working correctly in MAME? LOL ;-)
For each manufacturer (when required) there are two machines in the src, one with 1MB VRAM and one with 2MB VRAM.

Osso13:
Thanks for the PR. Closing because I will commit an updated readme later.

PR #14129, opened 2025-09-06
konami/twin16_v.cpp: Make Konami Twin 16 Video subsystem to device

cam900:
for use callback for interrupt, sprite and tile banking
- Use logmacro.h for logging function
- Use <algorithm> for replace memset and memcpy to std::fill* and std::copy*
- Reduce duplicates
- Fix variable type
- Fix spacings
- Minor optimization in sprite draw routine

konami/twin16.cpp: Cleanups & Updates
- consolidated driver into single file
- Fix notes in comments
- Fix metadata for majuu because it's japanese release
- Move driver initialization function into machine_start()
- Reduce duplicates
- Fix spacing
- Fix naming for capitalization consistency
- use BIT macro for single bit values


sjy555525:
I play the Devil World / Majuu no Ohkoku, Vulcan Venture / Gradius II: GOFER no Yabou has a Graphics Problem

Devil World / Majuu no Ohkoku



Vulcan Venture / Gradius II: GOFER no Yabou

angelosa:
> I play the Devil World / Majuu no Ohkoku, Vulcan Venture / Gradius II: GOFER no Yabou has a Graphics Problem

Happening with this PR or in general?

cam900:
@sjy555525 Possible fix is pushed, Please retest for these.

src/mame/konami/twin16_v.cpp
```diff
@@ -35,69 +43,114 @@ enum
...
+
+	m_fixed_tmap = &machine().tilemap().create(*this, tilemap_get_info_delegate(*this, FUNC(konami_twin16_video_device::fix_tile_info)), TILEMAP_SCAN_ROWS, 8, 8, 64, 32);
+	m_scroll_tmap[0] = &machine().tilemap().create(*this, tilemap_get_info_delegate(*this, FUNC(konami_twin16_video_device::scroll_tile_info<0>)), TILEMAP_SCAN_ROWS, 8, 8, 64, 64);
+	m_scroll_tmap[1] = &machine().tilemap().create(*this, tilemap_get_info_delegate(*this, FUNC(konami_twin16_video_device::scroll_tile_info<1>)), TILEMAP_SCAN_ROWS, 8, 8, 64, 64);
+
+	m_fixed_tmap->set_transparent_pen(0);
+	m_scroll_tmap[0]->set_transparent_pen(0);
+	m_scroll_tmap[1]->set_transparent_pen(0);
+
+	palette().set_shadow_factor(0.4); // screenshots estimate
```
cuavas: Should this be here?  It’s a function of the mixer – should it be done with palette initialisation in the system driver?

src/mame/konami/twin16_v.cpp
```diff
@@ -35,69 +43,114 @@ enum
...
+	const size_t spritebuffer_size = 0x1000 / 2;
+	m_fixram = make_unique_clear<uint16_t []>(fixram_size);
+	m_sprite_buffer = std::make_unique<uint16_t []>(spritebuffer_size);
+	for (int i = 0; i < 2; i++)
+	{
+		m_videoram[i] = make_unique_clear<uint16_t []>(videoram_size);
+		m_spriteram[i] = make_unique_clear<uint16_t []>(spriteram_size);
+		save_pointer(NAME(m_videoram[i]), videoram_size, i);
+		save_pointer(NAME(m_spriteram[i]), spriteram_size, i);
+	}
```
cuavas: Would it be more appropriate to use `memory_share_creator` and `memory_share_array_creator` for these?

src/mame/konami/twin16_v.cpp
cam900: It's leftover from original twin16 driver video_start(). I will move that in machine driver later.

cuavas:
It would probably be cleaner to make the video device implement `device_memory_interface` so the driver can install ROM/RAM rather than using a read callback.  I’ll merge it in its current state, anyway.

There are five games affected by this, most of which have multiple regional variants/versions.  It would be nice if someone could give them a play through and report any regressions here.

PR #14135, opened 2025-09-07
skeleton/elcirculo.cpp: add layout

stonedDiscord:
the LEDs roughly go in a circle

### DIFF
--- a/src/mame/skeleton/elcirculo.cpp
+++ b/src/mame/skeleton/elcirculo.cpp
@@ -66,7 +66,7 @@ public:
 	elcirculo_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
-		, m_leds(*this, "ld%u", 0U)
+		, m_leds(*this, "ld%u%u", 0U, 0U)
 		, m_lamps(*this, "il%u", 0U)
 	{
 	}
@@ -78,14 +78,16 @@ protected:
 
 private:
 	required_device<cpu_device> m_maincpu;
-	output_finder<38> m_leds;
+	output_finder<4,8> m_leds;
 	output_finder<8> m_lamps;
 
 	uint8_t m_led_row = 0;
+	uint8_t m_led_data = 0;
 
 	INTERRUPT_GEN_MEMBER(irq_gen);
 
 	void led_w(offs_t offset, uint8_t data);
+	void strobe_w(offs_t offset, uint8_t data);
 
 	void program_map(address_map &map) ATTR_COLD;
 };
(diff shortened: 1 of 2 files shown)


happppp:
Don't use output().set_value(), it's deprecated. Use output_finder, I see you already have output_finder in the driver but you're not using it? m_leds[m_led_row << 3 | bit] = BIT(...)

Are the lamps really selected by setting all 4 led row bits to 0? Not in the high nibble or somewhere else?

src/mame/skeleton/elcirculo.cpp
```diff
@@ -82,15 +95,53 @@ INTERRUPT_GEN_MEMBER(elcirculo_state::irq_gen)
...
+	if (offset == 1)
+	{
+		switch (data & 0x0f)
+		{
+			case 0x0e: m_led_row = 0; break;
+			case 0x0d: m_led_row = 1; break;
+			case 0x0b: m_led_row = 2; break;
+			case 0x07: m_led_row = 3; break;
+			case 0x00: m_led_row = 4; break;
+			default: logerror("row %02x\n",data); break;
```
cuavas: This looks like a one-hot mapping – a low bit drives a LED row.  All bits low driving the lamps doesn't make a great of sense.

src/mame/skeleton/elcirculo.cpp
```diff
@@ -82,15 +95,53 @@ INTERRUPT_GEN_MEMBER(elcirculo_state::irq_gen)
...
+	}
+	else if (offset == 0)
+	{
+		if (m_led_row <= 3)
+		{
+			for (int bit = 0; bit < 8; bit++)
+			{
+				m_leds[m_led_row*10 + bit] = !BIT(data, bit);
+			}
+		}
```
cuavas: You’ve got dead entries in your LEDs like this.  Have you considered using a two-dimensional output finder?

stonedDiscord: I did not know about these before

stonedDiscord:
I've loaded the binary into Ghidra and it writes to the LEDs like this
```
void FUN_9457(void)

{
  DAT_0400 = DAT_002a | DAT_002e;
  DAT_0401 = DAT_002b ^ 0xc0 | DAT_002f;
  DAT_0600 = DAT_002c ^ 0xff;
  DAT_0601 = (DAT_002d | DAT_0031) ^ 0xff;
  return;
}
```
I can't find any writes for the lamps, so I've removed them again for now.

src/mame/skeleton/elcirculo.cpp
```diff
@@ -82,15 +97,50 @@ INTERRUPT_GEN_MEMBER(elcirculo_state::irq_gen)
...
+			for (int bit = 0; bit < 8; bit++)
+			{
+				m_leds[m_led_row][bit] = !BIT(m_led_data, bit);
+			}
+		}
+		else
+		{
+			logerror("led_row %02x", m_led_row);
+		}
+}
```
cuavas: Your indentation is off.

Also, why not store the LED drive directly, and then update whichever rows are active?
```c++
void elcirculo_state::led_w(offs_t offset, uint8_t data)
{
	if (offset == 1)
	{
		m_led_row = data & 0x0f;
	}
	else if (offset == 0)
	{
		m_led_data = data;
	}
}

void elcirculo_state::strobe_w(offs_t offset, uint8_t data)
{
	for (unsigned row = 0; std::size(m_leds) > row; ++row)
	{
		if (!BIT(m_led_row, row))
		{
			for (int bit = 0; bit < 8; bit++)
			{
				m_leds[row][bit] = !BIT(m_led_data, bit);
			}
		}
	}
}
```

PR #14140, opened 2025-09-07
neogeo.xml: Added 3 working items

ArcadeShadow:
Bad Dudes - Burger Edition (20250628) [OzzyOuzo]
Double Dragon One (beta 2, 20250903) [OzzyOuzo]
In The Hunt (demo 20250518) [OzzyOuzo]

### DIFF
--- a/hash/neogeo.xml
+++ b/hash/neogeo.xml
@@ -11385,7 +11385,7 @@ license:CC0-1.0
 		</part>
 	</software>
 
-		<!-- Unlicensed Prototype -->
+	<!-- Unlicensed Prototype -->
 
 	<software name="diggerma">
 		<description>Digger Man (prototype)</description>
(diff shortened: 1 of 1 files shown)


cuavas (commented):
Does Itch.io really count as a publisher, or is it a storefront that people self-publish on?  I see it more as a distributor for self-published software.

Kind of like how GoG is a distributor owned by CD Project, which sells games published by other CD Project companies, but also acts as a storefront for games from other publishers.

ArcadeShadow:
> Does Itch.io really count as a publisher, or is it a storefront that people self-publish on? I see it more as a distributor for self-published software.
> 
> Kind of like how GoG is a distributor owned by CD Project, which sells games published by other CD Project companies, but also acts as a storefront for games from other publishers.

Fair enough. I replaced the Itch.io references by the developer's name.

PR #14141, opened 2025-09-07
Add toggle hotkey to autofire plugin

jackrjli:
I added an optional toggle hotkey for the autofire plugin that turns autofire on/off for a button. This is useful for games where you want to have both autofire and non-autofire but don't want different buttons for the two. For example, in Fire Shark the blue and green weapons benefit from autofire, but the red weapon breaks with autofire. The toggle hotkey lets you turn off autofire when using the red weapon without having to switch to a different fire button.

### DIFF
--- a/docs/source/plugins/autofire.rst
+++ b/docs/source/plugins/autofire.rst
@@ -62,6 +62,16 @@ may need higher numbers than 1 for the system to recognise the button being
 released and pressed again (e.g. 2 on frames and 2 off frames works for Alcon).
 Experiment with different values to get the best effect.
 
+Select **Toggle hotkey** to set a control to enable/disable autofire on the
+button.  This is optional, and may be skipped if not needed.  Press the UI Clear
+key or select the **Unset toggle hotkey** menu item (which appears when the
+toggle hotkey is set) to remove the hotkey.  When the toggle hotkey is set,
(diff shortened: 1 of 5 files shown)


cuavas:
I realise there are games where you want autofire in some situations but not others (e.g. Fire Shark that you mentioned, and also in Super X you *don’t* want autofire for the blue weapon at high power levels or any weapon with Super Power, but you *do* want autofire for the red and green weapons without Super Power and also for the blue weapon at low power levels).

But how much benefit does this approach actually provide in practice?  Instead of having separate controls for autofire and non-autofire, you need separate controls fire and toggle autofire.  It doesn’t actually save you a control – you need two controls either way.

jackrjli:
Yes, you'd still need two controls; the difference would be whether you'd tap or hold that second button. In theory you could save on buttons if you have multiple autofire buttons that you toggle with the same control, but I don't know of any games where you'd actually do that. I admit the desire for this is probably quite niche; I haven't seen (or looked for) anyone asking for this.

jackrjli:
I'm going to close this for now - if I come back to this, there are some things I'd want to improve with it, such as the lack of feedback when activating a toggle. That said, if anyone really wants this, I have the plugin posted as a zip file [here](https://github.com/jackrjli/mame/releases/tag/autofire-plugin_v0.0.5).

PR #14149, opened 2025-09-09
att6300p.cpp: Initial driver for AT&T 6300 Plus

donohoe00:
Initial driver for the AT&T 6300 plus.  It seems to run DOS just fine.

The 6300 Plus contains unique hardware to support a virtual 8086 environment allowing DOS to run concurrently with, and under the control of, a version of AT&T SVR 2.0 Unix that was specifically provided for this machine.  While I've attempted to implement the virtualization hardware, it's largely untested and would definitely need more work.  Unless someone unearths a copy of the right version of Unix (and the accompanying OS Merge software), the virtualization implementation cannot be completed and fully verified.  There are more details in the code comments about the exact version of Unix that's needed.

This is not your run-of-the-mill PC clone, and I'm expecting to get a lot of feedback, especially about how I hooked into accesses from the CPU to implement the virtualization.

Note that this PR includes a proposed change to how the i86/i286 registers are initialized upon reset.  This is obviously a risky change, but the BIOS fails when performing the memory check without it (since it fails to return from protected mode to real mode when testing memory above 640K).


src/mame/olivetti/att6300p_mmu.cpp
```diff
@@ -0,0 +1,305 @@
+// copyright-holders:D. Donohoe
+/**********************************************************************
```
startaq, line 3: Missing license header.

src/mame/olivetti/att6300p_mmu.h
```diff
@@ -0,0 +1,83 @@
...
+private:
+	devcb_write32 m_trapio;
+
+	const address_space_config m_mem_config;
+	const address_space_config m_io_config;
+	memory_access<20, 0, 0, ENDIANNESS_BIG>::specific m_mem8_space;
+	memory_access<24, 1, 0, ENDIANNESS_LITTLE>::specific m_mem16_space;
+	address_space *m_io;
+
+	void update_fastpath();
```
startaq: Don't mix variables and functions. Put either functions or variables at the top of this section.

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+	{ }
+
+	void att6300p(machine_config &config);
+
+protected:
+	virtual void machine_start() override ATTR_COLD;
+	virtual void machine_reset() override ATTR_COLD;
+
+private:
+	/*
```
startaq, line 100: This is really quite verbose. Maybe it can moved down to where it's used and be a bit more condensed (assuming this isn't reverse engineered but also mentioned in the manual).

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+		configure_mapping(m_mapping_sel);
+		m_mmu->set_protected_mode_enabled(enabled);
+	}
+}
+
+uint8_t att6300p_state::virt_r(offs_t offset)
+{
+	uint8_t reg = offset & 0xe0;
+
+	switch (reg)
```
startaq: It's usually preferred to let the `address_map` handle the decoding if possible and use individual functions instead.

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+	m_dmac->out_iow_callback<3>().set(FUNC(att6300p_state::dma_io_write<3>));
+	m_dmac->out_dack_callback<0>().set(FUNC(att6300p_state::dma_dack_w<0>));
+	m_dmac->out_dack_callback<1>().set(FUNC(att6300p_state::dma_dack_w<1>));
+	m_dmac->out_dack_callback<2>().set(FUNC(att6300p_state::dma_dack_w<2>));
+	m_dmac->out_dack_callback<3>().set(FUNC(att6300p_state::dma_dack_w<3>));
+	m_dmac->out_eop_callback().set(FUNC(att6300p_state::dma_tc_w));
+
+	PIC8259(config, m_pic);
+	m_pic->in_sp_callback().set_constant(1);
+	m_pic->out_int_callback().set(FUNC(att6300p_state::int_w));
```
startaq: You can replace this with `set_inputline(m_maincpu, INPUT_LINE_IRQ0);`

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+	m_ctc->zc_callback<0>().set(m_ctc, FUNC(z80ctc_device::trg1));
+	m_ctc->zc_callback<1>().set(FUNC(att6300p_state::zc01_cb));
+	m_ctc->zc_callback<2>().set(FUNC(att6300p_state::zc02_cb));
+
+	/* software lists */
+	SOFTWARE_LIST(config, "disk_list").set_original("ibm5150");
+}
+
+ROM_START( att6300p )
+	ROM_REGION16_LE(0x8000,"bios", 0)
```
startaq: Needs a space after the value (another one same below).

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+
+	ROM_REGION(0x400, "kbc", 0)
+	ROM_LOAD("d8741a_kbc_8w_v2.1.bin", 0x000, 0x400, CRC(e5a95ace) SHA1(f87f83ae1be790cde227514168f48bd7f806b2fa))
+ROM_END
+
+} // Anonymous namespace
+
+
+//    YEAR  NAME       PARENT   COMPAT  MACHINE   INPUT     CLASS           INIT        COMPANY     FULLNAME      FLAGS
+COMP( 1985, att6300p,  ibm5150, 0,      att6300p, att6300p, att6300p_state, empty_init, "AT&T",     "6300 Plus",  MACHINE_NOT_WORKING )
```
startaq: Very likely the `ibm5150` shouldn't be the parent, this seems to be a fully original machine.

Also add `MACHINE_SUPPORTS_SAVE` if the relevant state is saved.

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+	m_ctc->zc_callback<2>().set(FUNC(att6300p_state::zc02_cb));
+
+	/* software lists */
+	SOFTWARE_LIST(config, "disk_list").set_original("ibm5150");
+}
+
+ROM_START( att6300p )
+	ROM_REGION16_LE(0x8000,"bios", 0)
+
+	ROMX_LOAD("att6300p_bios_6j_v1.04_lo.bin", 0x0000, 0x4000, CRC(dd0b9335) SHA1(56783b84f34d900b3cc73d1f7f1291a4211271ec), ROM_SKIP(1))
```
startaq: Instead of `ROM_SKIP(1)` you can use `ROM_LOAD16_BYTE`.

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+	update_nmi();
+}
+
+void att6300p_state::update_nmi()
+{
+	m_nmi_active = (m_trapio_active ||
+	  (m_nmi_enable && BIT(m_ctrlport_b, 6) && !BIT(m_ctrlport_a, 4)));
+
+	if (m_nmi_active)
+		m_maincpu->set_input_line(INPUT_LINE_NMI, ASSERT_LINE);
```
startaq, line 623: Are you sure this isn't pulsed? I. e. something like `m_maincpu->pulse_input_line(INPUT_LINE_NMI, attotime::zero);`.

src/mame/olivetti/att6300p.cpp
```diff
@@ -0,0 +1,1050 @@
...
+#include "att6300p_mmu.h"
+
+#include "bus/isa/isa.h"
+#include "bus/isa/isa_cards.h"
+#include "cpu/i86/i286.h"
+#include "cpu/mcs48/mcs48.h"
+#include "imagedev/floppy.h"
+#include "machine/am9517a.h"
+#include "machine/z80ctc.h"
+#include "m24_kbd.h"
```
startaq: This should be moved to the top since it's not a generic include.

startaq (commented):
Thank you for your contribution! I just added some notes about things I've noticed.

src/mame/olivetti/att6300p.cpp
donohoe00: Thanks for the code review!

These registers are not documented anywhere that I could find, and I wrote the descriptions based on my understanding from the schematics.  I thought it best to put it near the register bit field enums, but I could move it down to where the register reads/writes are implemented.

src/mame/olivetti/att6300p.cpp
donohoe00: This was inherited from m24.cpp.  Like the M24 and the original 6300, this machine was at least intended to be PC compatible :-)

The lack of MACHINE_SUPPORTS_SAVE also came from m24.cpp.  I'll do some testing and see if saving actually works.

src/mame/olivetti/att6300p.cpp
donohoe00: My understanding is that NMI is to remain asserted, and gets cleared when TRAPCE register is accessed.

src/mame/olivetti/att6300p.cpp
donohoe00: Seems to work fine.  I can push the change when the other issues are resolved.

donohoe00: Turns out I already pushed this change!

donohoe00:
Everything should be addressed now - is there anything else I need to do in order to get this accepted?

PR #14151, opened 2025-09-09
sony/news_38xx.cpp: Get NWS-3860 working

briceonk:
This machine is now working, albeit without graphics and sound. This is a really interesting machine because it is a dual-architecture one. #14150 should be merged first, otherwise the NEWS-OS installer won't work. As I mentioned in the TODO section, the only real weirdness I have seen so far is around the format program, everything else has been pretty solid in my testing.


src/devices/machine/upd765.cpp
```diff
@@ -3130,6 +3130,7 @@ n82077aa_device::n82077aa_device(const machine_config &mconfig, const char *tag,
 	ready_connected = false;
 	select_connected = true;
 	select_multiplexed = false;
+	recalibrate_steps = 80;
```
angelosa: This sneaked in from #14150 PR

angelosa:
> This machine is now working, albeit without graphics and sound.

That's a contradiction.

src/devices/machine/upd765.cpp
briceonk: Fixed

briceonk:
> > This machine is now working, albeit without graphics and sound.
> 
> That's a contradiction.

The machine can run perfectly fine without a graphics card or sound module installed using the serial console and network. That said, I can put the `MACHINE_NOT_WORKING` tag back on.

rb6502:
To clarify: you don't need the NOT_WORKING flag if the machine was sold with no video or audio and was usable that way.  (And NOT_WORKING isn't exactly the right flag for something that works, boots an OS, and runs programs anyway).  A lot of UNIX servers like the SGI Challenge series had no video or audio capability and were admin'd strictly over serial console.

src/mame/sony/news_38xx.cpp
```diff
@@ -36,51 +59,82 @@
...
+			  {INPUT_LINE_IRQ6, {iop_irq::TIMER}},
+			  {INPUT_LINE_IRQ7, {iop_irq::FDCDRQ, iop_irq::PERR}}
+		  })
+		  , cpu_irq_line_map({
+			  {INPUT_LINE_IRQ0, cpu_irq::UBUS},
+			  {INPUT_LINE_IRQ1, cpu_irq::IOP},
+			  {INPUT_LINE_IRQ2, cpu_irq::TIMER},
+			  {INPUT_LINE_IRQ4, cpu_irq::WRBERR},
+			  {INPUT_LINE_IRQ5, cpu_irq::PERR}
+		  })
```
cuavas: Please use one of the conventions already in use in MAME for formatting.

Either this:
```c++
	foo(...)
		: base(...)
		, m_x(...)
		, m_y(...)
```

Or this:
```c++
	foo(...) :
		base(...),
		m_x(...),
		m_y(...)
```

Not something slightly different again with mixed space/tab indentation.

src/mame/sony/news_38xx.cpp
```diff
@@ -36,51 +59,82 @@
...
 	{
 	}
 
 protected:
 	// driver_device overrides
-	virtual void machine_start() override ATTR_COLD;
-	virtual void machine_reset() override ATTR_COLD;
+	void machine_start() override ATTR_COLD;
+	void machine_reset() override ATTR_COLD;
```
cuavas: Same here – convention in MAME is to explicitly include the `virtual` qualifier even when it’s implied by `override`.

src/mame/sony/news_38xx.cpp
```diff
@@ -93,30 +147,106 @@ class news_38xx_state : public driver_device
...
+	};
+	static constexpr u32 iop_nmi_mask = 1 << static_cast<u32>(iop_irq::FDCDRQ) |
+										1 << static_cast<u32>(iop_irq::SCC) |
+										1 << static_cast<u32>(iop_irq::SCSI0) |
+										1 << static_cast<u32>(iop_irq::SCSI1) |
+										1 << static_cast<u32>(iop_irq::LANCE) |
+										1 << static_cast<u32>(iop_irq::FDCIRQ) |
+										1 << static_cast<u32>(iop_irq::SLOT) |
+										1 << static_cast<u32>(iop_irq::SOFTINTR) |
+										1 << static_cast<u32>(iop_irq::AST);
```
cuavas: You really don’t need the `static_cast` here – you can just use `u32(iop_irq::FDCDRQ)` and avoid a lot of clutter.

Also, remember we generally use `SCREAMING_SNAKE_CASE` for constants.

src/mame/sony/news_38xx.cpp
```diff
@@ -93,30 +147,106 @@ class news_38xx_state : public driver_device
...
 	};
-	template <irq_number Number> void irq_w(int state);
-	void int_check();
 
-	u32 bus_error_r();
-	void timer_w(u8 data);
-	void timer(s32 param);
+	static constexpr std::array cpu_irq_names = {"UBUS"sv, "IOP"sv, "TIMER"sv, "FPA"sv, "WRBERR"sv, "PERR"sv};
+	static constexpr u32 cpu_nmi_mask = 1 << static_cast<u32>(cpu_irq::WRBERR) |
+										1 << static_cast<u32>(cpu_irq::IOP);
```
cuavas: Here as well – the `static_cast` is just clutter.

Also, please group constants and types before member functions within a section.

src/mame/sony/news_38xx.cpp
```diff
@@ -130,199 +260,631 @@ class news_38xx_state : public driver_device
...
-	bool m_int_state[3]{};
-	bool m_scc_irq_state = 0;
+	// IOP IRQ state
+	const std::map<int, std::vector<iop_irq> > iop_irq_line_map;
+	bool m_scc_irq_state = false;
+	u32 m_iop_intst = 0;
+	u32 m_iop_inten = 0;
+
+	// CPU IRQ state
+	const std::map<int, cpu_irq> cpu_irq_line_map;
```
cuavas: Do you really want `std::map` for these?  It’s pretty heavy and doesn’t have good locality.  Have you considered using constant arrays and looking up elements with `std::lower_bound`?  It’s more efficient when the content won’t change.

src/mame/sony/news_38xx.cpp
```diff
@@ -130,199 +260,631 @@ class news_38xx_state : public driver_device
...
+
+u8 news_38xx_state::park_status_r()
+{
+	const u8 park_status = m_parallel_fault << 6 |
+						   static_cast<bool>(m_serial[0]->dsr_r()) << 5 |
+						   m_parallel_busy << 4 |
+						   !is_irq_set<iop_irq::PARALLEL>() << 3 |
+						   static_cast<bool>(m_serial[0]->ri_r()) << 2 |
+						   static_cast<bool>(m_serial[1]->dsr_r()) << 1 |
+						   static_cast<bool>(m_serial[1]->ri_r());
```
cuavas: There’s no point using `static_cast<bool>(foo)` when you can just use `bool(foo)` (and the same for other simple types) and it’s a lot more cluttered.  You only really need `static_cast` and friends when pointers are involved.

But that aside, some compilers give warnings for doing integer maths on `bool` due to the implicit conversion to `int`.  You can avoid it with `(m_serial[0]->dsr_r() ? 0x20 : 0x00)`, etc. or similar.

src/mame/sony/news_38xx.cpp
```diff
@@ -436,39 +1017,44 @@ void news_38xx_state::nws3860(machine_config &config)
...
 
-	PORT_DIPNAME(0x20000000, 0x20000000, "Diagnostic Mode") PORT_DIPLOCATION("SW1:6")
+	PORT_DIPNAME(0x20000000, 0x20000000, "Diagnostic Mode")
+	PORT_DIPLOCATION("SW1:6")
 	PORT_DIPSETTING(0x20000000, DEF_STR(Off))
 	PORT_DIPSETTING(0x00000000, DEF_STR(On))
 
-	PORT_DIPNAME(0xc0000000, 0xc0000000, "Unused") PORT_DIPLOCATION("SW1:7,8")
+	PORT_DIPNAME(0xc0000000, 0xc0000000, "Unused")
+	PORT_DIPLOCATION("SW1:7,8")
```
cuavas: As with the other stuff, the rest of MAME puts the `PORT_DIPLOCATION` on the same line as the `PORT_DIPNAME`, and we generally tabulate it so the option values and names align under the default value and setting name, e.g.
```c++
	PORT_DIPNAME(0x07000000, 0x07000000, "Display") PORT_DIPLOCATION("SW1:1,2,3")
	PORT_DIPSETTING(         0x07000000, "Console")
	PORT_DIPSETTING(         0x06000000, "NWB-512")
	PORT_DIPSETTING(         0x03000000, "NWB-225A")
	PORT_DIPSETTING(         0x00000000, "Autoselect")
```

briceonk:
> To clarify: you don't need the NOT_WORKING flag if the machine was sold with no video or audio and was usable that way. (And NOT_WORKING isn't exactly the right flag for something that works, boots an OS, and runs programs anyway). A lot of UNIX servers like the SGI Challenge series had no video or audio capability and were admin'd strictly over serial console.

No graphics is a valid config for this machine. The software and firmware can handle missing sound (for example, on the NWS-5000X it is a separate board so it can be easily disconnected and thus can handle that case), but I think that, at least for the 3800 series, it was always included. I'll leave it in place for now, but can remove it too if that is the general consensus.

src/mame/sony/news_38xx.cpp
briceonk: I ran an autoformatter followed by srcclean and that messed it up. Will update to use one of those conventions.

src/mame/sony/news_38xx.cpp
briceonk: Sorry - was using clang-tidy and it complained about that. Updated to maintain style.

src/mame/sony/news_38xx.cpp
briceonk: Updated to match `news_68k_iop.cpp` and `news_r4k.cpp`

src/mame/sony/news_38xx.cpp
briceonk: Using `static_cast` instead of `(u32)` is a habit I picked up at work thanks to a static code analyzer that prefers avoiding C-style casts in C++

Fixed :)

src/mame/sony/news_38xx.cpp
briceonk: Fixed

src/mame/sony/news_38xx.cpp
briceonk: Sorry, this was a miss on my part after running the autoformatter. I reverted this section since I didn't change anything in it.

src/mame/sony/news_38xx.cpp
briceonk: Moved types and constants above the rest of the member functions.

src/mame/sony/news_38xx.cpp
briceonk: I will do so. I used map when I was developing because it made it easy to write/debug to my brain, but now is a good time to make it faster :)

briceonk: Since I had to iterate through all the items every time in `irq_check_iop`/`irq_check_cpu`, I wasn't 100% sure where `std::lower_bound` came into the picture, so I just went ahead and used `constexpr` arrays with a basic lookup for CPU, and `constexpr` arrays with a mask for IOP. 
I apologize for missing what you were alluding to, but if there is a specific algorithm you suggest for checking the active IRQs, I'd be happy to make any adjustments :)

briceonk:
Just for fun, here is a screenshot from when I was doing final testing after addressing code review feedback :)
image

Issue #14113, opened 2025-09-02
Artwork visual stack order doesn’t match hit test priority


cuavas:
This has been an issue ever since clickable artwork support was added.  Hit testing prioritises the first clickable element, which appears at the bottom of the visual stack.  This is confusing, and has the obvious effect of making piano keyboards a more difficult than it needs to be.  For example you can see @felipesanches getting annoyed about it in #14049 here: https://github.com/mamedev/mame/blob/02bdea66ea92f9bd5d0a18547dadafe8cdc3b5b7/src/mame/layout/esq1.lay#L495

Reversing hit test priority to match visual stacking is trivially easy – a one-line source change will do it.  However, there’s stuff that depends on the current semantics. Examples I can think of include:
* Piano keyboards, although fixing it will allow them to be simplified.
* Things punching holes in clickable elements or completely occluding them.  However at least the vast majority of these can be found easily because they use `clickthrough="no"` attributes on the occluding elements.

We can coordinate updating internal artwork, but there’s a risk of breaking external artwork as well.  As far as I know, external artwork for MAME’s traditional “arcade games” doesn’t generally depend on it.

So my question is, do others think it would be viable to change the semantics to make hit testing priority match visual stack order, or is there too much risk that of external artwork that depends on it, or internal artwork that depends on it that we wouldn’t catch?

galibert:
Artworks are entering a new era with all the interactive controls on them, it's better to fix those things early rather than enshrining unintuitive decisions.  We'll catch everything eventually, superpositions are rare anyway.

cuavas:
Fine, I’ll flip it.  We’re early in the cycle, let’s rip the bandaid off.  Can you take care of some of the piano keyboards?  Maybe @felipesanches and @devinacker can help.

galibert:
Sure.

devinacker:
Sure, I can go over my keyboard layouts whenever the change happens.

cuavas:
I made the code change and updated a lot of the affected layouts.  The following layouts will need someone else to look at them:

* All piano keyboards besides layout/esq1.lay which I already did
* layout/autorace.lay
* layout/linn_linndrum.lay
* layout/novag_supercon.lay
* layout/roland_tr707.lay
* layout/saitek_mark5.lay
* layout/saitek_mark6.lay

@m1macrophage this affects at least a couple of your layouts in case you want to look at them.

m1macrophage:
Thanks, I'll take a look.

[...]

PR #14166, opened 2025-09-12
update element order for various casio and yamaha keyboard layouts

devinacker:
(#14113)

This should be all of mine. I don't know offhand how many others need doing (aside from a few in the yamaha dir)


cuavas:
Thanks.

PR #13002, opened 2024-11-20
add new dump marywu.cpp

flama12333:
MARY-1/SUNRISE
diferent  version set but same music of marywu
32kb - duplicate indentical
dump is available in https://chamberplus.blogspot.com/2024/11/bar.html?sc=1732100220189.
dumped by chamberplus

also set Set inputs IP_ACTIVE_HIGH to IP_ACTIVE_LOW Except some
add P1 in to bypass error 30 marywu.cpp

unknown win way elec corp
dumped by stonedDiscord

PR #14168, opened 2025-09-13
Set inputs IP_ACTIVE_HIGH to IP_ACTIVE_LOW marywu.cpp

flama12333:
This will prevent to stuck at 1 post. error 30 bypassed by adding p1 in without function but will cause error 02. Update todo regarding add something.
I have dump of chamberplus that has different rom program version. see also
https://github.com/mamedev/mame/pull/13002


flama12333:
Closed due to duplicated

PR #14172, opened 2025-09-14
konami/k005885.cpp: Device-fied K005885, Add notes

cam900:
- Sprite emulation is still in each driver, because per-driver sprite quirks are remains (sprite RAM bank, draw list length, sprite size, etc)
- konami/ironhors.cpp: Fix notes, Use generic gfxdecode layout
- konami/ddribble.cpp, konami/finalizr.cpp, konami/ironhors.cpp, konami/jackal.cpp: Fix naming, Use BIT macro for single bit value

### DIFF
--- a/src/mame/konami/k005885.h
+++ b/src/mame/konami/k005885.h
@@ -12,6 +12,7 @@
 class k005885_device : public device_t, public device_video_interface, public device_gfx_interface
 {
 public:
+	using sprite_delegate = device_delegate<void (screen_device &screen, bitmap_ind16 &bitmap, const rectangle &cliprect, int size, bool flip, gfx_element *sgfx, int &code, int &color, int &sx, int &sy, bool flipx, bool flipy)>;
 	using tile_delegate = device_delegate<void (int layer, int attr, int &gfx, int &code, int &color, int &flags, int codebank)>;
 
 	k005885_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
@@ -29,11 +30,10 @@ public:
 	auto set_firq_cb() { return m_firq_cb.bind(); }
 	auto set_nmi_cb() { return m_nmi_cb.bind(); }
 	auto set_flipscreen_cb() { return m_flipscreen_cb.bind(); }
+	template <typename... T> void set_sprite_callback(T &&... args) { m_sprite_cb.set(std::forward<T>(args)...); }
 	template <typename... T> void set_tile_callback(T &&... args) { m_tile_cb.set(std::forward<T>(args)...); }
 
 	bool flipscreen() { return m_flipscreen; }
-	u8 ctrl_r(offs_t offset) { return (offset >= 5) ? 0 : m_ctrlram[offset & 7]; } // not from addressmap
-	void ctrl_w(offs_t offset, u8 data);
 
 	// scroll RAM
 	u8 scroll_r(offs_t offset) { return m_scrollram[offset & 0x3f]; }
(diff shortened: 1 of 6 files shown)


src/mame/konami/k005885.h
```diff
@@ -0,0 +1,106 @@
...
+	u8 scroll_r(offs_t offset) { return m_scrollram[offset & 0x3f]; }
+	void scroll_w(offs_t offset, u8 data) { m_scrollram[offset & 0x3f] = data; }
+
+	// tilemap RAM
+	u8 vram_r(offs_t offset) { return m_vram[offset]; }
+	void vram_w(offs_t offset, u8 data);
+
+	// sprite RAM
+	u8 spriteram_r(offs_t offset) { return m_spriteram[offset]; }
+	void spriteram_w(offs_t offset, u8 data) { m_spriteram[offset] = data; }
```
cuavas: Can you use device maps and possibly avoid a layer of function calls for these?

cuavas (commented):
I find it a bit concerning that despite merging functionality duplicated in four systems to a common location, this ends up adding about 20% more lines than it removes (including other general cleanup), and also the sprite drawing is still implemented in all the system drivers.

How much is actually being shared here?  If the sprite functionality is actually provided by the K005885 chip, what’s preventing it from being shared as well?

cam900:
> I find it a bit concerning that despite merging functionality duplicated in four systems to a common location, this ends up adding about 20% more lines than it removes (including other general cleanup), and also the sprite drawing is still implemented in all the system drivers.
> 
> How much is actually being shared here? If the sprite functionality is actually provided by the K005885 chip, what’s preventing it from being shared as well?

1. Tilemap registers and structure(2x32x32 width or 1x64x32 width tilemap - controlled by register?) are shared, also sprite RAM bank and flipscreen bits too.
2. I think sprite size(some has 16x16 minimum sprite size, some has 8x8), structure, code, color index and x position(some has 8 bit x position and some has 9 bit signed.) is depended by hardware.
3. Following to 2., code and color index are available to modify through callback in common sprite draw routine.

src/mame/konami/k005885.h
cam900: It will be, but some hardware allowed to reads these registers.

galibert:
Is there a schematic available with a 005885 on it?

cam900:
> Is there a schematic available with a 005885 on it?

https://arcarc.xmission.com/PDF_Arcade_Manuals_and_Schematics/Iron%20Horse.pdf has schematics for ironhors, last page mentioned 005885.

https://arcarc.xmission.com/PDF_Arcade_Manuals_and_Schematics/Double%20Dribble%20Schematics.pdf is for ddribble.

galibert:
Thanks, interesting chip.

It has a bus for graphics roms that's 16bits\*65536, where the games (at least ddribble) do banking to fit more.  It has 8bits\*65536 dram connected called "F", and a 8bits\*8192 of sram called "X" and "V". Those must include tilemaps and sprite lists, but the dram makes me wonder what it's used for.  There is also two ports from proms, one that gives 4 bits of VCB and VCF and expects 4 bits of VCD, and one with OCB and OCF and expects OCD.  Iron Horse has proms, ddribble directly connects B to D.  I wonder if the different attributes interpretations are due to that.  What is sure is that we don't seem to be using those proms in Iron Horse (but we have them).  The color output is 5 bits.

If we want to implement a common version we need to take the proms and the address banking into account I guess.  And it means understanding the prom ports, what they do.

PR #14174, opened 2025-09-14
alesis/midiverb.cpp: Adding the MIDIFEX and its layout.

m1macrophage:
New working systems
-----------------
Alesis MIDIFEX [*]

[*] I stumbled upon this [here](https://synth-diy.org/pipermail/synth-diy/2025-July/175631.html). Up to the MAME team on how you'd like to handle credits.

### DIFF
--- a/src/mame/alesis/midiverb.cpp
+++ b/src/mame/alesis/midiverb.cpp
@@ -4,6 +4,10 @@
 /*
 The MIDIverb is a digital delay & reverb unit.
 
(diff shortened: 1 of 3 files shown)


m1macrophage:
> [*] I stumbled upon this [here](https://synth-diy.org/pipermail/synth-diy/2025-July/175631.html). Up to the MAME team on how you'd like to handle credits.

While there is no urgency on the PR itself, the linked message implies that availability will be temporary. Mentioning it in case someone needs to validate the checksums.

PR #14176, opened 2025-09-15
mystwarr.cpp: Fix booting up with error for mmaulersu [sjy96525]

sjy555525:


sjy555525:
I used the same as mmaulers.nv booting up with error.


I used the new mmaulersu.nv booting up without error, so now playable for Monster Maulers (ver UAA).

cuavas:
The current default NVRAM does in fact trap it in a boot loop, and the SHA1 of after initialisation matches this.

Issue #14142, opened 2025-09-08
Multiple invalidations of a layout element don't work on most graphic backends


enikland3:
### MAME version

0.280

### System information

Ubuntu 24.04.3 LTS
Linux kernel 6.8.0-79-generic, x86_64
Mesa 25.0.7-0ubuntu0.24.04.2
AMD Radeon RX 550 / 550 Series (radeonsi, polaris12, ACO, DRM 3.57, 6.8.0-79-generic)
Video memory: 4096MB

### INI configuration details

```INI
default "./mame -cc", only rompath and samplepath adjusted
```

### Emulated system/software

Sega Master System, but not system specific

### Incorrect behaviour

Using the layout configuration below as an example:

----------------------------------------
<?xml version="1.0"?>

<mamelayout version="2">

	<element name="overlay" defstate="0">
		<rect state="0"><color red="1.0" green="1.0" blue="0" /></rect>
	</element>

	<view name="TESTLAYOUT">
		<screen index="0">
			<bounds x="0" y="0" width="4" height="3" />
		</screen>
		<element name="overlay" id="overlay" ref="overlay" blend="multiply">
			<bounds x="0" y="0" width="4" height="3" />
		</element>
	</view>

	<script><![CDATA[
		file:set_resolve_tags_callback(
				function ()
					local scr = machine.screens["screen"]
					local overlay = file.elements['overlay']
					local blendbitmap = emu.bitmap_argb32(scr.width, scr.height)
					local color = 0xff0000ff

					local function draw_overlay(state, bitmap)
						blendbitmap:resample(bitmap)
						print("frame " .. scr:frame_number() .. " blend color " .. string.format("0x%08x", color))
					end

					function prepare_items()
						blendbitmap:fill(color)
						color = color + 1
						overlay:invalidate()
						overlay:set_draw_callback(draw_overlay)
					end

					file.views['TESTLAYOUT']:set_prepare_items_callback(prepare_items)
				end)

	]]></script>
</mamelayout>
----------------------------------------

Running as:

mame -video opengl -verbose -plugins -plugin layout -frameskip 0 -window -nomaximize <driver>

The desired effect is get the system video colors blended with the color set by the value of the variable named 'color' of the layout script. For each frame the script increments the value of the 'color' variable, what should change the levels of the final output colors each time, what is not happening with most video backends.

Doesn't work as desired, blend color is changed once:
	-video opengl
	-renderdriver opengles2

Doesn't work as desired, blend color is changed from time to time:
[...]

Issue #11279, opened 2023-05-26
[sms1pal] wrongly reports 3 screens in -listxml


aretrogamer:
Hi all,

If you look at xml tags in -listxml of sms1pal, there are 2 mysterious LCD screens listed.
My question: how can a master system I have 2 LCD screens, here named left_lcd and right_lcd while it's output goes to a TV which is a raster display (see first display tag)?

`<display tag="screen" type="raster" rotate="0" width="268" height="240" refresh="49.701460" pixclock="5320342" htotal="342" hbend="61" hbstart="329" vtotal="313" vbend="46" vbstart="286" />
		<display tag="left_lcd" type="lcd" rotate="0" width="268" height="240" refresh="49.701460" pixclock="5320342" htotal="342" hbend="61" hbstart="329" vtotal="313" vbend="46" vbstart="286" />
		<display tag="right_lcd" type="lcd" rotate="0" width="268" height="240" refresh="49.701460" pixclock="5320342" htotal="342" hbend="61" hbstart="329" vtotal="313" vbend="46" vbstart="286" />`

the parent rom sms does NOT have those screens;

`<display tag="screen" type="raster" rotate="0" width="268" height="224" refresh="59.922743" pixclock="5369317" htotal="342" hbend="61" hbstart="329" vtotal="262" vbend="27" vbstart="251" />`

Can somebody explain this mystery?

FoxhackDN:
... isn't that for the Sega 3D Glasses?

aretrogamer:
That could make sense, but i'm not familiar with that hardware.
Then the question remains; why is the parent rom sms not having those 2 LCD screens?
Are there different master system I consoles made, one supporting the Sega 3D Glasses and the other not?

FoxhackDN:
Yeah. The Master System II had no card slot, so it didn't support the LCD glasses.

aretrogamer:
> Yeah. The Master System II had no card slot, so it didn't support the LCD glasses.

I'm not talking about the Master System II but the Master System I (sms). If you look in the XML, there the 2 LCD screens are missing...

`<display tag="screen" type="raster" rotate="0" width="268" height="224" refresh="59.922743" pixclock="5369317" htotal="342" hbend="61" hbstart="329" vtotal="262" vbend="27" vbstart="251" />`

So my question is: why is only the clone sms1pal having those LCD screens and not the parent sms?

aretrogamer:
Is the 3D glasses emulated in mame?
[...]

Issue #3492, opened 2018-04-27
Sega Master System SegaScope 3D on 3DTV


darkfalz79:
Hi there. I made this layout addition to sms1.lay to display SegaScope 3D on a Half-SBS mode suitable for 3DTVs. You obviously must enable the SegaScope in the configuration as well as binocular hack
to "both". And switch your TV into "Side by side" mode.

Works perfectly on my Sony HDTV - first time I am ever playing these SMS games in 3D. Unfortunately does not work on 3D Vision (maybe via 3DMigoto?) since you can't force those into a Half-SBS mode.

I would love it if you added it to the official SMS layout.

`
<?xml version="1.0"?>
<mamelayout version="2">
	<view name="SegaScope 3-D Glasses (3DTV H-SBS)">
		<screen index="1">
			<bounds x="2" y="0" width="6" height="9" />
		</screen>
		<screen index="2">
			<bounds x="10" y="0" width="6" height="9" />
		</screen>
	</view>
</mamelayout>
`

darkfalz79:
![1](https://user-images.githubusercontent.com/38789695/39370798-30dbd61a-4a83-11e8-99db-b66effeeabd1.png)
![2](https://user-images.githubusercontent.com/38789695/39370806-35130122-4a83-11e8-9e96-62bb67e5048a.png)

Taken in ddraw mode, sorry I couldn't get rendered screenshot working in this mode.

happppp:
This is more suitable for external artwork, I suggest you to get in touch with Mr Do: http://mrdo.mameworld.info/mame_artwork.php

Also, the SMS driver screen configuration for 3D glasses is implemented very wrong. It creates 3 physical screens, while in reality they're electronic shutters(for a better example, see tceptor2.lay). When the MAME sms driver is updated in the future to reflect this, your layout file will need to be updated as well.

darkfalz79:
I'm aware the layout relies on a kludge that already exists in the driver
(to output Sega Scope 3D as steady left/right images rather than
alternating shutters) but I figured since the layout already contains a
side by side mode I might as well offer a tweaked version to actually be
able to play these games in 3D on a compatible set.

I've been wanting to try these games since the system was released in the
80s and to finally be able to do it via MAME and the custom layout achieved
this long dream. Just a shame the library is so small and only a couple of
then have a well implemented 3D effect.

Cheers!

danielcamposramos:
[...]

PR #14184, opened 2025-09-17
Move switching of screen's current bitmap to after layout updates

enikland3:
Currently video_manager::frame_update() updates the screen object by calling finish_screen_updates(), that calls screen.update_quads(), where the current m_curbitmap is set as the bitmap to be used by the frame texture. The code then switches the m_curbitmap pointer to the other bitmap available to blit the next frame. The execution returns to after the last call was made in video_manager::frame_update() and continues through the code. One of the remaining calls is machine().osd().update(), where the layout is rendered. If the layout contains a script that calls screen:pixels() to obtain the data of pixels in the frame, it won't get the data of the last frame. To make this work, these changes move switching of screen's current bitmap to after the layout is updated.

### DIFF
--- a/src/emu/screen.cpp
+++ b/src/emu/screen.cpp
@@ -1779,7 +1779,6 @@ bool screen_device::update_quads()
 				}
 				m_texture[m_curbitmap]->set_bitmap(m_bitmap[m_curbitmap], m_visarea, m_bitmap[m_curbitmap].texformat());
 				m_curtexture = m_curbitmap;
-				m_curbitmap = 1 - m_curbitmap;
 			}
 
 			// brightness adjusted render color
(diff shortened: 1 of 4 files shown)


galibert:
It sounds like a very, very bad idea to call screen:pixels() in a layout in the first place.  What do you need that for?

enikland3:
Hi OG, thanks for the reply. This is one of three changes I made locally to allow to properly port the hack that displays Sega Master System 3D games without flicker from the driver code to a layout script. The hack is a nice workaround that allows to play 3D games without headache or vision impairment and allows to use a modern 3D display (or glasses) that converts side-by-side image into 3D, as a [user have already take advantage of that](https://github.com/mamedev/mame/issues/3492).

Natively the game displays the image intended for the left eye while controls the shutter glasses to block the right eye's view, then displays the right-eye image while blocks the left eye, in a cycle that keeps repeating while the game is running. This is simulated in MAME using a layout that has at least two views: a view for the standard TV screen and a view for the 3D glasses, that displays two images side-by-side, for each lenses, but because the game only displays the image for the opened lenses each time, it is displayed at a side and the other side shows a black color.

The current implementation of the hack in the driver uses 2 additional screens (left_lcd and right_lcd) that are used for each side of the layout view for the 3D glasses. When an image is drawn at a screen, a temporary bitmap keeps it and draws it again at the same screen in the next frame, when the game draws the other screen. The result is a simultaneous display of the images for the two lenses, why I called it the binocular hack. The hack works very well, but @happppp says [it is very wrong](https://github.com/mamedev/mame/issues/3492#issuecomment-387380983) and it was already questioned by a user (issue #11279 - deleted - subject:"[sms1pal] wrongly reports 3 screens in -listxml"). They have a good reason: it does not documents the hardware properly, one of the main goals of MAME. Moving the hack to the layout script would solve this problem.

The layout script version of the hack makes use of the screen:pixels() function to copy the frame to a bitmap, the element:invalidate() and element:set_draw_callback() functions to create a custom callback to drawn the rectange that would occlude the side of the blocked view, and the bitmap:resample() function to drawn the copied frame into that rectangle.

This change alone is not enough to move the hack to a layout script. I also need a solution for issue #14142 and I have one more change waiting that implements a function in luaengine_render.cpp to draw the output of screen:pixels() to a bitmap much faster than doing in pure Lua code. Performance is still halved when compared to the hack in the driver but framerate is still hundreds of FPS with modern PCs.

PR #14186, opened 2025-09-17
cpu/i960: store IRQ line states, immediately return if unchanged

gm-matthew:
If a particular IRQ line is already high, trying to assert another IRQ on the same line no longer has any effect; the IRQ line must be cleared/lowered first before another IRQ can be asserted. 

IRQ acknowledgement logic has been implemented for Namco System FL, since it is now required with the i960 update.

### DIFF
--- a/src/devices/cpu/i960/i960.cpp
+++ b/src/devices/cpu/i960/i960.cpp
@@ -4,6 +4,8 @@
 #include "i960.h"
 #include "i960dis.h"
 
+#include <algorithm>
+
 #ifdef _MSC_VER
 /* logb prototype is different for MS Visual C */
 #include <cfloat>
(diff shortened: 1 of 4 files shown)


src/devices/cpu/i960/i960.cpp
```diff
@@ -2396,6 +2402,8 @@ void i960_cpu_device::device_reset()
 	m_r[I960_FP] = m_program.read_dword(m_PRCB+24);
 	m_r[I960_SP] = m_r[I960_FP] + 64;
 	m_rcache_pos = 0;
+
+	std::fill(std::begin(m_irq_line_state), std::end(m_irq_line_state), 0);
```
happppp: Is this the physical pin state? Those don't get cleared by the chip at reset. And if they do, it's the surrounding hw that clears the pins.

It's good to clear them when MAME starts though, the correct place is device_start.

gm-matthew: I believe they do indeed represent the physical pin states. I had more or less copied the approach used by sh2.cpp but your point about the surrounding hardware clearing the pins makes sense.

src/mame/namco/namcofl.cpp
```diff
@@ -650,6 +656,12 @@ void namcofl_state::machine_reset()
 
 	std::fill_n(&m_workram[0], m_workram.bytes() / 4, 0);
 	m_mainbank.select(1);
+
+	const int irq_type[4] = { I960_IRQ0, I960_IRQ1, I960_IRQ2, I960_IRQ3 };
+	for (int i = 0; i < 4; i++)
+	{
+		m_maincpu->set_input_line(irq_type[i], CLEAR_LINE);
+	}
```
cuavas: Why not move your array of interrupt numbers to an anonymous namespace and make it `constexpr` so it doesn’t need to be repeated?
Why not use a rang-based `for` loop here to avoid pitfalls, like
```c++
	for (auto line : irqs)
		m_maincpu->set_input_line(line, CLEAR_LINE);
```

src/mame/sega/model2.cpp
```diff
@@ -269,6 +269,12 @@ void model2_state::machine_reset()
 	m_copro_fifo_out->clear();
 	m_geo_write_start_address = 0;
 	m_geo_read_start_address = 0;
+
+	const int irq_type[4] = { I960_IRQ0, I960_IRQ1, I960_IRQ2, I960_IRQ3 };
+	for (int i = 0; i < 4; i++)
+	{
+		m_maincpu->set_input_line(irq_type[i], CLEAR_LINE);
+	}
```
cuavas: This is a case for a range-based `for` loop as well.

angelosa:
Is it known if the i960 really has this internally rather than having external circuitry doing the job? i.e. is there a page from manual stating so?

cuavas:
> Is it known if the i960 really has this internally rather than having external circuitry doing the job? i.e. is there a page from manual stating so?

The code as it stands is just a buggy implementation of edge-triggered interrupts.  It’s wrong either way.  If the chip inherently has edge-triggered interrupts, then this is the correct fix, but it it doesn’t, it should still only be checking for changes, but should clear the condition when it gets the clear line state.

As far as I can tell, the 80960 core has level-sensitive interrupts.  You need a separate interrupt controller to implement edge-triggered interrupts.

However the i960 Jx has an integrated interrupt controller that can be configured for level-sensitive or edge-triggered interrupts.  It also has eight interrupt pins, and allows them to be treated as independent interrupt inputs, a bitfield allowing one of 240 interrupt interrupts to be signaled (68k style but with eight pins instead of three), or a hybrid mode with three independent interrupts and five pins used as a bitfield for signaling one of 32 interrupts.

What is MAME actually supposed to be emulating here?

gm-matthew:
> The code as it stands is just a buggy implementation of edge-triggered interrupts. It’s wrong either way. If the chip inherently has edge-triggered interrupts, then this is the correct fix, but it it doesn’t, it should still only be checking for changes, but should clear the condition when it gets the clear line state.
> 
> As far as I can tell, the 80960 core has level-sensitive interrupts. You need a separate interrupt controller to implement edge-triggered interrupts.
> 
> However the i960 Jx has an integrated interrupt controller that can be configured for level-sensitive or edge-triggered interrupts. It also has eight interrupt pins, and allows them to be treated as independent interrupt inputs, a bitfield allowing one of 240 interrupt interrupts to be signaled (68k style but with eight pins instead of three), or a hybrid mode with three independent interrupts and five pins used as a bitfield for signaling one of 32 interrupts.
> 
> What is MAME actually supposed to be emulating here?

The datasheets for the [MC](http://www.bitsavers.org/components/intel/i960/271080-006_80960MC_Advance_Information_Jan91.pdf), [SA, SB, KA and KB](http://www.bitsavers.org/components/intel/i960/i960_Processors_and_Related_Products_Jan95.pdf) variants of the i960 all specify in the pin description that in order to signal an interrupt, an interrupt pin must be deasserted for at least one bus cycle and then asserted for at least one additional bus cycle, which sounds to me like interrupts are edge-triggered for these variants.

If desired I can make it so that `execute_set_input()` only updates the interrupt line states and move the rest of the code over to `check_irqs()`.

> Is it known if the i960 really has this internally rather than having external circuitry doing the job? i.e. is there a page from manual stating so?

As the code currently stands, calling `set_input_line(IRQ, ASSERT_LINE)` more than once without clearing the line between interrupts still leads to multiple interrupts being fired which is impossible on real hardware; you can't raise the interrupt line to assert an IRQ if the interrupt line is already high.

src/mame/namco/namcofl.cpp
gm-matthew: I have moved the array of IRQs to the class body and made it a `static constexpr`. I have also made it so that a range-based for loop is used when clearing the IRQ lines during reset.

src/mame/sega/model2.cpp
gm-matthew: As above, I have implemented a range-based for loop.

ghost:
FYI the chip version on Model 2 is NG80960KB-25

PR #14189, opened 2025-09-19
sinclair/sprinter.cpp: Added access to internal z84 IO through the BUS

holub:


holub:
@ajrhacker do you mind to merge #14165 as well, we'll have the full house in same release then?

src/mame/sinclair/sprinter.cpp
```diff
@@ -1429,6 +1429,19 @@ void sprinter_state::init_taps()
...
+
+	m_maincpu->space(AS_IO).install_write_tap(0x0000, 0xffff, "cpu_io_w", [this](offs_t offset, u8 &data, u8 mem_mask)
+	{
+		// Internal z84 ports are not accesable through IO map, hence they need special case here
+		static const std::unordered_set<u8> z84_int = {
+			0x10, 0x11, 0x12, 0x13,
+			0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
+			0xee, 0xef,
+			0xf0, 0xf1, 0xf4
+		};
```
cuavas, line 1441: Eugh, can we not have dynamically allocated static things when it isn't necessary?  You can just use an ordered array, and then use `lower_bound` on it.  It's more efficient and avoids a DCLP in a hot function.

PR #14181, opened 2025-09-16
[tmp94c241] dasm: SFR symbol names

felipesanches:

PR #14190, opened 2025-09-19
[tlcs900 disasm] SFR symbol names for all other TLCS-900 variants:

felipesanches:
TMP95C061, TMP95C063 and TMP96C141

(follow-up to PR #14181)

### DIFF
--- a/src/devices/cpu/tlcs900/dasm900.h
+++ b/src/devices/cpu/tlcs900/dasm900.h
@@ -132,21 +132,33 @@ private:
 class tmp95c061_disassembler : public tlcs900_disassembler
 {
 public:
-	tmp95c061_disassembler() : tlcs900_disassembler() {};
+	tmp95c061_disassembler(std::pair<u16, char const *> const symbols[], std::size_t symbol_count);
+	template<size_t N> tmp95c061_disassembler(std::pair<u16, char const *> const (&symbols)[N]) : tlcs900_disassembler(symbols, N) {}
+
+private:
+	static const char *const s_sfr_names[];
 };
 
 
 class tmp95c063_disassembler : public tlcs900_disassembler
 {
 public:
-	tmp95c063_disassembler() : tlcs900_disassembler() {};
+	tmp95c063_disassembler(std::pair<u16, char const *> const symbols[], std::size_t symbol_count);
+	template<size_t N> tmp95c063_disassembler(std::pair<u16, char const *> const (&symbols)[N]) : tlcs900_disassembler(symbols, N) {}
+
+private:
+	static const char *const s_sfr_names[];
 };
 
 
 class tmp96c141_disassembler : public tlcs900_disassembler
 {
 public:
-	tmp96c141_disassembler() : tlcs900_disassembler() {};
+	tmp96c141_disassembler(std::pair<u16, char const *> const symbols[], std::size_t symbol_count);
+	template<size_t N> tmp96c141_disassembler(std::pair<u16, char const *> const (&symbols)[N]) : tlcs900_disassembler(symbols, N) {}
+
+private:
+	static const char *const s_sfr_names[];
 };
 
 #endif
--- a/src/devices/cpu/tlcs900/tmp95c061.cpp
+++ b/src/devices/cpu/tlcs900/tmp95c061.cpp
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:Wilbert Pol
+// copyright-holders:Wilbert Pol, Felipe Sanches
 /*******************************************************************
 
 Toshiba TMP95C061 emulation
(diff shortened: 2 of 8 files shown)


felipesanches:
This is ready for review.

src/devices/cpu/tlcs900/dasm900.h
```diff
@@ -132,21 +132,33 @@ class tmp94c241_disassembler : public tlcs900_disassembler
...
 
 class tmp96c141_disassembler : public tlcs900_disassembler
 {
 public:
-	tmp96c141_disassembler() : tlcs900_disassembler() {};
+	tmp96c141_disassembler(std::pair<u16, char const *> const symbols[], std::size_t symbol_count);
+	template<size_t N> tmp96c141_disassembler(std::pair<u16, char const *> const (&symbols)[N]) : tlcs900_disassembler(symbols, N) {}
+
+private:
+	static const char *const s_sfr_names[];
```
cuavas: Are the `s_sfr_names` things supposed to be here?  They don't seem to be defined.  Also you've now got an undefined constructor in each of these classes.  Do these classes actually serve any purpose, since they don't override anything and don't construct the base class differently?

src/devices/cpu/tlcs900/tmp95c061.cpp
```diff
@@ -1353,3 +1354,38 @@ void tmp95c061_device::dmemcr_w(uint8_t data)
...
+	{ 0x5e, "MSAR3" }, { 0x5f, "MAMR3" }, { 0x60, "ADREG0L" }, { 0x61, "ADREG0H" },
+	{ 0x62, "ADREG1L" }, { 0x63, "ADREG1H" }, { 0x64, "ADREG2L" }, { 0x65, "ADREG2H" },
+	{ 0x66, "ADREG3L" }, { 0x67, "ADREG3H" }, { 0x68, "B0CS" }, { 0x69, "B1CS" },
+	{ 0x6a, "B2CS" }, { 0x6b, "B3CS" }, { 0x6c, "BEXCS" }, { 0x6d, "ADMOD" },
+	{ 0x6e, "WDMOD" }, { 0x6f, "WDCR" }, { 0x70, "INTE0AD" }, { 0x71, "INTE45" },
+	{ 0x72, "INTE67" }, { 0x73, "INTET10" }, { 0x74, "INTET32" }, { 0x75, "INTET54" },
+	{ 0x76, "INTET76" }, { 0x77, "INTES0" }, { 0x78, "INTES1" }, { 0x79, "INTETC01" },
+	{ 0x7a, "INTETC23" }, { 0x7b, "IIMC" }, { 0x7c, "DMA0V" }, { 0x7d, "DMA1V" },
+	{ 0x7e, "DMA2V" }, { 0x7f, "DMA3V" }
+};
```
cuavas: You should put this in an anonymous namespace at the top of the file to avoid polluting the global symbol table.  Same for the other ones.

felipesanches:
@cuavas, thanks for the review. Yes, I agree with the concerns you raised. I restructured the code to try to address these points.

I also took the opportunity to add to unidasm these specialized disassemblers of TLCS900 variants.

felipesanches:
Bleep-bloop boing!

src/devices/cpu/tlcs900/dasm900.h
```diff
@@ -121,44 +122,179 @@ class tlcs900_disassembler : public util::disasm_interface
 class tmp94c241_disassembler : public tlcs900_disassembler
 {
 public:
-	tmp94c241_disassembler(std::pair<u16, char const *> const symbols[], std::size_t symbol_count);
-	template<size_t N> tmp94c241_disassembler(std::pair<u16, char const *> const (&symbols)[N]) : tlcs900_disassembler(symbols, N) {}
+	tmp94c241_disassembler() : tlcs900_disassembler(m_sfr_names) { };
 
 private:
-	static const char *const s_sfr_names[];
+	std::pair<u16, char const *> const m_sfr_names[185] = {
```
galibert: They should be static, otherwise they get reconstructed each time a dissassembler is intanciated.

felipesanches: Would this be OK?
image

felipesanches:
@cuavas, I saw your commit and comment today at https://github.com/mamedev/mame/commit/b8b0725074d624929b60b35f23a2aab8b36421ca#r166510006

You could have made that comment here and I would have addressed it swiftly.

felipesanches:
Rebased and fixed conflicts. This is ready for review.

felipesanches:
This one should now be trivial to merge because in it I have used for all other CPU variants the exact same code-style as @cuavas used when he refactored my earlier tmp94c241 SFR symbols code.

cuavas:
> @cuavas, I saw your commit and comment today at [b8b0725#r166510006](https://github.com/mamedev/mame/commit/b8b0725074d624929b60b35f23a2aab8b36421ca#r166510006)
> 
> You could have made that comment here and I would have addressed it swiftly.

I’m still rather irritated that a crashing disassembler made it into a release.  It crashes 100% of the time in debug builds (we intentionally pre-fill allocated memory with a pattern that makes it more likely to expose these issues), it was flagged by Coverity, and it’s a really obvious issue to see in the code.  I feel like I have egg on my face.  It’s embarrassing trying to write a release announcement when I know something like this is there.

felipesanches:
thanks!

PR #14193, opened 2025-09-20
segas32: fix NBG0 - NBG3 layer flip

mahoneyt944:
Thanks to @mamehaze for previous commits:

https://github.com/mamedev/mame/commit/3ca05f596d14fba033f33fc4ec1dbb28d8617070
https://github.com/mamedev/mame/commit/13b609deb2076a9c5df4e55dc042e1fc3c6f3988

### DIFF
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -644,9 +644,9 @@ int segas32_state::compute_clipping_extents(screen_device &screen, int enable, i
 void segas32_state::compute_tilemap_flips(int bgnum, int &flipx, int &flipy)
 {
 	// determine flip bits
-	int global_flip    = BIT(system32_videoram[0x1ff00 / 2], 9);
-	int layer_flip     = BIT(system32_videoram[0x1ff00 / 2], bgnum);
-	int prohibit_flipy = BIT(system32_videoram[0x1ff00 / 2], 8);
+	int global_flip    = BIT(m_videoram[0x1ff00 / 2], 9);
+	int layer_flip     = BIT(m_videoram[0x1ff00 / 2], bgnum);
+	int prohibit_flipy = BIT(m_videoram[0x1ff00 / 2], 8);
 
 	*flipx = (layer_flip) ? !global_flip : global_flip;
 
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -920,7 +920,7 @@ void segas32_state::update_tilemap_rowscroll(screen_device &screen, segas32_stat
 			{
 				const rectangle &visarea = screen.visible_area();
 				srcy = yscroll + visarea.max_y - y;
-				ylookup = cliprect.max_y - y;
+				ylookup = visarea.max_y - y;
 			}
 
 			/* apply row scroll/select */
--- a/src/mame/sega/segas32_v.cpp
+++ b/src/mame/sega/segas32_v.cpp
@@ -541,7 +541,7 @@ TILE_GET_INFO_MEMBER(segas32_state::get_tile_info)
 
 int segas32_state::compute_clipping_extents(screen_device &screen, int enable, int clipout, int clipmask, const rectangle &cliprect, extents_list *list)
 {
-	int flip = (m_videoram[0x1ff00/2] >> 9) & 1;
+	int flip = BIT(m_videoram[0x1ff00 / 2], 9);
 	rectangle tempclip;
 	rectangle clips[5];
 	int sorted[5];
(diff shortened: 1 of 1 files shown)


src/mame/sega/segas32_v.cpp
```diff
@@ -646,20 +643,19 @@ int segas32_state::compute_clipping_extents(screen_device &screen, int enable, i
...
 {
-	/* determine if we're flipped */
-	int global_flip = (m_videoram[0x1ff00 / 2] >> 9)&1;
-
-	flipx = global_flip;
-	flipy = global_flip;
+	/* determine flip bits */
+	int global_flip    = (system32_videoram[0x1ff00 / 2] >> 9) & 1;
+	int layer_flip     = (system32_videoram[0x1ff00 / 2] >> bgnum) & 1;
+	int prohibit_flipy = (system32_videoram[0x1ff00 / 2] >> 8) & 1;
```
happppp: For new or replaced code that does (x >> y & 1), we more commonly use BIT(x, y).
And use // for single line comments btw.

src/mame/sega/segas32_v.cpp
```diff
@@ -646,20 +643,19 @@ int segas32_state::compute_clipping_extents(screen_device &screen, int enable, i
...
-	flipy ^= layer_flip;
-	flipx ^= layer_flip;
+	if (layer_flip)
+		*flipx = !*flipx;
 
-	// this bit is set on Air Rescue (screen 2) title screen, during the Air Rescue introduction demo, and in f1en when you win a single player race
-	// it seems to prohibit (at least) the per-tilemap y flipping (maybe global y can override it)
-	if ((m_videoram[0x1ff00 / 2] >> 8) & 1) flipy = 0;
+	if (layer_flip && !prohibit_flipy)
+		*flipy = !*flipy;
```
cuavas: You shouldn’t be flipping it, then unflipping it again.

src/mame/sega/segas32_v.cpp
```diff
@@ -705,7 +701,7 @@ void segas32_state::update_tilemap_zoom(screen_device &screen, segas32_state::la
 //if (screen.machine().input().code_pressed(KEYCODE_X) && bgnum == 1) opaque = 1;
 	int flipx, flipy;
 
-	// todo determine flipping
+	/* determine flipping */
```
cuavas: Don’t replace line comments with C-style comments for single line comments.

src/mame/sega/segas32_v.cpp
```diff
@@ -914,21 +910,24 @@ void segas32_state::update_tilemap_rowscroll(screen_device &screen, segas32_stat
...
+				ylookup = cliprect.max_y - y;
 			}
 
 			/* apply row scroll/select */
 			if (rowscroll)
-				srcx += table[0x000 + 0x100 * (bgnum - 2) + y] & 0x3ff;
+				srcx += table[0x000 + 0x100 * (bgnum - 2) + ylookup] & 0x3ff;
 			if (rowselect)
-				srcy = (yscroll + table[0x200 + 0x100 * (bgnum - 2) + y]) & 0x1ff;
+				srcy = (yscroll + table[0x200 + 0x100 * (bgnum - 2) + ylookup]) & 0x1ff;
```
cuavas: Relying on the clipping rectangle matching the screen resolution isn’t safe, since you can get horizontal partial updates.

This whole part of the change looks sus.  The visible area should give you the current horizontal resolution.  Using the clipping rectangle for anything other than clipping is definitely wrong.

mahoneyt944: Updated to use visible area

src/mame/sega/segas32_v.cpp
mahoneyt944: Fixed

src/mame/sega/segas32_v.cpp
mahoneyt944: Fixed

mahoneyt944:
Hi, thank you for the quick replies, I updated the PR with the changes requested.

src/mame/sega/segas32_v.cpp
mahoneyt944: I fixed the comment line. I did not use BIT(x, y) however. I'm not sure how that would integrate into `compute_tilemap_flips`, or do you mean to swap x /y within the actual draw layers? If so perhaps that could be done in a follow up commit? I'm willing to make the requested change but didn't want to misconstrue the request.

holub: `= BIT(system32_videoram[0x1ff00 / 2], 8);`

happppp: BIT:
something >> shiftamount & 1
is equal to: BIT(something, shiftamount)

optionally, the 3rd param of BIT() is number of bits to mask, but I find that one harder to read.

mahoneyt944: Ok pushed update.

mahoneyt944:
I verified to the best of my ability that the tilemap flipping now works for all games and orientations. The best games to examine are radr, radm, and svf as previously noted with issues. I believe this is ready to merge. @happppp @cuavas any further concerns for me to address?

PR #14195, opened 2025-09-21
Set IP_ACTIVE_HIGH to low marywu.cpp

flama12333:
add p1 input. will prevent hang at 01 and bypass 30 error.  except the 0x80 at key2 and 0x40 at p1.  will shown error 02 due to hopper not implemented and something.


src/mame/misc/marywu.cpp
```diff
@@ -96,11 +95,15 @@ static INPUT_PORTS_START( marywu )
...
-	PORT_BIT(0xf0, IP_ACTIVE_HIGH, IPT_UNUSED )
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_BUTTON1) // K0
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_BUTTON2) // K1
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_BUTTON3) // K2
+	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_BUTTON4) // K3
+	PORT_BIT(0xf0, IP_ACTIVE_LOW, IPT_UNUSED )
+
+    PORT_START("P1")
+    PORT_BIT(0x40, IP_ACTIVE_HIGH, IPT_KEYPAD) PORT_NAME("P1.6") PORT_CODE(KEYCODE_C) // If press or IP_ACTIVE_LOW Will cause Error 30 if press
+    PORT_BIT(0x80, IP_ACTIVE_LOW,  IPT_KEYPAD) PORT_NAME("P1.7") PORT_CODE(KEYCODE_V) // If press during startup, it will cause error 76.
```
cuavas, line 106: We have schematics.  Why are you just making random guesses?

P1 bits are:
* /ENABLE (output)
* 入表 (output)
* 出表 (output)
* SSR (output)
* 入表檢查 (input)
* 出表檢查 (input)
* HOPPER SW (input)
* 洗分 SW (input)

flama12333: only to test if the input at the p1 does anything.
I had to bypass error to access k0 k1 k2 k3 button.

PR #14196, opened 2025-09-21
rs232.cpp: Add option for Heath H19 terminal

mgarlanger:
Adds the Heath H19 terminal to the rs232 option list.

h8/h_8_5.cpp: Change default connection from generic terminal to the Heath H19

### DIFF
--- a/src/devices/bus/rs232/heath_h19.cpp
+++ b/src/devices/bus/rs232/heath_h19.cpp
@@ -36,9 +36,10 @@ protected:
 private:
 	required_device<heath_tlb_connector> m_tlbc;
 
+	static void tlb_options(device_slot_interface &device);
 };
 
-static void tlb_options(device_slot_interface &device)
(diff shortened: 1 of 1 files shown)


src/devices/bus/rs232/heath_h19.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+	virtual void device_start() override ATTR_COLD { }
+	virtual void device_add_mconfig(machine_config &config) override ATTR_COLD;
+
+private:
+	required_device<heath_tlb_connector> m_tlbc;
+
+};
+
+static void tlb_options(device_slot_interface &device)
```
galibert: Can you put that as a static method of the class please?

mgarlanger: @galibert  Updated to a static method.

PR #14198, opened 2025-09-21
ussr/bk: emulate bk0011 hardware and b&w/color video switch.

shattered:
Improve interrupt handling in K1801VM1 CPU core.

Miscellaneous bugfixes (keyboard interrupt enable, cassette input).

### DIFF
--- a/src/devices/cpu/t11/t11.cpp
+++ b/src/devices/cpu/t11/t11.cpp
@@ -40,6 +40,14 @@ DEFINE_DEVICE_TYPE(K1801VM1, k1801vm1_device, "k1801vm1", "K1801VM1")
 DEFINE_DEVICE_TYPE(K1801VM2, k1801vm2_device, "k1801vm2", "K1801VM2")
 
 
+/*
+ * Not implemented:
+ *
+ * instruction timing (+ memory timing), bug compatibility (MOVB, MFPS),
+ * built-in timer, revision G (MUL instruction, timer IRQ),
+ * extra PSW bits, on-chip registers, multiprocessing support,
+ * traps: double bus error and bus error on vector fetch.
+ */
 k1801vm1_device::k1801vm1_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock)
 	: t11_device(mconfig, K1801VM1, tag, owner, clock)
 	, z80_daisy_chain_interface(mconfig, *this)
@@ -71,6 +79,7 @@ t11_device::t11_device(const machine_config &mconfig, device_type type, const ch
 		reg.d = 0;
 	m_psw.d = 0;
 	m_ppc.d = 0;
+	m_cp[0] = m_cp[1] = m_cp[2] = m_cp[3] = 0;
 }
 
 t11_device::t11_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock)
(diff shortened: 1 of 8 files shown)


src/devices/cpu/t11/t11.cpp
```diff
@@ -218,27 +227,35 @@ static const struct irq_table_entry irq_table[] =
 // PSW11 also masks IRQ1.  PSW10 also masks ACLO.
 void k1801vm1_device::t11_check_irqs()
 {
+	int8_t              mcir = MCIR_NONE;
+	uint16_t            vsel = 0;
+
```
galibert, line 232: You seem to have removed mcir and vsel from the state, but you didn't actually remove them from the class?  Or am I misreading something

src/mame/ussr/bk.cpp
```diff
@@ -55,10 +54,73 @@ void bk_state::bk0010fd_mem(address_map &map)
 	map(0120000, 0157777).ram();
 }
 
+void bk_state::bk0011_mem(address_map &map)
+{
+	map(0000000, 0177777).rw(FUNC(bk_state::trap_r), FUNC(bk_state::trap_w));
+
+	map(0000000, 0037777).ram().share(m_ram[0]);
+
+	map(0040000, 0077777).view(m_view1);
```
galibert, line 63: You can have for loops in memory maps, it's normal C++ code after all, that may make things a little less copypasty.

src/mame/ussr/bk.cpp
```diff
@@ -122,6 +187,44 @@ void bk_state::bk0010fd(machine_config &config)
...
+	m_maincpu->set_addrmap(AS_PROGRAM, &bk_state::bk0011_mem);
+	m_maincpu->set_daisy_config(daisy_chain);
+	m_maincpu->out_reset().set(FUNC(bk_state::reset_w));
+
+	m_kbd->halt_wr_callback().set([this] (int state) {
+		if (!m_stop_disabled) m_maincpu->set_input_line(t11_device::HLT_LINE, state);
+	});
+
+	subdevice<qbus_slot_device>("qbus:1")->set_default_option("by");
+	subdevice<qbus_slot_device>("qbus:1")->set_option_default_bios("by", "253");
```
galibert, line 203: Vas, that's expected to work and keep working in a machine config?

src/mame/ussr/bk_m.cpp
```diff
@@ -99,10 +204,55 @@ u32 bk_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, const r
...
+			{
+				for (u8 b = 0; b < 16; b++)
+					bitmap.pix(y, x * 16 + b) = BIT(code, b) << 2;
+			}
+		}
+	}
+	return 0;
+}
+
+TIMER_DEVICE_CALLBACK_MEMBER(bk_state::scanline_callback_bk11)
```
galibert, line 227: Shouldn't that use the standard partial screen update mechanism?

src/devices/cpu/t11/t11.cpp
shattered: They are still in the state -- check_irqs copies them to these temporary vars

src/mame/ussr/bk_m.cpp
shattered: Hmm, wasn't aware of that -- what's a good example?

src/mame/ussr/bk.cpp
shattered: Yes, but in this case key order doesn't match (6, 0, ..., 5 -> 0, 1, ..., 7)

galibert: Sure, but that order is used twice, afaict.  So it would fit into a static array in the loop.

src/mame/ussr/bk_m.cpp
galibert: look for update_partial or update_now.  Update partial tells to go up to a specific line, update now up to the current time.  It triggers a call to screen_update with an appropriate cliprect.  Avoids a copy.

angelosa: On the other hand your screen_update should also clip against the current cliprect otherwise will get over-copying (which translates to poor performance). 

Note that I don't think this is a bad approach of rendering (video chips usually prepare data per scanline, around hblank), but I wouldn't call `TIMER(config, "scantimer")` but rather instantiate an `emu_timer` following the current beam positions.

shattered: Thanks, I'll take a look for a future PR (have more updates planned for BK hardware).

PR #14212, opened 2025-09-24
sega/model2.cpp: setting a timer to 0xfffff disables it

gm-matthew:


galibert:
That doesn't seem insane, you have something that really suggests you're correct?

ajrhacker:
The only way this would make sense is if the timers were counting up to 0xfffff instead of down to 0.

Lord-Nightmare:
@ajrhacker I disagree, the timers may have an internal "TC" (Terminal Count) state which is active if all bits of the timer register are set (using a chained or multi input AND gate internally?), which is similar to how most 4-bit counters work. This would go active both if the counter was an up counter and it had reached the maximum count, or if the counter was a down counter and it had just underflowed past zero, in both cases masking the clock and preventing future counts from happening.

ajrhacker:
@Lord-Nightmare: I think you're mistaken about "how most 4-bit counters work." Most up-down counter ICs generate TC as carry (maximum count) in up mode and as borrow (zero count) in down mode. A few generate these as separate outputs to facilitate chaining when the up and down modes have separate clock inputs.

gm-matthew:
I submitted this commit as the timer interrupt handler for vf2, fvipers and schamp does nothing other than reset the timer to 0xfffff and clear the interrupt, hence the timer interrupts appeared to be redundant; the games work just fine without any timer interrupts at all, and all other games appear to be unaffected.

But I have recently discovered counter-evidence against the timers being disabled when set to 0xfffff in the initialization code for these three games: internal timer 2 is set to 0xfffff but is then later read to calculate a new timer value; it would not make sense to disable a timer and then use it to measure how much time has passed.

Closing due to the counter-evidence.

PR #14214, opened 2025-09-25
formats/ap2_dsk.cpp: add write support for the D13 format

SpecLad:


SpecLad:
Note to reviewers: most of the PR is refactoring `a2_16sect_format` to reuse most of its saving logic. I separated as much of the refactoring as I could into separate commits, so I would suggest reviewing commit-by-commit.

PR #14217, opened 2025-09-26
Fix rendering of LinnDrum's rounded corner outlines.

felipesanches:
# Before
Screenshot From 2025-09-26 04-13-54

# After
Screenshot From 2025-09-26 04-13-22

### DIFF
--- a/src/mame/layout/linn_linndrum.lay
+++ b/src/mame/layout/linn_linndrum.lay
@@ -145,39 +145,35 @@ copyright-holders:m1macrophage
 	<element name="screw"><rect><color red="0.19" green="0.18" blue="0.18"/></rect></element>
 	<element name="jack"><rect><color red="0.35" green="0.35" blue="0.36"/></rect></element>
 
-	<element name="outline_cassette_tempo">
+	<element name="outlines">
 		<image><data><![CDATA[
-			<rect stroke="#f28367" fill="none" stroke-width="3" width="124" height="400" rx="16" ry="16"/>
-		]]></data></image>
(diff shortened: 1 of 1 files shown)


m1macrophage:
Cool! Those corners were bugging me.

felipesanches:
> Cool! Those corners were bugging me.

I couldn't stop staring at them as well. I had to fix it otherwise I could not sleep! LOL

felipesanches:
In this case I decided to place all curves in a single SVG.

But in other cases, specially when we want to have many small SVG snippets for individual elements, it is  important to remember that:

(1) you need a root svg tag to declare the overall dimensions of the image, otherwise it will later be stretched and potentially distorted when referenced in an element with bounds that differ from whatever default svg document dimentions were inferred.
(2) if the curves of your drawing are touching your bounding box, then you gotta be extra careful to not let them be clipped out. It is important to remember that the thickness of an outline grows 50% inwards and 50% outwards. So a vertical line with `stroke-width="3px"` touching the the left border must have coordinate `x="1.5px"`. And this also obviously applies to the right border as well as to the top and bottom ones (y-axis values).

m1macrophage:
>  It is important to remember that the thickness of an outline grows 50% inwards and 50% outwards. 

Yeah, I realized that was the issue, but I did not know how to fix it. For newer layouts, I resorted to drawing two rounded rects, one inside the other, with `stroke="none"`. And I was slowly backporting to older layouts as I was working on them. But this PR taught me the `<svg width="" height="">` trick :).

PR #14218, opened 2025-09-26
Update Chinese_Simplified strings.po

Dirstac:
The current version of the language file contains a large number of incorrect translations and missing parts. This version has been re-translated and supplemented.

### DIFF
--- a/language/Chinese_Simplified/strings.po
+++ b/language/Chinese_Simplified/strings.po
@@ -1,6 +1,6 @@
 # Chinese Simplified translations for MAME package
 # MAME 套件的简体中文翻译
-# Copyright (C) 1997-2025 MAMEdev and contributors
+# Copyright (C) 1997-2024 MAMEdev and contributors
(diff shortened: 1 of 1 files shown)


language/Chinese_Simplified/strings.po
```diff
@@ -1,6 +1,6 @@
 # Chinese Simplified translations for MAME package
 # MAME 套件的简体中文翻译
-# Copyright (C) 1997-2025 MAMEdev and contributors
+# Copyright (C) 1997-2024 MAMEdev and contributors
```
felipesanches: This looks incorrect. Copyright year should remain as 2025, instead of being changed back to 2024.

Dirstac: You're right. I was negligent

Gently28:
[strings.zip](https://github.com/user-attachments/files/22706559/strings.zip)
吾道不孤！
https://bbs.xqemu.cn/forum.php?mod=viewthread&tid=4160

PR #14223, opened 2025-09-27
src/frontend/mame/ui/slotopt.cpp: add a menu option to copy the curre…

goldnchild:
…nt slot options to the clipboard

It can be difficult to set up a working command line with various slot options.  This would add a menu option to the Slot Options menu that would copy the current slot options to the clipboard.  This text could then be pasted as part of a command line to reproduce the configuration.

### DIFF
--- a/src/frontend/mame/ui/slotopt.cpp
+++ b/src/frontend/mame/ui/slotopt.cpp
@@ -257,8 +257,9 @@ bool menu_slot_devices::handle(event const *ev)
 	}
 	else if (ev->itemref == ITEMREF_COPYTOCLIPBOARD)
 	{
-		if (ev->iptkey == IPT_UI_SELECT)
+		if (ev->iptkey == IPT_UI_SELECT || ev->iptkey == IPT_UI_CLEAR)
 		{
+			bool inistyle = (ev->iptkey == IPT_UI_CLEAR);
 			std::stringstream outss;
 			for (device_slot_interface &slot : slot_interface_enumerator(m_config->root_device()))
 			{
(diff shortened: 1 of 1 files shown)


cuavas:
Have you not noticed that the slot devices menu *literally shows you what you need* on the command line or in an INI file to produce the current configuration?  You just put the stuff it shows in your INI file, skipping “[internal]” things.

Look, here’s an example:
image
```
rs232 
mb8877:0 525ssdd
mb8877:1 525ssdd
```

Here’s another one:
image
```
fdc:0 35dd
fdc:1 
printer
modem 
scsi:0 
scsi:1 
scsi:2 
scsi:4
scsi:5 
scsi:6 harddisk
nb9 mdc824
nba 
nbb 
nbc 
nbd 
nbe 
```

Or another system:
image
```
tty terminal
j7 imm4_90
j8 imm6_26
j9 
j10 
j11 
j12 
j13 
j14 
j15 
j16 
j17 
j18 
j19 
```

The only differences between INI and command line syntax are the leading hyphens and quoting/escaping rules (e.g. needing quotes for empty options).

Front-ends can also give interactive slot configuration purely using the `-listxml` output.  The included minimaws script provides a reference implementation for doing this, and can show INI or command line syntax, with or without explicit defaults (e.g. [try it out for the Commodore 64](https://arcade.vastheman.com/minimaws/machine/c64#heading-options)).

This feels like a convoluted non-solution to a contrived problem.

goldnchild:
Thanks for the information, Vas.

I've never used ini files before, always using the command line.

That minimaws looks very comprehensive down to being able to choose the device bioses.

I was just looking for a quick way to make a command line from within mame that doesn't require anything external to mame.  Yes, you can copy the information out of the Slot Devices menu by hand, but that gets tedious.



ok, been doing some experimenting, I just discovered the parameter "-writeconfig" which writes out the current slot configuration as an .ini file.  (all along I have been specifying everything manually because I didn't realize it can save the slot options to the ini)

All along it hasn't been saving the slot configuration because by default it doesn't writeconfig.

I kept looking for the .ini file, but by default it gets saved in my home directory in a hidden .mame directory

inipath                   $HOME/.mame;.;ini


So launching ./mame ti99_4a -writeconfig and changing the slots
and the ti99_4a.ini file shows the slot devices:

#
# SLOT DEVICES
#
ioport                    peb

#
# IMAGE DEVICES
#
cartridge                 
cassette1                 
cassette2                 
ioport:peb:slot2          32kmem


Interestingly, you have to disable -writeconfig manually as once it writes the config, it seems to put a writeconfig 1 into the ini file.

PR #14244, opened 2025-09-30
Add skeleton for ICE Full Court Fever hardware

clawgrip:
Star Basket [Juan Romero, ClawGrip]


ajrhacker:
From the schematics for Full Court Fever, the hardware is similar to ICE's Bozo board, though there are enough differences to justify a separate driver.

PR #14252, opened 2025-10-02
Add a Non Working Driver For Fruit Cocktail fruitctk.cpp

flama12333:
Currently shown error 02. Input was guessed after found ways to bypass.
info and driver in wip.
expected some mistake.
layout currently not correctly.
some code are from marywu.cpp
led controller are different than marywu.
Question:
which name choosen? only to prevent confusion.


flama12333:
I commited one mistake. We should separate patch now idk how? now currently moved. Lesson learned. Always check everything.

PR #14255, opened 2025-10-02
seibu/raiden2.cpp, seibu/r2dx_v33.cpp: Minor cleanups, some other Seibu Kaihatsu hardware minor cleanups

cam900:
seibu/raiden2.cpp: Minor cleanups:
- Fix bankswitching in reset
- Simplify CPU bankswitching method
- Use logmacro.h for logging
- Fix save state supports
- Reduce preprocessor defines
- Fix CPU ROM region size
- Reduce runtime tag lookups
- Fix namings
- Fix spacings
- Fix filenames
- Add notes

seibu/r2dx_v33.cpp: Minor cleanups:
- Split driver states per configurations
- Fix save state support
- Fix bankswitching in reset
- Use BIT macro for single bit value
- Use hexadecimal value for address value
- Reduce runtime tag lookups
- Add ROMREGION_ERASEFF for regions to prevent possible transparency issue
- Fix spacings
- Fix namings

seibu/seibu_helper.cpp: Split partial_carry_sum into seibu_helper.cpp for reduce including unnecessary functions

seibu/r2crypt.cpp: Allow variable ROM region and size in sprite ROM decryption functions

seibu/seibuspi_m.cpp, seibu/seibuspi_v.cpp: Fix spacings

seibu/legionna.cpp: Fix naming


src/mame/seibu/seibu_helper.h
cuavas, line 1: These functions should really have less generic names to avoid potential conflicts.

PR #14256, opened 2025-10-02
Faster winhybrid

janko-jj:
I've observed slow startups of Mame on my Windows machine, and I narrowed it down to the inefficient API use in src/osd/modules/input/input_winhybrid.cpp 

Specifically, it uses WMI which, as configured, starts a new process for a "WMI server", and I concluded it is the cause of multiple seconds delay.

I rewrote that code to use setupapi, which is a supported API, just with much less overhead. I verified that the lists of the iterated devices are identical in both cases (using WMI and my setupapi iteration) on my machine.

The changes are:

1) adding the linkage to the setupapi in `scripts/src/osd/windows.lua`

2) calling the API functions from setupapi which avoid the overheads which made these long "100% CPU, nothing can be done" delays.

The multiple commits are accidental, as I've fixed the build problems. Only two files changed in a single commit describes the changes correctly: it's just **speeding up winhybrid enumeration of the device IDs**.

The speedup of the startup on Windows is more than obvious on my machine (running Windows 8.1 Professional).

The picture shows the CPU usage without the fix and with it. During all the time of high CPU usage (almost 10 seconds without the fix on my machine) the mouse cursor spins and user just waits, seeing nothing and unable to do anything.

mame-faster-winhybrid

### DIFF
--- a/scripts/src/osd/windows.lua
+++ b/scripts/src/osd/windows.lua
@@ -36,6 +36,7 @@ function maintargetosdoptions(_target,_subtarget)
 		"psapi",
 		"shlwapi",
 		"uuid",
+		"setupapi"
 	}
 end
 
--- a/src/osd/modules/input/input_winhybrid.cpp
+++ b/src/osd/modules/input/input_winhybrid.cpp
@@ -17,7 +17,6 @@
 #include <oleauto.h>
 
 #include <setupapi.h>
-#pragma comment(lib, "setupapi.lib")
 
 namespace osd {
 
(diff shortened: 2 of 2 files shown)


cuavas:
* Microsoft officially recommends using WMI to do this.
* SetupAPI is supposed to be used by installer applications, this is very much off-label use.
* Microsoft’s documentation for SetupAPI has a big coloured warning on the front page saying, “SetupApi is not supported on all editions of Windows. When possible, you should use lower layer APIs…”

janko-jj:
"When possible, you should use lower layer APIs…"  WMI is actually a higher layer.

Would the changes which use would cfgmgr32 API (instead of setupapi) also be rejected? E.g. if I'd use CM_Get_Device_ID_List ?

https://learn.microsoft.com/en-us/windows/win32/api/cfgmgr32/nf-cfgmgr32-cm_get_device_id_lista

"Minimum supported client: Available in Microsoft Windows 2000 and later versions of Windows.
Target Platform: Universal"

holub:
Can we participate in the reasonable dialogue? https://github.com/orgs/mamedev/discussions/223

PR #14259, opened 2025-10-03
pgm3.cpp - bootstrap first decrypted block from internal Flash [Peter Wilhelmsen, David Haywood]

mamehaze:
this is done as the internal mask ROM part seems to have a few errors, so is a better starting point
this requires the use of AES_256_CBC ( https://github.com/halloweeks/AES-256-CBC ) which is MIT licensed (GPL compatible) and has been added to the 3rdparty folder.


src/mame/igs/pgm3.cpp
```diff
@@ -44,6 +44,7 @@
 #include "emupal.h"
 #include "screen.h"
 
+#include "aes256cbc/AES_256_CBC.h"
```
galibert: Trying to use that pseudo-header-only library in any other device will break the ODR.  So no, not that way.  Instanciate it in util, and provide an aes.h there with the functions declarations.

mamehaze: ok, I changed it to be more like nanosvg

src/lib/util/aes256cbc.h
```diff
@@ -0,0 +1,15 @@
...
+
+    AES256CBC helpers.
+
+***************************************************************************/
+#ifndef MAME_LIB_UTIL_AES256CBC_H
+#define MAME_LIB_UTIL_AES256CBC_H
+
+#include <aes256cbc/AES_256_CBC.h>
+
+#endif // MAME_LIB_UTIL_AES256CBC_H
```
galibert, line 28: You still defining the arrays and methods multiple times.  aes256cbc/AES_256_CBC.h should only be included from one .cpp in the whole of mame.

nanosvg has a define to avoid the problem, aes256cbc doesn't.

mamehaze: I was actually a bit confused by nanosvg

the .cpp has
#include <nanosvg/src/nanosvg.h>
#include <nanosvg/src/nanosvgrast.h>
(no guards)

and the .h file has
#include <nanosvg/src/nanosvg.h>
#include <nanosvg/src/nanosvgrast.h>
(with guards)

in this case I guarded the .h file (just like nanosvg) which contains the
#include <aes256cbc/AES_256_CBC.h>

and just include that header from the .cpp file, rather than duplicating that include there too

are you sure it's set up properly in nanosvg in the first place? it seems to break the rule you're telling me to apply

galibert: The nanosvg cpp also has:
#define NANOSVG_ALL_COLOR_KEYWORDS
#define NANOSVG_IMPLEMENTATION
#define NANOSVGRAST_IMPLEMENTATION

It's the _IMPLEMENTATION defines that tell the header to include the definitions, otherwise it doesn't.  aes256cbc doesn't have an equivalent define.

mamehaze: well I thought the other rule was that we're not meant to modify 3rdparty folder stuff with MAME specifics, so as there's no such defines in the library itself, I can't use them.

I've tried something else, which seems less than ideal to me, but should hopefully avoid the problem?

PR #14260, opened 2025-10-03
sinclair/specnext_dma.cpp: Fixed prescaler reg init

holub:


cuavas:
The description/commit message doesn’t seem to match the change here.

src/mame/sinclair/specnext_dma.cpp
```diff
@@ -60,55 +60,23 @@ void specnext_dma_device::write(u8 data)
 	else
 	{
 		int nreg = m_regs_follow[m_cur_follow];
-		z80dma_device::write(data);
 		if(nreg == REGNUM(2, 1))
 		{
 			if (data & 0x20)
 				m_regs_follow[m_num_follow++] = GET_REGNUM(ZXN_PRESCALER);
 		}
+		z80dma_device::write(data);
```
holub: It does match description. This is what was done for init fix. Rest is tidiness

PR #14263, opened 2025-10-04
bk: parallel expansion slot and carts

shattered:


cuavas:
This doesn’t meet basic standards:

* The device descriptions are far too generic, e.g. “Printer Interface” is completely unsuitable when these are globally identifying names across all of MAME.
* Superfluous semicolons on inline functions.
* Cold member functions not marked cold.
* Directly #including sound.h from headers for some weird reason (you’re supposed to get it via emu.h).
* Headers #include stuff only needed in implementation (e.g. speaker.h).
* Device classes declared in headers when they realistically will only be accessed through interfaces.
* `bk_covox_device::m_data` seems to be completely unnecessary – leftover copy/paste from loopback device?
* Unnecessary overridden member functions that just slow down compilation (e.g. all the empty `device_reset`).
* Useless `#define VERBOSE 0` in files that don’t have any logging and don’t even `#include "logmacro.h"` – this will confuse people looking for the additional logging switching it to 1 should enable.

I’m not cleaning this one up myself.  I’m sick of people just assuming I’ll do it.  If it isn’t cleaned up tomorrow, I’m reverting it.

rb6502:
@shattered You got this?

shattered:
Working on it.

shattered:
I think I've addressed all of these except this one (cannot find any?)
> Superfluous semicolons on inline functions.

This was copy/paste from bus/rc2014/sound.cpp, so I've fixed it as well
> Directly #including sound.h from headers

PR #14264, opened 2025-10-04
Added emulation of a CEM3320-based 4th-order LPF, and used it on the linndrum. 

m1macrophage:
* Added a generic 4-th order resonant LPF (VA_LPF4), based on the implementation in esq1.cpp.
  * Copied some of the code and comments  from `esq1.cpp`, with credit on the file copyright.
* Added a CEM3320_LPF4 specialization.
* Used the CEM3320_LPF4 on the linndrum.
* Implemented the corresponding VCF envelope generators on the linndrum.
* Fixed typos and comments.

The LinnDrum VCFs do not use resonance. Resonance functionality was tested by locally using the resonance configuration of the prophet5.

### DIFF
--- a/src/mame/linn/linndrum.cpp
+++ b/src/mame/linn/linndrum.cpp
@@ -2,9 +2,9 @@
 // copyright-holders:m1macrophage
 
 /*
-The LinnDrum (unofficially also known as LM-2) is a digital drum machine. It
-stores and plays digital samples (recordings of acoustic dumps), some of which
-are processed by analog filters or other analog circuits.
+The LinnDrum is a digital drum machine. It stores and plays digital samples
+(recordings of acoustic drums), some of which are processed by analog filters or
+amplifiers.
 
 The firmware runs on a Z80. It controls the UI (reads buttons, drives displays
 and LEDs), synchronization with other devices, cassette I/O, and voice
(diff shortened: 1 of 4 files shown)


src/mame/linn/linndrum.cpp
```diff
@@ -113,14 +108,20 @@ with a high sample rate.
 #define LOG_MIX              (1U << 7)
 #define LOG_PITCH            (1U << 8)
 #define LOG_HAT_EG           (1U << 9)
+#define LOG_CV_OFFSET        (1U << 10)
 
-#define VERBOSE (LOG_GENERAL)
+#define VERBOSE (LOG_GENERAL | LOG_CV_OFFSET)
```
rb6502: In general things should be checked in with this set to 0 to minimize the host CPU requirements.

m1macrophage: Note that the current VERBOSE setting will not cause any logging under normal operation.

LOG_CV_OFFSET will only log when the user adjusts certain trimmers (via the "sliders" menu) and it provides info to help with that calibration.

There are a couple of LOG() calls that are enabled by LOG_GENERAL, but those log situations that "should never happen". 

In general, I've been following the principle of no logging under normal operation for working drivers. I have been a bit more permissive for non-working ones, so as to provide evidence of operation.

But happy to set this to 0, or to just LOG_CV_OFFSET, if that's preferable.

rb6502: Things that should not happen generally are logerror() (logs but continues) or fatalerror() (halts MAME) depending on the severity.  That way error callouts are distinct from informational logging.

m1macrophage: I see. In this case the errors in question could only happen with buggy firmware. Would that fall under the logerror() category? Or should we just not log it?

rb6502: I'd use logerror() as long as it's something that you can recover from easily (for instance, by ignoring it).

m1macrophage: Thanks, I used logerror(). The errors are indeed recoverable / ignorable. I also renamed `LOG_CV_OFFSET` to `LOG_CALIBRATION`, to make the intention more obvious, and removed `LOG_GENERAL` from `VERBOSE`.

PR #14265, opened 2025-10-04
Prevent NTSC/PAL NES from loading Basic carts

FoxhackDN:
This change prevents loading all variants of Family Basic carts on non-Japanese hardware. Since stock systems lack the expansion port to use the keyboard (and consequentially, the audio port for the tape deck), this renders them unusable and causes the same issues that were mentioned in MAMETesters bug 8443.

### DIFF
--- a/hash/nes.xml
+++ b/hash/nes.xml
@@ -11837,7 +11837,8 @@ license:CC0-1.0
 		<info name="serial" value="HVC-BS"/>
 		<info name="release" value="19840621"/>
 		<info name="alt_title" value="ファミリーベーシック"/>
-		<sharedfeat name="exp_default" value="fc_keyboard"/>
+		<sharedfeat name="compatibility" value="fbasic" />
+		<sharedfeat name="exp_default" value="fc_keyboard" />
 		<part name="cart" interface="nes_cart">
 			<feature name="slot" value="hvc_basic" />
 			<feature name="pcb_model" value="FB-R-128-01" />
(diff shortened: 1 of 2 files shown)


angelosa:
I'm okay with the principle. However does this also means that _anything_ with an `exp_default` should be filtered out, which means all the Family Trainer games as well? If true I suggest to rename the compatibility filter as `exp` instead of `fbasic`.

FoxhackDN:
> I'm okay with the principle. However does this also means that _anything_ with an `exp_default` should be filtered out, which means all the Family Trainer games as well? If true I suggest to rename the compatibility filter as `exp` instead of `fbasic`.

... hold on a sec. I just noticed Stadium Events (US/PAL) has no `exp_default`, I don't think the game is even playable without the Power Pad. Is this by design?

etabeta78:
Stadium Events uses Power Pad that plugs into controller port, not the
Family Trainer controller that plugs into the expansion port

FoxhackDN:
> Stadium Events uses Power Pad that plugs into controller port, not the Family Trainer controller that plugs into the expansion port

Yeah, and I assumed `exp_default` would boot the game with the necessary device already plugged in.

If this isn't used for non-JP systems, then okay. I had no idea. I'll make the suggested changes later today and submit them after testing.

cuavas:
Is there a reason you didn’t express this positively in the code by adding an “exp” filter to systems with the expansion port?

FoxhackDN:
> Is there a reason you didn’t express this positively in the code by adding an “exp” filter to systems with the expansion port?

I only had examples where the negative was used as a reference - specifically, the CD-i and its filter to prevent discs that required the DVC expansion from being loaded. I'm not opposed to someone cleaning the code up, but I'm not experienced enough to know how to do it properly.

PR #14267, opened 2025-10-05
machine/nmk112.cpp: Convert bank method into memory_bank_creator and address_space finders

cam900:
- Reduce duplicates
- Fix spacings and argument data type
- atlus/cave.cpp, atlus/patapata.cpp, nmk/nmk16.cpp, nmk/nmkmedal.cpp, nmk/quizpani.cpp: Fix memory region size for NMK112 bankswitched OKI MSM6295 spaces

### DIFF
--- a/src/devices/machine/nmk112.cpp
+++ b/src/devices/machine/nmk112.cpp
@@ -18,14 +18,11 @@ DEFINE_DEVICE_TYPE(NMK112, nmk112_device, "nmk112", "NMK112")
 
 nmk112_device::nmk112_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock)
 	: device_t(mconfig, NMK112, tag, owner, clock)
-	, m_samplebank{ {*this, "samplebank_0_%u", 0U}, {*this, "samplebank_1_%u", 0U} }
-	, m_tablebank{ {*this, "tablebank_0_%u", 0U}, {*this, "tablebank_1_%u", 0U} }
+	, m_samplebank{ { *this, "samplebank_0_%u", 0U }, { *this, "samplebank_1_%u", 0U } }
+	, m_tablebank{ { *this, "tablebank_0_%u", 0U }, { *this, "tablebank_1_%u", 0U } }
 	, m_rom(*this, { finder_base::DUMMY_TAG, finder_base::DUMMY_TAG })
-	, m_oki0_space(*this, finder_base::DUMMY_TAG, -1)
-	, m_oki1_space(*this, finder_base::DUMMY_TAG, -1)
 	, m_page_mask(0xff)
-	, m_current_bank{0}
-	, m_size{0}
+	, m_size{ 0, 0 }
 {
 }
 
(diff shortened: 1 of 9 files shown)


cam900:
> The change to use real bank objects is a good idea, but I don’t think the method of installing into address spaces is ideal. Rather than adding the address space finders to the NMK112 device, you should add address map functions, e.g. something like:
> 
> In machine/nmk112.h:
> 
> ```c++
> ...
> class nmk112_device : public device_t
> {
> public:
> ...
> 	void oki0(address_map &map) ATTR_COLD;
> 	void oki1(address_map &map) ATTR_COLD;
> ...
> private:
> ...
> 	void oki_map(address_map &map, unsigned which) ATTR_COLD;
> ...
> };
> ...
> ```
> 
> In machine/nmk112.cpp:
> 
> ```c++
> ...
> void nmk112_device::oki0(address_map &map)
> {
> 	oki_map(map, 0);
> }
> 
> void nmk112_device::oki1(address_map &map)
> {
> 	oki_map(map, 1);
> }
> 
> void nmk112_device::oki_map(address_map &map, unsigned which)
> {
> 	for (int i = 0; i < 4; i++)
> 	{
> 		map((i << 16) | page_offset(which, i), (i << 16) | 0xffff).bankr(m_samplebank[which][i]);
> 		if (paged)
> 			map(i << 8, (i << 8) | 0xff).bankr(m_tablebank[which][i]);
> 	}
> }
> ...
> ```
> 
> Then when you want use it in a driver, you can do something like:
> 
> ```c++
> ...
> void cave_state::donpachi(machine_config &config)
> {
> ...
> 	OKIM6295(config, m_oki[0], 4.220_MHz_XTAL/4, okim6295_device::PIN7_HIGH); // pin 7 not verified
> 	m_oki[0]->add_route(ALL_OUTPUTS, "mono", 1.60);
> 
> 	OKIM6295(config, m_oki[1], 4.220_MHz_XTAL/2, okim6295_device::PIN7_HIGH); // pin 7 not verified
> 	m_oki[1]->add_route(ALL_OUTPUTS, "mono", 1.0);
> 
> 	nmk112_device &nmk112(NMK112(config, "nmk112", 0));
> 	nmk112.set_rom0_tag("oki1");
> 	nmk112.set_rom1_tag("oki2");
> 	nmk112.set_page_mask(1 << 0);    // chip #0 (music) is not paged
> 	m_oki[0]->set_addrmap(0, nmk112, &nmk112_device::oki0);
> 	m_oki[1]->set_addrmap(0, nmk112, &nmk112_device::oki1);
> }
> ```
> 
> This isn’t tested, and I’m coding in a web browser, but I hope you can see the basic idea.

Simply crash when OKI chip tries to play sample because address map is installed before ROM related stuff is initialized.

Bank initialization function is should be moved from device_start?

cuavas:
> Simply crash when OKI chip tries to play sample because address map is installed before ROM related stuff is initialized.

Which part is crashing?  The address map function should be able to use `bankr` before the bank entries are set up.  Just setting up where the banked memory is located in the address map doesn’t cause bank entries to be accessed.  The address map function shouldn’t need to access anything that depends on having the ROM available.

> Bank initialization function is should be moved from device_start?

No, the actual part where you call `configure_entry` etc. to set up the bank entries should happen at `device_start` time because you won’t have all your object finders resolved before that.

cam900:
> Which part is crashing? The address map function should be able to use `bankr` before the bank entries are set up. Just setting up where the banked memory is located in the address map doesn’t cause bank entries to be accessed. The address map function shouldn’t need to access anything that depends on having the ROM available.

When tries to access sample space from method of above. Memory bank finder should be optional_memory_bank_array for use `bankr`?

cuavas:
> > Which part is crashing? The address map function should be able to use `bankr` before the bank entries are set up. Just setting up where the banked memory is located in the address map doesn’t cause bank entries to be accessed. The address map function shouldn’t need to access anything that depends on having the ROM available.
> 
> When tries to access sample space from method of above. Memory bank finder should be optional_memory_bank_array for use `bankr`?

`bankr` should work with `memory_bank_array_creator` or `optional_memory_bank_array`.  I’ll look at it later, maybe I’m forgetting something important.

cuavas:
`address_map` gets upset about object finders in maps where the base device isn’t the address map’s base device, and it doesn’t pick up the base device from the address space constructor.  This means you need extra trampoline address maps in the driver using `.m` to install the device’s address map.  It’s probably possible to work around this, but not something I care about doing now.

The wrapping for non-power-of-two region sizes is wrong, but it’s no worse than before.  It can be fixed another time.

PR #14271, opened 2025-10-05
Update astrcorp.cpp adding passwords in ASCII format

jbtx77gr:
-adding passwords in ASCII format


src/mame/astrocorp/astrcorp.cpp
```diff
@@ -149,6 +149,98 @@ MAGIC BOMB\LB5.1 121302S (magibomb_lb51 RUN OK)
 MAGIC BOMB\NB4.5 061402S (magibomb_nb45 freeze)
 MAGIC BOMB\NB6.1 260404S (magibomb_nb61 RUN OK CGA/VGA)
 
+***ASTRO PASSWORDS***
+
+Ra's Scepter
```
angelosa, line 154: I'd like this to have romset ID formats, not game names. They may have changed passwords between revisions.

jbtx77gr: Astro uses stable passwords across all versions of each game — I’ve never come across anything different.
This applies even to versions made for different regions: international, U.S.-only, European, or even Russian releases.

So, in the settings, the user simply selects one of the 8 available options:
0 means no password, and 1–7 correspond to the preset passwords already programmed for each game.

If I ever find a version that behaves differently, I’ll make sure to mention it.

angelosa: That's a QA thing: we prefer using romset IDs so it's easier to track down when the next guy touches this driver. If you are positive about being identical between revisions then the parent set is fine, but this needs to be noted too.

jbtx77gr: As I mentioned earlier, this has nothing to do with the ROM set IDs.
For example, the codes for Monkey Land can be used across all sets and all versions.
I’m not sure how to make it any clearer — I’ve handled dozens of boards from different countries, and they all had the same codes.

Personally, it doesn’t really matter to me since I have all the manuals.
So, if you simply don’t want this information included for the users, just let me know and I’ll close the PR.

angelosa: Well, Ra's Scepter is already an Astro PC game, not `astrcorp.cpp` ...

jbtx77gr: yes, and some games are no dumped yet such Magic 7...

PR #14275, opened 2025-10-06
snk/hng64*.cpp: Cleanups and Updates

cam900:
- Use template for tilemap info function
- Use logmacro.h for logging function
- Fix side effects issue for debugger read
- Fix save state support
- Reduce duplicates
- Reduce runtime tag lookups
- Fix namings
- Fix spacings

### DIFF
--- a/src/mame/snk/hng64_v.cpp
+++ b/src/mame/snk/hng64_v.cpp
@@ -1453,21 +1453,21 @@ void hng64_state::video_start()
 	m_old_tileflags[2] = -1;
 	m_old_tileflags[3] = -1;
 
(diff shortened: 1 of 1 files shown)


src/mame/snk/hng64_3d.ipp
```diff
@@ -5,14 +5,28 @@
...
+#define LOG_TEXTURE      (1 << 4)
+
+#define VERBOSE (0)
+
+#include "logmacro.h"
+
+#define LOG3D(...)           LOGMASKED(LOG_3D, __VA_ARGS__)
+#define LOGFRAMEBUFFER(...)  LOGMASKED(LOG_FRAMEBUFFER, __VA_ARGS__)
+#define LOGDISPLAYLIST(...)  LOGMASKED(LOG_DISPLAY_LIST, __VA_ARGS__)
+#define LOGTEXTURE(...)      LOGMASKED(LOG_TEXTURE, __VA_ARGS__)
```
cuavas, line 20: This isn’t really safe to do in something that’s going to get #included somewhere else.  Does this even need to be a .ipp file?  Could it just be another .cpp file to implement the renderer classes?

PR #14276, opened 2025-10-06
snk/hng64.cpp: Further cleanups from https://github.com/mamedev/mame/pull/14275

cam900:
- Fix possibly transparent pen issue
- Minor optimization
- Reduce duplicates
- Use BIT for single bit values


cuavas:
This has conflicts.  Also, please make sure you test your changes.  The previous pull request failed validation and caused the games to fail to start.

mamehaze:
I still *really* disagree with this BIT macro stuff, it does nothing but obfuscate.

If I'm looking at the memory in the debugger I see hex values, having to convert them to and from bit positions because the code uses BIT is not fun or helpful at all.

I know this is policy, but it's really fucking stupid.

PR #14277, opened 2025-10-06
apple2gs: Improve IRQ emulation 

arekkusu42:
This PR fixes multiple problems with IRQs:

* g65816: fix IRQ to be level triggered
(this one is the impetus for the rest of this PR)
apple2gs in MAME 0.281 does not hang if a programmer \<_raises hand_> makes a error \<_hand still raised_> in an IRQ handler and fails to clear the proper IRQ source. This leads to much hair-pulling later, when real hardware hangs.

* apple2gs: improve VGC/MegaII IRQ emulation
There were several subtle errors, which isn't surprising given [Apple's scarce documentation](http://archive.org/details/Apple_IIgs_Firmware_Reference_HiRes/page/n309/mode/2up). A pile of unit tests revealed exactly how each softswitch works.

* apple2gs: implement undocumented MegaII interaction
Reading the MegaII counters clears the VGC's SCB status bit. [Strange but true](https://www.brutaldeluxe.fr/documentation/cortland/v5/Cortland%20Custom%20ICs%20-%20Wayne%20Lowry%20-%20Preliminary%20notes%20-%2019860214.pdf#page=45), and inconvenient if you try to measure when SCB IRQs occur using VERTCNT...

* apple2gs: improve C019 VBL behavior
The [previous change](https://github.com/mamedev/mame/pull/14177) fixed the polarity.  Now it triggers on the proper scanline.

* apple2gs: remove IOUDIS, fix AN3 and paddle                                                                                                           
MegaII C046 includes the AN3 state.  Now it matches the DHIRES state.  Also stop triggering the paddle when writing to C071-7D ROM, and read the floating bus (verified on hardware.)

Testing:
I wrote a small suite of unit tests to cover all of the above (a lot of comments in the source):
[IRQTest_251004.zip](https://github.com/user-attachments/files/22725574/IRQTest_251004.zip)
...which exposes many failures in 0.281 (only the first failing subtest of each test is printed):
IRQTest_0281_before
Everything passes after this PR, matching hardware behavior verified on an Apple IIgs ROM 3:
IRQTest_ROM3_pass

Because this touches the 65816 core, I also briefly tested the SNES. A handful of carts (and SA-1 carts) continue to work as well as before (which is to say, mostly OK... ctrigger has some graphical errors, and [tophant is completely broken](https://snes.nesdev.org/wiki/Tricky-to-emulate_games#:~:text=Missing%20sound%20effects,%20freezes).) The NES has good [docs & homebrew tests](https://www.nesdev.org/wiki/CPU_interrupts#Delayed_IRQ_response_after_CLI,_SEI,_and_PLP) for IRQs... the SNES, less so?

Other than saving programmer sanity by making MAME behave the same as real hardware when given a broken IRQ handler, there is no obvious user-visible change after this PR. Rastan music [continues to play](https://github.com/mamedev/mame/commit/5649a74f633d8a0872d4fa0bb09340fc572c29e9) as before.

TODO:
* The 1-sec and quarter-sec IRQ frequency is not verified (hardware does not trigger 1-sec on any particular scanline)
* `adbmicro_p2_out()` only touches a subset of `machine_reset()`, this seems wrong?
* The ZipGS emulation needs to add a slowdown in `raise_irq()`, I have this in a separate WIP PR

### DIFF
--- a/src/mame/apple/apple2gs.cpp
+++ b/src/mame/apple/apple2gs.cpp
@@ -3287,7 +3287,7 @@ void apple2gs_state::a2gs_es5503_map(address_map &map)
     ADB microcontroller + KEYGLU emulation
 
     Huge thanks to Neil Parker's writeup on the ADB microcontroller!
-    http://www.llx.com/~nparker/a2/adb.html
+    http://llx.com/Neil/a2/adb.html
 ***************************************************************************/
 
 u8 apple2gs_state::adbmicro_p0_in()


rb6502:
Regarding the subset of machine_reset(), that's intentional.

For reasons of MAME's background as an arcade emulator where the F3 reset key is intended to work even on hardware with no notion of a reset button or switch, machine_reset() is expected to completely fix the machine state regardless of what you'd gotten it into beforehand.  In the A2 case, it resets many more things than a normal keyboard reset (Control-Reset or Control-Open Apple-Reset) does.

Control-Reset and Control-Open Apple-Reset do not do that.  On the schematics, not all that many chips actually see the reset signal.  On the II and II Plus only the 6502 and the slots get the reset signal.  On the IIe/IIc/IIgs there's more to it than that; HIRES and 80COL are forced off, among many other things, and the language card is forced to a known state.  Total Replay inadvertently became a test case for a lot of the corner cases there since you can use Control-Reset to get back to its menu from inside a game.

cuavas:
Does making the 65816 interrupts level-sensitive break anything else?  It’s used in several other systems and slot cards.

rb6502:
They are level sensitive on real hardware, and I haven't seen any regressions from this.  Still need to check the Acorn Communicator though.

cuavas:
> They are level sensitive on real hardware, and I haven't seen any regressions from this. Still need to check the Acorn Communicator though.

Yes, I know they’re level-sensitive IRL, but there have been plenty of cases of MAME code depending on load-bearing bugs elsewhere.

cuavas:
There’s a Tube device with a 65816, and some TV games use it as well.

arekkusu42:
I made a list of potentially affected systems:
-listdevices filtered on 65.?816: 23 hits
-listdevices filtered on 65.?802: 0 hits
-listdevices filtered on 5a22: 41 hits
[potentially_affected_systems.txt](https://github.com/user-attachments/files/22782711/potentially_affected_systems.txt)

I have not investigated any of these other than apple2gs* and snes.

What level of testing are you looking for here?  I could set up a few of these and sanity-check that a couple of roms "appear to still work"-- but unlike apple2gs and snes, I don't own the physical hardware so have no baseline to compare with.

rb6502:
IRQ problems usually manifest as the system simply not booting, so a very quick smoke test is fine.

Also, all of the 5A22 hits are SNES, so they're already covered.

angelosa:
fwiw `snes sgng` uses both irqs during gameplay. If that's broken you will get something akin to this: 
0000x

arekkusu42:
sgng was one of my tested carts. It's OK.

arekkusu42:
Smoke-testing a few systems:
	abl4play (known: sound) boots, plays, "mountain bike rally" in-game tilemap scroll looks broken
	accomm (known: graphics) boots, calculator runs and works
	my1stddr (known: sound) boots, plays
	shtscore (known: sound) boots, plays
	trkfldch (known: sound) boots, attract mode runs, in-game tilemap scroll looks broken
	tv965 (known: sound) BLACK SCREEN, looks like looping around an NMI

Then I backed out the g65816 changes, rebuilt, re-ran problematic systems:
	abl4play	in-game tilemap scroll still broken
	trkfldch	attract mode tilemap scroll still broken
	tv965	BLACK SCREEN, looks like looping around an NMI
I wasn't able to locate information on how to configure tv965 to do anything.

arekkusu42:
Also FWIW, I had looked at these snes carts:
ctrigger
dbzhypdj
gradius3
kirby3
scvania4u
sf2tua
sgngu
smetroidj
smwu
starfox
testcart
tophant
zelda3u
Of these, the only interesting one was starfox, because extra instrumentation showed that it was re-triggering IRQ immediately after RTI.  But that happened before this PR too.

angelosa:
> tv965 BLACK SCREEN, looks like looping around an NMI

It's actually putting stuff in videoram from this old build I have, however no chargen is hooked up so no graphic is shown (missing ROM?)
image

arekkusu42:
A little more testing, of slot devices:
	bbcb -tube rc65816 (known: graphics) boots to "ReCo6502 65C816 512K", \*SETUP works, BASIC hello world works.
	c64 -exp supercpu -floppydisk scpukicks.zip boots, but to C64 BASIC V2. The -debug devices windows shows the exp>supercpu was loaded, but I'm thus far unable to figure out how to boot it into SUPERCPU DOS. LOAD "\*",8,1 of the scpukicks demo just hangs.  I get the same behavior when I back out the g65816 changes. Probably my user error? (a [wiki entry](https://wiki.mamedev.org/index.php/System-Specific_Setup_and_Information) from someone who understands the c64 would be helpful.) 

So, the result of that testing is:
	* tv965 should maybe be marked not working?
	* abl4play / trkfldch should maybe be marked imperfect graphics?
	* I don't see any obvious regressions from the g65816 IRQ changes in this PR.

arekkusu42:
...and I see [now](http://github.com/mamedev/mame/commit/a7c9b298d3daf8ae3d6da74cef44a76278fa10ea) that tv965 boots up to "0 001 001  FDX"...

PR #14280, opened 2025-10-06
seta/champbwl.cpp: New WORKING clone

sergiopolog:
New WORKING clone
Doraemon Surotto [hammy]

I added it as parent set as previously dumped one seems to be a prototype. It works on the exact same PCB but no hacks and external wirings are present. [https://i.imgur.com/vLPIgh1.jpeg](https://i.imgur.com/vLPIgh1.jpeg)

Sound ROMs are the same but others differ from previous set. I update file names for these ones according to the labels found on this PCB.

The game title for this one is "Doraemon Surotto"  (Doraemon Slot) and some differences can be spotted during gameplay.
Also, DSW settings are different. It uses two banks instead of only 1 and options changed from previous set.

Game freezes when winning, probably due to a bad hopper configuration, but I cannot figure out how to make it work.
Additionally, the game uses 3 different "coin inputs":  "Medal", "¥10" and "¥100" as can be seen in the Service Mode (F2), and I assigned `IPT_COIN2`, `IPT_COIN3` and `IPT_COIN4` for them, respectively.  I tried to use `IPT_COIN1` but for some reason, it doesn't work (it does nothing).


angelosa:
> Game freezes when winning, probably due to a bad hopper configuration

Is it a thing for the prototype set? Technically (optional) gambling features that doesn't cooperate should be `MACHINE_NOT_WORKING` but tbh I'm leaning towards making an own (yellow) flag, being imo not really _fatal_ ...

sergiopolog:
> > Game freezes when winning, probably due to a bad hopper configuration
> 
> Is it a thing for the prototype set? Technically (optional) gambling features that doesn't cooperate should be `MACHINE_NOT_WORKING` but tbh I'm leaning towards making an own (yellow) flag, being imo not really _fatal_ ...

No, it's not. On prototype set there is no reward, I think. Screen during gameplay changes a bit. 
On prototype set each "credit" gives you 3 plays and you need to reach "10 point" in total to win, but if you achieve it, game simply shows a win screen.
New set seems to be a purely slot machine and each credit gives you a single play and gives you a reward if you win (that's when the game freezes)

Prototype screen gameplay:
[https://i.imgur.com/iaam1Ya.png](https://i.imgur.com/iaam1Ya.png)

New set screen  (that section of icons to reach 10 points is missing on this one):
[https://i.imgur.com/uaW8z5I.png](https://i.imgur.com/uaW8z5I.png)


Regarding `MACHINE_NOT_WORKING` flag, let me know if I should add it or any other flag. I think something like `MACHINE_IMPERFECT_HOPPER` or similar should fit better on these cases, as video and sound emulation is perfect.

angelosa:
> Regarding MACHINE_NOT_WORKING flag, let me know if I should add it or any other flag. I think something like MACHINE_IMPERFECT_HOPPER or similar should fit better on these cases, as video and sound emulation is perfect.

I think it should be something more broad, like `MACHINE_IMPERFECT_GAMBLING`, which would translate in UI as `imperfectly emulated features: gambling`.

List of things on top of my head beyond hopper/ticket dispensers:
- stealth modes;
- key-in remote control;
- password system not yet documented/dealt with (worked around with a default backup save);
- unclear mechanism to free out of credit limit i.e. that doesn't cover an operator reset or credit limit bump (`86lions`, probably just hopper);
- lamps artwork not hooked up, if non mandatory i.e. the video display already highlights relevant gameplay portions;
- selectable payout options not yet handled (`magic10` in MKII supergame mode);

Case study `tinkerbl` / `blicks`: currently `MACHINE_NOT_WORKING` because it's lacking hopper and EEPROM.
https://github.com/mamedev/mame/blob/94e3ee38d44d51c901c86132e63c66302745e9a0/src/mame/sega/segasm1.cpp#L21-L23
There's an error with this QA priority order: puts hopper _before_ EEPROM. And chances are that any system types (arcade/console/computer/your washing machine) can possibly have a method to save settings, a ticket dispenser is 99% confined in selected Arcade sub-genres. Notice that MNW is correct in this case (not being able to save and being greeted everytime with a RAM BAD at startup + not being able to .inp playback/record are deal breaking), but once this bug is fixed is it really a bad idea to consider it "working" even without the hopper?

src/mame/seta/champbwl.cpp
```diff
@@ -434,7 +442,83 @@ static INPUT_PORTS_START( champbwl )
 INPUT_PORTS_END
 
 static INPUT_PORTS_START( doraemon )
```
cuavas: Can you use `PORT_INCLUDE` to avoid repeating the common parts shared between the two games?

src/mame/seta/champbwl.cpp
```diff
@@ -434,7 +442,83 @@ static INPUT_PORTS_START( champbwl )
...
+	PORT_DIPSETTING(    0x03, "2" )
+	PORT_DIPSETTING(    0x05, "3" )
+	PORT_DIPSETTING(    0x01, "4" )
+	PORT_DIPSETTING(    0x06, "5" )
+	PORT_DIPSETTING(    0x02, "6" )
+	PORT_DIPSETTING(    0x04, "7" )
+	PORT_DIPSETTING(    0x00, "8" )
+	PORT_DIPNAME( 0x08, 0x08, "Win Wave" )													PORT_DIPLOCATION("SW2:4")
+	PORT_DIPSETTING(    0x08, "Small" )
+	PORT_DIPSETTING(    0x00, "Big" )
```
cuavas: This is probably a poor translation of “payout rate variation” or similar.

src/mame/seta/champbwl.cpp
```diff
@@ -434,7 +442,83 @@ static INPUT_PORTS_START( champbwl )
...
+	PORT_DIPSETTING(    0x00, DEF_STR( On ) )
+	PORT_DIPNAME( 0x40, 0x40, DEF_STR( Unknown ) )											PORT_DIPLOCATION("SW2:7")
+	PORT_DIPSETTING(    0x40, DEF_STR( Off ) )
+	PORT_DIPSETTING(    0x00, DEF_STR( On ) )
+	PORT_DIPNAME( 0x80, 0x80, DEF_STR( Unknown ) )											PORT_DIPLOCATION("SW2:8")
+	PORT_DIPSETTING(    0x80, DEF_STR( Off ) )
+	PORT_DIPSETTING(    0x00, DEF_STR( On ) )	
+
+	PORT_START("IN0")   // f000
+	PORT_BIT( 0x01, IP_ACTIVE_LOW,  IPT_SERVICE2 )  PORT_NAME( "Data Clear" )
```
cuavas: This should probably be `IPT_MEMORY_RESET`.

src/mame/seta/champbwl.cpp
```diff
@@ -434,7 +442,83 @@ static INPUT_PORTS_START( champbwl )
...
+	PORT_BIT( 0x02, IP_ACTIVE_LOW,  IPT_COIN4    )  PORT_NAME( "¥100" )
+	PORT_BIT( 0x04, IP_ACTIVE_LOW,  IPT_COIN3    )  PORT_NAME( "¥10" )
+	PORT_BIT( 0x08, IP_ACTIVE_LOW,  IPT_COIN2    )  PORT_NAME( "Medal" )
```
cuavas: This seems to be missing `IPT_COIN1` – it’s generally desirable to start the coin inputs from 1.

src/mame/seta/champbwl.cpp
```diff
@@ -434,7 +442,83 @@ static INPUT_PORTS_START( champbwl )
...
+	PORT_BIT( 0x10, IP_ACTIVE_LOW,  IPT_UNKNOWN  )
+	PORT_BIT( 0x20, IP_ACTIVE_LOW,  IPT_UNKNOWN  )
+	PORT_BIT( 0x40, IP_ACTIVE_HIGH, IPT_SERVICE3 )  PORT_NAME( "Freeze" )
```
cuavas: Should this be `PORT_TOGGLE`?

src/mame/seta/champbwl.cpp
```diff
@@ -766,7 +876,9 @@ ROM_END
...
-GAME( 1989,  champbwla, champbwl, champbwl, champbwl, champbwl_state, empty_init, ROT270, "Seta / Romstar Inc.", "Championship Bowling (location test)",   MACHINE_SUPPORTS_SAVE )
+GAME( 1989,  champbwl,  0,        champbwl,  champbwl,  champbwl_state, empty_init, ROT270, "Seta / Romstar Inc.", "Championship Bowling",                   MACHINE_SUPPORTS_SAVE )
+GAME( 1989,  champbwla, champbwl, champbwl,  champbwl,  champbwl_state, empty_init, ROT270, "Seta / Romstar Inc.", "Championship Bowling (location test)",   MACHINE_SUPPORTS_SAVE )
 
-GAME( 1993?, doraemon,  0,        doraemon, doraemon, doraemon_state, empty_init, ROT0,   "Sunsoft / Epoch",     "Doraemon no Eawase Montage (prototype)", MACHINE_SUPPORTS_SAVE ) // year not shown, datecodes on pcb suggests late-1993
+// TODO: parent set freezes when winning during gameplay, probably due to bad hopper mapping. Needs further checks:
+GAME( 1993?, doraemon,  0,        doraemon,  doraemon,  doraemon_state, empty_init, ROT0,   "Sunsoft / Epoch",     "Doraemon Surotto",                       MACHINE_SUPPORTS_SAVE ) // year not shown, datecodes on pcb suggests late-1993
```
cuavas: It should be “Doraemon Slot” – we don’t back-transliterate loanwords.

src/mame/seta/champbwl.cpp
sergiopolog: Created a new port config calle `doraemon_base` which combine the common parts. Every set has its own config below for the specific parts.

src/mame/seta/champbwl.cpp
sergiopolog: Updated to "Payout rate variation"

src/mame/seta/champbwl.cpp
sergiopolog: Fixed

src/mame/seta/champbwl.cpp
sergiopolog: I updated the code to use `IPT_COIN1` for "¥100",  `IPT_COIN2` for "¥10"  and  `IPT_COIN3` for "Medal",   but Coin1 input doesn't work for an unknown reason (key 5 by default),  while Coin2 and Coin3 work fine.
On `doraemonp`, Coin1 works fine. I don't know why it doesn't work for this set... Any help on this?

src/mame/seta/champbwl.cpp
sergiopolog: Updated

src/mame/seta/champbwl.cpp
sergiopolog: Fixed

src/mame/seta/champbwl.cpp
cuavas: I’ll take a look, it should work.

cuavas:
I took a look at this.  The reason the ¥100 coin sometimes doesn’t work is that the corresponding coin lockout output is sometimes set.  Even when it responds and plays the credit sound, it won’t actually coin up, so there’s more wrong.  It isn’t just bad hopper connections.  It’s easier to just mark it not working for now until someone has more time to debug it.  It could be related to whatever the missing peripheral at F003 is.

PR #14282, opened 2025-10-07
seibu/seibuspi.cpp: Split driver state class per hardware configurations, Minor cleanups:

cam900:
- Reduce preprocessor defines
- Fix tilemap draw routine with screen cliprect
- Reduce literal tag usages
- Reduce duplicates
- Fix side effects for debugger reads
- Fix namings

seibu/seibucats.cpp: Fix crash when reading inputs


cuavas:
CATS should also be updated to include the standard mahjong matrix since it’s just the standard matrix with the bits in a different order.

PR #14283, opened 2025-10-07
seibu/feversoc.cpp: Minor cleanups

cam900:
- Fix sprite wraparound method
- Add notes for sprite hardware
- Simplify gfx decode layout
- Fix spacings
- Fix namings

### DIFF
--- a/src/mame/seibu/feversoc.cpp
+++ b/src/mame/seibu/feversoc.cpp
@@ -16,7 +16,7 @@ TODO:
 - Layout including lamps
 - Hopper only works in "COIN HOPPER" mode
 - Do button 5 or remaining DIPs actually do anything outside service mode?
-- Make device of RISE11 sprite hardware that also used in seibu/seibuspi.cpp
+- Make device of RISE11 sprite hardware also used in seibu/seibuspi.cpp
   and seibu/seibucats.cpp
 
 ============================================================================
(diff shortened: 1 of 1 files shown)


src/mame/seibu/feversoc.cpp
```diff
@@ -136,110 +141,111 @@ class feversoc_state : public driver_device
...
-
-		for(dx=0;dx<w;dx++)
-			for(dy=0;dy<h;dy++)
-				m_gfxdecode->gfx(0)->transpen(bitmap,cliprect,spr_offs++,colour,0,0,(sx+dx*16),(sy+dy*16),0x3f);
+		int sy = m_spriteram[offs + 1] & 0x01ff;
+		int sx = (m_spriteram[offs + 1] & 0x01ff0000) >> 16;
+		const u32 colour = (m_spriteram[offs + 0] & 0x003f0000) >> 16;
+		const u8 w = ((m_spriteram[offs + 0] & 0x07000000) >> 24) + 1;
+		const u8 h = ((m_spriteram[offs + 0] & 0x70000000) >> 28) + 1;
+		// TODO: flip sprite?
```
angelosa: The unused bits could be anything in the Seibu portfolio, adding this note has no effective value.

cam900: That because original RISE11 (also used at rfjet and clones in seibu/seibuspi.cpp; including functionally compatible RISE10 and SEI252) has per-sprite flip X/Y bit in sprite RAM, but seems to not used in feversoc.

angelosa: So the note should be `// TODO: merge implementation with Seibu SPI/Cats`, and I don't think it's necessary because it's already repeated in header. Game probably also uses DMA start address etc. as per the norm since legionna.cpp anyway ...

PR #14289, opened 2025-10-08
mcs51: fix destination RAM type in XCHD A, @R0/@R1

fabio-d:
- The initial value is read from internal RAM (IRAM_IR).
- The final value must be written to the internal RAM too (IRAM_IW instead of IRAM_W).

----

I and @jyaif have found an issue in the mcs51's implementation of the `XCHD` ("Exchange Digit") instruction: it reads the old value with `IRAM_IR` but writes the new value with `IRAM_W`.

Instead, it should have written the new value symmetrically with `IRAM_IW`:
XCHD documentation

We found the issue while investigating a bad Dynamical Redefinable Character Set (DRCS) rendering on the minitel2 (`src/mame/philips/minitel_2_rpic.cpp`).

Before this patch:
before

After this patch:
after

(from [MiniPavi](https://www.minipavi.fr/) service -> code `6212*DRCS` -> image 1)

cc @jfdelnero


mamehaze:
@fabio-d I'm guessing this needs applying to axc51/axc51ops.hxx as well, as that was cloned from mcs51 (as it was deemed cleaner than subclassing)

PR #14290, opened 2025-10-08
xavix progress - mostly sound

mamehaze:
Attempted to improve XaviX audio emulation with hardware studies [ramacat]
- sample loop start positions are no longer being treated as left/right channel start positions
- stop conditions are longer inverted, key-off no longer being treated as stop etc.
- attempted to implement various envelope effects and more (some use wave in the ROM, others are fixed function)
- some things sound much better (taitons2 etc.) while at times some notes hang until a new sample is played in other games (popira for example)

Added Save State support to the XaviX sound hardware emulation [David Haywood]

Fixed some regressions with SEEPROM handling in the driver [David Haywood]

### DIFF
--- a/src/mame/tvgames/xavix2.cpp
+++ b/src/mame/tvgames/xavix2.cpp
@@ -499,18 +499,14 @@ void xavix2_state::pio_update()
 
 void naruto_state::pio_update()
 {
-	if (BIT(m_pio_mask_out, 21))
-		m_i2cmem->write_sda(BIT(m_pio_dataw, 21));
-	if (BIT(m_pio_mask_out, 20))
-		m_i2cmem->write_scl(BIT(m_pio_dataw, 20));
+	m_i2cmem->write_sda(BIT(m_pio_mask_out, 21) ? BIT(m_pio_dataw, 21) : 1);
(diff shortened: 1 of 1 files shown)


mamehaze:
setting as draft, because while I've done a decent amount of cleanup on the submission, there are probably still things that could be improved.

mamehaze:
(also I'm fine with some stuff being a bit worse for now, as some of the guesses I made when I first tried to work this out were terrible.  this sets things up for progress in the right direction)

JimCarlTay:
Just tested this pull request and I can easily notice the sound improvements here. Meanwhile, is this PR limited only to the xavix and xavix_2000 drivers?

I'm interested in checking out if EEPROM error fixes can also be done for xavix2 and xavix_2002.

mamehaze:
the other stuff will be investigated, although anything attached to the SuperXaviX extra I/O ports is still maybe going to be rough, as the port direction stuff isn't fully understood / hooked up there anyway (although many of those still hook the I2C to the standard ports, not the extended ones)

at the very least I'd like to get that side of things back to where it was before the i2c change, as we need as many cases back to somewhat functional states to test / improve the sound further.

mamehaze:
there have been some significant functional improvements since the code in this PR, but getting this merged will serve as a base for those!

PR #14293, opened 2025-10-08
Bilinear luma filtering for SEGA Model 2 textured render path

Nitch2024:
Adding support for bilinear filtering for the model 2 rasterizer to look closer to the arcade rendering which has bilinear filtering.

### DIFF
--- a/src/mame/sega/model2rd.ipp
+++ b/src/mame/sega/model2rd.ipp
@@ -154,10 +154,15 @@ void MODEL2_FUNC_NAME(int32_t scanline, const extent_t& extent, const m2_poly_ex
 		int32_t u = uoz * z;
 		int32_t v = voz * z;
 		int  tr, tg, tb;
-		u16  t;
+		u32  t, tex1, tex2, tex3, tex4, frac1, frac2, frac3, frac4;
+#if defined(MODEL2_TRANSLUCENT)
+		u32  alp, alp1, alp2, alp3, alp4;
+#endif
 		u8 luma;
 		int u2;
 		int v2;
+		int du = 1;
+		int dv = 1;
 
 #if defined(MODEL2_CHECKER)
 		if ( ((x^scanline) & 1) == 0 )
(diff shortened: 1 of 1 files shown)


cuavas:
What’s the reference for this?  I’m leery of changes that just “look better”.

Also, the formatting in the added code is wildly inconsistent.

rb6502:
The if statements are being expanded to cover more than one line, and the rest of the file uses Allman-style bracing in those cases, so the formatting looks OK to me.

cuavas:
> The if statements are being expanded to cover more than one line, and the rest of the file uses Allman-style bracing in those cases, so the formatting looks OK to me.

Dude

```c++
		tex1 = get_texel(tex_x, tex_y, u2, v2, sheet );
		tex2 = get_texel(tex_x, tex_y, (u2 + du) % tex_width, v2, sheet);
		tex3 = get_texel(tex_x, tex_y, u2, ( v2 + dv ) % tex_height, sheet);
```

Stray space before closing parenthesis on one line, “English style” parenthesis spacing on the next, and a mixture of “English” and “European” on the line after.

galibert:
The bilinear filtering on the luminance is real, fwiw.  There's no filtering on the color itself, the code looks correct there too.

cuavas:
> The bilinear filtering on the luminance is real, fwiw. There's no filtering on the color itself, the code looks correct there too.

Do we know much about the exact implementation of the filtering?

galibert:
Not really.  It's tucked into big VLSIs with amusing names like "Mexico" and "Texas".

rb6502:
I know people have done limited testing on hardware to analyze what it's doing, because the last time I pressed X to doubt that any filtering exists I was presented with the PNGs.  But that's just from spotting usable places in the actual games, not running trojan programs.

gm-matthew:
Sega Model 2 actually performs trilinear filtering; you can see this in action during the course select screen in Daytona USA, as the texture LOD is gradually increased so that the course maps morph into detail ([video link](https://youtu.be/-Zh6DAq-rJ8?t=22)). Still, bilinear filtering (using the highest quality mipmap for now) is a step in the right direction.

An issue with regular bilinear filtering is that transparent texels can blend into the visible pixels, which would be especially pronounced on Model 2 as transparent texels have the maximum luma value. Here's a screenshot from my own build with this commit included; notice how the edges of the trees have highlights from blending with the transparent texels:



This does not occur on a real Model 2 board because it actually uses a technique to prevent the transparent texels from blending with the visible ones. Model 3, which had its video hardware designed by the same team that developed the texture mapping unit for Model 2, uses the same technique; this [post](https://web.archive.org/web/20181014114111/https://www.supermodel3.com/Forum/viewtopic.php?f=7&t=1585) on the old Supermodel forum discusses the issue.

> Not really. It's tucked into big VLSIs with amusing names like "Mexico" and "Texas".

That's fascinating to know, how did you discover this? It fits in with how on Model 3 the video ASICs are named after planets and on Hikaru they are named after continents.

Nitch2024:
> Stray space before closing parenthesis on one line, “English style” parenthesis spacing on the next, and a mixture of “English” and “European” on the line after.

Yes for some reason I missed that. Thanks, I will fix it.

> Sega Model 2 actually performs trilinear filtering; you can see this in action during the course select screen in Daytona USA, as the texture LOD is gradually increased so that the course maps morph into detail ([video link](https://youtu.be/-Zh6DAq-rJ8?t=22)). Still, bilinear filtering (using the highest quality mipmap for now) is a step in the right direction.

The effect looks trilinear indeed, need to check if this is a real single polygon or two blended polygons. This can also be done with fading in and out two superimposed bilinear maps.

> An issue with regular bilinear filtering is that transparent texels can blend into the visible pixels, which would be especially pronounced on Model 2 as transparent texels have the maximum luma value. Here's a screenshot from my own build with this commit included; notice how the edges of the trees have highlights from blending with the transparent texels:

Thanks, I implemented a similar algorithm to SuperModel and that works pretty good to reduce the highlights. Working on fixing a similar repeat issue and then will submit new code.

cuavas:
> Sega Model 2 actually performs trilinear filtering; you can see this in action during the course select screen in Daytona USA, as the texture LOD is gradually increased so that the course maps morph into detail ([video link](https://youtu.be/-Zh6DAq-rJ8?t=22)). Still, bilinear filtering (using the highest quality mipmap for now) is a step in the right direction.

Should we just hold off until you can do a proper implementation of trilinear filtering?

Nitch2024:
> Should we just hold off until you can do a proper implementation of trilinear filtering?

There would be work that needs to happen to support mipmaps first if I read the comments correctly in model2.cpp. So until that happens, I recommend to move forward with bilinear.

"    TODO:
    - Mip Mapping still needs to be properly sorted in the renderer;"

rb6502:
Can the specialized texture clamping discussed in the linked Supermodel post be applied so the trees don't look weird?

Nitch2024:
Yes it is already implemented and fixed the issue reported above in Daytona.

image

rb6502:
Yeah, that's much better.  Thanks!

PR #14296, opened 2025-10-08
sinclair/specnext_dma.cpp: Implemented prescaler

holub:


src/devices/machine/z80dma.h
```diff
@@ -112,14 +127,18 @@ class z80dma_device :   public device_t,
...
+	emu_timer *m_timer;
 
+	u16 &REG(unsigned m, unsigned s) noexcept { return m_regs[REGNUM(m, s)]; }
 	static constexpr unsigned REGNUM(unsigned m, unsigned s) { return (m << 3) + s; }
 
 	u16 m_addressA;
 	u16 m_addressB;
 	u16 m_count;
 	u16 m_byte_counter;
+	int m_dma_seq;
```
rb6502, line 141: Is this really externally visible, or is this some specific specnext thing?  In the latter case I'd prefer there be a derived device that exposes this via an accessor function instead of just having it out there for all Z80DMA users.

holub: it's not externally but 'protected', made it visible for the next's child

> Is this really externally visible, or is this some specific specnext thing? In the latter case I'd prefer there be a derived device that exposes this via an accessor function instead of just having it out there for all Z80DMA users.

PR #14300, opened 2025-10-08
acorn/bbcb.cpp, acorn/bbcbp.cpp: Make cm62024.bin ROM optional

### LINKED ISSUE #14297: BBC B Micro emulation not working unless a ROM file named cm62024.bin is present.
**Affected commit**: d26cb4aa3d235b0e9b908ab1b49ff9bf087c5138

**Problem**: 
The BBC Micro drivers (bbcb.cpp, bbcbp.cpp) now require cm62024.bin (Voice Synthesis Module ROM) as a mandatory ROM file. This prevents the emulator from running without it, even though this ROM is only needed for speech synthesis features which most software doesn't use.

**Expected behavior**: 
The cm62024.bin ROM should be marked as ROM_LOAD_OPTIONAL rather than ROM_LOAD, similar to how other optional peripherals are handled in MAME.

**Impact**:
- Users without this ROM file cannot run BBC Micro emulation
- The VSM region is already allocated with ROMREGION_ERASE00, so it will fill with zeros if the ROM is missing
- Most BBC Micro software does not use speech synthesis

**Proposed fix**:
Change all instances of:
```cpp
ROM_LOAD("cm62024.bin", 0x3c000, 0x4000, CRC(98e1bf9e) SHA1(b369809275cb67dfd8a749265e91adb2d2558ae6))
```

To:
```cpp
ROM_LOAD_OPTIONAL("cm62024.bin", 0x3c000, 0x4000, CRC(98e1bf9e) SHA1(b369809275cb67dfd8a749265e91adb2d2558ae6))
```

This would allow BBC Micro emulation to work for users who don't have the speech ROM, while still supporting it for those who do.

**Additional context**:
- The commit message mentions "TMS5220 speech is now working" which suggests speech synthesis was the focus of this change
- The cm62024.bin ROM is specifically for the Voice Synthesis Module
- Most BBC Micro games and software don't require speech synthesis
- I have verified this fix works correctly by testing with the modified code

rebroad:
**Problem**:
The BBC Micro drivers currently require cm62024.bin (Voice Synthesis Module ROM) as mandatory, preventing emulation from running without it, even though this ROM is only needed for speech synthesis features which most BBC Micro software doesn't use.

**Solution**:
Changed `ROM_LOAD` to `ROM_LOAD_OPTIONAL` for cm62024.bin in both bbcb.cpp and bbcbp.cpp.

**Benefits**:
- Allows BBC Micro emulation to work without the speech ROM
- Still supports speech synthesis when the ROM is present
- The VSM region is already allocated with `ROMREGION_ERASE00`, so it correctly fills with zeros when the ROM is missing (matching hardware behavior for systems without speech hardware)

**Testing**:
- Verified that BBC Micro emulation works correctly without cm62024.bin
- Confirmed that MAME displays "OPTIONAL cm62024.bin NOT FOUND" message and continues running
- Tested with Spy Hunter and other BBC Micro software

Fixes #14297


ajrhacker:
`ROM_LOAD_OPTIONAL` is more or less deprecated, so I don't recommend this change. A better idea would be to move the ROM definition into the `bbc_speech` slot device.

Pernod70:
The ROM is intentionally in the driver so that we know which machines used either the UK or US speech PHROM, The ROM has been in the romset for over 10 years without issue, and moving it to `bbc_speech` would mean creating `bbc_speech_uk` and `bbc_speech_us` devices, which would obviously cause the OP more issues!

There's nothing to fix here, and no valid reason for this change.

cuavas:
`ROM_OPTIONAL` is a liability.  It gives no indication when the ROM is missing and things just don’t work.

The correct solution *is* to make separate UK and US speech devices.

Issue #12468, opened 2024-06-11
MAME failing self test in Apple 2c with Rom 0 (apple2c0) firmware


univta0001:
### MAME version

0.266 (mame0266)

### System information

Windows 10 Home 22H2
Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz   3.40 GHz
16.0 GB

### INI configuration details

```INI
#
# UI SEARCH PATH OPTIONS
#
historypath               history;dats;.
categorypath              folders
cabinets_directory        cabinets;cabdevs
cpanels_directory         cpanel
pcbs_directory            pcb
flyers_directory          flyers
titles_directory          titles
ends_directory            ends
marquees_directory        marquees
artwork_preview_directory "artwork preview;artpreview"
bosses_directory          bosses
logos_directory           logo
scores_directory          scores
versus_directory          versus
gameover_directory        gameover
howto_directory           howto
select_directory          select
icons_directory           icons
covers_directory          covers
ui_path                   ui

#
# UI MISC OPTIONS
#
system_names              
skip_warnings             0
unthrottle_mute           0

#
# UI OPTIONS
#
infos_text_size           0.75
font_rows                 30
ui_border_color           ffffffff
ui_bg_color               ef101030
ui_clone_color            ff808080
ui_dipsw_color            ffffff00
ui_gfxviewer_color        ef101030
ui_mousedown_bg_color     b0606000
ui_mousedown_color        ffffff80
ui_mouseover_bg_color     70404000
ui_mouseover_color        ffffff80
ui_selected_bg_color      ef808000
ui_selected_color         ffffff00
ui_slider_color           ffffffff
ui_subitem_color          ffffffff
ui_text_bg_color          ef000000
ui_text_color             ffffffff
ui_unavail_color          ff404040

#
# SYSTEM/SOFTWARE SELECTION MENU OPTIONS
#
hide_main_panel           0
use_background            1
skip_biosmenu             0
skip_partsmenu            0
remember_last             1
last_used_machine         88games
last_used_filter          Unfiltered
system_right_panel        image
software_right_panel      image
system_right_image        snap
software_right_image      snap
enlarge_snaps             1
forced4x3                 1
info_audit_enabled        0
hide_romless              1
```

### Emulated system/software

mame apple2c0 -resolution0 800x800 -skip_gameinfo

### Incorrect behaviour

When the mame runs the self test for the Apple 2C rom 0x00, it is failing in the self test with the error message "GLU"
[...]

PR #14307, opened 2025-10-11
apple2: cleanup IOUDIS

arekkusu42:
This PR fixes a recent regression, and cleans up IOUDIS:

* apple2e: fix regression in 85e73ca reintroducing #12468
In 0.281, apple2c0 fails the GLU self-test again.  Reverting the RDIOUDIS polarity fixes it, and DGR still works, using examples [one](http://devonhubner.org/Applesoft_BASIC_Double_Low_Resolution_Graphics/) or [two](http://github.com/ibrumby/Crossrunner-Issues/issues/49).

* apple2e: IOUDIS and RDDHIRES only exist on the IIc
I did some [archeology](http://forums.bannister.org/ubbthreads.php?ubb=showflat&Number=115202#Post115202) here, and the TLDR is that Apple copy-pasted [wrong information](http://archive.org/details/apple-iie-technical-reference-manual-1987-edition/page/30/mode/2up) into the IIe Technical References, so everyone [was confused](http://github.com/mamedev/mame/commit/6a4854bbc0905281ff29c58457b2d7723a23373e).  AppleWin has a [similar issue](http://github.com/AppleWin/AppleWin/issues/636#issuecomment-480530115) open, which I've verified on a physical Apple IIe Enhanced.  But why stop there, when we can visualize the entire softswitch range at once:

Switches_apple2ee_HW
[Switches_251013.zip](https://github.com/user-attachments/files/22894536/Switches_251013.zip) (updated [per thread](http://www.applefritter.com/content/help-wanted-testing-softswitches#comment-113310))

This uses carefully controlled floating bus readback to indicate which softswitches (or the low 7 bits) are floating, which on the IIe is C020-C08F.  We can see that C07E/7F (and mirrors) are entirely floating, which definitively tells us that RDIOUDIS and RDDHIRES don't exist.

Here's 0.281 before this PR (via `apple2ee -gameio ""`, to match my hardware)
Switches_apple2ee_281_before
C078-7F clearly show that only the low 7 bits are floating, which was wrong.

* apple2video: rename dhires_w to reduce confusion
This is merely naming clarification.  I'm not sure if the IIc's RDDHIRES polarity is actually correct, I'm [looking for someone to test it](http://www.applefritter.com/content/help-wanted-testing-softswitches).

TODO:
* apple2c* allows -gameio compeyes (and it continues to function after this PR), but in reality this is [physically impossible to connect](http://www.digital-vision-inc.com/products/CE-AppleII/CE_Apple_pic.jpg).
* The SWITCHES test also shows errors with the C05x video softswitches. This happens because the (page2/mixed/lores etc) video mode is changed prior to the floating bus read, and gets random data (instead of the controlled HGR black/white.)  On real IIe hardware, the video switches are latched with a ~2 cycle delay, after the floating bus readback, and the test is completely deterministic (assuming you don't press a key while it is running.)

### DIFF
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -1786,11 +1786,11 @@ void apple2e_state::do_io(int offset, bool is_iic)
 			switch (offset)
 			{
 				case 0x5e:  // SETDHIRES
-					m_video->dhires_w(0);
+					m_video->an3_w(0);
 					break;
 
 				case 0x5f:  // CLRDHIRES
-					m_video->dhires_w(1);
+					m_video->an3_w(1);
 					break;
 			}
 		}
--- a/src/mame/apple/apple2gs.cpp
+++ b/src/mame/apple/apple2gs.cpp
@@ -1416,13 +1416,13 @@ void apple2gs_state::do_io(int offset)
 		case 0x5e: // AN3 off, DHIRESON
 			m_an3 = false;
 			m_gameio->an3_w(0);
-			m_video->dhires_w(0);
+			m_video->an3_w(0);
 			break;
 
 		case 0x5f: // AN3 on, DHIRESOFF
 			m_an3 = true;
 			m_gameio->an3_w(1);
-			m_video->dhires_w(1);
+			m_video->an3_w(1);
 			break;
 
 		case 0x69:  // 16-bit access to STATEREG
--- a/src/mame/apple/apple2video.cpp
+++ b/src/mame/apple/apple2video.cpp
@@ -133,7 +133,12 @@ void a2_video_device::res_w(int state)
 	m_hires = state;
 }
 
-void a2_video_device::dhires_w(int state)
+void a2_video_device::an2_w(int state)
+{
+	m_an2 = state;
+}
+
+void a2_video_device::an3_w(int state)
 {
 	// select double hi-res
 	screen().update_now();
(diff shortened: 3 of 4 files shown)


src/mame/apple/apple2e.cpp
```diff
@@ -2414,13 +2420,9 @@ void apple2e_state::c000_w(offs_t offset, u8 data)
...
 			}
 			do_io(offset, false);    // make sure it also side-effect resets the paddles as documented
-			break;
-
-		case 0x7e:  // SETIOUDIS
-			m_ioudis = true; break;
 
-		case 0x7f:  // CLRIOUDIS
-			m_ioudis = false; break;
+			assert(m_ioudis && "CLRIOUDIS does not exist on IIe");
```
ajrhacker, line 2424: Was this assert really meant to be left in?

arekkusu42: Yes, to warn any future developer against re-adding this functionality there.

I wanted something stronger than a comment; in the debug build, it will verify that m_ioudis hasn't somehow been switched off on the IIe, in any game writing PTRIG.  (In the release build of course it vanishes.)

rb6502:
@arekkusu42 , are you on the Apple II Slack?  There's definitely people there that could test on a IIc.

arekkusu42:
I'm not on Slack, thanks for the tip!

PR #14309, opened 2025-10-11
A couple of arcade additions (systemy2 motherboard ROM, tangtang clone)

mamehaze:
- added motherboard ROM for System Board Y2 [Brizzo, Peter Wilhelmsen]

new WORKING clones
------------
Tang Tang (ver. 0526, 26/05/2000, set 2) [Hammy]


mamehaze:
tangtanga internal version number is the same, but it's clearly a different build with lots moved around

PR #14311, opened 2025-10-11
minitel2: Improved ROM list

fabio-d:
Different revisions of the Minitel 2 came with slightly different ROMs. In addition to the ones already listed (Bv4 and Bv9), this commits adds references to Bv6 and Bv7 too.

The descriptions of the ROMs were updated to mention "France Telecom" (the "ft_" in their names), who commissioned this Minitel model.

A placeholder "Custom ROM (for homebrew development)" was added too, with the maximum possible size (64 KiB). This makes it easy use precompiled MAME binaries to test/debug an homebrew ROM too, by simply copying it into the roms directory under the name "devel.bin" and then booting with "-bios devel".

cc @jfdelnero

### DIFF
--- a/src/mame/philips/minitel_2_rpic.cpp
+++ b/src/mame/philips/minitel_2_rpic.cpp
@@ -2,7 +2,7 @@
 // copyright-holders: Jean-Francois DEL NERO
 /***************************************************************************
 
-    Minitel 2
+    Minitel 2 by Philips
 
     The Minitel is a small, on-line computer/Videotex terminal with multi-services that
     can be connected to any French telephone line. This terminal was widely used in France
@@ -11,7 +11,9 @@
     There are several models and versions. Most of them are based on 8051 compatible MCUs
     and EF9345 semi graphic video chip.
 
-    The current implementation is a Minitel 2 from "La RADIOTECHNIQUE PORTENSEIGNE" / RPIC (Philips)
+    The current implementation is a Minitel 2 (NFZ 400) from "La RADIOTECHNIQUE
+    PORTENSEIGNE" (RPIC, later acquired by Philips).
+
     More Minitel hardware related informations are available on this page :
     http://hxc2001.free.fr/minitel
 
@@ -60,6 +62,8 @@
 
 #include "emu.h"
 
+#include "bus/generic/carts.h"
+#include "bus/generic/slot.h"
 #include "bus/rs232/rs232.h"
 #include "cpu/mcs51/mcs51.h"
 #include "machine/clock.h"
@@ -113,6 +117,7 @@ public:
 	minitel_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
+		, m_romslot(*this, "romslot")
 		, m_ts9347(*this, "ts9347")
 		, m_palette(*this, "palette")
 		, m_i2cmem(*this, "i2cmem")
@@ -129,6 +134,7 @@ protected:
 
 private:
 	required_device<i80c32_device> m_maincpu;
+	required_device<generic_slot_device> m_romslot;
 	required_device<ts9347_device> m_ts9347;
 	required_device<palette_device> m_palette;
 	required_device<i2cmem_device> m_i2cmem;
@@ -178,6 +184,9 @@ void minitel_state::machine_start()
 	m_palette->set_pen_color(5, 120, 120, 120);
 	m_palette->set_pen_color(6, 200, 200, 200);
 	m_palette->set_pen_color(7, 255, 255, 255);
+
+	if (m_romslot->exists())
+		m_maincpu->space(AS_PROGRAM).install_read_handler(0x0000, 0xffff, emu::rw_delegate(*m_romslot, FUNC(generic_slot_device::read_rom)));
 }
 
 void minitel_state::port1_w(uint8_t data)
@@ -529,6 +538,9 @@ void minitel_state::minitel2(machine_config &config)
 	m_maincpu->port_in_cb<3>().set(FUNC(minitel_state::port3_r));
 	m_maincpu->port_out_cb<3>().set(FUNC(minitel_state::port3_w));
 
+	/* if populated, this ROM slot replaces the main ROM */
+	GENERIC_SOCKET(config, m_romslot, generic_linear_slot, "minitel_rom", "bin,rom");
+
 	I2C_24C02(config, m_i2cmem);
 
 	TS9347(config, m_ts9347, 0);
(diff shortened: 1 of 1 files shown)


src/mame/philips/minitel_2_rpic.cpp
```diff
@@ -564,23 +566,31 @@ void minitel_state::minitel2(machine_config &config)
...
+
+	ROM_SYSTEM_BIOS(3, "ft_bv9", "France Telecom Bv9")
+	ROMX_LOAD( "bv9.1402",           0x0000, 0x8000, CRC(ace5d65e) SHA1(c8d589f8af6bd7d339964fdece937a76db972115), ROM_BIOS(3) )
+
+	ROM_SYSTEM_BIOS(4, "demov1", "The Minitel Demo (jfdelnero)")
+	ROMX_LOAD( "demo_minitel.bin",   0x0000, 0x8000, CRC(607f2482) SHA1(7965edbef68e45d09dc67a4684da56003eff6328), ROM_BIOS(4) )
 
-	ROM_SYSTEM_BIOS(2, "ft_bv9", "Minitel 2 ROM Bv9")
-	ROMX_LOAD( "bv9.1402",           0x0000, 0x8000, CRC(ace5d65e) SHA1(c8d589f8af6bd7d339964fdece937a76db972115), ROM_BIOS(2) )
+	ROM_SYSTEM_BIOS(5, "devel",  "Custom ROM (for homebrew development)")
```
rb6502: I don't think there's any precedent for something like this.  Normally this is done by simply launching MAME directly with a driver from the command line ("mame minitel"), in which case it will warn about checksum mismatches but still start up.

fabio-d: The issue with simply replacing an existing one is that all of them are only 32 KB size, but the HW actually supports up to 64 KB if you put a big enough ROM chip. And MAME of course does what `ROMX_LOAD` says and simply loads the first 32 KB.

I've found this problem with [this custom ROM of mine](https://github.com/fabio-d/minitel-native/tree/43b069f0026d693e4013b5e9dda4ed322ce30683/examples/image_gallery) that shows a gallery of images in a loop. Its ROM size is bigger than 32K and I've tested that it does indeed work in a real Minitel 2.

```shell
$ cd roms/minitel2

# The original size of my custom ROM is 46857 bytes.
$ stat --format=%n=%s image_gallery.bin
image_gallery.bin=46857

# Let's pad it to 64 KB (doing this or not does not change the
# outcome of the experiment), but we all like round numbers :)
$ truncate -s 65536 image_gallery.bin

# Let's put it in place of both "ft_bv9" and "devel"
$ cp image_gallery.bin bv9.1402
$ cp image_gallery.bin devel.bin
$ stat --format=%n=%s bv9.1402 devel.bin
bv9.1402=65536
devel.bin=65536
```

Let's run MAME with `-bios ft_bv9`:
```
bv9.1402 WRONG LENGTH (expected: 00008000 found: 00010000)
bv9.1402 WRONG CHECKSUMS:
    EXPECTED: CRC(ace5d65e) SHA1(c8d589f8af6bd7d339964fdece937a76db972115)
       FOUND: CRC(c456b557) SHA1(da53d5bb05c10cbb7bda77ff51014edbfdcb3d94)
```

and this is the screen capture:
bv9

And let's run with `-bios devel` instead:
```
devel.bin NO GOOD DUMP KNOWN
```

and this is the screen capture:
devel

Note how after a few images the `-bios ft_bv9` invocation starts displaying just white screens until it cycles back to the first image (because the ROM area containing their data has not been loaded), whereas the `-bios devel` invocation keeps displaying them properly.

I get that I'm "abusing" MAME as a development platform and this `devel` thing is an ugly workaround for this edge case. Maybe we can call it something else, such as `rom64k`?

Of course, it's also fine if we decide that it's out of scope for MAME and it should not be supported at all :)

pmackinlay: The ROM region should match the address decoding done by the hardware. If the hardware decodes 16 address lines to the ROM device(s), then the ROM region should be defined as 64KiB. If fitting a smaller device is allowed (e.g., 32KiB), it should either be padded/filled with empty data (0x00 or 0xff, typically), or mirrored within the total range, depending on what the hardware does.

Depending on how the region is used, the pad/fill or mirror can be done within the region, or dynamically by whatever installs them in the address map. This approach allows larger and smaller ROMs to be used interchangeably as needed.

In this case, it sounds like the driver should be updated to reflect the 64KiB region supported by the hardware.

fabio-d: The hardware does expose address lines A0-A15 on the ROM socket ([this pic](https://forum.system-cfg.com/viewtopic.php?p=249020#p249020), pinout on the right). The factory ROM [comes as a 87C257](http://hxc2001.free.fr/minitel/index.html#minitel2_processeur), which interprets A15 as chip enable.

In short, the stock ROM leaves a hole in the upper half of the address space with no chip answering. But you can easily fit a different ROM chip, as the A15 line is exposed in the socket.

That said, what are the changes you recommend? I'm not sure I can correlate your advice to code. If we assume that we take the current version of the PR and remove the `devel` entry, we will have:
* Address map with the full 64 K: `map(0x0000, 0xffff).rom();`
* ROM region with 64 K too: `ROM_REGION( 0x10000, "maincpu", ROMREGION_ERASEFF )`
* All the ROM entries with 32 K length, e.g.: `ROMX_LOAD("bv9.1402", 0x0000, 0x8000, ... )`

In this configuration replacing the `bv9.1402` file with a 64 K ROM results in it being loaded only up to 32 K (see experiment above). What changes do you recommend?

Pernod70:
Instead of using a fake `devel` region maybe add a `GENERIC_SOCKET(config, m_rom, generic_linear_slot, "minitel_rom", "bin,rom");` to the machine. 
Then in `machine_reset()` you can check whether the user has manually loaded a ROM with `m_rom->exists()` and install it, overwriting your ROM region, with `install_read_handler(0x0000, 0xffff, emu::rw_delegate(*m_rom, FUNC(generic_slot_device::read_rom)));`. 
Ensure to check the loaded ROM has a power of 2 size, ie. 8K, 16K, 32K, 64K, and it will automatically mirror it for you as it's a `generic_linear_slot`.

rb6502:
Thanks Pernod, that's what I was going to suggest.

src/mame/philips/minitel_2_rpic.cpp
angelosa: There is a precedent anyway:

https://github.com/mamedev/mame/blob/d21dce3077b48696a1f2fb468a98e9a74e135cb7/src/mame/nec/pc8001.cpp#L828-L829

And it's imo a fair use of `ROM_LOAD_OPTIONAL` (the socket in this case is empty, the PC-8001 BIOS checks the header at this range to load the extension if present).

fabio-d:
**DO NOT MERGE THE CURRENT REVISION**

# GENERIC_SOCKET apprach

> Instead of using a fake `devel` region maybe add a `GENERIC_SOCKET(config, m_rom, generic_linear_slot, "minitel_rom", "bin,rom");` to the machine. Then in `machine_reset()` you can check whether the user has manually loaded a ROM with `m_rom->exists()` and install it, overwriting your ROM region, with `install_read_handler(0x0000, 0xffff, emu::rw_delegate(*m_rom, FUNC(generic_slot_device::read_rom)));`. Ensure to check the loaded ROM has a power of 2 size, ie. 8K, 16K, 32K, 64K, and it will automatically mirror it for you as it's a `generic_linear_slot`.

@Pernod70 I've just tried this and it worked, except that I also had to change `option_add_internal` to `option_add` in `src/devices/bus/generic/carts.cpp` to make it usable from the command-line.
```c++
device_slot_interface &generic_linear_slot(device_slot_interface &device)
{
	device.option_add("rom", GENERIC_ROM_LINEAR); // <-------- INSTEAD OF option_add_internal
	return device;
}
```

Without this extra change, I would get
```shell
$ ./mame -window minitel2 -romslot rom -romimage MY_LARGE_IMAGE.bin
Ignoring MAME exception: Unknown slot option 'rom' in slot 'romslot'
Unknown slot option 'rom' in slot 'romslot'
```

Did I miss something about your proposal? I've uploaded the code in the current revision, if you want to have a look

# ROM_LOAD_OPTIONAL approach

As an alternative, I've also tested the approach suggested by @angelosa of adding a `ROM_LOAD_OPTIONAL` for the upper half, and it works for me too:
```
// For loading custom ROMs larger than 32 KiB.
ROM_LOAD_OPTIONAL( "top32k.bin", 0x8000, 0x8000, NO_DUMP )
```

# Conclusion

All things considered, I think the `ROM_LOAD_OPTIONAL` approach looks simpler to me, as it's just a two-line change that doesn't involve overriding the ROM region in code. Having to split the large ROM files I'm testing in two halves is not an issue for me :slightly_smiling_face:

angelosa:
> DO NOT MERGE THE CURRENT REVISION

You can mark a PR as draft, if you don't want being merged.

Pernod70:
>
> @Pernod70 I've just tried this and it worked, except that I also had to change `option_add_internal` to `option_add` in `src/devices/bus/generic/carts.cpp` to make it usable from the command-line.
> 
> ```c++
> device_slot_interface &generic_linear_slot(device_slot_interface &device)
> {
> 	device.option_add("rom", GENERIC_ROM_LINEAR); // <-------- INSTEAD OF option_add_internal
> 	return device;
> }
> ```
> 
> Without this extra change, I would get
> 
> ```shell
> $ ./mame -window minitel2 -romslot rom -romimage MY_LARGE_IMAGE.bin
> Ignoring MAME exception: Unknown slot option 'rom' in slot 'romslot'
> Unknown slot option 'rom' in slot 'romslot'
> ```
> 
> Did I miss something about your proposal? I've uploaded the code in the current revision, if you want to have a look
> 

You don't need to change carts.cpp. Your command line should be:
`$ ./mame -window minitel2 -rom MY_LARGE_IMAGE.bin`

fabio-d:
> You don't need to change carts.cpp. Your command line should be: `$ ./mame -window minitel2 -rom MY_LARGE_IMAGE.bin`

Nice, it worked perfectly! Sorry for missing it out.

I've uploaded the PR :smiley:

Pernod70 (approved):

fabio-d:
Is there anything left for me to do on this PR? :)

src/mame/philips/minitel_2_rpic.cpp
```diff
@@ -564,23 +576,28 @@ void minitel_state::minitel2(machine_config &config)
...
+	ROM_SYSTEM_BIOS(2, "ft_bv7", "France Telecom Bv7")
+	ROMX_LOAD( "bv7.bin",            0x0000, 0x8000, CRC(2dde1628) SHA1(85500c06b93efb6e925dcb156248425fa3366ce6), ROM_BIOS(2) )
+
+	ROM_SYSTEM_BIOS(3, "ft_bv9", "France Telecom Bv9")
+	ROMX_LOAD( "bv9.1402",           0x0000, 0x8000, CRC(ace5d65e) SHA1(c8d589f8af6bd7d339964fdece937a76db972115), ROM_BIOS(3) )
 
-	ROM_SYSTEM_BIOS(2, "ft_bv9", "Minitel 2 ROM Bv9")
-	ROMX_LOAD( "bv9.1402",           0x0000, 0x8000, CRC(ace5d65e) SHA1(c8d589f8af6bd7d339964fdece937a76db972115), ROM_BIOS(2) )
+	ROM_SYSTEM_BIOS(4, "demov1", "The Minitel Demo (jfdelnero)")
+	ROMX_LOAD( "demo_minitel.bin",   0x0000, 0x8000, CRC(607f2482) SHA1(7965edbef68e45d09dc67a4684da56003eff6328), ROM_BIOS(4) )
```
cuavas, line 595: Stuff like this should be in a software list (if it should be in MAME itself at all).

fabio-d: What do you mean exactly?

The Minitel was not meant to load custom program originally. The ROM socket is not easily accessible to the user and @jfdelnero's [demo (2017)](https://github.com/jfdelnero/minitel/tree/master/minitel2/the_minitel_demo) is, to my knowledge, the first non-stock ROM that has ever been written.

An unmodified Minitel will only run its stock ROM, which makes it work as a thin client of the Videotex protocol (with the actual programs running on the server, at the other end of the phone line).

PR #14319, opened 2025-10-11
sega/model2.cpp: improve interrupt logic

gm-matthew:
- Interrupt control register only returns interrupts actively signalled to the i960; masked interrupts are not returned
- Sound IRQs are triggered when either TxRDY or RxRDY from the USART is active (Model 2B manual, section 6.2)
- Remove model2c_interrupt callback; no evidence that Model 2C behaves any differently to earlier board revisions
- Remove vcop issue from TODO list since it no longer occurs (tested)

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -11,15 +11,12 @@
 
     TODO:
     - Mip Mapping still needs to be properly sorted in the renderer;
-    - sound comms still needs some work (sometimes m68k doesn't get some commands or play them with a delay);
     - outputs and artwork (for gearbox indicators);
     - clean-ups;
 
     TODO (per-game issues)
     - doa, doaa: corrupted sound, eventually becomes silent;
     - dynamcopc: corrupts palette for 2d;
-    - fvipers: enables timers, but then irq register is empty, hence it crashes with an
-      "interrupt halt" at POST (regression, worked around);
     - hpyagu98: stops with 'Error #1' message during boot.
       Also writes to the 0x600000-0x62ffff range in main CPU program map;
     - lastbrnx: uses external DMA port 0 for uploading SHARC program, hook-up might not be 100% right;
@@ -29,7 +26,6 @@
               bypass it by entering then exiting service mode;
     - sgt24h: has input analog issues, steering doesn't center when neutral,
       gas and brake pedals pulses instead of being fixed;
-    - vcop: sound dies at enter initial screen (i.e. after played the game once) (untested);
 
     Notes:
     - some analog games can be calibrated in service mode via volume control item ...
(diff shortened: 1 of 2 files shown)


src/mame/sega/model2.cpp
```diff
@@ -2578,8 +2514,8 @@ void model2_state::model2_scsp(machine_config &config)
 	m_scsp->add_route(1, "speaker", 1.0, 1);
 
 	I8251(config, m_uart, 8000000); // uPD71051C, clock unknown
-//  m_uart->rxrdy_handler().set(FUNC(model2_state::sound_ready_w));
-//  m_uart->txrdy_handler().set(FUNC(model2_state::sound_ready_w));
+	m_uart->rxrdy_handler().set(FUNC(model2_state::sound_ready_w));
+	m_uart->txrdy_handler().set(FUNC(model2_state::sound_ready_w));
```
cuavas, line 2518: This needs to use an input merger or some other mechanism to combine the two signals – if you do it without one, the two signals will nuke each other.

Consider:

* rxrdy goes active – `m_sound_irq_pending` set to 1
* txrdy goes active – `m_sound_irq_pending` set to 1 again (no actual change)
* txrdy goes inactive – `m_sound_irq_pending` set to 0, but rxrdy is still active

In general, it’s never safe to connect two output lines to one input like this without combining them somehow.

src/mame/sega/model2.h
```diff
@@ -103,6 +103,8 @@ class model2_state : public driver_device
 	void init_srallyc();
 	void init_powsledm();
 
+	void sound_ready_w(int state);
```
cuavas, line 106: Internal handler functions shouldn’t be public.

src/mame/sega/model2.cpp
gm-matthew: A valid point; I shall make it so that a separate callback function is used for each, perhaps something like `sound_rxrdy_callback()` and `sound_txrdy_callback()`.

cuavas: You can use the stuff from src/devices/machine/input_merger.h for the glue logic at the cost of some extra trips through the scheduler when the signals change (pretty easy to find examples using them).

src/mame/sega/model2.h
gm-matthew: I mistakenly thought that it had to be public because I was getting a compile error making it protected, but I've realized that I can change the callback from `model2_state::sound_ready_w` to `model2o_state::sound_ready_w` in `model2o()` and it compiles just fine.

cuavas: Yeah, it’s a C++ism that you need to refer to inherited protected members via your own class name – the main effect is that you can’t take the address of an eclipsed inherited protected member function.

src/mame/sega/model2.cpp
gm-matthew: Alternatively I could make it so that both the callback function and `irq_enable_w()` call `m_uart->status_r()`, eliminating the need to store the state of the RxRDY and TxRDY lines as member variables in `model2_state`.

gm-matthew:
I have made the requested amendments in a new pull request: https://github.com/mamedev/mame/pull/14321

PR #14328, opened 2025-10-13
sinclair/specnext.cpp: Expansion bus control

holub:


src/mame/sinclair/specnext.cpp
```diff
@@ -2929,9 +2931,9 @@ void specnext_state::map_io(address_map &map)
 			m_dac[3]->data_w(data);
 	}));
 
-	map(0x0000, 0xffff).view(m_io_shadow_view);
-	subdevice<zxbus_device>("zxbus")->set_io_space(m_io_shadow_view[0], m_io_shadow_view[1]);
-	m_io_shadow_view.select(0);
+	map(0x0000, 0xffff).view(m_io_expbus_view);
+	m_io_expbus_view[0]; // exp bus disabled
+	subdevice<zxbus_device>("zxbus")->set_io_space(m_io_expbus_view[1], m_io_expbus_view[1]);
```
galibert, line 3047: Vas, how iffy is that?  I'm not at all sure of the state of initialization of subdevices at that point in the startup.

PR #14331, opened 2025-10-14
esq5505.cpp: Add VFX-family ROM & EEPROM Cartridge support, and improve floppy support.

cbrunschen:
This adds vfxcart.h and .cpp, which implement a cartridge slot device and a matching cartridge image device.

The VFX also uses DUART input 1 for the cartridge presence, just like the VFX-SD.

This PR also fixes recognition of the floppy drive, so a floppy can now be inserted and read from.

This includes loading the sequencer OS, sound and sequence files, and successfully playing back sequences and songs.

It also tracks the presence of the disk(image) in the floppy drive to synthesize the DSKCHG signal that the (VFS)SD(-1) expects, and combines that with the OTIS IRQ to generace the correct DOCIRQ. Similarly it handles the DSEL and MON signals from the keyboard to the floppy image.

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -177,10 +177,14 @@
 #include "sound/esqpump.h"
 #include "emupal.h"
 #include "speaker.h"
+#include "vfxcart.h"
 
 #include <cstdarg>
 #include <cstdio>
 
+// #define VERBOSE 1
+#define VERBOSE 0
+#include "logmacro.h"
 
 namespace {
 
@@ -211,10 +215,12 @@ public:
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
 		, m_duart(*this, "duart")
+		, m_otis(*this, "otis")
 		, m_esp(*this, "esp")
 		, m_pump(*this, "pump")
 		, m_fdc(*this, "wd1772")
 		, m_floppy_connector(*this, "wd1772:0")
+		, m_cartslot(*this, "cartslot")
 		, m_panel(*this, "panel")
 		, m_dmac(*this, "mc68450")
 		, m_mdout(*this, "mdout")
@@ -242,10 +248,12 @@ protected:
 private:
 	required_device<m68000_device> m_maincpu;
 	required_device<mc68681_device> m_duart;
+	required_device<es5505_device> m_otis;
 	required_device<es5510_device> m_esp;
 	required_device<esq_5505_5510_pump_device> m_pump;
 	optional_device<wd1772_device> m_fdc;
 	optional_device<floppy_connector> m_floppy_connector;
+	optional_device<ensoniq_vfx_cartslot> m_cartslot;
 	required_device<esqpanel_device> m_panel;
 	optional_device<hd63450_device> m_dmac;
 	required_device<midi_port_device> m_mdout;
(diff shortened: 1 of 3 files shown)


galibert:
I'm surprised at the complxity of the floppy interface.  What type of floppy drive is it, a custom one or a normal pc-like one?

Are you sure the dskchg is timer-based and not the usual "clear on first track change with a floppy in" that is implemented in floppy_image_device?

cbrunschen:
> I'm surprised at the complxity of the floppy interface.

So am I; it required a fair amount of digging and thinking, and I am so glad that I have the schematics, without which this would have been pretty impossible.

> What type of floppy drive is it, a custom one or a normal pc-like one?

It's a Sony MP-F11W-L3 drive, with a Shugart interface, and connecting to a WD1772 FDC. But it does seem to have at the very least that DSKCHG-related difference from most drives; see below.

The WD1772 doesn't support all of the signals on the floppy drive, so the rest are driven directly from, and received directly by, the system itself in various ways.

The DSEL (disk select) and MO (motor on) signals are driven from DUART OP4 (output 4).

The READY signal from the drive goes directly to DUART IP0. I've tried using the floppy_image_device's ready signal, but have had problems getting that to work; just using the presence/absence of an image file seems to work, though of course if that can be refined I'd be delighted. But for now, what I have, works reliably in my testing.

> Are you sure the dskchg is timer-based and not the usual "clear on first track change with a floppy in" that is implemented in floppy_image_device?

I an absolutely sure. 

I first found it documented for the EPS in Keir Fraser's FlashFloppy firmware, when people tried to connect it to an Ensoniq EPS and it [would not recognize disk changes](https://github.com/keirf/FlashFloppy/issues/240). 

But with this being so strange, I measured it with my cheap Saleae Logic clone logic analyzer on an SD-1/32. After a disk change - so, removed the previous disk, inserted a new disk, then went to try to read something from the disk: pressed the "Storage" button, then selected "Disk" by pressing the middle button below the screen, and then "Load" again with the middle button below the screen) - the keyboard signals (drives low) the DSEL and MO signals (enabling the drive and starting the motor), and the drive responds with a brief pulse: in my measurement with a cheap logic analyzer sampling at 4 MHz, all of two (2) samples were low, so a pulse of 500 nanoseconds and certainly no more than 1000 nanoseconds to 1 microsecond.

I'm including the sigrok trace recorded with [PulseView](https://sigrok.org/wiki/PulseView):.

[Load from floppy after new floppy inserted.zip](https://github.com/user-attachments/files/22909431/Load.from.floppy.after.new.floppy.inserted.zip)

Look at the 335088 mycrosecond timestamp: the MO and DSEL signals both go low, and at the same time, DSKCHG goes low, but only for 2 samples, and with no other signal changing during that very short window.

The DSKCHG signal is actually routed, via a gate and a transistor, into the same IRQ line on the CPU itself that is also used by the ES5505 OTIS sound generator chip. It so happens that on the SD-1, if that interrupt is taken (and I presume after the keyboard has checked the ES5505 interrupt status), eventually the SD-1 will assert (drive low) the DSEL (and MO) lines to the floppy drive and issue a STEP command, which will inturn reset DSKCHG even on a standard drive; but that of course might take longer than it does when the drive itself only briefly pulses the signal.

And while this would technically work on the SD-1, it would not work on the EPS, as documented by FlashFloppy users - where they added a configuration setting precisely to reset DSKCHG after a short time. So including this code is useful, possibly even necessary, to get the EPS and EPS16+ to work in the future.

Another trickiness is that while a disk is in the drive (and has been read from), the SD-1 pulses DUART OP4 about 32 times per second, for only a few microseconds each time - this of course driving the combined DSEL and MO pins to the drive. This is almost certainly to check whether the disk is still in the drive, since asserting DSEL should make the drive respond with asserting DSKCHG if the disk has been removed.

But while this pulse is short enough that the real drive doesn't seem to notice much, and in particular even though MO is signalled, the drive doesn't seem to have enough time to start spinning the motor, the emulated floppy_image_device can react immediately by playing the floppy motor sound. That's why I've added another timer that will, then DUART OP4 is signalled, delay actually driving MO for a bit - so that if it is just a short pulse, the floppy motor sound doesn't start to play.

I actually thought that this pulsing of DUART OP4 while the disk remained inserted, was because I had missed something, some signal from the drive that would tell the main CPU to stop doing that. Much to my surprise I found that exact behaviour in the real keyboard - as you can also see towards the end of the sigrok trace.

cbrunschen:
Also FYI: [the SD-1 mainboard schematics](https://drive.google.com/file/d/1J0BJdgOze7reBce29mVo4mTAuKU5kKVV) (linked from [http://zine.r-massive.com/ensoniq-technical-documents-and-schematics/](http://zine.r-massive.com/ensoniq-technical-documents-and-schematics/)).

cbrunschen:
Thank you!

PR #14333, opened 2025-10-14
esq5505.cpp and esqpump.cpp: Route es5505 bus 0 to a separate 'aux' stereo output.

cbrunschen:
The VFX-SD and later keyboards in the family have not only a main stereo output, but also a second 'aux' stereo output that bypasses effects processing and has separate output jacks from the keyboard. 

Screenshot From 2025-10-14 14-57-01

This allows some sound to be routed for separate effects processing or recording.

Screenshot From 2025-10-14 14-55-54

### DIFF
--- a/src/devices/sound/esqpump.cpp
+++ b/src/devices/sound/esqpump.cpp
@@ -4,6 +4,9 @@
 
   esqpump.cpp - Ensoniq 5505/5506 to 5510 interface.
 
+	Modeled specifically after the routing of and for use with
+	the VFX family of keyboards.
+
   By Christian Brunschen
 
 ***************************************************************************/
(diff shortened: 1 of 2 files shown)


cbrunschen:
Thank you!

src/devices/sound/esqpump.cpp
```diff
@@ -26,7 +29,12 @@ esq_5505_5510_pump_device::esq_5505_5510_pump_device(const machine_config &mconf
 
 void esq_5505_5510_pump_device::device_start()
 {
-	m_stream = stream_alloc(8, 2, clock(), STREAM_SYNCHRONOUS);
+	// The VFX only has a single pait of stereo outputs, 'Main'; these will be channels 0 and 1,
```
cuavas, line 32: “single pait”?

cbrunschen: "pait" is a typo for "pair".

PR #14335, opened 2025-10-14
further XaviX sound improvements and code cleanups [ramacat]

mamehaze:
Further XaviX sound improvements and code cleanups [ramacat]
- Last remaining stuck notes bugs fixed.
 - IRQ handling for audio engine rewritten to remove 10.3 magic number and to derive from the sequencer rate only:
   - Balls now line up perfectly on popira towards the end of the song.
   - No more re-syncs between CPU and audio engine in hikara, which caused a lumbering feel to the music on each resync.
 - Sequencer rate calculated from hardware behaviours.
 - WM1 (Noise Generator) corrected to better reflect hardware.
 - Various envelope logic bugs fixed.
 - Implement capacity limit for mixer with hardware accurate voice clipping.
 - Implement DAC broadcast and multiplex mixer modes.
 - Simplified pitch-bend stepping logic.
 - WM0 waves are now read from RAM, as they are CPU modifiable.
 - Gain calculations adjusted to avoid clipping on worst case titles.
 - Lots of clean-up, de-duplication, better naming, and correctness fixes.
 - Detailed logging put in place.
 - Annotations added.
 - Use BIT helpers.

### DIFF
--- a/src/mame/tvgames/xavix.h
+++ b/src/mame/tvgames/xavix.h
@@ -66,7 +66,6 @@ public:
 	double tempo_frequency(uint8_t tempo) const;
 	double tempo_tick_hz(uint8_t tempo) const;
 	uint32_t  phase_step_per_tick(uint32_t rate) const;
-	uint8_t   current_page() const;
 	uint32_t envelope_period_ticks(uint8_t tp) const;
 
 
@@ -81,8 +80,6 @@ protected:
 private:
 	// stream
 	sound_stream* m_stream = nullptr;
-	//uint32_t m_output_rate = 167791;
-	xavix_state* m_owner_state = nullptr;
 
 	// global timing
 	uint8_t m_tempo_div[4] = { 0, 0, 0, 0 };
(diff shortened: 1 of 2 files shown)


mamehaze:
(this is a significant improvement over the previous PR after some more extensive testing and finding problematic cases etc.)

mamehaze:
@ajrhacker you merged the previous XaviX sound PR, are you happy with this? (and the Jumping Popira cart fix PR)

any further functional tweaks will come after (maybe lowering overall mix etc.)

ajrhacker:
The one issue I see is the rather hacky method of hooking in `sound_wave_ram_read`.

mamehaze:
ah, yeah, I think that was some experimental code that crept in, I'll check with Ramacat

mamehaze:
ok, yeah, that's gone, it was some experimental code based on study of udance, which is the only one to enable that mode, but might just be doing so due to a bug (the screens we can see aren't meant to have audio, but it might be decoding audio in software into a RAM buffer once we get further into it)

PR #14337, opened 2025-10-15
jpmimpct: add layouts to Scrabble, Hangman, Coronation Street, Trivial Pursuit, and Treasure Quest - update Cluedo layout

as-tb-dev:
- This adds layouts for Scrabble, Hangman (JPM), Coronation Street Quiz Game, Trivial Pursuit and Treasure Quest
- The jpmimpct_state::digits_w() function has been updated to add bit mapping to make Scrabble work with the 16-segment LED.  No other working game with this driver appears to invoke this function to drive LEDs.
- Made minor updates to the Cluedo layout to correct minor lamp assignment issues, brightness issues, and to remove lamps with unknown functions

### DIFF
--- a/src/mame/jpm/jpmimpct.cpp
+++ b/src/mame/jpm/jpmimpct.cpp
@@ -106,6 +106,7 @@ Thanks to Tony Friery and JPeMU for I/O routines and documentation.
 
 #include "jpmimpct.lh"
 #include "cluedo.lh"
+#include "trivialp.lh"
 #include "scrabble.lh"
 #include "hngmnjpm.lh"
 #include "coronatn.lh"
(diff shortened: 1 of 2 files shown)


james-wallace-ghub:
In so much as I have any sway over these drivers, I am totally fine with these as they function, as well as the helper for 14 segs (this is something that would be beneficial for .lay to support natively, but that's another case for another time).

as-tb-dev:
Thanks.  Another good feature for .LAY files could be to have lamps drive the ledXseg components.  The 7-segment LEDs in Scrabble are driven by lamps.  The manual creation of the 7-segment LED in the .LAY file (to able to use lamps) does not look as good as what using the led7seg component (using 'digits') would look like.

as-tb-dev:
I did not realise I had uploaded trivialp.lay with my original PR submission - that layout was neither ready at the time nor linked into the updated jpmimpct.cpp.  It is ready now, so I just updated it and linked it in.

PR #14338, opened 2025-10-15
seibu/seibucats.cpp, seibu/seibuspi.cpp: Fix regression since commit 3bc4210656939f870e68f95e1990345571383150

cam900:
seibuspi_base_state::video_start() contains needs-to-initialized variables, and should be called after initialization routine in each child video_start() functions. seibu/seibuspi.cpp: Add notes


cuavas:
Is it possible to provide the configuration some other way?  For example, make palette and sprite RAM into `memory_share_creator` and supply sizes to the base class constructors, or have some way of supplying the setup information from the machine configuration functions.

Having stuff in the base class that depends on initialisation from the derived classes seems like an error-prone approach.

PR #14341, opened 2025-10-15
esq5505.cpp: Make the VFX family of keyboards' RAM non-volatile.

cbrunschen:
The VFX family of keyboards have a 3V lithium battery that keeps their RAM powered, which makes it effectively non-volatile.

Each has 64 kBytes of OS RAM (2 * 62256 32k x 8 SRAM). The VFX-SD and SD-1 have a base amount of sequencer memory, 64 kB in 2 * 62256 SRAM chips, mapped 0x330000 .. 0x33ffff.

Both of those can also accommodate the "SQX-70" sequencer memory expansion, which is simply 2 * 621001, 128k x 8 SRAM chips for another 256 kB of sequencer RAM, mapped 0x340000 .. 0x37ffff.

The SQX-70 is included by default in the SD-1/32 voice; and it's also included by default in the emulated vfxsd and sd1 devices as well as the sd132.

This PR:
- corrects the end address of Sequencer RAM that is mapped in the VFXSD memory map
- adds the device_nvram_interface to the esq5505_state class, but disabled by default
- enabled nvram for the VFX family of devices, in which case it loads and stores:
    - all of OS RAM, 64 kBytes
    - if present, all of Sequencer RAM, 64 + 256 = 320 kByte
   
  for a total of 64 kBytes for the VFX / 384 kBytes for the VFX-SD, SD-1 and SD-1/32

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -172,6 +172,7 @@
 #include "esqvfd.h"
 #include "machine/hd63450.h"    // compatible with MC68450, which is what these really have
 #include "machine/mc68681.h"
+#include "machine/nvram.h"
 #include "machine/wd_fdc.h"
 #include "sound/es5506.h"
 #include "sound/esqpump.h"
@@ -209,12 +210,11 @@ static void ATTR_PRINTF(1,2) print_to_stderr(const char *format, ...)
 }
 #endif
 
-class esq5505_state : public driver_device, public device_nvram_interface
+class esq5505_state : public driver_device
 {
 public:
 	esq5505_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
-		, device_nvram_interface(mconfig, *this, false)
 		, m_maincpu(*this, "maincpu")
 		, m_duart(*this, "duart")
 		, m_otis(*this, "otis")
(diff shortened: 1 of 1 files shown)


src/mame/ensoniq/esq5505.cpp
```diff
@@ -452,6 +463,49 @@ void esq5505_state::otis_irq(int irq)
 	}
 }
 
+void esq5505_state::nvram_default()
+{
+	std::fill(m_ram.begin(), m_ram.end(), 0);
+	if (m_seqram)
+		std::fill(m_ram.begin(), m_ram.end(), 0);
```
galibert: seqram there?

But I don't think that complexity is needed.  Just add a pair of NVRAM devices with names "seqram" and "osram" and they'll handle saving, loading and default for you.  See yamaha/ymmu50.cpp for a trivial example.

cbrunschen: > seqram there?

Absolutely, thank you for spotting that! 

> But I don't think that complexity is needed. Just add a pair of NVRAM devices with names "seqram" and "osram" and they'll handle saving, loading and default for you. See yamaha/ymmu50.cpp for a trivial example.

Even better! I wasn't aware of that, thank you for pointing that out, I learn something new every day.

It does mean a had to do a bit of restructuring: pretty much every device was using the vfx(config) setup in some fashion, and the ks32 was using the same setup as the sd132 but does not have a "seqram" region; so I factored out the common 21- and 32-voice setups and only add the NVRAM devices where I know that nvram is actually in use.

cbrunschen:
Thank you!

PR #14342, opened 2025-10-16
sound/va_vca.cpp: Defined a CEM3360_VCA specialization and used in on the linndrum.

m1macrophage:
Also, improved input handling: defined enums and added validation.

This should be a functional no-op. Just turned `CEM3360_VCA` into a device, instead of using VA_VCA with `.configure_cem3360()`.

### DIFF
--- a/src/devices/sound/va_vca.cpp
+++ b/src/devices/sound/va_vca.cpp
@@ -6,64 +6,75 @@
 
 #include <cfloat>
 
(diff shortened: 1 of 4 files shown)


m1macrophage:
Oops, I used`CEM3320_VCA` instead of `CEM3360_VCA` in the commit message. If it is easy to correct when merging, please considering doing so.

PR #14347, opened 2025-10-16
t11: handle unaligned accesses to I/O page.

shattered:
Certain devices on bk, uknc require this, as they do not handle the WTBT bus signal and always decode full address.

### DIFF
--- a/src/devices/cpu/t11/t11.cpp
+++ b/src/devices/cpu/t11/t11.cpp
@@ -138,13 +138,39 @@ void t11_device::WBYTE(int addr, int data)
 
 int t11_device::RWORD(int addr)
 {
-	return m_program.read_word(addr & 0xfffe);
+	if (addr < 0160000)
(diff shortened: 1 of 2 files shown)


src/devices/cpu/t11/t11.cpp
```diff
@@ -138,13 +138,39 @@ void t11_device::WBYTE(int addr, int data)
...
+		return m_program.read_word(addr & 0xfffe);
+	else // accessing I/O page
+	{
+		auto flags = m_program.lookup_read_word_flags(addr);
+		if (!flags)
+			return m_program.read_word(addr & 0xfffe);
+		else if (flags & UNALIGNED_WORD)
+			return m_program.read_word_unaligned(addr);
+		else if (flags & UNALIGNED_BYTE)
+			return m_program.read_byte(addr);
```
cuavas, line 151: Does this path actually work?  It looks like it causes it to always do a byte access, even if the address is word-aligned.

PR #14348, opened 2025-10-16
esqpump.cpp: clamp input samples from es5505.

cbrunschen:
The es5505 does not currently clamp the samples it generates, even though the real hardware does. This can cause audio glitches, especially if those samples are fed into the ESP for processing.

https://github.com/mamedev/mame/pull/14329 would fix that at the source, and at the same time bring es5505 (and es5506) generated samples into the ±1.0 range. But until that or something similar is in place, we need to clamp the samples ourselves.

Here's a sample from the SD-1/32's OS 4.10 disk, from the demo song `$SD-PALETTE`, sequence `-D-`, that demonstrates the glitches:

Without this PR, so without clamping: [without_clamping.wav](https://github.com/user-attachments/files/22959780/without_clamping.wav)

With this PR, so with clamping: [with_clamping.wav](https://github.com/user-attachments/files/22959785/with_clamping.wav)

### DIFF
--- a/src/devices/sound/esqpump.cpp
+++ b/src/devices/sound/esqpump.cpp
@@ -14,6 +14,8 @@
 #include "emu.h"
 #include "esqpump.h"
 
+#include <algorithm>
+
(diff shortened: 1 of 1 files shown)


cbrunschen:
Made redundant by https://github.com/mamedev/mame/pull/14329, closing.

PR #14351, opened 2025-10-17
toaplan/slapfght.cpp: Split driver classes to different hardware configuration, Minor cleanups

cam900:
- Fix save state support
- Move CPU bank initialize routine into machine_start()
- Remove unnecessary variable
- Reduce duplicates
- Fix filename in comments

### DIFF
--- a/src/mame/toaplan/slapfght_m.cpp
+++ b/src/mame/toaplan/slapfght_m.cpp
@@ -41,15 +41,16 @@ uint8_t slapfght_state::getstar_mcusim_status_r()
 {
 	static const uint8_t states[3]={ 0xc7, 0x55, 0x00 };
 
+	const uint8_t ret = states[m_getstar_status_state];
 	if (!machine().side_effects_disabled())
 	{
-		m_getstar_status = states[m_getstar_status_state];
+		m_getstar_status = ret;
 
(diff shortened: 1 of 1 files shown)


src/mame/toaplan/slapfght_m.cpp
```diff
@@ -39,21 +39,24 @@ void slapfght_state::scroll_from_mcu_w(offs_t offset, uint8_t data)
...
+	{
+		m_getstar_status = states[m_getstar_status_state];
 
-	m_getstar_status_state++;
-	if (m_getstar_status_state > 2) m_getstar_status_state = 0;
+		m_getstar_status_state++;
+		if (m_getstar_status_state > 2) m_getstar_status_state = 0;
+	}
 
 	return m_getstar_status;
```
cuavas: This won’t show the actual value that would be returned in the debugger with this change.  You need to do something like:
```c++
	static const uint8_t states[3] = { 0xc7, 0x55, 0x00 };

	const uint8_t result = states[m_getstar_status_state];
	{
		m_getstar_status = result;

		m_getstar_status_state++;
		if (m_getstar_status_state > 2) m_getstar_status_state = 0;
	}

	return result;
```

PR #14355, opened 2025-10-18
Layout for Access Virus B

felipesanches:
The knobs are still not interactive. I'll prepare a followup PR with lua scripting once we have further progress in the CPU core so that we can hook up bank switching to port 5 high nibble, and also the actual hooking of the knobs. It will be a bit easier to troubleshoot the interactive knobs after we have some feedback on the LCD. 

Screenshot From 2025-10-17 22-26-39

### DIFF
--- a/src/mame/access/acvirus.cpp
+++ b/src/mame/access/acvirus.cpp
@@ -1,5 +1,5 @@
 // license:BSD-3-Clause
-// copyright-holders:R. Belmont
+// copyright-holders:R. Belmont, Felipe Sanches
 /***************************************************************************
 
     acvirus.cpp - Access Virus series
@@ -65,7 +65,11 @@
 #include "emu.h"
 #include "cpu/mcs51/sab80c535.h"
 #include "machine/intelfsh.h"
+#include "video/hd44780.h"
+#include "emupal.h"
 #include "speaker.h"
+#include "screen.h"
+#include "virusb.lh"
 
 
 namespace {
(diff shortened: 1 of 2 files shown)


src/mame/access/acvirus.cpp
```diff
@@ -154,7 +214,7 @@ ROM_END
 
 
 CONS( 1997, virusa,     0, 0, virus, virus, acvirus_state, empty_init, "Access", "Virus A", MACHINE_NOT_WORKING|MACHINE_NO_SOUND )
-CONS( 1999, virusb,     0, 0, virus, virus, acvirus_state, empty_init, "Access", "Virus B (Ver. T)", MACHINE_NOT_WORKING|MACHINE_NO_SOUND )
+CONS( 1999, virusb,     0, 0, virusb, virusb, acvirus_state, empty_init, "Access", "Virus B (Ver. T)", MACHINE_NOT_WORKING|MACHINE_NO_SOUND )
```
angelosa, line 223: I know it's not your fault, but why this is a `CONS`? Shouldn't it be a `SYST`?

felipesanches: I don't know.

felipesanches:
I've just git-rebased and fixed conflicts after @galibert's recent update that added support for hooking up port 5.

The addresses (or ports) for the scanning of buttons and for hooking up the display are still unknown.

felipesanches:
One of the CI jobs in this PR took 6 hours and then was cancelled (presumably due to a timeout). It seems to me that's a glitch on the CI environment, and not really a problem with this PR.

https://github.com/mamedev/mame/actions/runs/18634437084/job/53123624235?pr=14355

For additional thoughts on slow CI jobs, I've opened issue #14377, but I think we're good to merge here.

Screenshot From 2025-10-20 00-24-14

felipesanches:
F... Yeah! :-D

PR #14356, opened 2025-10-18
Various driver baeed in ST0016: Move common ROM bankswitching routine into seta/st0016.cpp, Minor cleanups

cam900:
- Reduce literal tag usage and runtime tag lookups
- Remove unnecessary comments for internal handlers
- Fix spacings
- Fix initializers

seta/st0016.cpp: Use memory space for external ROM, Reduce duplicates, Add notes, Use BIT helper for single bit values

seta/jclub2.cpp: Fix filename in comments

seta/macs.cpp: Fix region size, Use set_addrmap for address map

seta/simple_st0016.cpp: Split mayjinsn games into its own driver class

seta/srmp5.cpp: Reduce preprocessor defines, Use lowercase hexadecimal digits, Move unique_ptr initializer into machine_start

seta/ssv.cpp: Fix game name and filename in comment

### DIFF
--- a/src/mame/seta/macs.cpp
+++ b/src/mame/seta/macs.cpp
@@ -74,7 +74,6 @@ class macs_state : public driver_device
 public:
 	macs_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
-		, m_cart_bank(0)
 		, m_ram2(*this, "ram2")
 		, m_maincpu(*this,"maincpu")
 		, m_cart1(*this, "slot_a")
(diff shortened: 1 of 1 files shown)


src/mame/seta/simple_st0016.cpp
```diff
@@ -180,55 +187,60 @@ void st0016_state::st0016_io(address_map &map)
...
 }
 
-void st0016_state::latch32_w(offs_t offset, uint32_t data, uint32_t mem_mask)
+void mayjinsn_state::latch32_w(offs_t offset, uint32_t data, uint32_t mem_mask)
 {
-	if(!offset)
+	if (!offset)
 		m_latches[2] |= 1;
 	COMBINE_DATA(&m_latches[offset]);
 	machine().scheduler().synchronize();
```
cuavas, line 205: Not an introduced issue, but this isn’t the right way to use `synchronize()` (it isn't actually waiting for synchronisation to do the update.  Same in the other place.

src/mame/seta/st0016.cpp
```diff
@@ -13,10 +13,12 @@ DEFINE_DEVICE_TYPE(ST0016_CPU, st0016_cpu_device, "st0016_cpu", "ST0016")
...
 void st0016_cpu_device::cpu_internal_map(address_map &map)
 {
-	//map(0x0000, 0x7fff).rom(); ROM, Fixed area
-	//map(0x8000, 0xbfff).bankr("rombank"); ROM, Bankswitched area
-	map(0xc000, 0xcfff).r(FUNC(st0016_cpu_device::sprite_ram_r)).w(FUNC(st0016_cpu_device::sprite_ram_w));
-	map(0xd000, 0xdfff).r(FUNC(st0016_cpu_device::sprite2_ram_r)).w(FUNC(st0016_cpu_device::sprite2_ram_w));
+	// ROM, Fixed area
+	map(0x0000, 0x7fff).lr8(NAME([this](offs_t offset) -> u8 { return m_extrom_cache.read_byte(m_fixedrom_offset + (offset & 0x7fff)); }));
+	// ROM, Bankswitched area
+	map(0x8000, 0xbfff).lr8(NAME([this](offs_t offset) -> u8 { return m_extrom_cache.read_byte((m_rom_bank << 14) | (offset & 0x3fff)); }));
```
cuavas, line 19: Going through an extra address space like this will definitely hurt performance.  Is anything other than ROM ever connected here?  It may be better to just allow configuring a ROM region, then setting up banking on start.

cuavas:
I merged this as the cleanup is useful.  We’ll take the performance hit for now, it can be improved later.

PR #14358, opened 2025-10-18
Add internal layout panels for the VFX family of keyboards (vfx, vfxsd, sd1 and sd132).

cbrunschen:
This includes implementing some more display commands that are used on the VFX family, and correcting edge-of-screen behaviour to match what is observed there.

Also adds underlines to the 1x22 character VFD, for consistency with the 2x40 displays.

### DIFF
--- a/src/mame/ensoniq/esqpanel.h
+++ b/src/mame/ensoniq/esqpanel.h
@@ -117,7 +117,7 @@ public:
 		VFX,
 		VFX_SD,
 		SD_1,
-		SD_1_32		
+		SD_1_32
 	};
 
 protected:
--- a/src/mame/layout/esq2by40_vfx.lay
+++ b/src/mame/layout/esq2by40_vfx.lay
@@ -9,32 +9,32 @@ license:CC0-1.0
 
   <!-- The VFD elements -->
   <element name="segments" defstate="0">
-    <led14seg>
-      <color red="0.45" green="1.0" blue="0.95" />
-    </led14seg>
+	<led14seg>
+	  <color red="0.45" green="1.0" blue="0.95" />
+	</led14seg>
   </element>
 
   <element name="dot" defstate="0">
-    <disk statemask="0x4000" state="0"><color red="0.06" green="0.12" blue="0.10" /></disk>
-    <disk statemask="0x4000" state="0x4000"><color red="0.45" green="1.00" blue="0.95" /></disk>
+	<disk statemask="0x4000" state="0"><color red="0.06" green="0.12" blue="0.10" /></disk>
+	<disk statemask="0x4000" state="0x4000"><color red="0.45" green="1.00" blue="0.95" /></disk>
   </element>
 
   <element name="underline" defstate="0">
-    <rect statemask="0x8000" state="0"><color red="0.06" green="0.12" blue="0.10" /></rect>
-    <rect statemask="0x8000" state="0x8000"><color red="0.45" green="1.00" blue="0.95" /></rect>
+	<rect statemask="0x8000" state="0"><color red="0.06" green="0.12" blue="0.10" /></rect>
+	<rect statemask="0x8000" state="0x8000"><color red="0.45" green="1.00" blue="0.95" /></rect>
   </element>
 
   <group name="vfd_cell">
-    <bounds x="0" y="0" width="342" height="572" />
-    <element ref="segments" name="~input~"><bounds x="50" y="69" width="214" height="311" /></element>
-    <element ref="dot" name="~input~"><bounds x="253" y="337" width="42" height="42" /></element>
-    <element ref="underline" name="~input~"><bounds x="43" y="441" width="183" height="25" /></element>
+	<bounds x="0" y="0" width="342" height="572" />
+	<element ref="segments" name="~input~"><bounds x="50" y="69" width="214" height="311" /></element>
+	<element ref="dot" name="~input~"><bounds x="253" y="337" width="42" height="42" /></element>
+	<element ref="underline" name="~input~"><bounds x="43" y="441" width="183" height="25" /></element>
   </group>
 
   <element name="vfd_background">
-    <rect>
-      <color red="0.0" green="0.0" blue="0.0" />
-    </rect>
+	<rect>
+	  <color red="0.0" green="0.0" blue="0.0" />
+	</rect>
   </element>
 
 	<view name="Default Layout">
--- a/src/mame/layout/sd1.lay
+++ b/src/mame/layout/sd1.lay
@@ -3,1195 +3,1195 @@
 
   <!-- Decoration definitions -->
   <element name="triangle_up">
-    <image>
-      <data>
-        <![CDATA[
-          <svg width="2" height="1" viewBox="0 0 2 1">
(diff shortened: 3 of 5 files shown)


src/mame/ensoniq/esqpanel.cpp
```diff
@@ -5,10 +5,26 @@
...
+#include "vfxsd.lh"
+#include "sd1.lh"
+
+#include <algorithm>
+
+
+#define VERBOSE 0
+#include "logmacro.h"
+
+template <typename Format, typename... Params> inline static void logerror(Format &&fmt, Params &&... args) {
```
galibert: what?

src/mame/ensoniq/esqvfd.cpp
```diff
@@ -116,6 +117,110 @@ static const uint16_t font[]=
 	0x0000, // 0000 0000 0000 0000 (DEL)
 };
 
+/**
+ * Character data as used in the VFX family. Bit 0..13 are ordered to match
+ * led14seg. Bit 14 activates the dot after a character; bit 15 the underline.
+ * See layout/{vfx|vfxsd|sd1}.lay . 
+ */
+static const uint16_t font_vfx[] = {
```
galibert: tradition by now is to make that a device rom

src/mame/ensoniq/esqpanel.cpp
```diff
@@ -679,28 +718,152 @@ esqpanel2x40_device::esqpanel2x40_device(const machine_config &mconfig, const ch
 
 void esqpanel2x40_vfx_device::device_add_mconfig(machine_config &config)
 {
-	ESQ2X40(config, m_vfd, 60);
+	ESQ2X40_VFX(config, m_vfd, 60);
+
+	const std::string &name = owner()->shortname();
```
galibert: switching on a name of something elsewhere has often been problematic over time.  Could you switch that to a config method called at the upper level in machinf config?

src/mame/ensoniq/esqpanel.cpp
```diff
@@ -679,28 +713,154 @@ esqpanel2x40_device::esqpanel2x40_device(const machine_config &mconfig, const ch
...
 {
-	ESQ2X40(config, m_vfd, 60);
+	ESQ2X40_VFX(config, m_vfd, 60);
+
+	if (m_panel_type == VFX)
+		config.set_default_layout(layout_vfx);
+	else if (m_panel_type == VFX_SD)
+		config.set_default_layout(layout_vfxsd);
+	else if (m_panel_type == SD_1 || m_panel_type == SD_1_32)
+	  config.set_default_layout(layout_sd1);
```
galibert: something went wrong with the indentation just there

src/mame/ensoniq/esqvfd.cpp
cbrunschen: OK - does that mean as a ROM file to go along with the dumped ROMs, Or something else? Could you point me to an example?

src/mame/ensoniq/esqpanel.cpp
cbrunschen: Indeed, removed.

src/mame/ensoniq/esqvfd.cpp
```diff
@@ -293,15 +450,78 @@ bool esq2x40_device::write_contents(std::ostream &o)
...
 esq2x40_device::esq2x40_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
-	esqvfd_device(mconfig, ESQ2X40, tag, owner, clock, make_dimensions<2, 40>(*this))
+	esq2x40_device(mconfig, ESQ2X40, tag, owner, clock)
 {
 }
 
+esq2x40_vfx_device::esq2x40_vfx_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock) :
+	esq2x40_device(mconfig, ESQ2X40_VFX, tag, owner, clock)
+{
+}
```
galibert, line 385: missing blank line

src/mame/ensoniq/esqvfd.cpp
galibert: The sin table from swp30.cpp for instance?  Or the font in mulcd.cpp.

src/mame/ensoniq/esqpanel.cpp
cbrunschen: I'm now passing the "panel type" in when I add the panel in esq5505_device's various configuration methods; is that better? 

I also tried just using `set_default_layout()` in the esq5505_device's various configuration methods, but then the layout tries to find its input ports etc directly in that device rather than in the panel where they actually live (on the real devices, the buttons are connected to the display panel board, not to the main CPU board).

galibert: Yep, passing though the constructor is a good method.  You're not going to break it by casually changing a device or driver name that way.

src/mame/ensoniq/esqvfd.cpp
cbrunschen: Like this?

[esq2x40_vfx.zip](https://github.com/user-attachments/files/22990910/esq2x40_vfx.zip)

src/mame/ensoniq/esqpanel.cpp
cbrunschen: Yes, some spaces snuck in instead of a tab.

src/mame/ensoniq/esqvfd.cpp
cbrunschen: Should be fixed

src/mame/ensoniq/esqvfd.cpp
galibert: Looks good

src/mame/ensoniq/esqpanel.cpp
rb6502: I always run the MAME srcclean tool on files before I commit them.  It helps normalize tabs/spaces and everything.

src/mame/ensoniq/esqvfd.cpp
cbrunschen: I tried to do the same thing with the existing font in there; but that resulted in every variant needing its own copy of that ROM file. That does work, but the duplication of that ROM seems a bit suboptimal. Also, the "esq2x40_device" doesn't seem to be used anywhere; instead there's an "esq2x16_device" in a separate file, esqlcd.cpp, which nevertheless inherits from esqvfd_device, but which clearly does not need the VFD font. I think that's a situation that warrants sorting out before trying to move the existing font into a ROM file.

src/mame/ensoniq/esqpanel.cpp
cbrunschen: I shall do that!

MooglyGuy:
Looks good to my eyes, but I leave it up to @galibert to merge since he's been doing the review. :)

galibert:
I'm waiting for the CI to pass, it gets reset with each push.

cbrunschen:
It's definitely better to make sure it's all good. CI and such are there for a reason!

cbrunschen:
Thank you!

felipesanches:
Thanks, @cbrunschen!

PR #14364, opened 2025-10-19
VET 3000, "The Video Effects Titler"

felipesanches:
by Tecnologia em Micro Sistemas (Brazilian company), 1988

* https://datassette.org/node/106385
* https://www.youtube.com/watch?v=DJXlqe2UXKs

Screenshot From 2025-10-18 21-24-14

### DIFF
--- a/src/mame/tms/vet3000.cpp
+++ b/src/mame/tms/vet3000.cpp
@@ -70,18 +70,19 @@ void vet3000_state::machine_start()
 };
 
 
-void vet3000_state::keyboard_w(u8 data){
+void vet3000_state::keyboard_w(u8 data)
+{
 	m_scan = data;
 };
 
 
 u8 vet3000_state::keyboard_r()
 {
+	u8 value = 0xff;
 	for (u8 i=0; i<7; i++)
 		if (!BIT(m_scan, i))
-			return m_row[i]->read();
-
-	return 0xff;
+			value &= m_row[i]->read();
+	return value;
 }
 
 


src/mame/tms/vet3000.cpp
```diff
@@ -0,0 +1,198 @@
...
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_C)
+	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_UNUSED)
+	PORT_BIT(0x10, IP_ACTIVE_LOW, IPT_OTHER) PORT_NAME("LEFT/RIGHT") PORT_CODE(KEYCODE_LEFT)
+	PORT_BIT(0x20, IP_ACTIVE_LOW, IPT_OTHER) PORT_NAME("CLEAR") PORT_CODE(KEYCODE_BACKSPACE)
+	PORT_BIT(0x40, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_V)
+	PORT_BIT(0x80, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_B)
+INPUT_PORTS_END
+
+
+#define MAIN_CLOCK 3.579545_MHz_XTAL
```
MooglyGuy: Would you be alright with making this a static member of the class instead of using a #define? I'm not super bothered about it either way, but it caught my eye by looking a bit out-of-place here.

src/mame/tms/vet3000.cpp
```diff
@@ -0,0 +1,199 @@
...
+void vet3000_state::keyboard_w(u8 data){
+	m_scan = data;
+};
+
+
+u8 vet3000_state::keyboard_r()
+{
+	for (u8 i=0; i<7; i++)
+		if (!BIT(m_scan, i))
+			return m_row[i]->read();
```
galibert: Isn't that an and of all the outputs for which a m_scan bit is zero?

src/mame/tms/vet3000.cpp
```diff
@@ -0,0 +1,199 @@
...
+};
+
+
+void vet3000_state::machine_start()
+{
+	save_item(NAME(m_scan));
+};
+
+
+void vet3000_state::keyboard_w(u8 data){
```
galibert: Random { at end of line.

src/mame/tms/vet3000.cpp
felipesanches: Yes, I think that's more accurate, indeed. I've just fixed it. Thanks!

src/mame/tms/vet3000.cpp
felipesanches: Also fixed the formatting now.

felipesanches:
Thanks!

src/mame/tms/vet3000.cpp
```diff
@@ -0,0 +1,200 @@
...
+static INPUT_PORTS_START( vet3000 )
+	PORT_START("ROW1")
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_1)
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_Q)
+	PORT_BIT(0x04, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_A)
+	PORT_BIT(0x08, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_LSHIFT) PORT_CODE(KEYCODE_RSHIFT)
+	PORT_BIT(0x10, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_7)
+	PORT_BIT(0x20, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_U)
+	PORT_BIT(0x40, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_J)
+	PORT_BIT(0x80, IP_ACTIVE_LOW, IPT_OTHER) PORT_CODE(KEYCODE_N)
```
cuavas, line 107: You should really be using `IPT_KEYBOARD` and `PORT_NAME` or `PORT_CHAR` for all the keys.  It’s not going to be a good experience like this.

PR #14365, opened 2025-10-19
sega/model2: rewrite model2rd.ipp to use C++ function templates

gm-matthew:
The existing code, originally written back when MAME used C, effectively mimics C++ function templates. Since MAME uses C++ nowadays, I thought it would make sense to rewrite it to use actual function templates.

Also I'm provisionally adding myself to the credits for the Model 2 driver; I hope I've done enough to earn it :-)

### DIFF
--- a/src/mame/sega/model2rd.ipp
+++ b/src/mame/sega/model2rd.ipp
@@ -5,9 +5,13 @@
     Sega Model 2 3D rasterization functions
 
 ********************************************************************/
+#ifndef MAME_SEGA_MODEL2RD_IPP
+#define MAME_SEGA_MODEL2RD_IPP
+
+#pragma once
 
 
-/* non-textured render path */
+// non-textured render path
 template <bool Checker, bool Translucent>
 void model2_renderer::draw_scanline_solid(int32_t scanline, const extent_t& extent, const m2_poly_extra_data& object, int threadid)
 {
@@ -15,16 +19,16 @@ void model2_renderer::draw_scanline_solid(int32_t scanline, const extent_t& exte
 	u32 *const p = &m_destmap.pix(scanline);
 	u8  *gamma_value = &state->m_gamma_table[0];
 
-	/* extract color information */
+	// extract color information
 	const u16 *colortable_r = &state->m_colorxlat[0x0000/2];
 	const u16 *colortable_g = &state->m_colorxlat[0x4000/2];
 	const u16 *colortable_b = &state->m_colorxlat[0x8000/2];
 	u32  color = object.colorbase;
 	u8   luma;
 	u32  tr, tg, tb;
-	int     x;
+	int  x;
 
-	/* if it's translucent, there's nothing to render */
+	// if it's translucent, there's nothing to render
 	if (Translucent)
 		return;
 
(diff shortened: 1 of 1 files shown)


galibert:
DId you benchmark it a little to see if it had any impact?

src/mame/sega/model2.h
```diff
@@ -684,63 +676,14 @@ class model2_renderer : public poly_manager<float, m2_poly_extra_data, 4>
...
-
-	scanline_render_func m_renderfuncs[8];
+	template <bool checker, bool translucent>
+	void draw_scanline_solid(int32_t scanline, const extent_t& extent, const m2_poly_extra_data& object, int threadid);
+
+	template <bool translucent>
+	u32 fetch_bilinear_texel(const m2_poly_extra_data& object, const u32 miplevel, const float fu, const float fv);
+
+	template <bool checker, bool translucent>
+	void draw_scanline_tex(int32_t scanline, const extent_t& extent, const m2_poly_extra_data& object, int threadid);
```
cuavas, line 686: Please `LlamaCase` for template parameter names.

src/mame/sega/model2_v.cpp
```diff
@@ -851,35 +852,19 @@ void model2_renderer::model2_3d_render(triangle *tri, const rectangle &cliprect)
...
+	case 0: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_solid<false, false>, this), tri->v[0], tri->v[1], tri->v[2]); break;
+	case 1: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_solid<false, true>, this),  tri->v[0], tri->v[1], tri->v[2]); break;
+	case 2: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_tex<false, false>, this),   tri->v[0], tri->v[1], tri->v[2]); break;
+	case 3: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_tex<false, true>, this),    tri->v[0], tri->v[1], tri->v[2]); break;
+	case 4: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_solid<true, false>, this),  tri->v[0], tri->v[1], tri->v[2]); break;
+	case 5: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_solid<true, true>, this),   tri->v[0], tri->v[1], tri->v[2]); break;
+	case 6: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_tex<true, false>, this),    tri->v[0], tri->v[1], tri->v[2]); break;
+	case 7: render_triangle<3>(vp, render_delegate(&model2_renderer::draw_scanline_tex<true, true>, this),     tri->v[0], tri->v[1], tri->v[2]); break;
```
cuavas, line 866: Delegates are kind of heavy due to all their flexibility, and `render_triangle` makes several copies of one.  I wonder if there’s a more lightweight way to do it.  (Not something to take on in this PR, just idle thoughts.)

cuavas:
> DId you benchmark it a little to see if it had any impact?

From looking at it, I doubt it would make a difference.  It generates the same set of functions, just without macros.  It always calls them via a delegate anyway, so they can’t ever get inlined.

gm-matthew:
> Did you benchmark it a little to see if it had any impact?

From my testing it doesn't seem to have any impact on performance.

cuavas (approved):
I approve of the code changes themselves.

@galibert how much has @gm-matthew contributed to these sources?  Is it enough for the copyright change?

angelosa:
> @galibert how much has @gm-matthew contributed to these sources? Is it enough for the copyright change?

Why not, it's fine to me given the non-trivial work done.

galibert:
It's fine for me too, including the copyright.  I wanted to be sure there wasn't a performance degradation because compilers are sometimes surprising.

PR #14375, opened 2025-10-19
Implement game exclusion support for hiscore plugin (enable/disable hiscore per game)

aaciii:
Added support for enabling/disabling hiscore support on a game-by-game basis. 

- Implemented functions to create/save and load exclusion lists in json format
- Updated existing plugin functions to check for a game exclusion before loading / saving high scores
- Updated menu options to enable/disable high-score support per game

I created this update because at least one game - Food Fight (foodf) - crashes upon reload when the existing hiscore plugin is used (due to an NVRAM corruption issue).   This made me want to be able to turn the plugin on and off on a per-game basis, however that is not possible (it's all-or-nothing).  There are likely other reasons that someone may want to turn off hiscore support for certain games.

This update is backward compatible with the existing 1.0.1 plugin; it does not modify its basic functions nor alter any of the existing files (json/dat).    This is a drop-in replacement.   Only the init.lua file has been modified.

This plugin update creates new file 'exclude_games.json' which stores the list of games for which hiscore should be disabled.

### DIFF
--- a/plugins/hiscore/init.lua
+++ b/plugins/hiscore/init.lua
@@ -1,16 +1,21 @@
--- hiscore.lua
--- by borgar@borgar.net, CC0 license
---
+-- hiscore.lua with exclusion support
 -- This uses MAME's built-in Lua scripting to implement
 -- high-score saving with hiscore.dat infom just as older
 -- builds did in the past.
 --
+-- original by borgar@borgar.net, CC0 license
+-- updated by aaciii@yahoo.com to enable/disable hiscore support per game
+--
+local json = require('json')
+local lfs = require('lfs')
+
 local exports = {
 	name = 'hiscore',
-	version = '1.0.1',
-	description = 'Hiscore',
-	license = 'CC0',
-	author = { name = 'borgar@borgar.net' } }
+    version = '1.1.0',
+    description = 'Hiscore with per-game exclusion support',
+    license = 'CC0',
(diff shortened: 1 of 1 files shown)


aaciii:
Hello @cuavas --  thanks for reviewing.   A few quick thoughts on your comments:

- a JSON file containing all the excluded games scales as well as any methodology to list ROMS that should be excluded; the list will "by definition" be a handful of named ROMs at best.    (In fact there is only one game I know of that REQURES this exclusion, because the hiscore plugin crashes the game.   That game is foodf.)    There are many instances in MAME where lists-of-roms are used ...  in fact the hiscore plugin ITSELF already uses a single file listing out all the (many) games that are hiscore compatible and the data needed to write scores to/from RAM.

- your coding & commenting standards are your prerogative, but when I modify someone's code, especially that of a single author, notating what was changed and why is prudent for their review.   As this is my first MAME submission I would have expected the prior author to want to understand what was modified (and as importantly, what was not modified).

galibert:
Explanations belong in the commit text, not the source file.  We're not in the age of cvs anymore ;-)

aaciii:
> Explanations belong in the commit text, not the source file. We're not in the age of cvs anymore ;-)

Fair enough...   by my count I added just 4 comments that mention changes from prior version (3 of which are "new function added for purpose X" and one of which is "re-ordered menu items."    I am happy to remove/revise those!

Most of the comments in the code (VAST majority) were already there when I came along :)

PR #14381, opened 2025-10-20
Really add 16:9 support to Sega Touring Car Championship

AnthonyRyuki:


galibert:
Did you actually bother to test it this time?

AnthonyRyuki:
I tested it and it works.

PR #14384, opened 2025-10-20
Follow-up to PR#14263 -- move implementation details out of headers etc.

shattered:


galibert:
CI failure, src/devices/bus/rc2014/sound.cpp does not see device_rc2014_card_interface.  Which means you either didn't even try to compile or changed something just before pushing.  Bad idea either way.

src/devices/bus/rc2014/sound.cpp
```diff
@@ -7,6 +7,7 @@
 ****************************************************************************/
 
 #include "emu.h"
+#include "sound.h"
```
galibert: ok, the naming is a bad idea, it collides with src/emu/sound.h

shattered: right, I copy pasted this into bk/ay.cpp and Vas pointed out that sound.h was not needed there

galibert: I mean you should rename sound.cpp/h from the rc2014 into something else so that sound.h is unambiguously the one from src/emu.

src/devices/bus/bk/loopback.h
```diff
@@ -14,30 +14,6 @@
 #include "parallel.h"
 
 
-//**************************************************************************
-//  TYPE DEFINITIONS
-//**************************************************************************
+DECLARE_DEVICE_TYPE(BK_LOOPBACK, device_qbus_card_interface)
```
cuavas, line 17: Ah, this doesn’t even match the definition – it’s declared as exposing `device_qbus_card_interface` here, but actual public interface is `device_bk_parallel_interface`.  Does this work at all?

shattered: Yes, it does

cuavas: It only “worked” by chance.  I can’t believe I have to add a pile of code to explicitly guard against doing this.

PR #14391, opened 2025-10-21
midw8080/8080bw.cpp: added support for Cosmo's starfield [Mike Coates]

Osso13:
Added stars and video effects to Cosmo based on Guru's reversed Cosmo PCB schematic [Mike Coates, Guru]


Osso13:
Original code was done on a 6 year old base, so I updated it to compile with the current codebase (no READ and WRITE macros, etc).

PR #14393, opened 2025-10-21
qbus: add DVK MFM hard disk controller HLE

shattered:


src/devices/bus/qbus/dvk_dwhle.cpp
```diff
@@ -0,0 +1,557 @@
...
+	clear_virq(m_bus->birq4_w, m_regs[REGISTER_SI], CSR_IE, m_drqa); } while (0)
+
+#define raise_drqb()	do { \
+	m_regs[REGISTER_SI] |= KZDSI_DRQ1; \
+	raise_virq(m_bus->birq4_w, m_regs[REGISTER_SI], CSR_IE, m_drqb); } while (0)
+
+#define clear_drqb()	do { \
+	m_regs[REGISTER_SI] &= ~KZDSI_DRQ1; \
+	clear_virq(m_bus->birq4_w, m_regs[REGISTER_SI], CSR_IE, m_drqb); } while (0)
+
```
galibert: Any reasons those aren't just normal automatically inlined methods?

shattered: Right, does this look fine now?

galibert: Way nicer.

PR #14394, opened 2025-10-21
mz2000_cass: Addition of 7 games to software list

paul75s:
mz2000_cass: Gang Man, Ice Block, Itasandrias, Justice Knight, Super Doors, Zero Fighter [TOSEC]

mz2000_cass: Polar Star II [TOSEC]

### DIFF
--- a/hash/mz2000_cass.xml
+++ b/hash/mz2000_cass.xml
@@ -230,7 +230,7 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
 	<software name="iceblock" supported="yes">
 		<description>Ice Block</description>
 		<year>1983</year>
-		<publisher>Db Soft</publisher>
+		<publisher>dB Soft</publisher>
 		<info name="alt_title" value="アイス・ブロック"/>
 		<info name="usage" value="IPL"/>
 		<part name="cass1" interface="mz_cass">


hash/mz2000_cass.xml
```diff
@@ -275,6 +328,19 @@ Loading error, boots with Z80 clock set at 3.2 MHz (80%)
 		</part>
 	</software>
 
+	<software name="plrstar2" supported="partial">
+		<description>Polar Star II</description>
+		<year>1983</year>
+		<publisher>Hyo-Kin Soft</publisher>
+				<notes><![CDATA[Loads but gameplay won't start]]></notes>
```
angelosa, line 335: That would be a supported no

paul75s: Modified to supported no

angelosa:
In order to merge this we need two things:
1. files being submitted to https://wiki.mamedev.org/index.php?title=Submitting_Source_Code#Submitting_ROM_or_disk_dumps , being .wav files a pointer to an external URL host should be fine (i.e. attaching those directly in the email won't work).
2. a credit source for the whatsnew, are these real dumps or like converted from .mzt?

paul75s:
> In order to merge this we need two things:
> 
> 1. files being submitted to https://wiki.mamedev.org/index.php?title=Submitting_Source_Code#Submitting_ROM_or_disk_dumps , being .wav files a pointer to an external URL host should be fine (i.e. attaching those directly in the email won't work).
> 2. a credit source for the whatsnew, are these real dumps or like converted from .mzt?

I will submit the files as described above. 

These are conversions from mzt - should I just confirm that when I submit the files?

angelosa:
> These are conversions from mzt - should I just confirm that when I submit the files?

Not necessary. I'm not sure about the source, probably available on IA anyway?

Other questions:
1. did you try any of these from direct .mzt format?
2. I have a Polar Star II .mzt, it doesn't work because that dump dislikes the cassette timing. Downclocking main CPU to 80% makes it boot (as in common with other entries in the list).

paul75s:
> > These are conversions from mzt - should I just confirm that when I submit the files?
> 
> Not necessary. I'm not sure about the source, probably available on IA anyway?
> 
> Other questions:
> 
> 1. did you try any of these from direct .mzt format?
> 2. I have a Polar Star II .mzt, it doesn't work because that dump dislikes the cassette timing. Downclocking main CPU to 80% makes it boot (as in common with other entries in the list).

Yes, I tried the original MZT for all these and they didn't work, all seen to be due to the cassette timing issue. 

As with a few other working items in the list, converting to WAV is resolving that timing issue.

wilbertpol:
> > > These are conversions from mzt - should I just confirm that when I submit the files?
> > 
> > 
> > Not necessary. I'm not sure about the source, probably available on IA anyway?
> > Other questions:
> > 
> > 1. did you try any of these from direct .mzt format?
> > 2. I have a Polar Star II .mzt, it doesn't work because that dump dislikes the cassette timing. Downclocking main CPU to 80% makes it boot (as in common with other entries in the list).
> 
> Yes, I tried the original MZT for all these and they didn't work, all seen to be due to the cassette timing issue.
> 
> As with a few other working items in the list, converting to WAV is resolving that timing issue.

That is then more an indication that there is a problem with our handling of those .mzt files.
We generally prefer to use files in wav format for direct recordings from original tapes.

paul75s:
> > > > These are conversions from mzt - should I just confirm that when I submit the files?
> > > 
> > > 
> > > Not necessary. I'm not sure about the source, probably available on IA anyway?
> > > Other questions:
> > > 
> > > 1. did you try any of these from direct .mzt format?
> > > 2. I have a Polar Star II .mzt, it doesn't work because that dump dislikes the cassette timing. Downclocking main CPU to 80% makes it boot (as in common with other entries in the list).
> > 
> > 
> > Yes, I tried the original MZT for all these and they didn't work, all seen to be due to the cassette timing issue.
> > As with a few other working items in the list, converting to WAV is resolving that timing issue.
> 
> That is then more an indication that there is a problem with our handling of those .mzt files. We generally prefer to use files in wav format for direct recordings from original tapes.

Agree, as these MZT files do work with other MZ2200 emulators. Also seems to be a wider issue with Sharp MZ series as I've found similar issues (and solutions) for MZ700/800/1500. 

However, I assumed this approach was acceptable as a number of other additions to this software list in recent months have been MZT to WAV conversions.

angelosa:
> 
> That is then more an indication that there is a problem with our handling of those .mzt files. We generally prefer to use files in wav format for direct recordings from original tapes.

afaik it's because they expect a different data format wrt MZ-700/MZ-800, that we don't handle.
https://www.sharpmz.org/mz-700/tapeproc.htm

paul75s:
> > That is then more an indication that there is a problem with our handling of those .mzt files. We generally prefer to use files in wav format for direct recordings from original tapes.
> 
> afaik it's because they expect a different data format wrt MZ-700/MZ-800, that we don't handle. https://www.sharpmz.org/mz-700/tapeproc.htm

I've managed to get a fair few non working MZ700/MZ800 MZF files working through conversion to WAV - thought I'd start initially with this handful of MZ2200 files but then add them to software lists if this approach is OK. 

I've got >100 software items working with the MZ2200 core, a mix of tapes, floppies and dumps. Keen to add them all to the relevant software lists but get the approach right as most have needed some clean up or conversion to get them working. 

What would be your preferred approach here: 
1.  Proceed with these WAV conversions so they work with the core as is in original tape format. 
2. Add the MZT files instead, so they're as I found them but only work with under clocking. 
3. Create dumps from the original BIN files within the MZT files and create a new software list for these.  These also work fine so provide running software, but as binary dumps rather than converted tapes. 
4. Take those BIN dumps and create new D88 floppies. This feels like the worst answer here but can do this.

hash/mz2000_cass.xml
```diff
@@ -214,6 +227,32 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
 		</part>
 	</software>
 
+	<software name="iceblock" supported="yes">
+		<description>Ice Block</description>
+		<year>1983</year>
+		<publisher>Db Soft</publisher>
```
angelosa: `dB-Soft`, assuming it's spelled that way on title screen and it's not like Vosque / Vosque 2000 where it's spelled `d.B Soft`

angelosa:
I'm okay with the WAV conversions for now, there are a bunch already done in hash list anyway. 

I presume the origin of these is "the infamous 200x P2P leak" or an actual group collection like TOSEC/Neo Kobe etc. upped these?

hash/mz2000_cass.xml
paul75s: Just checking and it's neither - my error, it actually shows as dB Soft (not Db Soft), so will correct this. I'm assuming the preference would be accuracy for the software item (so dB Soft) rather than go with d.B Soft for consistency with the existing entry for Vosque?

paul75s:
> I'm okay with the WAV conversions for now, there are a bunch already done in hash list anyway.
> 
> I presume the origin of these is "the infamous 200x P2P leak" or an actual group collection like TOSEC/Neo Kobe etc. upped these?

They're from a couple of TOSEC collections.

hash/mz2000_cass.xml
angelosa: Generally it's dB Soft, Vosque is a one-off really (lacking an actual scan of the tape cover).

paul75s: Amended to dB Soft

ICEknigh7:
Does inverting the polarity in the WAVs cause them to fail loading like the original MZTs?
Asking because: https://mametesters.org/view.php?id=8875

paul75s:
Apologies for a dumb newbie question - following approval of this change, is there anything further I need to do for this to be merged into master?

angelosa:
Approving a PR for MAME just means I'll eventually apply it if nobody says anything in a couple days.

paul75s:
> Approving a PR for MAME just means I'll eventually apply it if nobody says anything in a couple days.

Thanks, thought that was likely to be the case but wanted to double check.

Issue #9503, opened 2022-04-03
apple2e driver - disabled Zip Chip incorrectly re-enabled by timer


inexorabletash:
The IIc+ and non-IIc+ Zip Chip emulation differ in how writes to C05A/C05B affect the accelerator's behavior in response to temporary slowdowns. When disabling the accelerator through a write to C05A, the IIc+ version sets the `m_accel_fast` flag to false; the non-IIc+ version does not. The consequence of this is that when the temporary slowdown timer fires, the IIc+ accelerator remains disabled, but non-IIc+ accelerator gets re-enabled!

Repro in 6502 code:

```
  LDA #$5A
  STA $C05A
  STA $C05A
  STA $C05A
  STA $C05A ; unlock
  LDA #$00  ; anything but 5A and A5
  STA $C05A ; disable
  LDA #$A5
  STA $C05A ; lock

  STA $C070 ; trigger gameio timer
```

Verified (second hand, c/o @apple2geek) on a real IIe w/ Zip Chip, the above should leave the accelerator disabled when the 5ms timer fires. Note that the behavior should be the same regardless of whether the C070 access happens _before_ or _after_ the unlock/disable/lock (again, verified second hand on actual hardware).

From code inspection, the fix is likely to just be copying the two `m_accel_fast = ...;` lines from the IIc+ code into the normal code, for the 0x5a and 0x5b cases. 

cc @rb6502 @peterferrie

rb6502:
Thanks for the report!  I already have a preliminary fix, but nothing good comes of git commits after midnight so I'm not going to push it quite yet. 😅

rb6502:
Should be all set, that was definitely an oversight.  I will be getting to the Franklin function keys hopefully soon too.

PR #14397, opened 2025-10-22
apple: slightly improve Zip Chip emulation

arekkusu42:
I'm going to split my Zip Chip changes into two PRs. This first smaller one fixes problems in the existing implementation, and implements two useful status bits:

* apple2gs: fix Zip Chip inconsistencies
The [previous PR](http://github.com/mamedev/mame/pull/14388) made the IIc consistent with the IIe, now we do the same for the IIgs.  The IIgs had the same problem as #9503, and it was also erroneously re-locking the Zip when disabling via C05A. 

* apple: improve Zip Chip C05B
The disable bit allows querying the current state; this is very useful when writing vaporlock beam-racing tests ^\_^;
The delay bit allows measuring delays, which may only be useful to folks trying to write Zip Chip unit tests ^\_^;

These changes have been verified against a ZipGSX in an Apple IIgs ROM3.
The C05B status bit behaviors are described in the [1987 Zip manual](http://mirrors.apple2.org.za/ftp.apple.asimov.net/documentation/hardware/accelerators/zip_instruction_manual.pdf#page=12) (but due to inconsistent documentation I am not implementing bit 5 in this PR.)
The real annunciator side-effect behavior is more complex, and is handled in a WIP PR.

Testing:
* Total Replay continues to work, enabling the Zip Chip for decompression and disabling for play (or relying on CPS Follow on the IIgs.)
* Zip Chip Utilities are unchanged (System Check passes, Cache Diagnostics fail.)
* The ZipGS CDA and CDEV are improved: the initial "Sorry, no ZipGS present"/"No ZipGS detected" error is fixed, and the Misc Settings now display accurately.

I also have a WIP test suite verifying all of the remaining softswitches and side-effects, which identifies many failures in 0.281 _before_ this PR:
ZIPTEST_apple2gs_before

And a little better _after_ (disabling no longer locks, disable & delay status works, delays can be measured): 
ZIPTEST_apple2gs_after

Hardware looks like this:
ZIPTEST_apple2gs_HW
and with my WIP PR, ZipGS emulation passes all of this. However, I am holding off on that PR until I can [locate a volunteer](http://apple2infinitum.slack.com/archives/CABEM8JFK/p1760827313370039) to verify the behavior of the the 8-bit Zip Chip in a IIe or IIc+.  (There is enough difference between the Zip Chip and ZipGS that I can't just guess at the behavior.)

### DIFF
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -2105,18 +2105,18 @@ u8 apple2e_state::c000_r(offs_t offset)
 
 			if (m_accel_unlocked)
 			{
-				if (offset == 0x5b)
+				if (offset == 0x5b) // Zip status flags
 				{
+					// bits 0-1 are cache size; [8, 16, 32, 64]kB
+					u8 b01 = 0x03;
+					// bit 3 is set if a temporary delay is active due to slot or softswitch access
+					u8 b3 = m_accel_temp_slowdown ? 0x08 : 0x00;
+					// bit 4 is set if the Zip is disabled
+					u8 b4 = m_accel_fast ? 0x00 : 0x10;
 					// bit 7 is a 1.0035 millisecond clock; the value changes every 0.50175 milliseconds
 					const int time = machine().time().as_ticks(1.0f / 0.00050175f);
-					if (time & 1)
-					{
-						return 0x03;
-					}
-					else
-					{
-						return 0x83;
-					}
+					u8 b7 = time & 1 ? 0x80 : 0x00;
+					return b7 | b4 | b3 | b01;
 				}
 				else if (offset == 0x5c)
 				{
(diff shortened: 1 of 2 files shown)


arekkusu42:
cuavas cleaned up:
```
-					u8 b7 = time & 1 ? 0x80 : 0x00;
+					const u8 b7 = (time & 1) ? 0x80 : 0x00;
```
Whoops! Lesson learned.

cuavas:
> cuavas cleaned up:
> 
> ```
> -					u8 b7 = time & 1 ? 0x80 : 0x00;
> +					const u8 b7 = (time & 1) ? 0x80 : 0x00;
> ```
> 
> Whoops! Lesson learned.

In general it’s easier to analyse code in your head if you get hints like that to tell you that a local isn’t going to change after assignment.

PR #14399, opened 2025-10-22
sinclair/specnext.cpp: Improved interrupts handling

holub:


src/devices/machine/z80ctc.h
```diff
@@ -114,6 +114,7 @@ class z80ctc_device :   public device_t,
 
 	// internal helpers
 	u8 channel_int_state(int ch) const noexcept { return m_channel[ch]->m_int_state; }
+	u8 channel_mode(int ch) const noexcept { return m_channel[ch]->m_mode; }
 	void interrupt_check();
```
angelosa, line 118: Funny that this is called "internal helpers" then outside world can access it ... 
ETA: method looks protected, not exactly "internal" imo.

holub: They are protected

PR #14400, opened 2025-10-23
sequential_sixtrak.lay: Made knobs and wheels interactive.

m1macrophage:
To review the changes in the Slider script, see `linn_linndrum.lay`.

Updates to reference Slider script in `linn_linndrum.lay`:
* Added support for horizontal control of knobs.
* Added support for auto-center anolog controls. Useful for the pitch wheel.
* Shortened function names.

`sequential_sixtrak.lay`:
* Copied reference Slider script from `linn_linndrum.lay`.
* Adapted layout to make knobs and wheels interactive.

### DIFF
--- a/src/mame/layout/linn_linndrum.lay
+++ b/src/mame/layout/linn_linndrum.lay
@@ -592,19 +592,19 @@ copyright-holders:m1macrophage
 				local view = file.views["Default Layout"]
 				install_slider_callbacks(view)
 
-				add_simplecounter_knob(view, "knob_pot_tempo", "pot_tempo", 1.5)
-				add_simplecounter_knob(view, "knob_pot_volume", "pot_volume", 1.5)
+				add_knob(view, "knob_pot_tempo", "pot_tempo", 1.5, VERTICAL)
+				add_knob(view, "knob_pot_volume", "pot_volume", 1.5, VERTICAL)
 				for i = 1, 6 do
 					local knob_input = "pot_tuning_" .. i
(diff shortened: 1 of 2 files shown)


felipesanches:
Hey, @m1macrophage! That's so cool!

I was also looking for a solution for auto re-centering synth pitch wheels a couple months ago when I was working on the ESQ1 layout: https://github.com/mamedev/mame/pull/14049#issuecomment-3212443497

Nobody replied to my question at that time, but now I see that you found a solution to the same issue here, so I'll probably implement that on my synth layouts as well.

It is cool how the wheel smoothly re-centers gradually when we release the keys (right or left) after changing its position. But when interacting with the mouse cursor, the re-centering is abrupt and instantaneous. Have you considered how to achieve that same smooth "spring-like" transition also when using the mouse?

galibert:
Note that on both my hardware synths the physical pitch wheel returns to center very fast when you release it.

m1macrophage:
> Have you considered how to achieve that same smooth "spring-like" transition also when using the mouse?

I think this would either require simulating the return-to-center logic in the lua script, or (preferably) triggering the key release logic of the port from the lua script. I expect both are doable, but none of them seems trivial (unless I missed a relevant port/field lua function). So I haven't looked more closely for now.

To @galibert's point, the centering is pretty quick in practice anyway. Though the spring-like animation does *look* cooler :). And it is possible a quick, but not instant, centering will sound better.

PR #14402, opened 2025-10-23
current Plug and Play / Handhelds PR

mamehaze:
new NOT WORKING machines
-------------
Compact Cyber Arcade - Finding Dory [David Haywood, Team Europe]
H12 Pro 1000 in 1 Handheld Game Console [David Haywood, Team Europe]

new NOT WORKING clones
---------------
Compact Cyber Arcade - Paw Patrol (JL2365PA-5) [David Haywood, Team Europe]


mamehaze:
@ajrhacker h12p1000 probably needs a look to check if it's banking or a ROM issue.  If you select Warpman for example it blackscreens, even if the game data appears to be in the ROM.  There are ALU accesses and writes to unknown addresses around that time.  Definitely seems to be 369 as it accesses the sound ram area on startup.

ajrhacker:
That's not the "sound ram area"; it's the $4FFx register space that only exists on the yet-unknown VT model that still shares the VT369 driver for the moment.

mamehaze:
ah my mistake, I confused the upper accesses.

also how do I actually cancel a CL job, this one has clearly stalled, but as a regular user there appears to be no 'cancel' button anywhere.

PR #14418, opened 2025-10-26
att6300p.cpp: Fix DIP switches

donohoe00:
Confusion caused by the wrong assumption that an On switch would read as '1', and vice versa.

Wasn't breaking much since most of the DIP switches are ignored.


angelosa:
Are diplocations known for this?

donohoe00:
> Are diplocations known for this?

Not sure I understand the question.  They're shown in the schematics and their function is described in the hardware reference manual.  I also checked their correctness against comments and code in the BIOS listings.

angelosa:
MAME input ports have a system to make dips to show up graphically, around `PORT_DIPLOCATION`. May be a good idea to hook those up, it's one of those quality of life that helps immensely when comparing with real HW.

donohoe00:
> MAME input ports have a system to make dips to show up graphically, around `PORT_DIPLOCATION`. May be a good idea to hook those up, it's one of those quality of life that helps immensely when comparing with real HW.

Nice!  Makes it so much easier to verify whether they're implemented correctly.

PR #14425, opened 2025-10-27
Expose device properties to debugger -- taylorza

holub:
This allows to share more internal states with debugger.
image
... and makes possible do something like: `bp c000, mmu6==2a`

Brought by taylorza

### DIFF
--- a/src/emu/debug/debugcpu.cpp
+++ b/src/emu/debug/debugcpu.cpp
@@ -65,6 +65,12 @@ debugger_cpu::debugger_cpu(running_machine &machine)
 	m_symtable->add("wpdata", symbol_table::READ_ONLY, &m_wpdata);
 	m_symtable->add("wpsize", symbol_table::READ_ONLY, &m_wpsize);
 
+	/* add device symbols to the global symbol table */
+	for(auto &symbol : machine.root_device().get_device_symbols())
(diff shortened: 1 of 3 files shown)


ajrhacker:
I don't really like this approach of creating a brand new structure for devices to infiltrate the global symbol table with. I'd rather have the CPU (which, after all, is rather customized here) allow non-owned variables to be inserted into its state interface by calling back to the driver during its own `device_start()`, somewhat like I did with some of the Kawasaki KC80-based cores.

holub:
I did this through the cpu in the past. @cuavas doesn't like that either :(

holub:
It would be nice if we agree on cpu injection - I can rewrite this way.

Alternatively I can give a try and rewrite this without device modification. In case symbols lifecycle accepts this in driver's device_start.

CPU's way would be more useful as has the way to format the values out of the box. It's definitely doesn't work nicely for multicpu drivers.
In the long term ideally we need to expose the entire save state to debugger symbols automatically.

galibert:
Hmm, what happens if you use device_state_interface on a non-cpu?

holub:
State is exposed to active cpu? I'll give a try.

holub:
suppressed by #14898

PR #14429, opened 2025-10-27
sharp/mz2000.cpp: Addition of mz2000_snap software list

paul75s:
Introduce mz2000_snap software list containing an initial 16 working snapshots for MZ-2000/MZ-2200.

These are all sourced from various TOSEC repositories, either in snapshot format (Binary Land and Flicky) or converted from MZT, WAV or D88 cassette tape/floppy disk images which are currently unsupported in MAME, using DumpListEditor to extract the binaries.

Everything included is supported in this format. Proposed guidelines for what should be included here (vs using original cassette tape or floppy disk formats in existing software lists) included in the xml file.

### DIFF
--- a/hash/mz2000_snap.xml
+++ b/hash/mz2000_snap.xml
@@ -1,14 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
-
 <!-- 
 Software list of Sharp MZ-2000 snapshots, created to provide support for software that will run as a snapshot image in MAME but (currently) not in the original media.
-
-The initial snapshots fall into 4 categories:
-1. Software which can only be sourced in snapshot format.
-2. Software which can be sourced in original cassette tape format, but will not currently load in MAME, even using MZT to WAV conversion.
-3. Software which can be sourced in original cassette tape format, but will not currently load in MAME, but is included in the mz2000_cass software list in MZT format. MZT to WAV conversion may be successful but a snapshot has been included instead of removing the MZT file from mz2000_cass.
-4. Software which can be sourced in original floppy disk format, but will not currently load in MAME.
 -->
 <softwarelist name="mz2000_snap" description="Sharp MZ-2000 snapshots">
 


hash/mz2000_snap.xml
```diff
@@ -0,0 +1,195 @@
...
+
+<!-- 
+Software list of Sharp MZ-2000 snapshots, created to provide support for software that will run as a snapshot image in MAME but (currently) not in the original media.
+
+The initial snapshots fall into 4 categories:
+1. Software which can only be sourced in snapshot format.
+2. Software which can be sourced in original cassette tape format, but will not currently load in MAME, even using MZT to WAV conversion.
+3. Software which can be sourced in original cassette tape format, but will not currently load in MAME, but is included in the mz2000_cass software list in MZT format. MZT to WAV conversion may be successful but a snapshot has been included instead of removing the MZT file from mz2000_cass.
+4. Software which can be sourced in original floppy disk format, but will not currently load in MAME.
+-->
```
angelosa, line 5: Don't think we need this lengthy explanation, it's a 40 years old system with potentially plenty of lost media already so snapshots are an acceptable stopgap imo.

angelosa (approved):

hash/mz2000_snap.xml
paul75s: OK, will remove. Prompts a question on best approach for a few other software items I have working in WAV or snapshot format which are currently in the mz2000_cass software list (e.g. Vosque, Star Stream). Would you prefer I add the WAV files in place of the MZT files, preserving the original media format but removing MZT files that may be preferred in future as the emulation evolves, or add snapshots of these here to allow the MZT files to remain?

angelosa: If there are MZT files I'd keep them, using clones for alternative WAV files.

paul75s: Great, thanks - will do that.

paul75s: As this was approved a few weeks ago with no objections, can this now be merged?

angelosa: Because I didn't got the time to reformat the message for the additions in [whatsnew template spec](https://github.com/mamedev/build/blob/master/whatsnew/whatsnew_template.txt), as I did just [now](https://github.com/mamedev/mame/commit/0f769160d15d42d7158362f2aa2119c204664644).

paul75s: Great. Apologies for appearing to be chasing this, I know there's a huge amount of change to manage here - I'm just still getting my head around how everything works.

PR #14433, opened 2025-10-30
sega/model2: new trilinear implementation, add support for microtextures

gm-matthew:
The new trilinear implementation runs about 40% faster than the current one; most of the performance improvement comes from avoiding the slow `log2f()` function and also only fetching a second bilinear texel when required.

As well as running faster, the new implementation is a lot more accurate. For example, on the course select screen in Daytona USA, the maps morph into detail the same way they do on real hardware: https://youtu.be/R-4OZWhTWm0

The log table RAM has been hooked up and is used to calculate the mipmap level in conjunction with the per-pixel value of log2(z^2).

The dpdy parameter added to poly.h with the original trilinear implementation has been removed, as the new implementation does not require it.

### DIFF
--- a/src/mame/sega/model2.h
+++ b/src/mame/sega/model2.h
@@ -204,8 +204,6 @@ protected:
 	void geo_init(memory_region *polygon_rom);
 	u32 render_mode_r();
 	void render_mode_w(u32 data);
-	u16 lumaram_r(offs_t offset);
-	void lumaram_w(offs_t offset, u16 data, u16 mem_mask = ~0);
 	u8 lumaram_r(offs_t offset);
 	void lumaram_w(offs_t offset, u8 data);
 	u16 fbvram_bankA_r(offs_t offset);


Nitch2024:
Awesome. Great work! I was working on the same thing. You got there way faster than me.

[Edit] Those changes create some large black lines across the road noticeable in Sega Rally. After tracking down the issue it is related to the max mip level calculation. If higher than 5, it creates those black lines.

I can look into it if it helps, but if you are already looking into it happy to let you do it.

Nitch2024:
So for 256x256 max_level using (f2u((float)std::min(object.texwidth, object.texheight)) >> 23) - 128 is 7.

I attached 2 vram captures and you can see that for 256x256 there are at most 6 levels so the formula is off by one BUT the code uses the full range [0 to max_level] included, so there is an off by two basically.

Unless the number of mipmaps is available somewhere (I could not find it anywhere in the hardware doc) I think the formula should be (f2u((float)std::min(object.texwidth, object.texheight)) >> 23) - 128 - 2
 
vram0
vram1

gm-matthew:
> [Edit] Those changes create some large black lines across the road noticeable in Sega Rally. After tracking down the issue it is related to the max mip level calculation. If higher than 5, it creates those black lines.
> 
> I can look into it if it helps, but if you are already looking into it happy to let you do it.

These black lines briefly appear on real hardware as well ([source](https://youtu.be/tWdIiGuc51E?t=78)):

image

To me this shows that Model 2 can use more than six mipmaps, and chances are the AM3 developers didn't realize it. However, the black lines are more inconsistent and flicker a lot more on real hardware compared to MAME, so perhaps the mml calculation is not 100% accurate just yet.

Or perhaps it's because MAME is handling the polygons with more subpixel precision than real hardware.

Nitch2024:
You maybe right. This video captured from a real model 2 shows many of the issues I spotted today: https://www.youtube.com/watch?v=a9kprDUdxxs

So maybe your changes are pretty accurate.

image

Will try to see if I can get a higher quality capture from the arcade.

PR #14443, opened 2025-10-31
seibu/sei25x_rise1x_spr.cpp: Merge SEI25X, RISE1X sprite hardware emulation into its own device

cam900:
seibu/r2dx_v33.cpp, seibu/raiden2.cpp: Reduce duplicates

seibu/feversoc.cpp: Fix notes


src/mame/seibu/sei25x_rise1x_spr.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void sei25x_rise1x_device::device_reset()
+{
+}
+
+void sei25x_rise1x_device::alloc_sprite_bitmap()
+{
+	screen().register_screen_bitmap(m_sprite_bitmap);
+}
```
cuavas, line 63: Why does this need to be a function invoked manually?  What prevents it from being done in `device_start`?

PR #14444, opened 2025-10-31
esq5505.cpp: Add VFX-family ROM & EEPROM Cartridge support, and improve floppy support.

cbrunschen:
Second attempt, simpler and hopefully better.

I've tried to address the issues identified in the Revert commit message. Specifically, the cartridge is now represented by a single image device class, and vfxcart.h includes only diimage.h, not emu.h. All indentation should also now be tabs, not spaces, and there should be no trailing spaces (running srcclean changes nothing further).

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -362,7 +362,7 @@ TIMER_CALLBACK_MEMBER(esq5505_state::floppy_motor_on)
 		floppy_image_device *floppy = m_floppy_connector->get_device();
 		if (floppy)
 		{
-			floppy->mon_w(motor_on ? CLEAR_LINE : ASSERT_LINE);  // active low
+			floppy->mon_w(!motor_on);  // active low
 			m_panel->set_floppy_active(motor_on);
 		}
 	}
@@ -378,8 +378,7 @@ TIMER_CALLBACK_MEMBER(esq5505_state::floppy_dskchg_reset)
 void esq5505_state::update_floppy_inputs()
 {
 	// update the "Disk Ready" input
-	int state = (m_floppy_is_active && m_floppy_is_loaded) ? ASSERT_LINE : CLEAR_LINE;
-	m_duart->ip0_w(state);
+	m_duart->ip0_w(m_floppy_is_active && m_floppy_is_loaded);
 
 	// Also update the DOC IRQ in case there's a pending disk change to handle.
 	update_docirq_to_maincpu();
@@ -641,7 +640,7 @@ void esq5505_state::duart_output(uint8_t data)
 	    VFX-SD & SD-1 (32):
 	    bits 0/1/2 = analog sel
 	    bit 3 = SSEL (disk side)
-	    bit 4 = DSEL (Drive Delect and floppy Motor On)
+	    bit 4 = DSEL (Drive Select and floppy Motor On)
 	    bit 6 = ESPHALT
 	    bit 7 = SACK (?)
 	*/
@@ -671,8 +670,8 @@ void esq5505_state::duart_output(uint8_t data)
 		}
 		else
 		{
-			floppy->ss_w(((data & 8) >> 3) ^ 1); // bit 3, inverted -> floppy
-			m_floppy_is_active = (data & 16) != 0; // bit 4 is used to activate the floppy:
+			floppy->ss_w(!BIT(data, 3)); // bit 3, inverted -> floppy
+			m_floppy_is_active = BIT(data, 4); // bit 4 is used to activate the floppy:
 			if (m_floppy_is_active)
 			{
 				// immediately assert DISK SELECT (active low)


cbrunschen:
While I've tried to address the issues that were raised when the previous attempt at this was reverted, I would appreciate feedback on what I have missed, even in small things.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+	int state = loaded ? CLEAR_LINE : ASSERT_LINE;
+
+	// On VFX and later, DUART input bit 1 is 0 for cartridge present.
+	LOG("ip1 -> %d\n", state);
+	m_duart->ip1_w(state);
+}
+
+void esq5505_state::cartridge_load(ensoniq_vfx_cartridge *cart)
+{
+	(void) cart;
```
MooglyGuy: No need to explicitly void out unused function parameters, I believe any warnings that could be caused have been suppressed since forever.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+	LOG("Resetting floppy_dskchg -> %s\n", m_floppy_dskchg ? "true" : "false");
+	update_docirq_to_maincpu();
+}
+
+void esq5505_state::update_floppy_inputs()
+{
+	// update the "Disk Ready" input
+	int state = (m_floppy_is_active && m_floppy_is_loaded) ? ASSERT_LINE : CLEAR_LINE;
+#if VERBOSE
+	static int prev_state = 0;
```
MooglyGuy: I would use `LOGMASKED` for this rather than #if-ing it. #if'd-out code tends to rot. Should be plenty of examples to choose from when it comes to `LOGMASKED`, the only thing to know is that `LOG(...)` is effectively an alias for `LOGMASKED(LOG_GENERAL, ...)`, so user-defined verbose masks should start at `1U << 1`.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+	}
+	else
+	{
+		LOG("<No Floppy connector for loaded=%d>\n", loaded);
+	}
+}
+
+void esq5505_state::floppy_load(floppy_image_device *floppy)
+{
+	(void) floppy;
```
MooglyGuy: Same as above, don't need to explicitly void out unused function parameters.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+void esq5505_state::floppy_unload(floppy_image_device *floppy)
+{
+	(void) floppy;
+	floppy_loaded(false);
+}
+
+void esq5505_state::update_docirq_to_maincpu()
+{
+	bool floppy_dskchg_irq = m_floppy_is_active && m_floppy_dskchg;
+	if (floppy_dskchg_irq) LOG("docirq (m68k_irq1) due to disk change = %d\n", floppy_dskchg_irq);
```
MooglyGuy: MAME isn't super-consistent all the time about this sort of thing, but it would be slightly more preferred to put this on the following line.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+	if (floppy_dskchg_irq && m_floppy_is_loaded)
+	{
+		// The drives that Ensoniq use only _pulse_ DSKCHG for a brief time, when a disk is in the drive.
+		// schedule a reset.
+		LOG("Scheduling DSKCHG reset\n");
+		m_dskchg_reset_timer->adjust(attotime::from_nsec(500));
+	}
+	bool v = m_otis_irq || floppy_dskchg_irq;
```
MooglyGuy: `updated_irq` instead of `v`, perhaps?

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+	if (v != m_docirq)
+	{
+		LOG("docirq (m68k_irq1) -> %d\n", v);
+		m_maincpu->set_input_line(M68K_IRQ_1, v);
+		m_docirq = v;
+	}
+	else
+	{
+		// LOG("m68k irq1 -- %d\n", v);
```
MooglyGuy: Commented code can be deleted, or if you don't want it logged all the time, give it a separate bitfield and use `LOGMASKED`.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +308,160 @@ class esq5505_state : public driver_device
...
+{
+	if (irq != m_otis_irq)
+	{
+		// LOG("otis_irq -> %d\n", irq);
+		m_otis_irq = irq;
+		update_docirq_to_maincpu();
+	}
+	else
+	{
+		// LOG("otis_irq -- %d\n", irq);
```
MooglyGuy: Two more contenders for `LOGMASKED` here.

src/mame/ensoniq/vfxcart.cpp
```diff
@@ -0,0 +1,153 @@
...
+u8 ensoniq_vfx_cartridge::read(offs_t offset) {
+	m_state = state::IDLE;
+	auto v = m_storage[offset & MASK];
+	if (offset > 0x7f00) LOG("R %04x -> %02x\r\n", offset, v);
+	return v;
+}
+
+void ensoniq_vfx_cartridge::write(offs_t offset, u8 data) {
+	if (!m_is_writable) {
+		if (offset > 0x7f00) LOG("!W %04x  (%02x)\r\n", offset, data);
```
MooglyGuy: Suggest newline here and on the one above.

src/mame/ensoniq/vfxcart.cpp
```diff
@@ -0,0 +1,153 @@
...
+			break;
+		case state::CMD2:
+			if (offset == 0x5555 && data == 0xa0)
+				m_state = state::WR;
+			else
+				m_state = state::IDLE;
+			break;
+
+		case state::WR:
+			if (offset > 0x7f00) LOG("W %04x :  %02x\r\n", offset, data);
```
MooglyGuy: Newline.

src/mame/ensoniq/vfxcart.cpp
```diff
@@ -0,0 +1,153 @@
...
+	m_storage[0x7ffe] = 0x05;
+	m_storage[0x7fff] = 0x01;
+	fseek(0, SEEK_SET);
+	fwrite(&m_storage[0], m_storage.size());
+
+	// By definition, if we create a cartridge image, is kind of has to be a writable one: a completely empty
+	// read-only cartridge makes no sense!
+	// This allows users to create a cartridge image with a ".rom" or ".cart" extension, write to it _once_,
+	// in this session, until it is unloaded; and whenever in the future they load it again, it will be
+	// detected as ROM (_not_ EEPROM) by its file extension.
```
MooglyGuy, line 131: Is this behavior accurate to a real VFX (i.e., write-once cartridges)?

src/mame/ensoniq/esq5505.cpp
cbrunschen: Habit from other C++ code I work on. Removed.

src/mame/ensoniq/esq5505.cpp
cbrunschen: The #ifdef here is also guarding the static int prev_state, which is only there to track the previous state, so the logging doesn't spam non-changing updates; so it's not as straightforward as just replacing LOG with LOGMASKED.
But as this LOG was really only useful during development, I'm removing it; it's simple enough to add something similar back in the future if the need arises.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Removed.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Removed.

src/mame/ensoniq/vfxcart.cpp
cbrunschen: Done.

src/mame/ensoniq/vfxcart.cpp
cbrunschen: Done.

src/mame/ensoniq/vfxcart.cpp
cbrunschen: No, I have never encountered a write-once cartridge.

The idea is simply one of usability in a simulated environment: this lets you create a read-only cartridge (.cart or rom), but as an empty read-only cartridge makes little sense, it allows writing to it once; otherwise you'll have to first create it as a writable cartridge, then eject it and rename it (from .eeprom or .sc32 to .cart or .rom). This way, you can just create it, with the intention that it will be a read-only cartridge once it has been written to; and without running the risk of forgetting to rename the cartridge, loading it and accidentally overwriting.

cbrunschen: I changed this so that the vfxcart device will only create `.eeprom` or `.sc32` images, and will return a suitable error otherwise.

However, I note that MAME still creates the files, leaving them in place and empty.

MooglyGuy: So the user still has the option to create any given extension supported by the image device - .rom, .cart, .eeprom or .sc32 - but only .eeprom or .sc32 images are writable?

I'm admittedly unsure if there's a good workaround for this.

IMO, there isn't so much of a need for any differentiation in function like that. Support creation and writing to any of the above. Items that exist in a softlist have their underlying data inherently static, with only diffs against the fundamental data being stored, and in a separate file at that. Even if a user is able to write to a ROM-only cartridge, that's about the best one can do in the current implementation.

I'd like to hear from @cuavas or whoever rejected the initial implementation, because this seems like the exact sort of situation that calls for a cartslot-type bus device which supports two different `device_cartrom_slot_interfaces`, one that bounces out to an image device, and one that only does straight loading of ROM data. Per @cbrunschen, the original PR was rejected for being too complicated, but I'm not sure I would agree, given the two distinct types of pluggable cartridges.

As far as the UX of the VFX goes, users could buy blank, writable cartridges to make their own patches, or they could be non-writable cartridges preloaded with patches. It seems like a perfect fit for the intention behind the original PR, whereas going with a "pure" imagedev just results in annoying hoop-jumping on @cbrunschen's part.

cbrunschen: Users could buy either pre-made ROM cartridges (Such as Ensoniq's fairly ubiquitous VPC-100 series), or EEPROM cartridges such as Ensoniq's own StorCart-32. 
These were made on the same PCB, just with either a programmed 27256 series EPROM, or a 28256 EEPROM, and with two out of four zero-Ohm resistors placed to make the correct connections.

To the best of my knowledge, the actual contents of each is identical. The main difference is simply that writes to an EEPROM cartridge succeed, but attempted writes to a ROM cartridge fail.

To me, this is not really much different from a floppy image that may or may not be write protected. 

I had originally written a Slot device and an Image device, which could load either ROM or EEPROM cartridges; and I understand that this was considered overly complex, since the two could be combined into a single device. And I don't disagree; I wrote it the way I did largely because that seemed to be a patter among other slot and image devices.

That image device that could load both read-only and writable images, differentiated by their filetype, fairly arbitrarily chose as '.rom' or '.cart' for ROM cartridges, '.eeprom' or '.sc32' (for StorCart-32) for EEPROM ones. And in order to make things easier for the user to create a cartridge that is intended to be write-protected, it also allowed creating not only a writable .eeprom or .sc32 image, but also a .rom or .cart image, which would be loaded as effectively "once-writable"; with the idea being that the user would write to it once (copying over the entire contents of the internal RAM sounds and presets), and after it was unloaded, it would then be effectively read-only the next time it was loaded.

In the first round of this second PR I eliminated the separate slot device but left the image device mostly the same, and able to load either ROM (read-only) or EEPROM (writable) images, as well as create them as above.

Per comments earlier, I then changed the image device so that it will only create EEPROM cartridges that can actually be written to, and returns errors for attempts to create ROM cartridge images (instead of allowing them to be written once), which makes sense. However, MAME always creates the files, and returning an error from `call_create()` only prevents that file from being written to and loaded.

I don't really think this is a situation where there are two significantly different kinds of devices; After all, floppy disk images are not loaded into different image device classes, depending on their write protect status.. So consolidating both the slot and cartridge into a single image device class makes sense to me.

At the same time, I think that the pre-written cartridges and the EEPROM ones do deserve differentiation beyond whether the user chooses to load a particular image as read-only - which is something you can do with a floppy image - or, say, removing write permissions on the image file: a different file type (.rom/.cart vs .eeprom/.sc32) seems a good indicator of that, for me.

I'm not entirely sure what the best solution here is. Perhaps the code in `diimage.cpp` around line 994 should remove the file if the call to `call_create()` returns an error?


[edited to add]

Or perhaps rather, [this conditional](https://github.com/mamedev/mame/blob/master/src/emu/diimage.cpp#L869) should also remove any created file?

cbrunschen: Also, just to mention some theoretical added complexity that I'm not even really going to bother with: there are aftermarket cartridges that use larger EEPROM chips, and which use an external switch to provide values for the extra address lines - thus resulting in "multi-cartridges", like [this SC-2048](https://www.synthsquare.com/product/sc-2048) that is available to buy for the VFX family. One of these offers the advantage over switching between different cartridges, that you don't wear out the PCB edge connectors or the receptacle in the keyboard, while still having quick access to effectively 64 different cartridges at the flick of a switch.
This might well be better done as a separate image device, though the current device could fairly easily be extended to simply open any file that is a straight multiple of 32 kBytes in size, and allow some way for the user to switch between these on the fly.

MooglyGuy: We're getting into territory that I'm not even familiar enough with, and you're asking questions that I can't really give a clear-cut answer to. I'd rather someone more associated with steering MAME's trajectory, like RB, OG, or Vas, to weigh in on the matter.

But at the very end of your follow-up comment, you're describing yet a third type of cartridge, one that could pretty trivially be handled by a slot device, even including device-specific DIPs to select which page is currently visible. If anything, that's an argument in favor of going the slot-device route.

I agree that, fundamentally, if image creation fails for reasons forced by the image device, yet the file for the image was still successfully created, the resulting file should be deleted so as to not crap up the user's hard drive. Beyond that, I genuinely don't have any proverbial dog in this fight, other than to figure out how to get these changes into the driver in the least-unappealing way possible.

You *could* override `is_creatable` to prevent image creation, since `set_image_filename` is called prior to `determine_open_plan` is called.

...But to me, that's missing the forest for the trees.

If it was just a matter of cartridges that are identical component-wise but not electrically identical (you can't say they're identical; different resistors in different places is still different!), the above exactly how I would do it.

But now you're bringing up that there are aftermarket cartridges.

With a slot device and implementations for individual cartridge types, adding that third type would be easy. If this all only exists in the realm of diimage, that instantly makes life a far larger pain in the ass. And I don't like it when people who want to contribute decent changes to MAME are having a pain in the ass.

To me, it's an argument in favor of going right back with the originally-intended implementation of making the VFX cartridge slot an actual slot device, having and having three cartslot-device-type implementations: One that is a simple ROM loader, one that interfaces with an imagedev class, and one that interfaces with the same imagedev class, but also has input ports so that the user can select which page is presented on the bus.

That third option pushes me strongly in the direction of wanting the old, "too complex," implementation. But I want to hear RB, OG, or Vas weigh in, because the last thing I need in my life is getting my expectations rug-pulled by colleagues.

cbrunschen: For reference regarding the (EP)ROM vs EEPROM cartridges, there's for example [this blog post](https://web.archive.org/web/20210210221123/https://members.pro-ns.net/steve/vfxcart.html) about an adapter where you literally drop a chip into a ZIF socket, which refers to [this web page](https://web.archive.org/web/20210210221123/https://members.pro-ns.net/steve/vfxcart.html) where someone converted a ROM to an EEPROM cartridge.

cbrunschen: Gathering some thoughts:

- I largely draw parallels between these cartridges and their images, and floppies and their images.
- Like floppies, cartridges come in both writable and read-only varieties, but those are not considered different enough to require different image classes; a single image class handles both. 
- The aftermarket multi-cartridges I see as analogous to floppy-drive emulators: they allow you to choose between different storage media (real or virtual) to attach to the system, and yes, people do connect them to the original hardware, but they are also outside the remit of what emulating the system needs to do.
- The cartridge equivalent to the floppy drive is the cartridge slot, which literally just passes through reads and writes, so does not need a separate device class, as was pointed out on the original PR.
- This leads me to the design here: a single cartridge class that represents both the slot and the cartridge in it as loaded from an image, which can be either a writable or a read-only one.
- My little shortcut to allow users to create a write-once image is just that, a shortcut; I consider it a nice-to-have but not important. Users can always create a writable image, and then rename it to make it clear it should be read-only.
- I had clearly misunderstood the `is_{read,write,creat}able` methods; updating them to look at the currently set image `filetype()` lets us avoid trying to create read-only images in the first place, and indeed, creating or loading images of any other than the supported filetypes.

cbrunschen: Some further investigation leads me to reassess some of the above. As ever, where I'm wrong or have misunderstood something, please point this out so I can learn better.

`device_image_int::is_creatable()` seems to be called in two different contexts:
- from [within `device_image_interface::determine_open_plan()`](https://github.com/mamedev/mame/blob/master/src/emu/diimage.cpp#L682), when attempting to open or create an image file; at this point, the image file name has been set
- from [within `menu_control_device_image::menu_activated()`]e(https://github.com/mamedev/mame/blob/master/src/frontend/mame/ui/imgcntrl.cpp#L256), in order to determine whether or not to show a **[create]** menu item in the file manager; at this point, the image device either has the name and file type of whatever is currently loaded into the image device, or NULL if there's nothing loaded.

That means that in order to allow creation of images, I need to be able to return `true` from `is_creatable()` whenever it's called from `menu_control_device_image::menu_activated()`.

However, there doesn't seem to be a way to reliably detect that. So the best thing I could do is always return `true`.

That also means I can't check the image's `filetype()` when `is_creatable()` is called from  `device_image_interface::determine_open_plan()` to detect whether it is possible to create an image with a particular filename - i.e., I can't reject creation of `.rom` or `.cart` images here.

Incidentally, when I tried returning `false` from `is_creatable()` when the filetype was not one of the eeprom cartridge ones, MAME would show an error message "Error loading media image: no such file or directory", which was not the most informative.

So I can still reject creation of `.rom` or `.cart` images in `ensoniq_vfx_cartridge::call_create()`, but that then leaves MAME to first create the file, because `is_creatable()` returned true and allowed MAME to create the file in the first place; but at least here, MAME shows the error message that `call_create()` returned.

It seems that:
- `is_creatable()` is a bit overloaded to mean two different things;
- when `determine_open_plan()` does not find a file that it can successfully open, it results in a very generic error message

Interestingly enough, the latter also applies in another case. MAME's file manager lets me choose any file I want to load, even though I've specified the filetypes that I can handle in `ensoniq_vfx_cartridge::file_extensions()`. 

If I  try to return `false` from `is_readable()` for any file that has the wrong filetype in order to prevent loading non-images, then that again results in a message of "Error loading media image: no such file or directory" which is confusing, because I literally just chose an existing file. 

If instead I check the filetype in `ensoniq_vfx_cartridge::call_load()` then MAME will show the error message returned from `call_load()`.

So in order to give a user experience with the most informative error messages, and generally following [the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I can't really `return` false from any of the `is_{read,write,creat}able()` methods. 

- I have to return `true` from `is_readable()` to prevent an inaccurate "no such file or directory" error when trying to open an existing file that happens to have the wrong extension.

- I have to return `true` from `is_creatable()` to make sure that the [create] menu item always appears. 

- Since I can't return `false` from `is_creatable()`, I have to do something reasonable for any arbitrary file that I am asked to create, because MAME has already created the file by the time I can do anything else here. 

- So even if the file I'm asked to create is technically intended to be read-only - or even, of entirely the wrong filetype - I kind of have to treat it as writable at least once, at least on this occasion, or the user will have a bad experience.

What am I missing? What should I be doing differently?

cbrunschen:
Upon some further reflection, I've restored the "write-once" behaviour: it's useful, and it avoids the situation where the code has to back out of creating an image, leaving an empty file behind.
 
More than that, it is also fairly analogous to some functionality relating to floppy disk images: those can be mounted such that the original floppy image is untouched and changes are written to a separate file, for example. Much like write-once cartridges, such copy-on-write floppy disks don't exist in real life. But in an emulation context, both protect the image files the user is working with from being accidentally written to when they shouldn't, and thus prevent data loss.

src/mame/ensoniq/esq5505.cpp
MooglyGuy: It really is as simple as replacing `LOG` with `LOGMASKED`, however - `LOGMASKED` with a disabled mask bit will still count as reachable code to the compiler, so you won't get any warnings about unused variables.

MooglyGuy:
I don't have a strong preference here either way, but it would be good to be able to emulate those aftermarket cartridges. This still points me towards believing a slot device is the way to go. I'll wait for RB, OG, or Vas to weigh in. Not the end of the world if this waits a bit, hopefully.

src/mame/ensoniq/esq5505.cpp
cbrunschen: it's not about warnings but about having and updating a variable (`prev_state`) that is only there to limit logging to when the state actually changes. Neither the variable, nor the code updating it, are needed other than when logging is enabled.

Someone who reads the code can so much more clearly see that intent if the declaration of the variable and the code to updating it as well as the logging that uses it are all together inside a `#ifdef` guard. 

An intelligent compiler would hopefully see that, when `VERBOSE` is `0`, the variable is only ever written to and never read so that it can effectively be elided, but code is written not just for the compiler but also for programmers in the future who are trying to understand what is going on.

cbrunschen:
For hardware keyboards, the aftermarket cartridge multi-cartridge offers two main advantages: 
- it offers a writable cartridge _at all_, since the original EEPROM cartridges are rare and hard to come by
- it allows access to switching between different "virtual" cartridges, without needing to swap between physical ones, thus reducing the physical wear-and-tear on the cartridge slot and cartridge connectors

Neither of those is important in an emulation context:
- acquiring a new cartridge is as easy as using the `[create]` menu item in the File Manager
- there are no physical connectors, so there is no wear-and-tear

So I don't really foresee much need to emulate those; YAGNI seems to apply. 

If someone wishes to emulate these multi-cartridges in the future, there are other issues to resolve as well: Any multi-cartridge would also require some UI/UX for how to switch between the different "banks" of 32kB each, since that is really not part of the UI of the keyboard itself, and that UI also needs to be connected to the multi-cartridge. 

In that context, separating the current class into separate slot & cartridge classes at that time would be comparatively trivial (since the slot device literally just passes through read and write requests). 

Indeed an alternative design choice could be to extend the capabilities of the current class, so that it can handle some multiple of 32kiB in file size, and not requiring a split into separate slot & cartridge classes at all.

So on the balance I think that there is much more value in getting this functionality in place than debating a possible future extension for which there is little need in an emulation context, which may not even need the split into two separate classes, and if it does, then that split is easy to make at that time.

cbrunschen:
@rb6502  @galibert @cuavas , @MooglyGuy requests "for RB, OG, or Vas to weigh in".

MooglyGuy:
I pinged them earlier in the thread, they'll still have an e-mail about it. They'll respond when they have the time.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -258,6 +261,7 @@ class esq5505_state : public driver_device
 	required_device<esq_5505_5510_pump_device> m_pump;
 	optional_device<wd1772_device> m_fdc;
 	optional_device<floppy_connector> m_floppy_connector;
+	optional_device<ensoniq_vfx_cartridge> m_cart;
```
rb6502, line 264: A thought for future work: having more than one optional_device often means it would be cleaner to have a base class with nothing optional and then various inherited child classes that extend that in various ways.  Back when I wrote esq5505 originally I wasn't in that mindset and it does show.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +309,143 @@ class esq5505_state : public driver_device
...
+
+void esq5505_state::floppy_drives(device_slot_interface &device)
+{
+	device.option_add_internal("35dd", FLOPPY_35_DD);
+}
+
 void esq5505_state::floppy_formats(format_registration &fr)
 {
 	fr.add_mfm_containers();
 	fr.add(FLOPPY_ESQIMG_FORMAT);
```
rb6502, line 354: It's probably not a bad idea to add the HFE format too - that's become kind of the standard way to distribute synth floppies with nonstandard physical formats, since a lot of people still using real hardware have Gotek or similar drive replacements that use that format.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +309,143 @@ class esq5505_state : public driver_device
...
 
+TIMER_CALLBACK_MEMBER(esq5505_state::floppy_motor_on)
+{
+	bool motor_on = param;
+	if (m_floppy_connector)
+	{
+		floppy_image_device *floppy = m_floppy_connector->get_device();
+		if (floppy)
+		{
+			floppy->mon_w(motor_on ? CLEAR_LINE : ASSERT_LINE);  // active low
```
rb6502: !motor_on should work too, and is faster to read/interpret.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +309,143 @@ class esq5505_state : public driver_device
...
+{
+	m_floppy_dskchg = !m_floppy_is_loaded;
+	LOG("Resetting floppy_dskchg -> %s\n", m_floppy_dskchg ? "true" : "false");
+	update_docirq_to_maincpu();
+}
+
+void esq5505_state::update_floppy_inputs()
+{
+	// update the "Disk Ready" input
+	int state = (m_floppy_is_active && m_floppy_is_loaded) ? ASSERT_LINE : CLEAR_LINE;
```
rb6502: You can just collapse this to `m_duart->ip0_w(m_floppy_is_active && m_floppy_is_loaded);`

src/mame/ensoniq/esq5505.cpp
```diff
@@ -294,15 +309,143 @@ class esq5505_state : public driver_device
...
+}
+
+void esq5505_state::update_docirq_to_maincpu()
+{
+	bool floppy_dskchg_irq = m_floppy_is_active && m_floppy_dskchg;
+	if (floppy_dskchg_irq)
+		LOG("docirq (m68k_irq1) due to disk change = %d\n", floppy_dskchg_irq);
+	if (floppy_dskchg_irq && m_floppy_is_loaded)
+	{
+		// The drives that Ensoniq use only _pulse_ DSKCHG for a brief time, when a disk is in the drive.
```
rb6502, line 425: Is this 100% verified?  Most drives will raise DSKCHG on eject and then lower it the next time a seek happens with a disk in the drive, and MAME emulates that by default with the dskchg_r() method on the floppy image device.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -469,7 +641,7 @@ void esq5505_state::duart_output(uint8_t data)
 	    VFX-SD & SD-1 (32):
 	    bits 0/1/2 = analog sel
 	    bit 3 = SSEL (disk side)
-	    bit 4 = DSEL (drive select?)
+	    bit 4 = DSEL (Drive Delect and floppy Motor On)
```
rb6502: Delect? :-)

src/mame/ensoniq/esq5505.cpp
```diff
@@ -499,7 +671,23 @@ void esq5505_state::duart_output(uint8_t data)
 		}
 		else
 		{
-			floppy->ss_w(((data & 8)>>3)^1);
+			floppy->ss_w(((data & 8) >> 3) ^ 1); // bit 3, inverted -> floppy
```
rb6502: `floppy->ss_w(!BIT(data, 3));` would do the same thing much cleaner.

src/mame/ensoniq/esq5505.cpp
```diff
@@ -499,7 +671,23 @@ void esq5505_state::duart_output(uint8_t data)
...
+			m_floppy_is_active = (data & 16) != 0; // bit 4 is used to activate the floppy:
```
rb6502: Similarly, `m_floppy_is_active = BIT(data, 4);`

src/mame/ensoniq/esq5505.cpp
cbrunschen: It might be worth looking at this for a refactor more generally. The EPS Classic, for example, doesn't really belong here at all: it has no es5510 ESP, and the sound generator is the es5504 "DOC II", not the es5505 "OTIS".

src/mame/ensoniq/esq5505.cpp
cbrunschen: Probably a good idea, though the FlashFloppy firmware for Gotek drives explicitly supports the Ensoniq 800 kB format for [Ensoniq hosts](https://github.com/keirf/flashfloppy/wiki/Host-Platforms#ensoniq), so for these keyboards in particular, HFE may be less important.  Either way I'd prefer to read up more on that before I make such a change, separately from this PR.

src/mame/ensoniq/esq5505.cpp
cbrunschen: I think so; please see [this comment](https://github.com/mamedev/mame/pull/14331#issuecomment-3402898884) on the previous attempt at these changes:

> > Are you sure the dskchg is timer-based and not the usual "clear on first track change with a floppy in" that is implemented in floppy_image_device?
> 
> I an absolutely sure.
> 
> I first found it documented for the EPS in Keir Fraser's FlashFloppy firmware, when people tried to connect it to an Ensoniq EPS and it [would not recognize disk changes](https://github.com/keirf/FlashFloppy/issues/240).
> 
> But with this being so strange, I measured it with my cheap Saleae Logic clone logic analyzer on an SD-1/32. After a disk change - so, removed the previous disk, inserted a new disk, then went to try to read something from the disk: pressed the "Storage" button, then selected "Disk" by pressing the middle button below the screen, and then "Load" again with the middle button below the screen) - the keyboard signals (drives low) the DSEL and MO signals (enabling the drive and starting the motor), and the drive responds with a brief pulse: in my measurement with a cheap logic analyzer sampling at 4 MHz, all of two (2) samples were low, so a pulse of 500 nanoseconds and certainly no more than 1000 nanoseconds to 1 microsecond.
> 
> I'm including the sigrok trace recorded with [PulseView](https://sigrok.org/wiki/PulseView):.
> 
> [Load from floppy after new floppy inserted.zip](https://github.com/user-attachments/files/22909431/Load.from.floppy.after.new.floppy.inserted.zip)
> 
> Look at the 335088 mycrosecond timestamp: the MO and DSEL signals both go low, and at the same time, DSKCHG goes low, but only for 2 samples, and with no other signal changing during that very short window.
> 
> The DSKCHG signal is actually routed, via a gate and a transistor, into the same IRQ line on the CPU itself that is also used by the ES5505 OTIS sound generator chip. It so happens that on the SD-1, if that interrupt is taken (and I presume after the keyboard has checked the ES5505 interrupt status), eventually the SD-1 will assert (drive low) the DSEL (and MO) lines to the floppy drive and issue a STEP command, which will inturn reset DSKCHG even on a standard drive; but that of course might take longer than it does when the drive itself only briefly pulses the signal.
> 
> And while this would technically work on the SD-1, it would not work on the EPS, as documented by FlashFloppy users - where they added a configuration setting precisely to reset DSKCHG after a short time. So including this code is useful, possibly even necessary, to get the EPS and EPS16+ to work in the future.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: "Select" of course. Fixed.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: Done.

src/mame/ensoniq/esq5505.cpp
cbrunschen: For reading Ensoniq floppies, the GreaseWeazle software also explicitly supports [Ensoniq's floppy formats](https://github.com/keirf/greaseweazle/blob/master/src/greaseweazle/data/diskdefs_ensoniq.cfg), so they can be directly read to .img files for use with a FlashFloppy Gotek drive, or here in MAME.

src/mame/ensoniq/esq5505.cpp
rb6502: Yeah, for sure.

src/mame/ensoniq/esq5505.cpp
rb6502: Huh.  That's kind of wild.

cbrunschen: Yeah, the floppy signal routing in general and the DSKCHG pulsing behaviour in particular were interesting to trace and track down. Being able to measure on real hardware and seeing the behaviour really helped confirm that was what was actually happening, even after reading about it in the FlashFloppy firmware discussion and documentation.

cbrunschen:
Thank you!

PR #14445, opened 2025-10-31
Polar Star 2

stonedDiscord:
Shoken Polar Star 2 [stonedDiscord]

### DIFF
--- a/src/mame/skeleton/shoken_md06.cpp
+++ b/src/mame/skeleton/shoken_md06.cpp
@@ -72,7 +72,6 @@ void shoken_md06_state::program_map(address_map &map)
 void shoken_md06_state::io_map(address_map &map)
 {
 	map.global_mask(0xff);
-	map(0x3b, 0x3b).rw("oki", FUNC(okim9810_device::read), FUNC(okim9810_device::write));
 	map(0x40, 0x40).nopw();
 	map(0x80, 0x80).noprw();
 }
@@ -80,6 +79,7 @@ void shoken_md06_state::io_map(address_map &map)
 void shoken_md06_state::sound_io_map(address_map &map)
 {
 	map.global_mask(0xff);
+	map(0x3b, 0x3b).rw("oki", FUNC(okim9810_device::read), FUNC(okim9810_device::write));
 	map(0x40, 0x40).nopw();
 	map(0x80, 0x80).noprw();
 }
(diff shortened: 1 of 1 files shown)


stonedDiscord:
I am still a bit irritated by the fact that the AT27C010 without a window is labelled as Polar Star 2 but when I look at the Hex editor is says 2010 and Puerto Prinsesa.
Yet the label and the ROM match with that MK04 string. Really weird. Maybe it was upgraded at some point?

src/mame/skeleton/shoken_md06.cpp
```diff
@@ -42,27 +50,39 @@ class shoken_md06_state : public driver_device
...
+	map(0x09000, 0x0cfff).ram();
 	map(0xe0000, 0xe3fff).ram();
+	map(0xfcc00, 0xfffff).ram();
 }
 
 void shoken_md06_state::io_map(address_map &map)
 {
 	map.global_mask(0xff);
-	map.unmap_value_high();
+	map(0x3b, 0x3b).rw("oki", FUNC(okim9810_device::read), FUNC(okim9810_device::write));
```
Osso13: you put the Oki in the wrong io_map (you created sound_io_map below)

src/mame/skeleton/shoken_md06.cpp
```diff
@@ -138,20 +180,34 @@ void shoken_md06_state::petitlot(machine_config &config)
...
-	// TODO: sound related ICs aren't present?
+	okim9810_device &oki(OKIM9810(config, "oki", 16_MHz_XTAL / 12));
+	oki.add_route(ALL_OUTPUTS, "mono", 1.0); //divider guessed
 }
 
-
 ROM_START( petitlot )
 	ROM_REGION( 0x10000, "maincpu", 0 )
-	ROM_LOAD( "md06_ver4_1.ic6", 0x00000, 0x10000, CRC(ebc81f10) SHA1(28ac52aeadfbf792da95c01b16fb88f7a5eb1d4e) ) // 1xxxxxxxxxxxxxxx = 0xFF
+	ROM_LOAD( "md06_ver4_1.ic6", 0x00000, 0x10000, CRC(ebc81f10) SHA1(28ac52aeadfbf792da95c01b16fb88f7a5eb1d4e) ) // 1xxxxxxxxxxxxxxx = 0xff
```
Osso13, line 199: this is romcmp's output in a comment, it isn't subject to MAME's preference for lowercase hex.

Osso13 (commented):
Thanks, if you put the oki in the right io_map this should be ready to go.
Have you verified Petit Lot actually has no sound hardware, since you put that flag?

stonedDiscord:
No I could not verify it and was going off the description in the header that the Oki is unpopulated on Petit Lot. Didn't think about how the Z80 might be doing sound.

PR #14449, opened 2025-11-01
[CD-i] Fix XA audio peaking

Vincent-Halver:
This fixes an issue with audio peaking in Asterix's opening cutscene. It performs audio decoding in 32 bit and then clips to 16 instead of allowing overflow.

Before: Watch the opening cutscene. Notice that there is noticable audio popping.
After: No audio popping.

### DIFF
--- a/src/mame/igs/goldstar.cpp
+++ b/src/mame/igs/goldstar.cpp
@@ -24939,7 +24939,18 @@ ROM_END
   Tried any input on the real hardware without luck.
 
 */
-ROM_START( scmaster )
+ROM_START( scmaster )  // ver 1.1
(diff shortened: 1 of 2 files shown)


MooglyGuy:
Looks fine, but please squash your commits, and give the squashed commit a message more in the vein of typical whatsnew entries, something along the lines of:

-cdicdic: Fixed audio peaking by decoding in 32-bit and clipping to 16 later. [Vincent Halver]

-cdislavehle: Simplified mouse position-update logic. [Vincent Halver]

Vincent-Halver:
https://github.com/mamedev/mame/pull/14470

PR #14453, opened 2025-11-02
*/vrender0.cpp and related drivers/devices: Cleanup and Updates

cam900:
- Fix naming (Use snake_case as default naming convention for consistency)
- Fix spacings
- Fix variable and function order in header files
- Reduce defines

cpu/se3208/se3208.cpp:
- Make interrupt pin define as enum
- Use logmacro.h for logging feature
- Reduce duplicates

cpu/se3208/se3208dis.cpp:
- Fix naming for reduce confusion

machine/vrender0.cpp:
- Use memory_share_creator for texture/framebuffer RAM
- Add notes for DMA increment

sound/vrender0.cpp:
- Update per-channel cache when texture memory pointer is changed
- Use refefence instead pointer as possible

video/vrender0.cpp:
- Use template and function arrays for textured draw function
- Use device_memory_interface for external memory space
- Split shade and alpha function
- Minor fixes in single color fill function
- Use refefence instead pointer as possible

misc/crospuzl.cpp:
- Fix filename in comment (is_busy() is now moved into machine/nandflash.h)

misc/crystal.cpp:
- Add notes for flash memory emulations
- Remove unused finder
- Reduce literal tag usages

misc/ddz.cpp:
- Reduce literal tag usages

misc/menghong.cpp:
- Add notes for flash memory emulations

misc/psattack.cpp:
- Fix logging
- Reduce literal tag usages

misc/trivrus.cpp:
- Add notes for flash memory emulations

### DIFF
--- a/src/devices/video/vrender0.cpp
+++ b/src/devices/video/vrender0.cpp
@@ -255,17 +255,17 @@ device_memory_interface::space_config_vector vr0video_device::memory_space_confi
  IMPLEMENTATION
  *****************************************************************************/
 
-static inline u32 RGB32(u8 r, u8 g, u8 b) { return (r << 16) | (g << 8) | (b << 0); }
-static inline u16 RGB16(u8 r, u8 g, u8 b) { return ((r & 0xf8) << 8) | ((g & 0xfc) << 3) | ((b & 0xf8) >> 3); }
+constexpr u32 RGB32(u8 r, u8 g, u8 b) { return (r << 16) | (g << 8) | (b << 0); }
+constexpr u16 RGB16(u8 r, u8 g, u8 b) { return ((r & 0xf8) << 8) | ((g & 0xfc) << 3) | ((b & 0xf8) >> 3); }
 
-static inline u16 RGB32TO16(u32 rgb)
+constexpr u16 RGB32TO16(u32 rgb)
 {
 	return (((rgb >> (16 + 3)) & 0x1f) << 11) | (((rgb >> (8 + 2)) & 0x3f) << 5) | (((rgb >> (3)) & 0x1f) << 0);
 }
 
-static inline u8 EXTRACTR8(u16 src) { return ((src >> 11) << 3) & 0xff; }
-static inline u8 EXTRACTG8(u16 src) { return ((src >>  5) << 2) & 0xff; }
-static inline u8 EXTRACTB8(u16 src) { return ((src >>  0) << 3) & 0xff; }
+constexpr u8 EXTRACTR8(u16 src) { return ((src >> 11) << 3) & 0xff; }
+constexpr u8 EXTRACTG8(u16 src) { return ((src >>  5) << 2) & 0xff; }
+constexpr u8 EXTRACTB8(u16 src) { return ((src >>  0) << 3) & 0xff; }
 
 static inline u16 do_shade(u16 src, u32 shade)
 {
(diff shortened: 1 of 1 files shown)


src/lib/util/coretmpl.h
```diff
@@ -487,14 +487,17 @@ class fifo : protected std::array<T, N>
 		}
 	}
 
-	T const &dequeue()
+	T const &dequeue(bool allow_side_effects = true)
```
angelosa: Beyond going out of scope in _this PR_ I'm not convinced it's the right approach. Just suppress side effects in vr0uart instead of a pointless extra branch for everything using the FIFO class.

src/mame/misc/crospuzl.cpp
```diff
@@ -56,140 +56,142 @@ class crospuzl_state : public driver_device
...
-	uint32_t PIOlddr_r();
-	void PIOlddr_w(offs_t offset, uint32_t data, uint32_t mem_mask = ~0);
-	uint32_t PIOldat_r();
-	void PIOldat_w(offs_t offset, uint32_t data, uint32_t mem_mask = ~0);
-	uint32_t PIOedat_r();
+	u32 pio_lddr_r();
+	void pio_lddr_w(offs_t offset, u32 data, u32 mem_mask = ~0);
+	u32 pio_ldat_r();
+	void pio_ldat_w(offs_t offset, u32 data, u32 mem_mask = ~0);
+	u32 pio_edat_r();
```
angelosa: fwiw the capitalization was made this way for datasheet matching i.e. PIOLDAT / PIOEDAT.

src/devices/cpu/se3208/se3208dis.h
```diff
@@ -16,88 +16,88 @@ class se3208_disassembler : public util::disasm_interface
 	virtual offs_t disassemble(std::ostream &stream, offs_t pc, const data_buffer &opcodes, const data_buffer &params) override;
 
 private:
-	typedef uint32_t (se3208_disassembler::*_OP)(uint16_t Opcode, std::ostream &stream);
+	typedef u32 (se3208_disassembler::*_OP)(u16 opcode, std::ostream &stream);
```
cuavas, line 19: Names beginning with an underscore followed by an uppercase letter are reserved – `_OP` should be renamed.

src/devices/sound/vrender0.cpp
```diff
@@ -444,144 +452,144 @@ void vr0sound_device::channel_t::write(offs_t offset, u16 data, u16 mem_mask)
...
 			}
 			else
 			{
-				if (channel->Modes & MODE_8BIT)   //8bit
+				if (channel.modes & MODE_8BIT)   //8bit
 				{
-					sample = channel->Cache->read_byte(channel->CurSAddr >> 9);
-					sample = (s16)(((s8)(sample & 0xff)) << 8);
+					sample = channel.cache->read_byte(channel.cur_saddr >> 9);
+					sample = s16(s8(sample & 0xff) << 8);
```
cuavas, line 535: The second line here is pretty sus – I think it would be fine without the inner cast to `s8`:
```c++
					sample = s16((sample & 0xff) << 8);
```

src/devices/video/vrender0.cpp
```diff
@@ -163,119 +224,88 @@ void vr0video_device::device_start()
...
 {
 	return (((rgb >> (16 + 3)) & 0x1f) << 11) | (((rgb >> (8 + 2)) & 0x3f) << 5) | (((rgb >> (3)) & 0x1f) << 0);
 }
 
-#define EXTRACTR8(Src)  (((Src >> 11) << 3) & 0xff)
-#define EXTRACTG8(Src)  (((Src >>  5) << 2) & 0xff)
-#define EXTRACTB8(Src)  (((Src >>  0) << 3) & 0xff)
+constexpr u8 EXTRACTR8(u16 src) { return ((src >> 11) << 3) & 0xff; }
+constexpr u8 EXTRACTG8(u16 src) { return ((src >>  5) << 2) & 0xff; }
+constexpr u8 EXTRACTB8(u16 src) { return ((src >>  0) << 3) & 0xff; }
```
cuavas, line 268: These end up only setting five or six out of eight bits – should they use the `pal5bit` and `pal6bit` helpers that splat right instead?

PR #14454, opened 2025-11-02
[Access Virus] CPanel buttons for Virus A & B

felipesanches:


felipesanches:
This can be tested by holding the "S" key (corresponding to the "Transpose +" button) while booting. Right after the splash screen you'll get a key test mode. All button names are shown on the display.

Similarly, other buttons lead to additional power-on self-test / diagnostics routines.

PR #14455, opened 2025-11-02
esqpanel.cpp: Improve the Panel, and add three more views, including a working keyboard.

cbrunschen:
The panel view is improved after re-measuring on a real keyboard. Colors have been changed a bit in order to improve contrast:

Screenshot From 2025-10-30 15-24-43

Three new views have been added:

1. a Full view of the keyboard, including the controls to the left of the keyboard, specifically the Pitch-Bend and Modulation Wheels and the Patch Select buttons, as well as a representation of the floppy drive and the little "storage compartment" above it that is a bit of a characteristic shape.

Screenshot From 2025-10-30 15-23-58

2. a "Compact" view that includes all the controls and a 37-key (3-octave) keyboard that may be both legible and playable at the same time, on a reasonably-sized touch screen:

Screenshot From 2025-10-30 15-25-25

3. a "Tablet" view of the front-panel controls, with the display and buttons stacked vertically instead of stretched horizontally, as well as the "Patch Select" buttons. This view, together with a MIDI keyboard controller, this should give users a reasonably workable facsimile of the real thing: Keys, Pitch Bend and Modulation wheels on their MIDI keyboard controller; all the keyboard-specific controls and the Ensoniq-specific Patch Select buttons on the "Tablet" view:

Screenshot From 2025-10-30 15-25-56

The Full and Compact views also include a keyboard: the full 61 keys and a reduced 37 keys respectively.:

[Here's a video](https://www.dflund.se/~cb/mame/vfx_layouts.mp4)

These work, tracking pointer events, and generate not just key-down events with velocity, but also per-key "pressure" or "aftertouch". Velocity is simulated by position of the click/touch; Pressure by movement of the pointer on the key while down. Both are also animated: Velocity is shown as a colour between blue (1) and green (127), Pressure between yellow (1) and red (127):

This is implemented using analog i/o ports, one for each key, setting the value to pass the key's velocity and pressure. I wasn't able to find a specific type of analog i/o port so I am currently using `IPT_PADDLE`, which is probably not ideal, but it does work. I'd be delighted to use a more appropriate type, if there is one (or if I can relatively easily add one in a way that ).

### DIFF
--- a/src/mame/ensoniq/esqpanel.cpp
+++ b/src/mame/ensoniq/esqpanel.cpp
@@ -10,9 +10,10 @@
 #include "main.h"
 
 #include "esq2by40_vfx.lh"
-#include "sd1.lh"
 #include "vfx.lh"
 #include "vfxsd.lh"
+#include "sd1.lh"
+#include "sd132.lh"
 
 #include <algorithm>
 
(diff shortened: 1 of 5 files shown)


MooglyGuy:
@galibert This seems fine to me, if I hear no objections in the next 24-48 hours, I'll merge it.

cbrunschen:
Thank you!

PR #14459, opened 2025-11-02
apple: fix C06x softswitches

arekkusu42:
This PR supercedes #14456 and fixes the behavior of C060 and C063 softswitches:

* Apple II/II+ with SHIFT key mod enabled now act as expected ([MT#8917](http://mametesters.org/view.php?id=8917))

* SHIFT key mod is no longer applied to the IIe or IIe Enhanced (pad X6 is a [broken diamond](http://www.applefritter.com/files/styles/95-percent/public/mapetitecollection.com_P.A.L..jpg)) but is [still applied](http://web.archive.org/web/20040130132820/http://web.pdx.edu/~heiss/technotes/aiie/tn.aiie.09.html#:~:text=facilitating%20the%20shift-click%20mouse%20selection) to the IIe Platinum (pad X6 is a [connected bowtie](http://www.applefritter.com/files/styles/95-percent/public/Apple_IIe_platinum.jpg))

* Some cassettes (i.e. "hi" quality WAVs via "Apple Game Server") which previously failed to LOAD with CHKSUM ERROR, now work (and the "lo" WAVs continue to work):
before:
apple2p_cass_282_before
after:
apple2p_cass_282_after

Documentation:

* It is unintuitive that PB2 (and SW3 on the IIgs) behaves differently than PB0/PB1.

* Apple's documentation about the gameio pushbuttons says things like "the appropriate game input is connected to bit 7 of the data bus" without specifying if they are high or low. But the //c Tech Ref explicitly mentions [C063 in context of the shift-key mod](http://archive.org/details/AppleIIcTechnicalReference2ndEd/page/n497/mode/2up) and says "0 if it is pressed".

* Sather's [UTA2E Fig 7.2](http://archive.org/details/Understanding_the_Apple_IIe/page/n173/mode/2up) shows pull-down resistors on PB0 and PB1, but not PB2.

* Also, there is good [anecdotal corroboration of what's happening](http://comp.sys.apple2.narkive.com/ibs17S9K/apple-ii-s-paddle-3-button#:~:text=The%20cassette%20input%20is%20inverting) from respected source Michael J. Mahon.

* Sather also describes the LM741 inverting the cassette input, in [UTA2E 7-8.](http://archive.org/details/Understanding_the_Apple_IIe/page/n175/mode/2up)

Testing:

* The [previous SWITCHES test](http://github.com/mamedev/mame/pull/14307) shows that the default state of C06X softswitch reads (with or without a joystick attached) now matches hardware behavior, verified against a ][+ clone, //c, Apple IIe Enhanced and Apple IIgs ROM 3: both C060 and C063 default to high bit ON, except on a //c, C060 depends instead on the 40/80 switch.

* This PR is a [group effort](http://app.slack.com/client/T1J8S1LGH/CABEM8JFK): Shift key behavior was verified across various hardware by folks on Slack; kudos to: SpecLad, Will Scullin, Joshua Bell, and Tom Greene.

Caveats:
* This PR applies to common clones (and the tk3000) but I did not touch the tk2000 or the Apple 1 cassette.
* The shift-key mod could be added back to the IIe and IIe Enhanced via an `m_sysconfig` option bit, if someone wants that.

### DIFF
--- a/src/mame/apple/apple2.cpp
+++ b/src/mame/apple/apple2.cpp
@@ -468,8 +468,8 @@ u8 apple2_state::flags_r(offs_t offset)
 	// Y output of 74LS251 at H14 read as D7
 	switch (offset)
 	{
-	case 0: // cassette in (accidentally read at $C068 by ProDOS to attempt IIgs STATE register)
-		return (m_cassette->input() > 0.0 ? 0x80 : 0) | uFloatingBus7;
+	case 0: // cassette in, inverted (accidentally read at $C068 by ProDOS to attempt IIgs STATE register)
+		return (m_cassette->input() > 0.0 ? 0 : 0x80) | uFloatingBus7;
 
 	case 1:  // button 0
 		return (m_gameio->sw0_r() ? 0x80 : 0) | uFloatingBus7;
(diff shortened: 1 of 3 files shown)


arekkusu42:
(I would update MT#8917 but it seems to be impossible to create a MT account.)

rb6502:
I'll update the MT issue.

arekkusu42:
Additional corroboration: XPS Diagnostics //e 1.0.5 GAME PORT TEST checks PB0/PB1 with BMI, but PB2 with BPL. So the inverted C063 state was known in 1983.

PR #14460, opened 2025-11-02
esqvfd.cpp: Ignore 'ff' light commands. 'e8' also clears attributes.

cbrunschen:
'ff' light commands are handled in esqpanel.cpp; in esqvfd.cpp, ignore them
'e8' also cancels both blink and underline attributes.


cbrunschen:
Made couple of improvements to this PR as suggested offline:
- add `save_item()` calls to `device_start()`
- normalize braces to a single style across both header and implementation files
- reuse the existing `m_lastchar` instead of adding a new `m_ignore_next`

MooglyGuy:
Looking good. :) That's all of my feedback addressed, and it's nice to have the .cpp match the header now, bracing-wise.

cbrunschen:
Thank you!

PR #14461, opened 2025-11-02
[Access Virus] Analog input knobs for Virus A & B

felipesanches:


felipesanches:
On Virus B, this can be tested by holding "W" during boot, which starts the power-on "knobs test".

src/mame/layout/virusb.lay
```diff
@@ -60,7 +60,22 @@ license:CC0-1.0
 		<disk><color red="0.8" green="0.8" blue="0.8"/><bounds x="0" y="0" width="5" height="5"/></disk>
 	</element>
 
-	<element name="knob">
+
+	<element name="transparent"><rect><color red="0.0" green="0.0" blue="0.0" alpha="0.0"/></rect></element>
+	<element name="knob_outer"><disk><color red="0.0" green="0.0" blue="0.0" alpha="0.0"/></disk></element>
+	<element name="knob_value">
+		<simplecounter maxstate="100" digits="1"><color red="0.8" green="0.8" blue="0.8"/></simplecounter>
```
m1macrophage, line 67: The max value in `PORT_MINMAX` is set to 127. Should the same be used for `maxstate`?

felipesanches: I agree, in principle. But testing the proposed change locally I am seeing no difference.

PR #14465, opened 2025-11-03
apple2e: fix IIc mouse regression

### LINKED ISSUE #14463: apple2c - mouse only moves right
### MAME version

0.282

### System information

macOS via Ample

### INI configuration details

```INI

```

### Emulated system/software

apple2c, apple2cp

### Incorrect behaviour

In the apple2c variations (tried apple2c and apple2cp) the mouse cursor only moves right (i.e. not left, up, or down) in programs like MousePaint or Apple II DeskTop. 

Worked correctly in 0.281

### Expected behaviour

Normal mouse movement in apps that support it. 

### Steps to reproduce

_No response_

### Additional details

_No response_

arekkusu42:
This PR fixes #14463, which is a regression introduced by 92aa47f.

The "unconnected joysticks now read properly" portion of that commit fixed joystick 1 as advertised, however it incorrectly propagated similar logic to the IIc-specific mouse handling for C066/67.  The result is that the mouse stops functioning if no joystick is plugged in.

For folks affected by this in 0.282, a workaround is to launch with `-gameio joy`, which will restore mouse functionality.

### DIFF
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -2176,14 +2176,10 @@ u8 apple2e_state::c000_iic_r(offs_t offset)
 
 		case 0x66: // mouse X1 (IIc only)
 		case 0x6e:
-			if (!m_gameio->is_device_connected())
-				return 0x80 | uFloatingBus7;
 			return (m_x1 ? 0x80 : 0) | uFloatingBus7;
 
 		case 0x67: // mouse Y1 (IIc only)
 		case 0x6f:
-			if (!m_gameio->is_device_connected())
-				return 0x80 | uFloatingBus7;
 			return (m_y1 ? 0x80 : 0) | uFloatingBus7;
 
 		case 0x78: case 0x7a: case 0x7c: case 0x7e:  // read IOUDIS


colinleroy:
Confirming both the regression and fix :)

cuavas:
> For folks affected by this in 0.282, a workaround is to launch with `-gameio joy`, which will restore mouse functionality.

Why is it possible to have a mouse and joystick connected at the same time at all?  Don’t they use the same physical connector on the IIc?

colinleroy:
> > For folks affected by this in 0.282, a workaround is to launch with `-gameio joy`, which will restore mouse functionality.
> 
> Why is it possible to have a mouse and joystick connected at the same time at all? Don’t they use the same physical connector on the IIc?

It's not physically possible, indeed: they do share the same connector. But MAME doesn't deactivate the mouse handling when a joystick is connected so far.

arekkusu42:
They do use the same connector, which is why I had added this code in the first place-- in order to use the mouse, it should be plugged in, no?  However, MAME isn't emulating the mouse via a `gameio` device, instead `apple2e_state::update_iic_mouse()` is hooked up via analog `optional_ioports`.

So, there is probably a (preexisting) bug here where if `-gameio joy` is used, the mouse *should not* work.  I don't have a //c here to test the C066/67 interactions; I only restored the old behavior.

PR #14467, opened 2025-11-03
Add UML bit field extract instructions

cuavas:
This adds `BFXU`/`BFXS`/`DBFXU`/`DBFXS` instructions to the recompiler framework for extracting unsigned or signed bit fields (something @galibert wanted).  They works similarly to the three-parameter form of MAME’s `BIT` helper, or the ARM `ubfx`/`sbfx` instructions.

This does add one piece of undefined behaviour: if the `width` operand modulo the operand size in bits is zero, the result is undefined.  I don’t particularly like undefined behaviour, but this makes the back-end implementations substantially simpler.

These generate the most optimal code when the `shift` and `width` operands are both immediate values and the field doesn’t wrap around from MSB to LSB of the source, but the code for other cases is pretty decent for x86-64 and AArch64.  I didn’t put a great deal of effort into optimising for i686, particularly for the 64-bit forms – I’m really done caring, and i686 just doesn’t have enough registers to have much fun.

I updated the Hyperstone E1 recompiler to use the new instructions.  The majority of these cases get converted to `SHR` anyway (extracting the FP field from SR), so it doesn’t really affect the generated code much, but it’s a clearer seeing it expressed this way.

In addition to the main purpose, this also includes a few bits and pieces:
* Uses the x86-64 `LZCNT` instruction if available to implement the UML `LZCNT` instruction, and also optimises the `BSR`-based implementation. This isn’t a frequently-used instruction, but it’s still easy optimisation.
* Avoids unnecessary x86-64 REX prefixes in a few situations.
* Standardises on abbreviated integer type names in the x86-64 back-end as there was a mixture of standard and abbreviated names.
* Slightly optimises code i686 generation for `DROLAND` and `DROLINS` when flags need to be calculated.  This doesn’t happen frequently, but I had to mess with related code for other reasons, so it’s basically free.
* Slightly adjusts generated code for one case of the Hyperstone E1 `SET` instruction to produce tighter native code on i686 and x86-64.

### DIFF
--- a/src/devices/cpu/e132xs/e132xsdrc_ops.hxx
+++ b/src/devices/cpu/e132xs/e132xsdrc_ops.hxx
@@ -651,8 +651,8 @@ void hyperstone_device::generate_trap_exception_or_int(drcuml_block &block, uml:
 	UML_MOV(block, I4, DRC_SR);                                       // I4 = old SR
 
 	UML_MOV(block, I1, I4);                                           // I1 = SR to be updated
-	UML_ROLAND(block, I3, I4, 32 - FP_SHIFT, 0x7f);                   // I3 = old FP
-	UML_ROLAND(block, I2, I4, 32 - FL_SHIFT, 0xf);                    // I2 = old FL
+	UML_BFXU(block, I3, I4, FP_SHIFT, 7);                             // I3 = old FP
+	UML_BFXU(block, I2, I4, FL_SHIFT, 4);                             // I2 = old FL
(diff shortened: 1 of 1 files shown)


MooglyGuy:
I wouldn't sweat the UB for modular-zero shifts. It's already UB per-architecture anyway. Anyone who knows what they're doing and is implementing a DRC frontend is probably already going to be catching that case in order to implement whatever a given CPU does in that situation, anyway.

PR #14469, opened 2025-11-03
[Access Virus B] Control Panel LEDs

felipesanches:


felipesanches:
Testable via LED test mode, by holding 'Q' (mapped to "LFO2 Edit" button) during boot.

A bit glitchy (for yet unknown reasons), but if you ignore the spurious blinks, the correct LEDs can be seen being controlled by the test-mode.
Select current LED by pressing 'M' or 'N' (mapped to Program Value '+' and '-' buttons, respectively).

felipesanches:
This morning I found a solution to fix the glitches. I'm now cleaning up and preparing an additional commit for this PR including the fix.

PR #14470, opened 2025-11-04
CD-i: Fix XA Audio Peaking

Vincent-Halver:
-cdicdic: Fixed audio peaking by decoding in 32-bit and clipping to 16 later. [Vincent Halver]

-cdislavehle: Simplified mouse position-update logic. [Vincent Halver]

### DIFF
--- a/src/mame/philips/cdicdic.cpp
+++ b/src/mame/philips/cdicdic.cpp
@@ -41,17 +41,6 @@ TODO:
 #define VERBOSE         (0)
 #include "logmacro.h"
 
-
-namespace {
-
-[[maybe_unused]] constexpr int16_t clip_int16(int32_t sample)
-{
-	return int16_t(std::clamp<int32_t>(sample, -32768, 32767));
-}
-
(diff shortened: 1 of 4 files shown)


src/mame/philips/cdicdic.cpp
```diff
@@ -267,119 +256,33 @@ const uint8_t cdicdic_device::s_sector_scramble[2448] =
...
-			}
-		}
-
-		xa += 128;
-	}
+inline void rotate_samples(int16_t val, int16_t& a, int16_t& b, int16_t& output) {
+	b = a;
+	a = val;
+	output = a;
 }
```
cuavas, line 271: If this is correct, it’s a confusing way of doing it.  Why aren’t you assigning `output = val` directly?  I can’t see a way aliasing could require it to be done like this putting the value through `a` as an intermediate.

Vincent-Halver: This is correct. You could use `output = val;` if you prefer.

PR #14476, opened 2025-11-04
sega/model2: tilemap fixes

gm-matthew:
When emulating System 24 tilemaps (which Model 2 uses) MAME currently treats color 0 of any palette as transparent, but some Model 2 games use color 0 to encode tile-specific colors. Here is a screenshot of Behind Enemy Lines demonstrating the problem:

image

Here is what that screenshot is supposed to look like:

image

The [Sega System 24 Hardware Notes](https://segaretro.org/Sega_System_24_Hardware_Notes_(2013-06-16)) specifies the following when ABSEL=0:

> Zero value pixels from the topmost layer are considered transparent.
> Zero value pixels from the bottommost layer show color 0 of the specified
> palette.
> 
> If the topmost layer is blanked, it is considered transparent.
> If the bottommost layer is blanked, it is filled with color 0 of palette 0.

On Model 2, tilemap B is usually the bottommost layer unless it has the priority flag set, in which case either tilemap A or the framebuffer is the bottommost layer depending on whether tilemap A also has the priority flag set. However, I have observed that the polygon layer is never displayed if it is empty even if both tilemaps A and B have priority (Virtual On title screen). It is unknown whether tilemap A would instead be chosen if tilemap B has priority set and tilemap A does not, making tilemap A the bottommost layer; I have not observed any games that depend on this.

The easiest solution I have found is to display color 0 from the tile-specific palette of tilemap B if all three would otherwise be transparent; this seems to fix all tilegen issues. My current solution works by drawing the layers in this order:

- tilemap B without priority as opaque
- tilemap B with priority as opaque
- tilemap A without priority
- frame buffer
- tilemap B with priority
- tilemap A with priority

The solution is not quite as neat as I'd like it to be so I'm submitting this commit as a draft for now; any ideas to make the solution more elegant are welcome.

### DIFF
--- a/src/mame/sega/model2_v.cpp
+++ b/src/mame/sega/model2_v.cpp
@@ -2599,11 +2599,9 @@ u32 model2_state::screen_update_model2(screen_device &screen, bitmap_rgb32 &bitm
 	bitmap.fill(m_palette->pen(0), cliprect);
 	m_sys24_bitmap.fill(0, cliprect);
 
+	// draw tilemap B as opaque
 	for (int layer = 3; layer >= 2; layer--)
-	{
-		m_tiles->draw(screen, m_sys24_bitmap, cliprect, layer << 1, 0, 1);
-		m_tiles->draw(screen, m_sys24_bitmap, cliprect, (layer << 1) | 1, 0, 1);
-	}
+		m_tiles->draw(screen, m_sys24_bitmap, cliprect, layer << 1, 0, TILEMAP_DRAW_OPAQUE);
 
 	for (int layer = 1; layer >= 0; layer--)
 		m_tiles->draw(screen, m_sys24_bitmap, cliprect, layer << 1, 0, 0);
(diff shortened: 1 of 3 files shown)


angelosa:
There are several games that depends on background pen being chosen by lowermost layer, the most notable is `crkdown`. 
I also vaguely remember a bug of the like in Model 3, don't remember which game tho.

mamehaze:
Yeah the map display has been broken on Crackdown forever in MAME, bits that are meant to be solid are showing as transparent.

gm-matthew:
I've come up with a more elegant fix which uses the TILEMAP_DRAW_OPAQUE flag when calling `tilemap_t::draw()` and also adds the feature to `draw_rect()` .

I am satisfied with my current solution so I am declaring this commit as ready for review.

Regarding System 24 specifically, are there any other games besides Crack Down that are known to have similar tilemap bugs?

angelosa:
`qsww` draws like this at startup on master tip, which apparently is a transparent pen issue

angelosa (approved):
LGTM

PR #14478, opened 2025-11-04
adp/stella8085.cpp: New systems marked not working

stonedDiscord:
21 Point [stonedDiscord]
4 Asse F2 [stonedDiscord]
As [stonedDiscord]
Astro [stonedDiscord]
Bistro [stonedDiscord]
Club [stonedDiscord]
Disc II Plus [stonedDiscord]
Disc Olympia [stonedDiscord]
Disc Royal (DOB) [stonedDiscord]
Full House [stonedDiscord]
Herz As (F10) [stonedDiscord]
Max [stonedDiscord]
Multi [stonedDiscord]
Sesam [stonedDiscord]
Super Herz As [stonedDiscord]
Super Jackpot [stonedDiscord]
Super Pro [stonedDiscord]
Treff As [stonedDiscord]
Trio [stonedDiscord]

New clones marked not working
-----------------------------
4 Asse F1 [stonedDiscord]
Disc Royal [stonedDiscord]
Herz As (F8) [stonedDiscord]
Herz As [stonedDiscord]
Super Multi [stonedDiscord]


stonedDiscord:
# Other changes:
Moved Glücks-Stern to the correct driver (was stella8085 but belongs in stellafr as it's 68000 based).
Games on the old 4040 board use different IO ports, the KDC works on those now.
Fixed the service display (~~might be swapped with money~~ confirmed with test rom).
Intel 8256 will now return the interrupt level.
2 missing lamps in the disc2000 layout, there's 2 lamps below the risk buttons (labelled NICHTS (NOTHING)).

src/mame/adp/stella8085.cpp
```diff
@@ -124,40 +129,53 @@ void stella8085_state::machine_start()
...
+void stella8085_state::io_4040_map(address_map &map)
 {
-	map(0x0000, 0x4fff).rom();
-	map(0x5000, 0x5fff).ram();
-	map(0x6000, 0x633f).rw("rtc", FUNC(mc146818_device::read_direct), FUNC(mc146818_device::write_direct));
-	map(0x7000, 0x7fff).rom();
+	map(0x00, 0x00).w(FUNC(stella8085_state::io00));
+	map(0x70, 0x73).rw("ppi", FUNC(i8255_device::read), FUNC(i8255_device::write));
+	map(0x80, 0x81).rw("kdc", FUNC(i8279_device::read), FUNC(i8279_device::write));
+	map(0x90, 0x9f).rw("muart", FUNC(i8256_device::read), FUNC(i8256_device::write));
```
cuavas: Remember it’s better to use object finders rather than literal tags if you have them anyway.

src/mame/adp/stella8085.cpp
```diff
@@ -614,14 +624,46 @@ void stella8085_state::doppelpot(machine_config &config)
...
+	config.set_default_layout(layout_adpservice);
+
+	SPEAKER(config, "mono").front_center();
+	BEEP(config, "beeper", 0)
+		.add_route(ALL_OUTPUTS, "mono", 0.50);
 }
 
 ROM_START( bahia )
-	ROM_REGION( 0x8000, "maincpu", 0 )
+	ROM_REGION( 0x10000, "maincpu", 0 )
```
cuavas: If you’re making the ROM region bigger than the content, you should probably set a fill byte value for how empty space reads.

src/devices/machine/i8256.cpp
```diff
@@ -389,7 +435,7 @@ void i8256_device::write(offs_t offset, u8 data)
 			m_interrupts = m_interrupts & ~data;
 			break;
 		case I8256_REG_BUFFER:
-			LOG("I8256 write serial: %u\n", data);
+			LOG("I8256 write serial: %u %c\n", data, data);
```
cuavas: You want to be careful about this – you should sanitise the character before formatting with `%c`.

src/devices/machine/i8256.cpp
```diff
@@ -248,7 +268,29 @@ void i8256_device::device_reset()
...
+uint8_t i8256_device::acknowledge()
+{
+	const uint8_t vector = m_current_interrupt_level;
+	m_out_int_cb(CLEAR_LINE);
+	m_current_interrupt_level = -1;
+	if (BIT(m_command1,I8256_CMD1_8086)) // 8086 mode, TODO: only on second INTA
+		return 0x40 | vector;
+	else
+		return 0xc7 | (vector << 3); // 8085 mode c7 = rst 0, ff = rst 7
+}
```
cuavas, line 272: You’re forgetting side effect checks.

src/devices/machine/i8256.h
```diff
@@ -40,7 +40,7 @@
 class i8256_device : public device_t, public device_serial_interface
 {
 public:
-	static constexpr flags_type emulation_flags() { return flags::SAVE_UNSUPPORTED; }
+	static constexpr flags_type emulation_flags() { return flags::NONE; }
```
cuavas: You don’t need to implement this if there are no flags.

src/devices/machine/i8256.cpp
stonedDiscord: The datasheet nor the app note define what happens on INTA when no interrupt is pending.

stonedDiscord:
image
serial works now

stonedDiscord:
I don't think I have anything to add to this at this time.
Sadly it still gets stuck somewhere in the initialization procedure.
I've tested this branch plenty over the last weeks and it has helped very much in creating a [homebrew test ROM](https://github.com/stonedDiscord/weisse_cpu/tree/master/testrom) for the board and the emulated behavior matches what the real machine does.

stonedDiscord:
Can I get a re-review please?

stonedDiscord:
Going to split this up and get rid of that commit log.

PR #14480, opened 2025-11-04
midway/midvunit.cpp: shifter fix, dedicated neutral

Endprodukt:
This change adds a dedicated Neutral Gear input to the Midway V Unit driver.

Added “Neutral Gear” input definition to IN0 (bit 0x0020).

Updated gear handling logic in port0_r() to properly interpret the neutral input.

This works similar to the neutral gear in model 2.

### DIFF
--- a/src/mame/midway/midvunit.cpp
+++ b/src/mame/midway/midvunit.cpp
@@ -154,7 +154,6 @@ uint32_t midvunit_state::port0_r()
 	return (val << 16) | val;
 }
 
-
 /*************************************
  *
  *  ADC input ports


src/mame/midway/midvunit.cpp
```diff
@@ -132,22 +132,35 @@ void midvplus_state::machine_reset()
...
+		}
+		else if (!(val & 0x1000))
+		{
+			m_shifter_state = 4; // Gear 3
+		}
+		else if (!(val & 0x2000))
+		{
+			m_shifter_state = 8; // Gear 4
+		}
+
```
MooglyGuy: Please omit the braces so that the new code better-resembles the old code.

src/mame/midway/midvunit.cpp
```diff
@@ -132,22 +132,35 @@ void midvplus_state::machine_reset()
...
 		m_last_port0 = val;
```
MooglyGuy: With the removal of `diff`, `m_last_port0` is now entirely unused in the driver other than this assignment here. Depending on the compiler, this may pop an "assigned but never used" warning, so please remove the member entirely.

src/mame/midway/midvunit.cpp
```diff
@@ -742,15 +755,15 @@ static INPUT_PORTS_START( midvunit )
...
 	PORT_BIT( 0x0200, IP_ACTIVE_LOW, IPT_VOLUME_UP )
-	PORT_BIT( 0x0400, IP_ACTIVE_LOW, IPT_BUTTON8 ) PORT_NAME("4th Gear")    // 4th
-	PORT_BIT( 0x0800, IP_ACTIVE_LOW, IPT_BUTTON7 ) PORT_NAME("3rd Gear")    // 3rd
-	PORT_BIT( 0x1000, IP_ACTIVE_LOW, IPT_BUTTON6 ) PORT_NAME("2nd Gear")    // 2nd
-	PORT_BIT( 0x2000, IP_ACTIVE_LOW, IPT_BUTTON5 ) PORT_NAME("1st Gear")    // 1st
+	PORT_BIT( 0x0400, IP_ACTIVE_LOW, IPT_BUTTON9 ) PORT_NAME("4th Gear")    // 4th
+	PORT_BIT( 0x0800, IP_ACTIVE_LOW, IPT_BUTTON8 ) PORT_NAME("3rd Gear")    // 3rd
+	PORT_BIT( 0x1000, IP_ACTIVE_LOW, IPT_BUTTON7 ) PORT_NAME("2nd Gear")    // 2nd
+	PORT_BIT( 0x2000, IP_ACTIVE_LOW, IPT_BUTTON6 ) PORT_NAME("1st Gear")    // 1st
+	PORT_BIT( 0x0020, IP_ACTIVE_LOW, IPT_BUTTON5)  PORT_NAME("Neutral Gear")
```
MooglyGuy: Project preference is to keep ports in bit-order, so please don't move bit 0x0020 to be after bit 0x2000.

src/mame/midway/midvunit.cpp
```diff
@@ -132,22 +132,35 @@ void midvplus_state::machine_reset()
 uint32_t midvunit_state::port0_r()
 {
 	uint16_t val = m_in0->read();
-	uint16_t diff = val ^ m_last_port0;
 
 	if (!machine().side_effects_disabled())
 	{
-		// make sure the shift controls are mutually exclusive
```
happppp, line 139: Could keep this comment? It clarifies why the implementation is making an exception for the gearbox.

MooglyGuy: True, and the new implementation without the need for `diff` or the ternary operators actually makes the comment make more sense.

src/mame/midway/midvunit.cpp
Endprodukt: if I bring them in order the neutral gear will pop up unconnected to the gears in the menu, no?
  
  PORT_BIT( 0x0020, IP_ACTIVE_LOW, IPT_BUTTON5 ) PORT_NAME("Neutral Gear")
PORT_BIT( 0x0040, IP_ACTIVE_LOW, IPT_SERVICE1 )
PORT_BIT( 0x0080, IP_ACTIVE_LOW, IPT_COIN3 )
PORT_BIT( 0x0100, IP_ACTIVE_LOW, IPT_VOLUME_DOWN )
PORT_BIT( 0x0200, IP_ACTIVE_LOW, IPT_VOLUME_UP )
PORT_BIT( 0x0400, IP_ACTIVE_LOW, IPT_BUTTON9 ) PORT_NAME("4th Gear")    // 4th
PORT_BIT( 0x0800, IP_ACTIVE_LOW, IPT_BUTTON8 ) PORT_NAME("3rd Gear")    // 3rd
PORT_BIT( 0x1000, IP_ACTIVE_LOW, IPT_BUTTON7 ) PORT_NAME("2nd Gear")    // 2nd
PORT_BIT( 0x2000, IP_ACTIVE_LOW, IPT_BUTTON6 ) PORT_NAME("1st Gear")    // 1st

cuavas: This is going to mess stuff up because the system will actually be reading the fake neutral button in bit five of the coin/start/service port.

src/mame/midway/midvunit.cpp
```diff
@@ -132,28 +131,31 @@ void midvplus_state::machine_reset()
...
+		if (!(val & 0x0020))
+			m_shifter_state = 0; // Neutral
+		else if (!(val & 0x0400))
+			m_shifter_state = 1; // Gear 1
+		else if (!(val & 0x0800))
+			m_shifter_state = 2; // Gear 2
+		else if (!(val & 0x1000))
+			m_shifter_state = 4; // Gear 3
+		else if (!(val & 0x2000))
+			m_shifter_state = 8; // Gear 4
```
cuavas, line 148: Is there a good reason for not preserving detecting buttons being pressed?  Previously, if you pressed a gear button before releasing the previous gear, button it would shift immediately.  This change makes it only *downshift* immediately on pressing a button.  It makes more sense for the behaviour to be the same irrespective of the gears involved.

Endprodukt: it was wrong before. the gear would shift to neutral on double press (same gear twice)

Endprodukt: also I changed the code to use with a real shifter. that was my motivation to begin with.

Endprodukt: this has been tested with a shifter and keyboard. works just fine and as I would expect. 

btw I kindly as you to understand that this is my very first commit and I literally started working with visual studio yesterday. So please excuse mistakes. I did test before the pull request and it works.

cuavas: > it was wrong before. the gear would shift to neutral on double press (same gear twice)

That was intentional, to make it playable with buttons.  It isn’t intuitive, but it can be learned.

> also I changed the code to use with a real shifter. that was my motivation to begin with.

It won’t work with the original shifter that lacks a neutral switch.

cuavas:
I did it myself, it’s simpler: 6523b657fff094b188e6303626a3ec78081f433d

src/mame/midway/midvunit.cpp
Endprodukt: It's setup just like Daytona and Sega Rally.

PR #14481, opened 2025-11-05
CD-i: Add Audio Mixing

### LINKED ISSUE #12852: [CD-i] Sound mixing not faithful to the real machine
### MAME version

0.270 (mame0270-59-g9c54072486b)

### System information

Debian 12

### INI configuration details

_No response_

### Emulated system/software

CD-i (cdimono1)

### Incorrect behaviour

Multiple mono audio sources are played back with total stereo separation. For example, Zelda - Wand of Gamelon has the BGM on the left and SFX on the right.

### Expected behaviour

Zelda - Wand of Gamelon has the BGM and SFX mixed on both channels.

### Steps to reproduce

Play some Zelda - Wand of Gamelon and experience it during while playing a 2D level

### Additional details

_No response_

Vincent-Halver:
-cdicdic: Add audio attenuation (fixes #12852) [Vincent Halver]


galibert:
Compiling src/mame/philips/cdislavehle.cpp...
../../../../../src/mame/philips/cdislavehle.cpp:233:43: error: 'm_cdic' is a protected member of 'cdi_state'
  233 |                                                         dynamic_cast<cdi_state*>(m_owner)->m_cdic->atten_w(m_in_buf);
      |                                                                                            ^
In file included from ../../../../../src/mame/philips/cdislavehle.cpp:32:
../../../../../src/mame/philips/cdi.h:57:34: note: declared protected here
   57 |         optional_device<cdicdic_device> m_cdic;
      |                                         ^

Also, this is kind of atrocious code.  What are you trying to do exactly?

src/mame/philips/cdicdic.cpp
```diff
@@ -448,13 +446,36 @@ void cdicdic_device::play_audio_sector(const uint8_t coding, const uint8_t *data
...
 	{
+		uint16_t offset = 0;
 		for (uint16_t i = 0; i < SECTOR_AUDIO_SIZE; i += 128, data += 128)
 		{
-			play_xa_group(coding, data);
+			play_xa_group(coding, data, offset);
+			offset += 28 * num_samples;
+		}
+		int16_t sampleL = 0, sampleR = 0, outL = 0, outR = 0;
+		for (uint16_t i = 0; i < 18 * 28 * num_samples; i++) {
```
angelosa: File uses Allman style.

src/mame/philips/cdislavehle.cpp
```diff
@@ -227,6 +228,12 @@ void cdislave_hle_device::slave_w(offs_t offset, uint16_t data)
 				{
 					switch (m_in_buf[0])
 					{
+						case 0xc0: case 0xc1: case 0xc2: case 0xc3: case 0xc4: case 0xc5: case 0xc6: case 0xc7:
+						case 0xc8: case 0xc9: case 0xca: case 0xcb: case 0xcc: case 0xcd: case 0xce: case 0xcf:
+							dynamic_cast<cdi_state*>(m_owner)->m_cdic->atten_w(m_in_buf);
```
angelosa, line 233: This is unportable, you should rather provide this communication thru `devcb_write8::array<4>`.

src/mame/philips/cdicdic.cpp
```diff
@@ -318,50 +318,33 @@ void cdicdic_device::play_raw_group(const uint8_t *data)
...
 {
 	static const uint16_t s_4bit_header_offsets[8] = { 4, 5, 6, 7, 12, 13, 14, 15 };
 	static const uint16_t s_8bit_header_offsets[4] = { 4, 5, 6, 7 };
 	static const uint16_t s_4bit_data_offsets[8] = { 16, 16, 17, 17, 18, 18, 19, 19 };
 	static const uint16_t s_8bit_data_offsets[4] = { 16, 17, 18, 19 };
-
-	int16_t samples[28];
-
-	switch (coding & (CODING_BPS_MASK | CODING_CHAN_MASK))
+	uint8_t num_samples = coding&CODING_8BPS ? 4 : 8;
```
MooglyGuy: Missing spaces.

src/mame/philips/cdicdic.cpp
```diff
@@ -318,50 +318,33 @@ void cdicdic_device::play_raw_group(const uint8_t *data)
 	m_dmadac[1]->transfer(0, 1, 1, 28, samples);
 }
 
-void cdicdic_device::play_xa_group(const uint8_t coding, const uint8_t *data)
+void cdicdic_device::play_xa_group(const uint8_t coding, const uint8_t *data, const uint16_t idx)
 {
 	static const uint16_t s_4bit_header_offsets[8] = { 4, 5, 6, 7, 12, 13, 14, 15 };
```
MooglyGuy: Unsure if I originally added these or someone else did, but the s_ prefix is unnecessary, and `SHOUTY_CAPS` is prefered for constants; something like `HEADER_OFFSETS_4BIT` et al would be nicer.

src/mame/philips/cdicdic.cpp
```diff
@@ -377,10 +360,12 @@ void cdicdic_device::play_cdda_sector(const uint8_t *data)
 	{
 		samples[0][i] = (int16_t)((data[(i * 4) + 1] << 8) | data[(i * 4) + 0]);
 		samples[1][i] = (int16_t)((data[(i * 4) + 3] << 8) | data[(i * 4) + 2]);
+		m_samples[0][i] = samples[0][i];
+		m_samples[1][i] = samples[1][i];
 	}
 
-	m_dmadac[0]->transfer(0, 1, 1, SECTOR_SIZE/4, samples[0]);
-	m_dmadac[1]->transfer(0, 1, 1, SECTOR_SIZE/4, samples[1]);
+	m_dmadac[0]->transfer(0, 1, 1, SECTOR_SIZE/4, &m_samples[0][0]);
```
MooglyGuy: More missing spaces. If the line is being changed anyway, might as well fix the spacing along the way. The whole file could use a sweep for operator-spacing normalization.

src/mame/philips/cdicdic.cpp
```diff
@@ -448,13 +446,36 @@ void cdicdic_device::play_audio_sector(const uint8_t coding, const uint8_t *data
...
+			m_dmadac[1]->transfer(0, 1, 1, 1, &outR);
 		}
 	}
 }
 
+void cdicdic_device::atten_w(uint8_t* args) {
+	m_atten[0] = ((args[0] & 0x8) << 7) | args[1];
+	m_atten[1] = ((args[0] & 0x4) << 7) | args[2];
+	m_atten[2] = ((args[0] & 0x2) << 7) | args[3];
+	m_atten[3] = ((args[0] & 0x1) << 7) | args[4];
```
MooglyGuy, line 479: Following on from Kale's comment, it's pretty unlikely that the real hardware connection between the SLAVE and CDIC devices is 20 individual traces for the 20 bits of state represented by `args` here.

The correct way to do this would be to have `atten_w` take a single `uint8_t`, have a private member that increments on each call from 0 to 4 and wrapping around, in order to determine how to apply the incoming byte to the `m_atten` array. You would then have a `devcb_write8` with a public callback-registration function in `cdislavehle_device` - there are plenty of examples throughout the codebase on how to do this - and then in the machine configuration function for `cdi_state::cdimono1_base`, do -

`m_slave_hle->atten_callback().set(m_cdic, FUNC(cdicdic_device::atten_w));`

rather than grabbing into a protected class member in a way that doesn't compile.

Vincent-Halver: Added a temporary fix to expose the cdic for dynamic casting just so that it compiles so it can be validated it worked. Thanks for the feedback, I'll look into the callback change.

src/mame/philips/cdislavehle.cpp
Vincent-Halver: Acknowledged. If it is acceptable, I will use a follow up PR on this detail.

src/mame/philips/cdicdic.cpp
Vincent-Halver: Thank you for the example. Excluding the Mute bits here will result in popping sounds. Hotel Mario shows that this solution is *close*, but not entirely correct. I don't exactly know what the correct technical implementation is supposed to be yet.

If it is acceptable, I will commit this in the (largely) working state. I have left a TODO here.

MooglyGuy: > Thank you for the example. Excluding the Mute bits here will result in popping sounds. Hotel Mario shows that this solution is _close_, but not entirely correct. I don't exactly know what the correct technical implementation is supposed to be yet.
> 
> If it is acceptable, I will commit this in the (largely) working state. I have left a TODO here.

I'm not sure I follow, why do you need to exclute the mute bits?

Vincent-Halver:
I'm not sure how to properly implement the byte-write paradigm as suggested by Angelosa.

To reduce the complexity of the change, I suggest accepting this change as is, and I will try to remove the dynamic_cast in the next change.

Vincent-Halver:
Somehow my local copy became very messed up. I opened a different request.

PR #14483, opened 2025-11-05
Update goldstar.cpp cmast97 settings

jbtx77gr:
- adding cmast97 correct settings


src/mame/igs/goldstar.cpp
```diff
@@ -6335,150 +6335,181 @@ static INPUT_PORTS_START( cmast91 )
...
 
 	PORT_START("DSW1")
-	PORT_DIPNAME( 0x01, 0x01, "Double Up Game" )   PORT_DIPLOCATION("DSW1:!1")
-	PORT_DIPSETTING(    0x00, DEF_STR( Off ) )
-	PORT_DIPSETTING(    0x01, DEF_STR( On ) )
-	PORT_DIPNAME( 0x02, 0x00, "Double Up Type" )   PORT_DIPLOCATION("DSW1:!2")
+	PORT_DIPNAME( 0x01, 0x01, "Double Up Game" )          PORT_DIPLOCATION("DSW1:!1")
+	PORT_DIPSETTING(    0x01, "Used" )
+	PORT_DIPSETTING(    0x00, "Not Use" )
+	PORT_DIPNAME( 0x02, 0x00, "Double Up Game Type" )     PORT_DIPLOCATION("DSW1:!2")
```
RobertoFresca, line 6388: In the case of  things that are "yes/no" or "on/off", is better to use the def strings instead of some chinese "engrish". I know that is what the manual literally says, but the last thing we want is that someone jumps over due to the engrish sentences.

cuavas: It’s better to use standard descriptions in general.  Also we try to order settings consistently:
* Off before on
* Least generous to most generous
* Smallest to biggest
* etc.

PR #14488, opened 2025-11-06
adp/servicetastatur.cpp: Add layout

stonedDiscord:
I built a test ROM and put that on the hardware and in the emulator to test the key mapping and it matches.
The layout is clickable.
Serial is there but I haven't gotten it to interact with any of the stellafr machines yet, [this may or may not be an issue on real hardware](https://forum.goldserie.de/viewtopic.php?p=455737#p455737) as well

### DIFF
--- a/src/mame/adp/servicetastatur.cpp
+++ b/src/mame/adp/servicetastatur.cpp
@@ -40,11 +40,15 @@ ___| XTAL  80C31          +KEYPAD+       |__
 */
 
 #include "emu.h"
+
 #include "cpu/mcs51/i80c51.h"
 #include "machine/i2cmem.h"
 #include "video/hd44780.h"
+
 #include "emupal.h"
 #include "screen.h"
+
+#include "servicet.lh"
 
 namespace {
 
@@ -318,4 +322,4 @@ ROM_END
 
 } // anonymous namespace
 
-GAME( 1992, servicet, 0, servicet, servicet, servicet_state, empty_init, ROT0, "ADP", u8"Merkur Service Testgerät", MACHINE_NOT_WORKING | MACHINE_NO_SOUND_HW )
+GAMEL( 1992, servicet, 0, servicet, servicet, servicet_state, empty_init, ROT0, "ADP", u8"Merkur Service Testgerät", MACHINE_NOT_WORKING | MACHINE_NO_SOUND_HW, layout_servicet )
--- a/src/mame/layout/servicet.lay
+++ b/src/mame/layout/servicet.lay
@@ -93,39 +93,39 @@ license:CC0-1.0
 			<bounds x="3" y="3" width="16" height="2" />
 		</screen>
 
-		<element name="F1" ref="F1" inputtag="IN2">
+		<element name="F1" ref="F1">
 			<bounds x="4" y="6" width="2" height="2" />
 		</element>
 
-		<element name="F2" ref="F2" inputtag="IN2">
+		<element name="F2" ref="F2">
 			<bounds x="8" y="6" width="2" height="2" />
 		</element>
 
-		<element name="F3" ref="F3" inputtag="IN2">
+		<element name="F3" ref="F3">
 			<bounds x="12" y="6" width="2" height="2" />
 		</element>
 
-		<element name="F4" ref="F4" inputtag="IN0">
+		<element name="F4" ref="F4">
 			<bounds x="16" y="6" width="2" height="2" />
 		</element>
 
-		<element name="OK" ref="OK" inputtag="IN0">
+		<element name="OK" ref="OK">
 			<bounds x="19" y="8" width="3" height="2" />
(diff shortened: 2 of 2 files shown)


src/mame/adp/servicetastatur.cpp
```diff
@@ -119,90 +144,136 @@ void servicet_state::servicet_map(address_map &map)
...
-	// CPU drives each line HIGH and checks if connected lines are also HIGH = button pressed
+	/*
+	* the keypad scan works by writing 0xfb,0xfd,0xfe (pulling rows low) to port 1
+	* send seeing which bits end up low instead of high
+	*/
+	uint8_t data = m_port1;
 
-	// Column-to-Row scanning: When columns (P1.0-P1.2) are driven HIGH
-	for (int col = 0; col < 3; col++)
+	switch (m_port1)
```
startaq: Don't use a `switch` here, instead just check if any bit is low and multiplex the data. For example:

```
data = 0xff;

if (BIT(m_port1, 1)) data &= ioport("IN0")->read();
if (BIT(m_port1, 2)) data &= ioport("IN1")->read();
if (BIT(m_port1, 3)) data &= ioport("IN2")->read();
```

Also you already have `m_io_keys`, use it.

src/mame/adp/stellafr.cpp
```diff
@@ -308,6 +321,12 @@ void stellafr_state::mux_w(uint8_t data)
 		;
 	if (aw2)
 		;
+
+	m_st->write_rts(1);
+	m_st->write_dtr(enanz1);
+	m_st->write_rts(0);
```
startaq: That's certainly not what the hardware does. `rts` will just be 0 all the time (same for the code below).

src/mame/adp/servicetastatur.cpp
```diff
@@ -119,90 +144,136 @@ void servicet_state::servicet_map(address_map &map)
...
 
+INPUT_CHANGED_MEMBER(servicet_state::en_w)
+{
+	enable_in(newval);
+}
+
 uint8_t servicet_state::port3_r()
 {
-	uint8_t data = m_port3;
+	uint8_t data = ioport("P3")->read();
```
startaq: This needs `required_ioport`.

src/mame/adp/servicetastatur.cpp
```diff
@@ -119,90 +144,136 @@ void servicet_state::servicet_map(address_map &map)
...
-	PORT_BIT( 0x20, IP_ACTIVE_HIGH, IPT_KEYPAD) PORT_NAME("F4") PORT_CODE(KEYCODE_F4)
-	PORT_BIT( 0x40, IP_ACTIVE_HIGH, IPT_JOYSTICK_UP)    PORT_4WAY
+	PORT_BIT( 0x01, IP_ACTIVE_HIGH, IPT_UNUSED)
+	PORT_BIT( 0x02, IP_ACTIVE_LOW, IPT_UNUSED)
+	PORT_BIT( 0x04, IP_ACTIVE_LOW, IPT_UNUSED)
+	PORT_BIT( 0x08, IP_ACTIVE_LOW, IPT_UNUSED)
+	PORT_BIT( 0x10, IP_ACTIVE_LOW, IPT_START) PORT_NAME("OK")
+	PORT_BIT( 0x20, IP_ACTIVE_LOW, IPT_BUTTON4) PORT_NAME("F4")
+	PORT_BIT( 0x40, IP_ACTIVE_LOW, IPT_JOYSTICK_UP)    PORT_4WAY
+	PORT_BIT( 0x80, IP_ACTIVE_LOW, IPT_UNUSED)
```
startaq: Spacing is a bit messed up here. And you can shorten this to `PORT_BIT( 0x0f, IP_ACTIVE_LOW, IPT_UNUSED)`. "Random" `ACTIVE_HIGH` is weird too, just update the code used to read the inputs.

startaq (changes requested):

stonedDiscord:
I've removed serial from the PR as it was probably the wrong choice from the start.
It uses Clock, Enable, In and Out.
Shoehorning this into RS-232 was a dumb idea.
I'll probably write my own bus driver for this at a later time as the lamp outs and the coin mech use the same "protocol"

stonedDiscord:
For this one I also think it's ready to merge for now.
I've written a [test ROM](https://github.com/stonedDiscord/servicetastatur/tree/master/firmware) for this one as well and it matches the hardware (including the screen flicker).

stonedDiscord:
Made it work. Press 9 once. If you pick the wrong option it will freeze, the real hardware does this as well if it doesn't get a reply from the machine.

stonedDiscord:
I've cut it back to only do what it says in the title.

stonedDiscord:
Bump,
this is only adding the layout so it should be an easy review+merge.

PR #14490, opened 2025-11-07
segaic24: draw bottom tilemap layer as opaque

gm-matthew:
Applying my recent Model 2 tilemap fixes to System 24 and Model 1 as well; fixes the title screen of `qsww` and most of the issues in [MT06379](https://mametesters.org/view.php?id=6379).

The one remaining tilemap bug in Crack Down is the description of the time bomb not scrolling into view during attract mode; this is due to the scrolling of tall playfields not being implemented properly and I am working on a separate fix for this.

I noted that the System 24 tilegen issues could also be fixed by simply always drawing tilemap B as opaque before drawing the tilemap layers as normal, rather than drawing all the tilemap layers in reverse order:

`m_vtile->draw(screen, bitmap, cliprect, 6, 0, TILEMAP_DRAW_OPAQUE);`
`m_vtile->draw(screen, bitmap, cliprect, 4, 0, TILEMAP_DRAW_OPAQUE);`

However, since the [System 24 Hardware Notes](https://segaretro.org/Sega_System_24_Hardware_Notes_(2013-06-16)) specify that "zero value pixels from the _bottommost_ layer show color 0 of the specified palette", I have implemented the behavior as described.

### DIFF
--- a/src/mame/sega/model1_v.cpp
+++ b/src/mame/sega/model1_v.cpp
@@ -1649,8 +1649,9 @@ uint32_t model1_state::screen_update_model1(screen_device &screen, bitmap_rgb32 
 	screen.priority().fill(0);
 	bitmap.fill(m_palette->pen(0x400), cliprect);
 
-	m_tiles->draw(screen, bitmap, cliprect, 6, 0, 0);
-	m_tiles->draw(screen, bitmap, cliprect, 4, 0, 0);
+	// draw tilemap B as opaque
+	m_tiles->draw(screen, bitmap, cliprect, 6, 0, TILEMAP_DRAW_OPAQUE);
+	m_tiles->draw(screen, bitmap, cliprect, 4, 0, TILEMAP_DRAW_OPAQUE);
 	m_tiles->draw(screen, bitmap, cliprect, 2, 0, 0);
 	m_tiles->draw(screen, bitmap, cliprect, 0, 0, 0);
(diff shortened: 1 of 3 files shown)


mamehaze:
not sure if you've noticed (or care since I don't think it affects Model 2) but sgmast still has some layer offset problems which look similar to tilemap transpent issues but aren't. I think Charles MacDonald noted some years back, but one of the registers is meant to shift sprite(?) layers by 8 pixels.

He did do a writeup which can still be found at https://segaretro.org/Sega_System_24_Hardware_Notes_(2013-06-16) but I don't remember if it covers that.

gm-matthew:
> not sure if you've noticed (or care since I don't think it affects Model 2) but sgmast still has some layer offset problems which look similar to tilemap transpent issues but aren't. I think Charles MacDonald noted some years back, but one of the registers is meant to shift sprite(?) layers by 8 pixels.
> 
> He did do a writeup which can still be found at https://segaretro.org/Sega_System_24_Hardware_Notes_(2013-06-16) but I don't remember if it covers that.

The register in question is `HOUT`, which indeed can be used to scroll the sprite layer relative to the tilemap layers. By default it is set to `0xFFC6` but `sgmast` sets it to `0xFFCE` which is supposed to shift sprites 8 pixels to the right, but MAME doesn't emulate this currently.

PR #14491, opened 2025-11-07
Modernised Sord m5 driver

dlabi:
- fd5 floppy emulation works but only with rom hack
	- updated Sord m5 www links
	- added fd5 utility disk to software list - not original dump, made from program listings
	- added support of optional sram in Basic-F and Basic-G cartridges. Works only if shortname(softlist) is used
	- marked 32/64KB RAM expansions EM-5, 64KBI, 64KBF, 64KRX as supported in sofwarelist
	- reenabled and refactored memory banking
	- added support of E and F block types for cassettes

	- brno mod: switched to rom including basic-i
	- brno mod: reenabled and refactored memory banking

### DIFF
--- a/src/mame/sord/m5.cpp
+++ b/src/mame/sord/m5.cpp
@@ -535,7 +535,7 @@ void m5_state::mem64KBI_w(offs_t offset, u8 data) //out 0x6c
 	m_ram_mode = BIT(data, 0);
 
 	//if 32kb only mode don't map top ram
-	if (m_ram_mode && !BIT(m_DIPS->read(), 4))
+	if (m_ram_mode && !BIT(m_DIPS->read(), 2))
 		m_rom_view.select(2); // 64kb RAM
 	else
 
@@ -676,7 +676,7 @@ void m5_state::mem64KRX_w(offs_t offset, u8 data) //out 0x7f
 	m_bankr[1]->set_entry(BIT(m_ram_mode, 1));
 	m_bankr[2]->set_entry(BIT(m_ram_mode, 2));
 
-	if (BIT(m_DIPS->read(),1))
+	if (BIT(m_DIPS->read(),0))
 	{
 		m_bankr[4]->set_entry(BIT(m_ram_mode, 4));
 		m_bankr[5]->set_entry(BIT(m_ram_mode, 5));
@@ -727,9 +727,6 @@ void m5_state::m5_mem(address_map &map)
 	m_rom_view[1](0x0000, 0xffff).rw(FUNC(m5_state::cart_window_r),
 		FUNC(m5_state::cart_window_w));
 	m_rom_view[1](0x0000, 0x1fff).rom().region("monitor", 0x0000).unmapw(); // monitor rom
-	m_rom_view[1](0x2000, 0x3fff).bankr(m_bankr[1]); // cart rom
-	m_rom_view[1](0x4000, 0x5fff).bankr(m_bankr[2]); // cart rom
-	m_rom_view[1](0x6000, 0x6fff).bankr(m_bankr[3]); // cart rom
 	m_rom_view[1](0x7000, 0x7fff).ram().share("ram");
 
 	// for RAM only mode
@@ -1403,21 +1400,20 @@ void m5_state::machine_reset()
 			case MEM64KBI:
 			{
 				m_rom_view.select(1);
-				//if AUTOSTART is on then page out cart and start tape loading
-				bool autostart = BIT(m_DIPS->read(), 2);
-				bool ram32k_only = BIT(m_DIPS->read(), 4);
-				if (m_romcart)
-				{	//if not autostart map needed banks to rom cart, rest to ram
-					for (int i = 1; i < 4; ++i)
-					{
-						if ((i - 1) * 0x2000 < romcart_rom_size - 1 && !autostart)
-						{
-							m_bankr[i]->set_base(romcart_rom_base + (i - 1) * 0x2000);
-						}
-						else if (ram32k_only) m_bankr[i]->set_base(m_ignore_writes.get());
-						 else m_bankr[i]->set_base(ramcart_ram_base + (i - 1) * 0x2000);
-					}
+
+				bool autostart = BIT(m_DIPS->read(), 1);
+				bool ram32k_only = BIT(m_DIPS->read(), 2);
+
+				if (m_romcart && !autostart)
+				{
+					m_rom_view[1].install_read_bank(0x2000, 0x2000 + romcart_rom_size - 1, m_bankr[1]);
+					m_rom_view[1].unmap_write(0x2000, 0x2000 + romcart_rom_size - 1);
+					m_bankr[1]->configure_entry(1, romcart_rom_base + 0x0000);
+					m_bankr[1]->set_entry(1);
 				}
+
+				if (ram32k_only)
+					m_rom_view[1].unmap_readwrite(0x2000 + romcart_rom_size, 0x6fff);
 
 				break;
 			}
@@ -1428,7 +1424,6 @@ void m5_state::machine_reset()
 					//replace 64Kbf Basic-F with rom cart, works only with mode 6
 					for (int i = 1; i < 4; i++)
 						m_bankr[i]->configure_entry(2, romcart_rom_base + (i - 1) * 0x2000);
-
 				}
 				// ram entries
 				for (int i = 0; i < 5; i++)
@@ -1490,7 +1485,7 @@ void m5_state::machine_reset()
 				m_bankr[5]->configure_entry(2, ramcart_rom_base + 0x12000);
 
 				//select BASICs or MSX
-				if (BIT(m_DIPS->read(), 1))
+				if (BIT(m_DIPS->read(), 0))
 				{
 					//BASIC-G and BASIC-F
 					entry = 1;


MooglyGuy:
```
../../../../../src/mame/sord/m5.cpp:1092:23: error: 'void {anonymous}::m5_state::fdc_irq_timer_cb(osd::s32)' defined but not used [-Werror=unused-function]
 1092 | TIMER_CALLBACK_MEMBER(m5_state::fdc_irq_timer_cb)
      |                       ^~~~~~~~
../../../../../src/emu/schedule.h:25:46: note: in definition of macro 'TIMER_CALLBACK_MEMBER'
   25 | #define TIMER_CALLBACK_MEMBER(name)     void name(s32 param)
      |                                              ^~~~
```

You should probably fix this.

src/mame/sord/m5.cpp
```diff
@@ -4,39 +4,49 @@
 
     Sord m.5
 
-    http://www.retropc.net/mm/m5/
-    http://www.museo8bits.es/wiki/index.php/Sord_M5 not working
```
plaes, line 8: These links still work:
http://www.retropc.net/mm/m5/
https://web.archive.org/web/20180817030621/http://www.museo8bits.es/wiki/index.php/Sord_M5
https://web.archive.org/web/20120602100302/http://k5.web.klfree.net/content/view/10/11/

dlabi: m5.arigato.cz is successor of old k5.web.klfree.net and contains more info as all these deleted links. So there is no need to keep them here.

src/mame/sord/m5.cpp
```diff
@@ -704,145 +588,108 @@ void m5_state::mem64KBF_w(u8 data) //out 0x30
...
 	{
 		case 0:
-			program.unmap_write(0x0000, 0x6fff);
-			membank("bank1r")->set_base(memregion("maincpu")->base());
-			membank("bank2r")->set_base(m_cart_rom->base());
-			membank("bank3r")->set_base(m_cart_rom->base()+0x2000);
-			membank("bank4r")->set_base(m_cart_rom->base()+0x4000);
-			membank("bank5r")->set_base(ram_region->base()+0x8000);     membank("bank5w")->set_base(ram_region->base()+0x8000);
-			membank("bank6r")->set_base(ram_region->base()+0xc000);     membank("bank6w")->set_base(ram_region->base()+0xc000);
+			membank("bank1r")->set_entry(1); membank("bank1w")->set_entry(1);
```
galibert: Given you're modernizing, you should turn all of those into finders.

dlabi: I belived I've modernized it :-(  I will try to make it better but if you dont mind, could you please give me brief example how it should be?
are you refering to this? https://docs.mamedev.org/techspecs/object_finders.html 
thank you

galibert: Yes, search the code for uses of "required_memory_bank" too, it will give you examples of use.  membank("name") takes a noticeable amount of time to find the bank associated to the string.  Using a finder ensures that resolution is done only once at startup.

src/mame/sord/m5.cpp
plaes: Hum.. seems like m5.arigato.cz is currently dead (hopefully temporary) :(

Could at least add the Museo8bits.es via archive.org. This site looked quite informative.

dlabi: you are right m5.arigato.cz is currently dead, I've asked it's owner if it's temporary but I didn't get clear answer so I'll use archive.org for it too.

angelosa:
This PR has conflicts in hash files

src/mame/sord/m5.cpp
dlabi: I've replaced all named banks to arrays. Please take a look

einstein95:
Not sure why you've got a few files added under `3rdparty/luasocket/.vs/`, possibly VS Code is adding stuff

dlabi:
> Not sure why you've got a few files added under `3rdparty/luasocket/.vs/`, possibly VS Code is adding stuff

ups. that wasnt intended ;-/

angelosa:
That sounds Visual Studio, not Visual Studio Code. I guess this needs a stash in a different PR, and an actual .gitignore for any .vs folder

src/mame/sord/m5.cpp
dlabi: added back Museo8bits.es and m5.arigato.cz via archive.org

.gitignore
```diff
@@ -49,3 +49,4 @@ regtests/chdman/temp
 regtests/jedutil/output
 /CMakeLists.txt
 /samples
+/plugins/retsync
```
cuavas: Please remove this.

hash/m5_cart.xml
```diff
@@ -173,8 +173,8 @@ and why some of the dumps below have weird size?
 		<info name="serial" value="8" />
 		<info name="alt_title" value="ガッタンゴットン" />
 		<part name="cart" interface="m5_cart">
-			<dataarea name="rom" size="8192">
-				<rom name="guttang gottong (1981)(konami)(jp).bin" size="8192" crc="04f842ea" sha1="e5d328fa2765530849422eb3ae33628d2adf4ca9" offset="00000" />
+			<dataarea name="rom" size="9216">
+				<rom name="guttang gottong (1981)(konami)(jp).bin" size="9216" crc="4fcb8e75" sha1="976cb6b260d1276cf0ff01c62ba11c56420d0856" offset="00000" />
```
cuavas, line 177: Does this cartridge really contain an 8K ROM chip and a 1K ROM chip?

hash/m5_cass.xml
```diff
@@ -264,58 +265,58 @@ license:CC0-1.0
 		</part>
 	</software>
 
-	<software name="ssf" supported="no">
+	<software name="ssf" supported="yes">
 		<description>Solar System Forces</description>
-		<year>19??</year>
-		<publisher>&lt;unknown&gt;</publisher>
+		<year>1983</year>
+		<publisher>&lt;Brainhill Ltd. Tokyo Japan&gt;</publisher>
```
cuavas: The angle brackets in the publisher are to indicate that “<unknown>” is descriptive, i.e. the publisher is unknown.  If the publisher doesn’t actually use angle brackets, you shouldn’t leave them in.

src/devices/bus/m5/slot.cpp
```diff
@@ -48,7 +48,7 @@ void device_m5_cart_interface::rom_alloc(uint32_t size, const char *tag)
 {
 	if (m_rom == nullptr)
 	{
-		m_rom = device().machine().memory().region_alloc(std::string(tag).append(M5SLOT_ROM_REGION_TAG), size, 1, ENDIANNESS_LITTLE)->base();
+		m_rom = device().machine().memory().region_alloc(std::string(tag).append(":rom"), size, 1, ENDIANNESS_LITTLE)->base();
```
cuavas: Please don’t preserve the worst of stuff – don’t pass the tag around, and just use `subtag("rom")` if you want the tag `rom` relative to yourself.  I have no clue why this horrible piece of code has been copy/pasted so widely.

src/devices/bus/m5/slot.h
```diff
@@ -91,6 +94,33 @@ class m5_cart_slot_device : public device_t,
 	virtual std::string get_default_card_software(get_default_card_software_hook &hook) const override;
 
 	int get_type() { return m_type; }
+	device_m5_cart_interface* get_loaded_cart() const noexcept { return m_cart; }
```
cuavas: You shouldn’t need this – `device_single_card_slot_interface::get_card_device()` does the same thing.

src/mame/sord/m5.cpp
```diff
@@ -4,39 +4,57 @@
...
+	- fd5 floppy emulation works but only with rom hack
+	- updated Sord m5 www links
+	- added fd5 utility disk to software list - not original dump, made from program listings
+	- added support of optional sram in Basic-F and Basic-G cartridges. Works only if shortname(softlist) is used
+	- marked 32/64KB RAM expansions EM-5, 64KBI, 64KBF, 64KRX as supported in sofwarelist
+	- reenabled and refactored memory banking
+
+	- brno mod: switched to rom including basic-i
+	- brno mod: reenabled and refactored memory banking 
+
```
cuavas: No new changelogs in source files.  That’s what VCS history is for.

src/mame/sord/m5.cpp
```diff
@@ -300,6 +318,13 @@ Few other notes:
 #include "formats/m5_dsk.h"
 #include "formats/sord_cas.h"
 
+#define FD5_DEBUG 1
+#define BRNO_DEBUG 1
+#define VERBOSE 0
+
+#include "logmacro.h"
+
+#define FDC_MOTOR_TIMEOUT 3
```
cuavas: Please don’t use preprocessor macros where a `constexpr` will work.

src/mame/sord/m5.cpp
```diff
@@ -354,20 +385,23 @@ class m5_state : public driver_device
 	virtual void machine_start() override ATTR_COLD;
 	virtual void machine_reset() override ATTR_COLD;
 
+	TIMER_CALLBACK_MEMBER(FDC_MOTOR_TIMEOUT_cb);
+	emu_timer *m_motor_timer = nullptr;
```
cuavas, line 351: Please don’t intermix data members and member functions.  Put the timer with other siimilar things.

src/mame/sord/m5.cpp
```diff
@@ -704,145 +598,103 @@ void m5_state::mem64KBF_w(u8 data) //out 0x30
...
-			membank("bank3r")->set_base(m_cart_rom->base()+0x2000);
-			membank("bank4r")->set_base(m_cart_rom->base()+0x4000);
-			membank("bank5r")->set_base(ram_region->base()+0x8000);     membank("bank5w")->set_base(ram_region->base()+0x8000);
-			membank("bank6r")->set_base(ram_region->base()+0xc000);     membank("bank6w")->set_base(ram_region->base()+0xc000);
+			for (int i = 0; i < 4; i++)
+			{
+				m_bankr[i]->set_entry(1); m_bankw[i]->set_entry(1);
+			}
+			m_bankr[4]->set_entry(0); m_bankw[4]->set_entry(0);
+			m_bankr[5]->set_entry(0); m_bankw[5]->set_entry(0);
```
cuavas: Please just put one statement per line.  It’s less likely to be misread.

src/mame/sord/m5.cpp
```diff
@@ -704,145 +598,103 @@ void m5_state::mem64KBF_w(u8 data) //out 0x30
...
-			membank("bank1r")->set_base(ram_region->base()+0x0000);
-			membank("bank2r")->set_base(ram_region->base()+0x2000);
-			membank("bank3r")->set_base(ram_region->base()+0x4000);     membank("bank3w")->set_base(ram_region->base()+0x4000);
-			membank("bank4r")->set_base(ram_region->base()+0x6000);     membank("bank4w")->set_base(ram_region->base()+0x6000);
-			membank("bank5r")->set_base(ram_region->base()+0x8000);     membank("bank5w")->set_base(ram_region->base()+0x8000);
-			membank("bank6r")->set_base(ram_region->base()+0xc000);     membank("bank6w")->set_base(ram_region->base()+0xc000);
+			m_bankr[0]->set_entry(0); m_bankw[0]->set_entry(1);
+			m_bankr[1]->set_entry(0); m_bankw[1]->set_entry(1);
+			for (int i = 2; i < 4; i++)
+				m_bankr[i]->set_entry(0), m_bankw[i]->set_entry(0);
```
cuavas: This isn’t even consistent with the others, that’s liable to cause confusion.

src/mame/sord/m5.cpp
```diff
@@ -877,20 +745,21 @@ void m5_state::m5_io(address_map &map)
...
-	map(0x36, 0x36).mirror(0x08).portr("Y6");
-	map(0x37, 0x37).mirror(0x08).portr("JOY");
+	map(0x30, 0x30).mirror(0x00).portr("Y0").w(FUNC(m5_state::mem64KBF_w)); // 64KBF paging
+	map(0x31, 0x31).mirror(0x00).portr("Y1");
+	map(0x32, 0x32).mirror(0x00).portr("Y2");
+	map(0x33, 0x33).mirror(0x00).portr("Y3");
+	map(0x34, 0x34).mirror(0x00).portr("Y4");
+	map(0x35, 0x35).mirror(0x00).portr("Y5");
+	map(0x36, 0x36).mirror(0x00).portr("Y6");
+	map(0x37, 0x37).mirror(0x00).portr("JOY");
```
cuavas: Are you sure these aren’t mirrored?  I would have thought whoever wrote it wouldn’t have put the mirroring in explicitly if that address bit was decoded.

src/mame/sord/m5.cpp
```diff
@@ -1030,102 +899,227 @@ INPUT_PORTS_END
 //  I8255 Interface
 //-------------------------------------------------
 
+// 0x70
 u8 m5_state::ppi_pa_r()
 {
-	return m_fd5_data;
+		return m_fd5_data;
 }
```
cuavas, line 908: Indentation messed up.

src/mame/sord/m5.cpp
```diff
@@ -300,6 +318,13 @@ Few other notes:
 #include "formats/m5_dsk.h"
 #include "formats/sord_cas.h"
 
+#define FD5_DEBUG 1
+#define BRNO_DEBUG 1
+#define VERBOSE 0
```
cuavas: Please use `LOGMASKED` as intended rather than putting `if` around `LOG`.  Look at the idiomatic way things are done.

src/mame/sord/m5.cpp
```diff
@@ -1309,124 +1315,208 @@ void m5_state::machine_start()
...
-	// no cart inserted - there is nothing to do - not allowed in original Sord m5
-	if (m_cart_ram == nullptr && m_cart == nullptr)
+	// no cart inserted - there is nothing to do - Sord m5 won't power up
+	// but some owners bypassed this by shorting some lines on the motherboard and Sord power up and starts cassette loading
+	if (m_ramcart == nullptr && m_romcart == nullptr)
 	{
-	//  membank("bank1r")->set_base(memregion("maincpu")->base());
 		program.unmap_write(0x0000, 0x1fff);
-	//  program.unmap_readwrite(0x2000, 0x6fff); //if you uncomment this line Sord starts cassette loading but it is not correct on real hw
+		program.unmap_readwrite(0x2000, 0x6fff); //if you uncomment this line Sord starts cassette loading but it is not correct on vanilla Sord m5. see above comment
```
cuavas: This hack should at least be behind a machine configuration switch.

src/mame/sord/m5.cpp
```diff
@@ -1309,124 +1315,208 @@ void m5_state::machine_start()
...
 		program.unmap_readwrite(0x8000, 0xffff);
 		return;
 	}
 
+	if (m_romcart) {
+		romcart_rom_size = m_romcart->get_rom_size();
+		romcart_rom_base = m_romcart->get_rom_base();
+	}
```
cuavas: Brace positioning inconsistent with the rest of the file.

src/mame/sord/m5.cpp
```diff
@@ -1309,124 +1315,208 @@ void m5_state::machine_start()
...
-
+			{
+				m_rom_view.select(1);
 				//if AUTOSTART is on then page out cart and start tape loading
-				if (m_cart && ((m_DIPS->read() & 2) != 2))
-				{
-					program.install_read_handler(0x2000, 0x3fff, read8sm_delegate(*m_cart, FUNC(m5_cart_slot_device::read_rom)));
-					program.unmap_write(0x2000, 0x3fff);
+				bool autostart = (m_DIPS->read() & 2) == 2;
+				bool ram32k_only = (m_DIPS->read() & 4) == 4;
```
cuavas: Just use the `BIT` helper for testing bits.

src/mame/sord/m5.cpp
```diff
@@ -1309,124 +1315,208 @@ void m5_state::machine_start()
...
+				{	//if not autostart map needed banks to rom cart, rest to ram
+					for (int i = 1; i < 4; ++i) {
+						if ((i - 1) * 0x2000 < romcart_rom_size - 1 && !autostart)
+						{
+							m_bankr[i]->set_base(romcart_rom_base + (i - 1) * 0x2000);
+						}
+						else if (ram32k_only) m_bankr[i]->set_base(m_ignore_writes.get());
+						 else m_bankr[i]->set_base(ramcart_ram_base + (i - 1) * 0x2000);
+					}
 				}
```
cuavas: Inconsistent braces and indentation.

Also, if you’re using `set_base` on a memory bank, you’re probably doing it wrong.  You should use configured banking.  And sending ignored writes to a magic buffer isn’t how we do things.  You should actually unmap writes in that situation.

src/mame/sord/m5.cpp
```diff
@@ -1309,124 +1315,208 @@ void m5_state::machine_start()
...
+					if (romcart_ram_size > 0) {
+						uint32_t map_start = 0x8000;
+						uint32_t map_end = 0x8000 + romcart_ram_size - 1;
+						if (map_end > 0xffff) map_end = 0xffff;
+						program.install_ram(map_start, map_end, romcart_ram_base);
+					}
+					else program.unmap_readwrite(0x8000, 0xffff);
+				}
+				else program.unmap_readwrite(0x8000, 0xffff);
+			//}
```
cuavas: More weird indentation and bracing.

src/mame/sord/m5.cpp
```diff
@@ -1602,7 +1689,7 @@ void brno_state::brno(machine_config &config)
 	SCREEN(config, "screen", SCREEN_TYPE_RASTER);
 
 	// RAM disk (maximum is 1024kB 256x 4kB banks)
-	RAM(config, m_ram).set_default_size("512K").set_extra_options("256K,768K,1M");
+	RAM(config, m_ram).set_default_size("1M").set_extra_options("256K,768K,1M");
```
cuavas: This causes it to lose the 512K size.  Also, the default should really be a representative configuration, not absolute maximum.

.gitignore
dlabi: removed

hash/m5_cart.xml
dlabi: sounds ridiculous I know but 8k rom (with exact same checksums in current version of MAME) doesn't work for me even with 32kb ram. My dump has ~0x50 extra bytes and because Sord rejects ROMs not aligned to 1k there is also padding. Need to investigate more

src/mame/sord/m5.cpp
dlabi: not sure how to implement this

src/mame/sord/m5.cpp
dlabi: not sure where similar things are?

hash/m5_cass.xml
dlabi: removed

src/devices/bus/m5/slot.h
dlabi: replaced

src/mame/sord/m5.cpp
dlabi: removed

src/mame/sord/m5.cpp
dlabi: done

src/mame/sord/m5.cpp
dlabi: reverted back

src/mame/sord/m5.cpp
dlabi: done

src/mame/sord/m5.cpp
dlabi: reformated

src/mame/sord/m5.cpp
dlabi: done

src/mame/sord/m5.cpp
dlabi: reformated

src/mame/sord/m5.cpp
dlabi: reverted back

src/mame/sord/m5.cpp
dlabi: fixed

src/mame/sord/m5.cpp
dlabi: fixed

src/mame/sord/m5.cpp
dlabi: done

src/devices/bus/m5/slot.cpp
dlabi: replaced by subtag

src/mame/sord/m5.cpp
dlabi: I honestly don't know how to do it better(for 64kbf, 64krx). As far as I know it is not possible unmap write for a membank in certain configure entry. In another mode this same membank need rw again so one need to install bank again over and over. I saw this method using dummy buffer in other drivers too so it seemed to me as functional workaround. If there is cleaner way please advice

hash/m5_cart.xml
dlabi: obrazek
game realocates itself to ram at A000, it use length of 0x2100 what makes me think there were >8k rom. It is uncommon indeed

PR #14492, opened 2025-11-08
rare/kinst.cpp: Fix metadata

cam900:
Nintendo has co-copyright of games in these drivers, until Rare get copyright alone.
reference: https://flyers.arcade-museum.com/videogames/show/525

### DIFF
--- a/src/mame/rare/kinst.cpp
+++ b/src/mame/rare/kinst.cpp
@@ -862,6 +862,6 @@ void kinst_state::init_kinst2()
 // versions selectable by changing bioses
 
 //    YEAR  NAME      PARENT  MACHINE   INPUT   CLASS           INIT         SCREEN  COMPANY                  FULLNAME           FLAGS
-GAME( 1994, kinst,    0,      kinst,    kinst,  kinst_state,    init_kinst,  ROT0,   "Nintendo / Rare (Midway license)", "Killer Instinct", MACHINE_SUPPORTS_SAVE )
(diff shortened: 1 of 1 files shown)


happppp:
What did Nintendo do? The best I can see online is that they sold a license to Midway's parent company to allow them to use N64-based ("Ultra 64") arcade hardware. The only similar hardware here is the CPU? Should they really be in the company string, or is it like having an Manga/Anime publishing company in a Dragon Ball game just for licensing the right of the characters?

Anyway, if they must be in the company string, do Rare / Nintendo instead, eg. user sorts frontend by company, the game shows up under Rare, not Nintendo.

cam900:
kinst Attract mode shows 'Nintendo presents', after Rare logo scene and before 'Ultra 64' scene. kinst2 has similar scene in attract mode.(It shows Nintendo logo before Rare logo)
I think Nintendo was licensor for kinst/2.
Reference: https://en.wikipedia.org/wiki/Killer_Instinct_(1994_video_game)

rb6502:
The official credit is "Presented by Nintendo" and Nintendo of America's Ken Lobb had design input on both arcade KIs (and did voice acting and balance testing).  In the industry you'd generally break that down as Nintendo is the publisher and Rare's the developer.

Existing MAME practice leans toward only crediting the publisher in that scenario (daisenpu is developed by Toaplan and published by Taito, and only Taito is credited, and the same with NMK on the Jaleco Mega System 1 games) but my preference is to credit both.

happppp:
Midway is the publisher here though, anyway I take your word for it that Nintendo was more involved than I thought, and not just the for the branding and hardware rights.

PR #14494, opened 2025-11-08
Fix led 30 layout issues.

flama12333:
I had to change uint8_t to 7 at ay2_port_b_w.
and output_finder m_leds to 31.


src/mame/misc/marywu.cpp
```diff
@@ -190,6 +196,7 @@ void marywu_state::marywu(machine_config &config)
 	maincpu.set_addrmap(AS_DATA, &marywu_state::data_map);
 	//TODO: figure out what each bit is mapped to in the 80c31 ports P1 and P3
 	maincpu.port_in_cb<1>().set_ioport("P1");
+	maincpu.port_out_cb<3>().set(FUNC(p3_port_w));
```
ajrhacker: This is nonstandard syntax permitted by some compilers. A qualified name should be used here.

ajrhacker (changes requested):

ajrhacker:
The point still stands. To be specific, `FUNC(p3_port_w)` should be changed to `FUNC(marywu_state::p3_port_w)`.

```
../../../../../src/mame/misc/marywu.cpp:200:31: error: must explicitly qualify name of member function when taking its address
  200 |         maincpu.port_out_cb<3>().set(FUNC(p3_port_w));
      |                                      ^    ~~~~~~~~~
      |                                           marywu_state::
```

src/mame/misc/marywu.cpp
```diff
@@ -134,7 +134,7 @@ void marywu_state::ay2_port_b_w(uint8_t data)
 
 void marywu_state::p3_port_w(uint8_t data) // 1 led are used.
 {
-	for (uint8_t i = 0; i < 8; i++)
+	for (uint8_t i = 0; i < 0; i++)
```
ajrhacker: This causes the loop to be executed zero times, defeating the whole purpose of adding this as a callback.

Since external memory is hooked up to the MOVX space, it's unlikely P36 and P37 are used to drive LEDs, but some of the lower 6 bits might be.

flama12333:
Currently in draft.

flama12333:
I had to remove  p3_port_w due to an mistake without checking something.

PR #14498, opened 2025-11-09
Add software lists for the Ensoniq VFX-SD, SD-1 and SD-1/32 keyboards.

cbrunschen:
This PR builds on the floppy disk functionality improvements in https://github.com/mamedev/mame/pull/14444 and adds the Sequencer OS floppy images for each major version, as well as some demo disks.

Each can also use the floppy disks for its predecessor, mainly because
there are demo songs on the earlier OS disks that are not on the later ones,
but which still work perfectly well and sounds great.

The SD-1/32 software list also includes four demo disks:

1. A demo disk specifically for the SD-1/32

2. A demo disk for the SQ-series of keyboards with 32 voices
   This contains System Exclusive files foro use with 32-voice SQ series
   keyboards

3. Two multi-product demo disks
   These contain both sequencer files for the SD-1/32 and System
   Exclusive files for use with the SQ and KS series of keyboards.

### DIFF
--- a/hash/sd132_flop.xml
+++ b/hash/sd132_flop.xml
@@ -5,7 +5,7 @@ license:CC0-1.0
 -->
 <softwarelist name="sd132_flop" description="Ensoniq SD-1 32 Voice Disk Images">
 
-<software name="os400">
+<software name="os400" supported="yes">
 	<description>SD-1 Sequencer OS version 4.00</description>
 	<year>1991</year>
 	<publisher>Ensoniq</publisher>
@@ -18,7 +18,7 @@ license:CC0-1.0
 
 </software>
 
-<software name="os410">
+<software name="os410" supported="yes">
 	<description>SD-1 Sequencer OS version 4.10</description>
 	<year>1992</year>
 	<publisher>Ensoniq</publisher>
@@ -31,7 +31,7 @@ license:CC0-1.0
 
 </software>
 
-<software name="sd1_32_demo1">
+<software name="sd1_32_demo1" supported="yes">
 	<description>SD-1 32 Voice Demo Disk #1</description>
 	<year>1992</year>
 	<publisher>Ensoniq</publisher>
@@ -44,7 +44,7 @@ license:CC0-1.0
 
 </software>
 
-<software name="multi_demo1">
+<software name="multi_demo1" supported="yes">
 	<description>Multi-Product Demo Disk #1</description>
 	<year>1992</year>
 	<publisher>Ensoniq</publisher>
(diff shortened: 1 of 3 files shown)


hash/sd132_flop.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="sd132_flop" description="Ensoniq SD-1 32 Voice Disk Images">
+
+<software name="os400">
```
rb6502: Add `supported="yes"` or `"no"` to the `<software>` tag for each item to indicate if it's known working or not.  I completely forgot this for the original version of the MPC3000 list recently and it caused a minor issue with the release.

hash/sd1_flop.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
```
rb6502, line 1: I'm undecided on if it makes sense here, but you can have a single software list for a family of machines and use the `<sharedfeat name="compatibility">` tag, something like this:

`<sharedfeat name="compatibility" value="vfxsd,sd1,sd132" />`

Then in the driver on the `SOFTWARE_LIST()` config item you'd append

`.set_filter("sd1");` to mark a machine as able to use items with the `sd1` tag.

I use that feature extensively for the mac_cdrom.xml list.

hash/sd132_flop.xml
cbrunschen: Done.

Though, [the documentation](https://docs.mamedev.org/contributing/softlist.html) claims that

> **supported (optional)**
> One of the values `yes` (fully usable in emulation), `no` (not usable in emulation), or `partial` (usable in emulation with limitations). If the attribute is not present, it is equivalent to `yes`. 

Is that not accurate?

hash/sd1_flop.xml
cbrunschen: I saw that in some other software lists. But the nice thing about having multiple software lists is that, when choosing an item from a software list in the File Manager, the user first has to choose the software list from which to choose a floppy image:

Screenshot From 2025-11-10 07-04-50

This choice of software list seems to always happen, even if there's only a single software list:

image

So keeping them in different lists lets the user see that grouping, which I think is useful.

hash/sd132_flop.xml
cbrunschen: [softwarelist.dtd](https://github.com/mamedev/mame/blob/master/hash/softwarelist.dtd#L8) seems to corroborate this:

>		<!ATTLIST software supported (yes|partial|no) "yes">

cbrunschen:
Thank you!

PR #14500, opened 2025-11-09
Mark the Ensoniq SD-1/32 as working.

cbrunschen:
This PR: 
1. removes the markers of "imperfect sound" from the es5510 (and the related " esq_5505_5510_pump that really only ensures that the es5510 processes the samples generated by an es5505)

2. updates the 21-voice VFX family of keyboards to most have "imperfect timing" markers, since they do exhibit timing strangeness; but since the SD-1/32, the 32-voice variant, does not have those timing issues, marks that simply as working

4. marks the sd1 (21 voice) as a clone of the sd132 - because the sd132 is the most complete variant, and also the one that is working, so that is the one users should be guided towards

It does these things because, in my testing, the emulated sd132 sounds completely like the real one that I have; and now that it has working floppy and cartridge support, and is able to run the sequencer OS, and has layouts that give access to the full experience including all buttons, sliders, wheels, and even a full working keyboard, it really is a fully usable, working emulation of an Ensoniq SD-1/32 .

(Due to the timing issues, the 21-voice keyboards of course can't be marked as working, because they're not.)

And as it is the most complete, and working, member of the family, I think the sd132 should be marked as the "parent", and the not-quite-working sd1 (21 voice) as the "clone", so that users can more easily find the better, working version.

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -1242,13 +1242,13 @@ void esq5505_state::init_denib()
 } // Anonymous namespace
 
 
-CONS( 1988, eps,    0,   0, eps,   vfx, esq5505_state, init_eps,    "Ensoniq", "EPS",             MACHINE_NOT_WORKING )       // custom VFD: one alphanumeric 22-char row, one graphics-capable row (alpha row can also do bar graphs)
-CONS( 1989, vfx,    0,   0, vfx,   vfx, esq5505_state, init_denib,  "Ensoniq", "VFX",             MACHINE_IMPERFECT_TIMING )  // 2x40 VFD
-CONS( 1989, vfxsd,  0,   0, vfxsd, vfx, esq5505_state, init_denib,  "Ensoniq", "VFX-SD",          MACHINE_IMPERFECT_TIMING )  // 2x40 VFD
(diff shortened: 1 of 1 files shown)


cbrunschen:
For a comparison, I've played the same song - "32OVERVIEW" from the "Multi-Product Demo Disk #2" that was just added to the software list for the SD-1/32 - both in MAME and on a real SD-1/32. This "song" is comprised of four different little sequences, each with its own different effects program, including flanger, delay, distortion, chorus and different reverbs.

These are the results, both with the only post-processing to bring them to the same apparent volume level (and as MP3 files because that's what Github will accept):

[32overview_a.mp3](https://github.com/user-attachments/files/23480970/32overview_a.mp3)

[32overview_b.mp3](https://github.com/user-attachments/files/23480971/32overview_b.mp3)

I couldn't tell you which is which just from listening, but I would love to hear other people's opinion. Can you hear a significant difference? Is there something that sounds wrong in either recording?

rb6502:
B seems slightly louder but I don't hear any difference in the timbres or the effects processing.

cbrunschen:
> B seems slightly louder but I don't hear any difference in the timbres or the effects processing.

Thank you for listening and checking! I may well not have managed to get the two recordings to exactly the same volume level (I clearly need to learn to use Audacity better).

But that seems to corroborate my thinking, that the effect processing is clearly working as it should, since it results in sound that is not distinguishable between the real thing and the emulation.

With that working, the floppy and cartridge as well as the sequencer, and layouts that give full access to the panel and keyboard including velocity and aftertouch, I think that the sd132 should be considered working.

(The 21-voice keyboards must of course be noted for their timing issues, but apart from those, they actually also work - but of course, timing is a very important part of music, so that makes them not really fully usable.)

cbrunschen:
Thank you!

PR #14502, opened 2025-11-10
Add internal layout for Access Virus A / Declare inputs for other virus synths / Misc. improvements to the driver

felipesanches:
* Declare MIDI in/out ports
* Declare buttons and knobs for Virus C, CL, Rack and RackXL
* Restructure the driver to avoid excessive repetition of machine config declarations
* Declare a second version of the Virus C ROM


felipesanches:
And this is how the Layout for Access Virus A added in this PR looks:
Screenshot From 2025-11-09 21-49-13

PR #14504, opened 2025-11-10
esq5505.cpp: Also support HFE format flopppy images.

cbrunschen:
[As discussed](https://github.com/mamedev/mame/pull/14444#discussion_r2508227784), HFE floppy images are often used for synthesizer floppy images. Add support for using these in esq5505.cpp synthesizers. 

I verified that this works for the sd132 with an image file converted from `.img` to` .hfe` , loading the image into the HxC software with _Load Raw Image_ and choosing the _Ensoniq EPS 3"5 DD Floppy Disk_  format.


cbrunschen:
Thank you!

PR #14506, opened 2025-11-10
new WORKING clones - M.I.A. - Missing in Action (Data East, 12/22/88, US, prototype?)

mamehaze:
new WORKING clones
--------
M.I.A. - Missing in Action (Data East, US, 12/22/88, prototype?) [Mann of Wor, RetroCJT]


mamehaze:
this is from the thread at KLOV
https://forums.arcade-museum.com/threads/m-i-a-1988-data-east-bloody-wolf-lost-prototype-archival-thread.557358/#post-5034209

currently draft as I need to find out if these credits are correct, and I can't ask directly as the forum no longer accepts new registrations.

PR #14508, opened 2025-11-11
current Plug and Play work (SPG/GPL cleanups)

mamehaze:
- split a Generalplus Renderer off from the SunPlus renderer, and reduced the SunPlus renderer to just the features supported by the SPG2xx.  While the GPL does have similar features, it makes too many small changes to cleanly implement it as a class on top of the SPG renderer, and having it sitting in the same files makes the SPG code less clear.  [David Haywood]

### DIFF
--- a/src/devices/machine/spg_renderer.cpp
+++ b/src/devices/machine/spg_renderer.cpp
@@ -4,7 +4,7 @@
 #include "emu.h"
 #include "spg_renderer.h"
 
-DEFINE_DEVICE_TYPE(SPG_RENDERER, spg_renderer_device, "spg_renderer", "SunPlus / GeneralPlus video rendering")
+DEFINE_DEVICE_TYPE(SPG_RENDERER, spg_renderer_device, "spg_renderer", "SunPlus SPG2xx video rendering")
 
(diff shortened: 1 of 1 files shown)


mamehaze:
this should get things into a position where I can start to improve the GPL rendering a bit without making a mess of the older SPG implementation.

the 2 VT sets added both boot, but are left as NOT WORKING as I haven't verified the games contained within work well enough yet  https://bsky.app/profile/mamehaze.bsky.social/post/3m5c74xrjh222

mamehaze:
@ajrhacker can you give this a quick look over? I can't really move forward with improving the GPL renderer or other components of those systems until this part is in.

ajrhacker:
I think the SPG/GPL refactoring and the VT updates should really be two separate pull requests. The changes just aren't logically related, and I'll need time to look at the latter.

mamehaze:
Well I work on them together as there's every chance the next TE dump could be a SunPlus / GeneralPlus one relying on the changes, and it's my current non-arcade branch, but I'll split them out this time

ajrhacker:
g3_800 seems to use the same 4x16MB banking scheme as fapocket, though of course without the fake cartridge select input.

ajrhacker:
You might also want to update the SPG_RENDERER device name (line 7 of the file).

PR #14511, opened 2025-11-11
m24_kbd.cpp: Add support for proprietary mouse

donohoe00:
- Add support for the proprietary mouse which plugs into the M24 / 6300 keyboard.  The Logich Serial mouse, which is populated in serport1 by default, need to be removed for this to work (based this on what the Amstrad PC1512 does).
- Make 'hdc' the default device in the first ISA slot (as per drivers for other IBM and clone machines).
- Mark 6300 Plus as Working

### DIFF
--- a/src/mame/olivetti/m24_kbd.cpp
+++ b/src/mame/olivetti/m24_kbd.cpp
@@ -264,8 +264,7 @@ uint8_t m24_keyboard_device::p2_r()
 	// respect to the other, to produce the required "quadrature" encoding,
 	// which lets the MCU determine the direction of movement.
 	return (m_keypress << 7) | m_mousebtn->read() |
-	  ((mx+1) & 2) >> 1 | (mx & 2) |
-	  (my & 2) << 1 | ((my+1) & 2) << 2;
+      BIT(mx+1, 1) | BIT(mx, 1) << 1 | BIT(my, 1) << 2 | BIT(my+1, 1) << 3;
 }
 
 int m24_keyboard_device::t0_r()


src/mame/olivetti/m24_kbd.cpp
```diff
@@ -248,7 +256,16 @@ void m24_keyboard_device::p1_w(uint8_t data)
...
+	uint8_t mx = m_mousex->read();
+	uint8_t my = m_mousey->read();
+
+	// Generate appropriate square waves in response to changing mouse
+	// co-ordinates.  The +1 is to put one signal 90 degrees out of phase with
+	// respect to the other, to produce the required "quadrature" encoding,
+	// which lets the MCU determine the direction of movement.
+	return (m_keypress << 7) | m_mousebtn->read() |
+	  ((mx+1) & 2) >> 1 | (mx & 2) |
+	  (my & 2) << 1 | ((my+1) & 2) << 2;
```
angelosa: Isn't this just: `((mx & 3) ^ 1) | ((my & 3) ^ 1) << 2)`, roughly?

donohoe00: I think there have to be additions.  This version works and is probably less confusing:

      BIT(mx+1, 1) | BIT(mx, 1) << 1 | BIT(my, 1) << 2 | BIT(my+1, 1) << 3;

donohoe00: To explain what's going on: each signal comes from an optical sensor, which is triggered by a rotating wheel for a given axis.  If the mouse is moved at a constant speed along one axis, a square wave with a 50% duty cycle will be generated by the optical sensor.  However, there are two optical sensors for each axis (hence 2 signals per axis) and the second sensor is offset slightly from the first, such that the phase of the second signal is offset by 90 degrees with respect to the first.  I'm using bit 1 of the mouse coordinate that comes from the input system to generate one the square waves, and and bit 1 of [the coordinate plus one] to generate the square wave that's offset.

rb6502: That's standard quadrature signalling, and most mice work that way internally.   Data-entry knobs are often a single-axis instance, such as https://github.com/mamedev/mame/blob/ccdef41b3e3aba7a5f3979ec1ddcf0d52d48d608/src/mame/akai/mpc3000.cpp#L298-L335

donohoe00: Is there any action needed here?  The mouse support seems to work quite well (I successfully used it to lose a game of Reversi under Windows 1.03).  To my mind the theory is sound and the implementation is correct.  The keyboard MCU firmware is interpreting the signals correctly and generating the right pseudo scan codes and sending them to the motherboard MCU.  I could perhaps make the code more readable, or add a more explanatory comment...

rb6502: The "less confusing version" with BIT() that you referenced earlier would be good to include here, because it really is a lot more readable.

donohoe00: OK, Done!

src/mame/olivetti/m24_kbd.cpp
```diff
@@ -248,7 +256,15 @@ void m24_keyboard_device::p1_w(uint8_t data)
...
-	return (m_keypress << 7) | m_mousebtn->read();
+	uint8_t mx = m_mousex->read();
+	uint8_t my = m_mousey->read();
+
+	// Generate appropriate square waves in response to changing mouse
+	// co-ordinates.  The +1 is to put one signal 90 degrees out of phase with
+	// respect to the other, to produce the required "quadrature" encoding,
+	// which lets the MCU determine the direction of movement.
+	return (m_keypress << 7) | m_mousebtn->read() |
+      BIT(mx+1, 1) | BIT(mx, 1) << 1 | BIT(my, 1) << 2 | BIT(my+1, 1) << 3;
```
cuavas, line 267: This isn’t going to behave well if the user moves the mouse more than one unit in either direction during a frame interval.  MAME only updates input values once per frame interval, so for things like this you want to accumulate how much the mouse moves, and feed it to the MCU at a rate it can handle (e.g. see how the Apple II mouse card does things, although it has some additional logic between the quadrature inputs and its microcontroller).

donohoe00: Understood, I'll fix it up.  I wasn't seeing deltas greater than +/-1, but according to https://cpstest.org/polling-rate-test, the mouse I'm using tops out at around 45Hz!  I'm also running Linux/X11, so that could be a factor.

PR #14513, opened 2025-11-11
iteagle.cpp - Add bootrom variant

FakeShemp:
Many years ago I bought an update kit for Golden Tee Fore 2002. It came with a replacement bootrom labeled "GOLF FORE U15 V1.02". I just dumped it and noticed that it was different from the one in MAME.

Here's the thing though. When I try the 1.02 already in mame with gtfore02, it fails at bootup with checksum failures. My dump does boot though. Could it be a bad dump or are there some other shenanigans?

I've added my dump as an alt for now.


rb6502:
How extensive are the differences?  If it's just a few bytes it's likely the existing dump is bitrotted or a bad dump and can be replaced.

FakeShemp:
Very different. A substantial amount of changes are there.

rb6502:
Interesting.  Given the checksum failure I'm inclined to believe the existing one is a bad dump, so I'd just replace it anyway.  I assume the 1.02 already in MAME doesn't actually boot any games and the failure is consistent?

FakeShemp:
I've tested a bunch of other games and they mostly throw the checksum error for 1.02, except for the original Big Buck Hunter. I wonder why it doesn't error out. My dump seems to boot in all of them.

Though, I've notice another thing. On startup, the currently existing dump displays "Eagle 8MB System Boot ROM" while my dump only displays "Eagle System Boot ROM". This is in line with how 1.01 and 1.03 displays it as well. From 1.05 and later they all say "Eagle 40MB System Boot ROM". The existing dump seems to be of some other 1.02 variant, but maybe it's bad as well.

rb6502:
Interesting.  I'm going to merge this as-is since the differing messages indicate these may be legitimate different versions.

PR #14514, opened 2025-11-11
Add a Non Working Driver For Unknown Labeled 'MUSICBAR VER 201

flama12333:
dumped by @stonedDiscord 
mcu and eeprom are dumped.
adpcm rom are dumped.
Currently shown error 02. same error of marywu.cpp. k0 k1 k2 k3 can be access.


flama12333:
the hopper of mscbar was tested but there are issues
still counting  tickets dispensed after reset.
at K1 K3.
press p to change value amount
press d to payout
after complete payout will cause error 02.

flama12333:
I have submited the dump

flama12333:
if have question. let me known

PR #14516, opened 2025-11-12
mz80b_cass: 12 additions to software list

paul75s:
12 additions to software list. This includes 3 working clones in WAV format of non-working MZT files (Vosque is already included in the software list, MZT and WAV versions of SB-5520 BASIC and Puck & Puck game are added). Working WAVs converted from MZT using DumpListEditor

All other additions originally in WAV format.

Any software items which require BASIC loading first require the cassette image to be removed before they will run. This issue occurs with some MZ-2000/MZ-2200 software too, so unclear if this is an emulation issue, issue with the dumps or an issue with the original tapes or hardware. These items are playable but marked partially supported due to this issue.

Notes added to the software list to provide guidance on loading methods (in line with similar notes in the mz2000_cass software list).

All files sourced from TOSEC repositories, either UnRenamed Japanese Systems I or Rocks Preserved Games.

### DIFF
--- a/hash/mz80b_cass.xml
+++ b/hash/mz80b_cass.xml
@@ -51,7 +51,7 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
 	</software>
 
 	<software name="sb5520a" cloneof="sb5520" supported="yes">
-		<description>BASIC SB-5520 (Alt)</description>
+		<description>BASIC SB-5520 (alt)</description>
 		<year>1981</year>
 		<publisher>Sharp</publisher>
 		<info name="usage" value="IPL" />
@@ -187,7 +187,7 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
 	</software>
 
 	<software name="puckpucka" cloneof="puckpuck" supported="yes">
-		<description>Puck &amp; Puck (Alt)</description>
+		<description>Puck &amp; Puck (alt)</description>
 		<year>19??</year>
 		<publisher>K. Kuromusha</publisher>
 		<info name="usage" value="IPL" />
@@ -254,7 +254,7 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
 	</software>
 
 	<software name="vosquea" cloneof="vosque" supported="yes">
-		<description>Vosque (Alt)</description>
+		<description>Vosque (alt)</description>
 		<year>19??</year>
 		<!-- TODO: from title, check spelling on tape cover -->
 		<publisher>d.B Soft</publisher>


hash/mz80b_cass.xml
```diff
@@ -107,12 +246,37 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
...
+			<dataarea name="cass" size="5637132">
+				<rom name="vosque mz80b.wav" size="5637132" crc="1044e236" sha1="7f29b5ce3c92b9f680cb3b2361761d28644ae19f"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="walkwan" supported="partial">
+		<description>Walk Wan</description>
+		<year>19??</year>
+		<publisher>Unknown</publisher>
```
angelosa: This is Tecno Soft, either 1981 or 1982

hash/mz80b_cass.xml
```diff
@@ -98,6 +132,111 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
...
+			<dataarea name="cass" size="6007248">
+				<rom name="sos donbar.wav" size="6007248" crc="c0aaae32" sha1="4c728766bf1d1a9adf579ecb710da9e89e391668"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="spaceinv" supported="partial">
+		<description>Space Invader</description>
+		<year>19??</year>
+		<publisher>Unknown</publisher>
```
angelosa: `&lt;unknown&gt;`

hash/mz80b_cass.xml
```diff
@@ -98,6 +132,111 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
...
+				<rom name="galaxian.wav" size="1101654" crc="e37b9f7b" sha1="fba62570405e9f4efb2439555c278f99774843d8"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="icbm" supported="partial">
+		<description>ICBM</description>
+		<year>1981</year>
+		<publisher>Hudson Soft</publisher>
+		<notes><![CDATA[When loading completes remove the cassette image and the software will run]]></notes>
```
angelosa, line 165: Huh? Why?

paul75s: This seems to be an issue with all MZ-80B software that loads from BASIC (but not from IPL). The software loads but the system then appears to 'freeze'. However, removing the cassette image 'unfreezes' things and the software completes loading. This is why I have marked these as partially working. I've also added a line to the notes at the top of the file to try and explain this issue.

I've found the same issue with some but not all MZ-2000/MZ-2200 software that uses BASIC. Here I've managed to sort this issue for most but not all software by converting WAV to MZT - this is why I updated PR 14662 as I'd managed to correct much of the software that originally had this issue.

The conversion to MZT does not appear to work with MZ-80B software.

Although my assumption in both cases is that ejecting the cassette image emulates ejecting the physical tape and forces end loading to be recognised, I'm not clear if this is an emulation issue or issue with the dumps. The WAV to MZT conversion success for MZ-2000 makes me suspect it's the latter.

hash/mz80b_cass.xml
paul75s: All instances of this updated.

hash/mz80b_cass.xml
angelosa: By quick glance at CSCP sounds like tape port A writes are reversed for MZ-80B compared to MZ-2000/MZ-2200 (and MZ-2500 can do both ways). I'll take a look after this is merged.

hash/mz80b_cass.xml
paul75s: Confirmed 1982 and updated year/publisher

hash/mz80b_cass.xml
paul75s: Great, thanks. I'm sure there's a better answer to this but this (and the successful re-conversion to MZT in the MZ-2000 PR) are pretty much hitting my limits of solutions and this approach at least felt like a reasonable interim compromise.

angelosa:
> All files sourced from TOSEC repositories, either UnRenamed Japanese Systems I or Rocks Preserved Games.

Checked, I don't see most entries from there. The few I've found are in .mti format, given that in here they are in .wav instead you should upload those sets thru [CODE@](https://wiki.mamedev.org/index.php?title=Submitting_Source_Code#Submitting_ROM_or_disk_dumps)

hash/mz80b_cass.xml
```diff
@@ -98,6 +132,111 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
...
+		<info name="usage" value="IPL" />
+		<part name="cass1" interface="mz_cass">
+			<dataarea name="cass" size="11296">
+				<rom name="puck &amp; puck.mzt" size="11296" crc="0674fcc2" sha1="5bf6a86d35d89a4078d1716f86f76fdd79e27a84"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="puckpucka" cloneof="puckpuck" supported="yes">
+		<description>Puck &amp; Puck (Alt)</description>
```
cuavas: Lowercase for descriptive text that isn’t part of the title proper, besides proper nouns and initialisms.

hash/mz80b_cass.xml
```diff
@@ -107,12 +246,37 @@ CHECK SUM ERROR, boots with Z80 clock set at 3.2 MHz (80%)
...
-
 		<part name="cass1" interface="mz_cass">
 			<dataarea name="cass" size="25472">
 				<rom name="vosque mz80b.mzt" size="25472" crc="1dc3b210" sha1="3c87d52292e356686eca927cd26614a3ffe6739e"/>
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="vosquea" cloneof="vosque" supported="yes">
+		<description>Vosque (Alt)</description>
```
cuavas: This one, too.

hash/mz80b_cass.xml
paul75s: Probably a dumb question, but just to be certain before I amend it - is the change here Alt to alt?

angelosa: Yep, should be `alt`

cuavas: Yes.

paul75s: All instances of this updated

hash/mz80b_cass.xml
paul75s: All instances of this updated

paul75s:
> > All files sourced from TOSEC repositories, either UnRenamed Japanese Systems I or Rocks Preserved Games.
> 
> Checked, I don't see most entries from there. The few I've found are in .mti format, given that in here they are in .wav instead you should upload those sets thru [CODE@](https://wiki.mamedev.org/index.php?title=Submitting_Source_Code#Submitting_ROM_or_disk_dumps)

Apologies, should have been clearer in my initial comments - MTI files have been converted to WAV.

Apologies, I forgot to send the files for these updates - done now.

cuavas:
Is it really a good idea to use MTI-to-WAV conversions?  It could lead to real audio dumps of the cassettes never surfacing if people think the conversions are real tape audio dumps.  They should at least be marked in some way to indicate that they’re regenerated and not actual tape audio.

paul75s:
> Is it really a good idea to use MTI-to-WAV conversions? It could lead to real audio dumps of the cassettes never surfacing if people think the conversions are real tape audio dumps. They should at least be marked in some way to indicate that they’re regenerated and not actual tape audio.

What limited information I've found online suggests MTI files are simply WAV files and testing with DumpListEditor and Audacity (and indeed MAME) seems to confirm this, so here the files are unchanged, I've simply amended the file extension to WAV. Appreciate this sounds like a bit of a hack but seemed appropriate based on the above. I guess the other option would be to enable MTI as a supported file extension but this didn't necessarily seem any better if really these are WAV files.

However, if this approach seems OK I'm certainly happy to add comments for these to flag the originals were MTI 'format'.

PR #14517, opened 2025-11-13
add  adpcm bank code. huangyeh_m68k.cpp

flama12333:
the adpcm bankcode  was from another driver that still in pull request.  due to preliminary emulation status
I had to use pc = 061C to attract mode and check something at error log.

### DIFF
--- a/src/mame/skeleton/huangyeh_m68k.cpp
+++ b/src/mame/skeleton/huangyeh_m68k.cpp
@@ -145,8 +145,7 @@ void huangyeh_m68k_state::audio_program_map(address_map &map)
 
 void huangyeh_m68k_state::adpcm_bank_w(uint8_t data)
 {
-
-		m_oki->set_rom_bank(data & 0x01);
+	m_oki->set_rom_bank(data & 0x01);
 }
 
 void huangyeh_m68k_state::ramdac_map(address_map &map)


src/mame/skeleton/huangyeh_m68k.cpp
```diff
@@ -133,12 +136,19 @@ void huangyeh_m68k_state::audio_program_map(address_map &map)
...
 	map(0xf8a0, 0xf8a0).rw("oki", FUNC(okim6295_device::read), FUNC(okim6295_device::write));
 	map(0xf8d0, 0xf8d0).r(m_soundlatch, FUNC(generic_latch_8_device::read));
 //  map(0xf8e0, 0xf8e0).w soundlatch acknowledge or NMI ack
 	map(0xf8e0, 0xf8e0).lw8(NAME([this] (offs_t offset, u8 data) { m_soundlatch->acknowledge_w(); }));
 }
 
+void huangyeh_m68k_state::adpcm_bank(uint8_t data)
+{
+
+		m_oki->set_rom_bank(data & 0x01);
```
cam900: Function name for write to something is normally (function)_w, and Indentation consistency is should be correct. (in this case, it should be 1 tab, instead 2 tabs.)

flama12333: the adpcm rom has two bank. let me to check.

flama12333: I I already used the data & 0x01 after making the pull request.
 will cause the issues if I edit  data & 0x02

error log:
[:oki] Warning: requested bank 2 higher than actual bank count 2
[:oki] Requested to play invalid sample 4a

flama12333:
the adpcm pitch clock was verified.

src/mame/skeleton/huangyeh_m68k.cpp
MooglyGuy: You should check whatever program you're using for machine translation, because clearly you are misunderstanding what is being said.

The comment has to do with INDENTING. You should be using 1 four-space tab, not 2 tabs.

PR #14518, opened 2025-11-13
esq5505 Ensoniq VFX family layouts: Specify the bounds for each group.

cbrunschen:
Specify the bounds for each group in the Ensoniq VFX family layouts.

This fixes an issue where the VFX (non-SD)'s button area is stretched vertically, because it does not have the "System" label at the top, leaving an empty space that was not included in the bounds calculation. This resulted in the button area being stretched vertically to fill the space, and the buttons being taller than they should.

Before:
Screenshot From 2025-11-13 16-12-58
Screenshot From 2025-11-13 16-16-22

After:
Screenshot From 2025-11-13 16-12-22
Screenshot From 2025-11-13 16-16-00

### DIFF
--- a/src/mame/layout/sd1.lay
+++ b/src/mame/layout/sd1.lay
@@ -931,6 +931,7 @@
 
 	<!-- Group definitions -->
 	<group name="Sliders">
+		<bounds x="0" y="15" width="90" height="88.2" />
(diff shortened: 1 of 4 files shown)


cbrunschen:
Thank you!

PR #14519, opened 2025-11-13
Tek4404 wip

Elektraglide:
**REQUIRES** `-DUSE_TEK4404_CHANGES`  when compiling in order to work since it requires delays in ncr5385 and backing out resetting the serial interface in diserial.cpp

EDIT:  no longer needs special #ifdef to build and run successfully.

### DIFF
--- a/src/devices/machine/ncr5385.cpp
+++ b/src/devices/machine/ncr5385.cpp
@@ -177,7 +177,7 @@ void ncr5385_device::scsi_ctrl_changed()
 				ctrl & S_REQ ? " REQ" : "", ctrl & S_ACK ? " ACK" : "");
 
 		if (m_state != IDLE)
-			m_state_timer->adjust(attotime::from_usec(80));
+			m_state_timer->adjust(attotime::from_usec(30));
 	}
 	else
 	{
--- a/src/devices/bus/nscsi/hd.h
+++ b/src/devices/bus/nscsi/hd.h
@@ -20,9 +20,9 @@ protected:
 	virtual void device_reset() override ATTR_COLD;
 	virtual void device_add_mconfig(machine_config &config) override ATTR_COLD;
 
-	// experiment for slowing target device
-	virtual attotime scsi_data_byte_period() override { return attotime::from_nsec(770); }
-
+	// experiment for slowing target device (based on 1.2Mbps)
+	virtual attotime scsi_data_byte_period() override { return attotime::from_nsec(820); }
+	virtual attotime scsi_data_command_delay() override { return attotime::from_nsec(20000); }
 
 	virtual void scsi_command() override;
 	virtual uint8_t scsi_get_data(int id, int pos) override;
--- a/src/devices/machine/nscsi_bus.cpp
+++ b/src/devices/machine/nscsi_bus.cpp
@@ -327,7 +327,8 @@ void nscsi_full_device::step(bool timeout)
 			break;
 		buf_control_push()->action = BC_MSG_OR_COMMAND;
 		scsi_state = TARGET_NEXT_CONTROL;
-		step(false);
+		scsi_timer->adjust(scsi_bus_set_delay(), false);
+		//step(false);
 		break;
 
 	case RECV_BYTE_T_WAIT_ACK_1 << SUB_SHIFT:
@@ -404,7 +405,8 @@ void nscsi_full_device::step(bool timeout)
 			else {
 				scsi_state = TARGET_NEXT_CONTROL;
 			}
-			step(false);
+			scsi_timer->adjust(scsi_bus_set_delay(), false);
+			//step(false);
 			break;
 
 		case BC_DATA_OUT:
@@ -418,7 +420,8 @@ void nscsi_full_device::step(bool timeout)
 			else {
 				scsi_state = TARGET_NEXT_CONTROL;
 			}
-			step(false);
+			scsi_timer->adjust(scsi_bus_set_delay(), false);
+			//step(false);
 			break;
 
 		case BC_MESSAGE_1:


src/devices/machine/ncr5385.cpp
```diff
@@ -162,25 +194,35 @@ void ncr5385_device::device_reset()
...
 }
 
+attotime ncr5385_device::scsi_data_byte_period()
+{
+#ifdef USE_TEK4404_CHANGES
+	return attotime::from_nsec(200);
+#else
+	return attotime::zero;
+#endif
+}
```
angelosa: I hate C/C++ compile switches with a passion. `attotime::zero` is a fallback that obviously needs changing depending on the controller needs.
Also: where this magic number comes from?

src/devices/machine/ncr5385.h
```diff
@@ -63,6 +63,7 @@ class ncr5385_device
 	virtual void device_reset() override ATTR_COLD;
 
 	// ncsci_device implementation
+	virtual attotime scsi_data_byte_period();
```
angelosa: How this is not overriding the base method and manages to compile anyway? Is this `ncr5385_device` supposed to be a `nscsi_full_device`, and this function essentially do nothing here?

src/emu/diserial.cpp
```diff
@@ -204,8 +204,10 @@ void device_serial_interface::set_data_frame(int start_bit_count, int data_bit_c
 	{
 		m_rcv_bit_count++;
 	}
-
+#ifndef USE_TEK4404_CHANGES
 	receive_register_reset();
+#endif
```
angelosa: As above.

src/osd/modules/monitor/monitor_common.cpp
```diff
@@ -43,7 +43,10 @@ std::shared_ptr<osd_monitor_info> monitor_module_base::monitor_from_handle(std::
 
 void monitor_module_base::add_monitor(std::shared_ptr<osd_monitor_info> monitor)
 {
-	list().push_back(monitor);
+	// a set not a list
+	if (std::find(list().begin(), list().end(), monitor) == list().end())
+		list().push_back(monitor);
+
 	m_monitor_index[monitor->oshandle()] = monitor;
```
angelosa: These three OSD changes are for a separate PR.

src/mame/tektronix/tek440x.cpp
```diff
@@ -49,25 +52,477 @@
...
 #include "speaker.h"
 
+#include "tek4404.lh"
+
+#define LOG_GENERAL (1U << 0)
+#define LOG_MMU (1U << 1)
+#define LOG_FPU (1U << 2)
+#define LOG_SCSI (1U << 3)
+
+#define VERBOSE LOG_GENERAL
```
angelosa: Do not redefine `LOG_GENERAL`, that's already defined in `src/emu/logmacro.h`.
Also the canonical way is to express `VERBOSE` definition with parenthesis.

src/mame/tektronix/tek440x.cpp
```diff
@@ -49,25 +52,477 @@
...
+// mapcntl bits
+constexpr int MAP_VM_ENABLE = 4;
+constexpr int MAP_SYS_WR_ENABLE = 5;
+// mapcntrl result bits
+constexpr int MAP_BLOCK_ACCESS = 6;
+constexpr int MAP_CPU_WR = 7;
+
+#define OFF8_TO_OFF16(A)	((A)>>1)
+#define OFF16_TO_OFF8(A)	((A)<<1)
+
```
angelosa, line 95: Is there any real use of these macros? What's the catch in just using these shifts where it matters instead of obfuscation?

src/mame/tektronix/tek440x.cpp
```diff
@@ -49,25 +52,477 @@
...
+constexpr offs_t MAXRAM = 0x200000;	// +1MB
+//constexpr offs_t MAXRAM = 0x400000;	// +3MB
```
angelosa, line 97: This sounds a missing `ram_device` use.

src/mame/tektronix/tek440x.cpp
```diff
@@ -49,25 +52,477 @@
...
+
+DECLARE_DEVICE_TYPE(M68010_TEKMMU, m68010_tekmmu_device)
+
+class m68010_tekmmu_device : public m68010_device
+{
+	// HACK
+	u8 *m_map_control_ptr = nullptr;
+	u16 *m_map_ptr = nullptr;
+
+
```
angelosa, line 115: Whatever is this (custom MMU?) should probably be moved as a separate file instead of living here.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
 }
 
 
+static int inbuserr = 0;
```
angelosa, line 789: Absolutely no `static` for driver logic.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
...
+	if (!machine().side_effects_disabled())
+	{
+		if ((offset>>11) < 0x20)
+		{
+			LOGMASKED(LOG_MMU, "map_r 0x%08x => %04x\n",offset>>11, m_map[(offset >> 11)&0x7ff] );
+		}
+
+		// selftest does a read and expects it to fail iff !MAP_SYS_WR_ENABLE; its not WR enable, its enable..
+		// NB page 2.1-52 shows WrMapEn coming from latch
+		if (!BIT(m_latched_map_control, MAP_SYS_WR_ENABLE))
```
angelosa, line 974: This comment needs a revision.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
...
-	COMBINE_DATA(&m_map[offset >> 11]);
+	if ((offset>>11) < 0x20)
+	{
+		LOGMASKED(LOG_MMU, "map_w: %08x  <= %04x paddr(%08x) PID(%d) dirty(%d) write_enable(%d)\n",
+			(offset >> 11) & 0x7ff, data,
+			OFF16_TO_OFF8(BIT(data, 0, 11)<<11),BIT(data, 11, 3), data & 0x8000 ? 1 : 0, data & 0x4000 ? 1 : 0,
+			m_maincpu->pc());
+	}
+	
+	// NB page 2.1-52 shows WrMapEn coming from latch
```
angelosa, line 995: Repeats the same thing as above.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
...
+		if (BIT(data, MAP_VM_ENABLE) && (data & 15))
+		for(uint32_t i=0; i<2048; i++)
+		{
+			if (m_map[i])
+				LOGMASKED(LOG_MMU, "mapcntl_w: %08x -> paddr(%08x) PID(%d) dirty(%d) write_enable(%d)\n",
+				OFF16_TO_OFF8(i << 11), OFF16_TO_OFF8(BIT(m_map[i], 0, 11)<<11),
+				BIT(m_map[i], 11, 3), m_map[i] & 0x8000 ? 1 : 0, m_map[i] & 0x4000 ? 1 : 0);
+		}
+
+	}
```
angelosa, line 1064: This suppressed control block derailed. Please make sure to encapsulate it properly with curly braces.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
...
+{
+	if (0)
+	if (m_videocntl != data)
+	{
+		LOG("m_videocntl %02x\n", data);
+		LOG("m_videocntl VBenable   %2d\n", BIT(data, 6));
+		LOG("m_videocntl ScreenOn   %2d\n", BIT(data, 5));
+		LOG("m_videocntl ScreenInv  %2d\n", BIT(data, 4));
+		LOG("m_videocntl ScreenPan  %2d\n", data & 15);
+	}
```
angelosa: Merge these two `ifs`, also is this logging really useful? If you subscribed the relevant variables to `save_item` you can see the live state via the debugger CTRL+M instead.

src/mame/tektronix/tek440x.cpp
```diff
@@ -202,84 +783,549 @@ u32 tek440x_state::screen_update(screen_device &screen, bitmap_ind16 &bitmap, co
...
 	if (!m_kb_rclamp && m_kb_loop != BIT(data, 7))
 		m_keyboard->kdo_w(!BIT(data, 7) || m_kb_tdata);
 
 	m_kb_loop = BIT(data, 7);
+	m_diag = data;
 }
 
+// copied from stkbd.cpp
+int tek440x_state::mouseupdate()
+{
```
angelosa, line 1235: Is this a generic quadrature mouse that needs its own device?

src/mame/tektronix/tek440x.cpp
```diff
@@ -312,54 +1358,199 @@ void tek440x_state::kb_tdata_w(int state)
...
+	LOG("store_w\n");
+	m_novram->store(1);
+	m_novram->store(0);
+}
+
+void tek440x_state::palette(palette_device &palette) const
+{
+	palette.set_pen_color(0, rgb_t(0xec, 0xf4, 0xff));   // 2 color tiles
+	palette.set_pen_color(1, rgb_t(0x00, 0x00, 0x00));
```
angelosa: Those aren't tiles.

src/mame/tektronix/tek440x.cpp
```diff
@@ -409,22 +1613,51 @@ void tek440x_state::tek4404(machine_config &config)
...
+	m_screen->set_screen_update(FUNC(tek440x_state::screen_update));
+	m_screen->set_palette("palette");
+	m_screen->screen_vblank().set([this](int state)
+    {
+		printf("vblank %d\n", state);
+		if (state && m_vint_enable)
+		{
+			m_maincpu->set_input_line(M68K_IRQ_6, ASSERT_LINE);
+		}
+    });
```
angelosa: No `printf` in live code.

src/mame/tektronix/tek440x.cpp
```diff
@@ -409,22 +1613,51 @@ void tek440x_state::tek4404(machine_config &config)
...
+	PALETTE(config, "palette", FUNC(tek440x_state::palette),2);
+	
+	MOS6551(config, m_acia, 40_MHz_XTAL / 4 / 10);
+	m_acia->set_xtal(1.8432_MHz_XTAL);
+	m_acia->txd_handler().set("rs232", FUNC(rs232_port_device::write_txd));
+	m_acia->irq_handler().set_inputline(m_maincpu, M68K_IRQ_7);
+
+	I8255A(config, m_printer);
+	m_printer->in_pb_callback().set_constant(0x30);
+m_printer->in_pb_callback().set_constant(0xb0);		// HACK:  vblank always checks if printer status < 0
```
angelosa: De-aligned row here.

src/emu/diserial.cpp
pmackinlay: Instead of hacking diserial, do this:

```diff --git a/src/devices/machine/mc68681.cpp b/src/devices/machine/mc68681.cpp
index dbefa1507be..e19716e8661 100644
--- a/src/devices/machine/mc68681.cpp
+++ b/src/devices/machine/mc68681.cpp
@@ -1641,6 +1641,7 @@ void duart_channel::write_chan_reg(int reg, uint8_t data)

 void duart_channel::write_MR(uint8_t data)
 {
+       bool const frame_change = (MR_ptr == 0) ? ((MR1 ^ data) & 0x1f) : ((MR2 ^ data) & 0x0f);
        if (MR_ptr == 0)
        {
                MR1 = data;
@@ -1650,7 +1651,8 @@ void duart_channel::write_MR(uint8_t data)
        {
                MR2 = data;
        }
-       recalc_framing();
+       if (frame_change)
+               recalc_framing();
        update_interrupts();
 }```

File: src/devices/machine/ncr5385.cpp [hunk 18]
```diff
@@ -162,25 +194,35 @@ void ncr5385_device::device_reset()
...
+
 void ncr5385_device::scsi_ctrl_changed()
 {
 	u32 const ctrl = scsi_bus->ctrl_r();
 
 	static char const *const nscsi_phase[] = { "DATA OUT", "DATA IN", "COMMAND", "STATUS", "*", "*", "MESSAGE OUT", "MESSAGE IN" };
 
-	if ((ctrl & S_BSY) && !(ctrl & S_SEL))
+	if (ctrl & S_BSY)
```
pmackinlay, line 171: I have previously suggested this change, and I confirm it's good; I think this is the only change required to ncr5385.cpp.

src/devices/machine/ncr5385.cpp
```diff
@@ -761,6 +809,36 @@ int ncr5385_device::state_step()
 			// clear data and ACK
 			scsi_bus->data_w(scsi_refid, 0);
 			scsi_bus->ctrl_w(scsi_refid, 0, S_ACK);
+
+			if (m_cnt == 0)
+			{
+#ifdef USE_TEK4404_CHANGES
```
pmackinlay: I've tested simulating a slow hard disk instead of this change, and it can successfully complete the "diskrepair" and reach a login prompt. I don't have the knowledge of this system to test further, but I believe this change is wrong and the delay occur in the target, not this device.

src/devices/machine/ncr5385.cpp
pmackinlay: This change does not make sense and will have no effect; `ncr5385_device` is not derived from `nscsi_full_device`.

Elektraglide: yes, realised that later.

src/mame/tektronix/tek440x.cpp
Elektraglide: I can ifdef them out; they are a handy reminder of the bit meanings

src/emu/diserial.cpp
Elektraglide: With the above changes I get:
```
Tektronix 4404
Self Test Error-Keyboard Failure or Not. Attached [02, 0002]
Fatal Error - Cannot Initialize Keyboard_
```
Any ideas?

pmackinlay: The only other place that resets the receive register is when changing the receive baud rate, so it must be that; do the same thing.

src/devices/machine/ncr5385.cpp
Elektraglide: I found having logging with timestamps in ncr5385 very useful.

src/devices/machine/ncr5385.cpp
Elektraglide: OK.  I've had a few _false dawns_ at being able to just boot.    Running a script that does "cp /system.boot /tmp" 20-30 times seems to always flush out timing issues.

Elektraglide:
> The ACB-4000 device which acts as a SCSI target used in this system no doubt accounts for the delays assumed by the host software. The ACB-4000 is documented as only sustaining up to 1.3MBps for SCSI data transfers (rather than the maximum 5MBps for a SCSI-I device).
> 
> You can simulate such a slow hard disk by overriding `scsi_data_byte_period()` in a SCSI target and supplying a suitable value. As an experiment, you can modify src/devices/nscsi/hd.h by adding the following:
> 
> `virtual attotime scsi_data_byte_period() override { return attotime::from_nsec(770); }`
> 
> This will cause the SCSI ACK signal for the default SCSI hard disk target to be delayed, producing a data rate of ~1.3MBps, and appears to eliminate reported SCSI issues.
> 
> Moving forward, adding a new SCSI device which represents a "slow" hard disk is the right interim measure until the ACB-4000 is emulated.

OK tried this and my disk thrashing script succeeds.  Hurrah.   Thank you.

**BUT**, XFI_ACK delay version takes **1m12s** to run the copying script,  the hd.hpp changes that to **5m25s**.!!   5x slower,  I'm guessing I can tweak that 770ns delay?

Elektraglide:
> **BUT**, XFI_ACK delay version takes **1m12s** to run the copying script, the hd.hpp changes that to **5m25s**.!! 5x slower, I'm guessing I can tweak that 770ns delay?

Ah  no, making the delay less than 770ns hangs the machine on boot...

src/devices/machine/ncr5385.cpp
angelosa: And I'd like colorized output in `logmacro.h`, which (*) doesn't belong to the device core but should be an opt-in option of said module if possible.

(*) beyond not having a Windows to test its specific ANSI escape sequences.

pmackinlay:
Trying to speed up something which was failing because it was too fast seems illogical. A 5x slowdown doesn't seem out of order since you're talking about dropping back to ~1.3MB/s from whatever it was getting without limits before; how long does the script take on hardware?

The 770 wasn't random; it was chosen specifically to represent the documented throughput of the ACB-4000.

Again you seem to be ignoring that MAME is fundamentally about accuracy; usability of the end result is a nice side effect. If you want it to run at unrealistic speed, try unthrottling it or make local hacks, but don't expect those to make it upstream.

Another option would be to fix the broken software, which might even allow your hardware to use a modern/fast storage device.

Elektraglide:
> Trying to speed up something which was failing because it was too fast seems illogical. A 5x slowdown doesn't seem out of order since you're talking about dropping back to ~1.3MB/s from whatever it was getting without limits before; how long does the script take on hardware?
> 
I will test on tek4404 h/w; I'm not in the country right now

src/mame/tektronix/tek440x.cpp
Elektraglide: OK.  FYI the ethernet driver has printfs

src/mame/tektronix/tek440x.cpp
Elektraglide: OK.  I use offside formatting to easily find non-permanent changes.

src/mame/tektronix/tek440x.cpp
angelosa: Doesn't mean it's something that should be there ...

src/emu/diserial.cpp
Elektraglide: In duart_channel::write_chan_reg(), right?   (Obv,  I have not a clue how this chip works.)

src/mame/tektronix/tek440x.cpp
Elektraglide: Rmoved it

src/emu/diserial.cpp
Elektraglide: Given we check if any params have changed and early out, I just added a test if its also already syncronized which works but as I said before, I have no idea how this chip works so asking me to make changes like this is probably a bad idea.

```
	// no params have changed and already syncronised
	if (m_rcv_flags & RECEIVE_REGISTER_SYNCHRONISED)
		return;
		
	receive_register_reset();
```

Elektraglide:
This PR now builds and works without need for magic #ifdef for Tek4404 and doesn't change ncr5385.
[I noticed tek4404 was using wrong clock frequency for ncr5385 - dont think it materially changes the problem - but good to use the documented crystal.]

Elektraglide:
> > Trying to speed up something which was failing because it was too fast seems illogical. A 5x slowdown doesn't seem out of order since you're talking about dropping back to ~1.3MB/s from whatever it was getting without limits before; how long does the script take on hardware?
> 
> I will test on tek4404 h/w; I'm not in the country right now

Oh and just so to add some numbers to this:

My _ACK_ hack  is a transfer rate of **64k Bytes/sec**
Using scsi_data_byte_period(770) is a transfer rate of **14k Bytes/sec**

That does seem very very slow.

Lord-Nightmare:
> > > Trying to speed up something which was failing because it was too fast seems illogical. A 5x slowdown doesn't seem out of order since you're talking about dropping back to ~1.3MB/s from whatever it was getting without limits before; how long does the script take on hardware?
> > 
> > 
> > I will test on tek4404 h/w; I'm not in the country right now
> 
> Oh and just so to add some numbers to this:
> 
> My _ACK_ hack is a transfer rate of **64k Bytes/sec** Using scsi_data_byte_period(770) is a transfer rate of **14k Bytes/sec**
> 
> That does seem very very slow.

which one these is more accurate to the real hardware?

PR #14521, opened 2025-11-13
Add billboard to 16:9 aspect ratio setting for Indy 500

AnthonyRyuki:


AnthonyRyuki:
Well, this game doesn't quite use the billboard feature like the other Sega games. I was actually wondering why it was showing letters that made no sense. Instead of a 7-segment display for each player, it has a set of starting lights, as shown in this picture: https://64.media.tumblr.com/8340cdc4dedb0bbbc8df01114a25afc3/tumblr_ps6qrimpRQ1tp2r9i_1280.jpg
Still, I think this is a step in the right direction. I'll see if I can implement this properly...

AnthonyRyuki:
I just can't do it. Apparently, each bit of the m_digits variable controls each light, but I can't figure out which ones, and even if I could, I don't have what it takes to code it.
I'm sorry about wasting your time.

PR #14522, opened 2025-11-13
sharp/mz700.cpp: Enable mz700_cass software list for Sharp MZ-800

paul75s:
As Sharp MZ-800 is fully backwards compatible with MZ-700, change enables mz700_cass as a compatible software list for MZ-800.

This avoids duplication in the mz800_cass software list, particularly of software that functions better on MZ-800 but is also MZ-700 compatible (and may over time, with fuller emulation of MZ-700 capabilities such as PCG run consistency on both machines).

### DIFF
--- a/src/mame/sharp/mz700.cpp
+++ b/src/mame/sharp/mz700.cpp
@@ -428,7 +428,7 @@ void mz800_state::mz800(machine_config &config)
 
 	config.device_remove("cass_list");
 	SOFTWARE_LIST(config, "cass_list").set_original("mz800_cass");
-	SOFTWARE_LIST(config, "cass_list_700").set_original("mz700_cass");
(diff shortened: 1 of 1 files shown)


src/mame/sharp/mz700.cpp
```diff
@@ -428,6 +428,7 @@ void mz800_state::mz800(machine_config &config)
 
 	config.device_remove("cass_list");
 	SOFTWARE_LIST(config, "cass_list").set_original("mz800_cass");
+	SOFTWARE_LIST(config, "cass_list_700").set_original("mz700_cass");
```
angelosa: Should be `set_compatible`

paul75s: Updated to compatible - did have that debate with myself but made the wrong call.

PR #14533, opened 2025-11-15
rendfont.cpp,rendlay.cpp: add font support to `<text>` components

cbrunschen:
For discussion: improves the layout `<text>` component to allow specifying the font and style to use, with fairly minimal changes to the layout rendering system.

This adds a "font" attribute to the `<text>` layout component, defaulting to `"default"` if nothing is specified.

It also changes the `render_font` constructor to accept not just a single "filename" but a comma-separated list of "filenames". These actually don't seem to be file names at all, but font specifications that may be symbolic (like the default `"Liberation Sans|Regular"` in `font_osd.cpp`), which includes a `|` separator between font family and style; so adding another separator between different font specifications seems a not unreasonable thing, but is of course open to be changed.

This is in turn used in the layouts for the Ensoniq VFX family of keyboards, including the `sd132`, to use the correct Bold, Italic and Regular styles of a suitable Helvetica-compatible sans-serif font, falling back to "default" of nothing else is found. This includes a first choice of an explicitly nonexistent font to demonstrate that the list of fonts is in fact traversed.

Screenshot From 2025-11-15 11-16-26

The labels along the thick coloured line - including '**Performance**', '**Sequencer**' and '**Programming**' - are all in Bold; the labels above the buttons - '_Volume_', '_Pan_', '_Timbre_', etc - are all Italic; and the extra little labels between white lines, 'Tracks' and 'Edit', are in Regular style (neither Bold nor Italic).

### DIFF
--- a/docs/source/techspecs/layout_files.rst
+++ b/docs/source/techspecs/layout_files.rst
@@ -517,17 +517,17 @@ text
     stretched horizontally to fill its bounds. If the ``align`` attribute is
     absent, the text will be centred.
 
-		You can also specify the font to use using the ``font`` attribute. This 
-		contains a comma (``,``)-separated list of fonts to use in order of preference.
-		Font styles (Regular, Bold, Italic,oldItalic) are specified after the font name
-		separated by a vertical bar (``|``).
+    You can also specify the font to use using the ``font`` attribute. This 
+    contains a comma (``,``)-separated list of fonts to use in order of preference.
+    Font styles (Regular, Bold, Italic,oldItalic) are specified after the font name
+    separated by a vertical bar (``|``).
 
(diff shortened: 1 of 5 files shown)


cbrunschen:
I've verified that the current version of this also works under Windows. 

Here's what I get with the `win` font provider, before:
win_before
and after:
win_after

And with the `dwrite` provider, before:
dwrite_before

and after:
dwrite_after

cbrunschen:
A friend very kindly ran this on a Mac as well:
dc9e9bfd-f0c2-4616-a4ad-8ed59ecfd1c6

While this is a JPEG it clearly shows botb Regular, Bold and Italic fonts working, as intended.

PR #14534, opened 2025-11-15
ibm5170.xml - Add a few SPCS games

FakeShemp:
## New working software list items (ibm5170.xml)
Alfons påhittelåda (Sweden) [FakeShemp]
Cheops pyramid (Sweden) [conradtrout @ archive.org]
Handla med Alfons (Sweden) [FakeShemp]
Måla med Alfons (Sweden) [Max lundell @ archive.org]
Mitt första datorprogram (Sweden) [retrotrollet @ archive.org]
Pussla med Alfons (Sweden) [Max lundell @ archive.org]
Alfons på äventyr i trafiken (Swedish) [Max lundell @ archive.org]
Kom till Alfons hemliga kalas (Sweden) [Max lundell @ archive.org]

### DIFF
--- a/hash/ibm5170.xml
+++ b/hash/ibm5170.xml
@@ -3322,7 +3322,7 @@ license:CC0-1.0
 	</software>
 
 	<software name="pcdos2ksv" cloneof="pcdos2k">
-		<description>IBM PC DOS 2000 (Sweden)</description>
+		<description>IBM PC DOS 2000 (Swedish)</description>
 		<year>1998</year>
 		<publisher>IBM</publisher>
 		<info name="version" value="7.00, Revision 1" />


hash/ibm5170.xml
```diff
@@ -10361,6 +10361,53 @@ The disks have had their OEM IDs modified by Windows
...
+			</dataarea>
+		</part>
+		<part name="flop2" interface="floppy_3_5">
+			<dataarea name="flop" size = "9906489">
+				<rom name="Alfons paahittelaada (Swedish) (Disk 2).mfi" size="9906489" crc="b97f6654" sha1="1beabe7fdf4c619b336d1682398d510cb581d352"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="alfonstra" supported="no">
```
angelosa: Needs a reason about why `supported="no"`, preferably in a notes field.

FakeShemp: Disks work now as I've converted them to IMD. They are 82-track disks which caused mame to fail in parsing them. I've added a note about this.

hash/ibm5170.xml
```diff
@@ -10361,6 +10361,56 @@ The disks have had their OEM IDs modified by Windows
 		</part>
 	</software>
 
+	<software name="alfonspl">
+		<description>Alfons påhittelåda (Swedish)</description>
```
angelosa: `description` tags should be by country not by language i.e. Sweden rather than Swedish.

FakeShemp: None of the other entries do, hence I did it this way. Should I change this to Sweden, keep it consistent and use Swedish, or update all entries? The last one would require some effort and probably should be a separate PR.

angelosa: I see only `pcdos2ksv` that has the `Swedish` wrong tag.

FakeShemp: I was looking at the other languages as well, like German.

angelosa: That's out of scope for this PR yes :)

FakeShemp: Should be solved now.

cuavas: It does depend on whether it’s a release licensed for a certain country or a release in a certain language, though.

PR #14543, opened 2025-11-17
Layout for the Patinho Feio computer

felipesanches:
This was developed by @felipe-27 and @nicolysampaio under my guidance during the past few months.

The driver itself can still be improved a lot, but this PR focuses mostly on adding their improved layout, which looks like this:

image

While the version currently available on MAME looks like this:

image

And, as a reference, an actual photo of the real computer is here:
image

With more info available at https://forum.fiozera.com.br/t/resgate-historico-do-computador-patinho-feio-usp-1971/58

### DIFF
--- a/src/mame/usp/patinhofeio.cpp
+++ b/src/mame/usp/patinhofeio.cpp
@@ -13,6 +13,8 @@
 
 #include "patinho.lh"
 
+namespace {
+
 class patinho_feio_state : public driver_device {
 public:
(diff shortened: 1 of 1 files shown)


felipesanches:
Since this is a collaborative PR, here is the list of authors to be credited:
* Felipe Nascimento (@felipe-27)
* Nicoly Sampaio (@nicolysampaio)
* Felipe Sanches (@felipesanches - that's myself)

PR #14547, opened 2025-11-18
Fix MT05574, MT05445 [sjy96525]

sjy555525:


happppp:
Did you test both?
See reaction on the ddragon bug report where it's said the Japanese manual is likely a misprint. And the common meaning for "Free Play" is that no coins are needed.

And the comments in the jdredd driver about the dipswitches, do they work fine afterall?

PR #14550, opened 2025-11-19
rendlay.cpp: Improve accuracy and performance calculating text aspect ratio.

cbrunschen:
Also adds another text alignment option to stretch the text to fill its bounds horizontally.

The current code calculating the aspect ratio for text involves looping, calculating the width of text for a variety of decreasing aspect ratios until the text's width is less than the available bounds. However, this string width calculation performs the same loop over the text each time, finally multiplying by the candidate aspect ratio. That text width calculation thus really only needs to be done once. Further, instead of trying different aspect ratios, the ratio can simply be calculated directly by dividing the width of the bounds by the string's width. This also calculates a more accurate aspect ratio, rather than always resulting in an aspect ratio of (0.95)^n.

For example, when trying to fit a 101-pixel wide text into a 100-pixel wide space, the current code would result in an aspect ratio of 0.95, making the text 96 pixels wide,  leaving 4 pixels unused;  the new code will instead calculate the aspect ratio as 100/101 == 0.9900990099... , making the text use the full 100 available pixels.

This in turn allows us to easily calculate the ratio also if we want to not just schrink but also stretch the text to fill the available space, so we add that as another text alignment option, number 3 (three).

This PR also demonstrates this in the VFX family of layouts, where on the default Full view, the text "MUSIC PRODUCTION SYNTHESIZER" ("DYNAMIC COMPONENT SYNTHESIZER" on the VFX), the logo-like text "ensoniq", and the keyboard-specific markers "VFX", "VFX-SD", "SD-1" and "3 2 V O I C E" can now be made to fill their available space horizontally as they should, making it all look that much more like the real thing. One example from the sd132, before:

Screenshot From 2025-11-19 16-34-01

after:

Screenshot From 2025-11-19 16-33-18
 
and the real thing:

image

While still not perfect, the "after" at least has the corrects widths and horizontal alignments.

### DIFF
--- a/docs/source/techspecs/layout_files.rst
+++ b/docs/source/techspecs/layout_files.rst
@@ -512,9 +512,10 @@ text
     Draws text in using the UI font in the specified colour.  The text to draw
     must be supplied using a ``string`` attribute.  An ``align`` attribute may
     be supplied to set text alignment.  If present, the ``align`` attribute must
-    be an integer, where 0 (zero) means centred, 1 (one) means left-aligned, and
-    2 (two) means right-aligned.  If the ``align`` attribute is absent, the text
-    will be centred.
+    be an integer, where 0 (zero) means centred, 1 (one) means left-aligned,
+    2 (two) means right-aligned, and 3 (three) means that the text will be
+    stretched horizontally to fill its bounds. If the ``align`` attribute is
+    absent, the text will be centred.
 led7seg
     Draws a standard seven-segment (plus decimal point) digital LED/fluorescent
     display in the specified colour.  The low eight bits of the element’s state
--- a/scripts/build/complay.py
+++ b/scripts/build/complay.py
@@ -452,7 +452,7 @@ class LayoutChecker(Minifyer):
             if 'string' not in attrs:
                 self.handle_error('Element text missing attribute string')
             align = self.check_int_attribute(name, attrs, 'align', None)
-            if (align is not None) and ((0 > align) or (2 < align)):
+            if (align is not None) and ((0 > align) or (3 < align)):
                 self.handle_error('Element text attribute align "%s" not in valid range 0-2' % (attrs['align'], ))
             self.check_component(name, attrs)
         elif 'simplecounter' == name:
--- a/src/emu/rendlay.cpp
+++ b/src/emu/rendlay.cpp
@@ -3219,16 +3219,8 @@ protected:
 					bitmap_argb32 tempbitmap(dest.width(), dest.height());
 
 					// get the width of the string
-					float aspect = 1.0f;
-					s32 width;
-
-					while (1)
-					{
-						width = font->string_width(ourheight / num_shown, aspect, m_stopnames[fruit]);
-						if (width < bounds.width())
-							break;
-						aspect *= 0.95f;
-					}
+					s32 width = font->string_width(ourheight / num_shown, 1.0f, m_stopnames[fruit]);
+					float aspect = width > bounds.width() ? (float) bounds.width() / (float) width : 1.0f;
 
 					float curx = bounds.left() + (bounds.width() - width) / 2.0f;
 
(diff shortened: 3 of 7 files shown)


ghost:
Nice. 
Q: Would this break existing lays?
For example where the width= boxes are wider than the text area due to not being able to see where the boundary is because it's invisible, and the text wasn't stretched due to the issue you mentioned in the old code?
If it is applied automatically then a whole bunch of lays will look very bad until they are fixed ;-)

Edit: Nevermind, I see it's an align="3" option which is new. Ok this solves the no-stretch issue.
Hopefully it gets into mainline rather being ignored like a lot of other cool stuff ;-)

btw, in your example case SD-1 logo thing, the way to do that is remove -1, place SD and place the 1 separately further over. The triangle can be done with a simple svg path that even AI can write in 2 seconds :-)

```
<!-- equilateral triangle with all sides 100 pixels -->
<element name="triangle">
<image state="0"><data><![CDATA[
<svg width="100" height="100" viewBox="0 0 100 100">
  <path d="M0,0 L100,0 L50,86.6 Z" fill="red" stroke="red" stroke-width="2"/>
</svg>
]]></data></image>
</element>
```

cbrunschen:
> Nice. Q: Would this break existing lays? For example where the width= boxes are wider than the text area due to not being able to see where the boundary is because it's invisible, and the text wasn't stretched due to the issue you mentioned in the old code? If it is applied automatically then a whole bunch of lays will look very bad until they are fixed ;-)

The stretching is absolutely not applied automatically - it has to be explicitly requested by specifying `align="3"` in the layout file.

So, no, this will not break any existing layouts.

> Edit: Nevermind, I see it's an align="3" option which is new. Ok this solves the no-stretch issue. Hopefully it gets into mainline rather being ignored like a lot of other cool stuff ;-)
> 
> btw, in your example case SD-1 logo thing, the way to do that 

I'm not trying to recreate the actual SD-1 logo, because that would not be accepted, per [this discussion on the layout for the ESQ-1](https://github.com/mamedev/mame/pull/14049#issuecomment-3235114388). I am simply placing text in the corresponding place, specifically the name by which each keyboard is commonly known; and I want that text to take approximately the same amount of space as the log would. That's all.

(If I did want to put something there that looks a lot closer to the real logos, I have significantly better options available - they just won't be sent for inclusion in MAME because they would be rejected.)

cbrunschen:
Thank you!

felipesanches:
This PR introduced multiple cases of text clipping in many layouts. Here are a few affected layouts:

* kn5000
* virusa
* virusb
* virusc
* virusrck

And here's one before/after comparison (this one is from virusrck):

# Before
Screenshot From 2025-11-30 17-03-49

# After
Screenshot From 2025-11-30 17-01-50

felipesanches:
I'd like to suggest that we could try to figure out a way to automatically calculate the equivalent bound values for text that is affected, so that automatic update of the layouts can be performed to preserve the original rendering output.

felipesanches:
Alternatively, the old method could be restored and possibly the new approach activated only when a given .lay file uses a version number bumped up from 2 to 3. And then authors would use the new version whenever they work on new layouts or when they manually update the existing ones to render properly.

happppp:
I can confirm clipping, @cbrunschen can you fix it? Here's a simple test case

```
<?xml version="1.0"?>
<mamelayout version="2">

	<element name="text0"><text string="CCCDDD" align="0"><color red="1" green="1" blue="1" /></text></element>
	<element name="text1"><text string="CCCDDD" align="1"><color red="1" green="1" blue="1" /></text></element>
	<element name="text2"><text string="CCCDDD" align="2"><color red="1" green="1" blue="1" /></text></element>
	<element name="text3"><text string="CCCDDD" align="3"><color red="1" green="1" blue="1" /></text></element>

	<view name="Test Layout">
		<bounds left="0" right="120" top="0" bottom="60" />

		<element ref="text0"><bounds x="10" y="10" width="20" height="10" /></element>
		<element ref="text1"><bounds x="10" y="20" width="20" height="10" /></element>
		<element ref="text2"><bounds x="10" y="30" width="20" height="10" /></element>
		<element ref="text3"><bounds x="10" y="40" width="20" height="10" /></element>

		<element ref="text0"><bounds x="10" y="50" width="100" height="2" /></element>
		<element ref="text1"><bounds x="10" y="52" width="100" height="2" /></element>
		<element ref="text2"><bounds x="10" y="54" width="100" height="2" /></element>
		<element ref="text3"><bounds x="10" y="56" width="100" height="2" /></element>

	</view>
</mamelayout>
```

[simon.zip](https://github.com/user-attachments/files/23841889/simon.zip)

Screenshot (806)

felipesanches:
Thanks @cbrunschen for the extremely quick fix! This is now looking great ;-)

cbrunschen:
Sorry for the silly mistake - thank you for reporting the issue and  @happppp for help with the fix!

PR #14551, opened 2025-11-19
mz700_cass, mz800_cass: 43 additions and other updates to software lists

paul75s:
22 additions to mz800_cass software list, including 1 clone and 4 items already included in mz700_cass which are not/partially supported on MZ-700 but fully supported on MZ-800.

21 additions to mz700_cass software list, including 1 clone.

4 items in mz800_cass previously marked not working updated to working.

The 4 items in mz700_cass replicated in mz800_cass updated in mz700_cass to partially (3) or not (1) working (previously no supported status set).

Virtually everything here will load on MZ-700 as very little truly MZ-800 specific software is supported by the current state of MZ-800 emulation. MZ-800 will however support PCG, which is not supported by the current state of MZ-700 emulation.

As emulation of both matures, these software lists probably need a re-think, but right now items added to MZ-800 (and any tidy up of supported statuses) reflects the current state of emulation of both systems and feels consistent with existing decisions on which software list to add items to. Largely therefore that means additions to the MZ-800 list are:
- The handful of software which will run on MZ-800 right now but is not supported on MZ-700.
- Software supported by both but which relies heavily on PCG, so is significantly more playable on MZ-800.
- Software which explicitly states it is for MZ-800 (even if it actually runs on MZ-700).

Additions sourced from http://www.mz-800.scav.cz/.

### DIFF
--- a/hash/mz700_cass.xml
+++ b/hash/mz700_cass.xml
@@ -1052,7 +1052,7 @@ Easidata (English Version)
 	<software name="blizzard" supported="yes">
 		<description>Blizzard</description>
 		<year>19??</year>
-		<publisher>Unknown</publisher>
+		<publisher>&lt;unknown&gt;</publisher>
 		<part name="cass" interface="mz_cass">
 			<dataarea name="cass" size="23681">
 				<rom name="blizzard.mzt" size="23681" crc="bf76135c" sha1="d29d5299e70e5a07ee32155750623b0dc8ec6053"/>
@@ -1153,7 +1153,7 @@ Easidata (English Version)
 	</software>
 	
 	<software name="carrier2a" cloneof="carrier2" supported="yes">
-		<description>Carrier ][ (Joystick version)</description>
+		<description>Carrier ][ (joystick version)</description>
 		<year>1984</year>
 		<publisher>BBG Software</publisher>
 		<part name="cass" interface="mz_cass">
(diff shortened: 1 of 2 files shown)


hash/mz700_cass.xml
```diff
@@ -1966,7 +2142,7 @@ Easidata (English Version)
 		</part>
 	</software>
 
-	<software name="nakamoto">
+	<software name="nakamoto" supported="partial">
```
cuavas, line 2147: It would really help the next person if you added a comment saying what’s broken.

hash/mz800_cass.xml
```diff
@@ -38,6 +78,39 @@ MZ-821 Sharp BASIC Trainer (Germany):   Account Manager: Einnahmen/Ausgaben-Verw
 		</part>
 	</software>
 
+	<software name="building" supported="yes">
+		<description>Building Hopper</description>
+		<year>1983</year>
+		<publisher>Unknown</publisher>
```
cuavas: Please use markup if the publisher is unknown, and not a group called “Unknown”.

hash/mz800_cass.xml
```diff
@@ -50,6 +123,50 @@ MZ-821 Sharp BASIC Trainer (Germany):   Account Manager: Einnahmen/Ausgaben-Verw
...
+		<part name="cass" interface="mz_cass">
+			<dataarea name="cass" size="20177">
+				<rom name="galao.mzt" size="20177" crc="fad2a016" sha1="d3e334f7155fbe031a859128bd3b7fb3d08f92a4"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="galaoa" cloneof="galao" supported="yes">
+		<description>Galao (Joystick version)</description>
```
cuavas: Lowercase for descriptive text that isn’t part of title proper.

hash/mz700_cass.xml
paul75s: Added a comment to all the items I've flagged partial to highlight they need PCG emulation to display correctly.

hash/mz800_cass.xml
paul75s: All instances of this amended in both lists.

hash/mz800_cass.xml
paul75s: All instances of this amended in both lists.

PR #14555, opened 2025-11-20
Update copyright year megacrane.cpp

flama12333:
at string:
MEGA_CRANE ELAUT NV 05/01/01
VERSION 0239 REV_0

### DIFF
--- a/src/mame/skeleton/megacrane.cpp
+++ b/src/mame/skeleton/megacrane.cpp
@@ -148,4 +148,4 @@ ROM_END
 } // anonymous namespace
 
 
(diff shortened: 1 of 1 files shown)


stonedDiscord:
That's what the sticker on the EPROM said.
It's possible the ROM was updated to support the switch to Euro.

Other sources say the machine was introduced in 1996
https://clawmachinewiki.fandom.com/wiki/Mega_Crane

angelosa:
So what's the year for this? Alternatively just add a `1997?` with a note?

PR #14559, opened 2025-11-20
mz2000_flop: 11 additions to software list

paul75s:
Add 11 items to mz2000_flop, 8 working,  2 partially working, 1 not working.

These were not working but fixed using MZDiskExplorer.

Additions include:
- 5 compilations of mostly Carry Lab games.
- 2 disks of individual games, Miranda and Uootoy.
- A compilation with a mix of BASIC versions and games.
- A disk based version of MZ-1Z001 BASIC (not MZ-2Z001 disk BASIC).
- A data disk of MZ-1Z002 samples which is not working (or at least unable to test it is) as it requires disk BASIC (presumably MZ-2Z002).

### DIFF
--- a/hash/mz2000_flop.xml
+++ b/hash/mz2000_flop.xml
@@ -309,7 +309,7 @@ FRONT LINE: works
 		</part>
 	</software>
 
-	<software name="ipls2d01" supported="yes">
+	<software name="ipls2d01" supported="partial">
 		<description>Compilation Disk IPLS2D-001</description>
 		<year>200?</year>
 		<publisher>Unknown</publisher>
@@ -343,7 +343,7 @@ PLAZMA LINE: works
 		</part>
 	</software>
 
-	<software name="mz1z001" supported="yes">
+	<software name="mz1z001" supported="partial">
 		<description>BASIC MZ-1Z001 DISK</description>
 		<year>2023</year>
 		<publisher>Sharp</publisher>


hash/mz2000_flop.xml
```diff
@@ -191,4 +191,192 @@ Requires Disk BASIC
...
+Compilation of (mostly) Carry Soft games.
+HIROTON WARS: works
+SPACE CRUISER: works
+LOVE ANT LOVE: works
+FIGHTER 2000: works
+HP OHSHO 2200: works
+BUG FIRE 2200: works
+PETTER.B 2200: works
+POLAR STAR II: works
+AMA TENNIS 2200: works
```
angelosa, line 246: Is this one in color unlike `mz2000_cass:amtennis`?

hash/mz2000_flop.xml
```diff
@@ -191,4 +191,192 @@ Requires Disk BASIC
...
+	</software>
+
+	<software name="ipls2d01" supported="yes">
+		<description>Compilation Disk IPLS2D-001</description>
+		<year>200?</year>
+		<publisher>Unknown</publisher>
+		<notes><![CDATA[
+Compliation of various games and BASIC versions.
+BASIC MZ-1Z001: works
+BASIC SB-5520: not working (probably for MZ-80B?)
```
angelosa, line 319: `supported="partial"` if any item of a given collection fails for any reason.

hash/mz2000_flop.xml
```diff
@@ -191,4 +191,192 @@ Requires Disk BASIC
...
+				<rom name="mz1z001.d88" size="305328" crc="31905cb3" sha1="b39554c8cfb50656b6c37f69ae233f237a472582"/>
+			</dataarea>
+		</part>
+	</software>
+
+	<software name="1z002dta" supported="no">
+		<description>BASIC MZ-1Z002 Data</description>
+		<year>19??</year>
+		<publisher>Unknown</publisher>
+		<notes><![CDATA[Requires disk BASIC - presumably MZ-2Z002]]></notes>
```
angelosa, line 362: These two seems part of the same set. Also 2023?

hash/mz2000_flop.xml
paul75s: Yes, this version supports colour displays, so I also amended amatennis in mz2000_cass to flag compatability is for mono screens only as part of PR 14512.

hash/mz2000_flop.xml
paul75s: They are different at least insofar as information on screen differs (although unclear what, if any functional differences there are).

The version on it's own disk shows:

BASIC interpreter MZ-1Z001 DISK
Copyright 2023 by SHARP Corp.

The version on the compilation disk shows:

BASIC interpreter MZ-1Z001 V1.0D
Copyright 1982 by SHARP Corp.

Both appear to be clones of MZ-1Z001 BASIC in mz2000_cass, but I presume I can't show the standalone disk as a clone of an item in a different software list?

angelosa: Do they respond to a `FILES` inquiry?

paul75s: Neither does. As far as I can see they are cassette BASIC on disk (I think disk BASIC would be MZ-2Z001). I'm updating the compilation disk to show partial support anyway per above comment, think it may be sensible to do the same with the standalone disk.

hash/mz2000_flop.xml
paul75s: Updated to partial. Also done the same with the standalone MZ-1Z001 BASIC disk.

PR #14562, opened 2025-11-21
{vfx,vfxsd,sd1,sd132}.lay: Many cosmetic updates. Show floppy and/or cartridge if loaded.

cbrunschen:
VFX family buttons are actually 15.8mm wide, not 15mm. Regenerated all four layouts.

Before:
Screenshot From 2025-11-21 14-48-03

Real VFX-SD:
image

After:
Screenshot From 2025-11-21 14-47-26

The difference is quite small, especially if you look at the groups of buttons on the right side of the panel.

But if you look at the space between the **Seq** and **0** buttons below the screen, the "After" picture looks a lot more like the real VFX-SD than the "Before" picture.

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -336,9 +336,6 @@ void esq5505_state::cartridge_loaded(bool loaded)
 	// On VFX and later, DUART input bit 1 is 0 for cartridge present.
 	LOG("ip1 -> %d\n", state);
 	m_duart->ip1_w(state);
-
-	// Also upadte the panel.
-	m_panel->set_cartridge_loaded(loaded);
 }
 
 void esq5505_state::cartridge_load(ensoniq_vfx_cartridge *cart)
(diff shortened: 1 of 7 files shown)


cbrunschen:
Now that there is that little bit of extra space beside/between the labels above the buttons, added the double lines above the labels of those buttons that lead to multiple pages of functionality:

image

cbrunschen:
Regenerated with improved vertical label placement:
- there's a bit of space between the top of the button and the label above it
- multi-line labels have less vertical space between them

Positions and sizes of labels and Multi-page markers has been recalculated based on metrics from the "Arimo" fonts, which are metric compatible with Helvetica, Arial and Liberation Sans, and should hopefully look reasonable across MAME's supported platforms.

The pseudo-logo text ("VFX", "VFX-SD", "SD-1", "3 2 V O I C E" and "ensoniq" has also been positioned and sized to fill the appropriate vertical space, and likewise, stretched (using `align="3"`) to fill the space horizontally.

image

image

cbrunschen:
Add code & layout components to show a floppy disk in the drive if one is loaded, and a cartridge in the cartridge slot if one is loaded.

Both empty:

Screenshot From 2025-12-04 16-02-43

Both loaded:

Screenshot From 2025-12-04 16-02-27

cbrunschen:
These layout & Lua script updates significantly improve the UI for the VFX Family of keyboards, including showing cartridge and floppy representations when either is loaded. 
The script updates also include performance improvements to the key color animation that indicates simulated velocity / pressure.

PR #14564, opened 2025-11-21
xavix.cpp (audio tweaks, fix regression in rad_mtrk)

mamehaze:
xavix.cpp (audio improvements) [ramacat]
 - remove the udance hack, it was stopping rad_mtrk going in game, and with the sample mode disabled elsewhere udance does't play garbage anyway
 - provide different 'tempo' defaults depending on machine, it isn't clear what's going on here as some match closest to hardware with 0x40, others with 0x80


mamehaze:
(was set to draft on request as we thought there was a regression with taitons2 and kikikaikai, but it turns out they just have the music in that disabled by default, and it isn't a regression at all, so undrafted it again)

PR #14565, opened 2025-11-21
esqpanel.cpp, esq5505.cpp: change VFX-family buttons from IPT_KEYBOARD to IPT_KEYPAD.

cbrunschen:
Also removes the KEYBOARD_HACK keyboard inputs from the VFX family of keyboards, which now have all buttons (which were triggered by KEYBOARD_HACK inputs) available on the built-in layout front panel.


cbrunschen:
Thank you!

PR #14570, opened 2025-11-22
Update igs_m027.cpp - fix ccly tile ROM loading

mamehaze:


mamehaze:
(github got very confused by this, due to a change in the file while I was working in the web editor)

PR #14571, opened 2025-11-22
Update igs_m027.cpp - fix ccly tile ROM loading

mamehaze:


mamehaze:
I've gone over all the other sets, and I think they're loaded correctly, it was just this one and the fix that already got merged that were wrong

PR #14577, opened 2025-11-23
att6300p.cpp: Add support for 1.2MB drive (525hd)

donohoe00:
The standard configurations for the 6300 Plus have drive A as a 1.2MB floppy, with either a hard drive or a 360K floppy as a secondary drive.  One thing that's unique to the 6300 Plus is that port 0x65 is used for data rate control.

Since the FDC is on the motherboard, isa8_fdc_6300p_device is unique to the 6300 Plus... I'm wondering if it should go in a machine-specific file, instead of in fdc.cpp?

### DIFF
--- a/src/mame/olivetti/att6300p_fdc.cpp
+++ b/src/mame/olivetti/att6300p_fdc.cpp
@@ -22,6 +22,10 @@ void isa8_fdc_6300p_device::device_add_mconfig(machine_config &config)
 	upd765a_device &upd765a(UPD765A(config, m_fdc, 8'000'000, false, false));
 	upd765a.intrq_wr_callback().set(FUNC(isa8_fdc_6300p_device::fdc_irq_w));
 	upd765a.drq_wr_callback().set(FUNC(isa8_fdc_6300p_device::fdc_drq_w));
+
+	// According to "Getting Started with you AT&T 6300 Plus", when the
+	// system came in a two-floppy configuration, drive A (lower drive)
+	// was 1.2MB, and drive B was 360K.
 	FLOPPY_CONNECTOR(config, "fdc:0", att6300p_floppies, "525hd", isa8_fdc_device::floppy_formats).enable_sound(true);
 	FLOPPY_CONNECTOR(config, "fdc:1", att6300p_floppies, "525dd", isa8_fdc_device::floppy_formats).enable_sound(true);
 }


rb6502:
I'd say just move this to the actual driver since it's on the motherboard.  It's small enough that it can probably just go right in the driver, or you can add it as a separate device in mame/olivetti.

donohoe00:
OK, I did the latter - should be good to go.

rb6502:
Almost there.  For motherboard devices added via a slot mechanism, you can use set_fixed() to prevent the user from changing them.  For the most directly applicable example, the Kaypro 16 PC luggable in pc.cpp does this to lock down the ISA slots it uses:

subdevice<isa8_slot_device>("isa1")->set_fixed(true);

donohoe00:
ISA8_SLOT(config, "mb2", 0, m_isabus, att6300p_mb_isa_devices, "fdc_6300p", true);

I thought that's what the last parameter was doing?  If I change it to "false", I can change the device in the "Slot Devices" menu (well, it lets me choose between fdc_6300p or nothing, since there's only one option), otherwise the menu won't let me change it.

galibert:
isa8_slot_device(const machine_config &mconfig, const char *tag, device_t *owner, uint32_t clock, T &&isa_tag, U &&opts, const char *dflt, bool fixed)

Yes, the last bool parameter is the value of "fixed".

src/mame/olivetti/att6300p_fdc.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+}
+
+void isa8_fdc_6300p_device::device_add_mconfig(machine_config &config)
+{
+	upd765a_device &upd765a(UPD765A(config, m_fdc, 8'000'000, false, false));
+	upd765a.intrq_wr_callback().set(FUNC(isa8_fdc_6300p_device::fdc_irq_w));
+	upd765a.drq_wr_callback().set(FUNC(isa8_fdc_6300p_device::fdc_drq_w));
+	FLOPPY_CONNECTOR(config, "fdc:0", att6300p_floppies, "525hd", isa8_fdc_device::floppy_formats).enable_sound(true);
+	FLOPPY_CONNECTOR(config, "fdc:1", att6300p_floppies, "525dd", isa8_fdc_device::floppy_formats).enable_sound(true);
```
angelosa, line 30: Is `525hd` and `525dd` a mistake or the sizes are really uneven by factory default? (I'd add a comment if latter)

donohoe00:
I didn't mean to push that last commit to this PR/branch since it's unrelated.  But the commit is good from my point of view.  Let me know if I should back it out and submit a new PR for it.  Sorry for the confusion.

src/mame/olivetti/att6300p_fdc.cpp
donohoe00: That's the way they are by factory default, according to the User Guide.  I can add a comment.

angelosa:
> I didn't mean to push that last commit to this PR/branch since it's unrelated. But the commit is good from my point of view. Let me know if I should back it out and submit a new PR for it. Sorry for the confusion.

Looks a fairly standard x86 A20 gate control. It's fine as long as it's mentioned in the PR OP message.

src/mame/olivetti/att6300p_fdc.h
```diff
@@ -0,0 +1,30 @@
...
+	virtual void device_start() override ATTR_COLD;
+
+	void map(address_map &map) ATTR_COLD;
+	void rc_map(address_map &map) ATTR_COLD;
+
+    uint8_t rc_r();
+    void rc_w(uint8_t data);
+
+private:
+    int m_rate;
```
cuavas, line 29: `m_rate` is registered for save states.  You should use an explicitly sized integer type for portability.

PR #14586, opened 2025-11-26
xavix - remove system timer disable hack, try some different logic

mamehaze:


MooglyGuy:
LGTM, removing hacks is always a good thing in my book.

PR #14587, opened 2025-11-26
current Plug and Play PR - 1 set + preparation for further ELAN cleanups

mamehaze:
new NOT WORKING machines
-------------
VJ Starz Dance Mat [Team Europe, David Haywood]

- fixed a regression with qix in rad_sinv, which suggests a difference in EU3A05 and EU3A13 behavior when banking
- use subclass for ELAN EU3A13 video as there seem to be some subtle differences, removed hacks in driver


mamehaze:
This is the first stage in refactoring / tidying up the ELAN support.

The software running on these SoCs is quite annoying to work with though, much of it is badly written and assumes default behaviors / states, not writing registers unless they're used etc.

I have experimented with a more extensive clean-up before this, but reverted my progress on that in favour of breaking the work into smaller chunks.

mamehaze:
@ajrhacker can you check this over?  my next steps are getting rid of address bank (to use an address space) and converting the SOCs into devices that derive from the CPU (which is what I'd already started, but decided to not push at this stage)

PR #14589, opened 2025-11-26
Fix VFX and SD1 duart clock

giulioz:
According to the VFX schematics the DUART clock is 5MHz, not 4. Setting it to 5 fixes the demo songs playback speed.

image

The SD-1 32 voices uses 4MHz instead.
image

### DIFF
--- a/src/mame/ensoniq/esq5505.cpp
+++ b/src/mame/ensoniq/esq5505.cpp
@@ -874,6 +874,7 @@ void esq5505_state::sd132(machine_config &config, int panel_type)
 
 	// Like the SD-1 but with its own panel type
 	sd1(config, panel_type);
+	m_duart->set_clock(4000000);
 
 	m_maincpu->set_clock(clock);
 	m_otis->set_clock(clock);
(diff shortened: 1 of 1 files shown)


cbrunschen:
That looks like a great find!

cbrunschen:
I've just verified that this fixes the timing issues on all of the 21-voice VFX-family keyboards! Thank you for finding this!

With this change you should probably also be able to `MACHINE_IMPERFECT_TIMING` markers on lines 1268, 1278 and 1271, marking those, too, as fully working

giulioz:
@cbrunschen 

Nice! Thanks for the double check.
I think though that putting an imperfect audio flag is in order, I did some A/B testing with the original SD-1 and some things sound off. One example is the distortion effect in the `RUDE GUITAR` patch which sounds quite different. It might be a bug in the dsp emulator implementation.

cbrunschen:
... actually, this change alone breaks the timing on the sd132!

Add

`m_duart->set_clock(4000000);`

on line 878 ?

cbrunschen:
These have all recently had the IMPERFECT_SOUND removed because they sound, to me, just like the real thing, including effects, in every thing that I've tried. 

Where can I find the `RUDE GUITAR` sound?

giulioz:
@cbrunschen 

You can load the SD-1 v4.1 SD1-PALETTE demo song (https://forum.vintagesynth.com/viewtopic.php?t=103003) and it will be played there. You can compare it with this: https://www.youtube.com/watch?v=KtV5J6ZLKuU

image

If you want to try the sound alone it's in the first internal page:
image

giulioz:
You can also try the sound "GRIND ORGAN" which also sounds wrong (aha! another one that uses distortion)
image

cbrunschen:
To me, the $D1-PALETTE song, and all others I've tried, sound pretty indistinguishable between the real and the emulated SD-1. Otherwise I would not have proposed marking the sd132 as working.

giulioz:
@cbrunschen I can clearly hear some noise with that patch that differs from the original. I can record an A/B if you want. You can also clearly hear the GRIND ORGAN patch clicking, especially in the low notes.

cbrunschen:
One thing to keep in mind is that the real keyboard of course also has its own digital-to-analog conversion happening, followed by some amount of amplification through several stages of op-amps which may also shape the sound a bit. 

Of course the more precise examples you can give the better.

giulioz:
Absolutely, but I think it's still weird that the distortion effect produces such weird digital noises. I can try to record the digital output of my SD-1 and check how much it differs.

PR #14591, opened 2025-11-27
sd1 (21 voice) BIOS 4.xx now correctly detects speed.

cbrunschen:
This allows the 4.xx BIOS to correctly configure the emulated sd1 for 21 voices and thus play at the correct pitch.

Thus the SD-1 Sequencer OS 4.xx floppies now also apply to the SD-1.

So this also consolidates the `sd1_flop` and `sd132_flop` software lists into a single `sd1_flop` list - but still keeps the `vfxsd_flop` list separate because that is still sufficiently different.

The consolidated `sd1_flop` software list uses
`<sharedfeat name="compatibility" ... />` to allow the different sd1 versions to show only those floppies that apply, respectively: The sd132 does not show the Sequencer OS 3.00 floppy; the sd1 does not show the demo disks.

Arguably this is not quite right because all the demos would at least technically work even on the 21-voice SD-1, in particular those for other keyboards (SQ and KS series) which only use the SD-1's sequencer; but any SD-1/32 demos would run out of voices and thus sound worse than they should if played on an SD-1 (21-voice), so it seems a safer bet to make those only available on the SD-1/32.

(Edited to add:)
This PR also adds the text printed on the labels of the floppies as `<notes>`. I sadly don't have a physical SD-1 Sequencer OS 3.00 floppy so I was not able to include the print from that label, but I expect it would have been very similar to what's on the SD-1 Sequencer OS 4.00 and 4.10 floppies.

### DIFF
--- a/hash/sd1_flop.xml
+++ b/hash/sd1_flop.xml
@@ -49,8 +49,17 @@ P/N 8506000201-A MODEL VSD-201
 
 	<software name="os300" cloneof="os410" supported="yes">
 		<description>SD-1 Sequencer OS version 3.00</description>
-		<year>1990</year>
+		<year>1991</year>
 		<publisher>Ensoniq</publisher>
+		<notes><![CDATA[
+SEQUENCER O.S. DISK - VERSION 3.00
+(REQUIRES ROM O.S. VER. 3.00 OR HIGHER)
+THIS DISK ALSO CONTAINS
+MULTIPLE SOUND PROGRAM FILES,
+PRESET FILES AND DEMO SEQUENCES
+
+P/N 8506000101-B MODEL VSD-200 (C) 1991
+]]></notes>
 
 		<part name="flop" interface="floppy_3_5">
 			<dataarea name="flop" size="819200">


galibert:
I see nothing in that patch that would make a difference in whatever method the bios uses to detect speed.

cbrunschen:
It's [the previous patch to esq5505.cpp](https://github.com/mamedev/mame/pull/14589) that makes the detection code in the 4.xx BIOSes work.

Now that the detection works, this patch simply makes those BIOS ROMs and their matching Sequencer OS floppy disks available on the sd1 (21-voice) as well.

rb6502:
I'm not a fan of rug-pulling an entire software list after one version.  That's one of the reasons I suggested using the compatibility tag initially.  Changes are obviously necessary sometimes, but I don't think we've ever added a list and then deleted it in the next version, which is what's being proposed here.

cbrunschen:
Would you prefer if I restored the `sd132_flop` list, but moved the SD-1 Sequencer OS 4.x floppies to the sd1_flop list?

angelosa:
That's an alternative, if SD-32 has compatibility with `SD-1` then you can use `set_compatible` in former as extra `SOFTWARE_LIST` tag.

```c++
void esq5505_state::sd132(machine_config &config, int panel_type)
[...]
        SOFTWARE_LIST(config, "sd132_flop").set_original("sd132_flop");
	SOFTWARE_LIST(config.replace(), "sd1_flop").set_compatible("sd1_flop");
```

cbrunschen:
Restored the `sd132_flop` software list, which now contains all of the demo disks.  The SD-1 OS disks are now all in the `sd1_flop` list. All disks are made available to all compatible keyboards: there are demos on the earlier disks that are missing from the later ones, but which continue to work.

cbrunschen:
This PR uses the fact that the sd1 (21-voice), with its corrected DUART clock per https://github.com/mamedev/mame/pull/14589 , correctly detects the number of voices when used with the 4.xx BIOS versions, which in turn allow the Sequencer OS 4.xx to run as well. Accordingly I've moved the 4.xx Sequencer OS floppy entries from the `sd132_flop` software list to the `sd1_flop` software list.

It also adds the text from the labels of the actual floppies that I have, as `<notes>` within for the respective floppy entries.

I would greatly appreciate if someone could take another look.

PR #14592, opened 2025-11-28
sega/model2: Various improvements

gm-matthew:
- I have moved `geo_parse()` and the frame number update to the start of vblank; this fixes high score scrolling in `segawski`, 2D palette in `dynamcopc`, and 3D graphics in `bel` not displaying after first booting up.
- Z-Sort mode value uses an exponent point to adjust the z-value scale. Without adjustment, some of the background objects in `srallyc` exceed the maximum exponent and so are not displayed properly.
- I have merged `model2_3d_process_quad()` and `model2_3d_process_triangle()` into a single template function, since most of the code was identical.
- Original Model 2 does have FIFO control; it uses the same 315-5673 buffer controller as Model 2A and 2C.
- Burst stall must be cleared when resetting the i960, otherwise games can potentially hang on soft reset.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -325,9 +325,9 @@ void model2_state::palette_w(offs_t offset, u16 data, u16 mem_mask)
 {
 	COMBINE_DATA(&m_palram[offset]);
 	u16 palcolor = m_palram[offset];
-	u8 r = m_colorxlat[0x0080 / 2 + ((palcolor >> 0) & 0x1f) * 0x100];
-	u8 g = m_colorxlat[0x4080 / 2 + ((palcolor >> 5) & 0x1f) * 0x100];
-	u8 b = m_colorxlat[0x8080 / 2 + ((palcolor >> 10) & 0x1f) * 0x100];
+	u8 r = m_colorxlat[(0x0080 >> 1) + (((palcolor >> 0) & 0x1f) << 8)];
+	u8 g = m_colorxlat[(0x4080 >> 1) + (((palcolor >> 5) & 0x1f) << 8)];
+	u8 b = m_colorxlat[(0x8080 >> 1) + (((palcolor >> 10) & 0x1f) << 8)];
 	r = m_gamma_table[r];
 	g = m_gamma_table[g];
 	b = m_gamma_table[b];
@@ -485,9 +485,9 @@ u32 model2_tgp_state::copro_sincos_r(offs_t offset)
 	offs_t ang = m_copro_sincos_base + offset * 0x4000;
 	offs_t index = ang & 0x3fff;
 	if (ang & 0x4000)
-		index = std::min(0x4000 - (int)index, 0x3fff);
+		index = std::min(0x4000 - int(index), 0x3fff);
 	u32 result = m_copro_tgp_tables[index];
-	if(ang & 0x8000)
+	if (ang & 0x8000)
 		result ^= 0x80000000;
 	return result;
 }
@@ -521,7 +521,7 @@ u32 model2_tgp_state::copro_isqrt_r(offs_t offset)
 	u8 bexp = (m_copro_isqrt_base >> 24) & 0x7f;
 	u8 exp = (result >> 23) + (0x3f - bexp);
 	result = (result & 0x807fffff) | (exp << 23);
-	if(!(offset & 1))
+	if (!(offset & 1))
 		result &= 0x7fffffff;
 	return result;
 }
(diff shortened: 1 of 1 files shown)


happppp:
Note that moving geo_parse to the start of scanline 384 means it's after screen_update. So 1 frame delay compared to before, is that something you intended?

angelosa:
Generally speaking a renderer displays the previously prepared framebuffer and swaps drawing at vblank time, it's pretty much settled that at least one frame of lag should happen (in this case there's also a 30 Hz setting that iirc most games uses).

gm-matthew:
> Note that moving geo_parse to the start of scanline 384 means it's after screen_update. So 1 frame delay compared to before, is that something you intended?

The emulated screen for Model 2 uses the attribute `VIDEO_UPDATE_AFTER_VBLANK` so `screen_update_model2()` runs after scanline 424, not scanline 384.

When we eventually emulate the geometry DSP to get Last Bronx and Top Skater working properly it will involve the geometry DSP processing the T&L for frame N while the i960 and copro are processing the game logic for frame N+1, so an extra frame of latency will be inevitable.

angelosa:
> The emulated screen for Model 2 uses the attribute `VIDEO_UPDATE_AFTER_VBLANK` so `screen_update_model2()` runs after scanline 424, not scanline 384.

That should be renamed to just `screen_update`, if you don't mind :)

gm-matthew:
> That should be renamed to just `screen_update`, if you don't mind :)

Done :-)

I also renamed `screen_vblank_model2()` while I was at it; I considered just removing it since it is unused, but then again perhaps we could migrate the code from `model2_interrupt()` over to `screen_vblank()` and hook it up accordingly.

src/mame/sega/model2_v.cpp
```diff
@@ -166,10 +166,10 @@ static inline void apply_focus( model2_state::geo_state *geo, poly_vertex *p0)
...
 
 /* 1.8.23 float to 4.12 float converter, courtesy of Aaron Giles */
-inline u16 model2_state::float_to_zval( float floatval )
+inline u16 model2_state::float_to_zval( float floatval, s32 z_adjust )
 {
-	int32_t fpint = f2u(floatval);
-	int32_t exponent = ((fpint >> 23) & 0xff) - 127;
+	s32 fpint = f2u(floatval);
+	s32 exponent = ((fpint >> 23) & 0xff) - ((z_adjust >> 23) & 0xff);
 	u32 mantissa = fpint & 0x7fffff;
```
cuavas, line 173: I know I’m being annoying, but can you change the functions to use “English” rather than “European” spacing around parentheses when your touching thing?  It would be nice to eventually get consistency on this.  (`foo(T a)` rather that `foo( T a )`)

src/mame/sega/model2_v.cpp
```diff
@@ -322,7 +322,8 @@ u32 model2_state::polygon_count_r()
  *
  *******************************************/
 
-void model2_state::model2_3d_process_quad( raster_state *raster, u32 attr )
+template <unsigned NumVerts>
+void model2_state::model2_3d_process_polygon( raster_state *raster, u32 attr )
```
cuavas: Philosophically, since this only supports triangles and quads, would it be better to just use a `bool` to indicate one way or the other?  Or should there be a `static_assert` at the top of the body to ensure the number is supported?

angelosa:
> > That should be renamed to just `screen_update`, if you don't mind :)
> 
> Done :-)
> 
> I also renamed `screen_vblank_model2()` while I was at it; I considered just removing it since it is unused, but then again perhaps we could migrate the code from `model2_interrupt()` over to `screen_vblank()` and hook it up accordingly.

Believe it was made that way to mimic what `segas24_state` does, however Model 2 shouldn't use horizontal sync events and `configure_scanline` is pretty much a bad abstraction anyway (namely it doesn't really control the scanline composition, throws line events at the start of the display period), and if it happens it would need a custom `emu_timer`.

So yes for the conversion but make sure that nothing breaks in the process.

src/mame/sega/model2_v.cpp
gm-matthew: I've gone ahead and removed all the excess spaces between parentheses in model2_v.cpp.

src/mame/sega/model2_v.cpp
gm-matthew: Added a `static_assert` to the top of the function.

gm-matthew:
> Believe it was made that way to mimic what `segas24_state` does, however Model 2 shouldn't use horizontal sync events and `configure_scanline` is pretty much a bad abstraction anyway (namely it doesn't really control the scanline composition, throws line events at the start of the display period), and if it happens it would need a custom `emu_timer`.
> 
> So yes for the conversion but make sure that nothing breaks in the process.

I left it with `VIDEO_UPDATE_AFTER_VBLANK` since the System 24 tilemap generator latches the scroll registers during the last scanline before active display, or just before vertical blanking ends.

I've tested with the vblank code moved from `model2_interrupt()` to `screen_vblank()` and it works as expected; vblank IRQs are asserted at the correct time.

PR #14593, opened 2025-11-28
nwk-tr.cpp: Wheel FFB outputs for Thrill Drive, Racing Jam 1/2

Endprodukt:
Hooks outputs for Thrill Drive, RJ 1+2

### DIFF
--- a/src/mame/konami/nwk-tr.cpp
+++ b/src/mame/konami/nwk-tr.cpp
@@ -192,9 +192,9 @@ Notes:
      TEXELFX - 3DFX 500-0004-02 BD0665.1 TMU (QFP208)
      PIXELFX - 3DFX 500-0003-03 F001701.1 FBI (QFP240)
       001604 - Konami Custom (QFP208)
-       MC44200FT - Motorola MC44200FT 3 Channel Video D/A Converter (QFP44)
+   MC44200FT - Motorola MC44200FT 3 Channel Video D/A Converter (QFP44)
      MACH111 - AMD MACH111 CPLD (Stamped '03161A', PLCC44)
-    PLCC44_SOCKET - empty PLCC44 socket
+   PLCC44_SOCKET - empty PLCC44 socket
       AV9170 - Integrated Circuit Systems Inc. Clock Multiplier (SOIC8)
       AM7201 - AMD AM7201 FIFO (PLCC32)
         PAL1 - AMD PALCE16V8 (stamped 'N676B4', DIP20)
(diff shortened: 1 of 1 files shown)


happppp:
Do you have a better name for this? I assume you copy pasted it from pcbdigit, which are 7segs mounted on the PCB. What do these outputs do, name it after that? If it's 8 separate bits, best separate the outputs too, and if it's 1 thing combined, eg. motor strength (just guessing), you won't need an array, ```output_finder<>``` instead.

Endprodukt:
changed name to "wheel" but removing the array will not work. At least for me it wont compile. 

it's force feedback outputs for the wheel motor.

happppp:
IMO wheel_motor is a better name then. Is it all 8 bits controlling a single thing, or separate bits controlling separate parts of the motor? Or even multiple mechanical elements.

How to do a single output finder:
```
m_my_output(*this, "abcd")
output_finder<> m_my_output;
m_my_output.resolve();
m_my_output = someval;
```

Endprodukt:
Thank you very much for explaining. It's very much appreciated as I'm a noob :)

Hydreigon223:
U.S. Konami cabs were known for using a Happ Motor Controller pcb. I'm aware Konami did make their own FFB board too but the protocol remains unchanged from Happ's motor pcb with two added functions. This kind of protocol was used in midnrun, windheat, gticlub, racingj, thrilld, code1d, gticlub2, thrilld2 and xtrial.

The standard protocol consists of a 5-bit port. Bits 0-3 determine the wheel strength and bit 4 determines determines the wheel direction (0 = left, 1 = right). Writing a non-zero value to bits 0-3 enables the motor.

Some Konami games set either bits 5 or 7 depending on the game. Bit 7, used by all Konami games, supposedly enables the motor. Since writing a non-zero value to bits 0-3 also enables the motor, the use for this function is situational. Bit 5, used in mindrun, windheat and gticlub, supposedly centers the motor. While this could have been useful, most newer games don't ever set this bit.

Eventually, this kind of FFB protocol should be made into a device because of maxspeed and ftspeed in dc_atomiswave.cpp utilizing this in addition to the previous Konami games mentioned. Later PC based arcade games way out of MAME's reach like Chase HQ2 and Ford Racing Full Blown would also use this protocol.

Edit, maxspeed and ftspeed would arrange the bits differently with bit 0 assigned to the wheel direction and bits 1-4 assigned to wheel strength.

happppp:
Ok, I guess the thing that receives the output can mask the relevant bits.
Out of curiosity, did you try it with any force feedback device? Maybe with this? https://github.com/Endprodukt/FFBArcadePluginMAME

Endprodukt:
of course, what you see there is my version of ffbplugin. works perfectly with ffb wheel.

Issue #14600, opened 2025-11-30
emu.register_periodic blocked on Windows when in MAME debugger


taylorza:
### MAME version

0.283 (LLP64)

### System information

Windows 11 Pro 10.0.26200
64 GB RAM
BGFX
English

### INI configuration details

```INI

```

### Incorrect behaviour

Callback functions registered in LUA scripts/plugins using `emu.register_periodic` are not triggered regularly while MAME is stopped in the MAME Debugger. The callback is triggered on windows inputs such as mouse movement. On Linux the callbacks are invoked regularly regardless of the debugger state.

### Expected behaviour

Periodic events should be triggered regardless of the MAME debugger state.

### Steps to reproduce

On Windows:
1. Create a simple LUA script to register a periodic function that logs to the console to demonstrate the issue
2. Launch MAME with the script using the -script and -debug option
3. Notice that the LUA script is logging to the console fairly rapidly
4. Break into the debugger
5. Notice that the logging either slows down or stops depending one where the mouse is. Regardless of the mouse location, moving the mouse will kickstart the messages being logged but stop/slowdown as soon as you stop moving the mouse.
6. Resume from the debugger and the messages start to flow again

The issue is in debugwin.cpp `GetMessage` is blocking waiting for a windows message to arrive, this blocks the loop that generates calls `emulator_info::periodic_check();`

taylorza:
I created a new pull request to address this issue
https://github.com/mamedev/mame/pull/14603

PR #14601, opened 2025-11-30
Service periodic callbacks on Windows while MAME debugger is active

taylorza:
Ensure that LUA periodic callbacks continue to be regularly serviced on Windows builds while the MAME debugger is in a `stop` state.

### DIFF
--- a/src/osd/modules/debugger/debugwin.cpp
+++ b/src/osd/modules/debugger/debugwin.cpp
@@ -206,23 +206,28 @@ void debugger_windows::wait_for_debugger(device_t &device, bool firststop)
 
 	// get and process messages
 	MSG message;
-	GetMessage(&message, nullptr, 0, 0);
 
-	switch (message.message)
+	// Crude implementation of GetMessage with timeout
+	if (MsgWaitForMultipleObjects(0, nullptr, false, 16, QS_ALLINPUT) == WAIT_OBJECT_0)
 	{
-	// check for F10 -- we need to capture that ourselves
-	case WM_SYSKEYDOWN:
-	case WM_SYSKEYUP:
-		if (message.wParam == VK_F4 && message.message == WM_SYSKEYDOWN)
-			SendMessage(GetAncestor(GetFocus(), GA_ROOT), WM_CLOSE, 0, 0);
-		if (message.wParam == VK_F10)
-			SendMessage(GetAncestor(GetFocus(), GA_ROOT), (message.message == WM_SYSKEYDOWN) ? WM_KEYDOWN : WM_KEYUP, message.wParam, message.lParam);
-		break;
+		if (PeekMessage(&message, nullptr, 0, 0, PM_REMOVE)) {
+			switch (message.message)
+			{
+				// check for F10 -- we need to capture that ourselves
+				case WM_SYSKEYDOWN:
+				case WM_SYSKEYUP:
+					if (message.wParam == VK_F4 && message.message == WM_SYSKEYDOWN)
+						SendMessage(GetAncestor(GetFocus(), GA_ROOT), WM_CLOSE, 0, 0);
+					if (message.wParam == VK_F10)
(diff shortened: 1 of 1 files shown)


MooglyGuy:
Change looks alright, but "Fix for #14600" isn't a particularly descriptive message. Could I convince you to find your words, describing your change and the rationale behind it, so that I can merge it?

src/osd/modules/debugger/debugwin.cpp
```diff
@@ -206,23 +206,28 @@ void debugger_windows::wait_for_debugger(device_t &device, bool firststop)
...
-		if (message.wParam == VK_F10)
-			SendMessage(GetAncestor(GetFocus(), GA_ROOT), (message.message == WM_SYSKEYDOWN) ? WM_KEYDOWN : WM_KEYUP, message.wParam, message.lParam);
-		break;
-
-	// process everything else
-	default:
-		winwindow_dispatch_message(*m_machine, message);
-		break;
+		if (PeekMessage(&message, nullptr, 0, 0, PM_REMOVE)) {
+			switch (message.message)
```
cuavas, line 214: Doesn’t this naïve code break if you get multiple messages queued?  From the documentation for `MsgWaitForMultipleObjects`:

> **MsgWaitForMultipleObjects** does not return if there is unread input of the specified type in the message queue after the thread has called a function to check the queue. This is because functions such as [PeekMessage](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-peekmessagea), [GetMessage](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/fax/-mfax-faxaccountincomingarchive-getmessage-vb), [GetQueueStatus](https://learn.microsoft.com/en-us/windows/desktop/direct3d10/id3dx10threadpump-getqueuestatus), and [WaitMessage](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-waitmessage) check the queue and then change the state information for the queue so that the input is no longer considered new. A subsequent call to **MsgWaitForMultipleObjects** will not return until new input of the specified type arrives. The existing unread input (received prior to the last time the thread checked the queue) is ignored.

Also:
* What actually triggers the periodic check with the Qt debugger on Linux when waiting for the debugger?
* What happens with the Qt debugger on Windows?
* What about the Cocoa debugger on macOS?  (Yes, I know it interacts poorly with SDL.)
* What about the ImGUI debugger?
* How did you choose the timeout?
* The formatting doesn’t match the surrounding code.

taylorza:
@cuavas and @MooglyGuy - I hit a snag on my end and inadvertently deleted the head repository for this pull request. I resubmitted the pull request. 

Apologies for the inconvenience!

src/osd/modules/debugger/debugwin.cpp
janko-jj: I can answer this question: "What actually triggers the periodic check with the Qt debugger on Linux when waiting for the debugger?"

The current QT implementation has ( src/osd/modules/debugger/debugqt.cpp#L174 ):

```C
	osd_sleep(osd_ticks_per_second() / 1000 * 50);
	qApp->processEvents(QEventLoop::AllEvents, 1);
```

Note that this handling is also suboptimal and probably different from what was expected: the above sleep constant results **only in 20 calls per second** observed, as the sleep time between two possible calls is whole 50ms. That is even less frequent than while the guest machine runs (when around 50 calls per second happen!) in spite of the fact that the guest machine doesn't run and the time is spent on the long sleeps instead.

This and more is in: #15272 where it is demonstrated how periodic calls can completely stop both on Windows and on Mac.

Regarding topic of "choosing the timeout": I'd expect that on all platforms at least 50 calls per seconds should be reasonable. If all platforms achieve 50 calls per second when the emulated machine runs, there's no reason to so drastically limit the number of calls when the emulated machine doesn't run (which is when the debugger window opens and these "no calls" times happen).

PR #14603, opened 2025-11-30
Service periodic callbacks on Windows while MAME debugger is active - Issue #14600

taylorza:
When the MAME debugger is active and emulator in the stopped state, LUA periodic callbacks are not serviced regularly. This fix ensures that the callbacks continue to be invoked and aligns with the LINUX build.

This is specific to Windows using the Windows Native UI for the MAME Debugger.


holub:
@cuavas he got rid of MsgWaitForMultipleObjects in this follow up. Do you still think the fix is problematic?

cuavas:
> @cuavas he got rid of MsgWaitForMultipleObjects in this follow up. Do you still think the fix is problematic?

But now he made it not actually wait at all.  That’s going to cause its own set of problems.

PR #14605, opened 2025-11-30
skeleton/st25.cpp: New systems marked not working

stonedDiscord:
---
Blizzard [stonedDiscord]
Tango [stonedDiscord]
Tobago [stonedDiscord]
Ballermann 6 [stonedDiscord]
Unimint Arena [stonedDiscord]
City Tower [stonedDiscord]
Jamaica [stonedDiscord]
Majesto [stonedDiscord]
Pur Pur Royal [stonedDiscord]
Robin [stonedDiscord]
Big Action 3000 E [stonedDiscord]
Super Take Off E [stonedDiscord]
Booster Speed [stonedDiscord]
Colossos [stonedDiscord]
Matrixx [stonedDiscord]
Avanti [stonedDiscord]
Rotamint Macao [stonedDiscord]
Multiclassic [stonedDiscord]
Rotamint Alpha [stonedDiscord]
Super Pasch [stonedDiscord]
Big Kick [stonedDiscord]
Galaktica [stonedDiscord]
Amarillo [stonedDiscord]
Xeno [stonedDiscord]


src/mame/skeleton/st25.cpp
```diff
@@ -0,0 +1,491 @@
+// license:BSD-3-Clause
+// copyright-holders: stonedDiscord
```
MooglyGuy: No need for a space between the colon and the contributor name. Seems to be inconsistent in the codebase itself, but I'm using some of Vas's own devices/drivers for reference.

src/mame/skeleton/st25.cpp
```diff
@@ -0,0 +1,491 @@
...
+
+/*
```
MooglyGuy: Would it be possible to turn this into the more-common comment block format (with a forward-slash and a bunch of asterisks which can be found in many drivers/devices)?

src/mame/skeleton/st25.cpp
```diff
@@ -0,0 +1,491 @@
...
+{
+	map(0x00000, 0x1ffff).ram();
+	map(0x20000, 0x7ffff).rom().region("maincpu", 0x20000);
+	map(0x80000, 0xeffff).ram();
+	map(0xfc000, 0xfffff).rom().region("maskrom", 0);
+}
+
+void st25_state::io_map(address_map &map)
+{
+	//OUT
```
MooglyGuy: If these aren't hooked up yet, it would be better to describe the not-yet-hooked outputs in a comment block at the top of the function, rather than a series of commented-out lines.

MooglyGuy (changes requested):
Some fairly minor things that would be nice to have before this goes in - otherwise, looks great!

src/mame/skeleton/st25.cpp
stonedDiscord: I think this makes it easier to toggle them on/off during development.
0x3000 is not commented out because it spams the machine log really bad and I've been using the comments like that when I was trying to figure things out (didn't get very far).

src/mame/mame.lst
```diff
@@ -44201,6 +44201,32 @@ sols5e
...
+majesto
+matrixx
+multiclassic
+purpurr
+robin
+stakeoffe
+superpasch
+tango
+tobago
+xeno
```
cuavas, line 44414: Please try to keep the short names of parents within eight characters when it can be done reasonably.  The point of relaxing the limit wasn’t to encourage “shortname inflation”, it was to give more breathing space for cases where there’s e.g. a prefix for the platform eating up several characters (like PlayChoice 10).

PR #14606, opened 2025-11-30
vtech_innotab_cart.xml: Add a list of games

vtarchive:


hash/vtech_innotab_cart.xml
```diff
@@ -11,8 +11,85 @@ license:CC0-1.0
 
 	<!-- pinout is identical to Storio cartridges -->
 
-	<software name="mmousec" supported="no">
-		<description>Mickey Mouse Clubhouse (UK)</description>
+
+	Language:
+	********0 = US / Canada / Hong Kong
+	********1 = Unused
```
angelosa: This won't pass validation, it seems a comment block.

vtarchive: It removed

angelosa: Again, that's not what I meant, and it's not how XML works. If you don't enclose comment blocks with `<!-- whatever your text is ... -->` it won't work.

vtarchive: ok thanks

vtarchive (commented):
i have to add a entire list of innotab games

vtarchive (commented):
I have added full InnoTab list

PR #14608, opened 2025-11-30
Update goldstar.cpp add dsw settings for cb3c

jbtx77gr:
- add dsw settings for cb3c


cuavas:
How are you supposed to get past the “INIT MACHINE” screen?  The machine is marked working, but when I attempt to run it, it just sits at this screen, and pushing all the buttons doesn’t seem to help.

jbtx77gr:
@cuavas When the “INIT MACHINE” message appears on the screen, press the Confirm and Record buttons at the same time. A second screen will show up — then press Key-Out to enter the game.
You only need to do this the very first time you start the game.

Robbbert:
There is no Confirm or Record buttons listed in the inputs.

However, holding 9 and 0 together for a moment shows a screen with a lot of FFFF. Now press W.

RobertoFresca:
Yep... Confirm/Record seems the names listed in the manual. These are the two service buttons 9 & 0.  Keyout is the W.

PR #14618, opened 2025-12-01
Fix floating value options when Windows is set to certain languages

AnthonyRyuki:
If Windows is set to those languages where floating values are displayed with a ',' instead of a '.' (such as Italian and German), the options in General Settings->Advanced Options with a floating value do not work correctly and they can only display either 1 or 0, with the latter value not allowing any changes, so those values can only be changed by modifying them in the mame.ini file.

This is a fix for that.

### DIFF
--- a/src/osd/windows/winmain.cpp
+++ b/src/osd/windows/winmain.cpp
@@ -172,7 +172,7 @@ static int is_double_click_start(int argc);
 
 int main(int argc, char *argv[])
 {
-	std::setlocale(LC_ALL, "");
+	std::setlocale(LC_ALL, "en_US.UTF-8");
 	std::vector<std::string> args = osd_get_command_line(argc, argv);
 
 	// use small output buffers on non-TTYs (i.e. pipes)


rb6502:
I would imagine globally forcing the app locale like this is going to have bad side effects on non-US English operating systems.  options.cpp is using iostreams with `imbue(std::locale::classic())` which as far as I can tell should already prevent this from happening.

cuavas:
Hard “no” on this.  MAME should respect the user’s locale settings for sort order, number formats, etc.  It currently does on Windows and macOS.  Breaking that is regressive.

AnthonyRyuki:
How about just leaving that line alone and adding this line underneath, so it only changes the decimal dot?

`std::setlocale(LC_NUMERIC, "en_US.UTF-8");`

EDIT: Never mind, that wouldn't actually fix the bug.

cuavas:
There was an actual bug in src/lib/util/options.cpp where it converted float values to strings with the global locale, but validated and parsed them with the “classic” locale.  It should be fixed by 6907b0eb4dfcb0ad3b0cb17c86923f4d56608771.  There may be additional bugs on top of that, but you could see whether that fixes the issue.

cuavas:
5211c22324242d1d82a94034af5952f9ea40988c fixes more issues in the UI layer.

AnthonyRyuki:
The issue is now fixed, thank you very much.

PR #14619, opened 2025-12-01
Namcos22 hook up outputs, name changes

Endprodukt:
Namcos22 outputs, name changes, based on SailorSats work.

### DIFF
--- a/src/mame/namco/namcos22.cpp
+++ b/src/mame/namco/namcos22.cpp
@@ -3634,6 +3634,7 @@ void namcos22_state::machine_start()
 	m_mcu_out.resolve();
 	m_cpuled_out.resolve();
 	m_wheel_motor.resolve();
+
 	m_portbits[0] = 0xffff;
 	m_portbits[1] = 0xffff;
 	m_dsp_upload_state = NAMCOS22_DSP_UPLOAD_READY;


happppp:
The existing m_mcu_out can be used here for the single-bit System22 outputs (use a for loop with BIT to set the 8 outputs).

Endprodukt:
So just take care of the mcu_out but you are okay with the wheel stuff?

happppp:
If it's like the Konami games where the thing that polls the output from MAME can mask the relevant bits, yes I'm ok with that.

Endprodukt:
How about this

u8 out0 = m_shareram[0x20/2] & 0xff;
u8 out1 = m_shareram[0x42/2] & 0xff;

for (int i = 0; i < 8; i++)
    m_mcu_out[i] = BIT(out0, i);

for (int i = 0; i < 8; i++)
    m_mcu_out[8 + i] = BIT(out1, i);

I'm really new to this stuff and just want to help.

happppp:
Yeah, but you only need 1 loop.

```
// lamps, coin counters, and also leds for acedrive and victlap
u16 outputs = m_shareram[0x42/2] << 8 | (m_shareram[0x20/2] & 0xff);
```
then loop through 16 bits.

Endprodukt:
@happppp done, let me know if this works for you :)

PR #14620, opened 2025-12-02
sinclair/specnext_dma.cpp: Avoided 'Unknown base register' error

holub:


galibert:
Is there something that actually filters the write?

holub:
I'm reconsidering the solution.

PR #14621, opened 2025-12-02
mame/luaengine.cpp: Access machine.side_effects_disabled

holub:


cuavas:
Is it a good idea to implement this as a property, given it’s likely to be used in fairly hot handlers?

I really don’t like things going in without consideration when external stuff will start to depend on it, and then deprecating or changing it becomes a major pain.

holub:
This PR expected to drive this discussion because the other places we tried didn't trigger anyone.

How do you see this to be redone the right way?

cuavas:
> This PR expected to drive this discussion because the other places we tried didn't trigger anyone.

Have you considered that I’ve already got a mountain of other crap to deal with, and it takes time to get to stuff?  For example writing up #14624 properly, which is a fairly serious regression that’s *already been festering for months*, was fairly time-consuming.  I really feel like I’m fighting an losing battle with stuff being broken and things marked working when they clearly don’t.  No-one likes it if I revert things or just go marking things not working either.

Also, I avoid the bannister forum because every time I look at it, there’s someone mouthing off about something they have no clue about.  It’s just not worth trying to deal with.

> How do you see this to be redone the right way?

The first thing to remember about properties is that they get instantiated for every instance – you can’t have a single one in the metatable to use for all instances of a class.  That’s important for object that you might have a large number of instances of, but isn’t really a big deal for the machine class since I’d like to hope that you don’t create a large number of instances of it.  (Each time you invoke `manager.machine`, you get a separate proxy object instance, but if you keep one in a Lua variable you can re-use it for the entire session.)

But consider:

```lua
local m = manager.machine

...

if not m.side_effects_disabled then
    ...
end
```

Now every time it hits `m.side_effects_disabled` it needs to do a hash table lookup to find the property to invoke it.

You can naïvely use a method like this:

```lua
local m = manager.machine

...

if not m:side_effects_disabled() then
    ...
end
```

But that doesn’t gain you anything.  The trick is that you can look up the method and store it in a local variable (locals are accessed by register index):

```lua
local m = manager.machine
local sed = m.side_effects_disabled

...

if not sed(m) then
    ...
end
```

Now you’ve avoided a hash table lookup on the critical path.  If you have multiple instances of the same class, you can use the same local reference to the method on all of them.

And now I’m even further behind on trying to get through the mountain of stuff that’s already been dumped on me…

holub:
Vas, I really appreciate what you are doing and happy to help if it's up to me. Don't even try to think I'm trying to push dirt on purpose.

The plan is (correct me if I'm wrong) rewrite this as a : method and share the last example as a best practice in the doc?

cuavas:
Yeah, I think changing it to a method probably a good idea.

I don’t think we have a good place for those kinds of tips in the docs.  In this case, it’s not really a MAME-specific thing, just a thing you need to consider with Lua scripting in general.  Would it be sensible to make a Lua scripting tips page on the MAMEdev wiki with examples of various things?

holub:
Maybe we do this page right here in this repo on mamedev's github?
As for me wiki is already less convenient for community editing as required extra registration.

cuavas:
I don’t really think MAME’s documentation is the rightt place for general Lua optimisation tips, though.

holub:
I'm suggesting wiki in-place not MAME's doc.
Like here: https://github.com/redcode/ZXSpectrum/wiki

galibert:
Honestly, if you put a lua-level tap in a path that's so hot that the call to test for side effects is noticeable, you had a problem before doing the call.  The lua scripting is not supposed to be performance critical in the first place.

holub:
People want to use MAME as a development platform. It's hard to bring useful platform-specific tooling to our debugger.
Lua looks like a solution and relief for them and I think performance drop is something they can accept in exchange to this opportunity.

cuavas:
> Honestly, if you put a lua-level tap in a path that's so hot that the call to test for side effects is noticeable, you had a problem before doing the call. The lua scripting is not supposed to be performance critical in the first place.

Yeah, but I’d rather not leave performance gains on the table.  Avoiding a hash table lookup on a hot path is better than not avoiding it.

PR #14623, opened 2025-12-02
Fix riscv64 build

belegdol:
This removes the first hurdle towards having mame compile on riscv64 again.

### DIFF
--- a/makefile
+++ b/makefile
@@ -397,6 +397,10 @@ ifeq ($(findstring aarch64,$(UNAME)),aarch64)
 ARCHITECTURE :=
 endif
 
+ifeq ($(findstring riscv64,$(UNAME)),riscv64)
+ARCHITECTURE :=
+endif
+
 # Autodetect BIGENDIAN
 # MacOSX
 ifndef BIGENDIAN


cuavas:
What problem is having `ARCHITECTURE` set causing?  Are you adding `PTR64=1` to the build options to force it to see that the unrecognised architecture has 64-bit pointers?

belegdol:
Having `ARCHITECTURE` set results in the build system trying to pass `-m32` to the compiler.
I can retest with `PTR64=1` later. I just have a qemu VM available though so my testing capacity is very limited as it is very slow.

rb6502:
ARCHITECTURE should generally be set to _x86 for 32-bit architectures and _x64 for 64-bit (this is exactly what PTR64 does).  Yes, that naming is legacy cruft at this point, but it does work for AArch64 so it should work for RISC-V.

belegdol:
> it does work for AArch64

https://github.com/mamedev/mame/blob/20b4455da6142e9d7ef77d803f4095b3bd434e0c/makefile#L396-L398

I vaguely recall that this is due to genie insisting on adding  `-m32` resp. `-m64` for _x86 and _x64 targets.

cuavas:
This is kind of annoying.  For a while, `-m32` and `-m64` were supported options for all the platforms that mattered that had 32-bit and 64-bit variants (x86, PowerPC, SPARC, etc.) but it seems that those options aren’t supported, at least on Linux, for newly added architectures.

How do you actually build 32-bit executables on 64-bit ARM or RISC-V Linux, anyway?  Or do they not even support multilib?

rb6502:
Debian, Fedora, and SuSE all only support riscv64.

belegdol:
Just for reference, setting `ARCHITECTURE` to nil was the solution before:
https://github.com/mamedev/mame/commit/91a9904a84703d57cd649c9c66a4bea446298b2b
https://github.com/mamedev/mame/commit/fe162f615b8e5639e5255abe1522bb6a7bb81510

I agree with @cuavas that this is somewhat annoying, it appears that gcc only supports `-m32` and `-m64` on platforms where you can actually build a binary of a different bitness. It was never supported on `armv7hl` or `aarch64`:
https://github.com/mamedev/mame/commit/fd5fb05aceeea0822ae268a8af0ce54001c244f8.
I also recall it was mighty annoying to figure out at which step of the build process these flags are getting inserted. If I get this right, for gcc genie generates a 32, 64 and native debug and release variants, without too much care whether the generated platform actually supports `-m32` on `-m64`:
https://github.com/mamedev/mame/blob/9b3478f25ddd55b30b204d8af30af96a43606cf0/3rdparty/genie/src/tools/gcc.lua#L67-L77
Six targets (debug and release for each) are then created in the generated makefiles, and it is up to whoever calls the generated makefiles to figure out which target is correct.

belegdol:
Looks like this is actually a very old problem:
https://github.com/mamedev/mame/commit/d4ec50198c75fea966ce23f3c02ee5cf61a9cdfe

cuavas:
The process for generating the build files is kind of ugly, and basically on life support.  I want to shift to CMake, but in doing so I don’t want to carry over the current problems, and I don’t want an approach that attempts to make CMake look like GENie/premake as that largely defeats the purpose.  Given how big MAME is, it’s a big projects, and it’s hard to get a solid chunk of time to work on a big project with the amount of fires that need to be fought.

cuavas:
I’ll merge this for now – I’ll take this over the other one as I’d like to see where we’re getting the warnings about alignment requirements increasing for casts.  That really will bite us if we just keep suppressing the warning for new targets.  We did have MAME building for SPARC with the warning enabled before Sun ran out of steam.

cuavas:
Well, I started looking at those alignment warnings in 3071a892c39ff60437f1f1a959f2a08a9cca941d, and it's going to be a long road to sorting out this mess.

PR #14626, opened 2025-12-02
mame/luaengine.cpp: Refactored machine:side_effects_disabled() as a method instead of property.

holub:
#14621 followup


ajrhacker:
Might I ask what the advantages are of doing this, besides consistency of syntax?

holub:
> Might I ask what the advantages are of doing this, besides consistency of syntax?

see https://github.com/mamedev/mame/pull/14621#issuecomment-3603132901

PR #14627, opened 2025-12-02
z80dma.cpp: Relaxed fatal error which were used for unknown commands catching.

holub:


holub:
I believe this fatal stayed enough to catch any commands we suspected MAME is missing. Time to be less destructive for the user.

src/devices/machine/z80dma.cpp
```diff
@@ -850,10 +850,11 @@ void z80dma_device::write(u8 data)
 				logerror("Z80DMA Unknown WR6 command %02x\n", data);
 			}
 		}
-		else if (data == 0x8e) //newtype on Sharp X1, unknown purpose
-			logerror("Z80DMA Unknown base register %02x\n", data);
 		else
-			fatalerror("Z80DMA '%s' Unknown base register %02x\n", tag(), data);
+		{
+			// 0x8e: newtype on Sharp X1, unknown purpose
+			logerror("Z80DMA '%s' Unknown base register %02x\n", tag(), data);
```
cuavas, line 856: `logerror` is a `device_t` member function and always includes the tag.  You don’t need to duplicate it in the message.

holub: Will change in one of the following commits

angelosa: What are the other SWs that triggers this btw?

cuavas: I’ve already got it staged locally.  I realise it was just an oversight converting from throwing `emu_fatalerror` to `logerror`.

holub: > What are the other SWs that triggers this btw?

The Next's DMA extends this one. Recently we found an example where multiple software write 0xCA and this command is not documented anywhere (not only this one, but few other are crashing same way):
https://github.com/mikedailly/mod_player/blob/0c46930c3c0a2aa4679718452935853fe89c1774/mod_player/utils.asm#L627
> db $ca			; unknown

DMA in Spectrum is cursed. Many soft is based on examples from Magazines which either had typo or used rare clone of z80dma chip.

PR #14633, opened 2025-12-03
casio/superjr.cpp: Fix ROM_LOAD length to match expected jd320 ROM size

qufb:


galibert:
Why is the region that larger than the rom?

ghost:
Probably a mistake or not understanding ROMs. Filename is TC531001 which is a common Toshiba mask ROM compatible with a 27C010 EPROM which is 0x20000 in size.

angelosa:
Driver mounts over a `0x40000` sized area, with a `// banked?` note. Should probably be 0x80000 for all and padded with `ROMREGION_ERASEFF` afaict.

cuavas:
If it mirrors smaller ROMs, it may be easier to set the region size to match the size of the ROM in each case, and set it up correctly in `machine_start`.

qufb:
The program space memory map was based off jd361, which has the larger region size. I still need to further analyze these ROMs to check if there are data references beyond this mapping. But I agree that the region sizes should be made consistent for the other systems.

@cuavas can you check if the latest changes are aligned with your suggestion?

PR #14637, opened 2025-12-04
model2 driveio_comm_out

Endprodukt:
This does not implement force feedback.
It only exposes the raw 8-bit value that the game writes to the Model 2 Drive I/O register.

This byte is real machine state and part of the communication protocol with the external motor driver PCB on original hardware. No interpretation or scaling is performed.

Providing this raw value is useful for debugging, hardware research, and future emulation of the actual motor-driver board. Other MAME drivers expose similar raw protocol bytes even when the downstream hardware is not yet emulated.

A consideration for the merge would be highly appreciated.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -144,6 +144,7 @@ void model2_state::machine_start()
 	debug_init();
 
 	m_lamps.resolve();
+	m_driveio_comm_out.resolve();
 	
 	save_item(NAME(m_intreq));
 	save_item(NAME(m_intena));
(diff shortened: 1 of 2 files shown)


happppp:
I already disagreed to this, another person can give you feedback and if they think it's good, merge the PR. If they don't think this value should be sent to output, then what? 3rd try's the charm maybe? ;D

Endprodukt:
I didn't mean to be disrespectful. I see a real use for this output besides just ffb data and we will not see an emulated ffb drive board any time soon. 

If you absolutely see no benefit in it I will close the the pr and that's it.

jamesriley72281:
Sounds helpful to me. Not sure why it would get denied.

happppp:
Because the output in reality is from the partially emulated 'drivecpu' (the CPU handling the force feedback), not the internal comms latch between maincpu and drivecpu.

Hydreigon223:
Add the fact the host reads data from the 'drivecpu' for info like the drive board's steering pot among other things (Wave Runner would wire all four pots to both the host and drive board).

Making this (and Daytona 1's drive board) a device is the only option I can suggest.

Endprodukt:
Way above my skill level. Thanks for the suggestions and input.

PR #14640, opened 2025-12-07
ibm5170: Mark Crystal Caves as a baddump; fix grdchessdx media type

chungy:
The file dates for Crystal Caves show as 2005.  I believe this is unlikely to be original media, unless the guy mastering the disks in 1991 had his clock set forward by 14 years by accident.  (Not impossible, but I doubt it.)

Grandmaster Chess Deluxe has its entries set for a 5¼″ floppy drive. The images should be for a 3½″ drive instead.

### DIFF
--- a/hash/ibm5170.xml
+++ b/hash/ibm5170.xml
@@ -12123,9 +12123,9 @@ Differences between clone/parent:
 		<publisher>Apogee Software</publisher>
 		<info name="developer" value="Apogee Software" />
 		<info name="version" value="1.0a" />
+		<!-- baddump: file dates are in 2005, this is unlikely to be unmodified original media -->
 		<part name="flop1" interface="floppy_3_5">
 			<dataarea name="flop" size = "1474560">
-				<!-- File dates are in 2005; this is unlikely to be unmodified original media -->
 				<rom name="Crystal Caves [Apogee Software] [1991] [3.5HD] [Disk 1 of 1].img" size="1474560" crc="f4360e24" sha1="93a63c8bb0764ba83829b30263ce8870d1902c9f" status="baddump"/>
 			</dataarea>
 		</part>


hash/ibm5170.xml
```diff
@@ -12125,7 +12125,8 @@ Differences between clone/parent:
 		<info name="version" value="1.0a" />
 		<part name="flop1" interface="floppy_3_5">
 			<dataarea name="flop" size = "1474560">
-				<rom name="Crystal Caves [Apogee Software] [1991] [3.5HD] [Disk 1 of 1].img" size="1474560" crc="f4360e24" sha1="93a63c8bb0764ba83829b30263ce8870d1902c9f"/>
+				<!-- File dates are in 2005; this is unlikely to be unmodified original media -->
+				<rom name="Crystal Caves [Apogee Software] [1991] [3.5HD] [Disk 1 of 1].img" size="1474560" crc="f4360e24" sha1="93a63c8bb0764ba83829b30263ce8870d1902c9f" status="baddump"/>
```
angelosa: The more-or-less standardized way is to prefix the comment with `<!-- baddump: why_message -->`, no idea if there are plans to support such comments as actual info nodes in the future but imo better be consistent than sorry.

chungy: That pattern doesn't exist in this software list, but I can use it here :)

DopefishJustin:
Crystal Caves is an Apogee mail order title, most of which were kept in print for many years after release. For example I have a 3.5" HD copy of Duke Nukem (1991) with 1999 file dates, and a Commander Keen CD-ROM mastered in 2001, both ordered directly from Apogee. So 2005 isn't out of the question for an official reprint IMO. Their online store was still offering the game for sale on floppy disk as late as 2009: http://web.archive.org/web/20090725092312/http://3drealms.stores.yahoo.net/games.html

chungy:
Would a reprint really modify the file dates? Would they not keep a master image of the disks and recreate it the same every time?

2005 is pretty late for floppy disks, too.

FoxhackDN:
If they're imaging the disk on Windows, it could absolutely be modified by the OS _before_ they take it out. Windows is like that.

DopefishJustin:
Have a look at ORDERFRM.TXT on the disk:
```
       ==================================================================
       APOGEE/3D REALMS ORDER FORM - FAX SHEET (Fax number: 972-278-4670)
                      PAGE 1 OF 2 - OCT 21, 2005 Revision
       ==================================================================
```

chungy:
Fascinating. I still wonder why the install files have a 2005 date, especially since DOS/Windows systems default to preserving the date. I think you're right about it not being a bad dump after all.

Oh, it thickens, the PKSFX binary has a date of 1999-03-01. I'm surprised Apogee was like this for releasing new copies, I thought it'd be easier to just rewrite the same masters.

cuavas:
It’s quite possible they didn’t have master disks as such, and the process for doing a run of floppies used files as the source.

PR #14641, opened 2025-12-07
Added `tag` attribute to Info XML `<device_ref>` elements

npwoods:


cuavas:
What’s the size of the full XML before/after this change?  Also, what’s the maximum memory usage of `xmllint --valid --noout whatever.xml` before/after this change to get an idea of the effect it has on memory usage for building a DOM tree?

npwoods:
Before:  305,855,410 bytes (20,367,190 gzipped); memory usage 4673296 KB
After:  319,004,735 bytes (21,176,991 gzipped); memory usage 4856420 KB

Running `xmllint` showed that I needed to add a corresponding entry to the DTD; fixed

npwoods:
[#14641](https://github.com/mamedev/mame/pull/14641) was submitted a week ago and aside some questions about XML bloat (which were answered), there has not been any conversation.  I therefore conclude:
1. People are fine with adding `tag` to `<device_ref>`
2. Trying to remove the `emu_options` reference in `machine_config` has not been deemed to be a worthwhile endeavor.

As a consequence I am closing https://github.com/mamedev/mame/pull/14639 and merging [#14641](https://github.com/mamedev/mame/pull/14641)

PR #14643, opened 2025-12-07
m5_cart, m5_cass: 7 additions, 4 improved dumps and enhancements

paul75s:
Additions and enhancements to Sord M5 cassette and cart software lists:
- 7 additions to m5_cass (all working)
- 4 bad dumps replaced with working dumps in m5_cart (Up Up Balloon was not working, other 3 revised dumps did work with 36K on older versions of MAME but not on current versions and should not require additional RAM).
- All items in m5_cass updated with requirement for BASIC-I or BASIC-G cart, as appropriate.
- Some minor additional tweaks to metadata.

All files are from dumps provided by bsittler.

### DIFF
--- a/hash/m5_cass.xml
+++ b/hash/m5_cass.xml
@@ -10,13 +10,26 @@ license:CC0-1.0
 -->
 
 <softwarelist name="m5_cass" description="Sord M5 cassettes">
+	
+	<software name="3dsquash">
+		<description>3D Squash</description>
+		<year>19??</year>
+		<publisher>Sord</publisher>
+		<info name="usage" value="Mount together with &quot;BASIC-I&quot; and load with TAPE"/>
+		<sharedfeat name="requirement" value="m5_cart:basici"/>	
+		<part name="cass1" interface="m5_cass">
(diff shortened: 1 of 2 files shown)


hash/m5_cass.xml
```diff
@@ -314,19 +390,35 @@ license:CC0-1.0
...
 			</dataarea>
 		</part>
 	</software>
 
+	<software name="ufogame">
+		<description>UFO Game</description>
+		<year>19??</year>
+		<publisher>Takara</publisher>
+		<info name="usage" value="Mount together with &quot;BASIC-I&quot; and load with TAPE"/>
+		<info name="alt_title" value="ゲーム"/>
```
cuavas, line 407: I highly doubt the Japanese title of this game is simply “Game”.

paul75s: This was information provided by the source of the original dumps, which has been reliable and thorough for other dumps. As I don't have any other information to verify/amend this, I have removed the alternative title and opened PR 14655 for this change as the original change has been merged already.

PR #13102, opened 2024-12-20
hash/ibm51?0.xml: MS-DOS 6.22 variants

chungy:
A beta build of MS-DOS floats around as an alternative set.  Added the German supplemental disk to msdos622de (most of the disk is still in English, but notably DOS Shell is translated into German).

Most of the sets on WinWorld appear to be genuine, albeit the Russian translation has obvious signs of likely unintentional tampering, so that's the singular entry being marked as bad dumps here.

New working software list items (ibm5150.xml)
---------------------------------------------
MS-DOS (Version 6.22, 360K 5.25", Upgrade) [winworldpc.com]
MS-DOS 6.22 Supplemental Kit [winworldpc.com]

New working software list items (ibm5170.xml)
---------------------------------------------
MS-DOS (Version 6.22, 1994-05-09 beta) [winworldpc.com]
MS-DOS (Version 6.22, Dutch) [winworldpc.com]
MS-DOS (Version 6.22, French) [winworldpc.com]
MS-DOS (Version 6.22, Italian) [winworldpc.com]
MS-DOS (Version 6.22, Russian) [winworldpc.com]
MS-DOS (Version 6.22, Spanish) [winworldpc.com]
MS-DOS (Version 6.22, Swedish) [winworldpc.com]

PR #10017, opened 2022-07-01
New working software list additions

heroku-miraheze:
New working software list additions
-----------------------------------
Windows NT 4.0 Workstation
Windows NT 4.0 Server
Windows XP Professional (with Service Pack 3)

PR #14644, opened 2025-12-07
pc: split ibm5150 and ibm5170 software lists into categories

chungy:
This is a big one, but I have made sure that no software additions nor removals take place here. I even made sure a couple errors discovered in the process were corrected separately (PR #14640).

There is some precedent for categorization of software lists, the Amiga lists already have a few like this. My motivation here was to simplify the process of using the MAME UI to use specific kinds of software. Rather than Doom, Turbo Pascal, and Windows all being in one massive list, it could be easier to navigate to "DOS games", "Development", and "Operating Systems" for each of them respectively, with shorter lists focused on these types of software.

There's also potentially endless kinds of categories to make, but I tried to do my best with what exists in MAME's software lists now, and hopefully what new entries could slot into. I could possibly already foresee, say, the `ibm5170_cdrom_apps` list being split into per-OS categories like I've done with games and floppy applications, but it is so short right now I thought I would not bother. I feel that the MAME UI is already pushing to my personal tastes with what I have. :-)

Couple major decisions I've made:
* Games are listed as "DOS" for as long as they include a DOS version. Windows can run DOS games, but not vice-versa.
* I'm fully aware that what is and isn't an operating system will never be fully settled. For the purpose of the software lists, I figured that if it can host unique applications, it counts. Maybe you don't think Windows 3.0 is an OS, but it's definitely not a game either. I value usability over nitpicking these details.

**The following open pull requests would be affected and would require fixing if this PR is merged:**
* #10017 
* #13102
* #14534 

It'll also require significant intervention from users to split the lists. One possibility is to use the old lists as a source for rebuilding new ones; you can also copy the old directories to all the new names, and then do a full scan to rid the unnecessary files. (This might be costly on disk space if you don't have a reflink-supporting file system like btrfs or zfs)

MAME UI before the change:
Screenshot_20251207_120920
Screenshot_20251207_120925

MAME UI after the change:
Screenshot_20251207_121043
Screenshot_20251207_121047


ajrhacker:
I believe an approach like this was rejected many years in the past, back when softlists for some major systems were still missing or much less filled out than they are now. I don't think too highly of it. It's a lot of churn that will "also require significant intervention from users to split the lists," and I think that "to simplify the process of using the MAME UI to use specific kinds of software" is a goal that can and should be met by less intrusive means. The admissions that the categories are "potentially endless" and that some of the distinctions "will never be fully settled" are a bit telling.

angelosa:
I don't think this is the right approach either. The only split that practically needs to be done at current time from ibm5170 lists is  `generic_windows_cdrom` and perhaps `generic_windows_flop_35`. There are several entries that can be perfectly installed to PC-98, FM Towns and DOS/V at very least, beyond the Jp code page requirement. 
On its own `pc98_cd` already contains 94 "hybrids", that needs to be checked one by one and is already a less intrusive starting point.

angelosa:
Also: software categorization should be done as filter inside the XMLs really, then it's a matter of internal UI/external frontend managing them.

ArcadeShadow:
I agree with the opinions given.
I don't think it's the best way to address the subject. Splitting it into several files does not make maintenance easier. The division should be by system/storage media.
The UIs should handle the software filters, based on information contained in the hash file.

chungy:
> The only split that practically needs to be done at current time from ibm5170 lists is `generic_windows_cdrom` and perhaps `generic_windows_flop_35`.

Perhaps a `generic_cdrom` to handle multiplatform discs. `chexquest` and `mfma_addsub` are two such discs that are duplicated in both lists.

chungy:
The consensus went against this idea pretty firmly. Thanks for your input everyone!

PR #14645, opened 2025-12-07
sega/model2.cpp: when writing to IRQ mask register, wait two cycles before checking for sound IRQ; fixes MT09313

gm-matthew:
Virtua Cop 2 runs the following code when playing a music track or some sound effects:

```
0005A524: 90803000 005D05FC ld      0x5d05fc,g0			; load sound buffer queue size
0005A52C: 5CB01E01          mov     1,g6
0005A530: 92B03000 00E80004 st      g6,0xe80004			; mask all IRQs except vblank
0005A538: 92B03000 00E80004 st      g6,0xe80004
0005A540: 3C7C202C          cmpibl  15,g0,0x5a56c
0005A544: 90A03000 005D0600 ld      0x5d0600,g4
0005A54C: 59805010          addo    g0,1,g0				; increment sound buffer queue size
0005A550: 5CA81614          mov     g4,g5
0005A554: 8CA52001          lda     0x1(g4),g4
0005A558: 58A5088F          and     15,g4,g4
0005A55C: 92203915 005D0610 st      r4,0x5d0610[g5*4]
0005A564: 92A03000 005D0600 st      g4,0x5d0600
0005A56C: 8CB00401          lda     0x401,g6
0005A570: 92B03000 00E80004 st      g6,0xe80004			; unmask sound IRQ
0005A578: 92803000 005D05FC st      g0,0x5d05fc			; save sound buffer queue size
0005A580: 92B03000 00E80004 st      g6,0xe80004
```
During the sound interrupt handler, the game checks the sound buffer queue size. If it is greater than zero, it reads the next byte from the sound buffer queue, transmits it to the SCSP and decrements the queue size by one; if it is zero (queue is empty), it bails.

[MT09313](https://mametesters.org/view.php?id=9313) is caused by MAME asserting a sound interrupt the instant it is unmasked, before the game has a chance to save the incremented queue size. The sound interrupt handler reads the queue size, mistakenly thinks that the queue is empty and does not transmit the data to the SCSP, resulting in sound effects and/or music tracks not playing. This occurs since commit 99f8b5e05cc0f983a5d52f3e2bae46358e1ef789 when I rewrote the interrupt logic.

It seems there must be a slight delay between unmasking an interrupt and the same interrupt being fired; the exact delay is unknown, but it only needs to be enough for the following `st g0,0x5d05fc` instruction to be executed before the sound interrupt is asserted.

I have implemented it so that when writing to the IRQ mask register, instead of immediately checking to see if a sound interrupt should be asserted, we use a timer to wait for two cycles before doing so (the `st` instruction requires two cycles to execute).

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -138,6 +138,16 @@ TIMER_DEVICE_CALLBACK_MEMBER(model2_state::model2_timer_cb)
 	m_timerrun[TNum] = 0;
 }
 
+TIMER_CALLBACK_MEMBER(model2_state::check_sound_irq)
+{
+	const u32 line = 1 << 10;
+	if ((m_uart->status_r() & 0x03) && (m_intena & line))
+	{
+		m_intreq |= line;
+		irq_update();
+	}
+}
+
 void model2_state::machine_start()
 {
 	// initialize custom debugger pool, @see machine/model2.cpp
@@ -165,6 +175,9 @@ void model2_state::machine_start()
 
 	save_item(NAME(m_geo_write_start_address));
 	save_item(NAME(m_geo_read_start_address));
+
+	m_irq_delay_timer = timer_alloc(FUNC(model2_state::check_sound_irq), this);
+	m_irq_delay_timer->adjust(attotime::never);
 }
 
 void model2_tgp_state::machine_start()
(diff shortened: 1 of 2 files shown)


gm-matthew:
> With this change, because you update `m_intena` immediately and only defer checking for already-asserted UART interrupts until the timer fires, you get a situation where if the sound interrupt _changes to asserted_ in the window between `irq_enable_w` and `check_sound_irq`, it won’t honour the delay and will fire immediately. Is that intentional?

Are you referring to the possibility that `sound_ready_w()` could be called immediately after updating `m_intena`, thus firing a sound IRQ before the timer is scheduled? I must admit that I overlooked this possibility. The chances of it affecting `vcop2` are very small; it would have to unmask sound IRQs exactly 8000 cycles (25 MHz CPU clock / 3125 Hz MIDI clock) after transmitting a byte to the SCSP.

An alternative solution would be to implement the delay within the i960 core when firing an interrupt; the real i960 is pipelined and would not be capable of effectively triggering an external interrupt without at least a few cycles of latency. I shall look into the feasibility of implementing such a solution, and if I am successful we can revert this commit.

cuavas:
You can also feed the new value for the interrupt mask register to the callback via the timer parameter, or keep a separate member variable for the pending mask value and update the active mask value when the timer fires.

PR #14646, opened 2025-12-08
Make buildable on RISC-V Linux

rockyzhang:
Modified makefile and scripts/genie.lua to make it possible to build for RISC-V Linux.

### DIFF
--- a/makefile
+++ b/makefile
@@ -397,6 +397,10 @@ ifeq ($(findstring aarch64,$(UNAME)),aarch64)
 ARCHITECTURE :=
 endif
 
+ifeq ($(findstring riscv64,$(UNAME)),riscv64)
+ARCHITECTURE :=
+endif
+
 # Autodetect BIGENDIAN
 # MacOSX
 ifndef BIGENDIAN
--- a/scripts/genie.lua
+++ b/scripts/genie.lua
@@ -1087,6 +1087,12 @@ if (_OPTIONS["PLATFORM"]=="arm64") then
 	}
 end
 
+if (_OPTIONS["PLATFORM"]=="riscv64") then
+	buildoptions {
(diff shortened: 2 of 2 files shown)


rb6502:
Isn't this a duplicate of https://github.com/mamedev/mame/pull/14623 ?

rockyzhang:
> Isn't this a duplicate of #14623 ?

No, the was also modifying the generate script to make sure it is generating the correct compiler flag for RISC-V. Otherwise, there would be compiler failing on type cast.

belegdol:
As I only have access to emulation, I cannot really test the full compile. #14623 was the initial attempt to get the build going further. If `-Wno-cast-align` is required on riscv64, this PR is more correct.

rockyzhang:
> As I only have access to emulation, I cannot really test the full compile. #14623 was the initial attempt to get the build going further. If `-Wno-cast-align` is required on riscv64, this PR is more correct.

Yes, this is needed for the error/warnings on the pointer cast. Arm did this as well:

[arm/arm64 -Wno-cast-align](https://github.com/mamedev/mame/blob/b89fd58d0be6b8680d2c4cec965d72f1f5336b8b/scripts/genie.lua#L1078C1-L1088C4)

cuavas:
I’d rather see if those warnings can actually be addressed if possible.  We had MAME building with those warnings enabled on SPARC when that was a thing, but it seems to have rotted since then.  Hopefully we can enable the warnings for AArch64 as well.  You never know, it might even fix some of the ARM-specific issues.

rb6502:
This was addressed in https://github.com/mamedev/mame/commit/48eaefac071d52f2ef96efc7a2f90070e3e7d028 .  Enabling the warnings can be done separately.

PR #14651, opened 2025-12-08
machine/dp8390.cpp: push CR bit 5 high on DMA complete events

angelosa:


angelosa:
As per MX98902A datasheet I'm seeing:

image

0241 (Edit)

(writes first then reads back)
Screenshot From 2025-12-08 13-22-41

ajrhacker:
I have no comment on this. I haven't analyzed any network cards.

cuavas:
@pmackinlay is this sane?

rb6502:
This works fine with the Mac DP8390x cards.

PR #14652, opened 2025-12-08
Added the Dutch version of Kidizoom Twist

vtarchive:


mamehaze:
This is just the same ROM as the other set, which seems VERY unlikely as these have user data stored on it.

Are you the same troll who was adding fake 'Dutch' vtech sets before and trying to corrupt the database?

All your PRs are VERY suspicious.

I'm rejecting this anyway.

galibert:
Same sha/crc, different size?  No way.

mamehaze:
no, you haven't.  you've added a duplicate set of the German release with incorrect information.

PR #14653, opened 2025-12-08
Dumped the Dutch version of Kidizoom Twist's user data

vtarchive:
NEW NOT_WORKING machine addition:
Kidizoom Twist (Netherlands, blue camera) (VTArchive)

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -47341,6 +47341,7 @@ jak_ddhp
 jak_prhp
 kidizmb
 kidizmp
+kidizmtb
 neopad
 pdcm2
 pocketmp


vtarchive:
it's real not fake

mamehaze:
what did you use to dump it?

vtarchive:
HxD.

mamehaze:
So you dumped a NAND ROM with a Hex Editor?

vtarchive:
yes.

mamehaze:
Could you better explain the process.  HxD is not a dumping tool.

The closest HxD has in terms of functionality is the ability to open drives to view them.  It is not even designed for imaging drives. 

None of that explains how you used it to dump a NAND ROM.

vtarchive:
i used a tool to dump the whole nand. i used hxd to open a drive and save as a file named "kidizoomtwistdutchfw.bin"

mamehaze:
again this process does not make any sense.  why is HxD needed?

any standard dumping tool would come with software

what 'tool' are you using?

vtarchive:
im using HXD to rip NANDs from Kidizooms.

rb6502:
That doesn't make any sense.  How are you connecting the NAND to the PC to read it?

vtarchive:
i did manage to connect the NAND via PC and copied the RAW contents

mamehaze:
did you connect it to an ISA slot?

mamehaze:
Anyway, if you excuse my scepticism here, the last person we had trying to add a bunch of 'Dutch' Vtech sets, was doing nothing but adding fake entries with copy+paste data from other sets.  When questioned they just gave nonsensical answers.  They had to be banned from contributing to the project because they were making a mess.

Your first attempt to add this was exactly the same.  An objectively incorrect 'Dutch' Vtech set addition which was nothing but a fake entry copy and pasted from elsewhere.  Your other 'Dutch Vtech' PR was deleting valid entries in tables without even mentioning it in the description, which looked like vandalism.

Now you're providing answers to questions that make no sense at all to those with decades of experience in dumping this kind of thing.  It feels like you're being evasive, not providing details that would help your cause.

Trust in this PR is very low as a result.

vtarchive:
i finally submitting one that is real.

mamehaze:
So you admit the others were fake?

vtarchive:
could be. i have the kidizoom twist myself

happppp:
@vtarchive show us your tool please (the thing you used to read the NAND), a photo of it with your VTech toy.

cuavas:
> it's real not fake



> i did test and it has user data so i might be real.

You’ve gone from saying it is real to might be real.  This really doesn’t inspire confidence.

vtarchive:
it's real, checked my kidizoom twist

cuavas:
You haven’t actually explained how you connected the NAND Flash to a PC.  If you make one more comment without giving a straight answer, I’m closing this.

vtarchive:
> @vtarchive show us your tool please (the thing you used to read the NAND), a photo of it with your VTech toy.

HxD, it's free and can dump NANDs of VTech toys with a USB port. useful if you dont plan to open the system.
https://drive.google.com/file/d/1pDGh0g-Pq1faLGRR1mTb830s2_kJMBt_/view?usp=sharing

cuavas:
What USB endpoints does the VTech toy present itself as?  If it’s USB Mass Storage class, what steps are you taking to ensure Windows doesn’t automatically modify the content?  Has it been confirmed that this actually gives a raw view of the NAND Flash in the device?

vtarchive:
unmodified after dumping and still works, its confirmed that this actually gives a raw view.

cuavas:
> unmodified after dumping and still works, its confirmed that this actually gives a raw view.

How did you confirm it gives a raw view?  What steps did you take to ensure Windows will not modify the content?

vtarchive:
i've confirmed, the first step is opening the toy using HXD and save as a bin file but nothing is modified since that point

cuavas:
That doesn’t confirm that Windows didn’t write to it as soon as it was connected.

vtarchive:
i've confirmed, but i saved as a different location from the camera and nothing has changed.

mamehaze:
Even in this conversation we've gone from

"Used a tool"
to
"Plugged it into a PC via USB"

none of this really builds any trust, the communication keeps changing.

Is this even a technique likely to dump the firmware, or just a user storage area, which isn't a dump of the system at all.  It would seem a bit strange if the entire firmware was exposed over USB.

happppp:
It doesn't sound implausible to be able to open the system partition as raw data and then extract the firmware from that.
Firmware exposed over USB?: Of course it is, how else would a firmware update work otherwise.

mamehaze:
maybe, although other devices I have that connect to the PCB aren't exposing their entire internal firmware that way.

why the OP didn't just explain this in the first place, rather than making up some story about a tool I don't know though.

I'm still treating with scepticism though due to the admission that previous stuff they've sent has been fake.

vtarchive:
i didn't expose the NAND,

ghost:
@cuavas 
I'm not commenting on the legitimacy of the dump or method used but I want to clear up something about ROMs in general which clearly some here do not understand. I don't believe for a second that Windows could modify any ROM without the devices consent, even a NAND flash ROM. It's not a DOS-readable HDD or DOS-readable anything. It's a ROM which has a non-standard format and a write-enable pin that must be toggled to the correct state along with several other pins (Reset, CE, OE, Ready/Busy etc etc). Those pins are all controlled by the hardware. Multiple pins must be set to a specific state in order to put the device into write mode and if those pins are not set correctly or it is hard-wired to read-mode after programming then it will never be writable. The programming info is in any datasheet for that device. You can't just plug a device containing flash ROM into USB and it gets exposed to writing. That simply does not happen with any ROM. The device may support updating via USB but it is controlled by specific updating software that tells the hardware to change modes and sets specific pins to specific states. This is exactly the same as updating any PC BIOS or updating device firmware. Windows can't write to the BIOS unless special device-specific software is run to update the BIOS / firmware from within Windows.
Anyway, if Windows was able to write to the device and corrupt it... A). The device would no longer be working and the owner has already said it still works after dumping, and B). Someone could easily go through the dump and locate where Windows (supposedly) dirtied the data since it is known what Windows does when a device is plugged in. Windows does not just write random bytes to non-standard media. 
Take another example of an arcade HDD that is written in raw mode where the arcade board reads it directly, such as Gauntlet Dark Legacy (there are hundreds of other similar example HDD dumps in MAME). The HDD is not in any Windows-readable format and as such Windows can't access it nor write to it. 

The solution to all of this nonsense is to take the ROM file (and only the ROM file), ignore the PR and someone add it and test it or put it in the archives until such time that someone can add it/test it. In other words treat it like the other 50000 devices that were added to MAME in the past.

PR #14659, opened 2025-12-09
properly identify falconun

mamehaze:
I got a mail from the guy who dumped this, chasing up on if there had been any progress (and also again asking why the Mustache Boy thing he also dumped hadn't been added yet)

IMHO the Mustache Boy set is perfectly legit, but I'm not going down that path of trying to convince people saying otherwise.

I've renamed this thing to what it turned out to be, because there's no point in having it mislabeled in MAME.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -43751,9 +43751,6 @@ waveterm
 @source:skeleton/extrema_m27sv.cpp
 multgmxl
 
-@source:skeleton/falconun.cpp
-falconun
-
 @source:skeleton/fanucs15.cpp
 fanucs15
 
@@ -44098,6 +44095,9 @@ pm68k
 
 @source:skeleton/powerpack.cpp
 powerpack
+
+@source:skeleton/predatrs.cpp
+predatrs
 
 @source:skeleton/proteus3.cpp
 proteus3
--- a/src/mame/skeleton/predatrs.cpp
+++ b/src/mame/skeleton/predatrs.cpp
@@ -13,7 +13,7 @@
 
 
 /*
-Unknown Falcon\Williams host & satellite terminal gambling game.
+Williams / QVS - The Predators
 
 From a unfunctional set of 3 host PCBs and 5 matching Falcon terminal PCBs
 and auxiliary PCBs without any ROM.
@@ -25,16 +25,16 @@ Recovered ROMSs are from a mixture of PCBs and could be incomplete.
 
 namespace {
 
-class falconun_state : public driver_device
+class predatrs_state : public driver_device
 {
 public:
-	falconun_state(const machine_config &mconfig, device_type type, const char *tag) :
+	predatrs_state(const machine_config &mconfig, device_type type, const char *tag) :
 		driver_device(mconfig, type, tag),
 		m_maincpu(*this, "maincpu"),
 		m_screen(*this, "screen")
 	{ }
 
-	void falconun(machine_config &config);
+	void predatrs(machine_config &config);
 
 private:
 	required_device<m6802_cpu_device> m_maincpu;
@@ -46,25 +46,25 @@ private:
 	uint32_t screen_update(screen_device &screen, bitmap_rgb32 &bitmap, const rectangle &cliprect) { return 0; }
 };
 
-static INPUT_PORTS_START( falconun )
+static INPUT_PORTS_START( predatrs )
 INPUT_PORTS_END
 
-void falconun_state::memmap(address_map &map)
+void predatrs_state::memmap(address_map &map)
 {
 	map(0x0000, 0x07ff).ram();
 	map(0x2000, 0x2003).rw("pia", FUNC(pia6821_device::read), FUNC(pia6821_device::write));
 	map(0xc000, 0xffff).rom().region("falcon_terminal", 0xc000);
 }
 
-void falconun_state::slavemap(address_map &map)
+void predatrs_state::slavemap(address_map &map)
 {
 	map(0x000, 0xfff).rom().region("slave", 0);
 }
 
-void falconun_state::falconun(machine_config &config)
+void predatrs_state::predatrs(machine_config &config)
 {
 	M6802(config, m_maincpu, 40'000'000 / 4); // TODO 10 MHz for a M6802??
-	m_maincpu->set_addrmap(AS_PROGRAM, &falconun_state::memmap);
+	m_maincpu->set_addrmap(AS_PROGRAM, &predatrs_state::memmap);
 	m_maincpu->set_ram_enable(false);
 
 	INPUT_MERGER_ANY_HIGH(config, "mainirq").output_handler().set_inputline(m_maincpu, M6802_IRQ_LINE);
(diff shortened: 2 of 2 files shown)


happppp:
Did you forget that @smf- asked you to stop chasing this?

mamehaze:
I've done no work on it, but I don't really see the benefit of MAME having misinformation in it when something has been correctly identified.  The correct information isn't harming anybody.

Since I've still got the original dumper chasing me up on it, I just made this PR to acknowledge what we have learned.

I don't know why politics are being dragged into this, it's just rude towards the person who found this.  Why is MAME showing favouritism for dumps of a 40 year old game? MAME's business is to accurately describe what we know, not engage in petty politics.

The dumper provided 2 legitimate looking dumps, this and a Mustache Boy set.  The Mustache Boy set got inexplicably rejected as a region hack (no real reason to think it is, the region byte was not known about before the dump) and this which having identified it now seems to be being treated as unwanted too.

We've got actual idiots to deal with when it comes to fake sets, like the guy trying to add the VTech stuff.  Maybe reserve the hate for that kind of person, not people who have been going through random PCBs they've picked up to dump them.

mamehaze:
Anyway, I'll probably make a PR for the Mustache Boy set, because the situation is ridiculous. Mamedev was asked if they needed more info / photos etc. (for both these boards) didn't reply, then only asked for photos once these boards had been sold a year later when the dumper followed up to ask why nothing had been added.

This set got added at this point (by Osso I think) but Mustache Boy was still ignored. I was asked about Mustache Boy at the time because the dumper got fed up with replies from Mamedev, but I said common sense would likely prevail if they pushed the issue, as it was clearly Mamedev's fault for not requesting photos when they were originally offered.

happppp:
IMO Mustache Boy has more chance to get added if you can confirm that it's the exact same ROM contents as the version you were hoarding.

mamehaze:
I've not been hoarding any mustache boy sets?  I'm not aware of any prior dumps of that version at all apart from the one dumped by this guy.

happppp:
In that case I'd say there's a good chance it's a 1-bit ROM hack by someone.
MAMEdev never received any emails from this guy 1 year before.

mamehaze:
I'm assured the mails were sent and was forwarded the original mail (sent to code@mamedev) when the guy was complaining about it being rejected.

If your email server / forwarder dropped them that's still Mamedev's fault.

Seibu regions are typically 1 byte changes.  The region byte was not known about prior to it being dumped (weirdly documentation of it was added to the code days after the dump was rejected)  I don't see why any of this is suspicious.  Far more suspicious looking Seibu sets have been added in the past.

MooglyGuy:
Either way, the point that hap is dancing around here is that @philipjbennett wrote a completely-working driver for The Predators back in the day, it simply didn't go in as the person who provided the prototype ROMs was insistent that it not go in until MAME supported positional audio.

MAME now supports positional audio. Either @smf- is reinventing the wheel or he's blowing the dust off of Phil's driver, but either way, simply bulk-renaming a driver and changing a couple strings is something best characterised as "rearranging deck chairs on the Titanic". There really isn't any point to it. It's not a slight against *you*, it's just that it's already being handled, and we already know how sensitive smf is when it comes to dealing with merge conflicts if he already has something in-flight.

The Mustache Boy thing is a different story, submit a PR and it should be pretty easy for me or anyone to dig into the code and see what the single changed byte affects.

mamehaze:
Actually I wrote the driver some 15-20 years ago, and Phil fixed up some things in it.  That's why I recognised it when somebody else started to show an interest in the dump, treating the collision map like a bitmap display is the easiest mistake to make when trying to figure it out.  I'd got an entire blog post ready to post about it when my site was still on mameworld because it's *really* funky hardware.

But yes, the person who dumped it 15-20 years ago originally said it could be in MAME a year or so later, but kept moving the goalposts.  The last set of goalposts were positional audio, so technically this is fine by them even if their opinion really doesn't matter since they stalled so long another set of boards turned up.

Hammy1986:
The Mustache Boy set is legit, check out raiden 2 sets to see the same thing that seibu did.
Sure, one could hack all the sets but when they are confirmed to be on board too, the version should be added.

mamehaze:
Yeah.  At worst, it could be marked as hack? if you're really suspicious, but I see no reason to do so.

I was sent over the archive of the dumps this guy had done.  About 8 sets plus some loose ROMs.  6 of them matched sets in MAME, nothing to add, the loose ROMs looked like junk, the others were mustache boy and this, both of which look like legitimate sets.

MooglyGuy:
> Yeah. At worst, it could be marked as hack? if you're really suspicious, but I see no reason to do so.
> 
> I was sent over the archive of the dumps this guy had done. About 8 sets plus some loose ROMs. 6 of them matched sets in MAME, nothing to add, the loose ROMs looked like junk, the others were mustache boy and this, both of which look like legitimate sets.

That's pretty much my take on it, but none of this is germane to the discussion on this particular PR.

The Predators is being handled by Phil Bennett (previously) and smf (now), so if it turns out that the ROM in `falconun.cpp` is a hash-match for one of the ROMs in the sets they're working with, then the driver will likely be deleted once they submit a working driver, now that the positional-speaker blocker is out of the way.

Alternatively, if it turns out that it's not a hash-match, then it would probably make sense to keep the driver as-is.

At this point I think it would make sense to just close this PR and see what the future results end up looking like, since smf is actively working on a driver for The Predators, so all merging this would do is potentially introduce a merge conflict that would be awkward to work around.

Does that path forward sound reasonable enough? I won't close it on my own volition as I don't want to step on your toes, @mamehaze.

mamehaze:
Well I figure the driver that's been worked on will be off the old driver I did (with Phil) not related to this one at all.  I doubt it's going to cause merge conflicts.  I don't especially see the harm in properly identifying the set here though, MAME is meant to be factual at the end of the day, and this is no longer unknown.

Once the new (old) driver goes in this just gets deleted.  In the meantime, it should probably at least have the correct description.

MooglyGuy:
So are you confirming that the single ROM in `falconun.cpp` matches at least one of the ROMs from the dump of The Predators that was provided to you?

The Mustache Boy discussion is not germane to this PR and should be occurring through other means. This is not the place for it.

mamehaze:
I've managed to dig up the old driver, don't have the old romset handy.  Looks like one file differs?
[pred.zip](https://github.com/user-attachments/files/24146141/pred.zip)

MooglyGuy:
For context, it's precisely what I theorized, and what @mamehaze was trying to get across: That the ROMset for Wliiams's prototype "The Predators" did in fact include this ROM.

As the overall driver and ROMset are presently being handled by @smf- based on a copy of the old driver that @philipjbennett seems to have ceded ownership of, it is a ROMset that will be dealt with in its own time. There is even a hash-match against what was in this skeleton driver vs. what's being handled as a prelude to a fully-working driver. Thus there isn't much point in keeping this current "[manufacturer]unk.cpp" driver around.

I hope this is a sentiment that Haze shares, but that is the complete story as to what's been going on behind the proverbial curtain. It's being handled, the hash will be be subsumed into another driver fairly shortly, and all the creation of a same-named driver would introduce is a potential merge conflict, which I hope we can all agree is not something anyone wants.

So, despite an initial disagreement, everything seems to be proceeding forward as it should. Let's try to leave it on this positive note.

mamehaze:
Phil cleaned up the driver.  Remember this is *really* old; work on the driver started over 15 years ago and the original one I made predates the proper device system and had a lot more code duplication, but most of the code in what I've attached is still my code just refactored, so I'd consider myself the primary copyright owner on that version of the driver even if Phil's name is listed first.  I was asked to write the driver, and enlisted the help of others to take it a bit further (then ended up being frozen out of future updates)

I still think the set that's in MAME should be properly identified though, even if it's just changing the manufacturer & name strings, as it makes no sense to leave it as 'unknown' when it isn't.

I'm not sure the exact vintage of this version of the driver, as I found it loose on an old IDE drive, but it's probably a fully playable game (despite some visual bugs) with a 2010 era of MAME or similar.

MooglyGuy:
It will be, when smf submits the updated version of Phil's code. Phil was kind enough to confirm that, yes, this is in fact a single ROM in the overall The Predators set, so your instincts were correct. It just doesn't make much sense to rename the set if smf is going to come rolling in with a brand new driver. I appreciate your keen eye for this sort of thing, I wonder sometimes if there are other situations in the codebase that have a similar sort of vibe. But as I theorized and now have confirmed, it's just proverbially re-arranging deck chairs on the Titanic, so it's a far better use of your time to focus on other things. I appreciate you being understanding about it, because it really isn't intended as a slight against you, it's just a matter of not introducing a potential merge conflict when an entirely-working driver for the very same set is (potentially) on the cusp of being pushed.

mamehaze:
I'm not sure what you mean by 'single ROM'. From what I can see it's a
functional/complete set which is just missing the pal dump and has one ROM
with a different hash (which could have been a bad dump in the old set)

MooglyGuy:
Sorry about that, Haze, I completely misread the PR. Didn't expand the diff far enough, that's my bad.

At any rate, it's been confirmed that the files are identical to the dump of The Predators that'll be going in soon enough. You made a good and accurate call to want to rename the set, and I'm sorry for being doubtful about it. Given that the fully-working driver is having the dust brushed off it for submission soon (ish?), I appreciate the understanding to close this out so that the former goes in more smoothly. You're a good egg, dude.

PR #14660, opened 2025-12-09
new WORKING machine - Jammin' (Atari prototype)

mamehaze:
new WORKING machine 
------
Jammin' (prototype) [SynaMax, Dutchman2000, David Haywood]

### DIFF
--- a/src/mame/nintendo/dkong.cpp
+++ b/src/mame/nintendo/dkong.cpp
@@ -1812,7 +1812,7 @@ void jammin_state::jammin(machine_config &config)
 	// sound hardware
 	SPEAKER(config, "speaker", 2).front();
 
-	YM2151(config, m_ym2151, I8035_CLOCK / 2); // matches video reference
+	YM2151(config, m_ym2151, MASTER_CLOCK / 20); // matches video reference
 	m_ym2151->add_route(0, "speaker", 0.60, 0);
 	m_ym2151->add_route(1, "speaker", 0.60, 1);
 }


ajrhacker:
This is Donkey Kong-based hardware and is probably worth folding into that driver.

mamehaze:
Ahh, interesting, not something I would have guessed.  I'll see about
moving it over there.

synamaxmusic:
> This is Donkey Kong-based hardware and is probably worth folding into that driver.

Excellent catch.  It makes perfect sense since the original VHS footage shows the game running in a Nintendo cabinet.

mamehaze:
It's now been moved into the DK driver

mamehaze:
jam1 jam2 jam3

jam4 jam5 jam6

I now consider this a working game.  I think the sprite colours are OK and the 4bpp sprite graphics + extra palette (that I can't see how to use at all) might have been for a different hardware type.  Left them in anyway in case somebody figures out they can be used, but there's not really good enough reference material to draw any conclusions.  What we have looks reasonable within the confines of DK hardware conversion.

Dips aren't filled in, but I'm not sure the game developed to the stage of them being implemented.

ajrhacker:
I don't think the 4Kx8x8 split is actually realistic. Actual PCBs have a limited number of EPROM sockets, and even Nintendo's sequels on the same hardware used 8Kx8 EPROMs.

mamehaze:
I could entertain the idea that it was 2 16kb ROMs on a sub-board given it was 1985.  I just wanted to move away from 'random odd-sized binary blob with large unused area in the middle' which definitely isn't what would have been on a PCB.

I'd be very surprised if there was no sub-board given the addition of the YM.

mamehaze:
I've gone with that instead.

synamaxmusic:
The ROM split also makes sense as I had to tell the linker to start the data segment at 8000h.  Before that, it was placing machine code at 6000h (which is allocated for RAM).

mamehaze:
btw apparently the music pitch doesn't quite match the OG recordings.  I used the System2 clock for the YM, as some of the tunes are the same as System 2 games, so I thought maybe Atari used the same Xtal, but assuming the footage we have is at the correct speed, maybe they didn't.  This isn't something I can easily tune myself, but might be something to look at for somebody with better ears.

synamaxmusic:
Can confirm that the pitch appears to be a semitone or so higher than the original video.  Both the emulated audio and VHS audio have nearly identical tempos so at least the song speed is very accurate.

mamehaze:
ok ramacat did some testing, 3Mhz matches the recordings, and we can reasonably get that

*edit* 3.075 (same as Z80) is even closer, so we're 99% sure it's using that, not the 6/2

mamehaze:
@ajrhacker done with this now.  Everything matches the footage we have.

PR #14664, opened 2025-12-11
fix risc-v build error casued by cast-align warnings

rockyzhang:
RISC-V, like arm/arm64 is an architecture requires more strict alignment.

Disable the warning like arm/arm64 to make it buildable.

### DIFF
--- a/scripts/genie.lua
+++ b/scripts/genie.lua
@@ -1087,6 +1087,12 @@ if (_OPTIONS["PLATFORM"]=="arm64") then
 	}
 end
 
+if (_OPTIONS["PLATFORM"]=="riscv64") then
+	buildoptions {
+		"-Wno-cast-align",
+	}
+end
+
 local subdir
 if (_OPTIONS["target"] == _OPTIONS["subtarget"]) then
 	subdir = _OPTIONS["osd"] .. "/" .. _OPTIONS["target"]


belegdol:
I believe @cuavas preferred to actually fix the code causing the warnings rather than disabling them. As such, it could help if you provided a build log with `-Wno-error=cast-align`. This way we would know which code parts generate the warnings.

cuavas:
Disabling the warnings doesn't fix the issues.  We need to start addressing this rather than just suppressing the warnings.

rockyzhang:
I'm collecting the log with -Wno-error=cast-align, but seems there are quite some fix needs to be done.

MooglyGuy:
> I'm collecting the log with -Wno-error=cast-align, but seems there are quite some fix needs to be done.

Feel free to post the results on a publicly-accessible spot, perhaps they're things that other people can help with.

rockyzhang:
[make.log.gz](https://github.com/user-attachments/files/24146753/make.log.gz)

here is the build log.

belegdol:
While it is using non-standard compiler flags, here is the build log of mame package - first successful build on risc-v: https://riscv-koji.fedoraproject.org/koji/buildinfo?buildID=83097

PR #14670, opened 2025-12-13
snk/hng64.cpp: added lamp and ffb wheel outputs

Endprodukt:
added outputs for lamps and wheel motor.

### DIFF
--- a/src/mame/snk/hng64.h
+++ b/src/mame/snk/hng64.h
@@ -350,7 +350,7 @@ private:
 	std::vector<polygon> m_polys;  // HNG64_MAX_POLYGONS
 
 	output_finder<> m_wheel_motor;
-	output_finder<64> m_lamps_out;
+	output_finder<8> m_lamps_out;
 
 	u16 main_latch[2]{};
 	u16 sound_latch[2]{};


src/mame/snk/hng64.cpp
```diff
@@ -2329,14 +2331,25 @@ u8 hng64_state::ioport3_r()
 
 DEFINE_DEVICE_TYPE(HNG64_LAMPS, hng64_lamps_device, "hng64_lamps", "HNG64 Lamps")
 
-hng64_lamps_device::hng64_lamps_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock)
+hng64_lamps_device::hng64_lamps_device(const machine_config& mconfig, const char* tag, device_t* owner, u32 clock)
```
MooglyGuy: Please undo this change. Checking files that are exclusively owned by Vas, the reference and pointer tokens should be right-aligned to the variable name, so it can be inferred that the way it currently is is the preferred formatting.

src/mame/snk/hng64.cpp
```diff
@@ -2332,11 +2334,22 @@ DEFINE_DEVICE_TYPE(HNG64_LAMPS, hng64_lamps_device, "hng64_lamps", "HNG64 Lamps"
...
 }
 
 void hng64_lamps_device::device_start()
 {
+	m_lamps.resolve();
+}
+
+void hng64_lamps_device::lamps_w(offs_t offset, u8 data)
+{
+	const u32 base = (offset & 7) * 8;
```
MooglyGuy: `<< 3` instead of `* 8` would be preferred here.

src/mame/snk/hng64.cpp
```diff
@@ -2332,11 +2334,22 @@ DEFINE_DEVICE_TYPE(HNG64_LAMPS, hng64_lamps_device, "hng64_lamps", "HNG64 Lamps"
...
+	for (int bit = 0; bit < 8; bit++)
+	m_lamps[base + bit] = BIT(data, bit);
```
MooglyGuy: Please indent the inner statement of the loop.

Endprodukt:
> Your changes add an additional feature; please address these points first.

ignore the model 2 stuff. I was working on the same branch and it got accidentally committed.

happppp:
It looks to me like the driver author intended to have the lamps hooked up at where ```hng64_lamps_device``` outputs to.
(It's not how I would do it by the way, that lamps device is odd..)

Endprodukt:
> It looks to me like the driver author intended to have the lamps hooked up at where `hng64_lamps_device` outputs to. (It's not how I would do it by the way, that lamps device is odd..)

yeah it's a bit weird indeed but to be fair I don't fully understand what you would like me to do.

happppp:
the hng64 lamps device is from @mamehaze - he can tell what his intention with that thing is.

Endprodukt:
> the hng64 lamps device is from @mamehaze - he can tell what his intention with that thing is.

that's fine but what is the problem with my PR :D

mamehaze:
I don't actually remember making it into a device, but it was a while back.

I do remember not being able to find all the outputs for Beast Busters 2 and wondering if there was a bug in the IO MCU core causing data to not be output properly (maybe shifted, or a bit dropped)

I make no claim on this either way, do as you want with it.  If you're also finding lamps are missing on some games, please do check the IO core, as it has no other users and was thrown together quite quickly.

happppp:
Your PR:
cpu write -> {lamps output} + lamp device -> lamp device callback -> N/A

Expected in current driver:
cpu write -> lamp device -> lamp device callback -> {lamps output}

Endprodukt:
thank you as always for taking your time. I changed this quite a bit and hope this works for you :)

happppp:
I don't think you need 64 lamps. Other than that, yeah this is a better place for the lamp outputs. You tested it?

Endprodukt:
all is tested indeed. I wasn't sure how many lamps I should use at first. 8 covers them all!

happppp:
FWIW CI's clang-arm64 build failure is not from this PR, a compile error in imgtool.

Endprodukt:
Thanks for merging!

PR #14671, opened 2025-12-13
ibm5150, ibm5170, ibm5170_cdrom: Norton Utilites 1983–2002

chungy:
This covers basically the entire original run of the Norton Utilities as a standalone product.  Version 2002 is also pretty much the last that has good functionality within MAME itself.

While version 2002 dropped Windows 95 support, it still includes many DOS programs that run all the way back to MS-DOS 5.0.  Notably the DiskEdit program supports FAT32, VFAT, and all that allows some forensics to be done to determine good/bad dumps.

### DIFF
--- a/hash/ibm5150.xml
+++ b/hash/ibm5150.xml
@@ -3941,7 +3941,7 @@ Fatal error: Incorrect layout on track 39 head 0, expected_size=100000, current_
 			</dataarea>
 		</part>
 		<part name="flop2" interface="floppy_3_5">
-			<dataarea name="flop" size="727280">
+			<dataarea name="flop" size="737280">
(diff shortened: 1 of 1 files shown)


ajrhacker:
Can you fix all the entries added to ibm5150.xml with multiple parts named "flop1"?

chungy:
> Can you fix all the entries added to ibm5150.xml with multiple parts named "flop1"?

Wish I saw this message instead of hacking up minimaws to give me where the CI errors were happening :)

But yes, thanks.

ajrhacker:
By the way, who should be credited for these?

chungy:
I thought that's all in the commit message; the typical pattern I've seen in `git log` is to credit where it was obtained.

PR #14675, opened 2025-12-14
Add 'Progear (Europe 010117)' (missing decryption key)

clawgrip:
New clones marked not working
-----------------------------
Progear (Europe 010117) [Eduardo Anay]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -15859,6 +15859,7 @@ pgear
 pgearr1
 progear
 progeara
+progeare
 progearj
 progearjbl
 progearjd


MooglyGuy:
Any chance of getting at MAMEdev to find the key, rather than submitting it in an encrypted state?

sergiopolog:
Great add! For some time, this set was in doubt to even exist.
I think it should be moved as parent set when the decryption key will be extracted like all other CPS2 sets which `Euro` region is marked as parent, but for now it's good as a clone as it's currently in 'not working' state.

PR #14677, opened 2025-12-16
New Machine - Head On 2 (Sega Slimline)/Head On Part 2 (Nintendo)

SNESNESCUBE64:
This new machine is the vertically oriented version of head on 2. As far as I can tell this was found in the Sega Slimline cabinets (which is where the dump from this came from) and a single example of a Nintendo Head On Part 2 machine. The Nintendo one was particularly interesting because it ran on stock Head On N hardware. The only noticeable difference I observed with it running on Head On N hardware is that it cannot have six lives. This is because despite having 8 dip switch switches, only one is actually hooked up on the PCB. This is acknowledged by the instruction card in the machine as the only options are 4 and 5 lives. Otherwise, it plays like a hybrid of Head On N and Head On 2. It is meant to only be played on cocktail tables, the screen flipping is hard coded. 

The only effected MAME driver is the vicdual driver. The machine config had to be added to the public header and the bare minimum had to be added to the vicdual source file (machine config, rom config, and input ports). I did both a full build and vicdual focused build, it does not seem like any additional compiler warnings were added. The game plays in MAME just like it does on a real machine with the exception being the dot sound (which does not seem to be full implemented). The best way I can describe the dot sound is it sounds like a quick pulse of the bonus sound. This is a per-existing condition though.

Regarding the ROMs, the actual program ROMs are unique. However, the timing PROMs are the same as Head On.

### DIFF
--- a/src/mame/sega/vicdual.cpp
+++ b/src/mame/sega/vicdual.cpp
@@ -1678,7 +1678,7 @@ static INPUT_PORTS_START( headon2sl )
 	PORT_DIPNAME( 0x03, 0x01, "Lives" )   PORT_DIPLOCATION("SW1:1,2")
 	PORT_DIPSETTING(    0x00, "4" )
 	PORT_DIPSETTING(    0x01, "5" )
-	//PORT_DIPSETTING(    0x02, "5" ) /* Duplicate Value - Each switch set adds a life */
+	PORT_DIPSETTING(    0x02, "5 (Duplicate)" ) /* Duplicate Value - Each switch set adds a life */
 	PORT_DIPSETTING(    0x03, "6" )
 INPUT_PORTS_END
 


SNESNESCUBE64:
Adding image
image

src/mame/mame.lst
```diff
@@ -42465,6 +42465,7 @@ frogs
 headon
 headon1
 headon2
+headon2sl
```
Osso13: please put this in alphabetical order

src/mame/sega/vicdual.cpp
```diff
@@ -1740,6 +1790,18 @@ void vicdual_state::headon2(machine_config &config)
 	headon_audio(config);
 }
 
+void vicdual_state::headon2sl(machine_config &config)
```
Osso13: isn't this identical to headonn(machine_config &config)?
if this is the case, you don't need to duplicate it, you can just use the latter.

src/mame/sega/vicdual.cpp
```diff
@@ -1632,6 +1634,54 @@ static INPUT_PORTS_START( headon2 )
...
+	PORT_BIT(0xc0, IP_ACTIVE_LOW, IPT_UNUSED)
+
+	PORT_COIN_DEFAULT
+
+	
+	PORT_START("FAKE_LIVES.0")
+	PORT_DIPNAME( 0x03, 0x01, "Lives" )   PORT_DIPLOCATION("SW1:1,2")
+	PORT_DIPSETTING(    0x00, "4" )
+	PORT_DIPSETTING(    0x01, "5" )
+	//PORT_DIPSETTING(    0x02, "5" ) /* Duplicate Value - Each switch set adds a life */
```
Osso13: current orientation is to leave all switches accessible to the user, even if they are duplicates, so please don't comment this out

src/mame/mame.lst
SNESNESCUBE64: Updated this to make alphabetical order

src/mame/sega/vicdual.cpp
SNESNESCUBE64: There are two reasons I did this:
1. Official documentation does not acknowledge that specific combination. I did test it to see what would happen and found that each dip switch set adds a life. 
image

2. Other drivers within the vicdual driver also do this. For example: sspaceat, headon2, and invho2 (The Sega Slimline "Special Dual" cabinet that has invinco and head on 2) all have the duplicate dip switch options commented out.

Should I match those machines for dip switches or should I leave the duplicate option knowing that it wasn't intended on being used and doesn't benefit the user?

src/mame/sega/vicdual.cpp
SNESNESCUBE64: Removed duplicate settings in both source and header files, good catch.

src/mame/sega/vicdual.cpp
angelosa: `PORT_DIPSETTING( 0x02, "5 (duplicate)" )`

SNESNESCUBE64: Uncommented the dip switch option, labeling it as a duplicate. If this is the way, moving forward the other games that comment out the dip switch options should also be investigated.

Osso13:
Thanks! One last thing before merging: who gets credited for this?

New working clones
------------------
Head On 2 (Sega Slimline) [???]

SNESNESCUBE64:
You can just put "SNESNESCUBE64". Thanks!

PR #14531, opened 2025-11-15
sequential/sixtrak.cpp: Emulated audio and promoted to working.

m1macrophage:
New functionality:
* Wheel RC circuits.
* Autotune circuit.
* Audio.

Systems promoted to working
----------------------------
Sequential Circuits Six-Trak (Model 610) Rev B/C

PR #14680, opened 2025-12-16
sound/cem3394.cpp: implement self-oscillation when no waveforms are enabled

m1macrophage:
Detecting the special case of a filter with high resonance and no input waveforms, and generating a sine wave to implement self-oscillation.

Fixes the regression reporter [here](https://github.com/mamedev/mame/pull/14531#issuecomment-3653253168). Also fixes patches 98 and 99 on the `sixtrak`, and enables ever more sounds in `sente6vb` .

Finally, corrected the value of `m_pulse_width` when PW is at 100%.  Though this does not make a difference, since `m_pulse_width` is not used when the PW is 100%.

### DIFF
--- a/src/devices/sound/cem3394.cpp
+++ b/src/devices/sound/cem3394.cpp
@@ -259,7 +259,6 @@ double cem3394_device::filter(double input, double cutoff)
 		m_filter_out[0] *= scale;
 		m_filter_out[1] *= scale;
 	}
-
 	return output;
 }
 


galibert:
That's kinda gross...

m1macrophage:
Open to other ideas. 

I tried a few things before resorting to this. For context, the problematic case is when the filter is configured with no input and full resonance. The fixes below, and in this PR, were only applied in this configuration.

I tried exciting the filter with noise and a pulse wave. Both fixed self-oscillation, but in both cases, the fake input came through to the output.  As an aside, an incorrect pulse input seems to be why self-oscillation was (sometimes) working before #14497.

I also tried just using DC as an input, but that didn't work.

The "least gross" approach that worked was to avoid clamping the resonance to 0.99 when there are no inputs. https://github.com/mamedev/mame/blob/098b4edc4c2147dcc97f83ee7f10f604f7fef9fb/src/devices/sound/cem3394.cpp#L221-L223

But AFAICT, that's not guaranteed to always work (though it did in my tests).  Furthermore, it looks like the clamping was done for stability, and removing it might increase the risk of the filter exploding and losing audio. So it felt a bit risky.

m1macrophage:
> The "least gross" approach that worked was to avoid clamping the resonance to 0.99 when there are no inputs [...] it felt a bit risky.

If the risk is acceptable, I can go with this approach after doing some more testing. And can revisit if there are reports of lost audio or missing sounds.

There seem to be additional safeguards for filter stability, which might help mitigate the risk:
https://github.com/mamedev/mame/blob/098b4edc4c2147dcc97f83ee7f10f604f7fef9fb/src/devices/sound/cem3394.cpp#L248-L256

galibert:
I'm quite surprised you can have zero input in the first place.  The oscillators are usually free-running and the envelope/gating happens after the filter.  Checking the 3394 schematics, it seems to be the case.  The VCA only switches between external and internal, but other than that there's always some signal doing through as far as I can see.  Am I wrong?

The self-oscillation filters out the input, so it's not a problem to get the sine if the digital implementation of the filter manages self-resonance in the first place.

m1macrophage:
Thanks for taking a closer look.

> I'm quite surprised you can have zero input in the first place. 

Yeah, it is a bit unexpected. Zero input is possible because of how waveform selection works. The waveform CV selects between "none", "triangle", "triangle + saw" and "saw" (note the switches drawn as open in the block diagram). The pulse waveform is controlled by a pulse-width CV, and is zero (well, constant) if PW is 0% or 100%.

So when the waveform CV is "none", PW CV is 0% (or 100%), and the external input is switched off, there will be zero AC to the filter.

That zero input + max resonance configuration is used as a sine oscillator, for some of the FM patches. Those are the ones currently not working.

> The self-oscillation filters out the input, so it's not a problem to get the sine if the digital implementation of the filter manages self-resonance in the first place.

The existing implementation works  well when there is an input, though it doesn't seem to cancel the input completely. This *might* be due to the resonance clamping I mentioned in the previous post.

It is the zero input configuration that does not work.

Removing resonance clamping (when the input is zero) seems to fix that. But It feels a bit more  risky. Nonetheless,  I could pursue this "no clamp" approach (with some more testing), instead of the "explicit sine oscillator" approach in this PR, and revisit if there are reported issues. Let me know if that's preferable.

galibert:
I don't think you should resonance-clamp, like, ever.  There's no reason for it, I think...

m1macrophage:
I removed the sine generator, and I removed resonance clamping when there is no input. This is a minimal change on the preexisting filter implementation, in order not to affect other sounds.

But if it is preferable, I could remove this entire section instead:
https://github.com/mamedev/mame/blob/64bba73c17ca26e5d9221cb209fdb1d625fb7fb0/src/devices/sound/cem3394.cpp#L216-L223

That would get rid of all resonance clamping. That `outscale` adjustment might not be *too* accurate anyway [*].

I've tested both options, and I didn't run into any filter instability, probably due to that protection I mentioned in an earlier reply.

Note that, with both options, self-oscillation is not entirely reliable [*]. Specifically, the level is not always the same. So while this is an incremental improvement (I've found more sounds working compared to 0.282 and 0.283), it is not perfect 


> I don't think you should resonance-clamp, like, ever. There's no reason for it, I think...

Yeah, that's more of a "hack" to address instability.  Clamping the feedback gain and/or saturating the feedback signal (and/or something else?) might be better strategies [*]


[*] In addition to those imperfections, I think the current implementation is a 2-pole rather than a 4-pole filter (though not 100% sure, not familiar with that implementation). So the filter code could use some improvements. While working on the `prophet5`, I would like to get the cem3320 filter in a better state. Once that's done, I could look into consolidating the two implementations.

galibert:
Yeah, looks like a Oberheim 12dB/decade rather than a Moog 24dB/decade.  Now, the thing to know is that I wrote the esq1 filter a while ago, before I know about that stuff and self-oscillation and friends.  I recommend you look at https://github.com/ddiakopoulos/MoogLadders and also Supercollider's MoogFF and MoogLadder, try them and see what you like most.  The papers are very interesting too.  Then we can see about which one we want to reimplement for mame.

m1macrophage:
Thanks for the pointers!

Regarding this PR, up to you if you would like to merge. In its current form (sine oscillator has been removed), it mostly fixes the regression in 0.283,  and adds new sounds compared to 0.282 and 0.283. Notably, the post-tuning filter chirp at startup for the sente6vb, and program 98 for the sixtrak.

cuavas:
I’ll take it, and risk the wrath/reversion of @galibert if he isn’t happy.

galibert:
No wrath needed, I have no problem with that.

galibert:
My aim is just that we end up with good primitives for analog synth emulation... and then we add them wholesale :-)

PR #14685, opened 2025-12-18
skeleton/digijet.cpp: New clone marked not working

stonedDiscord:
new NOT WORKING clones
------------------
Volkswagen Digijet (Polo) [stonedDiscord]


stonedDiscord:
Things I like about this:
image
Things I don't like about this:
- The inputs (Pedal/Paddle)
- How the ADC is addressed

src/mame/skeleton/digijet.cpp
```diff
@@ -2,71 +2,249 @@
...
+
+void digijet_state::prg_map(address_map &map)
+{
+	map(0x000, 0x7ff).rom().region("maincpu", 0);
+};
+
+uint8_t digijet_state::read_adc(offs_t offset)
+{
+	return m_io_adc[offset & 0x03]->read(); //FIXME this is a hack
+	// The address lines connected to the ADC are A8-A11 but those can't be accessed externally by this CPU
```
ajrhacker: P20-P23 outputs are time-multiplexed with A8-A11 on MCS-48 parts executing from external memory.

I thought that the ADC hookup might use the PROG expander protocol, but from other comments that doesn't sound like the case here.

stonedDiscord:
I don't know how to fix this "properly".
I have a reversed schematic here if anyone wants to take a look: https://github.com/stonedDiscord/Fahrzeugplatinen/tree/main/Digijet

stonedDiscord:
I have removed the ADC hack and the pedal inputs.

PR #14689, opened 2025-12-19
Lift address-bus width limit to 64 bits.

MooglyGuy:
This switches the `offs_t` declaration from `u32` to `u64`, updates the global mask in `address_map::map_validity_check`, and fixes the compile errors that fell out of the change.

Most of it was associated with logging, so I added explicitly-narrowing casts where most applicable, and switched other situations over from using `u32` for addresses to using `offs_t`.

It would be nice if @pmackinlay would have a look, as ROMP, NS32xxx, and M88x00 had more changes than most.

### DIFF
--- a/src/devices/bus/coco/coco_dwsock.cpp
+++ b/src/devices/bus/coco/coco_dwsock.cpp
@@ -162,7 +162,7 @@ u8 beckerport_device::read(offs_t offset)
 			//logerror("beckerport_device: data read 1 byte (0x%02x).  %i bytes remaining.\n", data&0xff, m_rx_pending);
 			break;
 		default:
-			fprintf(stderr, "%s: read from bad offset %d\n", __FILE__, offset);
+			fprintf(stderr, "%s: read from bad offset %04x\n", __FILE__, (uint16_t)offset);
 	}
 
 	return data;
@@ -193,7 +193,7 @@ void beckerport_device::write(offs_t offset, u8 data)
 			//logerror("beckerport_write: data write one byte (0x%02x)\n", d & 0xff);
 			break;
 		default:
-			fprintf(stderr, "%s: write to bad offset %d\n", __FILE__, offset);
+			fprintf(stderr, "%s: write to bad offset %04x\n", __FILE__, (uint16_t)offset);
 	}
 }
 
--- a/src/devices/bus/isa/ex1280.cpp
+++ b/src/devices/bus/isa/ex1280.cpp
@@ -167,7 +167,7 @@ TMS340X0_SCANLINE_RGB32_CB_MEMBER(isa16_ex1280_device::scanline_update)
 
 TMS340X0_TO_SHIFTREG_CB_MEMBER(isa16_ex1280_device::to_shiftreg)
 {
-	printf("address to shiftreg: %08x\n", address);
+	printf("address to shiftreg: %08x\n", (uint32_t)address);
 	if (address < 0xa00000)
 	{
 		memcpy(shiftreg, &m_vram[address >> 4], 0x400);
@@ -181,7 +181,7 @@ TMS340X0_TO_SHIFTREG_CB_MEMBER(isa16_ex1280_device::to_shiftreg)
 
 TMS340X0_FROM_SHIFTREG_CB_MEMBER(isa16_ex1280_device::from_shiftreg)
 {
-	printf("address from shiftreg: %08x\n", address);
+	printf("address from shiftreg: %08x\n", (uint32_t)address);
 	if (address < 0xa00000)
 	{
 		memcpy(&m_vram[address >> 4], shiftreg, 0x400);
--- a/src/devices/machine/rtc4513.h
+++ b/src/devices/machine/rtc4513.h
@@ -1,0 +1,180 @@
+// license:BSD-3-Clause
+// copyright-holders:Ryan Holtz
+/***************************************************************************
+
+    rtc4513.h - Epson RTC-4513 real-time clock emulation
+
+***************************************************************************/
+
+#ifndef MAME_MACHINE_RTC4513_H
+#define MAME_MACHINE_RTC4513_H
+
+#pragma once
+
+#include "dirtc.h"
+
+
+class rtc4513_device : public device_t,
+                       public device_nvram_interface,
+                       public device_rtc_interface
+{
+public:
+	// construction/destruction
+	rtc4513_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
+
+	int data_r() { return m_data; }
+	void data_w(int state) { m_data = state; }
+	void clk_w(int state);
+	void ce_w(int state);
+
+	auto irq_cb() { return m_irq_cb.bind(); }
+
+protected:
+	// device-level overrides
+	virtual void device_start() override;
+	virtual void device_reset() override;
+
+	// device_nvram_interface overrides
+	virtual void nvram_default() override;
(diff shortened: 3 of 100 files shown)


pmackinlay:
@MooglyGuy I will take a proper look in a couple of weeks, unfortunately can't until then.

PS: I fat-fingered the "ready for review" button, sorry.

pmackinlay:
I have had a quick look, and none of the changes stand out in any of my stuff as being problematic. FWIW, Alpha also needs >32-bit physical addresses, and this will allow unwinding an address-space hack. From memory, even mips3 has 34(?) bits of physical space which we currently don't handle.

src/devices/bus/coco/coco_dwsock.cpp
```diff
@@ -162,7 +162,7 @@ u8 beckerport_device::read(offs_t offset)
 			//logerror("beckerport_device: data read 1 byte (0x%02x).  %i bytes remaining.\n", data&0xff, m_rx_pending);
 			break;
 		default:
-			fprintf(stderr, "%s: read from bad offset %d\n", __FILE__, offset);
+			fprintf(stderr, "%s: read from bad offset %04x\n", __FILE__, (uint16_t)offset);
```
cuavas: In general, I’d change things like this to:
```c++
			util::stream_format(std::cerr, "%s: read from bad offset %04x\n", __FILE__, offset);
```

That way you get type safety, etc. from the strformat.h magic and don’t need the casts.

(This one in particular should be a `logerror` though, and not have the file name in it – it shouldn’t be sending that stuff to `stderr`.)

src/devices/bus/isa/ex1280.cpp
```diff
@@ -167,7 +167,7 @@ TMS340X0_SCANLINE_RGB32_CB_MEMBER(isa16_ex1280_device::scanline_update)
 
 TMS340X0_TO_SHIFTREG_CB_MEMBER(isa16_ex1280_device::to_shiftreg)
 {
-	printf("address to shiftreg: %08x\n", address);
+	printf("address to shiftreg: %08x\n", (uint32_t)address);
```
cuavas: Same comment for this – to send stuff to standard output with type safety, you can use:
```c++
	util::stream_format(std::cout, "address to shiftreg: %08x\n", address);
```

But the stuff in this device should probably be logmacro.h stuff, not sending stuff to standard output anyway.

src/mame/funworld/funworld.cpp
```diff
@@ -3562,7 +3562,7 @@ uint8_t royalcrdf_state::royalcrdf_opcode_r(offs_t offset)
 	if(offset < 0x800)
 		data = bitswap<8>(data ^ 0x22, 2, 6, 7, 4, 3, 1, 5, 0);
 
-	unsigned idx {bitswap<4>(offset, 8, 5, 2, 1)};
+	uint64_t idx {bitswap<4>(offset, 8, 5, 2, 1)};
```
cuavas, line 3565: The narrowing is safe in this case since the result is only ever four bits – not sure how much performance impact making `idx` 64 bits would have.

src/devices/machine/rtc4513.h
cuavas, line 1: Was this new file supposed to be part of this PR?

cuavas:
Keep in mind this will need some non-trivial DRC surgery:
* The emulated memory read instructions (`READ`, `READM`, `FREAD`) need to be changed to treat the second operand `addr` as 64-bit irrespective of instruction size.
* The emulated memory write instructions (`WRITE`, `WRITEM`, `FWRITE`) need to be changed to treat the first operand `addr` as 64-bit irrespective of instruction size.
* The C back-end needs to be updated to ensure it stores/retrieves `addr` as a 64-bit value in its intermediate format.
* The i686 back-end needs to be updated to:
  * Ensure it allocates enough space in each stack frame for the larger arguments for memory access calls (affects `HANDLE`, `RET`, entrypoint and exit point generation).
  * Stack `addr` as 64 bits, adjust offsets for other arguments, and adjust `stdcall` cleanup.
* The x86-64 back-end needs to be updated to:
  * Make sure it isn’t truncating the global mask, etc. in its address space parameters.
  * Make sure the `addr` handling is 64-bit clean.
  * Still use 32-bit operand sizes if the global mask fits in 32 bits as the encodings are more compact.
* The AArch64 back-end needs to be updated to:
  * Make sure it isn’t truncating the global mask, etc. in its address space parameters.
  * Make sure the `addr` handling is 64-bit clean.

MooglyGuy:
> Keep in mind this will need some non-trivial DRC surgery:

As-is it requires no DRC surgery at all, so any of those changes can be tackled without much time pressure.

My aim here is to be able to take a stab at an interpreter core for MIPS R10000, which has a 40-bit physical address bus. Unless someone is sitting on a DRC for a CPU that has an address bus width over 32 bits, it should not be perceived as a blocker.

src/devices/machine/rtc4513.h
MooglyGuy: Clearly, it was not.

src/mame/funworld/funworld.cpp
MooglyGuy: Safe or not, it's a compile error on Windows GCC as it weas written. So, you tell me: Should I make it a `uint64_t`, or should I cast to an unsigned?

src/devices/bus/coco/coco_dwsock.cpp
MooglyGuy: I'll be happy to address the string_format and stream_format things as a bulk change some other time, as there's a lot more than just these handfuls of spots that need reworking. There are commented-out printfs that have obviously long-since rotted, there are printfs that definitely could be logmacro, a whole bunch of that sort of thing that would be better addressed separately. The intent behind my changes was simply to address each compile error in turn as it came up.

cuavas:
> > Keep in mind this will need some non-trivial DRC surgery:
> 
> As-is it requires no DRC surgery at all, so any of those changes can be tackled without much time pressure.
> 
> My aim here is to be able to take a stab at an interpreter core for MIPS R10000, which has a 40-bit physical address bus. Unless someone is sitting on a DRC for a CPU that has an address bus width over 32 bits, it should not be perceived as a blocker.

It will still break i686 because the stack layout for calling address space read/write methods changes when the address changes from 32 bits to 64 bits. On Windows i686, the stack pointer will end up at the wrong place because of `stdcall` adjustment. The function signature change requires calls to be updated.

Thinking about it, this also affects the calls to the debugger instruction hook that the back-end generates, as that has an `offs_t` parameter.

This will break all CPUs using a recompiler whether they want bigger address spaces or not.

src/mame/funworld/funworld.cpp
cuavas: Is it onlyv getting upset because it uses uniform initialiser syntax? Does it also get upset by:
```c++
unsigned idx = bitswap<4>(offset, 8, 5, 2, 1);
```
etc.?

Compilers are weird about when they give narrowing warnings and when they don't.

MooglyGuy:
> This will break all CPUs using a recompiler whether they want bigger address spaces or not.

So I simply hallucinated sf2049, panicprk, xfiles, and windheat all booting and running just fine. Got it.

galibert:
Sounds to me like it's a good reason to get rid of the i686 recompiler.  The only reason we kept it was that it was no annoying enough to maintain yet.  Blocking 64-bits addresses is annoying enough in my book.

MooglyGuy:
Closing in favor of an updated draft PR.

PR #14692, opened 2025-12-20
minitel2: Implement sound output for DTMF tones and beeps

fabio-d:
In the minitel2, sound is generated by a speaker connected to the TS7514 modem chip. A subset of its features are implemented in this commit:
 - Mode selection, using MCBC and DTMF lines.
 - Shift register for input control data, using the RTS (clock) and PRD (data) lines.
 - Modem register RDTMF, to select what pair of frequencies to generate while in DTMF mode.
 - Modem register RWLO, selecting what to play on the speaker. In particular, monitoring of the transmitted DTMF tones and beep generation have been implemented. Monitoring of TX/RX data tones and volume control has NOT been implemented.
 - Modem register RPTF, selecting a filter on the transmitted frequencies. Only the filter on DTMF tones (to select only one of the two frequencies in the current pair) has been implemented.

This subset of features is enough to play all the sounds generated by the "demov1" ROM and the original ROM (beep on invalid keypress, DTMF dialing tones).

Note that, in order to hear DTMF tones in the original ROM, the "FV" (_fréquence vocale_) option must be selected in the configuration screen:
DC_FV

This commit also simplifies the pre-existing logic to generate a carrier-detect signal when the line control relay is closed.

cc @jfdelnero

### DIFF
--- a/src/mame/philips/minitel_2_rpic.cpp
+++ b/src/mame/philips/minitel_2_rpic.cpp
@@ -25,10 +25,10 @@
     - 24C02 EPROM
     - Modem serial interface.
     - The rear serial port (prise péri-informatique).
+    - Sound output.
 
     What is not yet implemented :
 
-    - Sound output.
     - Screen should go blank when switched off
 
     The original firmware and the experimental demo ROM are currently both working.
@@ -74,10 +74,13 @@
 #include "emupal.h"
 #include "screen.h"
 #include "softlist.h"
+#include "speaker.h"
 
 #include "logmacro.h"
 
 namespace {
+
+const int MODEM_SAMPLE_RATE = 48000;
 
 // 14174 Control register bits definition
 enum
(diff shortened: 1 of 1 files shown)


angelosa:
Is the TS7514CP modem fairly common so it may warrant a device-ization instead of plonking code in driver? iirc at least C=64 have Minitel compatible [software](https://www.zimmers.net/cbmpics/ouser6.html), and other computers of the era follow suit (been a while I went into this rabbit hole, i.e. digging mags around it).

fabio-d:
> Is the TS7514CP modem fairly common so it may warrant a device-ization instead of plonking code in driver?

I'm not aware of any other usage, all the references I can find on the Internet are for this Minitel 2.

I have seen a real Minitel 1 too (made by RTIC, later acquired by Philips), so essentially the previous revision of this Minitel 2 [btw, I have it working in MAME too, I'll have to send a PR someday! :D]. It has a TS7513 chip which, despite the similar numbering, is very different from the TS7514.

Minitels made by Alcatel used their own custom modem chips.

> iirc at least C=64 have Minitel compatible [software](https://www.zimmers.net/cbmpics/ouser6.html), and other computers of the era follow suit (been a while I went into this rabbit hole, i.e. digging mags around it).

This one seems to use an AM7911. Source: board picture at https://bizzarri.altervista.org/blog/adattatore-telematico-6499-per-c64/

fabio-d:
Oh, I forgot to mention that it is also used in the color version of the Minitel 2: https://forum.museeminitel.fr/t/minitel-2-couleur-philips/1462

It looks similar enough to the one we have that, when one day we get to implement it, I hope they can share code. But I haven't had a chance to look at it closely yet.

angelosa (approved):
This went in the backburner for a lot of time. I don't see anything bad here, beyond the modem sound stuff being inside the driver state, which I can close an eye for the time being.

PR #14705, opened 2025-12-22
segach.cpp: gets past tuning errors [superctr]

superctr:
segach.cpp: gets past tuner error 000A

segach.cpp: make tcu reg map 8-bit

segach.cpp: add authorization bits

segach.cpp: now gets past tuning

### DIFF
--- a/src/devices/bus/megadrive/cart/segach.cpp
+++ b/src/devices/bus/megadrive/cart/segach.cpp
@@ -1,5 +1,5 @@
 // license: BSD-3-Clause
-// copyright-holders: Angelo Salese
+// copyright-holders: Angelo Salese, superctr
 /**************************************************************************************************
 
 Sega Channel Game no Kanzume "digest" RAM cart, developed by CRI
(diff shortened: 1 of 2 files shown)


src/devices/bus/megadrive/cart/segach.cpp
```diff
@@ -163,15 +165,18 @@ void megadrive_segach_us_device::time_io_map(address_map &map)
 				logerror("Attempting to read with TCU DIR unset\n");
 				return 0xffff;
 			}
-			return space(0).read_word(m_tcu_index, mem_mask);
+			u16 data = space(0).read_byte(m_tcu_index, mem_mask);
+			m_tcu_index = (m_tcu_index + 1) & 0x1ff;
```
angelosa: If it's byte access doesn't need the `mem_mask`.
Also this needs to be guarded against `machine().side_effects_disabled()`, given it's done just above I guess it could be changed as:
```c++
			if (!m_tcu_dir)
			{
				logerror("Attempting to read with TCU DIR unset\n");
				return 0xffff;
			}
			u16 data = space(0).read_byte(m_tcu_index);
    			m_tcu_index = (m_tcu_index + 1) & 0x1ff;
			return data;
```

angelosa: Unless hmm, does it access two bytes with a word read/write?

superctr: It reads/writes words to this address but only 8-bits are used.

angelosa: So yeah, just the move of `machine().side_effects_disabled()` is needed then.

PR #14709, opened 2025-12-23
vfxsd (esq5505.cpp): add reconstructed drum waveform ROMs to waverom2 region.

cbrunschen:
These are not actual ROM dumps - I have sadly not been able to successfully dump these ROMs. However, the same drum waveforms that are used by the VFX-SD are also present in the SD-1 and SD-1/32, albeit at different addresses within the waveform ROMs. So these "ROMs" are reconstructed from the SD-1 waveform ROMs by copying the relevant waveform data to the locations where the VFX-SD expects to find them.

While this means that these are not the full ROM dumps, these reconstructed ones do allow the VFX-SD to play the drum sounds that differentiate it from the VFX.


rb6502:
Two things: reconstructed ROMs should be marked BAD_DUMP to advertise that real dumps are needed (that won't stop the driver from loading them and running), and the filenames should be the real chip locations, assuming that's known either from direct reference or schematics.

cbrunschen:
> Two things: reconstructed ROMs should be marked BAD_DUMP to advertise that real dumps are needed (that won't stop the driver from loading them and running), 

Done.

> and the filenames should be the real chip locations, assuming that's known either from direct reference or schematics.

Sadly I don't know exactly which chips they correspond to; I don't have VFX-SD schematics (only VFX and SD-1/32 ones, and the waveform ROMs differ) - and when I look at photos of the VFX-SD PCB, there are only 4 real chips (u54, u55, u56 and u59) that apparently cover those 3 64kSample ranges, so I'm guessing there's more nybble complexity going on here as well - which means that at least some of the reconstructed dumps correspond to a smaller number of real chips.

But in the absence of those real chips, these reconstructed dumps let the VFX-SD play all its correct sounds.

(Edited to add):

on a picture of the PCB, there are these markings on the PCB and the chips:

- U54: DRUM ED77
- U55: HI 8298
- U56: LO 5D71
- U59: NYB 6BCD

ghost:
Can you link the pcb pic? I found a not-very-good pic of what could be the VFX-SD PCB and I see a bunch of additional black mask ROMs (likely soldered in), hence why they are not dumped.
I also found another set of ROMs that don't appear to be in the driver. I can't link here but you can find them by googling "Ensoniq VFX SD (v2.00) OS ROMs and KPC"

cbrunschen:
> I found a not-very-good pic of what could be the VFX-SD PCB and I see a bunch of additional black mask ROMs (likely soldered in), hence why they are not dumped. 

All of the waveform ROMs on the VFX family of keyboards are soldered in place. I am still using my VFX-SD (which is working perfectly fine) so de-soldering the ROMs risks damaging them or the PCB and I'm not particular looking to do that. I tried using a Test Clip to connect to & read the ROMs but those attempts were sadly not particularly successful. 

The current set of ROMs came from the mainboard of an SD-1 that was already damaged and had some parts already removed, so further removing its ROMs was not a problem. Some of the waveform ROMs are identical between the VFX, VFX-SD and SD-1 - those are the ones that are used across all the variants. What's missing is those that were on the VFX-SD only, whose waveforms were incorporated into the SD-1's ROMs but in different locations - which is what I've used in this attempt at a reconstruction.

> I also found another set of ROMs that don't appear to be in the driver. I can't link here but you can find them by googling "Ensoniq VFX SD (v2.00) OS ROMs and KPC"

The OS ROMs are very much in the driver (`vfxsd_200_lower.bin` and `vfxsd_200_upper.bin`); "KPC" is the Keyboard Processor Chip, which reads the keys on the keyboard including the coils that measure key pressure, and communicates serially with the mainboard; this is essentially subsumed by the "esqpanel" which sort of emulates the combination of the Keyboard and the actual front panel (which runs on an Rockwell R6500/11, a 6502-based microcontroller with embedded mask ROM that can't really be dumped except perhaps by decapping the chip). So the KPC ROM is not really used or needed here. These are all chips that are socketed on the main and keyboard PCBs, respectively, so were easy to dump.

ghost:
The ROMs I referred to have crc32's that are not in MAME, that was my point. If ROMs are on the PCB then they should be in MAME even if they are unused.

Desoldering ROMs poses no threat at all if done correctly ;-)

To read ROMs in-circuit using a clip you have to ensure the connection from the clip to the ROM chips is perfect. This can be done by lightly sanding the upper part of the ROM legs with 400-grit sandpaper so the legs are shiny. But you also must increase the voltage to the chips by 0.1V-0.8V or even up to 1V (measure PCB under load and adjust to spec) because the current draw on the EPROM programmer is too much as it also powers all the other stuff on the PCB when the ROM is read. The power supply inside the EPROM programmer is not designed to power another large circuit board as well as its own PCB. When the voltage is increased so that it reads the chips with 5.0V-5.1V (or even 5.5V which is still in spec) then you should get correct and consistent multiple reads from the ROMs. 

A better way is to power the PCB separately from an external power supply (measure PCB voltages to ensure they are within spec 5.0V +-10%) and disconnect the VCC pin on the clip then you can read out the chips using the clip with no additional load on the EPROM programmer's internal power supply. You may also need to link the clip ground and external PSU ground to ensure the circuit shares a common ground.

Also check the board to ensure there are not multiple ROMs connected and enabled together. CE is often grounded on all ROMs but OE is usually controlled by logic or a PAL and if the board is wired so two 8-bit ROMs are read as 16-bits combined then you'll get reads from multiple chips when you try to read one. Just beep out the OE pin on each ROM checking if that pin goes to a different place on the PCB. You'll have to disable any other ROMs that are enabled at the same time as the chip you are trying to read by clipping the OE leg at the PCB (not at the plastic chip level) and lifting it or some other way of disabling it. The leg can easily be resoldered back to the PCB after reading is done. Or if you are lucky and the ROM is enabled by a PAL just remove the PAL if it is socketed.

Regardless of how you do it the proper way to read chips in-circuit is to disable the board so it can't run. This can be done by removing the main program ROMs or removing the crystal or some other way of turning the board off (disable power-on-reset circuit) so the CPU doesn't try to run the board while the ROMs are being read. Alternatively if the CPU is in a socket or main program RAM is socketed just remove it.

galibert:
Can you read the roms from the main program (yamaha usually can for rom testing, dunno about ensoniq).  And can you change the main program rom (which tends to be socketed on yamaha)?

rb6502:
OG: there is an undocumented method of reading the sample ROMs through the ES5505 that the Taito games use (they store non-waveform patch data in the sample ROMs).  The 68K program is socketed, AFAIK.

cbrunschen:
> Can you read the roms from the main program (yamaha usually can for rom testing, dunno about ensoniq). And can you change the main program rom (which tends to be socketed on yamaha)?

The CPU does not have access to the sample ROMs; only the es5505 accesses those, indirectly, through the es5701 'Super GLU' which, among other things, assembles the high bytes + low nybble into 12-bit samples.

cbrunschen:
> OG: there is an undocumented method of reading the sample ROMs through the ES5505 that the Taito games use (they store non-waveform patch data in the sample ROMs). The 68K program is socketed, AFAIK.

The es5505 accesses the samples through the es5701 'Super GLU' which assembles bytes+nybbles together into 12-bit samples, for those samples that are present on the VFX and the VFX-SD.

cbrunschen:
> The ROMs I referred to have crc32's that are not in MAME, that was my point. If ROMs are on the PCB then they should be in MAME even if they are unused.
> 
> Desoldering ROMs poses no threat at all if done correctly ;-)

Feel free to buy a VFX-SD and desolder any ROMs or read them in whatever fashion you choose.

rb6502:
> The es5505 accesses the samples through the es5701 'Super GLU' which assembles bytes+nybbles together into 12-bit samples, for those samples that are present on the VFX and the VFX-SD.

Yes, and that's a lossless process that could be easily reversed from the samples the 5505 sees.  I'll still apply this in the meantime, but it's also good to have a plan to get the real data.

PR #14713, opened 2025-12-23
Move more items into generic_cdrom.xml

chungy:
Two discs that were duplicated between Mac and PC lists (Chex Quest and My First Math Adventure) were moved to generic_cdrom. The id Anthology was moved to generic_cdrom, uniting the set to represent the physical release better.

All multi-architecture Windows NT 3.1, 3.5, and 4.0 discs were moved to generic_cdrom. Compatibility tags have been added to all entries. New "IBM5170" and "MIPS_MAGNUM" tags were defined in the pc/*.cpp and microsoft/jazz.cpp drivers. "ALPHA" and "PREP" tags are in the generic_cdrom.xml file, but have not been attached to any drivers.

### DIFF
--- a/hash/generic_cdrom.xml
+++ b/hash/generic_cdrom.xml
@@ -21,7 +21,7 @@ TODO:
 		<description>Air Warrior Plus (v1.16b)</description>
 		<year>1995</year>
 		<publisher>Kesmai Corporation</publisher>
-		<sharedfeat name="compatibility" value="IBM5170,MC68020" />
+		<sharedfeat name="compatibility" value="IBM5170,MC68020,MC68030,MC68040" />
 		<part name="cdrom" interface="cdrom">
 			<diskarea name="cdrom">
 				<disk name="air warrior plus v1.16b (1995)(on-line)[pc-mac]" sha1="8e4e21b7ac87016874e13d1c9fa02969757cf4ad" />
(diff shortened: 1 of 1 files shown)


angelosa:
The idea was to be slower in adding stuff to `generic_cdrom` (like few entries at a time), not rush things asap.
Also I don't like the tags used, like I don't want `IBM5170` and don't want Mac CPU style tagging in there (it's confusing as it gets). And preferably they should be lowercase (like the other `generic` file is).

chungy:
Could we meet half-way and at least move the Windows NT images? they're rather large, so I don't want to have duplicate items in different lists. I find it useful to have them available for the `mmr4000le` driver (in `src/mame/microsoft/jazz.cpp`), without having to type out file paths to the ibm5170_cdrom set.

The compatibility tags could be adjusted of course. Not all NT4 images support MIPS; and not all of them support Alpha either, so filtering them out seemed to be a good idea depending on driver support.

angelosa:
Are those really `jazz` rather than `MIPS_MAGNUM`?
As for `ibm5170`, just call it `ibmpc`.

chungy:
I think it should probably be `MIPS_R4000`. Magnum (and Jazz) are specific machines with the supported CPUs. There are at least a couple other MIPS computers that supported Windows NT and MAME could hypothetically support them in the future. It should be best to state the generic CPU instead of a model of computer.

Not sure when I'll have time to prepare an NT-only PR, pretty busy today :) If not today, then tomorrow.

chungy:
I will close this pull request, a replacement has been made at PR #14724. The Macintosh software could be done at a latter date, but there isn't a whole lot to deal with. :)

PR #14724, opened 2025-12-25
hash: Move Windows NT from ibm5170_cdrom.xml to generic_cdrom.xml

chungy:
A replacement for PR #14713, containing only the Windows NT software.

Compared to the previous pull request, the following changes are made:
 * The compatibility tags are lowercase.
 * `ibmpc` specifies the PC platform
 * Rather than a specific kind of machine, `mipsr4000` is used to specify the CPU type required by Windows NT. If MAME gains additional MIPS workstation drivers that support NT, they could automatically use these.


angelosa:
`pc98` also needs to be modified, with the existing SW items hooking `ibmpc,pc98` compatibility.

hash/generic_cdrom.xml
```diff
@@ -169,12 +174,1004 @@ pc98: access [PC-9801-73] registers
...
+
+	<!-- Did Service Pack 1 ever have a CD-ROM release? -->
+	<software name="winnt40sp2">
+		<description>Windows NT 4.0 Service Pack 2</description>
+		<year>1996</year>
+		<publisher>Microsoft</publisher>
+		<info name="release" value="19961213" />
+		<info name="language" value="English" />
+		<info name="usage" value="This updates the operating system, but is not the operating system installer itself." />
+		<sharedfeat name="compatibility" value="alpha,ibmpc,prep" />
```
angelosa, line 1065: I guess I'll apply then modify by myself, and I need two pieces of information:
1. what `alpha` exactly stands for? Does a driver exists for it?
2. what `prep` stands for?

ajrhacker: "Alpha" most likely refers to the DEC Alpha AXP series, which currently has a skeleton driver in MAME (dec/jensen.cpp).

"PReP" was apparently IBM's **P**owerPC **Re**ference **P**latform.

chungy: Yeah, Windows NT supported four architectures: i386, MIPS, Alpha, and PowerPC. MAME does not have any drivers for the latter two NT architectures, at least for now.

Notably for NT4 service packs, MIPS did not survive beyond SP1. PowerPC did not survive beyond SP2.

angelosa: So jensen needs the software list hook, prep needs a TODO note on top of `generic_cdrom.xml` about it.

chungy:
I was wrong about "any" drivers for alpha and prep, they're just not working. I've adjusted the list to mark almost all NT as partially supported (per the comment on top of the XML file) and software list hooks to the two drivers that could support them in the future.

PR #14728, opened 2025-12-26
Rename Alien Arena instances and add documentation why both instances…

MajorHavocOnKLOV:
… exist

### DIFF
--- a/src/mame/midway/williams.cpp
+++ b/src/mame/midway/williams.cpp
@@ -3993,8 +3993,11 @@ GAME( 1983, blasterkit, blaster,  blastkit,         blastkit, blaster_state,   e
 
 GAME( 1985, spdball,    0,        spdball,          spdball,  williams_state,  empty_init,    ROT0,   "Williams", "Speed Ball - Contest at Neonworld (prototype)", MACHINE_SUPPORTS_SAVE )
 
-GAME( 1985, alienar,    0,        alienar,          alienar,  williams_state,  empty_init,    ROT0,   "Duncan Brown", "Alien Arena",                    MACHINE_SUPPORTS_SAVE )
-GAME( 1985, alienaru,   alienar,  alienar,          alienar,  williams_state,  empty_init,    ROT0,   "Duncan Brown", "Alien Arena (Stargate upgrade)", MACHINE_SUPPORTS_SAVE )
+// Do not merge the Alien Arena instances. There is a legal reason they both exist.
+// The parent is intentionally not using ROMs we don't have a license to distribute.
+// This was done so that the game could be distributed on mamedev.org. We can't distribute the original stargate sound ROM, so it's only playable without sound using the ROMs that can be distributed.
+GAME( 1985, alienar,    0,        alienar,          alienar,  williams_state,  empty_init,    ROT0,   "Duncan Brown", "Alien Arena (without sound)",                    MACHINE_SUPPORTS_SAVE )
+GAME( 1985, alienaru,   alienar,  alienar,          alienar,  williams_state,  empty_init,    ROT0,   "Duncan Brown", "Alien Arena", MACHINE_SUPPORTS_SAVE )
 
 GAME( 1987, lottofun,   0,        lottofun,         lottofun, williams_state,  empty_init,    ROT0,   "HAR Management", "Lotto Fun", MACHINE_SUPPORTS_SAVE )
 


cuavas:
Besides the excessively long comment line, this isn’t the place for this.  Also, the only way to legitimately get an Alien Arena with sound would be to convert a Stargate, since Duncan Brown could not legally distribute the Williams sound program.  Maybe it should be called “Stargate conversion” rather than “upgrade”.

MajorHavocOnKLOV:
Thanks for the feedback. I got to play the game (a LONG time ago) and was disappointed to find all the videos on Youtube have no audio. But now I finally understand why the two instances are in mame and named that way.

ICEknigh7:
This might sound too obvious but maybe the version with no audio should not be the parent?

cuavas:
> This might sound too obvious but maybe the version with no audio should not be the parent?

We already went through this – it’s the parent because it only contains material developed by Duncan Brown, and as such is also the version we can legally distribute.

cuavas:
> Thanks for the feedback. I got to play the game (a LONG time ago) and was disappointed to find all the videos on Youtube have no audio. But now I finally understand why the two instances are in mame and named that way.

Another option might be to rename the clone to say “(with Stargate sound)” or something to give a hint about what it uses from Stargate.

ICEknigh7:
> it’s the parent because it only contains material developed by Duncan Brown, and as such is also the version we can legally distribute.

I don't see a reason there for it not to be a clone (can be legally distributed, so what?) but I see a good reason for it not to be the parent (will never have sound for legal reasons, will always be the "worse" version out of the two).

happppp:
"We already went through this" That was with a single set, not 2 sets.
Why can't the broken sound one be clone? Distribute the alienarclone.zip romset without the sound roms and MAME should be able to load it fine.

cuavas:
> > it’s the parent because it only contains material developed by Duncan Brown, and as such is also the version we can legally distribute.
> 
> I don't see a reason there for it not to be a clone (can be legally distributed, so what?) but I see a good reason for it not to be the parent (will never have sound for legal reasons, will always be the "worse" version out of the two).

Just stop it.  I don’t want to have to lock the discussion here.  The parent won’t be changed.

happppp:
Let's say we were to find out that wboy (Wonder Boy) is actually a prototype, and one of the clones is the final revision. Then the logical thing would be to swap parent/clone. You'd agree to that, yes? No matter how many years we had it as a parent set and users have to rename their rom files. It's a similar thing here.

cuavas:
No, it’s different.  That’s discovering that two versions of a game from Sega have a different relationship to what we thought.

We’ve always known that Duncan Brown released the ROMs with the intention of them being used for Stargate conversions.  The parent has always been the content actually developed by Duncan Brown, with no Williams content included.  Nothing has been “discovered” here.

No-one’s had an issue with the situation that’s been in place for well over a decade.

It’s probably a good idea to rename the version with Stargate sound to make it more obvious what the difference is, though.

MajorHavocOnKLOV:
I pushed a branch with the rename. Thanks!

mamehaze:
I'd be inclined to make the less functional version the clone too, not sure why you're so tetchy about the issue.

A more important question to ask might be was the game coded from scratch at all, or is it using functions from the original stargate / defender elsewhere in the code to get it going, in which case we probably shouldn't be distributing it at all.  I only mention this because I've looked over a lot of seemingly original games on existing platforms and many of them are built around the code powering the original games, rather than being fully original code.

synamaxmusic:
> I'd be inclined to make the less functional version the clone too, not sure why you're so tetchy about the issue.
> 
> A more important question to ask might be was the game coded from scratch at all, or is it using functions from the original stargate / defender elsewhere in the code to get it going, in which case we probably shouldn't be distributing it at all. I only mention this because I've looked over a lot of seemingly original games on existing platforms and many of them are built around the code powering the original games, rather than being fully original code.

From looking at the gameplay footage, it looks very different from other Williams titles.  So even if it is reusing some code, I feel that it's minor.  Duncan even took the time to draw his own font graphics rather than reusing the Williams font, as seen in the PDF hosted on the website: 

https://www.mamedev.org/roms/alienar/alien_arena_stuff.pdf

Considering that copyright and piracy were big concerns for arcade manufacturers in 1984, I don't think Duncan would be so reckless to steal code from Williams and then file a copyright for his game.  I can try looking at a disassembly to see how much similarities I can deduce with Alien Arena and other Williams games but again, I have a hunch that it's minor.

Regardless, from a historical and preservation perspective, I feel that presenting this game without sound as the parent is misleading as users that are not familiar with the game will assume that the sound code is lost rather than going out of their way to grab the Stargate sound ROM and plugging it into the ROM set.

Like what MajorHavoc said, I was disappointed to see that all the gameplay reference videos online don't have any audio and that can have unintended consequences when it comes to preserving game history.

cuavas:
For practical reasons, the version we can distribute needs to be the parent, not least because the "tiny" subtarget can't include a clone without its parent.

synamaxmusic:
Understood, I'll make a video for my YouTube channel so that there can be a proper example of how the game is supposed to sound.

PR #14731, opened 2025-12-26
seibu/raiden2.cpp, seibu/r2dx_v33.cpp: Updates

cam900:
seibu/raiden2.cpp:
- Add priority callback for xsedae
- Use single pass draw routine for non-blendable hardware
- Split driver state classes per hardware configurations
- Reduce duplicates
- Add notes

seibu/r2dx_v33.cpp:
- Fix spacings

### DIFF
--- a/src/mame/seibu/raiden2_v.cpp
+++ b/src/mame/seibu/raiden2_v.cpp
@@ -307,10 +307,11 @@ u32 xsedae_state::screen_update(screen_device &screen, bitmap_rgb32 &bitmap, con
 	if (BIT(~m_tilemap_enable, 4))
 		m_spritegen->draw_prio(screen, bitmap, cliprect, m_spriteram->buffer(), m_spriteram->bytes());
 
-#if DUMP_COP
-	if (machine().input().code_pressed_once(KEYCODE_Z))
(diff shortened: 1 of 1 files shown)


src/mame/seibu/raiden2_v.cpp
```diff
@@ -252,6 +285,32 @@ void raiden2_state::tilemap_draw_and_blend(screen_device &screen, bitmap_rgb32 &
...
+		m_foreground_layer->draw(screen, bitmap, cliprect, 0, 4);
+
+	if (BIT(~m_tilemap_enable, 3))
+		m_text_layer->draw(screen, bitmap, cliprect, 0, 8);
+
+	if (BIT(~m_tilemap_enable, 4))
+		m_spritegen->draw_prio(screen, bitmap, cliprect, m_spriteram->buffer(), m_spriteram->bytes());
+
+	if (machine().input().code_pressed_once(KEYCODE_Z))
+		if (m_raiden2cop) m_raiden2cop->dump_table();
```
angelosa: This stuff shouldn't be enabled in production, same in `raiden2_state::screen_update`

PR #14732, opened 2025-12-27
misc/gamtor.cpp: Add some system base ROMs [ClawGrip]

clawgrip:


src/mame/misc/gamtor.cpp
```diff
@@ -132,6 +134,52 @@ uint16_t gamtor_vga_device::offset()
...
+
+const tiny_rom_entry *gamtor_vga_device::device_rom_region() const
+{
+	return ROM_NAME(gamtor_vga);
+}
+
+
+// Money Controls Universal Hopper model MK4, standard on the Gaminator cabinet.
+// TODO: Needs to be moved to own file
+class mc_mk4_hopper_device : public ticket_dispenser_device
```
cuavas: I don’t know what you’re trying to do, but making this derive from the ticket dispenser device isn’t the way to do it.

clawgrip: I've derived from hopper_device instead

clawgrip: What I'm trying to do is create a Hopper class for a specific hopper model, so it can serve as a placeholder for the dumped ROM and it can be preserved.

Thx!

happppp: What he means is, don't derive it from that class at all. The generic hopper/ticket device is meant to be the mechanical part and sensor. If there's hopper hardware with an MCU controlling it, that can be its own device, with a machine_config that adds HOPPER.

clawgrip: Ahhhh... Ok, done!

clawgrip: Is it OK now? Thanks!

clawgrip: @happppp , @cuavas Are there any issues on this PR still needing to be fixed? Thanks!

PR #14733, opened 2025-12-27
hash/casio_rompack.xml: Some info corrections/additions

ICEknigh7:
Corrected some titles in the list of releases, added a few years and entries. [CASIO ROM PACK @Wiki, devan_wolf]

### DIFF
--- a/hash/casio_rompack.xml
+++ b/hash/casio_rompack.xml
@@ -11,187 +11,170 @@ license:CC0-1.0
 	MT-18, MT-28, MT-85, MT-86, MT-88, MT-800, MT-820, MT-830, CT-800, CT-805, CT-810, CT-840,
 	SK-8, SK-8A, KS-03, KS-02 and the DH-280 and DH-800 digital horns.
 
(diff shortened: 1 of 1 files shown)


angelosa:
It may be a good idea to add the source(s) of this info.
Also as I've said on Discord, this whole header table is something that is iffy for a data container like .xml, and the diff for moving around the markers is unreadable as proof.

ICEknigh7:
Want me to delete the whole table or...?

Already mentioned the sources, it's those two between the "[ ]".

ICEknigh7:
Personally, I do find the table a bit redundant since it keeps the dumps which have already been found and thus makes updating the softlist double the work. Maybe we should change it to your typical "Missing dumps" list which should hopefully keep getting smaller until it doesn't occupy as much space (more than 90 dumps have already been made since last update, which would be removed from the table).

angelosa:
Making a list with only the undumped games rather than a (poor man's) table sounds a good idea yeah.

ICEknigh7:
Done.

PR #14737, opened 2025-12-27
new NOT WORKING machine (Relief Pitcher System 1)

mamehaze:
new NOT WORKING machines
-------------
Relief Pitcher (System 1, prototype) [anonymous, Dutchman2000, SynaMax]

### DIFF
--- a/src/mame/atari/atarisy1.h
+++ b/src/mame/atari/atarisy1.h
@@ -10,19 +10,22 @@
 
 #pragma once
 
+#include "atarimo.h"
+#include "slapstic.h"
+
 #include "machine/6522via.h"
 #include "machine/74259.h"
 #include "machine/adc0808.h"
 #include "machine/gen_latch.h"
 #include "machine/input_merger.h"
-#include "slapstic.h"
 #include "machine/timer.h"
 #include "sound/tms5220.h"
 #include "sound/ymopm.h"
-#include "atarimo.h"
+
 #include "emupal.h"
 #include "screen.h"
 #include "tilemap.h"
+
 
 class atarisy1_state : public driver_device
 {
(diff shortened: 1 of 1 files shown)


mamehaze:
set as draft until the real PROMs are dumped, because the graphics rely on them to look remotely correct.  Was dumped from an actual PCB, owner wishes to remain anonymous.

relief_pcb

mamehaze:
also this is an entirely different game to the 1992 release, which is basically a sequel to this unreleased game

Dutchman2000:
PROMS are correct, but for an earlier LSI board, the structure is similar to other SYS1 games and are NOT corrupted. The correct PROM files for the PS and BS are now in SynaMax's hands.

synamaxmusic:
The PROM located at A7 is the same as all previous dumps, but the latest A5 dump is different (all upper nibbles are now 0xC0).  However, when I load the new dumps to MAME, we still get corruption, though it does look slightly different than what I saw previously.

synamaxmusic:
Thanks to Dutchman, the proper PROMs have been recovered.  We'll wait for Haze to return so the inputs can be implemented.

mamehaze:
I've added the correct PROMs.

I don't understand why the game hangs on 'please center joystick' though, or stalls in attract mode in a similar batting situation.

It doesn't seem to program the ADC (if it has one) in a way that generates interrupts, but maybe I'm just not understanding the ADC properly.  Apparently this used hall-effect joysticks, like Road Runner, but using that hookup doesn't work.

There's no slapstic populated so it isn't a protection issue.

Undrafting for now so the correct ROMset can be merged anyway.

synamaxmusic:
Phil B and I both confirmed that all the visible checksums written on the PROMs match up with what we have.

Dutchman2000:
I can confirm that the game uses 2 hall effect joysticks.  The source code is also available if needed.

mamehaze:
yeah, I can see where both joysticks map and they work in test mode.

can you confirm the ADC type?

the problem is it only generates an ADC interrupt if it reads/writes from the ADC, and it stops doing that quite early on.

the ADC only works in test mode if from a cold start too, if you enter test mode from the game it doesn't play nice with the ADC either.

If I force ADC interrupts in the debugger by reading from the addresses in question it generates the interrupt and the game gets past the screen.

maybe the ADC hookup in here is just entirely wrong, even if it works for other games.  @ajrhacker this seems more up your street?

(as an aside, this driver really needs a cleanup, there are tag lookups everywhere)

Dutchman2000:
Here are some source files that may help.
[SYS1HDWRMAP.TXT](https://github.com/user-attachments/files/24509785/SYS1HDWRMAP.TXT)
[CONTROL.TXT](https://github.com/user-attachments/files/24534790/CONTROL.TXT)
[system1.txt](https://github.com/user-attachments/files/24534792/system1.txt)

PR #14739, opened 2025-12-28
sinclair/specnext.cpp: Refactored INT/IRQ and implemented configurable DMA halt (delay)

holub:


src/devices/cpu/z80/z80n.cpp
```diff
@@ -31,6 +31,14 @@ z80n_device::z80n_device(const machine_config &mconfig, const char *tag, device_
 {
 }
 
+void z80n_device::nmi(int state)
+{
+	if (state != CLEAR_LINE)
+		set_service_attention<SA_NMI_PENDING, 1>();
+	else
+		set_service_attention<SA_NMI_PENDING, 0>();
+}
```
cuavas, line 40: Does this CPU really allow NMI to be deasserted externally (rather than the common edge-triggered behaviour)?

holub: Multiface button NMI holding while button pressed.

cuavas: Yeah, but NMI is edge-triggered for most Z80-like CPUs, so dropping the NMI input won’t actually deassert the NMI condition.  A logical 0->1 transition (after taking active low inputs into account) on the input triggers the condition, and it won’t be retriggered until there’s another 0->1 condition, but a 1->0 transition doesn’t clear the condition.

Unless I’m missing something (which is entirely possible), this looks like it’s treating NMI as a level-sensitive input.

holub: Honestly that's beyond my expertise. Don't forget z80n is software cpu so anything is possible. I can only judge by testcases I've tried, it's sometime very exhausting to do everything "right" by comparing to Verilog, sorry.

galibert: The verilog is available somewhere?  I can probably check whether it's edge.

holub: https://gitlab.com/SpectrumNext/ZX_Spectrum_Next_FPGA/-/tree/master/cores/zxnext/src/cpu?ref_type=heads

galibert: https://gitlab.com/SpectrumNext/ZX_Spectrum_Next_FPGA/-/blob/master/cores/zxnext/src/cpu/t80n.vhd?ref_type=heads#L1665

It's edge indeed.

cuavas: Why does it have this member function at all, rather than just using `set_input_line`?

holub: I'm positive set_input_line didn't work for me in some case. Let me try to recall what was it.

holub: Thank you both for the inspiration!
#15086

PR #14743, opened 2025-12-29
vbaby_cart.xml: Add the United Kingdom and French versions

vtarchive:


vtarchive (commented):
i did add the backyardigans subtitle

PR #14745, opened 2025-12-29
new WORKING machine (Moon Raker)

mamehaze:
new WORKING machine
------------
Moon Raker [Dillweed, SNESNESCUBE64]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -28466,6 +28466,7 @@ mlander
 mlandera
 moonbase
 moonbasea
+mraker
 orbite
 ozmawars
 ozmawars2
--- a/src/mame/midw8080/8080bw.cpp
+++ b/src/mame/midw8080/8080bw.cpp
@@ -1603,6 +1603,38 @@ static INPUT_PORTS_START( rollingc )
 	PORT_DIPSETTING(    0x00, "RC=5000 / MB=2000" )
 INPUT_PORTS_END
 
+
+static INPUT_PORTS_START( mraker )
+	PORT_INCLUDE( sicv_base )
+
+	PORT_MODIFY("IN0")
+	// nothing in here seems to matter
+
+	PORT_MODIFY("IN1")
+	PORT_BIT( 0x01, IP_ACTIVE_HIGH, IPT_COIN1 )
+	PORT_BIT( 0x02, IP_ACTIVE_HIGH, IPT_START2 )
+	PORT_BIT( 0x04, IP_ACTIVE_HIGH, IPT_START1 )
+	PORT_BIT( 0x08, IP_ACTIVE_HIGH, IPT_JOYSTICK_UP )
+	PORT_BIT( 0x10, IP_ACTIVE_HIGH, IPT_BUTTON1 )
+	PORT_BIT( 0x20, IP_ACTIVE_HIGH, IPT_JOYSTICK_LEFT )
+	PORT_BIT( 0x40, IP_ACTIVE_HIGH, IPT_JOYSTICK_RIGHT )
+	PORT_BIT( 0x80, IP_ACTIVE_HIGH, IPT_JOYSTICK_DOWN )
+
+	PORT_MODIFY("IN2")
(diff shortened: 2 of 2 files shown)


mamehaze:
could probably do with somebody looking at the code to see if it reads any dip settings from anywhere, but it's possible it has no settings at all.  There's a Galaxian style starfield that apparently needs adding, and it leans on the existing samples, which may not be right for this game (I don't have them to test)

mraker_pcb

mamehaze:
if https://www.youtube.com/watch?v=4kA_Ib7ZYgY is correctly hooked up, blue and green may need swapping too
neither scheme matches the flyer picture though https://flyers.arcade-museum.com/videogames/show/5915

currently the driver uses the same scheme as Rolling Crash / Moon Base, and Rolling Crash has a rainbow pattern, which I guess was used to figure that out.

ghost:
The colors in that vid are *definitely* wrong. It looks very bad and there are too many bright pinks. The score at the top left should be pale blue/cyan where it is yellow in the video.
I think red and blue are incorrectly swapped.
Edit: Although the video shows the player ship is blue and that matches the flyer. So could be different color PROMs as well as rgb swaps.

ghost:
Also check the recent Cosmo src changes that adds a Galaxian-style starfield based on my fully-reversed schematic. Space Invaders hardware can't produce a starfield by itself so there is probably an additional top board (like Cosmo has) that does extra stuff, possibly also with additional sounds too. It's also possible the color PROMs are different as they are usually socketed and on the middle CPU board on that hardware. Try all the dumped PROMs from other SI-based games in case it's using PROMs from an existing ROM set. It's a pity the rest of the boards is not available.

mamehaze:
it is not believed this hardware type uses PROMs, it appears to have been replaced by RAM.

wd40yasu:
uploaded schematics (scanned by Andrew Welburn of Dumping Union)

docs/schems/moon raker schematics.pdf

ghost:
Nice! No PROMs means perfect colors are possible. Back to the schematic to work out the rest of the emulation details ;-)
As I expected there's a 4th board with a (currently missing) 2708 (1kB) ROM doing some extra stuff (possibly the starfield only?). But it should be possible to get 98% of the game perfect if someone can be coaxed into working on the discrete sound emulation which looks very Space Invaders-ish :-)

happppp:
Yes, it has a starfield according to schematics (and the PCB video too) with a ROM on it.
So, add a "stars" region, with NO_DUMP 0x400 ROM "2708.j" // label unknown

Imperfect graphics flag too, and the // should have a galaxian style starfield? comment can be removed. Or at least remove the questionmark, and put the comment at the end of the game entry instead of above it. And it's not galaxian style, that one doesn't have a ROM for the stars

happppp:
Could you remove this?: PORT_INCLUDE( sicv_base )

Otherwise you get duplicate l/r/button1 controls because these are unused:
INVADERS_CONTROL_PORT_P1
INVADERS_CONTROL_PORT_P2
INVADERS_CAB_TYPE_PORT

mamehaze:
I'm leaving now for a holiday, going to be away 1-2 weeks in the middle of nowhere with minimal tech.  If you want those changes made quickly please make them.

happppp:
Hmm, ok I'll do it, have a nice vacation.

happppp:
FWIW: Some small gfx glitches I saw, are the same on the PCB video, so those are not a bad dump: The "A" from RAKER on titlescreen, the bottom-right of "C" char, the top-right of "3" char.

ghost:
Maybe the starfield is similar or the same as Cosmo?

TheDillestWeed:
It depends. It seems that the MR-4 board does the star patterns, or it's at least involved directly with video. However, the way it works is very interesting. It seems to take data from the ROM and use some counting logic and a RAM chip to load it into video summation in a sequence, which would explain how they get their "twinkling" effect. Though I find it funny it's only a half-second loop pattern if you reference the youtube video.

It's almost similar to how Atom Smasher does it, which also appears to use a ROM. However, that hardware scrolls a fixed image.

PR #14749, opened 2025-12-31
apple2gs: fix 1-sec and 1/4-sec IRQs

arekkusu42:
([More](https://github.com/mamedev/mame/pull/14590) small prerequisites split out from a larger 50Hz video timing WIP PR):

This commit fixes the 1-second and 1/4-second IRQs to match observed hardware behavior:

* MegaII 1/4-sec IRQs (used by i.e. AppleTalk) use the same timing mechanism as the [FLASH text rate](https://github.com/mamedev/mame/pull/14554), which means they are actually 3.75 Hz (NTSC) or 3.125 Hz (PAL), not 4.0 Hz, and always trigger on scanline 256 (wrapped to 250 or 200.) This is hinted at by the Apple IIgs Firmware Reference, which says the "[3.75Hz interrupt](https://archive.org/details/Apple_IIgs_Firmware_Reference_HiRes/page/n309/mode/2up)" "occurs [every 0.26667 second](https://archive.org/details/Apple_IIgs_Firmware_Reference_HiRes/page/n197/mode/2up) for AppleTalk use."

* VGC 1-sec IRQs (used by i.e. the CDA Control Panel) really trigger every second via the RTC, and are not tied to NTSC/PAL or any particular scanline.  Using the C02E/2F video counters on hardware to measure the drift between IRQs relative to the video scanner, we can observe `-[1319..1326]` 1 MHz cycles (about 20.3 scanlines, with 7-cycle jitter) every second.  This corresponds to the difference between 60 Hz and the actual 59.92 Hz CRT refresh rate.  In another experiment (not part of this PR) which changes [A2GS_1M](https://github.com/mamedev/mame/blob/535cec072009e79b22f187977d4d001830e7036c/src/mame/apple/apple2gs.cpp#L97) to 1020484, we can observe in MAME a very close `-[1312..1319]` cycle drift: accurate down to the precision of the average frequency 32-bit integer approximation.

<BR>

This commit also fixes [a subtle off-by-one bug](https://github.com/mamedev/mame/blob/535cec072009e79b22f187977d4d001830e7036c/src/mame/apple/apple2gs.cpp#L1471) in `apple2gs_state::get_vpos()`, which only affected scanlines 246-256.  The new logic is NTSC/PAL-agnostic.

<BR>

This commit also changes C044/45 to always return zero (not floating bus).  These [MMDELTAX/Y registers](https://archive.org/details/Apple_IIgs_Firmware_Reference_HiRes/page/n307/mode/2up) have some behavior described in [Cortland documentation](https://www.brutaldeluxe.fr/documentation/cortland/v5/Cortland%20Custom%20ICs%20-%20Wayne%20Lowry%20-%20Preliminary%20notes%20-%2019860214.pdf#page=29), but AFAICT this is unimplemented on the MegaII as shipped in the IIgs-- no interrupts or mouse deltas are generated regardless of what C041 INTEN bits 0-2 are set to.

Before this change, AppleTalk hangs on boot; to reproduce:
1) Set slot 1 to AppleTalk in Control Panel
2) boot GS/OS 6.0.1 or 6.0.4 with AppleShare installed (an easy reproducer is the 6.0.4 Live.Install.po)
-> AppleTalk enables 1/4-second C041 IRQ during boot, and selects SCCAREG/DATA registers, but then spins forever polling C044 expecting zero, at 02/B9A4 (6.0.1), or 02/BABF (6.0.4).

It appears that AppleTalk is setting C041 bit 0, and then using C044/45 to somehow detect activity on an SCC channel, and waiting if the channel is busy.  The IIgs schematic shows that [MegaII MSEX1/MSEY1 tap SCC RXDA/RXDB](http://cini.classiccmp.org/pdf/Apple/Apple%20IIgs%20schematic.pdf#page=4), but it's unclear exactly how these (signed, clamped?) counters increment.

<BR>

Testing:
My [previous IRQTEST](https://github.com/mamedev/mame/pull/14277) is updated here to carefully verify the 1-second and 1/4-second IRQ frequency, against both NTSC and PAL display modes.  MAME 0.284 fails:
IRQTest_0284_before
[IRQTest_251229.zip](https://github.com/user-attachments/files/24390286/IRQTest_251229.zip)

After this PR, 1-second IRQs match hardware behavior and pass, and 1/4-second IRQs work as expected in NTSC mode; a separate WIP PR implements PAL mode.

### DIFF
--- a/src/mame/apple/apple2gs.cpp
+++ b/src/mame/apple/apple2gs.cpp
@@ -473,7 +473,7 @@ private:
 
 	// clock/BRAM
 	u8 m_clkdata = 0, m_clock_control = 0;
-	u8 m_clock_frame = 0;
+	void rtc_vgc(int cko);
 
 	void lcrom_update();
 	void do_io(int offset);
@@ -772,7 +772,6 @@ void apple2gs_state::machine_start()
 	save_item(NAME(m_textcol));
 	save_item(NAME(m_clock_control));
 	save_item(NAME(m_clkdata));
-	save_item(NAME(m_clock_frame));
 	save_item(NAME(m_motors_active));
 	save_item(NAME(m_slotromsel));
 	save_item(NAME(m_diskreg));
@@ -833,7 +832,6 @@ void apple2gs_state::machine_reset()
 	m_ramrd = false;
 	m_ramwrt = false;
 	m_video->set_newvideo(0x01); // verified on ROM03 hardware
-	m_clock_frame = 0;
 	m_slot_irq = false;
 	m_clkdata = 0;
 	m_clock_control = 0;
@@ -1053,18 +1051,6 @@ TIMER_DEVICE_CALLBACK_MEMBER(apple2gs_state::apple2_interrupt)
 
 		m_adbmicro->set_input_line(m5074x_device::M5074X_INT1_LINE, ASSERT_LINE);
 
-		m_clock_frame++;
-
-		// quarter second?
-		if ((m_clock_frame % 15) == 0)
-		{
-			m_intflag |= INTFLAG_QUARTER;
-			if (m_inten & 0x10)
-			{
-				raise_irq(IRQS_QTRSEC);
-			}
-		}
-
 		// 3.5 motor off timeout
 		if (m_motoroff_time > 0)
 		{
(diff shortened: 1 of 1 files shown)


arekkusu42:
If anyone wants to get LocalTalk working in MAME, it looks like [GSPort has functional AppleTalk emulation](https://david-schmidt.github.io/gsport/appletalk.html), and some [notes about C044/45](https://apple2infinitum.slack.com/archives/CPSGNGE05/p1686604587116699).  For now, "not hanging on boot" is forward progress.

rb6502:
Mini vMac has GSPort-compatible LocalTalk emulation as well, and I gather it's possible to bridge it to real hardware now as well.  Vas called dibs on that ages ago but he's pretty busy otherwise.

PR #14750, opened 2025-12-31
innotv_innotabmax.cpp: Add note about undumped system

vtarchive:
InnoTab Max has backwards compatiblity with most InnoTab titles
https://www.innotabsoftware.com/

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -48609,7 +48609,7 @@ gl8008cx
 gtbmcx
 
 @source:vtech/innotv_innotabmax.cpp
-innotv
+innotvuk
 
 @source:vtech/inteladv.cpp
 cars2lap


vtarchive:
why is this failing?

angelosa:
Because you are changing the romset ID without changing the correlated `ROM_START` and actually reflecting the change in [mame.lst file](https://github.com/mamedev/mame/blob/master/src/mame/mame.lst)

angelosa:
Also, you should test whatever you change, locally. I suspect you are not doing that ...

happppp:
It still won't compile correctly, because mame.lst wasn't updated.
IMO the romset rename is unnecessary, we normally only add region postfix to the romset name when we have 2 or more regions. eg. setname and setnamej (Japan).

vtarchive:
it works.

PR #14752, opened 2026-01-01
Canon X07 - fix VRAM access functions in T6834 subsystem emulation

Aurelien34:
Access by the Z80 main CPU to the VRAM is done through the T6834 subsystem, which hosts its own RAM.

In MAME, the subsystem functions are implemented using a simple switch/case statement.
Functions 0x05 and 0x06 allow subsystem RAM access. Access to the specific VRAM location did not take drawing functions impacts on VRAM.
With this changes, VRAM accesses (and only VRAM accesses) target the LCD component memory directly.

### DIFF
--- a/src/mame/canon/x07.cpp
+++ b/src/mame/canon/x07.cpp
@@ -103,6 +103,15 @@ void x07_state::t6834_cmd (uint8_t cmd)
 				data = 0x0a;
 			else if(address == 0xd000)
 				data = ioport("BATTERY")->read();
+			else if (address >= 0xe000 && address < 0xe200 && ((address & 0xf) != 0xf)) // VRAM and col 0-119 ?
+			{
+				data = 0;
+				const int y = (address & 0x1ff) >> 4;
+				const int x = (address & 0x000f) * 8;
+				// Read video data from LCD
+				for (int p = 0; p < 8; ++p)
+					data |= (m_lcd_map[y][x+p] ? (1 << (7-p)) : 0);
+			}
 			else
 				data = m_t6834_ram[address & 0x7ff];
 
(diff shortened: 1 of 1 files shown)


Aurelien34:
And it's working:
image

src/mame/canon/x07.cpp
```diff
@@ -103,6 +103,15 @@ void x07_state::t6834_cmd (uint8_t cmd)
 				data = 0x0a;
 			else if(address == 0xd000)
 				data = ioport("BATTERY")->read();
+			else if (address >= 0xe000 && address < 0xe200 && ((address & 0xf) != 0xf)) // VRAM and col 0-119 ?
+			{
+				data = 0;
+				int y = (address & 0x1ff) >> 4;
```
rb6502: y and x can be const.  Same thing in the second case below.  Looks fine otherwise!

Aurelien34: Done, thanks for your feedback!

angelosa (approved):
The memory system should really use a standard memory map, this is something not reflected in header. I'll add it later after this is merged.

Aurelien34:
> The memory system should really use a standard memory map, this is something not reflected in header. I'll add it later after this is merged.

I am doing some code on the X07 in the context of a game jam. I might have some other T6834 functions to implement or fix in the coming weeks.
Shoud I wait for you to makr changes to the memory management?
To do things correctly, the memery space should be clearly defined, the gralhic functions should access the VRAM section, and the VRAM should be transfered to the LCD every frame.
As the T6834 is not really implemented as a controller executing a program in ROM, it is hard to do things correctly :-(

angelosa:
Sorry but I don't have spare time to look at this in the short term, I'm just waiting for the CI to complete before applying.

Aurelien34:
> Sorry but I don't have spare time to look at this in the short term, I'm just waiting for the CI to complete before applying.

I wasn't trying to tell you what to do, I'm just pointing out that this part of the driver is nowhere near perfect and there’s a huge amount of work to do to fix it.
Thanks for your work on this!

PR #14754, opened 2026-01-01
sinclair/specnext_copper.cpp: Optimized execution for sequential NOPs

holub:


src/mame/sinclair/specnext_copper.cpp
```diff
@@ -79,31 +79,42 @@ void specnext_copper_device::data_w(u16 addr, u8 data)
...
+		m_copper_dout, m_copper_list_data and m_copper_list_addr are set to the future value.
+		Ideally they must be recalculated if data_w() happens. However, in practice racing
+		with the copper's counter is unreasonable, so we can avoid that complexity till required.
+		*/
+		++times;
+	} while (m_copper_dout == 0 && m_copper_list_addr < 0x400);
 
 	m_copper_list_addr %= 0x400;
-	m_timer->adjust(attotime::from_hz(clock()));
+	m_timer->adjust(times * attotime::from_hz(clock()));
```
cuavas: Can you use `m_timer->adjust(clocks_to_attotime(times));` to do this?

holub: sure

PR #14760, opened 2026-01-02
cpu/drccache.cpp: Defer allocating to start, allow forcing W^X mode

cuavas:
This makes a few changes to the DRC cache.  I expect this won't be contentious, but @galibert or @MooglyGuy might have something to say about it.

This adds a user-facing option to disable use of read/write/execute pages even if they're permitted on the host system (-[no]drc_rwx).  This is useful for debugging or profiling W^X mode on Windows/Linux/whatever when RWX pages are permitted.  People may also want to use it if they want to make MAME marginally less insecure or something (it might prevent certain classes of VM escape bug).

This defers DRC cache allocation from construction to start time.  This has two main benefits right now:
* It improves the performance of `-validate`, `-listxml`, the internal UI, etc. because it's no longer asking the virtual memory system for megabytes of memory it has no need for and messing with page tables.
* It means systems/devices can override the cache size for CPUs at configuration time, which may improve performance if they have a large working set.  To this end, MIPS III and PowerPC get `set_drc_cache_size(size_t)` configuration member functions to set the desired cache size.

This adds helpers for allocating structures or objects in the DRC cache.  This makes syntax a bit prettier in CPU cores.  It prints an error message if the structure/object could possibly be misaligned (changing the heap management approach to correctly align things that aren't satisfied with `max_align_t` alignment is too intrusive to do in this PR).

This adds some DRC cache stats to verbose output on destruction.  Debug builds have more counters than release builds.  This can give you an idea of whether the cache is big enough.  Note that cache use will be higher with `-drc_log_uml`, and a lot higher with `-drc_log_native`, especially on i686 and x86-64.

There's also some general cleanup in some of the CPU cores, but nothing major.

### DIFF
--- a/docs/source/commandline/commandline-all.rst
+++ b/docs/source/commandline/commandline-all.rst
@@ -3923,7 +3923,9 @@ Core Misc Options
 
 **-[no]drc**
 
-    Enable DRC (dynamic recompiler) CPU core if available for maximum speed.
+    Enable DRC (dynamic recompiler) CPU cores if available.  Turn this option
+    off to use interpreter CPU cores if available.  This option does not affect
+    CPUs that only support one core type.
 
     The default is ON (**-drc**).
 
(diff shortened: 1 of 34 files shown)


ajrhacker:
While I can't say I'm well-versed in the technical details of MAME's DRC, this is the kind of change I enthusiastically approve of.

cuavas:
Well I got one enthusiastic approval and no complaints.  I guess I'll take it.

galibert:
Do we want that to be options?  I don't like the idea of having more cases to test for, which we won't actually do most of the time.  Even for the cache size, I seem to remember the maximum is 32M or something?  Why would we ever want less, given nowadays ram sizes?

cuavas:
> Do we want that to be options? I don't like the idea of having more cases to test for, which we won't actually do most of the time.

I could make it use W^X mode all the time, but that hurts performance when things need to generate code frequently if you're on Windows or Linux. I generally don't like to throw away performance.

Previously, there was no option, so you need to edit the code if you wanted to test W^X mode on a system that permits RWX pages, which people are even less likely to do. If W^X mode gets broken, we'll hear about it from the Mac users pretty quickly (e.g. @rb6502 won't be able to run emulated PowerPC systems).

The existence of W^X mode doesn't affect developers the vast of the time, anyway. It's taken care of in the utility classes, and you don't even notice when you're writing a CPU core.

> Even for the cache size, I seem to remember the maximum is 32M or something? Why would we ever want less, given nowadays ram sizes?

The default is 32MiB for some CPUs, but whatever the default is for MIPS III is more than enough for the Midway MIPS/Voodoo games, but woefully inadequate for IRIX. Apparently running IRIX with less than 160MiB cases it to constantly flush the cache and recompile everything due to the much larger working set. In general, games using tight code have far smaller working sets than general-purpose operating systems running desktop environments.

I don't want to be bumping the default cache size for CPUs to whatever UNIX desktops need when most people are just using them to run a few 3D arcade games.

galibert:
You sure it would be a problem to go 256M?  I mean, it's just address space, it doesn't become real memory unless it's actually used.  So if we don't go and poke at every other byte just for fun, it shouldn't be noticeable.

cuavas:
That's platform-dependent. Linux and macOS will over-commit without backing and just kill processes if it runs out. Windows won't over-commit – it will reject mapping requests if it doesn't have RAM or swap backing available.

PR #14763, opened 2026-01-02
machine/spi_sdcard.cpp: Added configurable read/seek delays

holub:


src/mame/sinclair/specnext.cpp
```diff
@@ -3817,10 +3817,12 @@ void specnext_state::tbblue(machine_config &config)
 	SPI_SDCARD(config, m_sdcards[0], 0);
 	m_sdcards[0]->set_prefer_sdhc();
 	m_sdcards[0]->spi_miso_callback().set(FUNC(specnext_state::spi_miso_w));
+	m_sdcards[0]->set_delays_ext(0, 1); // This is a minimum dalay which allows Tha Atic Atac's loader to work.
```
angelosa: `dalay`

rb6502: LOL.  I almost thought that said "Dalek".  But yes, please fix the spelling of "delay".

holub: I did. This box shows original state.

cuavas:
Speaking of this device, it’s suffering from a misunderstanding of how the interface actually works.  The SD Card state machine (represented in `spi_sdcard_device::sd_state`) only applies in native SD Card interface mode.  In SPI mode, there’s only one “state”, and commands can happen in any sequence.

There are also major issues with how the device reports errors and operation results back to the host.

cuavas:
Actually, this change is broken.

Now if you don’t call `set_delays_ext`, you lose the `SPI_DELAY_RESPONSE` delay from the default argument value in `send_data`, because it passes the default zero values from `m_delay_seek` and `m_delay_data`.

I need to at least hack around this for the release.

cuavas:
In fact, this change doesn’t increase the delay at all.  Previously, all those calls to `send_data` were using the default `delay` value of 1 (from `SPI_DELAY_RESPONSE` in the header).  The only place it had no delay at all is the one explicitly commented “zero delay” at what’s now line 425 (and it absolutely must be zero in that situation).

The net result of this change is that *all* the calls to `send_data` get zero delay, except for the one on line 444 which gets 1 (which is what it got previously anyway from the default argument value), and then only for Spectrum Next, and not for any other machines because nothing else sets delays.

Since this doesn’t do what it purports to do and changes behaviour for every system using the SD Card, I’m reverting it.  Whatever was going wrong with Spectrum Next wasn’t the need for additional delay if removing delays fixed it.

It’s possible there are other situations where an immediate response is required, but I’d have to go through the interface standard documentation to find them.

holub:
It's not for Spectrum Next only. We have tested SD Cards from different vendors and these delays are different for each Vendor/Product. In MAME we just have it set to "min" but none of the cards we tested doesn't have it so quick amd idling for some period.

cuavas:
Read what I wrote, read the code.  I know the delay depends on the card.

But your change *decreases* delays because it replaces the default `delay` value of 1 (from `SPI_DELAY_RESPONSE` in the header) with the default `m_delay_seek`/`m_delay_data` values of zero for all the systems that haven’t overridden these values.

Your change *decreases* the delay in almost all cases.

cuavas:
To further elaborate, there are two thing involved in these “delays”: the number of SPI clock cycles it takes to turn around the interface after receiving a command, and the time time it takes to actually execute the command, which is largely independent of the SPI clock.

The number of cycles to turn around the interface and send a checksum result after a write *must* be zero – this is specified in the standard, and `linux4004` depends on this being correct.  The actual write will take longer to complete – the host needs to poll status to determine when the write has completed, and whether it succeeded.  MAME does not emulate this at all, it fakes a bad checksum response if the write fails, and always treats successful writes as completing immediately.

In practice, the number of cycles to turn around the interface after a read command is always at least one, and a number of systems in MAME have firmware/software that depends on it.  This PR broke these systems by changing that to zero for all systems that didn’t explicitly call `set_delays_ext` to increase it.

The other part of the timing is the time to actually access the Flash memory.  This depends on many things, including the model of card, the controller it uses, the type of Flash memory it uses (e.g. SLC vs MLC vs TLC), the age of the Flash memory, how many times it has been rewritten, etc.  A key point is that this is independent of the SPI clock frequency, so you can’t specify it in terms of that.  You need to specify it in real time units.

To handle it realistically, you need to get the current time from `machine().time()` on receiving a valid command, and calculate the completion time based on that.  Then don’t start sending a response until after the completion time has passed *and* the required number of SPI cycles to turn around the interface have occurred (with the exception of commands that require immediate checksum responses).

holub:
I screwed with defaults, agreed and sorry.
Surprisingly AticAtac we tried to fix with this, works even without now (looks like delays were bandaid but not actual treatment).
If you want I can fix defaults in reverted code, but we'll have these delays without actual consumers, so I'm sceptical to bring it back.

PR #14765, opened 2026-01-03
sinclair/specnext.cpp: Added PS/2 Keyboard mapping

taylorza:
Updated the keyboard mapping to match the physical hardware used with a PS/2 keyboard

### DIFF
--- a/src/devices/cpu/sharc/sharc.cpp
+++ b/src/devices/cpu/sharc/sharc.cpp
@@ -682,9 +682,9 @@ void adsp21062_device::device_start()
 	m_core->fp0 = 0.0f;
 	m_core->fp1 = 1.0f;
 
-	save_item(NAME(m_core->pc));
 	save_pointer(NAME(&m_core->r[0].r), std::size(m_core->r));
 	save_pointer(NAME(&m_core->reg_alt[0].r), std::size(m_core->reg_alt));
+	save_item(NAME(m_core->pc));
 	save_item(NAME(m_core->mrf));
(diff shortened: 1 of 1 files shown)


src/mame/sinclair/specnext.cpp
```diff
@@ -3042,6 +3042,26 @@ void specnext_state::map_regs(address_map &map)
 INPUT_PORTS_START(specnext)
 	PORT_INCLUDE(spec_plus)
 
+	// PS2 Keyboard Mapping
```
angelosa: I'm assuming it's PS/2, not PS2.

angelosa:
Are the default mappings / charsets you used not clashing with existing keyboard inputs?

holub:
It's fixing up default as it's technically different keyboard. E.g. Symbol and Control shifts swapped to reflect this transition. Original 40-keys postion them opposite.

We decided switch to PS/2 mapping, that's not mapping of original Next's keyboard.
https://www.specnext.com/ps2-keyboard/

src/mame/sinclair/specnext.cpp
taylorza: You are correct. Thanks, I updated the code comment.

holub:
Resolve the conflict please

taylorza:
@holub - the conflict has been resolved

holub:
Great!
Gents, let's merge it, I see no reason to hold it more.

holub:
> Great! Gents, let's merge it, I see no reason to hold it more.

+1

PR #14770, opened 2026-01-03
Fix floating value options when the OS is set to certain languages

AnthonyRyuki:
As changing the entire locale in MAME to a different option than the OS setting is problematic, I wonder if just changing it temporarily to parse the values is OK.
Interestingly, this problem only occurs with the floating values in the "Advanced Options" menu. The other floating values aren't affected.
As a reminder, these options do not work correctly when the OS is set to those languages where floating values are displayed with a ',' instead of a '.'.


cuavas:
It should be really obvious why this doesn't work.

PR #14771, opened 2026-01-03
concept: Use 1.7 ROM by default

eschaton:
Set the Corvus Concept 1.7 (Version 1 Level 7) ROMs as the default.

With this, I've been able to successfully boot, install, and use the Corvus Concept OS 1.2 from Bitsavers.

### DIFF
--- a/src/mame/concept/concept.cpp
+++ b/src/mame/concept/concept.cpp
@@ -196,23 +196,26 @@ ROM_START( concept )
 	ROM_REGION16_BE(0x100000,"maincpu",0)   /* 68k rom and ram */
 
 	// concept boot ROM
-	ROM_SYSTEM_BIOS(0, "lvl8", "Level 8" )  // v0?
-	ROMX_LOAD("bootl08h", 0x010000, 0x1000, CRC(ee479f51) SHA1(b20ba18564672196076e46507020c6d69a640a2f), ROM_BIOS(0) | ROM_SKIP(1))
-	ROMX_LOAD("bootl08l", 0x010001, 0x1000, CRC(acaefd07) SHA1(de0c7eaacaf4c0652aa45e523cebce2b2993c437), ROM_BIOS(0) | ROM_SKIP(1))
+	ROM_DEFAULT_BIOS("v17")
+
+	ROM_SYSTEM_BIOS(0, "lvl6", "Level 6" )  // v0?
+	ROMX_LOAD("cc06h", 0x010000, 0x1000, CRC(66b6b259) SHA1(1199a38ef3e94f695e8da6a7c80c6432da3cb80c), ROM_BIOS(0) | ROM_SKIP(1))
+	ROMX_LOAD("cc06l", 0x010001, 0x1000, CRC(600940d3) SHA1(c3278bf23b3b1c35ea1e3da48a05e877862a8345), ROM_BIOS(0) | ROM_SKIP(1))
 
 	ROM_SYSTEM_BIOS(1, "lvl7", "Level 7" )  // v0? v1?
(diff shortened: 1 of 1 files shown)


startaq:
Can you order them chronologically and use `ROM_DEFAULT_BIOS` to select the default to use? Also, it would be good to add a comment that this version works successfully with OS 1.2.

eschaton:
Can do, thanks for the feedback!

src/mame/concept/concept.cpp
```diff
@@ -196,23 +196,26 @@ ROM_START( concept )
...
-	ROMX_LOAD("cc06h", 0x010000, 0x1000, CRC(66b6b259) SHA1(1199a38ef3e94f695e8da6a7c80c6432da3cb80c), ROM_BIOS(2) | ROM_SKIP(1))
-	ROMX_LOAD("cc06l", 0x010001, 0x1000, CRC(600940d3) SHA1(c3278bf23b3b1c35ea1e3da48a05e877862a8345), ROM_BIOS(2) | ROM_SKIP(1))
+	ROM_SYSTEM_BIOS(2, "lvl8", "Level 8" )  // v0?
+	ROMX_LOAD("bootl08h", 0x010000, 0x1000, CRC(ee479f51) SHA1(b20ba18564672196076e46507020c6d69a640a2f), ROM_BIOS(2) | ROM_SKIP(1))
+	ROMX_LOAD("bootl08l", 0x010001, 0x1000, CRC(acaefd07) SHA1(de0c7eaacaf4c0652aa45e523cebce2b2993c437), ROM_BIOS(2) | ROM_SKIP(1))
+
+	// the 1.7 ROM (at least) works with Corvus OS 1.2
+	ROM_SYSTEM_BIOS(3, "v17", "Version 1 Level 7" )
+	ROMX_LOAD("ccb17h.bin", 0x010000, 0x1000, CRC(6dd9718f) SHA1(7427d218dc13b1b319866d5a1e59e152fab082b3), ROM_BIOS(3) | ROM_SKIP(1))
+	ROMX_LOAD("ccb17l.bin", 0x010001, 0x1000, CRC(107a3830) SHA1(0ea12ef13b0d11fcd83b306b3a1bb8014ba910c0), ROM_BIOS(3) | ROM_SKIP(1))
```
cuavas, line 216: Why the change to the labels here?  We usually use "<label>.<location>" format for these.

eschaton: Sorry, I'm not up on the terminology—which strings are you referring to as labels? I'm happy to change it however you like.

The Corvus code sometimes says "Version 1 Level 7" and sometimes "V1.7" and sometimes "1.7" while the code in the file uses "lvl" for the other values. For the filenames, I just lowercased the names of the files on Bitsavers, I could use exactly the same name or use the names from the removed code, it's all fine by me.

cuavas: The quoted strings, e.g. `"ccb17h.bin"`, are called “labels” – MAME does use as file names when searching for assets.  We usually try to format these by putting the label/mask code/etc. from the chips (in lowercase with restricted character set) before the final dot, and the PCB location afterwards.

For example `"mon_4-000-v_2.1.a1"` was from a chip labeled “MON 4-000-v 2.1” at PCB location A1 in an Intel development system.

Do you know what the labels and PCB locations for these chips are?

eschaton: Thanks for the explanation! I can check my Concepts for the PCB locations. (I assume a socket number like U37 would be ideal while grid location would be acceptable?)

Unfortunately I don't have the 1.7 ROMs in my Concepts so I can't check them for label/mask code/etc., and I don't think anything like that was recorded for the dumps on Bitsavers. Al does that when he can, so I expect the ROMs may have been hand-labeled.

PR #14774, opened 2026-01-04
DRC: Minimise the number of pages that need protection changes

cuavas:
This changes how the misleadingly named "DRC hash table" allocates lookup table blocks.  This changes it to allocate larger blocks of memory from the "permanent allocation" area at the end of the cache, rather than allocating smaller "temporary" blocks intermixed with executable code.

This has a number of effects:
* The lookup tables are no longer marked executable while generated code is running.
* Nothing needs to write to previously allocated "temporary" cache memory, so at most one page needs to be marked writable when switching to generating code (except in the case of a cache flush, which obviously requires everything to be marked writable).
* You may be able to get a better idea of the actual amount of generated executable code from cache statistics when the megabytes of lookup tables are separate.

I doubt this will actually have much impact on performance in practice, as even changing protection on a single page requires an expensive page table flush.  But I still think it's a cleaner approach to managing memory.

### DIFF
--- a/src/devices/cpu/drcbeut.cpp
+++ b/src/devices/cpu/drcbeut.cpp
@@ -34,6 +34,8 @@ using namespace uml;
 //-------------------------------------------------
 
 drc_hash_table::drc_hash_table(drc_cache &cache, uint32_t modes, uint8_t addrbits, uint8_t ignorebits) :
+	m_next_l1_block(0),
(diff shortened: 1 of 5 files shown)


cuavas:
I've seen no objections, and I have other DRC stuff to do that will benefit from having this in place.

PR #14775, opened 2026-01-04
pippin.xml, pippin_flop.xml, pc1000.xml, etc...: Various metadata cleanups

ArcadeShadow:
pippin.xml: Removed unnecessary CHD file extension.
pippin_flop.xml: Added manufacturer in the software list name and rename country abbreviation.
pc1000.xml: Fixed parent/cloneof relationships
bbc_cass.xml: Lowercase on descriptive word
aim65_cart.xml: Fixed parent/cloneof relationships
gba.xml: rename "Jackie Chan Adventures - Legend of the Darkhand" to "Jackie Chan Adventures - Legend of the Dark Hand" and fixed parent/cloneof relationship of "Mortal Kombat Advance"

### DIFF
--- a/hash/pippin_flop.xml
+++ b/hash/pippin_flop.xml
@@ -5,7 +5,7 @@ license:CC0-1.0
 
 NOTE: This list is here only to document available dumps and it's not used (yet) by MAME!
 -->
-<softwarelist name="pippin_flop" description="Apple / Bandai Pippin disk images">
+<softwarelist name="pippin_flop" description="Apple Pippin disk images">
 
 <!-- The Apple/Bandai Pippin Atmark/@world didn't ship with a floppy drive, it was sold as an add-on -->
 
--- a/hash/pippin.xml
+++ b/hash/pippin.xml
@@ -3,7 +3,7 @@
 <!--
 license:CC0-1.0
 -->
-<softwarelist name="pippin" description="Apple / Bandai Pippin CD-ROMs">
+<softwarelist name="pippin" description="Apple Pippin CD-ROMs">
 
 <!-- Todo: Some (maybe all?) Pippin discs are compatible with macs
 


ArcadeShadow:
In the pc1000.xml software list file, "General Knowledge 2" and "Fantasy Trivia" have the same rom CRC/SHA1. 
Is it normal?

angelosa:
I have no experience with this system, but there's a pattern where some carts with identical crc/sha1 with this format:
```xml
	<software name="biblekno">
		<description>Bible Knowledge</description>
		<year>1988</year>
		<publisher>VTech</publisher>
		<part name="rom" interface="genius_cart">
			<dataarea name="rom" size="0x20000">
				<rom name="27-5410-00.ic1" size="0x20000" crc="fc1fa151" sha1="54cd1311aa9f58607836b152826e782de6bfe098" offset="0" />
			</dataarea>
		</part>
	</software>

	<software name="explor">
		<description>Explorations</description>
		<year>1991</year>
		<publisher>VTech</publisher>
		<part name="rom" interface="genius_cart">
			<dataarea name="rom" size="0x20000">
				<rom name="27-5410-00.ic1" size="0x10000" crc="fc1fa151" sha1="54cd1311aa9f58607836b152826e782de6bfe098" offset="0" />
				<rom size="0x10000" offset="0" loadflag="continue"/> <- this difference
			</dataarea>
		</part>
	</software>
```
I presume your catch is because one of the two needs the extra bank swapped, is it some kind of flippy-like for carts that really needs to be fixed in driver instead?

hash/pippin_flop.xml
```diff
@@ -5,12 +5,12 @@ license:CC0-1.0
 
 NOTE: This list is here only to document available dumps and it's not used (yet) by MAME!
 -->
-<softwarelist name="pippin_flop" description="Pippin disk images">
+<softwarelist name="pippin_flop" description="Apple / Bandai Pippin disk images">
```
cuavas: I’m not sure I’d really want to do this.  It was only distributed by Bandai in Japan.  We don’t generally put regional distributors in software list names.

cuavas:
> I presume your catch is because one of the two needs the extra bank swapped, is it some kind of flippy-like for carts that really needs to be fixed in driver instead?

In some cases where the program was small, they ordered one batch of mask ROMs with the content for both software titles, and inverted address lines in the cartridges to select between the two.  The cartridge wiring determines how the ROM gets loaded.

hash/pippin_flop.xml
ArcadeShadow: Do you suggest to put only "Apple Pippin" to both software list names (Pippin disk images and Pippin CD-ROMs)?

FoxhackDN:
> pippin.xml: Removed unnecessary CHD file extension.

... whoops, my bad, sorry.

hash/pippin_flop.xml
cuavas: Yeah, I’d leave it as just Apple if you really want to put a company name in.  Apple had the idea that the Pippin platform would be licensed to a number of companies and become a family of products, but in the end the only licensees were Mitsubishi (manufactured hardware for Bandai and Katz Media, but never sold devices themselves), Bandai (Pippin ATMARK in Japan, Pippin \@WORLD in US/Europe), and Katz Media (Katz Media Player 2000, not sold to the public directly, very rare).

ArcadeShadow: Done it. Ready to merge.

PR #14781, opened 2026-01-06
ibm/ibm5100.cpp: added layout for ibm5110 [Guru]

Osso13:


felipesanches:
Oh, wow! That's so cool!

image

felipesanches:
Here's some feedback:

* There are a few white rectangles being drawn slightly out of the key caps.
* It seems to me that the power and the reset switches do not work
* The display color can be inverted, but the default in this layout seems to be white background. I would prefer black background as a default, but I am not sure what makes more sense for this specific device. What do others here think about it?
* I haven't looked at the code yet, but I may send additional feedback later regarding code style.

felipesanches:
Oh, I think the "rectangles" are actually notdef glyphs! (glyphs not available on the font used by MAME)

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+	<!-- Knob Elements -->
+	<element name="knob"><disk></disk></element>
+	<element name="knob-position"><rect></rect></element>
+
+	<!-- Keyboard Key Component Layers -->
+	<element name="keybase">
+		<image><data><![CDATA[
+		<svg width="280" height="340" viewBox="0 0 280 340">
+			<g transform="scale(0.345 0.4048)">
+				<path d="M 52.6663 64.32523
```
cuavas, line 239: Please use readable coordinates for SVG, and don’t use things that need extra layers of transform.  Layout files are supposed to be intelligible for humans.  Same goes for all the SVG.

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+			A30,30 0 0 1 316,526
+			H30
+			A30,30 0 0 1 0,496
+			V40
+			A40,40 0 0 1 40,0
+			Z"
+			fill="#40435a" stroke="#40435a" stroke-width="1"/>
+				<line x1="72" y1="340" x2="10" y2="517" stroke="#4c4f6b" stroke-width="10"/>
+				<line x1="270" y1="340" x2="335" y2="517" stroke="#4c4f6b" stroke-width="10"/>
+		</svg>
```
cuavas, line 426: Indentation’s weird here.

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+				 A40,40 0 0 0 356,0
+				 H40
+				 A40,40 0 0 0 0,40
+				 Z"
+				 fill="#40435a" stroke="#40435a" stroke-width="1"/>
+				<line x1="152" y1="343" x2="87" y2="520" stroke="#4c4f6b" stroke-width="10"/>
+				<line x1="318" y1="343" x2="383" y2="520" stroke="#4c4f6b" stroke-width="10"/>
+				<line x1="80" y1="118" x2="8" y2="255" stroke="#4c4f6b" stroke-width="10"/>
+				<line x1="142" y1="120" x2="65" y2="267" stroke="#4c4f6b" stroke-width="10"/>
+		</svg>
```
cuavas, line 451: Here as well.  Please clean up all the indentation.

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+		]]></data></image>
+	</element>
+	<element name="keycap-grey">
+		<image state="0"><data><![CDATA[
+			<svg width="190" height="190" viewBox="0 0 195 195">
+				<rect x="0" y="0" width="190" height="190" rx="50.0" ry="50.0" fill="#717788" stroke="#717788" stroke-width="1"/>
+			</svg>
```
cuavas, line 458: All of these have a width 1 stroke that's half way outside the width/height of the image.  Also, the viewbox is larger than the image dimensions (applies to most of the SVG fragments).  Please keep the graphics inside the images and make the viewboxes make sense.

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+				<rect x="0" y="0" width="400" height="190" rx="50.0" ry="50.0" fill="#474a63" stroke="#474a63" stroke-width="1"/>
+			</svg>
+			]]></data>
+		</image>
+		<image state="1"><data><![CDATA[
+			<svg width="400" height="190" viewBox="0 0 405 195">
+				<rect x="0" y="0" width="400" height="190" rx="50.0" ry="50.0" fill="#4d506c" stroke="#4d506c" stroke-width="1"/>
+			</svg>
+			]]></data>
+		</image>
```
cuavas, line 480: Generally, if you’ve got a single shape that doesn’t change and you just want to change the colour depending on the shape, you should draw the shape in white and set the colour for each shape.  You can make this file a lot shorter by doing that.

src/mame/layout/ibm5110.lay
```diff
@@ -0,0 +1,1984 @@
...
+			A20 20 0 0 0 95.37,-3
+			H1582.63
+			A20 20 0 0 0 1602.63,-18
+			L1673,100
+			A10 10 0 0 1 1663,110
+			H10 Z"
+			fill="#232323" stroke="#555555" stroke-width="5"/>
+		</svg>
+		]]></data></image>
+	</element>
```
cuavas, line 1228: More indentation that’s all over the place.

cuavas:
> Oh, I think the "rectangles" are actually notdef glyphs! (glyphs not available on the font used by MAME)

You can use `-artwork_font` to specify a font with more coverage if you don’t get glyph substitution with your font provider.

happppp:
I understand why you're ranting, but I don't like your threats.

Messy indentation: yeah, easy to clean up.
Drawing outside SVG viewbox: I don't mind, whether or not done intentionally, it's allowed by SVG standard (FWIW the LCD screen SVGs in MAME deliberately draw the white bg to outside the viewbox to prevent aliasing issues (black boarder) when plonking it on a white rect in external artwork)
SVG is not human-readable: I don't really care, as long as you don't embed large SVGs, eg. 10KB.

White rects for Felipe: with the exception of cnchess and ngomoku, I always make sure that the UTF8 chars I use are available in the Liberation Sans font, so they work on Linux.

Opinions may vary: In general, not specifically this, does MAME need full keyboards in internal .lay files? I don't mean for non-standard things like synths, but standard PC style keyboards with minor differences to the ones we have at home. Let's say a ZX Spectrum one, on one hand it's nice to have a quick reference for the BASIC shortcuts, on the other hand just google it?

happppp:
In general, if embedded keyboard references are going to be a thing in MAME sometime in the future. Then I'd rather have them look consistent between eachother. eg. a simple abstract design that looks same for apple2, spectrum, c64, ti99, ... Just the key locations and labels.

Note that this opinion is not a blocker for this PR.

felipesanches:
@Gurudumps Please focus on objectively answering any questions you get on PRs (even when you eventually consider them absurd). Also avoid rants. Your goal here should ideally be to get your contribution merged. If it means you have to reply to what you may consider as silly question, simply do it and get your PR merged! Causing a stir won't be helpful in getting the PRs merged.

Having said that, I'd like to reiterate my support for your work. I think that what you've done here is good stuff, even though it may not be perfect according to many criteria. I still think it improves what currently exists on master branch so I hope it gets merged soon, even if additional tweaks may happen later.

PR #14783, opened 2026-01-07
lisa: move lisa_m.cpp contents to lisa.cpp

eschaton:
Move the contents of `lisa_m.cpp` into `lisa.cpp` since that was just a side-effect of having separate device/machine directories. (The exception is `lisa_model_t` which I moved into `lisa.h` and tweaked `lisa_state` to use.) No functional change intended.


galibert:
I have a massive commit here that does that among a lot of other things...

PR #14784, opened 2026-01-08
storio.cpp: use the system's real size

vtarchive:
i checked the bios in flash player and it's 480x272


mamehaze:
the resolution of the flash files does not directly imply the system output resolution

please stop making changes you're not qualified to make.  a comment would have sufficed.

PR #14786, opened 2026-01-08
Add Revolution X revision 2

clawgrip:
New working systems
-------------------
Revolution X (revision 2.0 9/8/94) [Victor Fernandez (City Game)]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -28821,6 +28821,7 @@ wwfmaniap
 
 @source:midway/midxunit.cpp
 revx
+revx1
 revxp5
 
 @source:midway/midyunit.cpp


mamehaze:
FWIW the comments Clawgrip added here are from me.

From what I can see RevX does *need* the main CPU ROMs in both regions, but there's probably a better way to do it than loading the ROMs again when they're on the PCB only once.  The region endianess might be a problem for ROM_COPY though (not sure) but even if it is, it could be done in an init.  This is an existing issue in the code, so shouldn't get in the way of the PR, but did need highlighting, because it's a bit confusing to see.

For the sound program, it really looks like the board was upgraded from an earlier/proto board at some point, and the sound ROM was mistakenly not replaced, as the main CPU code here explicitly looks for version 2 or higher, and the sound ROM present reports version 1, which will give an error on startup (skippable, but clearly not what is expected)  Using the V2 sound ROM makes a lot more sense (nothing has been lost, the proto set in MAME uses the V1)

PR #14788, opened 2026-01-09
majrjh （majrjhdx）助力Correction, Supplement, and Improvement it.

Gently28:
Added Correction and Supplement  input ports for "majrjh"，Also applies to “majrjhdx”.if i wrong,i hope someone to do it that based on the following image !


cuavas:
Do you have a scan of the actual manual for this that we can see?

Gently28:
> 你有我们能看到的实际说明书扫描件吗？




 
SW4-10 writed OFF is color 【カラー】，fact was Green【緑】!

cuavas:
Thanks, I filled in the settings that can be seen in those pages in 01bce4b5b731c66bd327557446ad55b2becfa74a.  That confirms half the DIP switch settings, and all but three of the other DIP switch settings could be identified easily, following usual Dynax conventions.

PR #14792, opened 2026-01-10
namcops2.cpp: Document offline English version of Battle Gear 3

FMecha:
Dongle ID `B3900068A` [(in this PR)](https://github.com/mamedev/mame/pull/12093) was originally thought to be offline English version, but turned out to be the Betson US English one (region `A` with Honda disclaimers). mokonaXVI recently pointed out about [this post](https://www.arcade-projects.com/threads/what-kind-of-wires-i-o-boards-are-needed-to-run-system-246-racing-games.904/#post-12531) which has a different dongle ID (`B3900071A`) for Battle Gear 3, which is likely to be [the offline English version](https://youtu.be/d_ibssSC4Fs) (region `E`).

Unfortunately, the post does not include HDD ID, but much like the US version required editing for different server and Honda-related disclaimers, I believe the offline one has a different HDD ID as well to replace online support with local leaderboards.


FMecha:
Per [Undumped Wiki:](https://undumped.miraheze.org/wiki/Battle_Gear_3_(Export))

>According to Guru, "Screen shows version 2.00E but using 2.03J HDD".

Probably same HDD as "J" (Japanese)/"O" (English with NESYS support, known to have operated in [Hong Kong](https://web.archive.org/web/20030416095745/http://www.bg3.org/cgi-bin/bg3/index.cgi?pg=loc) and [Malaysia](https://web.archive.org/web/20040422202631/http://www.angeltowns.com/members/mrgc/frames.html)) regions, then? Not sure if it will work with V2.04J HDD.

PR #14797, opened 2026-01-11
Fixing the overload order for the LUA wrapper for `options.entries[...]:value(...)`

npwoods:
Overloads specified with `sol::overload()` are evaluated in sequential order.  Prior to this change, the overload for `float` was before the overload for `int`.  This means that in practice, the overload for `float` was always called even if the value and the option in question was an integer.  Because the wrappers error if the types do not match, this would make `option_type::INTEGER` entries unusable.

This change flips the order and makes the overload for `int` work with both `option_type::INTEGER` and `option_type::FLOAT`.

### DIFF
--- a/src/frontend/mame/luaengine.cpp
+++ b/src/frontend/mame/luaengine.cpp
@@ -1357,17 +1357,17 @@ void lua_engine::initialize()
 			else
 				e.set_value(val ? "1" : "0", OPTION_PRIORITY_CMDLINE);
 		},
+		[this](core_options::entry &e, int val) {
+			if(e.type() != core_options::option_type::INTEGER && e.type() != core_options::option_type::FLOAT)
+				luaL_error(m_lua_state, "Cannot set option to wrong type");
+			else
+				e.set_value(string_format("%d", val), OPTION_PRIORITY_CMDLINE);
+		},
 		[this](core_options::entry &e, float val) {
 			if(e.type() != core_options::option_type::FLOAT)
 				luaL_error(m_lua_state, "Cannot set option to wrong type");
 			else
 				e.set_value(string_format("%f", val), OPTION_PRIORITY_CMDLINE);
-		},
-		[this](core_options::entry &e, int val) {
-			if(e.type() != core_options::option_type::INTEGER)
-				luaL_error(m_lua_state, "Cannot set option to wrong type");
-			else
-				e.set_value(string_format("%d", val), OPTION_PRIORITY_CMDLINE);
 		},
 		[this](core_options::entry &e, const char *val) {
 			if(e.type() != core_options::option_type::STRING && e.type() != core_options::option_type::PATH && e.type() != core_options::option_type::MULTIPATH)


cuavas:
Should it really be giving an error when setting a float option to an integer value?  That seems to violate the principle of least surprise.

cuavas:
Also that whole thing shouldn't be using capturing lambdas

cuavas:
e25ef95c4628fcf2c712e82d7ba4475a4a6faa89 cleans it up a bit.  Note that there’s nasty code duplication here.  The proper way to get integer and float option values is with `core_options::int_value` and `core_options::float_value` but these do the option lookup as well as the parsing.  There isn’t currently a way to do the parsing using the same code when you have an option object.

cuavas:
Also, Lua errors aren’t being propagated to the console properly – they always cause a panic.  That’s a bug in itself.

npwoods:
The problem is that LUA is a very weakly typed language, and goes through great lengths to treat "numbers" the same.

The logic for handling numbers with `sol::override` seems to be "_Can I change the represent the number as the specified type without any loss of information?_".  This is what prompted this change; all `int` values can be represented as a `float` and thus if the `float` handler is first, it becomes impossible to write code that manipulates options of type `option_type::INTEGER`.  However, because some `float` values can be converted to `int` without loss of information, so the order cannot be flipped without introducing the reverse problem.

I don't like this aspect of `sol` but we have to take what we're given.

cuavas:
The sol overload mechanism is complicated because it can’t read your mind.  You need to be careful with it if you want optional parameters, etc. as well.  That’s just life.

The fact that if `foo` is an integer option, `foo:value(0)` raised an error was clearly a bug.

My point though is that assuming `foo` is a float option, should `foo:value(0)` (as opposed to `foo:value(0.0)`) raise an error?  The code as written seems to be designed on the assumption that the answer is “yes”.  However I feel that violates the principal of least surprise.

npwoods:
With your change, `foo:value(0.0)` also raises an error.  This is what I'm trying to avoid.

Here is some LUA code that demonstrates this:
```
local exports = {
	name = "dummy",
	version = "0.0.1"}

local dummy = exports

function dummy.startplugin()
	-- please ignore the specific options being set; there are better ways to do each of those; the
	-- purpose of this is to demonstrate the problem of option setting overloads and LUA
	emu.register_start(function()
		manager.options.entries["speed"]:value(20.0) -- OPTION_SPEED is option_type::FLOAT, but because the `int` handler is first, this triggers an error
	end)
	emu.register_start(function()
		manager.options.entries["frameskip"]:value(5) -- this used to error though, but doesn't anymore
	end)
end

return exports
```

cuavas:
No it doesn't.  I tested it and it accepts it.

```
[MAME]> print(opts.entries['frameskip']:value())
0
[MAME]> print(opts.entries['beam_dot_size']:value())
1.0
[MAME]> print(opts.entries['readconfig']:value())
true
[MAME]> print(opts.entries['beam_dot_size']:value(0.5))

[MAME]> print(opts.entries['beam_dot_size']:value())
0.5
[MAME]> print(opts.entries['beam_dot_size']:value(1.0))

[MAME]> print(opts.entries['beam_dot_size']:value())
1.0
[MAME]> print(opts.entries['frameskip']:value())
0
[MAME]> print(opts.entries['frameskip']:value(1))

[MAME]> print(opts.entries['frameskip']:value())
1
[MAME]> print(opts.entries['frameskip']:value(0))

[MAME]> print(opts.entries['mouseprovider']:value())
auto
```

Anyway, I found *another* bug in the options code itself that I have to fix while I’m at it.

npwoods:
This is what I get with a fresh clean MAME build.

It isn't clear to me how to reconcile this with your findings.  I'm curious if anybody else can reproduce my findings @Lord-Nightmare @rb6502 @tlindner 
image

PR #14802, opened 2026-01-13
nintendo/snesb.cpp: promote piratdwb to working

devinacker:
There's still some unknown I/O here and there but the game seems to coin up and play perfectly fine now.

Machines promoted to working
----------
The Pirates of Dark Water (SNES bootleg) [Devin Acker]

### DIFF
--- a/src/mame/nintendo/snesb.cpp
+++ b/src/mame/nintendo/snesb.cpp
@@ -32,6 +32,7 @@ TODO:
  - denseib,2: fix gfx glitches, missing texts
  - venom    : gfx glitches on second level
  - wldgunsb : remove hack for continue counter (values at 0x781010 and 0x781012 are expected to be initialized on reset/boot)
+ - piratdwb : is coinage supposed to be configurable?
 
 ***************************************************************************
 
@@ -360,8 +361,7 @@ INPUT_CHANGED_MEMBER(snesb_state::piratdwb_coin_w)
 {
 	// 68k program expects two NMIs per coin
 	// (TODO: is this really supposed to be 2 coins/credit then?)
-	if (oldval != newval)
-		m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 1);
+	m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 1);
 }
 
 
@@ -502,7 +502,7 @@ void snesb_state::piratdwb_68k_map(address_map &map)
 	map(0x100000, 0x100001).portr("COIN");
 //	map(0x140000, 0x140000).w // ?
 	map(0x180000, 0x180000).lw8(
-		NAME([=](offs_t, u8) { m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 0); })
+		NAME([this](offs_t, u8) { m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 0); })
 	);
 }
 
@@ -1058,7 +1058,7 @@ static INPUT_PORTS_START( piratdwb )
 	PORT_DIPUNUSED( 0x01, IP_ACTIVE_LOW ) //
 	PORT_DIPUNUSED( 0x02, IP_ACTIVE_LOW ) // supposed to be for coinage (table at 83F69F), but not actually used
 	PORT_DIPUNUSED( 0x04, IP_ACTIVE_LOW ) //
-	PORT_DIPNAME( 0x38, 0x38, DEF_STR( Difficulty ) )
+	PORT_DIPNAME( 0x38, 0x30, DEF_STR( Difficulty ) )
 	PORT_DIPSETTING(    0x38, "1 (Easiest)" )
 	PORT_DIPSETTING(    0x30, "2" ) // "Normal" (default) on SNES
 	PORT_DIPSETTING(    0x28, "3" ) // "Hard" on SNES
@@ -1198,7 +1198,7 @@ void snesb_state::piratdwb(machine_config &config)
 	base(config);
 	m_maincpu->set_addrmap(AS_PROGRAM, &snesb_state::piratdwb_map);
 
-	GENERIC_LATCH_8(config, "latch");
+	GENERIC_LATCH_8(config, "latch").set_separate_acknowledge(true);
 
 	M68000(config, m_68k, 16_MHz_XTAL / 2);
 	m_68k->set_addrmap(AS_PROGRAM, &snesb_state::piratdwb_68k_map);


src/mame/nintendo/snesb.cpp
```diff
@@ -1029,11 +1044,37 @@ static INPUT_PORTS_START( tmntmwb )
...
+	PORT_DIPUNUSED( 0x04, IP_ACTIVE_LOW ) //
+	PORT_DIPNAME( 0x38, 0x38, DEF_STR( Difficulty ) )
+	PORT_DIPSETTING(    0x38, "1 (Easiest)" )
+	PORT_DIPSETTING(    0x30, "2" ) // "Normal" (default) on SNES
+	PORT_DIPSETTING(    0x28, "3" ) // "Hard" on SNES
+	PORT_DIPSETTING(    0x20, "4" )
+	PORT_DIPSETTING(    0x18, "5" )
+	PORT_DIPSETTING(    0x10, "6" )
+	PORT_DIPSETTING(    0x08, "7" )
+	PORT_DIPSETTING(    0x00, "8 (Hardest)" )
```
angelosa: I guess this would go at least 2 if not 3 by default.

src/mame/nintendo/snesb.cpp
```diff
@@ -1029,11 +1044,37 @@ static INPUT_PORTS_START( tmntmwb )
...
+	PORT_DIPNAME( 0xc0, 0x00, DEF_STR( Lives ) )
+	PORT_DIPSETTING(    0xc0, "1" )
+	PORT_DIPSETTING(    0x80, "2" )
+	PORT_DIPSETTING(    0x40, "3" )
+	PORT_DIPSETTING(    0x00, "4" )
+
 	PORT_START("COIN")
```
angelosa, line 1076: 4 lives are too much by Arcade standards, probably 3?

src/mame/nintendo/snesb.cpp
```diff
@@ -347,6 +354,17 @@ uint8_t snesb_state::wldgunsb_72443a_r() // in-game
...
+// Pirates of Dark Water
+
+INPUT_CHANGED_MEMBER(snesb_state::piratdwb_coin_w)
+{
+	// 68k program expects two NMIs per coin
+	// (TODO: is this really supposed to be 2 coins/credit then?)
+	if (oldval != newval)
+		m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 1);
+}
+
```
angelosa, line 366: Is it possible to control the timing so to give different coinage settings?

src/mame/nintendo/snesb.cpp
```diff
@@ -1909,4 +1947,4 @@ GAME( 1996, legendsb,     0,        extrainp,     legendsb, snesb_state, init_le
...
 GAME( 1997, venom,        0,        venom,        venom,    snesb_state, init_venom,     ROT0, "bootleg",         "Venom & Spider-Man - Separation Anxiety (SNES bootleg)",        MACHINE_IMPERFECT_SOUND | MACHINE_IMPERFECT_GRAPHICS )
 GAME( 1996, wldgunsb,     0,        wldgunsb,     wldgunsb, snesb_state, init_wldgunsb,  ROT0, "bootleg",         "Wild Guns (SNES bootleg)",                                      MACHINE_IMPERFECT_SOUND | MACHINE_IMPERFECT_GRAPHICS ) // based off Japanese version
-GAME( 199?, piratdwb,     0,        piratdwb,     piratdwb, snesb_state, init_anheizs,   ROT0, "bootleg (Conny)", "The Pirates of Dark Water (SNES bootleg)",                      MACHINE_NOT_WORKING | MACHINE_IMPERFECT_SOUND | MACHINE_IMPERFECT_GRAPHICS )
+GAME( 199?, piratdwb,     0,        piratdwb,     piratdwb, snesb_state, init_piratdwb,  ROT0, "bootleg (Conny)", "The Pirates of Dark Water (SNES bootleg)",                      MACHINE_IMPERFECT_SOUND | MACHINE_IMPERFECT_GRAPHICS )
```
angelosa, line 1950: If the coinage stuff isn't resolved this still deserves `MACHINE_UNEMULATED_PROTECTION`

Hammy1986:
Excellent! Great to see this oddball setup is up and running!
The 68000 on there made it looks like one of their MD bootlegs, to see SNES code was a surprise :D

About the Conny MD bootlegs, i'm pretty sure they are missing an IRQ, if one of the "unused" ones is forced then coin inputs work.

src/mame/nintendo/snesb.cpp
```diff
@@ -347,6 +354,17 @@ uint8_t snesb_state::wldgunsb_72443a_r() // in-game
...
 
+// Pirates of Dark Water
+
+INPUT_CHANGED_MEMBER(snesb_state::piratdwb_coin_w)
+{
+	// 68k program expects two NMIs per coin
+	// (TODO: is this really supposed to be 2 coins/credit then?)
+	if (oldval != newval)
+		m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 1);
+}
```
cuavas, line 365: These are only called when the value changes, hence the name.

src/mame/nintendo/snesb.cpp
```diff
@@ -473,22 +497,13 @@ void snesb_state::piratdwb_68k_map(address_map &map)
...
-//      NAME([this] (offs_t offset, u8 data) { m_maincpu->space(AS_PROGRAM).write_byte(0x7ffff?, data); })
-//  );
-//  map(0x180000, 0x180000).lw8(
-//      NAME([this] (offs_t offset, u8 data) { m_maincpu->space(AS_PROGRAM).write_byte(0x7ffff?, data); })
-//  );
-
+//	map(0x140000, 0x140000).w // ?
+	map(0x180000, 0x180000).lw8(
+		NAME([=](offs_t, u8) { m_68k->set_input_line(INPUT_LINE_NMI, m_prot_nmi = 0); })
+	);
```
cuavas, line 506: Please don’t use auto-capture in lambdas.  Explicitly capture everything you need.

src/mame/nintendo/snesb.cpp
```diff
@@ -464,6 +482,12 @@ void snesb_state::tmntmwb_map(address_map &map)
 void snesb_state::piratdwb_map(address_map &map)
 {
 	snesb_map(map);
+
+	map(0xc00000, 0xc00000).r("latch", FUNC(generic_latch_8_device::read));
+	map(0xc00001, 0xc00001).w("latch", FUNC(generic_latch_8_device::acknowledge_w));
```
cuavas, line 487: The way you’ve got it now, reading the latch will acknowledge it as well as writing to the acknowledge address.  Is that actually what you want, or did you mean to call `set_separate_acknowledge(true)` in machine configuration so reading it doesn’t implicitly acknowledge?

src/mame/nintendo/snesb.cpp
devinacker: 4 is the default (really the only) option in the SNES original, but I can change it to something else if you think it makes sense.

src/mame/nintendo/snesb.cpp
devinacker: Not really - the 68k NMI handler is just set up to add a credit on every second interrupt regardless of timing. I'm starting to suspect it's supposed to be two coins per credit like the comment suggests, but the game doesn't really make it clear at all.

devinacker:
```
ROM:00000500 NMI:
ROM:00000500                 movem.l d0-d5/a0-a6,-(sp)
ROM:00000504                 move.w  #$280,d0
ROM:00000508
ROM:00000508 loc_508:                                ; CODE XREF: NMI+10↓j
ROM:00000508                 bsr.s   sub_540
ROM:0000050A                 nop
ROM:0000050C                 subi.w  #1,d0
ROM:00000510                 bne.s   loc_508
ROM:00000512                 cmpi.b  #0,d6
ROM:00000516                 beq.s   loc_51E         ; first NMI
ROM:00000518                 bsr.s   AddCredit       ; second NMI
ROM:0000051A                 clr.l   d6
ROM:0000051C                 bra.s   loc_522
ROM:0000051E ; ---------------------------------------------------------------------------
ROM:0000051E
ROM:0000051E loc_51E:                                ; CODE XREF: NMI+16↑j
ROM:0000051E                 move.b  #$FF,d6         ; first NMI
ROM:00000522
ROM:00000522 loc_522:                                ; CODE XREF: NMI+1C↑j
ROM:00000522                 move.b  #0,($180000).l
ROM:0000052A                 movem.l (sp)+,d0-d5/a0-a6
ROM:0000052E                 rte
```

devinacker:
@Hammy1986 Do you happen to have any photos of the board?

src/mame/nintendo/snesb.cpp
devinacker: Right, it (seemingly) should acknowledge on writes only.

src/mame/nintendo/snesb.cpp
devinacker: I'm not sure if the coinage thing is related to protection at all, though. The 68k-side code is extremely small and always adds 1 credit per two NMIs (specifically, the 68k-to-SNES latch is written with the number of credits to add, but it's always 1). The SNES side has a coinage lookup table like the other bootlegs, but the value is just written to RAM and never referenced again. Feels more like just a sloppy/unfinished hack to me.

Hammy1986:
IMG_20250908_085021 (2)

Here's what i have on hand, the board will be with me at some point.

src/mame/nintendo/snesb.cpp
mamehaze: maybe it's meant to generate 1 NMI when the input is triggered, and one when it's released (although that would be 2 NMIs in very quick succession)

would be a weird way to do things, but using a 68000 on a SNES bootleg just to drive coinage is a pretty weird thing to do in the first place.

cuavas: That’s what his code currently does…

devinacker: Yeah, I figured it was probably the most "player-friendly" way to handle it without knowing how the original PCB behaves, since the game itself doesn't really provide much feedback otherwise.

mamehaze:
Seems logical. 2 coin 1 credit as a default is more common on Spanish/euro
games, not really Asian bootlegs, especially not if it's hardcoded, so from
a point of origin perspective I'm favouring the current solution

PR #14804, opened 2026-01-13
misc/3do.cpp: split Madam and Clio into own devices

angelosa:


src/mame/misc/3do_madam.cpp
```diff
@@ -0,0 +1,326 @@
...
+	// bit 23 Internal ARM
+	// bit 15-8 manufacturer ID (0x02: MEC -> Panasonic, 0x09: Sanyo, 0x0b: Goldstar)
+	// bit 0 wirewrap system
+	m_revision = 0x01020200;
+	m_msysbits = 0x51;
+
+	save_pointer(NAME(m_pip), 16);
+	save_pointer(NAME(m_fence), 4);
+	save_pointer(NAME(m_mmu), 64);
+	save_pointer(NAME(m_mult), 40);
```
cuavas, line 51: Why are you using `save_pointer` with determinate arrays?

angelosa: Saving for debugger aid, none of this stuff is actually tested yet (and [mult isn't exactly memory mapped](https://3dodev.com/documentation/hardware/opera/madam/matrix_engine)). Unless you mean using `sizeof` for determining the length.

cuavas: My point is, they aren’t indeterminate arrays.  `save_pointer` is error-prone in that case because size mismatches can make it blow up.

angelosa: Oh, wasn't aware that `save_item` supports C sized arrays, for some reason I thought they needed `save_pointer`.

PR #14809, opened 2026-01-14
gbcolor: add South Park prototype ROM

chungy:
ROM from here: https://archive.org/details/spgam120

Might make sense as a child of Mary-Kate & Ashley and/or Maya the Bee, but those games already don't have a parent/child relationship. Since this was first developed with the South Park IP, maybe leave it as its own thing.

New working software list items (gbcolor.xml)
---------------------------------------------
South Park (prototype) [DaKoolDood]

### DIFF
--- a/hash/gbcolor.xml
+++ b/hash/gbcolor.xml
@@ -20402,6 +20402,21 @@ license:CC0-1.0
 		</part>
 	</software>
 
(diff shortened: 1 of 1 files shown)


mamehaze:
It was intended to be a unique release, even if it did get reworked into something else later, so I'd leave it as its own thing

PR #14810, opened 2026-01-15
Update atarisy1.cpp - set reliefs1 to use atarisy1r_state so that controls work

mamehaze:


mamehaze:
https://github.com/mamedev/mame/commit/cdd8da5ca1e772425c62d88fc81308d6814965eb the original note when atarisy1r_state was added stated that LSI CART 2 hardware doesn't have the IRQ3 circuits, but Phil B provided images showing otherwise, and the game here, which is an LSI CART 2 relies on it.

mamehaze:
lsicart2 copy

PR #14812, opened 2026-01-15
Fixed inputs for Relief Pitcher (System 1)

synamaxmusic:
The joysticks were swapped and the buttons now have proper mapping.  Confirmed with the owner that the sound ROMs report an error message in service mode but this is a false positive since the checksums match up with the PCB photo.


src/mame/atari/atarisy1.cpp
```diff
@@ -611,10 +611,10 @@ static INPUT_PORTS_START( reliefs1 )
...
 
 	PORT_START("F60000")    // F60000
-	PORT_BIT( 0x0001, IP_ACTIVE_LOW, IPT_BUTTON1 )
-	PORT_BIT( 0x0002, IP_ACTIVE_LOW, IPT_BUTTON2 )
-	PORT_BIT( 0x0004, IP_ACTIVE_LOW, IPT_BUTTON3 )
-	PORT_BIT( 0x0008, IP_ACTIVE_LOW, IPT_BUTTON4 )
+	PORT_BIT( 0x0001, IP_ACTIVE_LOW, IPT_START1 )
+	PORT_BIT( 0x0002, IP_ACTIVE_LOW, IPT_START2 )
+	PORT_BIT( 0x0004, IP_ACTIVE_LOW, IPT_BUTTON1 ) PORT_NAME("P1 Action")
+	PORT_BIT( 0x0008, IP_ACTIVE_LOW, IPT_BUTTON2 ) PORT_NAME("P2 Action")
```
cuavas: Are these supposed to be player 1 and player 2 buttons?  If so, it should just be:
```c++
	PORT_BIT( 0x0004, IP_ACTIVE_LOW, IPT_BUTTON1 ) PORT_PLAYER(1)
	PORT_BIT( 0x0008, IP_ACTIVE_LOW, IPT_BUTTON1 ) PORT_PLAYER(2)
```

synamaxmusic: The buttons are labeled as "Action" on the control panel, so I followed what was done for peterpak and indytemp.
image

cuavas: This is a straightforward stick plus one button, though.  It fits into MAME’s framework for conventional arcade controls better.

PR #14814, opened 2026-01-15
Add speaker to GRiD Compass

vklachkov:


src/mame/gridcomp/gridcomp.cpp
```diff
@@ -612,9 +642,9 @@ ROM_END
 ***************************************************************************/
 
 //    YEAR  NAME      PARENT    COMPAT  MACHINE   INPUT     CLASS           INIT        COMPANY           FULLNAME           FLAGS
-COMP( 1982, grid1101, 0,        0,      grid1101, gridcomp, gridcomp_state, empty_init, "GRiD Computers", "Compass 1101",    MACHINE_NO_SOUND_HW | MACHINE_IMPERFECT_CONTROLS )
```
rb6502: With the volume DAC not supported, MACHINE_IMPERFECT_SOUND should also be included.

vklachkov: I would like to add volume control support, but I don't see MAME devices being able to set their own volume level. If you know, could you point me in the right direction on where to look?

happppp: mysounddevice->set_output_gain(channel number or ALL_OUTPUTS, volume in 0.0f to 1.0f range);
(if it's a user analog volume knob, then don't bother, volume control is already in MAME's UI itself)

startaq: How is it handled in hardware? A simple example can be found here: https://github.com/mamedev/mame/blob/85b1ac1b8e0ea62885491b8fd7979620013a54f4/src/mame/dataeast/deco156.cpp#L110

vklachkov: The laptop has two 8-bit DAC0832s. One is for sound, the other for volume control. I wanted to find the function where I could pass a normalized volume value. Sorry, I somehow missed the `set_output_gain` function. It seems that's what I need.

vklachkov: Thank you all for your help! I'll make a patch a bit later. I need to write some test program for this. I don't know of any program with volume control)

rb6502: Depending how the hardware works, it could be as simple as multiplying the two DACs together and using the result as a 16-bit sample.

vklachkov: Decided not to implement volume control for now. Added MACHINE_IMPERFECT_SOUND.

vklachkov:
Ready for review. Such sound is better than no sound.

happppp:
For an 8-bit DAC I'd recommend this: https://github.com/mamedev/mame/blob/master/src/devices/sound/dac.h
(use DAC_8BIT_R2R)

Oh, and no need for a trampoline, instead like: ```w("dac", FUNC(dac_8bit_r2r_device::write));```

vklachkov:
@happppp thanks for the advice! Indeed, MAME already had the DAC0832. I replaced my own implementation with the ready-made component. It works, and the code is cleaner and better now.

vklachkov:
Rebased. If there are no comments, could you please merge this PR?

src/mame/gridcomp/gridcomp.cpp
```diff
@@ -231,6 +234,20 @@ void gridcomp_state::grid_modem_w(offs_t offset, uint8_t data)
...
 }
 
+void gridcomp_state::grid_sound_w(offs_t offset, uint8_t data)
+{
+	if (offset & 0b01) {
+		LOG("VOLUME DAC <- %02x\n", data);
+		// Not supported.
+	}
+
+	// GRiD makes sounds when writing to dfe4:2 or dfe4:4.
```
galibert, line 244: does it make sound when writing to 6 or 7?

vklachkov: I don't know. Only two programs produce sound on the Compass. I don't have the modem/sound board on hand to check.

galibert: are there schematics?

vklachkov: No. @JDat started drawing, but never finished.

PR #14815, opened 2026-01-15
add ym2203 support for magical tonic II

flama12333:
can confirm that tonic II uses ym2203 which is opn. xtal is not verified. not hooked up.


Osso13:
this is now superceded by 483cc38e72ccac426d2cc22d288e1570298573d6

PR #14819, opened 2026-01-16
Add a Non Working Driver For  Feitalin Game Square

flama12333:
firmware stuck24

dumped by someone
the microcontroller and eeprom such adpcm etc are dumped. sadly the firmware refuses to start due to timer? stuck at ljmp 378.
hardware info wip


stonedDiscord:
The file is missing the .cpp extension and the mame.lst entry is also missing.

flama12333:
> The file is missing the .cpp extension and the mame.lst entry is also missing.

fixed.

PR #14823, opened 2026-01-17
Taito System L driver files consolidation into single file.

TheLostNinja:
I think, **taito_l_v.cpp** and **taito.h** is too small for an independent sub-driver status.


JimCarlTay:
This PR can now be closed since Osso13 has made these same changes.

PR #14826, opened 2026-01-18
sequential/prophet5.cpp: Incorporated the CEM3320 filter.

m1macrophage:
sound/va_vcf.cpp:
* va_vcf_device: Added functions for accessing frequency and resonance gain.
* cem3320_lpf4_device: Removed feedback resistor from the constructor. Most designs use the datasheet recommendation (100K).
* cem3320_lpf4_device: Input stream can be a current, or a voltage by using a resistor. The first configuration is used by the prophet5, while the second one is used by the linndrum.

linn/linndrum.cpp:
* Adapted to the interface changes above.

sequential/prophet5.cpp:
* Incorporated CEM3320, and emulated a subset of its frequency control and trimmers.
* Created and used enums for CV indices.
* Renamed some inputs for better UI organization.

EDIT: Newly working controls: Cutoff and Resonance knobs, Keyboard (tracking) button.

------

Next up: I will look into a proper filter implementation, as suggested [here](https://github.com/mamedev/mame/pull/14680?notification_referrer_id=NT_kwHOChHyK9oAJFJlcG9zaXRvcnk7MTQzMDMwNDg7SXNzdWU7MzczMzYyMzE0Nw#issuecomment-3687433354)

### DIFF
--- a/src/mame/sequential/prophet5.cpp
+++ b/src/mame/sequential/prophet5.cpp
@@ -36,7 +36,8 @@ and adc_r() for details.
 TODO: Outline of voice architecture.
 
 This driver is based on the Prophet 5 Rev 3.0 technical manual, and is intended
-as an education tool.
+as an education tool. To that end, names for variables, I/O handlers, enums,
+etc., generally match signal names in the schematics.
 
 There is no audio. Running with `-oslog -output console` will display CVs, and
 voice and LED control signals.
@@ -55,10 +56,10 @@ illuminated). The synth is unresponsive while this is happening.
 #include "machine/timer.h"
 #include "sound/dac.h"
 #include "sound/flt_rc.h"
-#include "sound/flt_vol.h"
 #include "sound/mixer.h"
 #include "sound/mm5837.h"
 #include "sound/va_vca.h"
+#include "sound/va_vcf.h"
 #include "video/pwm.h"
 #include "speaker.h"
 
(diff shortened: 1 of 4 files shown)


src/mame/sequential/prophet5.cpp
```diff
@@ -1479,23 +1638,48 @@ INPUT_PORTS_START(prophet5)
...
+	PORT_START("cv_in_filter")  // J6, filter cutoff CV input.
+	// An input of 0V has the same effect as the input not being connected.
+	PORT_ADJUSTER(0, "CV IN: FILTER")
+		PORT_CHANGED_MEMBER(AUDIO_TAG, FUNC(prophet5_audio_device::filter_cv_in_changed), 0)
+
+
+	static std::string voice_tag[5];
 	static std::string volume_name[5];
+	static std::string filt_scale_name[5];
+	static std::string filt_offset_name[5];
```
cuavas, line 1660: This isn’t safe at all.  Remember MAME goes multi-threaded for various reasons (e.g. for `-validate` and when caching stuff for the internal UI).  Static `std::string` like this isn’t thread safe, and since it’s going to reassign the strings every time through, you have no idea whether they’ll remain valid.

m1macrophage: Ah, thanks, I hadn't realized this code path could be used by multiple threads.

When I originally worked on this, I tried defining the per-voice input ports in their own `INPUT_PORTS_START`, and having a `prophet5_voice_device::device_input_ports() { return INPUT_PORTS_NAME(...); }`. That felt cleaner, but the ports across the voice devices had the same name, which didn't work well for the UI.

Thinking about this now, duplicate names could be avoided by doing something like:
```
INPUT_PORT_START(prophet5_voice_io)
{
    prophet5_voice_device &voice = dynamic_cast<prophet5_voice_device &>(owner);
    PORT_START(...)
    PORT_ADJUSTER(..., voice.get_volume_trimmer_name()) 
    // get_volume_trimmer_name() returns a name parameterized on the voice number
   ...
}
```

Would that be acceptable?

An other possible fix would be for `ioport_field::ioport_field()` to copy the port name instead of storing a pointer to it. But I assume there is a reason for the current setup.

Let me know if there is some other established solution to this problem, or if one of the above options works.

m1macrophage: I sent a fix in #14895.

PR #14832, opened 2026-01-20
hash/jazz.xml: new Windows NT hard disk images

chungy:
For Windows NT 3.1, 3.51, and 4.0, these disk images were created by using the installation discs defined in generic_cdrom.xml.

All of these images have the following properties:
 * Half-gigabyte hard disk image
 * Single FAT partition for both ARC osloader and operating system
 * Installation date and time reflect those found on the installation CDs.

Like my FreeDOS 1.3 and 1.4 hard disks defined in ibm5170_hdd.xml, each of these was also created while recording an INP and can be recreated from these input recordings. They are located at <https://chiselapp.com/user/chungy/repository/mame-reprod-chd/dir?ci=c15788dc4dd087c9&name=jazz-nt>

Windows NT 3.5 has consistently failed to install from a blank hard disk, only producing a BSOD at the end of text mode setup.  Not just from the installation image provided by generic_cdrom, but I tried several variants from MSDN disks, both debug and release builds, and the same behavior across all of them.  It works as an upgrade from NT 3.1, but I am not willing to depend on that for this sort of image. There is an existing winnt35 image dating to 2020, but I do not know how Patrick Mackinlay first created this image (possible MAME regression?).

### DIFF
--- a/hash/jazz.xml
+++ b/hash/jazz.xml
@@ -6,15 +6,15 @@ license:CC0-1.0
 <softwarelist name="jazz" description="Jazz software">
 
 	<software name="winnt31">
-	    <description>Windows NT 3.1</description>
-	    <year>1994</year>
-	    <publisher>Microsoft</publisher>
+		<description>Windows NT 3.1</description>
+		<year>1994</year>
+		<publisher>Microsoft</publisher>
 
-	    <part name="hdd" interface="scsi_hdd">
-	        <diskarea name="harddriv">
-	            <disk name="winnt31" sha1="efaf73e45eca35e45f447d41d12bbd091c298d8e" writeable="yes" />
-	        </diskarea>
-	    </part>
+		<part name="hdd" interface="scsi_hdd">
+			<diskarea name="harddriv">
+				<disk name="jazz_winnt31" sha1="efaf73e45eca35e45f447d41d12bbd091c298d8e" writeable="yes" />
+			</diskarea>
+		</part>
 	</software>
 
 	<software name="winnt35">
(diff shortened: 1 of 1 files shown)


chungy:
One other amendment to the properties the images have: the system was shut off after the GUI-mode setup was finished. The first boot of the operating system (in Event Viewer) will always reflect the time that you start the VM. :-)

It can be like you waited 26 years after installing NT before actually booting it!

chungy:
May as well tag @pmackinlay here. If you happen to know how you made a Windows NT 3.5 image almost six years ago, I'd be interested. :-) (If you don't mind, posting on my [fossil forum](https://chiselapp.com/user/chungy/repository/mame-reprod-chd/forumpost/2026-01-19T04:14:52z) would be my preference, but here is fine too.)

cuavas:
@pmackinlay can you look at this?  You’re the resident expert on these systems.

pmackinlay:
This generally looks fine, but I don't have time to test; I'm fine to merge on the assumption they have been tested.

I also don't recall how I got 3.5 to install, it's been too long.

cuavas:
The thing with CHD diff name conflicts is a known issue.  I haven’t decided exactly what to do about it yet.

chungy:
> The thing with CHD diff name conflicts is a known issue. I haven’t decided exactly what to do about it yet.

I immediately think that frontends (which could be as simple as a bash function...) can set `-diff_directory` per-driver, but that's just a bandage. I have a belief that MAME shouldn't need frontends for basic usage.

Still, it's a problem not connected to this PR. :-)

PR #14833, opened 2026-01-20
sega/model2.cpp: delay IRQ mask update by feeding new value to timer callback

gm-matthew:
This is basically an update to commit 456c228869283dcca16e17d9450d76e75c505009; instead of updating the IRQ mask register straight away, it feeds the new value to the timer callback as suggested by @cuavas. There should no longer be any possibility of `sound_ready_w()` bypassing the 2 cycle delay and firing a sound interrupt too early.

### DIFF
--- a/src/mame/sega/model2.cpp
+++ b/src/mame/sega/model2.cpp
@@ -22,8 +22,10 @@
               bypass it by entering then exiting service mode;
     - sgt24h: has input analog issues, steering doesn't center when neutral,
       gas and brake pedals pulses instead of being fixed;
-    - vcop2: stage select has tilemap priority issue, tilemap B (city model) has priority bit set,
-             yet it should appear underneath tilemap A ("shoot to select") which does not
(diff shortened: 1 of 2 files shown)


cuavas:
Unrelated to this, but on the topic of Model 2, I noticed that Fighting Vipers won’t boot when MAME is running on an ARM host.  I don’t think it’s SHARC DRC related.  The SHARC DRC seems to work well enough on ARM hosts for Konami NWK-TR, and it’s unaffected by `-drc_use_c`.

angelosa:
> Unrelated to this, but on the topic of Model 2, I noticed that Fighting Vipers won’t boot when MAME is running on an ARM host. I don’t think it’s SHARC DRC related. The SHARC DRC seems to work well enough on ARM hosts for Konami NWK-TR, and it’s unaffected by `-drc_use_c`.

This bs perhaps?

https://github.com/mamedev/mame/blob/b952c56fc4db51e7a44ac6a14467bd4db19cd81d/src/mame/sega/model2.cpp#L668-L706

PR #14836, opened 2026-01-21
Add a piece of instruction to Megami Tensei for MSX.

TheLostNinja:
Please advice me about multiline notes.


angelosa:
Usage is not suited for manual booklet style instructions about how to play a given game (and out of scope in general). 
Unless that's a thing required for testing (like how to install a given SW or how to create an user disk) I don't think there's any use here.

ajrhacker:
I agree with @angelosa here: usage instructions should be directed towards getting software to boot.

On a not strictly related note, I considered putting the defective image of the Enhancer ][ Utility Diskette that can be found on another site into apple2_flop_misc.xml together with usage notes to repair a bad byte in the critical DOWNLOAD program with specific BLOAD, POKE and BSAVE commands. It was of course more efficient to just put the resulting corrected image in the software list.

PR #14838, opened 2026-01-21
cave/cv1k*.cpp: Fix device naming and Minor cleanups

cam900:
cave/cv1k_v.cpp:
- Change ep1c* files filename to cv1k_v*, because This file is emulation core for cave CV1000 hardware specific and more closer to describe actual purpose, also ep1c is part family name of Altera Cyclone (1) FPGAs.
- Use logmacro.h for replace plain logerror function
- Use reference to replace pointers
- Reduce some duplicates
- Remove unnecessary device finder
- Use util::sext for signed values
- Fix spacings
- Add notes

cave/cv1k.cpp:
- Fix metadata for describe region
- Reduce unnecessary functions
- Use shorter typename value for consistency
- Reduce duplicates
- Fix naming
- Fix spacings
- Fix loggings
- Reduce literal tag usages
- Add notes


src/mame/cave/cv1k_v.h
```diff
@@ -1,24 +1,23 @@
...
-#define MAME_CAVE_EP1C12_H
+// emulation of Cave CV1000 blitter hardware
+// programmed to Altera Cyclone EP1C12 FPGA
+
+#ifndef MAME_CAVE_CV1K_V_H
+#define MAME_CAVE_CV1K_V_H
 
 #pragma once
 
 #define DEBUG_VRAM_VIEWER 0 // VRAM viewer for debug
```
cuavas: Can you give this macro a better name?  This is far too generic, given it affects a single device.

PR #14840, opened 2026-01-22
posixdir.cpp: implement XDG_*_HOME variables.

howl:
This change makes possible to define XDG variables that don't need to exist. When using them in the configuration file avoids warnings about not found variables and also sets them to their default values as per XDG spec when they are not defined.

### DIFF
--- a/src/osd/modules/file/posixdir.cpp
+++ b/src/osd/modules/file/posixdir.cpp
@@ -229,6 +229,31 @@ directory::ptr directory::open(std::string const &dirname)
 } // namespace osd
 
 
+std::string xdg_env;
+
+const char * getenv_xdg(const char * env_name)
+{
+	const char * env_value = std::getenv(env_name);
+
+	if (env_value) return env_value;
+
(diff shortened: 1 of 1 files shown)


cuavas:
No.  You’re throwing thread safety out the window, as well as adding non-standard behaviour.

howl:
@cuavas thread safety ok, about non-standard behaviour, don't think so.

XDG spec is standard https://specifications.freedesktop.org/basedir/latest/

And here you have a list of apps even tyrian2000 implementing this https://wiki.archlinux.org/title/XDG_Base_Directory#Supported

https://github.com/opentyrian/opentyrian/commit/39559c3

So this behaviour is exactly the only standard one.

The way I implemented need to be redone, will try to do myself but I don't have the necessary skills.

Following XDG will allow to clean configuration files that follow XDG, for example the official fedora mame package has defined double the paths because of this. If someone defines the XDG_*_HOME variables should be used but if not actually they aren't set as default when not defined so is needed to add a second path with the .local... stuff.

PR #14843, opened 2026-01-22
LUA:  Fixing an issue setting float options

npwoods:
LUA only has a `number` type, and as such currently the handler for options of type `int` is going to intercept floating point numbers, even if they can't be cast to integers.  This can be reproduced with the following command:

```
manager.options.entries['beam_dot_size']:value(2.5)
```

With this change, there is a single handler for LUA type `number` that takes a `float` and can set integer options provided that the `float` can be represented as an `int` losslessly.

### DIFF
--- a/src/frontend/mame/luaengine.cpp
+++ b/src/frontend/mame/luaengine.cpp
@@ -1360,19 +1360,14 @@ void lua_engine::initialize()
 			else
 				e.set_value(val ? 1 : 0, OPTION_PRIORITY_CMDLINE);
 		},
-		[] (core_options::entry &e, sol::this_state s, int val)
-		{
-			if (e.type() != core_options::option_type::INTEGER)
-				luaL_error(s, "Cannot set option to wrong type");
-			else
-				e.set_value(val, OPTION_PRIORITY_CMDLINE);
-		},
 		[] (core_options::entry &e, sol::this_state s, float val)
 		{
-			if (e.type() != core_options::option_type::FLOAT)
-				luaL_error(s, "Cannot set option to wrong type");
-			else
+			if (e.type() == core_options::option_type::FLOAT)
 				e.set_value(val, OPTION_PRIORITY_CMDLINE);
+			else if (e.type() == core_options::option_type::INTEGER && static_cast<int>(val) == val)
+				e.set_value(static_cast<int>(val), OPTION_PRIORITY_CMDLINE);
+			else
+				luaL_error(s, "Cannot set option to wrong type");
 		},
 		[] (core_options::entry &e, sol::this_state s, const char *val)
 		{


npwoods:
image

cuavas:
> LUA only has a `number` type, and as such currently the handler for options of type `int` is going to intercept floating point numbers, even if they can't be cast to integers. This can be reproduced with the following command:

That isn’t true since Lua 5.3, which introduced an integer type, which is a 64-bit twos complement integer, regardless of the host architecture.  We’re using Lua 5.4 in MAME.  You can verify this easily enough:

```
$ mamed.exe -console
       /|  /|    /|     /|  /|    _______
      / | / |   / |    / | / |   /      /
     /  |/  |  /  |   /  |/  |  /  ____/
    /       | /   |  /       | /  /_
   /        |/    | /        |/  __/
  /  /|  /|    /| |/  /|  /|    /____
 /  / | / |   / |    / | / |        /
/ _/  |/  /  /  |___/  |/  /_______/
         /  /
        / _/

mame 0.284
Copyright (C) Nicola Salmoria and the MAME team

Lua 5.4
Copyright (C) Lua.org, PUC-Rio

[MAME]> print(math.type(1))
integer
[MAME]> print(math.type(1.5))
float
[MAME]>
```

> ```
> manager.options.entries['beam_dot_size']:value(2.5)
> ```

Sol’s overload helper works with integer and float in Lua 5.3.  That works fine without your change:

```
$ git describe
mame0284-307-g3ef49944de8
$ mamed.exe -console
       /|  /|    /|     /|  /|    _______
      / | / |   / |    / | / |   /      /
     /  |/  |  /  |   /  |/  |  /  ____/
    /       | /   |  /       | /  /_
   /        |/    | /        |/  __/
  /  /|  /|    /| |/  /|  /|    /____
 /  / | / |   / |    / | / |        /
/ _/  |/  /  /  |___/  |/  /_______/
         /  /
        / _/

mame 0.284
Copyright (C) Nicola Salmoria and the MAME team

Lua 5.4
Copyright (C) Lua.org, PUC-Rio

[MAME]> opts = manager.options
[MAME]> print(opts.entries['beam_dot_size']:value())
1.0
[MAME]> opts.entries['beam_dot_size']:value(2.5)
[MAME]> print(opts.entries['beam_dot_size']:value())
2.5
[MAME]>
```

> With this change, there is a single handler for LUA type `number` that takes a `float` and can set integer options provided that the `float` can be represented as an `int` losslessly.

Your premise is wrong, and whatever is going wrong for you is caused by some other local issue.  I’m reverting this.

npwoods:
@cuavas can you tell me what compiler you're using?  I reproduced your findings regarding `math.type()`, so it looks like that part of my analysis is incorrect.  That said, I've noticed that if I build and run with `clang-cl`, it works (this is after your revert of course):
image

I still get the error with a fresh MinGW build

cuavas:
I just tested with:
* clang 20.1.8, libc++, UCRT, arm64
* GCC 10.3.0, libstdc++, MSVCRT, x86-64

And it works with both.

happppp:
MINGW64 GCC 15.2.0 local compile: I can repro the error with ```manager.options.entries['beam_dot_size']:value(2.5)```
Official MAME 0.284: it works fine.

edit: lemme paste the whole thing for mame 0.284

```
mame ace -console
       /|  /|    /|     /|  /|    _______
      / | / |   / |    / | / |   /      /
     /  |/  |  /  |   /  |/  |  /  ____/
    /       | /   |  /       | /  /_
   /        |/    | /        |/  __/
  /  /|  /|    /| |/  /|  /|    /____
 /  / | / |   / |    / | / |        /
/ _/  |/  /  /  |___/  |/  /_______/
         /  /
        / _/

mame 0.284
Copyright (C) Nicola Salmoria and the MAME team

Lua 5.4
Copyright (C) Lua.org, PUC-Rio

[MAME]> print(manager.options.entries['beam_dot_size']:value())
1.0
[MAME]> manager.options.entries['beam_dot_size']:value(2.5)
[MAME]> print(manager.options.entries['beam_dot_size']:value())
2.5
[MAME]>
```

and the local compile
```
[MAME]> manager.options.entries['beam_dot_size']:value(2.5)
error:  [string "..."]:2: Cannot set option to wrong type
```

holub:
current master, clang 21.1.8 (1+b1)
```
[MAME]> manager.options.entries['beam_dot_size']:value(2.5)
error:  [string "..."]:2: Cannot set option to wrong type
```

cuavas:
> Official MAME 0.284: it works fine.

MAME 0.284 should have the opposite problem – an error attempting to set an integer option using an integer, unless there’s dark magic at work.

cuavas:
See if d4fb65d44adb56a054a162de948f2f4298f9da7f affects it.

holub:
It works now!

happppp:
It works here as well on my local compile (the beam_dot_size test)

PR #14844, opened 2026-01-22
pc98.xml: New software list item.

TheLostNinja:
_(redacted -Kale)_ was created by Neko Project 21/W (cause MAME always creating a blank images in unformatted form, which PC-98 Windows installers cannot work with) and filled by NEC Win95 default means.

### DIFF
--- a/hash/pc98.xml
+++ b/hash/pc98.xml
@@ -1276,7 +1276,7 @@ Utility disk file "BMENU." sports ugly GFX cyan fills (especially noticeable whe
         	<publisher>Microsoft</publisher>
         	<part name="flop1" interface="floppy_3_5">
             		<dataarea name="flop" size="1281968">
-                		<rom name="win95b.img" size="1281968" crc="106cc9fa" sha1="2dde4d7d51494ab5e7805aefcea26ad80a4ba104"/>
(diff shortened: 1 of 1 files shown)


TheLostNinja:
_EDIT: no roms, software disks, copyrighted work attachments. Use the [submission form](https://wiki.mamedev.org/index.php?title=Submitting_Source_Code#Submitting_ROM_or_disk_dumps) instead._

angelosa:
Isn't the root issue that Windows 95/98 expects a 1.44MB formatted disk and that's doable thru a `format` switch that escape me off the bat?

TheLostNinja:
> Isn't the root issue that Windows 95/98 expects a 1.44MB formatted disk and that's doable thru a `format` switch that escape me off the bat?

I tried to create a blank FD image inside slot 2 (-fdc_2hd:1 35hd) - in **pc98** format as **u35dshd**, but its size is 160 KB and after ejection and insertion later second slot recognizes it as invalid image (but for some reason 5"25 first slot calmly receives).

angelosa:
I did this myself in https://github.com/mamedev/mame/commit/5d74f33ff0d16512bd4dcfe93e9e7d98122906cf , closing.

PR #14845, opened 2026-01-22
WIP Furby Connect addition

lib-fluff-so:


mamehaze:
@lib-fluff-so 
I'm in the process of refactoring the GPL162xx code.

If you're still interested in this being added in some form, could you provide me with contact details so that you can send the required files across.

PR #14849, opened 2026-01-23
DRC: Add support for generating invariant code

cuavas:
This allows back-ends and CPU cores to keep invariant code across code cache resets.  Impact for CPU core developers is minimal.  Implemented for Hyperstone E1 and PowerPC.  This makes a cache reset substantially cheaper for PowerPC because you’re no longer immediately liable for over 100 KiB of helper functions that won’t have changed since last time.

It should be easy enough for other people to adapt other CPU cores that generate relatively large amounts of invariant code.

Also cleans up a bunch of crud.


cuavas:
If this breaks mapvar recover (used in PowerPC exception handling, etc.), yell and I’ll sort it out.  It shouldn’t break it, but that code is pretty awful, and you feel dirty any time you touch it.

PR #14851, opened 2026-01-23
Skeleton driver for Arachnid "Super Shuffle"

clawgrip:
Arachnid Super Shuffle [Victor Fernandez (City Game)]

### DIFF
--- a/src/mame/pc/sshuffle.cpp
+++ b/src/mame/pc/sshuffle.cpp
@@ -61,9 +61,6 @@ void sshuffle_state::sshuffle_io(address_map &map)
 {
 }
 
-static INPUT_PORTS_START(sshuffle)
-INPUT_PORTS_END
-
 void sshuffle_state::sshuffle(machine_config &config)
 {
 	PENTIUM4(config, m_maincpu, 100'000'000); // Actually an AMD Sempron SDX140HBK13GQ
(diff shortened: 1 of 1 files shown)


src/mame/pc/sshuffle.cpp
```diff
@@ -0,0 +1,92 @@
...
+	map(0x000e0000, 0x000fffff).rom().region("bios", 0xe0000);
+	map(0xfff00000, 0xffffffff).rom().region("bios", 0);
+}
+
+void sshuffle_state::sshuffle_io(address_map &map)
+{
+}
+
+static INPUT_PORTS_START(sshuffle)
+INPUT_PORTS_END
```
angelosa: fwiw I don't think we ever need `INPUT_PORTS` for any PC based Arcade, even less for PCIe targets (input devices will be from cards). So in these cases you remove this define then push a null input def at bottom:
`GAME(2011, sshuffle, 0, sshuffle, 0, sshuffle_state, empty_init, ROT0, "Arachnid", "Super Shuffle", MACHINE_NO_SOUND | MACHINE_NOT_WORKING)`

clawgrip: Done, thanks a lot!!!

PR #14852, opened 2026-01-23
generic_cdrom.xml: New software list items.

TheLostNinja:
**New working software list additions**

- Road Rash (USA, Europe)
- Disney's Hercules Action Game (USA)

**New partially working software list additions**

- Claw (Spain v1.1)

### DIFF
--- a/hash/generic_cdrom.xml
+++ b/hash/generic_cdrom.xml
@@ -84,8 +84,8 @@ pc98: InstallShield is painfully slow unless floppy is inserted
 		-->
 		<description>Disney's Hercules Action Game (USA)</description>
 		<year>1997</year>
-		<developer>Eurocom Developments</developer>
 		<publisher>Disney Interactive</publisher>
+		<info name="developer" value="Eurocom Developments" />
 		<sharedfeat name="compatibility" value="ibmpc,pc98" />
 		<part name="cdrom" interface="cdrom">
 			<diskarea name="cdrom">
@@ -209,8 +209,8 @@ pc98: has color and graphic issues during gameplay
 		-->
 		<description>Road Rash (USA, Europe)</description>
 		<year>1996</year>
-		<developer>Papyrus Design Group</developer>
 		<publisher>Electronic Arts</publisher>
+		<info name="developer" value="Papyrus Design Group" />
 		<sharedfeat name="compatibility" value="ibmpc,pc98" />
 		<part name="cdrom" interface="cdrom">
 			<diskarea name="cdrom">


hash/generic_cdrom.xml
```diff
@@ -55,6 +74,26 @@ pc98: InstallShield is painfully slow unless floppy is inserted
...
+	<software name="hercules">
+		<!--
+		redump.org/disc/64876
+		<rom name="Disney's Hercules Action Game (USA) (Track 1).bin" size="307914432" crc="1233a831" sha1="4a067edcc5673be944c68d6564da6f67ef184c09"/>
+		<rom name="Disney's Hercules Action Game (USA) (Track 2).bin" size="364176624" crc="02f6272f" sha1="bd3051ec23f9484c48e26159ec3ed22afd28e355"/>
+		<rom name="Disney's Hercules Action Game (USA).cue" size="263" crc="df17a4c6" sha1="6d71ac964a00da261aadebed5db1ebb60a12ec33"/>
+		-->
+		<description>Disney's Hercules Action Game (USA)</description>
+		<year>1997</year>
+		<developer>Eurocom Developments</developer>
```
angelosa: This will fail validation.

hash/generic_cdrom.xml
```diff
@@ -162,6 +201,24 @@ pc98: has color and graphic issues during gameplay
 		</part>
 	</software>
 
+	<software name="roadrash">
```
angelosa, line 204: This requires a Pentium class machine to begin with, as per `roadrashi` in `ibm5170_cdrom.xml`. Also is it really installable on `pc98`?

hash/generic_cdrom.xml
```diff
@@ -34,9 +34,28 @@ pc98: InstallShield is painfully slow unless floppy is inserted
...
+			<diskarea name="cdrom">
+				<disk name="claws" sha1="d134487f49fb429075e4daebbc881befdefffc98" />
+			</diskarea>
+		</part>
+	</software>
+
+	<!-- Very slow on pc9821ap2 -->
 	<!-- TODO: has transparency issues on main menu -->
-	<software name="clawj" supported="partial">
+	<software name="clawj" cloneof="claws" supported="partial">
```
angelosa, line 58: This should have a regular World release, which should be the parent _if_ it's installable on pc98, otherwise it doesn't belong to this list.

hash/generic_cdrom.xml
TheLostNinja: > This will fail validation.

Because of ' symbol?

angelosa: No, because `<developer>` doesn't exist in the DTD schema. It's `<info name="developer" value="<whatever>" />`, plus also needs to be moved below publisher node.

hash/generic_cdrom.xml
angelosa: I tried, it first moans about the language, then aborts with DirectDraw not available, which is either down to a lack of video acceleration emulation or it's just incompatible.
Either way this doesn't belong to `generic_cdrom.xml` at current time.

TheLostNinja: About the language - game just choosing the default English if the current system language is not included in the game language base.

I'll try to install RR to Neko Project 21\W with emulated window accelerator to ensure, is this game incompatiable with PC-98 at all, or with basic 9821 video subsystem only.

TheLostNinja: UnCHDfied **win95o2** with **PC-9821Xe10, Xa7e, Xb-10 built-in** window accelerator gives font warning "Badloc: No error".

angelosa: That means you didn't test this, and my original point still stands. I only accept _known confirmed/advertised games on `generic_cdrom.xml`_, anything else goes to their own list (and this is valid for the other two games as well).

angelosa:
Closing this, it's all stuff that needs moving in `ibm5170_cdrom.xml` anyway, which means redoing from scratch.
I thought that `Cross-platform` was explicit enough to avoid people pushing single platform disks in there ...

PR #14853, opened 2026-01-23
Skeleton driver for Coastal Amusements "Sea Wolf - Next Mission"

clawgrip:
Coastal Amusements Sea Wolf - Next Mission [Victor Fernandez (City Game)]

### DIFF
--- a/src/mame/pc/seawolfnm.cpp
+++ b/src/mame/pc/seawolfnm.cpp
@@ -1,14 +1,14 @@
 // license:BSD-3-Clause
 // copyright-holders:
 
-/**********************************************************************************************************************************
+/****************************************************************************
 
 Skeleton driver for "Sea Wolf - Next Mission", from Coastal Amusements Inc.
 
 Based on an Foxconn "45CMX" PCB:
 -Intel Celeron SL9XN 1.8GHz/512/800/06 CPU.
 -1024MB DDR2-667 RAM.
--North Bridge : Intel® 945GC/945G chipset, South Bridge : Intel® ICH7.
+-North Bridge: Intel 945GC/945G chipset; South Bridge: Intel ICH7.
 
 Jaton Video-PX8400GS-EX video board with 512 MB RAM (NVIDIA GeForce 8400GS).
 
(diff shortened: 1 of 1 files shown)


src/mame/pc/seawolfnm.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+ROM_START(seawolfnm)
+	ROM_REGION32_LE(0x80000, "bios", 0)
+	ROM_LOAD("25l4005.bin", 0x00000, 0x80000, CRC(2a6b7cf2) SHA1(0be15605c686bec5109b1cbe91b45edca6c46d35))
+
+	ROM_REGION(0x8000, "io", 0)
+	ROM_LOAD("sea_wolf_1_89c5131a.bin", 0x0000, 0x8000, NO_DUMP)
+
+	DISK_REGION("ide:0:hdd")
+	DISK_IMAGE("sea_wolf_the_next_mission_0.9.9_12-25-08_coastal_amusements_inc", 0, SHA1(dc47d60ba04d33aece91838598828edbf79e8eb5)) // Image tested on real hardware, may contain operator or players game
```
angelosa: `BAD_DUMP` then, it will eventually need a roundtrip to flush that data.

src/mame/pc/seawolfnm.cpp
```diff
@@ -0,0 +1,84 @@
...
+	required_device<cpu_device> m_maincpu;
+
+	void seawolfnm_io(address_map &map) ATTR_COLD;
+	void seawolfnm_map(address_map &map) ATTR_COLD;
+};
+
+
+void seawolfnm_state::seawolfnm_map(address_map &map)
+{
+}
```
angelosa, line 53: x86 basics part 1:
```c++
map(0x00000000, 0x0009ffff).ram();
map(0x000e0000, 0x000fffff).rom().region("bios", size_offset);
map(rom_start,  0xffffffff).rom().region("bios, 0);
```
Where `size_offset` is BIOS ROM size minus 0x20000 (in this case 0x60000), `rom_start` is 0xfffffffff - rom size + 1 (0xfff80000). People will get angry that MAME locks up on execution because the memory map is unimplemented (and a P4 will be very eager to do that).

clawgrip: Done, thanks a lot!

PR #14856, opened 2026-01-24
cpu/sharc.cpp, sega/model2.cpp: move 16-bit IOP write handling to SHARC core

gm-matthew:
SHARC uses a default host bus width of 16 bits for the IOP registers. Some Model 2B games set the width to 32 bits using bits 4-5 of the SYSCON register before writing to the necessary IOP registers, and then change it back to 16 bits before uploading the DSP program.

All Model 2B games (except Power Sled) tested and successfully booting.

Source: [ADSP-2106x SHARC Processor User’s Manual](https://www.analog.com/media/en/dsp-documentation/processor-manuals/50836807228561adsp2106xsharcprocessorusersmanual_revision2_1.pdf), page E-26 (634 in PDF), Host Packing Mode

### DIFF
--- a/src/devices/cpu/sharc/sharc.cpp
+++ b/src/devices/cpu/sharc/sharc.cpp
@@ -13,6 +13,9 @@
 
 #include "emuopts.h"
 
+//#define VERBOSE 1
+#include "logmacro.h"
+
 
 #define DISABLE_FAST_REGISTERS      1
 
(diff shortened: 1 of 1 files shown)


src/devices/cpu/sharc/sharc.h
```diff
@@ -332,6 +332,8 @@ class adsp21062_device : public cpu_device
 		uint32_t curlcntr;
 		uint32_t lcntr;
 		uint8_t extdma_shift;
+		uint32_t iop_write_num;
+		uint32_t iop_data;
```
cuavas, line 336: Should these be in save states?  If not, why not?

gm-matthew: Indeed they should.

cuavas:
This *does* fix Fighting Vipers on ARM hosts.  @angelosa was correct about which part of the code was somehow tripping undefined behaviour.

PR #14857, opened 2026-01-24
SDL3

cuavas:
This is (hopefully) the SDL3 branch rebased onto master.  I’m not sure if it really is or not, because the revision history was a complete mess, with about 350 cherry-picked mainline commits, and two instances of each of the actual SDL3 commits because the changes were reverted in between for reasons that are beyond me.

@rb6502 please check this and if it’s actually what you wanted to do, either edit this PR to have a sensible title and description, or close this PR and make a new one, then squash and merge it with a sane commit message.


happppp:
Original diff here: https://github.com/mamedev/mame/compare/d787b4ac6d1e...d8b971f77911
stats are the same (number of files changed, number of lines added, number of lines removed)

src/osd/modules/input/input_dinput.cpp
```diff
@@ -107,8 +107,7 @@ Rz          Rudder
 #include "util/corestr.h"
 
 #ifdef SDLMAME_WIN32
-#include <SDL2/SDL.h>
-#include <SDL2/SDL_syswm.h>
+#include <SDL3/SDL.h>
 #endif
```
cuavas, line 111: I think this file may have been missed for adding support for building with either SDL2 or SDL3?

rb6502: Correct, the files for Windows SDL2 vs SDL3 are still in my Windows VM.  I'm putting together a more complete/correct version.

src/osd/modules/render/drawogl.cpp
```diff
@@ -1246,20 +1246,11 @@ int renderer_ogl::draw(const int update)
 
 		GLsizei iScale = 1;
 
-		#ifndef SDLMAME_SDL3
```
rb6502, line 1249: This breaks SDL3, the enclosing #ifndef should not have been removed.

cuavas: Check below – I changed the other condition.  It isn’t necessary for SDL2 either because 2.0.14 is required, 2.0.10 is no longer supported.  It’s only being used for OSD_MAC now

src/osd/modules/render/drawogl.cpp
```diff
@@ -1246,20 +1246,11 @@ int renderer_ogl::draw(const int update)
...
+		    assuming you do support Retina.
 		*/
-		#if defined(SDLMAME_MACOSX) && !defined(OSD_MAC)
-		SDL_version sdlVers;
-		SDL_GetVersion(&sdlVers);
-		// Only do this if SDL is not at least 2.0.11.
-		if ((sdlVers.major == 2) && (sdlVers.minor == 0) && (sdlVers.patch < 11))
-		#endif
-		#if defined(SDLMAME_MACOSX) || defined(OSD_MAC)
+		#if !defined(SDLMAME_MACOSX) && defined(OSD_MAC)
```
cuavas, line 1253: This part:
```c++
		#if !defined(SDLMAME_MACOSX) && defined(OSD_MAC)
```

It’s disabled altogether for SDLMAME on macOS now, because it was only used for SDL < 2.0.11 before, and we don’t allow building with 2.0.10 at all any more (input_sdl.cpp at least depends on newer SDL than that to compile at all).

The outer `#ifndef SDLMAME_SDL3` shouldn’t be needed.

Or does SDL3 not `#define SDLMAME_MACOSX` any more?  If that’s the case, what’s the proper way to detect SDLMAME on macOS at compile time?

PR #14858, opened 2026-01-24
posixdir.cpp: implement XDG_*_HOME variables

howl:
Third try doing thread safe @cuavas.

SDL3 changes entered in my branch so I thought better to start form zero and even simplify the logic before in another previous commit.

This implements the XDG standard as is, for direct reference: https://specifications.freedesktop.org/basedir/latest/

When the following env are unset or empty they should be taken by programs with the respective default values:
$XDG_DATA_HOME = $HOME/.local/share
$XDG_CONFIG_HOME = $HOME/.config
$XDG_STATE_HOME = $HOME/.local/state
$XDG_CACHE_HOME = $HOME/.cache

You can see lots of common apps that have implemented this standard and the respective commits here:
https://wiki.archlinux.org/title/XDG_Base_Directory#Supported

Implementing this helps to not being necessary to define paths like this:
cfg_directory $XDG_CONFIG_HOME/mame/cfg;$HOME/.config/mame/cfg;$HOME/.mame/cfg

And being simply like:
cfg_directory $XDG_CONFIG_HOME/mame/cfg;$HOME/.mame/cfg

.mame retained for pre XDG compat but is very uncommon in fedora mame.

### DIFF
--- a/src/osd/modules/file/posixdir.cpp
+++ b/src/osd/modules/file/posixdir.cpp
@@ -291,10 +291,36 @@ std::string osd_subst_env(std::string_view src)
 
 			if (!var.empty())
 			{
-				const char *const exp = std::getenv(var.c_str());
-				if (exp)
+				bool env_found = false, is_xdg_env = false;
+				const char* const xdg_envs[4][2] = {
+					{"XDG_CONFIG_HOME", "/.config"},
+					{"XDG_DATA_HOME",   "/.local/share"},
+					{"XDG_STATE_HOME",  "/.local/state"},
+					{"XDG_CACHE_HOME",  "/.cache"}};
+
+				for (const char* const* xdg_env: xdg_envs)
+					if (var == xdg_env[0])
+					{
+						is_xdg_env = true;
+						break;
+					}
+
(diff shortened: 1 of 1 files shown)


howl:
Set again as draft, I'm going to complain the standard also the set but empty part.

howl:
Now also takes into account if XDG_*_HOME variables are set but empty.

For my side is actually finished but all the suggestion changes that should be needed to comply are welcome.

howl:
@cuavas could you review this? I tried my best to improve logic and then add the intended behaviour to comply with the XDG spec.

howl:
@cuavas apart from the thread safety actually resolved there is anything else needed to be done for this?

howl:
Hi @cuavas could you please review this one?

Here https://bugzilla.redhat.com/attachment.cgi?id=2134080&action=diff in core output paths you can see how useful is to implement XDG spec env dirs.

Actually fedora, without that srpm change, uses them but when they don't exists gives a warning and follows up to \$HOME/.config/%{name} style path definition but that makes lots of warning arising. Even with the warns makes sense to define them as if they are set allow to move the mame paths as per XDG. With PR we achieve three things, be XDG compilant, avoid warnings when using XDG envs when they are unset or empty (part of the XDG spec behaviour) and simplifies path definitions.

PR #14861, opened 2026-01-25
Save some more items inside Sega's chips I hope would be helpful for System 1(6/8) state saving.

TheLostNinja:
I'm looking for you verdict of testing results.


happppp:
315_5195: I don't think m_curregion and m_open_bus_recurse need to be saved, and the others can't be saved (I expect another compile error after your last try).

This may help with savestates: https://github.com/mamedev/mame/commit/5e0c6f4f44994320eec933068440b6024b6bfe00

happppp:
Closing. See my comment above about 315_5195. And the ones in sega16sp are configuration variables that don't need to be saved (none of the drivers change these offsets at runtime)

PR #14864, opened 2026-01-26
Save Sound Blaster 16 state.

TheLostNinja:
Now [Mega Cop](https://www.old-games.ru/game/6304.html) (at least, its floppy version) is fully beatable in "perfectionist mode" without any digital sound disappearing after another new load. :)

### DIFF
--- a/src/devices/bus/isa/sb16.cpp
+++ b/src/devices/bus/isa/sb16.cpp
@@ -805,6 +805,19 @@ void sb16_lle_device::device_start()
 	save_item(NAME(m_dma16_cnt));
 	//save_item(NAME(m_adc_fifo)); /*temporarely decision*/
 	//save_item(NAME(m_dac_fifo));
+	save_item(NAME(m_adc_fifo_head));
+	save_item(NAME(m_adc_fifo_tail));
+	save_item(NAME(m_dac_fifo_head));
+	save_item(NAME(m_dac_fifo_tail));
+	save_item(NAME(m_adc_r));
+	save_item(NAME(m_dac_r));
+	save_item(NAME(m_adc_h));
+	save_item(NAME(m_dac_h));
+	save_item(NAME(m_irq8));
+	save_item(NAME(m_irq16));
+	save_item(NAME(m_irq_midi));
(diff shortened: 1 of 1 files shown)


TheLostNinja:
And CD version too.

src/devices/bus/isa/sb16.cpp
```diff
@@ -787,6 +787,37 @@ void sb16_lle_device::device_start()
...
+	save_item(NAME(m_adc_fifo_ctrl));
+	save_item(NAME(m_ctrl8));
+	save_item(NAME(m_ctrl16));
+	save_item(NAME(m_mpu_byte));
+	save_item(NAME(m_dma8_len));
+	save_item(NAME(m_dma16_len));
+	save_item(NAME(m_dma8_cnt));
+	save_item(NAME(m_dma16_cnt));
+	//save_item(NAME(m_adc_fifo)); /*temporarely decision*/
+	//save_item(NAME(m_dac_fifo));
```
angelosa, line 807: `temporary` to begin with, I smell because can't save an union so you got compile errors? (I legit don't even know how, putting aside that those FIFOs should really use `util::fifo` instead and that the hookup itself should be moved in a separate device).

TheLostNinja: I looked inside **cpu/nec/nec.cpp**, where unions are used too - and there it gets saving in a largest form.
**nec.h**:
`union necbasicregs
	{                   /* eight general registers */
		uint16_t w[8];    /* viewed as 16 bits registers */
		uint8_t  b[16];   /* or as 8 bit registers */
	};

	necbasicregs m_regs;`
**nec.cpp**:
`save_item(NAME(m_regs.w));`

Maybe, with **m_(adc/dac)_fifo** should behave by the same way?

angelosa: I'll decide later what to do with these, but for now just remove that comment.

MooglyGuy:
Closing, as user was blocked from the project due to repeated refusal to learn from past feedback or address feedback on pull requests, instead spamming the project with more and more new pull requests with the same issues.

PR #14866, opened 2026-01-27
tvgames/xavix.cpp: Document ekara alternate input dip

einstein95:
BGM Down/Up buttons not affected, surprisingly.

Requesting feedback from @mamehaze.


mamehaze:
I think this may be related to e-pitch, which has a slightly different layout.

I recall the inputs being funky on a few sets, so maybe some only support one of the layouts?

PR #14872, opened 2026-01-28
Foundin' a base for PC-98 state saving.

TheLostNinja:
This platform is also has quite enough of "tough" arcades, where saving is if not necessary, then desireable at least.

### DIFF
--- a/src/devices/video/upd7220.cpp
+++ b/src/devices/video/upd7220.cpp
@@ -758,6 +758,10 @@ void upd7220_device::device_start()
 	save_item(NAME(m_disp));
 	save_item(NAME(m_gchr));
 	save_item(NAME(m_pattern));
+	save_item(NAME(m_dma_type));
+	save_item(NAME(m_dma_mod));
+	save_item(NAME(m_dma_data));
+	save_item(NAME(m_dma_transfer_length));
 	save_item(NAME(m_mask));
 	save_item(NAME(m_pitch));
 	save_item(NAME(m_ra_addr));
--- a/src/mame/nec/pc9801.cpp
+++ b/src/mame/nec/pc9801.cpp
@@ -1658,15 +1658,6 @@ MACHINE_START_MEMBER(pc9801_state,pc9801_common)
 {
 	m_rtc->cs_w(1);
 	m_rtc->oe_w(1);
-}
-
-MACHINE_START_MEMBER(pc9801_state,pc9801f)
-{
-	MACHINE_START_CALL_MEMBER(pc9801_common);
-
-	m_fdc_2hd->set_rate(500000);
-
-	m_sys_type = 0x00 >> 6;
 
 	save_item(NAME(m_uart_irq_mask));
 	save_item(NAME(m_uart_irq_pending));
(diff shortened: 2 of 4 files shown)


TheLostNinja:
State load in Rude Breaker lefts a flickering trail of previous bitmap. Let's search for an undersaved items in other source files.

TheLostNinja:
Did I opened an older build by mistake, or you still having the trail bug staying at its point too?

JimCarlTay:
I haven't tested. It's because I know nothing about how to run a PC-98 computer system on MAME, and to load a game there. Is it as straightforward as just loading a floppy on an X68000?

TheLostNinja:
> I haven't tested. It's because I know nothing about how to run a PC-98 computer system on MAME, and to load a game there. Is it as straightforward as just loading a floppy on an X68000?

It's a game from **pc98_cd:dstat_10**, installable through the Windows (althought the game itself is for PC-98 DOS). First you should install the **msdos62** as it described [here](https://forums.bannister.org/ubbthreads.php?ubb=showflat&Number=112996#Post112996) (**ex**cluding the _neccdm.sys_ driver installation phase - because CD support in **msdos62** is well enough out of the box actually - and with the correctives that prescribed in _config.sys_ CD driver must be DEVICE=A:¥DOS¥NECCD**C**.SYS /D:CD_101, and for the **pc98_cd:win95** setup.exe launching without of problems before the "[dont_remember_path]\print.[dont_remember extension]" line must be the _"rem "_ prescribed). Afterward, the Windows 95 installation has very little difference with analogical procedure on IBM PCs.

angelosa:
This just stomps on feet: most of this stuff needs to be composed rather than being in base driver (the use of legacy `MACHINE_START` / `MACHINE_RESET` calls is indicative of that). On top of my head:
- `m_analog16` is from PC-9801-24 (optional on earlier models);
- `mouse` is PC9871 port;
- anything related to the earlier FDC 2HD is yet another C-Bus card (which somehow doesn't work [when moved](https://github.com/angelosa/mame_scratch/blob/main/src/pc98/fdd_2hd.cpp)), likewise for the integrated 3 mode i/f ...
- PEGC integrates with VGA chip(s), we already have a few cases where it warrants a split (Flashback and Silent Service 256 color mode version).

I didn't compose big header(s) on top just to see stuff like this ...

TheLostNinja:
> This just stomps on feet: most of this stuff needs to be composed rather than being in base driver (the use of legacy `MACHINE_START` / `MACHINE_RESET` calls is indicative of that). On top of my head:
> 
>     * `m_analog16` is from PC-9801-24 (optional on earlier models);
> 
>     * `mouse` is PC9871 port;
> 
>     * anything related to the earlier FDC 2HD is yet another C-Bus card (which somehow doesn't work [when moved](https://github.com/angelosa/mame_scratch/blob/main/src/pc98/fdd_2hd.cpp)), likewise for the integrated 3 mode i/f ...
> 
>     * PEGC integrates with VGA chip(s), we already have a few cases where it warrants a split (Flashback and Silent Service 256 color mode version).
> 
> 
> I didn't compose big header(s) on top just to see stuff like this ...

Then we need to plan the per-device splitting.

MooglyGuy:
What's this "we"?

angelosa:
Closing, don't have any use of this at current time.

PR #14873, opened 2026-01-28
Miniforce and CPU1

JoakimLarsson:
Added a SYS68K-CPU1 panel, wrired up the last ACIA, fixed layout to reflect what serial port is connected to what and some cleaning up


MooglyGuy:
Why are we allowing AI-generated commits now? @cuavas @galibert

cuavas:
Yeah, I’m leaning towards just reverting this, too.  The whole change seems misguided.

galibert:
Can we do conditional-depending-on-what-slots-are-there layouts?

PR #14876, opened 2026-01-28
Sys68k cpu20

JoakimLarsson:
Added panel view for sys68k-cpu20 with reset and abort buttons

### DIFF
--- a/src/emu/layout/sys68k_cpu1.lay
+++ b/src/emu/layout/sys68k_cpu1.lay
@@ -105,9 +105,4 @@ Matches real hardware: RESET and ABORT pushbuttons, HALT LED, P3/P4/P5 connector
 		<screen tag="p4:terminal:terminal_screen"><bounds x="52" y="0" width="460" height="345" /></screen>
 	</view>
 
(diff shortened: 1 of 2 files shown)


src/devices/bus/vme/sys68k_cpu20.cpp
```diff
@@ -467,6 +485,43 @@ void vme_sys68k_cpu20_card_device_base::device_reset()
 				m_boot_mph.remove();
 			}
 		});
+
+	// Detect connected RS232 devices and update connector status outputs
```
happppp, line 489: That looks hacky IMO. Why does it need to know this? Is it only for aesthetics? I saw the same thing in your previous commit.

edit: maybe can be done with luascript in the .lay file? Still hacky, but better there than in the device.

JoakimLarsson: Might be, since there are no possibility to view any cables when plugged in it is helpful to know what device is plugged in where so I just added some text to indicate that. Manuals usually talk about plugin the terminal in port p2 and printer in p4 or whatever. But there might be a better way to know that, apart from remember it

PR #14877, opened 2026-01-28
hash/gba.xml: mark Metroid Zero Mission as partially supported

chungy:
The main game works fine, but the “Original Metroid” mode unlocked by beating the main game (the prototype ROMs let you set cleared flags and unlock it immediately) has severe graphical issues, rendering it almost unplayable.

### DIFF
--- a/hash/gba.xml
+++ b/hash/gba.xml
@@ -22238,7 +22238,8 @@ Hangs during cutscene after beating Maxim for best ending path (i.e. before real
 		</part>
 	</software>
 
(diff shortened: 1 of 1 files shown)


cuavas:
Really?  I could have sworn I played it years ago in MAME in the Japan version (`metroid0j`) without issues.  Has it regressed?  Are the graphical issues specific to the prototype?

chungy:
No, I played normally on `metroid0u` and it's broken.

Screenshot_20260128_081157

PR #14888, opened 2026-01-29
snesb.cpp: Correct The Pirates of Dark Water Year [sjy96525]

sjy555525:


sjy555525:
The Pirates of Dark Water titlescreen says 1996.

mamehaze:
I'm not sure how much we can trust the copyright displayed on these bootlegs to be honest, but we may as well use the dates for lack of any better information.

(a lot of the cheap clone SNES/Genesis chipsets I've seen come from the early 2000s instead, although maybe these were earlier clone hardware)

PR #14895, opened 2026-01-30
sequential/prophet5.cpp: moved voice trimmers to dedicated, per-voice input ports.

m1macrophage:
Fixes the threading issue raised in PR [14826](https://github.com/mamedev/mame/pull/14826#discussion_r2740165546), and provides better encapsulation.

But let me know if some alternative is preferable.

### DIFF
--- a/src/mame/sequential/prophet5.cpp
+++ b/src/mame/sequential/prophet5.cpp
@@ -47,6 +47,7 @@ illuminated). The synth is unresponsive while this is happening.
 */
 
 #include "emu.h"
+
 #include "cpu/z80/z80.h"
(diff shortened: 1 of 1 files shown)


m1macrophage:
... this approach assumes that `prophet5_voice_device` will outlive the port configuration defined by `device_input_ports()`. Or at the very least, it assumes that the port names won't be accessed after the device object is deallocated.

If that is not a safe assumption, I can look at alternatives.

m1macrophage:
To answer my own question: the device's ioports seem to be deallocated before the device. Assuming this is deliberate, rather than happenstance, this change should be safe.

PR #14896, opened 2026-01-30
Sega OutRun HW driver files consolidation into single file. 

TheLostNinja:
Since **segaorun.h** file is not cross-referred by any other driver, and all three files is small enough, their splitness looks like a pure anachronism.

### DIFF
--- a/src/mame/sega/segaorun.cpp
+++ b/src/mame/sega/segaorun.cpp
@@ -300,18 +300,7 @@ Notes:
 
 #include "outrun.lh"
 
-
-//**************************************************************************
-//  CONSTANTS
-//**************************************************************************
-
-const auto MASTER_CLOCK = XTAL(40'000'000);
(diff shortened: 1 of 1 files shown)


TheLostNinja:
Looks like this driver is one of most archaic inside MAME, hasn't been seriously rewrote by anybody for the years.

happppp:
The driver doesn't look outdated to me. Aaron did good work with his Sega drivers rewrite.

TheLostNinja:
> The driver doesn't look outdated to me. Aaron did good work with his Sega drivers rewrite.

I just remember (althought memory may deceive me) the examples of drivers, the intervals between commits inside which could be 7 (!!!) years even.

cuavas:
If something is working well and there isn’t a better way of doing things, there’s no reason to change it for the sake of changing it.

PR #14898, opened 2026-01-30
emu/debug: Allowed device_state_interface to participate in global symtable state

holub:
After this, for device which implements device_state_interface we can access state from global symtable.
CPU (exec) devices continue to publish their state to separate CPU symtables (child of GLOBAL).

~~Also suggesting to rename `machine().debugger().cpu()` to `machine().debugger().state()` as "cpu" confusing term in this context as far as we can have multiple cpu.~~ <<< will do separately, if requested

### DIFF
--- a/src/emu/debug/debugcpu.cpp
+++ b/src/emu/debug/debugcpu.cpp
@@ -488,7 +488,7 @@ device_debug::device_debug(device_t &device)
 	, m_state(nullptr)
 	, m_disasm(nullptr)
 	, m_flags(0)
-	, m_symtable(std::make_unique<symbol_table>(device.machine(), symbol_table::CPU_STATE, &device.machine().debugger().cpu().global_symtable(), &device))
+	, m_symtable(nullptr)
 	, m_stepaddr(0)
 	, m_stepsleft(0)
 	, m_delay_steps(0)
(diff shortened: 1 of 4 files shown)


holub:
I'd like those to be delivered/moved forward: ~#14908~ ~#14898~ ~#14765~ ~#14754~ ~#14740~ ~#14739~ ~#14328~ ~#13122~
Who can I blackmail for this?

galibert:
I don't have a problem with renaming state() to cpu(), but I do have a problem with not renaming debugger_cpu to debugger_state then.  And the filename too.

holub:
Yeah, I was waiting for feedback on this change.
So... revert the field or rename the class as well?

holub:
I made an executive decision and only kept functional changes in this PR.
If anybody support "rename", I can bring it as a separate change.

galibert:
For other things I want to do I'm going to end up massively reorganizing that crap, so tabling the rename is a good thing

cuavas:
So what happens when two devices in a system have a state item with the same name?

holub:
This is unpleasant situation where last registered win. But we are far away from this situation as the device needs to be "state" and not be executable. I believe now number of devices which match this condition are close to 0.

holub:
> So what happens when two devices in a system have a state item with the same name?

An option to minimize conflicts if you will: #14950

PR #14827, opened 2026-01-18
Add the rudiments of statesaving to Sega (Mega) CD and Neo-Geo CD.

TheLostNinja:
I'm sure, many of MAME user for the years or even decades await a possibility to beat a CD version of first 2 Metal Slugs without of recaptured prisonners count loss. (S/M)ega CD also has a some "tough" arcades, like Bari-Arm, Devastator, Cyborg 009, (Dennin\Robo) Aleste, Batman Returns and Cadillacs'n'Dinosaurs, where statesaving if not necessary, then desireable at least.

PR #14901, opened 2026-01-31
A bit of saving inside Taito's Under Fire & Chase Bombers HW drivers.

TheLostNinja:
This light fix may supress a such side effect of plugin-assisted reload as an occasional involuntary shot at the crosshair's location. If you think that driver deserved a SUPPORT_SAVE flag, please add a corresponsive commit.

### DIFF
--- a/src/mame/taito/undrfire.cpp
+++ b/src/mame/taito/undrfire.cpp
@@ -311,7 +311,9 @@ void undrfire_state::video_start()
 
 	m_frame_counter = 0;
 
-	//save_item(NAME(m_dislayer));
+#ifdef MAME_DEBUG
+	save_item(NAME(m_dislayer));
+#endif
 }
 
 /***************************************************************


happppp:
What light fix, the savestates? It won't have effect on possibly gun reload glitches, unless you mean the glitch happens after loadstate.

TheLostNinja:
> What light fix, the savestates? It won't have effect on possibly gun reload glitches, unless you mean the glitch happens after loadstate.

They're just after loadstates.

happppp:
I will ask once more: Why do you use the CI compile, and not a local compile environment?

And do you actually test the application after? (in this case, the CI build) - Over here it was evident that you did not test the changes: https://github.com/mamedev/mame/pull/14827

TheLostNinja:
In the case of https://github.com/mamedev/mame/pull/14827 I tested the build made by MAMETeam official MSYS2 kit. Sometimes I trying to build thus a new own branches before post a PR when it's need, but GitHub mostly builds the emulator faster than MSYS2 on my PC even with all involved threads, and compilation reassignment to GitHub artifact maker aslo helps to avoid a PC work slowdowns.
Besides that, MSYS2 compilers is more "omnivore" than artifact maker, and sometimes the build of commit sucсessfully built locally beforehand may be broken on GitHub, so the success of local build is far from a guarantee that you made all right.

happppp:
On your own PC, it's not like you have to do a full (time consuming) compile after each change. Unlike the CI build. Look over here for example, 5 failed tries (full compiles for CI), it would be faster to do it yourself: https://github.com/mamedev/mame/pull/14861

I don't think your workflow and relying on github artifacts to test is a good idea.

TheLostNinja:
Looks like the reason why **m_dislayer** saving crashes a compilation process is not a variable format itself, but the same #ifdef MAME_DEBUG requirement as for declaration. :)

TheLostNinja:
Althought the involuntary shots after reload button pressing didn't disapper (now I noticed that they happens if to move a crosshair during reload - looks like the real reason is 3-shot burst of default game gun), I think that few more items won't be superfluous. Is save support really full now?

FlykeSpice:
> On your own PC, it's not like you have to do a full (time consuming) compile after each change. Unlike the CI build. Look over here for example, 5 failed tries (full compiles for CI), it would be faster to do it yourself: #14861
> 
> I don't think your workflow and relying on github artifacts to test is a good idea.

Well. sometimes your (potato) machine isn't even able to compile the full project at all, like I do. My 8gb pc would always hang up when it reached the point to compile the bespoke ``luaengine.cpp`` source file. 

In those cases I think using Github CI to compile and test your changes is warranted (ofc just dont open unready PR but use makeshift branches on your fork just to build the artifacts)

MooglyGuy:
Closing, as user was blocked from the project due to repeated refusal to learn from past feedback or address feedback on pull requests, instead spamming the project with more and more new pull requests with the same issues.

PR #14904, opened 2026-02-01
sound/cem3394.cpp: Using the new filter implementation in va_vcf.cpp.

m1macrophage:
This corrects the filter type to a 24dB/oct, and makes self-oscillation more reliable.

sound/va_vcf.cpp: Implemented (optional) gain compensation as resonance increases, since the CEM3394 does that.

In order to reuse the VA_LPF4  implementation, I made it possible to have a "streamless" VA_LPF4 . See documentation for `configure_streamless()` and related code. An alternative is to create a VA_LPF4_STREAMLESS device (open to other names), and then have both VA_LPF4 and CEM3394 use that. Let me know if the second option is preferable.

### DIFF
--- a/src/devices/sound/cem3394.cpp
+++ b/src/devices/sound/cem3394.cpp
@@ -306,7 +306,12 @@ void cem3394_device::sound_stream_update(sound_stream &stream)
 
 void cem3394_device::device_add_mconfig(machine_config &config)
 {
-	VA_LPF4(config, m_vcf).configure_drive(1.0).configure_bass_gain_comp(0.5);
+	VA_LPF4(config, m_vcf).configure_drive(1.0);
+
+	// According to the datasheet, the filter maintains the apparent loudness
+	// constant as resonance increases. The value below was selected by trial
+	// and error, to qualitatively match the preceding statement.
+	m_vcf->configure_bass_gain_comp(0.2);
 }
 
 void cem3394_device::device_start()


galibert:
The option I would have used is derivation actually, with a common filter class from which both VA_LPF4 and CEM3394, and possibly the other synth-in-a-chip devices like the CEM3396 would derive from.  The root class doesn't have to be a device_t derivative either, to allow inheriting (composing) from both it and other future band-limited oscillator and adsr base classes.

Or possibly a non-device_t object the devices would include.  Which would make multiple oscillator instances easier.  But would make sate saving a little more annoying...

What would you think about that approach?

m1macrophage:
Thanks, I gave it some thought.

A drawback of my current approach is  that `VA_LPF4` is too many things at once. Its interface changes depending on whether it is configured as "streamles". The "multiple base classes" approach you proposed doesn't suffer from that.

A drawback with multiple base classes that don't derive from `device_t` is that we lose functionality that comes "for free". State saving (as you pointed out), lifecycle management (`device_start()`, `device_reset()`, etc), functions such as `logerror()`, `logfatal()`, `machine().time()`, `machine().random()` etc. There are workarounds of course, but those base classes start looking device-like without being a `device_t`.

Considering all the above, I think a "best of both worlds" approach is for VA_LPF4, VA_VCO (future), etc., to just inherit from `device_t` (but **not** `device_sound_interface`). Specific filter and oscillator devices can just nest them. And CEM3394 and similar can nest multiple of those devices, as needed.

More concretely, CEM3394 will, for now, look similar to this PR. But there will be more functionality (e.g. VCO) moved out in the future. And VA_LPF4 will be split into two classes, with the "streaming" variant (subclass of `device_sound_interface`) nesting the streamless VA_LPF4.

Happy to try this out. And open to discussing more of course.

pmackinlay:
Implementing `device_interface` will give you most, if not all of the missing functionality you mention.

m1macrophage:
> [...] Or possibly a non-device_t object the devices would include. Which would make multiple oscillator instances easier. But would make sate saving a little more annoying...

I think I misinterpreted "include" to mean "inherit", due to the inheritance suggestion earlier on. But I believe the suggestion here is to include as an instance variables. I agree that composing as instance variables is preferable, and using 2 instances of the same class is a good reason why.

Is there a strong reason to avoid making those instance variable a `device_t` (`required_device`)? Assuming I make the LPF4 device more "lightweight", and remove the streaming functionality.


> Implementing device_interface will give you most, if not all of the missing functionality you mention.

Indeed, it seems to cover a lot. It looks like some stuff (save states) might require extra legwork, though maybe I missed something. Would it be preferable for LPF4 to inherit from `device_interface` rather than `device_t`, even if it is used as an instance variable in CEM3394?

galibert:
The difficulty with device_t is that you can't inherit multiple times from it, and diamond inheritance really doesn't work with it.  Also, you can't have a derived-from-device_t instance variable, everything device_t must be instanciated from a machine config or equivalent and appear in the device tree.

As for save state, as long as you have some device handy, it's not too bad.  Check tilemap_t::init_common for a non-device_interface, and device_execute_interface::interface_post_start for a device_interface (which is nicer).

Do you want to current version merged or are you going to change it?

m1macrophage:
> Do you want to current version merged or are you going to change it?

I would not object to a merge.

Regardless of the ideal end state, I think this is an improvement in sound accuracy and code reuse. 

As I work on the prophet5, I will add more VA devices (CEM3310, OTA-based EGs, band-limited VCO/CEM3340). Once the prophet5 is done, I can revisit this discussion, mock up a few options for how to reuse in CEM3394, and see what folks prefer.


> The difficulty with device_t is that you can't inherit multiple times from it, and diamond inheritance really doesn't work with it [...]

Yup, I understand. The options that use device_t would not allow for reuse-by-inheritance. 


> As for save state, as long as you have some device handy, it's not too bad. Check tilemap_t::init_common for a non-device_interface, and device_execute_interface::interface_post_start for a device_interface (which is nicer).

Cool, thanks for the concrete examples!

PR #14907, opened 2026-02-01
Foundin' a base for Exidy 440 System state saving.

TheLostNinja:
Would be especially useful for Crossbow. If somebody think that driver deserved a SUPPORT_SAVE flag after these changes, please add a corresponsive commit.


TheLostNinja:
If to save the **crossbow** at the map, the menu entries are disappearing. What items did I missed?

dinkc64:
I wonder if those buffers that get allocated in video_start() get saved automatically or not?

cuavas:
The banking needs to be changed to use configured banking and views.  The run-time tag lookups need to go, too.  There’s no point patching up code that needs to be thrown out wholesale.

PR #14908, opened 2026-02-01
bus/spectrum: Add Kempston Mouse Interface slot device

dotkoval:
* classic 2-button mouse interface
* available for classic ZX Spectrum models (48/128/+2/+3)

### DIFF
--- a/src/devices/bus/spectrum/kempmouse.cpp
+++ b/src/devices/bus/spectrum/kempmouse.cpp
@@ -14,29 +14,19 @@
 
 **********************************************************************/
 
-
 #include "emu.h"
 #include "kempmouse.h"
 
-//**************************************************************************
-//  DEVICE DEFINITIONS
-//**************************************************************************
-
 DEFINE_DEVICE_TYPE(SPECTRUM_KEMPMOUSE, spectrum_kempmouse_device, "spectrum_kempmouse", "Kempston Mouse Interface")
 
-
-//-------------------------------------------------
-//  INPUT_PORTS( kempmouse )
-//-------------------------------------------------
-
 static INPUT_PORTS_START( kempmouse )
-	PORT_START("mouse_x") /* 0xFBDF */
+	PORT_START("mouse_x")       // 0xFBDF
 	PORT_BIT(0xff, 0, IPT_MOUSE_X) PORT_SENSITIVITY(30)
 
-	PORT_START("mouse_y") /* 0xFFDF */
+	PORT_START("mouse_y")       // 0xFFDF
 	PORT_BIT(0xff, 0, IPT_MOUSE_Y) PORT_REVERSE PORT_SENSITIVITY(30)
 
-	PORT_START("mouse_buttons") /* 0xFADF */
+	PORT_START("mouse_buttons") // 0xFADF
 	// The right key was the primary key in most implementations in the 1980s–1990s.
 	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_BUTTON1) PORT_NAME("Mouse Button Right") PORT_CODE(MOUSECODE_BUTTON2)
 	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_BUTTON2) PORT_NAME("Mouse Button Left") PORT_CODE(MOUSECODE_BUTTON1)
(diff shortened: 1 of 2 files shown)


src/devices/bus/spectrum/kempmouse.cpp
```diff
@@ -0,0 +1,106 @@
...
+	PORT_START("mouse_x") /* 0xFBDF */
+	PORT_BIT(0xff, 0, IPT_MOUSE_X) PORT_SENSITIVITY(30)
+
+	PORT_START("mouse_y") /* 0xFFDF */
+	PORT_BIT(0xff, 0, IPT_MOUSE_Y) PORT_REVERSE PORT_SENSITIVITY(30)
+
+	PORT_START("mouse_buttons") /* 0xFADF */
+	// The right key was the primary key in most implementations in the 1980s–1990s.
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_BUTTON5) PORT_NAME("Mouse Button Right") PORT_CODE(MOUSECODE_BUTTON2)
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_BUTTON4) PORT_NAME("Mouse Button Left") PORT_CODE(MOUSECODE_BUTTON1)
```
angelosa: That doesn't mean it should map arbitrarily. It's `IPT_BUTTON1` / `IPT_BUTTON2`, numbers are relative to your device, leaving the user to assign inputs if the defaults aren't satisfactory.

holub: I think that mapping matches what I did before for the clones without exp slot.
I would prefer to keep them consistent and fix everywhere in one shot.
Let's not make him responsible for my mistakes.

src/devices/bus/spectrum/kempmouse.cpp
```diff
@@ -0,0 +1,106 @@
+// license:BSD-3-Clause
+// copyright-holders: Oleksandr Kovalchuk
+// Thanks to: Nigel Barnes (based on kempjoy implementation)
+/**********************************************************************
+
+    Kempston Mouse Interface (original, 2-buttons)
+
+**********************************************************************/
+
+// Original Kempston mouse interface description:
```
holub: Move up, inside header comment

src/devices/bus/spectrum/kempmouse.cpp
dotkoval: Makes sense! Thanks for pointing this out. I've updated it to use `IPT_BUTTON1`/`IPT_BUTTON2`. I kept the `PORT_CODE(MOUSECODE_...)` defaults to ensure correct 2-button behaviour, as otherwise `BUTTON2` maps to the middle mouse button by default.

src/devices/bus/spectrum/kempmouse.cpp
dotkoval: Rgr! Fixed in the latest commit :)

src/devices/bus/spectrum/kempmouse.cpp
```diff
@@ -0,0 +1,107 @@
+// license:BSD-3-Clause
+// copyright-holders: Oleksandr Kovalchuk
+// Thanks to: Nigel Barnes (based on kempjoy implementation)
```
Pernod70: No need for the Thanks, it's a very common device template.

dotkoval: Just to check - should I remove the "Thanks" line?
I think expressing appreciation is never a bad thing :) but if it doesn't match the usual template or project rules, I'll gladly remove it.

Pernod70: Yes, remove, it's thanking me though I haven't contributed to this file.

dotkoval: Got it! Done.

src/devices/bus/spectrum/kempmouse.cpp
```diff
@@ -0,0 +1,106 @@
...
+	PORT_START("mouse_buttons") /* 0xFADF */
+	// The right key was the primary key in most implementations in the 1980s–1990s.
+	PORT_BIT(0x01, IP_ACTIVE_LOW, IPT_BUTTON1) PORT_NAME("Mouse Button Right") PORT_CODE(MOUSECODE_BUTTON2)
+	PORT_BIT(0x02, IP_ACTIVE_LOW, IPT_BUTTON2) PORT_NAME("Mouse Button Left") PORT_CODE(MOUSECODE_BUTTON1)
+	PORT_BIT(0xfc, IP_ACTIVE_HIGH, IPT_UNUSED)
+INPUT_PORTS_END
+
+
+//-------------------------------------------------
+//  input_ports - device-specific input ports
```
holub: I think we don't like these mumbling comments anymore. Anybody confirm?

src/devices/bus/spectrum/kempmouse.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+DEFINE_DEVICE_TYPE(SPECTRUM_KEMPMOUSE, spectrum_kempmouse_device, "spectrum_kempmouse", "Kempston Mouse Interface")
+
+
+//-------------------------------------------------
+//  INPUT_PORTS( kempmouse )
+//-------------------------------------------------
+
+static INPUT_PORTS_START( kempmouse )
+	PORT_START("mouse_x") /* 0xFBDF */
```
holub: ...and block comments for single line

src/devices/bus/spectrum/kempmouse.cpp
galibert: I personally don't see the point to them

dotkoval: Cleaned up the comments according to your suggestions.

holub: Looks good for me now

PR #14911, opened 2026-02-02
tms320c2x: some optimizations

FlykeSpice:
This is a sucessor to my previous PR https://github.com/mamedev/mame/pull/14479 after @cuavas broke it with his renaming changes

Here the benchmark results from clang-x64 build (done with ``-bench 300``):
```
benchmarking victlapa...
254.33 -> 254.85 (0.52)
250.91 -> 254.92 (4.01)
251.28 -> 254.37 (3.09)
251.70 -> 254.06 (2.36)

benchmarking adillor...
148.92 -> 147.59 (-1.33)
148.62 -> 148.87 (0.25)
148.78 -> 149.23 (0.45)
146.78 -> 149.00 (2.22)

benchmarking alpinerd...
116.44 -> 115.98 (-0.46)
116.24 -> 115.92 (-0.32)
116.06 -> 116.39 (0.33)
115.86 -> 116.06 (0.20)

benchmarking cybrcomm...
165.43 -> 168.59 (3.16)
170.65 -> 172.03 (1.38)
170.15 -> 172.40 (2.25)
170.37 -> 171.26 (0.89)

benchmarking raverace...
139.77 -> 140.15 (0.38)
139.00 -> 140.47 (1.47)
138.32 -> 140.83 (2.51)
139.02 -> 139.45 (0.43)

benchmarking acedrive...
141.75 -> 141.79 (0.04)
140.69 -> 142.13 (1.44)
140.32 -> 141.50 (1.18)
141.10 -> 141.96 (0.86)

benchmarking winrun91...
244.38 -> 250.36 (5.98)
244.07 -> 249.12 (5.05)
244.93 -> 248.70 (3.77)
247.36 -> 250.57 (3.21)

benchmarking winrungp...
240.20 -> 240.21 (0.01)
234.34 -> 240.38 (6.04)
239.03 -> 240.08 (1.05)
237.82 -> 240.64 (2.82)
```
(The bechmarks may have been influenced by CPU throttling while running them, so they can't be 100% accurate)

### DIFF
--- a/src/devices/cpu/tms320c2x/tms320c2x.cpp
+++ b/src/devices/cpu/tms320c2x/tms320c2x.cpp
@@ -542,7 +542,6 @@ inline void tms320c2x_device::GETDATA(int shift, int signext)
 
 	/* next ARP */
 	if (m_opcode.b.l & 0x80) MODIFY_AR_ARP<IgnoreARPHack>();
-
 }
 
 inline void tms320c2x_device::PUTDATA(uint16_t data)
--- a/src/devices/cpu/tms320c2x/tms320c2x.cpp
+++ b/src/devices/cpu/tms320c2x/tms320c2x.cpp
@@ -312,7 +312,7 @@ void tms320c2x_device::dxr_w(uint16_t data)
 
 	if(TXM) {
 		if(FSM)
-			m_waiting_for_serial_frame = 1;
+			m_waiting_for_serial_frame = true;
 		else
 			m_IFR |= 0x20;
 	}
@@ -370,12 +370,14 @@ inline void tms320c2x_device::MODIFY_DP(int data)
 	m_STR0 |= (data & DP_REG);
 	m_STR0 |= 0x0400;
 }
+
 inline void tms320c2x_device::MODIFY_PM(int data)
 {
 	m_STR1 &= ~PM_REG;
 	m_STR1 |= (data & PM_REG);
 	m_STR1 |= m_fixed_STR1;
 }
+
 inline void tms320c2x_device::MODIFY_ARP(int data)
 {
 	m_STR1 &= ~ARB_REG;
@@ -386,12 +388,11 @@ inline void tms320c2x_device::MODIFY_ARP(int data)
 	m_STR0 |= 0x0400;
 }
 
-uint16_t tms320c2x_device::reverse_carry_add(uint16_t arg0, uint16_t arg1 )
+uint16_t tms320c2x_device::reverse_carry_add(uint16_t arg0, uint16_t arg1)
 {
 	uint16_t result = 0;
 	int carry = 0;
-	int count;
-	for (count = 0; count < 16; count++)
+	for (int count = 0; count < 16; count++)
 	{
 		int sum = (arg0 >> 15) + (arg1 >> 15) + carry;
 		result = (result << 1) | (sum & 1);
@@ -402,6 +403,7 @@ uint16_t tms320c2x_device::reverse_carry_add(uint16_t arg0, uint16_t arg1 )
 	return result;
 }
 
+template <bool IgnoreARPHack = false>
 inline void tms320c2x_device::MODIFY_AR_ARP()
 { /* modify address register referenced by ARP */
 	switch (m_opcode.b.l & 0x70)        /* Cases ordered by predicted useage */
@@ -437,7 +439,7 @@ inline void tms320c2x_device::MODIFY_AR_ARP()
 			break;
 	}
 
-	if (!m_mHackIgnoreARP)
+	if (!IgnoreARPHack)
 	{
 		if (m_opcode.b.l & 8)
 		{ /* bit 3 determines if new value is loaded into ARP */
@@ -484,7 +486,7 @@ inline void tms320c2x_device::CALCULATE_ADD_OVERFLOW(int32_t addval)
 
 inline void tms320c2x_device::CALCULATE_SUB_OVERFLOW(int32_t subval)
 {
-	if ((int32_t)((m_oldacc.d ^ subval) & (m_oldacc.d ^ m_ACC.d)) < 0)
+	if (int32_t((m_oldacc.d ^ subval) & (m_oldacc.d ^ m_ACC.d)) < 0)
 	{
 		SET0(OV_FLAG);
 		if (OVM)
@@ -497,13 +499,14 @@ inline void tms320c2x_device::CALCULATE_SUB_OVERFLOW(int32_t subval)
 inline uint16_t tms320c2x_device::POP_STACK()
 {
 	uint16_t const data = m_STACK[m_stack_limit];
-	for (unsigned i = m_stack_limit; 0 < i; --i)
+	for (unsigned i = m_stack_limit; i > 0; --i)
 		m_STACK[i] = m_STACK[i - 1];
 	return data;
 }
+
 inline void tms320c2x_device::PUSH_STACK(uint16_t data)
 {
-	for (unsigned i = 0; m_stack_limit > i; ++i)
+	for (unsigned i = 0; i < m_stack_limit; ++i)
 		m_STACK[i] = m_STACK[i + 1];
 	m_STACK[m_stack_limit] = data;
 }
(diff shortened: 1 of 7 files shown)


src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -497,13 +499,14 @@ inline void tms320c2x_device::CALCULATE_SUB_OVERFLOW(int32_t subval)
 inline uint16_t tms320c2x_device::POP_STACK()
 {
 	uint16_t const data = m_STACK[m_stack_limit];
-	for (unsigned i = m_stack_limit; 0 < i; --i)
+	for (unsigned i = m_stack_limit; i > 0; --i)
```
cuavas: The reason for putting these the other way around is that it makes an `=` vs `==` typo always cause an “assigning to rvalue” error.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -619,7 +614,7 @@ void tms320c2x_device::add()
 void tms320c2x_device::addc()
 {
 	m_oldacc.d = m_ACC.d;
-	GETDATA(0, 0);
+	GETDATA();
```
cuavas: I don’t think making this less explicit really helps.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -629,12 +624,12 @@ void tms320c2x_device::addc()
 void tms320c2x_device::addh()
 {
 	m_oldacc.d = m_ACC.d;
-	GETDATA(0, 0);
+	GETDATA();
 	m_ACC.w.h += m_ALU.w.l;
-	if ( (uint16_t)(m_oldacc.w.h) > (uint16_t)(m_ACC.w.h) ) {
+	if ((uint16_t)m_oldacc.w.h > (uint16_t)m_ACC.w.h) {
```
cuavas: If you’re going to be reformatting casts, it’s better to reformat them to C++ function syntax than the other way around – it’s less error-prone.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -853,19 +848,19 @@ void tms320c2x_device::cmpr()
 	switch (m_opcode.b.l & 3)
 	{
 		case 0:
-			if ( (uint16_t)(m_AR[ARP]) == (uint16_t)(m_AR[0]) ) SET1(TC_FLAG);
+			if (m_AR[ARP] == m_AR[0]) SET1(TC_FLAG);
```
cuavas: The casts here are to make it explicit that it’s doing an unsigned comparison.  They’re for the developer, not the compiler.  They don’t slow things down.  Please don’t make code less explicit for no reason.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -889,17 +884,17 @@ void tms320c2x_device::cnfp()  /** next two fetches need to use previous CNF val
...
 void tms320c2x_device::conf()
 {
-	// Disabled on TMS320C25
+	// Disabled on tms320c2x
 }
 
 void tms320c26_device::cnfd()
 {
-	// Disabled on TMS320C26
+	// Disabled on tms320c26
```
cuavas: The official names of the chips are uppercase.  Please don’t do this.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -1510,7 +1486,7 @@ void tms320c2x_device::tblr()
 		m_PFC = m_ACC.w.l;
 	}
 	m_ALU.w.l = m_cache.read_word(m_PFC);
-	if ( (CNF0) && ( (uint16_t)(m_PFC) >= 0xff00 ) ) {}   /** TMS320C25 only */
+	if (CNF0 && (uint16_t)m_PFC >= 0xff00) {}   /** TMS320C25 only */
```
cuavas: Once again don’t just make code less explicit – the parentheses around the comparison help *developer* parse the expression faster even if compilers don’t care.

src/devices/cpu/tms320c2x/tms320c2x.cpp
```diff
@@ -1843,70 +1813,68 @@ inline int tms320c2x_device::process_IRQs()
...
-			m_idle = 0;
+			m_idle = false;
 			m_IFR &= (~0x20);
 			SET0(INTM_FLAG);
-			return m_tms320c2x_irq_cycles;
+			return irq_cycles;
 		}
 	}
-	return m_tms320c2x_irq_cycles;
+	return 0;
```
cuavas: Are you sure this is safe?  what if the outer `if ( (INTM == 0) && (m_opcode.w.l != 0xce00) && (m_IFR & m_imr) )` passes but none of the inner tests pass?  Previously it would have returned `(3*CLK)`.  If that’s never supposed to happen, it should do something to make sure the developer know bad things have happened.

cuavas (commented):
So essentially, the optimisation is:

* Templating `GETDATA` to get rid of `m_mHackIgnoreARP` and one test for instructions that have the “next ARP” flag set.  This is probably fine, but impact would be minimal at best.
* Getting rid of `m_external_mem_access` and testing `m_memaccess` instead.  Are there potentially members of the family that would need a different test?
* Slightly shuffling the code for handling repeated instructions.  Probably fine, but once again probably very small impact.

src/devices/cpu/tms320c2x/tms320c2x.cpp
FlykeSpice: According to the datasheet, if an interrupt occured, one vector interrupt *must* be serviced, it should be impossible to fallthrough all the tests. Should I place a fatalerror right below indicating it reached a unreachable state?

src/devices/cpu/tms320c2x/tms320c2x.cpp
FlykeSpice: Got caught up during mechanical replacement, gonna revert it.

src/devices/cpu/tms320c2x/tms320c2x.cpp
FlykeSpice: The reason is those registers datatype are already unsigned 16-bit in the processor, so I thought it was being very reduntant. But sure I will revert it.

BenoitAdam:
Hey

did tms320c5x also broke somewhere because of the renaming ?

taito jc system games seems to launch but maybe there is some errors also, i'm not sure

here are some sheets from Texas Instruments : 

**TMS320C2x/C2xx/C5x
Optimizing C Compiler
User’s Guide**
https://www.ti.com/lit/ug/spru024e/spru024e.pdf

**TMS320c5x specific :** 

https://www.ti.com/lit/ds/symlink/tms320lbc53s.pdf

https://www.ti.com/lit/ds/sprs030a/sprs030a.pdf

FlykeSpice:
> Hey
> 
> did tms320c5x also broke somewhere because of the renaming ?
> 
> taito jc system games seems to launch but maybe there is some errors also, i'm not sure

What do you mean? No, it only "broke" my prev PR nothing else (because it was referencing the source file before the renaming).

If you getting errors it must be on your end, as I tested the games and they are working fine.

Anyway, it seems the Github Actions outage has been finally resolved and I'm gonna restart.

cuavas:
I cribbed the obviously safe parts of this in 2cabaf1541c070b9ba1b61ef98f2b12ed3e24a4f if you want to rebase.  I’m still not comfortable with removing the placeholder callbacks in the System 22 driver as it’s a reminder that they *should* be hooked up.

FlykeSpice:
> I cribbed the obviously safe parts of this in [2cabaf1](https://github.com/mamedev/mame/commit/2cabaf1541c070b9ba1b61ef98f2b12ed3e24a4f) if you want to rebase. I’m still not comfortable with removing the placeholder callbacks in the System 22 driver as it’s a reminder that they _should_ be hooked up.

Thanks, yeah that is why I just _commented_ them out rather than entirely remove the lines hooking up the callbacks as a reminder they are unused right now but they should probably be doing something now.

Should at least the callback functions definitions be restored but with their hooks commented out, and flagged with ``[[maybe_unused]]`` attribute?

PR #14912, opened 2026-02-02
epson/qx10.cpp: Mirror 2k rom across 8k address space

brijohn:
On a real QX-10 the 2k IPL ROM is mirrored across an 8k address range. 

This can be seen when using hdpart.com utility to partition a hard disk. It will check the installed rom size by doing a byte comparison between 0x000-0x7ff and 0x800-0xfff, which will be the same for a system running on a 2k rom.  This check is used to determine if they should show the "Change Bootable" menu item to allow toggle the boot flag on a partition.


brijohn:
Alright, i'll take a look at updating the banking in this as part of this fix.

brijohn:
Closing this in favor #14920

PR #14915, opened 2026-02-03
plugins/hiscore/hiscore.dat: add support to Double Point

TheLostNinja:
How can this game still not acquire the hiscore save support after 20+ years of indwelling in the MAME base?


angelosa:
hiscore.dat has no active maintainer (and actually surprised we have this in repository).

dinkc64:
With Leezer's site (hosted from now-defunct mameworld.info) gone, it's an unfortunate situation.

cracyc:
Added.

PR #14919, opened 2026-02-03
Update gaelco.cpp - change refresh rate on bigkrnk to meausred one

mamehaze:
gaelco.cpp - use measured screen hz for bigkarnk [Recreativas.org]

### DIFF
--- a/src/mame/gaelco/gaelco.cpp
+++ b/src/mame/gaelco/gaelco.cpp
@@ -864,7 +864,7 @@ void bigkarnk_state::bigkarnk(machine_config &config)
 
 	// Video hardware
 	screen_device &screen(SCREEN(config, "screen", SCREEN_TYPE_RASTER));
-	screen.set_refresh_hz(60);
+	screen.set_refresh_hz(58.74);
 	screen.set_vblank_time(ATTOSECONDS_IN_USEC(2500) /* not accurate */);
 	screen.set_size(32*16, 32*16);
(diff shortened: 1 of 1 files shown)


mamehaze:
Not sure if this is tied to any subtle protections as is the case in Thunder Hoop, but ClawGrip had Big Karnak measured by the folks at Recreativas.org and it's 58.74.

The protection definitely still fails sometimes in MAME with the current implementation and screen set to 60hz, so this may or may not improve the situation.

framerate

happppp:
Odd that it doesn't come close to thoop's measurement of 57.42Hz.
Can they measure hsync too? If you assume the common hsync of 15625Hz, then you get 15625/58.74 = 266.002 - (24MHz/4)/(384*266) = 58.74

ajrhacker:
> Odd that it doesn't come close to thoop's measurement of 57.42Hz.

That's because Thunder Hoop uses a different PCB, with no sound CPU or YM3812 and a 26 MHz XTAL instead of the 8 MHz XTAL.

happppp:
Maybe this then for those boards?: Again assuming 15625 hsynz, makes it possibly (26MHz/4)/(416*272) = 57.44

mamehaze:
that could work I think, there was a small range of acceptable values by the Thunder Hoop protection check code (which is flagged on the stage 2 boss, with the 'fail' flag check happening on either stage 4 or 5 when you die)

PR #14921, opened 2026-02-04
Rest of NIX, PlayMark and Sanritsu drivers files consolidation into single file.

TheLostNinja:
I think, since their header files is not cross-referred by any other driver and all four files of each driver is small enough, their splitness looks like a pure anachronism.


src/mame/nix/fitfight.cpp
```diff
@@ -84,14 +84,185 @@ Stephh's notes :
 */
 
 #include "emu.h"
-#include "fitfight.h"
+#include "emupal.h"
```
angelosa: > Can this at least be done by people who’ve read the reasons for organising #includes the way we do?
https://github.com/mamedev/mame/commit/d8179a687d1da77caf92a58796bc752c330e96fa#r176214886

i.e. `#include emu.h` at top, then devices finally MAME internals. Separated by one line each and A-Z sorted.

TheLostNinja: > > Can this at least be done by people who’ve read the reasons for organising #includes the way we do?
> > [d8179a6#r176214886](https://github.com/mamedev/mame/commit/d8179a687d1da77caf92a58796bc752c330e96fa#r176214886)
> 
> i.e. `#include emu.h` at top, then devices finally MAME internals. Separated by one line each and A-Z sorted.

No problem.

TheLostNinja:
I've hurried a bit about playmark.h - **powerball.cpp** uses it too.

src/mame/playmark/playmark.cpp
```diff
@@ -90,14 +90,403 @@ A |__     |_MCT1413P_|       |
...
+#define TILES_PER_PAGE_Y    (0x20)
+#define TILES_PER_PAGE_X    (0x20)
+#define PAGES_PER_TMAP_Y    (0x1)
+#define PAGES_PER_TMAP_X    (0x4)
+
+namespace {
+
+class playmark_base_state : public driver_device // base class for powerbal.cpp, too
+{
+public:
```
cuavas: This won’t work.  `playmark_base_state` needs to be in a header where playmark/powerbal.cpp can get it.  There’s even a comment there.  You can move `playmark_base_state` to playmark/playmark_base.h and playmark/playmark_base.cpp if you really want to, but it definitely can’t be made local like this.

cuavas (changes requested):

angelosa:
Before you go forward and fix the include guard: `pmcommon` is an unhappy name for a file, also genuinely not sure if there's a point in having this kind of merge if it _adds_ new files instead of reducing them (beyond the fact that those should be composed devices anyway).

cuavas (commented):
Only the `playmark_base_state` is shared.  `playmark_state` is not.  Putting `playmark_state` in the “common” source file is just making things confusing.

PR #14923, opened 2026-02-04
misc/4enraya.cpp: Add IDSA 'Spanish Darts' (not working)

clawgrip:
Spanish Darts [TobiSham]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -29121,6 +29121,7 @@ sprtdart
 4enrayaa
 bowlgum
 chicgum
+spadarts
 strker
 tourpgum
 unkpacg


src/mame/misc/4enraya.cpp
```diff
@@ -769,6 +785,61 @@ ROM_START( 4enrayaa )
...
+ROM_START( spadarts )
+	ROM_REGION( 0x10000, "maincpu", 0 )
+	ROM_LOAD( "spanish_darts_idsa_91_db98_27c256.bin", 0x0000, 0x8000, CRC(e47ecdfc) SHA1(8dbbd0e866c9fc044dc4687501944ce4966dc288) )
+	ROM_REGION( 0x6000, "chars", 0 )
+	ROM_LOAD( "spanish_darts_idsa_91-1_27c64.bin",     0x0000, 0x2000, CRC(727cc152) SHA1(cd5bb9427745d898546b7dba04161f3713d7e79f) )
+	ROM_LOAD( "spanish_darts_idsa_91-2_27c64.bin",     0x2000, 0x2000, CRC(148f1edf) SHA1(ea6873164d086058af86e5707468a434ef9ed0cc) )
+	ROM_LOAD( "spanish_darts_idsa_91-3_27c64.bin",     0x4000, 0x2000, CRC(7c1d8910) SHA1(2a7f26cc43b87a8d191ca6d3bd225a8d5e6d8a22) )
+
+	ROM_REGION( 0x0020,  "pal_prom", 0 )
+	ROM_LOAD( "n82s123n.bin",                          0x0000, 0x0020, BAD_DUMP CRC(dcbd2352) SHA1(ce72e84129ed1b455aaf648e1dfaa4333e7e7628) ) // Not dumped, loaned from 4enraya
```
angelosa, line 839: If the PROM is different then it should be a `NO_DUMP`, but the trouble here is that who knows what will happen without a specific machine config derivation. Does it work any better with any other config declared in this driver?

PR #14924, opened 2026-02-05
DIP switch sheet for mhhonban

Gently28:
add somebody to do under it……， ：-）


happppp:
Hi, since there is no code contribution, it is better placed here: https://github.com/mamedev/mame/issues
Also, could you make the topic more descriptive in the future? So instead of "Update niyanpai.cpp", write something like "DIP switch sheet for mhhonban"

cuavas:
Addressed in c67391436b3b9536d1d74b74a334d239a8a66349.

PR #14925, opened 2026-02-05
Three missed FM Towns family missing items saving.

TheLostNinja:
If somebody think that driver deserved a SUPPORT_SAVE flag after these changes, please add a corresponsive commit.


cuavas:
There were bigger issues with serial IRQ than that, see d24b0cf898d1759c746593af1c9d461705c5d6a8.  There are more issues preventing save state support, too.

PR #14928, opened 2026-02-06
raccoon: Add imperfect timing flag, dodghero: Add imperfect sound flag

JimCarlTay:
This is a scaled-down version of PR #14890 that focuses only on adding the Imperfect Timing flag to raccoon and the Imperfect Sound flag to dodghero.

Refer to commit messages for proof/justification.

### DIFF
--- a/src/mame/sigma/sigmab98.cpp
+++ b/src/mame/sigma/sigmab98.cpp
@@ -2394,18 +2394,18 @@ ROM_END
 ***************************************************************************/
 
 // Sigma Medal Games
-GAME( 1997, dodghero, 0,        dodghero, sigma_1b, sigmab98_state, empty_init,    ROT0, "Sigma",             "Minna Atsumare! Dodge Hero",           0 )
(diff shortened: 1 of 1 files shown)


cuavas:
I don’t think `raccoon` needs the imperfect timing flag.  It’s marked not working, and there’s a bunch of other stuff going wrong, e.g. whatever makes it generate excessive amounts of recompiled code if you disable single-instruction mode.  Not working means all bets are off

Imperfect sound isn’t the right flag for `dodghero`.  It’s not really a case of bad sound emulation – it’s bad timing that happens to cause one sound to play incorrectly, and could be throwing other stuff off as well.  Depending on how @ajrhacker, maybe multiple games on that platform should have the imperfect timing flag.

JimCarlTay:
So, I have two options here.

I can either revert the previous commits on this PR and add imperfect timing flags to most or all sigmab98 games, or I can close this and let ajrhacker do it.

If the former option is chosen, I will rename and sync this PR accordingly or make a new one with these changes.

JimCarlTay:
Closed because a new pull request by cam900 may cause conflicts with this one.

PR #14929, opened 2026-02-06
pc/play4texash.cpp: Additional notes; change name to be same as manual

einstein95:
Manual from https://www.recreativas.org/archivo/play4-texas-hold-em-sleic-manual-es.pdf

Software itself appears to have been last updated on 2011-10-18, unsure if I should change the year to 2011 in this case.


angelosa:
Manual also has 2009 dates, I'd just bump the year and add notes in case new dump(s) surfaces.

einstein95:
> Manual also has 2009 dates, I'd just bump the year and add notes in case new dump(s) surfaces.

Done

PR #14930, opened 2026-02-06
emu/debug/debugstate.cpp: Renamed debugger_cpu -> debugger_state (chore)

holub:


ajrhacker:
I don't see this as a clean solution, since `debugger_state` can be confused with `device_state_interface`.

holub:
Another option is `debugger_globals`

galibert:
I'm trying to decouple the debugger from the rest of the code to allow the debugger primitives (breakpoints, etc) to be accessible from the driver for RE.  That kind of naming change would collide with my changes, so it would be nice if it could be postponed a little, it's complex enough as it is.

holub:
Don't worry about it, PR is gone. I'm excited to see your next masterpiece.

PR #14932, opened 2026-02-07
Some missed PSX CD block items saving.

TheLostNinja:
Let's see, will the emu crashes with timer error messages after random state load be repeat.

### DIFF
--- a/src/devices/cpu/psx/mdec.cpp
+++ b/src/devices/cpu/psx/mdec.cpp
@@ -71,9 +71,6 @@ void psxmdec_device::device_start()
 	save_item(NAME(p_n_quantize_y));
 	save_item(NAME(p_n_quantize_uv));
 	save_item(NAME(p_n_cos));
-	save_item(NAME(p_n_quantize_y));
-	save_item(NAME(p_n_quantize_uv));
-	save_item(NAME(p_n_cos));
 	save_item(NAME(n_decoded));
 	save_item(NAME(n_offset));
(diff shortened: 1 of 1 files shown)


TheLostNinja:
Well, the MAME timer error crashes, the real scourge of savemilkers, had left behind finally. Now I'll look into PS1 components to estimate their states save completion - maybe, I'll have a success to eliminate the occasional "trails" (sometimes distorted and thus horrifying not less than enemies or locations) of previously played sample after state load in **revil** and its clones, where statesaving is very useful for ammo, medicals and ink preservation, and stucks of gameplay (althought the music still keeps to play) after random state load in **swep1jpb**.

JimCarlTay:
GitHub was down a few hours ago so that's the likely reason the latest build of this PR failed.

TheLostNinja:
> GitHub was down a few hours ago so that's the likely reason the latest build of this PR failed.

How to relaunch the artifact build?

JimCarlTay:
You can make a new commit if you want to address something that you need to fix here.

TheLostNinja:
The build failure is _error: '[m_sf/m_mac**x**]' was not declared in this scope_, althought **psx.cpp** refers **gte.h**.

cuavas:
I don’t know whether these members actually persist across timeslices, or if they're only valid temporarily during execution of a single instruction and hence should be excluded from saved states.  I have a feeling `m_berr` in the PSX CPU is the latter.  Someone more competent can look at this stuff.

PR #14934, opened 2026-02-07
DRC: Restore floating point mode across calls to C code

cuavas:
Since I brought this up in a discussion and got no reaction, creating a PR to force the issue.

UML has an instruction for changing the floating point rounding mode.  The back-ends all have ways to implement this.  The thing is, this can affects most floating point arithmetic.  If this isn't restored across calls to C functions, code will be running with the recompiled code's floating point environment rather than what would usually be the default floating point environment.  In the case of the debugger hook function, this can include MAME's rendering and UI code, as well as the debugger UI code.

There's also the risk of weird things happening if the C runtime library's floating point environment gets out of sync with the the host CPU's floating point control state, or code in a memory handler calling a maths library function that needs to temporarily change the host CPU's floating point control state and restores it to the state the C runtime library thinks it should be in, not the state the generated code thinks it left it in.

This makes the generated code for the AArch64, x86-64 and C back-ends restore the floating point environment across calls to C code.  This will obviously have some performance impact on emulated memory reads/writes and calls to C functions.

@galibert can you give an opinion on this?

### DIFF
--- a/src/devices/cpu/drcbearm64.cpp
+++ b/src/devices/cpu/drcbearm64.cpp
@@ -1224,6 +1224,7 @@ void drcbe_arm64::emit_narrow_memread(a64::Assembler &a, const be_parameter &add
 		a.lsr(REG_PARAM1.w(), REG_PARAM1.w(), FLAGS_REG.w());
 	else
 		a.lsr(REG_PARAM1, REG_PARAM1, FLAGS_REG);
+	a.mov(FLAGS_REG, a64::xzr);
 }
 
 void drcbe_arm64::emit_narrow_memwrite(a64::Assembler &a, const be_parameter &addrp, const parameter &spacesizep, const memory_accessors &accessors) const
@@ -1943,15 +1944,22 @@ void drcbe_arm64::op_debug(a64::Assembler &a, const uml::instruction &inst)
 
 		be_parameter pcp(*this, inst.param(0), PTYPE_MRI);
 
-		Label skip = a.new_label();
+		Label const skip = a.new_label();
 
 		emit_ldr_mem(a, temp, &m_device.machine().debug_flags);
 		a.tbz(temp, 1, skip); // DEBUG_FLAG_CALL_HOOK
+
+		emit_ldr_mem(a, SCRATCH_REG1, &m_near.saved_fpcr);
+		a.mrs(FLAGS_REG, a64::Predicate::SysReg::kFPCR); // flags are clobbered anyway - save FP control here
+		a.msr(a64::Predicate::SysReg::kFPCR, SCRATCH_REG1);
 
 		get_imm_relative(a, REG_PARAM1, m_debug_cpu_instruction_hook.obj);
 		mov_reg_param(a, 4, REG_PARAM2, pcp);
 
 		call_arm_addr(a, m_debug_cpu_instruction_hook.func);
+
+		a.msr(a64::Predicate::SysReg::kFPCR, FLAGS_REG);
+		a.mov(FLAGS_REG, a64::xzr);
 
 		a.bind(skip);
 	}
@@ -2235,9 +2243,16 @@ void drcbe_arm64::op_callc(a64::Assembler &a, const uml::instruction &inst)
 	Label skip;
 	emit_skip(a, inst.condition(), skip);
 
+	emit_ldr_mem(a, SCRATCH_REG1, &m_near.saved_fpcr);
+	a.mrs(FLAGS_REG, a64::Predicate::SysReg::kFPCR); // flags are clobbered anyway - save FP control here
+	a.msr(a64::Predicate::SysReg::kFPCR, SCRATCH_REG1);
+
 	get_imm_relative(a, REG_PARAM1, (uintptr_t)paramp.memory());
 	get_imm_relative(a, TEMP_REG1, (uintptr_t)funcp.cfunc());
 	a.blr(TEMP_REG1);
+
+	a.msr(a64::Predicate::SysReg::kFPCR, FLAGS_REG);
+	a.mov(FLAGS_REG, a64::xzr);
 
 	if (inst.condition() != uml::COND_ALWAYS)
 		a.bind(skip);
@@ -2693,6 +2708,11 @@ void drcbe_arm64::op_read(a64::Assembler &a, const uml::instruction &inst)
 	auto const &accessors = m_memory_accessors[spacesizep.space()];
 	bool const have_specific = (uintptr_t(nullptr) != accessors.specific.read.function) || accessors.specific.read.is_virtual;
 
+	emit_ldr_mem(a, SCRATCH_REG1, &m_near.saved_fpcr);
+	if (!have_specific || ((1 << spacesizep.size()) >= accessors.specific.native_bytes))
+		a.mrs(FLAGS_REG, a64::Predicate::SysReg::kFPCR); // flags are clobbered anyway - save FP control here
+	a.msr(a64::Predicate::SysReg::kFPCR, SCRATCH_REG1);
+
 	if (have_specific && ((1 << spacesizep.size()) == accessors.specific.native_bytes))
 	{
 		emit_memaccess_setup(a, addrp, accessors, accessors.specific.read);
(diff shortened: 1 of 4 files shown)


galibert:
I'd rather have the rounding mode in the UML instruction itself (e.g. stateless) and leave the management of the state, if it's a state (is it on mmx?), to the backend.

p1pkin:
change rounding mode per instruction may affect performance, at least in my past tests save/restore x86 MXCSR register per instruction or per block had a price.

@cuavas for sure restoring rounding mode when exit DRC or call external code is a right move. couple months ago I've faced issue caused by not restoring it (in demul) when calling Direct3D code, which seems relies on to have it not changed, especially when handling FP math edge cases, NaN values, etc.

cuavas:
> change rounding mode per instruction may affect performance, at least in my past tests save/restore x86 MXCSR register per instruction or per block had a price.

Yeah, it causes SSE operations to be serialised, so it's better not to do it too frequently.

> @cuavas for sure restoring rounding mode when exit DRC or call external code is a right move. couple months ago I've faced issue caused by not restoring it (in demul) when calling Direct3D code, which seems relies on to have it not changed, especially when handling FP math edge cases, NaN values, etc.

OK, that means we definitely need to save/restore around calling the debugger hook, as that can call into Direct3D.

cuavas:
> I'd rather have the rounding mode in the UML instruction itself (e.g. stateless) and leave the management of the state, if it's a state (is it on mmx?), to the backend.

We don't use MMX – single precision only, doesn't provide the operations wev need. Floating point rounding mode is a state on any CPU designed in the last 30 years. Most CPUs provide the four IEEE754 modes, or some subset of them (e.g. SHARC only provides to nearest and towards zero, not the other two).

You really think every floating point arithmetic instruction should have the rounding mode it wants encoded in it? Remember this affects just about every floating point arithmetic instruction, e.g. floating point add, when the exponent of the result is higher than the source and you lose precision. This would quadruple the number of floating point arithmetic instructions to implement, and require the back-end to do more analysis to determine when an expensive mode change is needed. At the very least, it would have to do a mode change on every emulated instruction sequence that performs floating point arithmetic, because a HASHJMP could get there from anywhere, and similarly it would have to in every helper function invoked with CALLH or EXH, and also after helper functions return.

That would also mean emulated CPUs that can control rounding mode would need to recompile all their code any time the emulated CPU's rounding mode changes. That's a big change for PPC and MIPS.

Do you really think that's practical, and the performance cost is worth it?

It also doesn't avoid the issue of C code potentially getting confused if the floating point mode is out of sync with what the C runtime library thinks it should be.

rb6502:
I'm with Vas, I don't love encoding the mode into the instruction.  Flushing the generated code on a rounding mode switch is probably not going to happen much in arcade games, but it likely would happen constantly when booting IRIX or Linux or whatever.  Midway Skins Game runs on Linux, and I think some other MIPS based games do as well.

I do think SETFMOD should check if we're already in that mode to avoid unnecessary switches, and similarly the switch when going out to a memory accessor should check if the emulated program has the same rounding mode that the C++ code had set on entry to the recompiled code.  That way we minimize unnecessary switches.

This also gives a stronger incentive to use memory system bypasses with the DRC where possible.  It might be useful if you could query the memory system and find out in advance if the read/write you're about to make is going straight to RAM/ROM or through a handler, but that lookup may well cost the same as setting the rounding mode in the first place.

cuavas:
> I do think SETFMOD should check if we're already in that mode to avoid unnecessary switches,

I could make the back-ends do that, but it means at least an extra load and dependent conditional branch on SETFMOD.  Do you think it’s worth it?  The generated code shouldn’t be using SETFMOD frequently anyway.

> and similarly the switch when going out to a memory accessor should check if the emulated program has the same rounding mode that the C++ code had set on entry to the recompiled code

That’s kind of complicated to do without making the cure worse than the disease.  I would have thought the cost of changing the mode there wouldn’t be too bad there because you’ve got the memory access call setup that won’t be using SSE anyway.  Or does `ldmxcsr` cause a complete pipleline stall on x86-64 CPUs these days?

rb6502:
> I could make the back-ends do that, but it means at least an extra load and dependent conditional branch on SETFMOD. Do you think it’s worth it? The generated code shouldn’t be using SETFMOD frequently anyway.

I can put some instrumentation on the drcbec version to see how often it's called and how many of those calls are redundant.  As far as the benefit, I assume a load and branch is going to be cheaper than serializing the SSE pipes even if the branch predictor gets it wrong.  That may not be true though.

> That’s kind of complicated to do without making the cure worse than the disease. I would have thought the cost of changing the mode there wouldn’t be too bad there because you’ve got the memory access call setup that won’t be using SSE anyway. Or does `ldmxcsr` cause a complete pipleline stall on x86-64 CPUs these days?

It's probably true that a mode switch there isn't going to be horrible.

galibert:
Yeah, I didn't realize rounding-as-a-mode was not something ia32-specific.  I'm surprised when you say we don't use mmx/sse, I thought it was the only option available on amd64.

Anyway, it probably wouldn't be too complicated to add a variant on the handlers in emumem that fiddle with the rouding mode when needed (e.g. on delegates handlers and taps) for drc use.  Which would entirely remove the cost for all normal memory accesses.  Adding a variant that also handled flags for instance didn't represent any particular problem.  The only complicated part is how the new handler can reach the appropriate drc state fast (it only has the space handy when created, but then the space has the cpu, so there's a series of layers to cross).

rb6502:
So I have some preliminary data from GitHub master.  With SETFMOD in drcbec printf'ing the current and new mode, every MIPS and PowerPC driver I tried calls it dozens of times a frame, hundreds in some cases, and in 99.9% of cases it's setting the mode that's already current.  My assumption is that SETFMOD comes from recompiled code; is there a boilerplate use of it that might be giving those results?

cuavas:
> Yeah, I didn't realize rounding-as-a-mode was not something ia32-specific. I'm surprised when you say we don't use mmx/sse, I thought it was the only option available on amd64.

We use SSE, but not MMX. They're very different things. You can still use x87 FP in x86-64 as well, but we don't.

> So I have some preliminary data from GitHub master. With SETFMOD in drcbec printf'ing the current and new mode, every MIPS and PowerPC driver I tried calls it dozens of times a frame, hundreds in some cases, and in 99.9% of cases it's setting the mode that's already current. My assumption is that SETFMOD comes from recompiled code; is there a boilerplate use of it that might be giving those results?

PowerPC and MIPS have a SETFMOD in their initial setup when jumping into recompiled code. It's part of setting up three desired virtual machine state. Really, any CPU that does FP arithmetic should be doing this, as the initial state on entry is undefined. So you're seeing one for each time it returns to the scheduler for whatever reason (end of timeslice, time expired, etc.) which all add up. That would easily explain dozens, or even hundreds for a system that needs tight sync between emulated CPUs.

Do MIPS and PowerPC unconditionally use SETFMOD when the emulated FP control register gets written? If the do, you'd see one for every guest OS context switch as well. That would add up quickly for \*NIX-like operating systems. On that case, I'd question whether the test for whether anything changed should be in the SETFMOD or in the generated code.

This also reminds me that UML has no other FP mode state, e.g. no denormal-to-zero state, and also no instructions for testing NaN, denormal, etc. which means all that stuff needs to be done using integer operations in the generated UML code.

cuavas:
> I'm surprised when you say we don't use mmx/sse, I thought it was the only option available on amd64.

Maybe you're thinking of AArch64 which drops ARM scalar FP unit (need to use NEON SIMD unit)? AMD didn't drop x87 or MMX, although they did make SSE2 a baseline feature.

happy6861707079:
> is there a boilerplate use of it that might be giving those results?

A compiler when doing a floating point to integer conversion will save the fp control register and likely be setting the mode to the value it just saved (if nothing else ever changes the mode), as the compiler needs to ensure consistent results from these conversions and not make any assumptions about the fp mode.

cuavas:
> > is there a boilerplate use of it that might be giving those results?
> 
> A compiler when doing a floating point to integer conversion will save the fp control register and likely be setting the mode to the value it just saved (if nothing else ever changes the mode), as the compiler needs to ensure consistent results from these conversions and not make any assumptions about the fp mode.

That's not universally true:
* SSE2 provides an instruction that performs FP to integer conversion rounding towards zero irrespective of the current mode so that `int(some_float)` can be implemented efficiently (i.e. not needing to change the mode before/after)
* AArch64 has four FP to integer conversion instructions, all of which ignore the current mode
* The C `lround` functions have different semantics to the IEEE754 "to nearest" mode, so they need to be implemented with FP arithmetic glue and can't just use conversion instructions

It's true that when using x87 FP, a C compiler needs to change the mode any time you intermingle FP to integer conversions with FP arithmetic, but that's just one of the many ways x87 sucks.

happy6861707079:
But with respect to the emulated MIPS and PowerPC architectures, the instruction sequences from the contemporary compilers will likely contain some conversions that will result in the fp mode being written with the same value that it already contains.

cuavas:
> But with respect to the emulated MIPS and PowerPC architectures, the instruction sequences from the contemporary compilers will likely contain some conversions that will result in the fp mode being written with the same value that it already contains.

Ah, right. PPC has `fctiwz`, but MIPS would probably be doing mode changes at least occasionally for that.

cuavas:
I’m going to merge this in the interests of, “first make it work, then you can worry about making it fast.”  I’ll look at options for reducing impact later.

cuavas:
I've started to look at the performance implications of changing the rounding mode, and the results are all over the place.  It seems to have a lot more impact on some emulated systems than others, and the impact also depends on the power profile selected on the host CPU.  I've added code to mitigate the performance impact to the 64-bit ARM backend which should be working as of f3b970688582c7b94615d88d2c1d9dc7130fb3ed.  I'll look at x86-64 when I get a chance.

cuavas:
f6c613d3ecaa25f95c178ec0724b5297c2882673 avoids some unnecessary changes to the SSE mode.  It gives a small but measurable improvement on an i7 (not as radical as the differences on a Snapdragon X).  I’ll have a look at the impact on memory system calls later.

cuavas:
I tried making the x86-64 back-end avoid restoring MXCSR across memory system calls if it doesn’t need to, and it was a wash at best, if not slightly worse.  I think it’s as good as it’ll get for now.

PR #14935, opened 2026-02-07
casio_rompack.xml: Added 11 entries

ICEknigh7:
Added new Japan Rom Pack dumps:

- Family Selection (RO-581) [Edward d-tech]
- DH Tunes & Rhythms (RO-582) [whc2001 (Ian Wang)]
- Pops 2 (RO-603P) [kosei_515]
- Kodomo no Uta Best 1 / SOUND KIDS Douyou 1 (RO-609K / RO-701) [whc2001 (Ian Wang)]
- Kodomo no Uta Best 2 / SOUND KIDS Douyou 2 (RO-610K / RO-702) [Edward d-tech]
- Kodomo no Uta Best 3 / SOUND KIDS Douyou 3 (RO-611K / RO-703) [whc2001 (Ian Wang)]
- Animation Themes 8 / SOUND KIDS Anime no Uta 1 (RO-613A / RO-705) [whc2001 (Ian Wang)]
- Animation Themes 9 / SOUND KIDS Anime no Uta 2 (RO-615A / RO-707) [kosei_515]
- Young Best Hits 13 (RO-617Y) [whc2001 (Ian Wang)]
- Fusion (RO-801F) [kosei_515]
- Standards (RO-803S) [kosei_515]

### DIFF
--- a/hash/casio_rompack.xml
+++ b/hash/casio_rompack.xml
@@ -16,7 +16,7 @@ license:CC0-1.0
 	==============================================================================================
 	JAPANESE ROM PACKS
 	==============================================================================================
-	RO-104		19??	Enka Vol.1 Karaoke Hit Shuu
+	RO-104		19??	Enka Vol.1 Karaoke Hit-shū
 	RO-105		19??	New Music 2
 	RO-107		19??	New Music 3
 	RO-109		1984	New Music 4
@@ -209,7 +209,7 @@ license:CC0-1.0
 		17. 中の島ブルース
 	-->
 	<software name="ro106" supported="no">
-		<description>Enka Vol.2 Karaoke Hit Shuu (RO-106)</description>
+		<description>Enka Vol.2 Karaoke Hit-shū (RO-106)</description>
 		<year>19??</year>
 		<publisher>Casio / Sight &amp; Sound International Inc.</publisher>
 		<info name="alt_title" value="演歌 Vol.2 カラオケヒット集"/>


ICEknigh7:
Reverted the tabs-turned-spaces which were I assume erroneously changed when fixing conflicts before merging my previous PR, so the current changes are not showing up as cleanly as they should, but it's easy to see what's changed and what's not anyway.

hash/casio_rompack.xml
```diff
@@ -211,26 +190,26 @@ license:CC0-1.0
...
+		12. 鴎という名の酒場
+		13. 水割り
+		14. そんな夕子にほれました
+		15. 襟裳岬
+		16. 浮草ぐらし
+		17. 中の島ブルース
 	-->
 	<software name="ro106" supported="no">
-		<description>Enka Vol.2 Karaoke Hit-shū (RO-106)</description>
+		<description>Enka Vol.2 Karaoke Hit Shuu (RO-106)</description>
```
angelosa: Revert this change please.

hash/casio_rompack.xml
```diff
@@ -16,123 +16,102 @@ license:CC0-1.0
...
+	RO-704		1990	Sound Kids, Douyou 4
+	RO-706		1990	Sound Kids, Hit Songs 1
+	RO-708		19??	Sound Kids, Hit Songs 2
+	RO-802E		1988	Enka 10
+	RO-804M		1989	Digital Horn Lessons
 	==============================================================================================
 	WORLD ROM PACKS
 	==============================================================================================
-	RO-202?             Movie Themes? (unreleased)
-	RO-203?             Rock On? (unreleased)
```
angelosa, line 96: I don't see these two in the XML list?

hash/casio_rompack.xml
```diff
@@ -16,123 +16,102 @@ license:CC0-1.0
...
-	RO-704      1990    Sound Kids, Nursery rhymes [4]
-	RO-705      1990    Sound Kids, Anime Songs [1]
-	RO-706      1990    Sound Kids, Hit Songs [1]
-	RO-707      19??    Sound Kids, Anime Songs [2]
-	RO-708      19??    Sound Kids, Hit Songs [2]
-	RO-801F     1988    Fusion
-	RO-802E     1988    Enka [10]
-	RO-803S     1988    Standards
-	RO-804M     1989    Introduction to Digital Horn, Digital Horn Lessons
+	RO-104		19??	Enka Vol.1 Karaoke Hit Shuu
```
angelosa: Same deal as below.

hash/casio_rompack.xml
ICEknigh7: I'm seeing that MAME spells Japanese long vowels by doubling them:
image

hash/casio_rompack.xml
ICEknigh7: They ended up just being conjectures. "Movie Themes" and "Rock On" appeared in some early catalogs and fans decided to fill the missing product codes with them, so not real info.

hash/casio_rompack.xml
ICEknigh7: MAME seems to use "uu" in place of "ū":
image

hash/casio_rompack.xml
ICEknigh7: Also, cartridges called "Rock On" and "All Time Movie Hits" were released later so it could be that they were rescheduled.

angelosa:
Given that table is "unstructured" what happens in git is that it can't tell what is an anchor and what not, leading to potential conflicts and an hard time when checking the diffs due of the resulting verticality.
XML comment block is not the right place for markdown table wannabe, it's not intended for that. What you could do (in a separate PR) is to encapsulate in a `<softwarelist>` root `notes`, like what wswan.xml do at the top which is probably the best compromise here:
https://github.com/mamedev/mame/blob/2c42ec323d1099912d4850f3676f2540650b14f0/hash/wswan.xml#L6-L14

hash/casio_rompack.xml
angelosa: I believe these are low hanging fruits being in `src` given that u8 support for `GAME` and friends is relatively recent, `hash` files uses several instances of ū

image

ICEknigh7: The current xml in this PR is consistent with other softlists and with the main MAME system list, and if I made this change I'd have to change all the double vowels in the softlist and god knows which other Japanese things which I am not an expert with. This is something that should be done whenever somebody decides to unify romanization across MAME, it's really not something for this PR.

ICEknigh7:
Literally only the Wonderswan softlist does that, so maybe that one should be made consistent with the rest...

ICEknigh7:
Ready to go.

PR #14939, opened 2026-02-08
video/tmap038.cpp: Fix notes

cam900:
Low byte in First word of tile RAM is used for hi bits of tile code: 8x8 only?


cuavas:
Since you’ve previously shown interest in the VRender0 chipset family, have you looked at src/mame/namco/namcos6.cpp?  You might be able to get it working.

angelosa:
namco/namcos6.cpp requires a NAND device that works.

PR #14940, opened 2026-02-08
Adds a callback mechanism for handling vector updates.  This allows

davepl:
This PR adds a simple callback mechanism to the vector rendering issue so that the LUA script can register for and receive vector events, such as moveTo, lineTo, etc.  

The only cost when there are no hooks active is a single null check.

The LUA engine can register for the various vector update events, and then display them on an HP vector scope, for example.

### DIFF
--- a/src/devices/video/vector.cpp
+++ b/src/devices/video/vector.cpp
@@ -61,7 +61,7 @@ float vector_options::s_beam_dot_size = 0.0f;
 float vector_options::s_beam_intensity_weight = 0.0f;
 
 namespace {
-vector_device::hook_callback s_hook_callback;
+    vector_device::hook_callback s_hook_callback;
 }
 
 void vector_options::init(emu_options& options)


MooglyGuy:
Looks nice and tidy to me. I'd like to get input from @galibert since he's the resident `screen_device` wonk, but if he doesn't have any issues, I'll be happy to approve this.

galibert:
The vector rendering interface should include timings, and currently doesn't.  I would find it annoying to have that lack enshrined in the lua interface.

davepl:
They’re not required or used by this project, and I balk at adding unused functionality.  I agree timings could be useful, but without a consumer, I think they should be added when needed.

Unless those timings should be passed in band with the current vector callbacks, then I could see the value in adding them up front. Let me know specifically what you mean by timings and what you’re looking for.

- Dave

cuavas:
We can’t keep adding direct src/frontend -> src/devices dependencies on specific devices.  It prevents excluding them from subtargets.  It also won’t scale with debugger and Lua bindings, as it would mean instantiating all the support stuff for every device with bindings, whether they’re used in a session or not.  It needs some kind of mechanism for allowing a device type to register for an aspect when it’s first encountered in a live session (i.e. not just instantiating a device for `-validate`, `-romident`, or to display general info for the system selection menu).

MooglyGuy:
@galibert The vector rendering interface *should* have timings, but it doesn't, and that's irrelevant when it comes to showing off MAME's infrastructure in the context of a flashy demo that utilizes X/Y scopes and vector hardware. DaveP here literally just attached a screenshot of this PR running perfectly fine on a vector scope.

It's not going to cover all possible use cases of all possible vector output devices, but there's nothing about the implementation that can't be extended going forward once per-vector timing has been put into the core.

@cuavas I understand where you're coming from when it comes to wanting a perfectly clean architecture, but let's stay in our respective lanes. At the end of the day, when I end up implementing a renderer abstraction layer that targets DX9-12, Vulkan, OGL and Metal, I'm the one who's going to have to clean up all of this screen_device and associated stuff, not you. You're edge-case-blocking some good functionality that can easily be carried forward.

If either of you guys want to revert my approval of this PR, then do so at your own peril, but there's nothing about this that adds a notable amount of tech debt, and it comes with the side benefit of interfacing MAME's current codebase against legacy vector hardware. Don't let perfect be the enemy of perfectly good.

Issue #4904, opened 2019-04-16
Register Value Reset After Assigning Value


Abystus:
![RegisterSetFail](https://user-images.githubusercontent.com/1529049/56174783-dba31f80-5fb9-11e9-9847-0f0281fb1a87.gif)

You'll notice that V0 is loaded with 08, I then set it to 01 and step. Upon stepping, the register value will reset to the original value (08) before being set. This has been an for several versions now, and is very easy to reproduce across several different processors.

MooglyGuy:
This is a known issue with all cores that use the dynamic recompiler (DRC) system. Other than MIPS-based processors, which use the DRC system, what other processors do you experience this issue with?

Abystus:
> This is a known issue with all cores that use the dynamic recompiler (DRC) system. Other than MIPS-based processors, which use the DRC system, what other processors do you experience this issue with?

This has varied from build to build. I've had issues with it occurring on TMS34010, TMS32032, and MIPS, though at this moment I can only get it to reproduce on MIPS. The funny thing is, in the 32 bit version of MAME (same build version as the 64 bit one) you can set register values in recompiled processors like MIPS as you can see below:

![RegisterSetWork](https://user-images.githubusercontent.com/1529049/56179378-19f50a80-5fcb-11e9-8cc0-c2c39a9ce507.gif)

tedgreen99:
The MIPS3 DRC uses it's own copy of the registers called fast iregs which is why changing the debugger state variables doesn't update the actual used variables.
One simple solution would be to add a load_fast_iregs call after line 1230 in mips3drc.cpp, although this may have some performance implications.  It would be better if the debugger could somehow signal that something has been modified and the copy needs to be done.

> 
	/* if we are debugging, call the debugger */
	if ((machine().debug_flags & DEBUG_FLAG_ENABLED) != 0)
	{
		UML_MOV(block, mem(&m_core->pc), desc->pc);                              // mov     [pc],desc->pc
		save_fast_iregs(block);
		UML_DEBUG(block, desc->pc);                                         // debug   desc->pc
		load_fast_iregs(block);
	}

PR #14944, opened 2026-02-10
cpu/mips3: Fix DRC debugger interaction for register and memory modifications

nekuz0r:
# Summary

When debugging MIPS3 code running under the DRC, two things were broken: changing a register value in the debugger had no effect (the DRC kept using its own cached copy), and patching instructions in memory didn't do anything either since the DRC just kept running its already-translated code.

# Details

The DRC maps hot registers to internal UML integer registers for performance. The problem is that the debugger writes to the memory-backed register array, not those internal registers — so any change made during a breakpoint was immediately overwritten when execution resumed. The old code knew about this and even logged an error message.

Similarly, there was no mechanism to tell the DRC that memory contents had changed. If you patched an opcode through the debugger, the DRC would happily keep running the stale translated block.

**Register reload**: A `m_drc_iregs_dirty` flag is set in `state_import` when the debugger writes to a register that lives in a fast UML ireg. After each `UML_DEBUG` hook, the generated code tests this flag and reloads the internal registers from the memory-backed array when set. The flag is declared as `uint32_t` rather than `uint8_t` because `UML_MOV`/`UML_TEST` operate on 32 bits wide data, using a smaller type would corrupt the adjacent `m_fpmode` table.

**Cache invalidation**: `device_debug_setup` hooks the symbol table's `set_memory_modified_func` callback to set `m_drc_cache_dirty`, which is checked at the top of `execute_run` and triggers a full cache flush before the next execution slice.

# Related issue

https://github.com/mamedev/mame/issues/4904

### DIFF
--- a/src/devices/cpu/mips/mips3.cpp
+++ b/src/devices/cpu/mips/mips3.cpp
@@ -17,6 +17,8 @@
 
 #include "emuopts.h"
 
+#include "debug/express.h"
+
 #include <cmath>
 
 #define ENABLE_OVERFLOWS            (0)
@@ -180,6 +182,7 @@ mips3_device::mips3_device(const machine_config &mconfig, device_type type, cons
 	, m_drcfe(nullptr)
 	, m_drcoptions(0)
 	, m_drc_cache_dirty(0)
+	, m_drc_iregs_dirty(0)
 	, m_entry(nullptr)
 	, m_nocode(nullptr)
 	, m_out_of_cycles(nullptr)
@@ -237,6 +240,14 @@ void mips3_device::device_stop()
 	{
 		m_drcuml = nullptr;
 	}
+}
+
+void mips3_device::device_debug_setup()
+{
+	if (!m_isdrc) return;
+	debug()->symtable().set_memory_modified_func(
+		[this]() { m_drc_cache_dirty = true; }
+	);
 }
 
 /***************************************************************************
@@ -826,7 +837,7 @@ void mips3_device::state_import(const device_state_entry &entry)
 		// this refers to HI as R32 and LO as R33 because I'm lazy
 		const unsigned regnum = entry.index() - MIPS3_R0;
 		if (m_regmap[regnum].is_int_register())
-			logerror("debugger R%u = %08X, must update UML I%u\n", regnum, m_core->r[regnum], m_regmap[regnum].ireg() - uml::REG_I0);
+			m_drc_iregs_dirty = true;
 	}
 }
 
--- a/src/devices/cpu/mips/mips3drc.cpp
+++ b/src/devices/cpu/mips/mips3drc.cpp
@@ -38,6 +38,7 @@
 #define DEBUG_STRICT_VERIFY 0
 
 
+
 /***************************************************************************
     MACROS
 ***************************************************************************/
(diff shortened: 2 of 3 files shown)


nekuz0r:
Anybody to review ?
I have been using this modification for weeks while working on my Killer Instinct custom boot rom, no issues so far.

src/devices/cpu/mips/mips3.cpp
```diff
@@ -826,7 +837,7 @@ void mips3_device::state_import(const device_state_entry &entry)
 		// this refers to HI as R32 and LO as R33 because I'm lazy
 		const unsigned regnum = entry.index() - MIPS3_R0;
 		if (m_regmap[regnum].is_int_register())
-			logerror("debugger R%u = %08X, must update UML I%u\n", regnum, m_core->r[regnum], m_regmap[regnum].ireg() - uml::REG_I0);
+			m_drc_iregs_dirty = true;
```
rb6502, line 840: The rest of this file uses Allman brace style, so this should too for consistency.

src/devices/cpu/mips/mips3.cpp
```diff
@@ -239,6 +242,14 @@ void mips3_device::device_stop()
 	}
 }
 
+void mips3_device::device_debug_setup()
+{
+	if (!m_isdrc) return;
```
rb6502, line 247: This also should be Allman style.

src/devices/cpu/mips/mips3drc.cpp
```diff
@@ -1366,9 +1367,15 @@ void mips3_device::generate_sequence_instruction(drcuml_block &block, compiler_s
...
+		uml::code_label skip_reload = compiler.labelnum++;
 		UML_MOV(block, mem(&m_core->pc), desc->pc);                              // mov     [pc],desc->pc
 		save_fast_iregs(block);
-		UML_DEBUG(block, desc->pc);                                         // debug   desc->pc
+		UML_DEBUG(block, desc->pc);                                              // debug   desc->pc
+		UML_TEST(block, mem(&m_drc_iregs_dirty), 1);                        // test    [debugger_iregs_dirty],1
+		UML_JMPc(block, COND_Z, skip_reload);                                    // jmp     skip_reload,Z
+		UML_MOV(block, mem(&m_drc_iregs_dirty), 0);                         // mov     [debugger_iregs_dirty],0
+		load_fast_iregs(block);                                                  // <load fast integer registers>
+		UML_LABEL(block, skip_reload);                                           // skip_reload:
```
cuavas, line 1378: Considering how frequently this is going to occur, would it be better to turn it into a helper function to keep generated code size down?

src/devices/cpu/mips/mips3.cpp
```diff
@@ -239,6 +242,14 @@ void mips3_device::device_stop()
...
+	debug()->symtable().set_memory_modified_func(
+		[this]() { m_drc_cache_dirty = true; }
+	);
```
cuavas, line 250: You don't want to be doing this for every memory modification from the debugger - it's too expensive when most of the time, people will just be modifying data memory.

nekuz0r: The cache invalidation is triggered when editing the ':maincpu' program space, not through any other spaces/regions.
What is your suggestion ?

src/devices/cpu/mips/mips3.cpp
nekuz0r: There's quite a few one line if/else statements in the file without braces, i kept it as it, should i update it and update the other occurrences or leave it ?

src/devices/cpu/mips/mips3drc.cpp
nekuz0r: Not sure to understand what's expected to be done, any guidance ?

PR #14948, opened 2026-02-11
toaplan/toaplan_dsp.cpp: Make common toaplan DSP interface to device, Minor cleanups:

cam900:
toaplan/toaplan1.cpp:
- Reduce duplicates
- Improve logging infos
- Fix DSP ROM region size

toaplan/twincobr.cpp:
- Improve logging infos
- Fix spacings
- Make fsharkbt_dsp_* functions to fsharkbt set only
- Fix DSP ROM region size

### DIFF
--- a/src/mame/toaplan/toaplan1.cpp
+++ b/src/mame/toaplan/toaplan1.cpp
@@ -617,37 +617,37 @@ Stephh's and AWJ's notes (based on the games M68000 and Z80 code and some tests)
 
 /***************************** 68000 Memory Map *****************************/
 
-void toaplan1_state::bcu_map(address_map &map, offs_t offset)
+void toaplan1_state::bcu_map(address_map &map)
 {
-	map(offset + 0x01, offset + 0x01).w(FUNC(toaplan1_state::bcu_flipscreen_w));
-	map(offset + 0x02, offset + 0x03).rw(FUNC(toaplan1_state::tileram_offs_r), FUNC(toaplan1_state::tileram_offs_w));
-	map(offset + 0x04, offset + 0x07).rw(FUNC(toaplan1_state::tileram_r), FUNC(toaplan1_state::tileram_w));
-	map(offset + 0x10, offset + 0x1f).rw(FUNC(toaplan1_state::scroll_regs_r), FUNC(toaplan1_state::scroll_regs_w));
+	map(0x01, 0x01).w(FUNC(toaplan1_state::bcu_flipscreen_w));
+	map(0x02, 0x03).rw(FUNC(toaplan1_state::tileram_offs_r), FUNC(toaplan1_state::tileram_offs_w));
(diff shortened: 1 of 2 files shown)


src/mame/toaplan/toaplan1.cpp
```diff
@@ -617,22 +617,47 @@ Stephh's and AWJ's notes (based on the games M68000 and Z80 code and some tests)
...
-	map(0x100002, 0x100003).rw(FUNC(toaplan1_rallybik_state::tileram_offs_r), FUNC(toaplan1_rallybik_state::tileram_offs_w));
-	map(0x100004, 0x100007).rw(FUNC(toaplan1_rallybik_state::tileram_r), FUNC(toaplan1_rallybik_state::tileram_w));
-	map(0x100010, 0x10001f).rw(FUNC(toaplan1_rallybik_state::scroll_regs_r), FUNC(toaplan1_rallybik_state::scroll_regs_w));
-	map(0x140000, 0x140001).portr("VBLANK");
-//  map(0x140000, 0x140001).w(?? video frame related ??)
-	map(0x140003, 0x140003).w(FUNC(toaplan1_rallybik_state::intenable_w));
-	map(0x140008, 0x14000f).w(FUNC(toaplan1_rallybik_state::bcu_control_w));
-	map(0x144000, 0x1447ff).ram().w(FUNC(toaplan1_rallybik_state::bgpalette_w)).share(m_bgpaletteram);
-	map(0x146000, 0x1467ff).ram().w(FUNC(toaplan1_rallybik_state::fgpalette_w)).share(m_fgpaletteram);
+	bcu_map(map, 0x100000);
```
angelosa: Ugly.

`map(0x100000, end_offset).m(*this, FUNC(toaplan_state::bcu_map));`

Which incidentally also means you don't have to `offset` inside.

cam900: Fixed in recent commit.

src/mame/toaplan/twincobr.cpp
```diff
@@ -1479,19 +1462,25 @@ ROM_START( gulfwar2a )
 	ROM_LOAD( "82s123.b24", 0x240, 0x020, CRC(4fb5df2a) SHA1(506ef2c8e4cf45c256d6831a0a5760732f2de422) )    // Tile to sprite priority ??
 ROM_END
 
+void twincobr_state::init_fsharkbt()
+{
+	m_dsp->dsp().space(AS_IO).install_readwrite_handler(2, 2,
+		read16smo_delegate(*this, FUNC(twincobr_state::fsharkbt_dsp_r)),
+		write16smo_delegate(*this, FUNC(twincobr_state::fsharkbt_dsp_w)));
+}
```
cuavas, line 1470: It feels kind of ugly to have to patch in something like this when we’re usually trying to reduce the number of these init functions that we need.

PR #14950, opened 2026-02-12
emu/debug/debugcpu.cpp: Prefixed symbol names in global space with owner devices path in order to minimize conflicts

holub:


holub:
Vas, this is regarding your previous concern that new debugger symbols participants potentially can bring name conflicts.
If nobody interested, I'll drop the PR.

PR #14955, opened 2026-02-13
sinclair/glukrs.h: Fixed region size

holub:


angelosa:
`-validate` failure looks a CI env issue

PR #14956, opened 2026-02-13
input/input_sdl3.cpp: Fixed assert on new instance of joystick connected

holub:


holub:
Otherwise if joystick attached it failed with:
`mamed: ../../../../../src/osd/modules/input/input_sdl3.cpp:991: void osd::(anonymous namespace)::sdl_joystick_device_base::set_instance(SDL_JoystickID): Assertion '0 > m_instance' failed.`

PR #14962, opened 2026-02-14
Renaming of gumbo.cpp driver.

TheLostNinja:
From now it will wear the name of company whose only hardware architecture is described by this driver.


cuavas:
This is pointless churn.

PR #14945, opened 2026-02-10
tk2000: fix cassette and page2

arekkusu42:
(side-effects of my ongoing tinkering with apple2video timing...)

* cassette input is [documented as bit7 of C010](http://www.apple-iigs.info/doc/fichiers/TK2000%20II%20Manual%20teecnico.pdf#page=75) KBIN (not C06X like apple2p)
* HGR page2 is [documented as A000-BFFF](http://www.apple-iigs.info/doc/fichiers/TK2000%20II%20Manual%20teecnico.pdf#page=97) (not 4000-5FFF like apple2p)

This is enough to get various software working in tk2000 and mpf2 via `LOADT` of cassette audio ("TK2000 GAME SERVER"):
mpf2_BOLO

tk2000_Karateka

tk2000_Minotauro

tk2000_Suicida

PR #14964, opened 2026-02-14
tk2000: promote to WORKING

arekkusu42:
After PR #14945, it looks like the remaining TODOs are:

* emulate the expansion connector
	In addition to the original [Microdigital](https://www.facebook.com/groups/5251478676/posts/10159483337733677/) [disk interface card](https://www.facebook.com/groups/5251478676/posts/10161275857838677/), one expansion is the "[Tomato Board](https://web.archive.org/web/20140704015358/http://www.victortrucco.com/apple/tomatoboard/tomatoboard.asp)" which converts to a standard Apple II expansion slot, which can then take a DiskII controller (and [compatible SD card reader](https://www.youtube.com/watch?v=IhGf7qAi5K0))

* emulate the joystick port
	There was a "[TK Stick](http://www.tk90x.com.br/Perifericos_TK2000.html)"; this port is different from Apple II gameio and is wired [directly to KBIN and KBOUT](https://mirrors.apple2.org.za/ftp.apple.asimov.net/documentation/hardware/machines/TK2000%20II-Manual%20teecnico.pdf#page=75) (effectively, a second small keyboard.)

* emulate cassette MOTOR softswitch
	It appears the firmware does not automatically trigger either MOTOR on LOADT or LOADA, but some games (i.e. BOLO, Estacao Orbital) trigger MOTOR B after loading from tape, presumably to rewind? 

* emulate dual RESET keys
	Both keys need to be [pressed simultaneously to reset](https://mirrors.apple2.org.za/ftp.apple.asimov.net/documentation/hardware/machines/TK2000%20II-Manual%20de%20operacao.pdf#page=29).  The keyboard seems otherwise fully functional, including graphical chars (via CTRL-B, then SHIFT-char or SHIFT-CTRL-char)

* emulate the color/bw softswitch
	The firmware triggers this on TEXT/GR/HGR.  This is trivially implemented, thanks to the [existing CEC code](https://github.com/mamedev/mame/blob/7912c408e44a5602c6a7270719d403fc0c465b72/src/mame/apple/apple2video.cpp#L646).  It should *not* be beam race-able, since the color-killer circuit needs a (probably-CRT-specific) number of scanlines to take effect; "within a frame" seems accurate enough.
	
* correct PAL colors
	Brazil uses [PAL-M](https://en.wikipedia.org/wiki/PAL-M) (NTSC-60 but PAL coloration), and the firmware remaps GR colors to the possible HGR colors.  The tk2000 manual [describes this remapping](https://mirrors.apple2.org.za/ftp.apple.asimov.net/documentation/hardware/machines/TK2000%20II-Manual%20de%20operacao.pdf#page=135), but it does not match the current results in MAME using normal NTSC coloration. [This video](
https://youtu.be/0v8_qkCALFM?t=1624) shows real hardware coloring (assuming the user has a matching PAL-M TV?)

	The mpf2 was sold in [multiple regions](https://en.wikipedia.org/wiki/Microprofessor_II) (not Brazil), but there is some evidence that [it also shows PAL coloration](
https://www.nightfallcrew.com/30/05/2015/multitech-micro-professor-mpf-ii-boxed/).

Expected:
```
0       black   preta
1       blue    azul
2       green   verde
3       white   branca
4       blue    azul
5       red     vermelho
6       cyan    cyan
7       white   branca
8       green   verde
9       blue    azul
10      green   verde
11      white   branca
12      white   branca
13      red     vermelho
14      cyan    cyan
15      white   branca
```
Emulated:
tk2000_GR_colors

<br>
This commit implements the color/bw switch, adds TODO notes, and then since none of the remaining issues are particularly blocking users from playing games, removes MACHINE_NOT_WORKING.  The incorrect coloration can undoubtedly be corrected in an HLSL post-processing chain, however with the default SW presentation, I think it warrants MACHINE_IMPERFECT_GRAPHICS.  (Save states also appear to work OK.)

### DIFF
--- a/src/mame/apple/tk2000.cpp
+++ b/src/mame/apple/tk2000.cpp
@@ -619,5 +619,5 @@ ROM_END
 
 
 /*    YEAR  NAME    PARENT  COMPAT  MACHINE  INPUT   CLASS         INIT        COMPANY         FULLNAME */
-COMP( 1984, tk2000, 0,      0,      tk2000,  tk2000, tk2000_state, empty_init, "Microdigital", "TK2000 Color Computer", MACHINE_IMPERFECT_GRAPHICS | MACHINE_SUPPORTS_SAVE )
-COMP( 1982, mpf2,   tk2000, 0,      tk2000,  tk2000, tk2000_state, empty_init, "Multitech",    "Microprofessor II",     MACHINE_IMPERFECT_GRAPHICS | MACHINE_SUPPORTS_SAVE )
+COMP( 1984, tk2000, 0,      0,      tk2000,  tk2000, tk2000_state, empty_init, "Microdigital", "TK2000 Color Computer", MACHINE_IMPERFECT_COLORS | MACHINE_SUPPORTS_SAVE )
+COMP( 1982, mpf2,   tk2000, 0,      tk2000,  tk2000, tk2000_state, empty_init, "Multitech",    "Microprofessor II",     MACHINE_IMPERFECT_COLORS | MACHINE_SUPPORTS_SAVE )


arekkusu42:
Whoops, MACHINE_IMPERFECT_COLORS is more appropriate.

arekkusu42:
One more issue is that the XTAL looks wrong.  Photos ([one](https://amxprojects.com//wp-content/upload/2025/09/image-27.png), [two](https://www.1000bit.it/lista/m/microdigital/tk2000_2/tk-2000%20(5).jpg)) show it to be 14.30244 MHz which is slightly different from the Apple II EuroPlus.  But then, all of the Apple II clocks are slightly wrong in MAME...

PR #14966, opened 2026-02-15
shangkid.cpp driver files consolidation into single file.

TheLostNinja:
Also, looks like Taiyo had an experience of manufacturing only this hardware.


cuavas:
This doesn’t compile, it introduces pointless churn, and you still haven’t read and/or understood the C++ contributing guidelines.

PR #14970, opened 2026-02-16
tlcs900: Add TMP94C241 LDC control register mappings for DMA

felipesanches:
## Summary

- Add TMP94C241 DMA control register encodings to the LDC instruction handler in `900tbl.hxx`
- Refactor disassembler CR register names from hardcoded switches into per-variant symbol tables
- No changes to existing TMP96C141/TMP95C061/TMP95C063 behavior

## Problem

The `LDC` (Load Control Register) instruction uses an immediate byte to select which internal register to access. The TMP94C241 places its DMA registers at different offsets than the TMP96C141/TMP95C061/TMP95C063:

| Register | TMP96C141/TMP95C061/TMP95C063 | TMP94C241 |
|----------|-------------------------------|-----------|
| DMAM0-3 (8-bit) | 0x22, 0x26, 0x2A, 0x2E | 0x42, 0x46, 0x4A, 0x4E |
| DMAC0-3 (16-bit) | 0x20, 0x24, 0x28, 0x2C | 0x40, 0x44, 0x48, 0x4C |
| DMAD0-3 (32-bit) | 0x10, 0x14, 0x18, 0x1C | 0x20, 0x24, 0x28, 0x2C |
| DMAS0-3 (32-bit) | 0x00, 0x04, 0x08, 0x0C | 0x00, 0x04, 0x08, 0x0C |

Without the TMP94C241 cases, `LDC cr,DMAMn` writes go to a dummy register and DMA never configures. This blocks the Technics KN5000 (the only driver currently using TMP94C241) from performing HDMA-based firmware payload transfers.

## Changes

**`900tbl.hxx`**: Add TMP94C241 `case` entries to the six existing CR switch blocks in `prepare_operands()`. The encoding values are disjoint between variants so the new cases coexist harmlessly — unreachable cases have zero runtime cost in the compiler's jump table. Each case is annotated with a comment identifying which variant(s) use that encoding.

**`dasm900.h`**: Add `cr_sym` struct and `cr_name()` lookup, paralleling the existing SFR symbol table pattern.

**`dasm900.cpp`**: Replace six hardcoded switch blocks for `O_CR8`/`O_CR16`/`O_CR32` with table-driven lookups via `cr_name()`.

**`tmp94c241.cpp`**, **`tmp96c141.cpp`**, **`tmp95c061.cpp`**, **`tmp95c063.cpp`**: Each variant defines its own `*_cr_syms[]` table passed to the disassembler constructor.

## Test plan

- [x] Verified existing TMP96C141/TMP95C061/TMP95C063 CR register disassembly is unchanged
- [x] Verified TMP94C241 DMA register writes now reach the correct internal registers
- [x] KN5000 HDMA payload transfer completes successfully with this fix applied

## Reference

Toshiba TMP94C241F Data Sheet, Section 5 "DMA Controller", Table 5-1 "DMA Register Map".

### DIFF
--- a/src/devices/cpu/tlcs900/900tbl.hxx
+++ b/src/devices/cpu/tlcs900/900tbl.hxx
@@ -3930,16 +3930,28 @@ void tlcs900_device::prepare_operands(const tlcs900inst *inst)
 		m_imm1.d = RDOP();
 		switch( m_imm1.d )
 		{
-		case 0x22:
+		case 0x22:	// TMP96C141/TMP95C061/TMP95C063
 			m_p1_reg8 = &m_dmam[0].b.l;
 			break;
-		case 0x26:
+		case 0x26:	// TMP96C141/TMP95C061/TMP95C063
 			m_p1_reg8 = &m_dmam[1].b.l;
 			break;
-		case 0x2a:
+		case 0x2a:	// TMP96C141/TMP95C061/TMP95C063
 			m_p1_reg8 = &m_dmam[2].b.l;
 			break;
-		case 0x2e:
+		case 0x2e:	// TMP96C141/TMP95C061/TMP95C063
+			m_p1_reg8 = &m_dmam[3].b.l;
+			break;
+		case 0x42:	// TMP94C241
+			m_p1_reg8 = &m_dmam[0].b.l;
+			break;
+		case 0x46:	// TMP94C241
+			m_p1_reg8 = &m_dmam[1].b.l;
+			break;
+		case 0x4a:	// TMP94C241
+			m_p1_reg8 = &m_dmam[2].b.l;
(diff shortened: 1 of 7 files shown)


felipesanches:
This is the first in a series of 5 planned PRs that I have here waiting for submission that will bring the Technics KN5000 driver to a much better state:
* All GUI + control panel buttons and LEDs responsive via HLE;
* HDMA transfers and subcpu payload init implemented;
* but still without sound emulation (which will require additional research).

Earlier this week I've decided to split the code into these smaller chunks because I don't want to have too much to be reviewed in each PR. And I believe this first one is pretty straight-forward. Please let me know what you think.

felipesanches:
thanks!

PR #14971, opened 2026-02-16
sound/[va_eg, va_ops]: Added CEM3310 and misc stream processing helpers.

m1macrophage:
sound/va_eg.cpp:
* Added CEM3310 implementation.

sound/va_ops.cpp:
* Added device for converting a constant (e.g. control voltages) to a stream, for processing by subsequent stream processing devices.
* Added device for scaling and offseting a signal, for modeling certain op-amp and resistor network circuits.

These are verified on the prophet5 implementation. Submitting as separate PR to keep the size manageable, and since the prophet5 changes build off of a pending PR (#14895)

### DIFF
--- a/src/devices/sound/va_eg.cpp
+++ b/src/devices/sound/va_eg.cpp
@@ -8,7 +8,6 @@
 // The envelope is considered completed after this many time constants.
 static constexpr const float TIME_CONSTANTS_TO_END = 10;
 
-DEFINE_DEVICE_TYPE(VA_RC_EG, va_rc_eg_device, "va_rc_eg", "RC-based Envelope Generator")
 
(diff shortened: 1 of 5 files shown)


src/devices/sound/va_eg.cpp
```diff
@@ -158,3 +157,188 @@ void va_rc_eg_device::snapshot()
 	}
 	m_t_end_approx = m_t_start + attotime::from_double(TIME_CONSTANTS_TO_END * m_r * m_c);
 }
+
+
+cem3310_device::cem3310_device(const machine_config &mconfig, const char *tag, device_t *owner, float rx, float cx)
```
galibert, line 162: Perhaps, in general, such devices should be in an appropriately named file rather than the generic one.  That would be a cem3310.h/cpp in that case.

m1macrophage: Sounds good. I'll follow up with that change.

PR #14976, opened 2026-02-17
Fix MZ-80B cassette loading/multi-block load hang (format + motor-gate)

Aesthetic-Engine:
> Fixes MZ-80B cassette loading. MZ-80B uses 1800 baud (332.75µs / 667µs pulses), not 1200 baud. Adds dedicated mz80b_cassette_formats at 6000 Hz. Tested with session3_engine.mzf — program loads successfully. Suggests removing MACHINE_NOT_WORKING from mz80b.

### DIFF
--- a/src/lib/formats/mz_cas.cpp
+++ b/src/lib/formats/mz_cas.cpp
@@ -20,18 +20,25 @@
 #define SHORT_PULSE 2
 #define LONG_PULSE  4
 
+/* MZ-80B @ 48 kHz: 16 samples = 333µs, 32 samples = 666µs (8x MZ-700 resolution) */
+#define SHORT_PULSE_MZ80B 16
+#define LONG_PULSE_MZ80B  32
+
 #define BYTE_SAMPLES (LONG_PULSE+8*LONG_PULSE)
+#define BYTE_SAMPLES_MZ80B (LONG_PULSE_MZ80B+8*LONG_PULSE_MZ80B)
 
 #define SILENCE     8000
 
 /* long gap and tape mark */
 #define LGAP        22000
+#define LGAP_MZ80B  10000   /* MZ-80B uses 10,000 pulses (vs 22,000 for MZ-700) */
 #define LTM_1       40
 #define LTM_0       40
 #define LTM_L       1
 
 /* short gap and tape mark */
 #define SGAP        11000
+#define SGAP_MZ80B  5000    /* MZ-80B uses 5,000 pulses (proportional to LGAP) */
 #define STM_1       20
 #define STM_0       20
 #define STM_L       1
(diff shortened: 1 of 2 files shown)


Aesthetic-Engine:
Fixes MZ-80B cassette loader hang on multi-block loads (e.g. Monster House (1983) by Falcom). The loader’s “wait for silence” loop at $04BD never exits because MAME freezes the waveform when the motor stops instead of decaying to 0V. Gating the tape data bit with motor state restores correct behavior.

angelosa:
https://sharpmz.org/mz-700/tapeproc.htm

I know that MZ-80B needs the tape signals to be inverted, as we discussed in https://github.com/mamedev/mame/pull/14516#discussion_r2632161508

Beyond that I don't have domain knowledge for tape stuff so asking to @ajrhacker for assistance here.

src/lib/formats/mz_cas.cpp
```diff
@@ -285,6 +329,187 @@ static int fill_wave(int16_t *buffer, int length, const uint8_t *code, int)
...
+		bytecount = 0;
+		header = 0;
+		csum = 0;
+		beg = buffer + count;
+	}
+
+	if( length < BYTE_SAMPLES_MZ80B )
+		return -1;
+
+	if (*code & 0x01) csum++;
```
rb6502, line 499: You could cast *code up to a uint32_t and use MAME's population_count_32() helper, which counts the 1 bits in a single instruction on modern CPUs.  Would be a little bit cleaner.

src/lib/formats/mz_cas.cpp
```diff
@@ -20,18 +20,25 @@
...
+#define SHORT_PULSE_MZ80B 16
+#define LONG_PULSE_MZ80B  32
+
 #define BYTE_SAMPLES (LONG_PULSE+8*LONG_PULSE)
+#define BYTE_SAMPLES_MZ80B (LONG_PULSE_MZ80B+8*LONG_PULSE_MZ80B)
 
 #define SILENCE     8000
 
 /* long gap and tape mark */
 #define LGAP        22000
```
rb6502, line 33: All of these constants should be converted to static constexpr to match modern MAME practice.

rb6502:
I'll give AJR final say, but this looks good to me aside from the two trivial notes I just made.

src/lib/formats/mz_cas.cpp
```diff
@@ -285,6 +329,187 @@ static int fill_wave(int16_t *buffer, int length, const uint8_t *code, int)
 	return count;
 }
 
+/* MZ-80B fill_wave: 48 kHz, LGAP 10k, SGAP 5k, 16/32 sample pulses for sharp edges */
+static int fill_wave_mz80b(int16_t *buffer, int length, const uint8_t *code, int)
+{
+	static int16_t *beg;
+	static uint16_t csum = 0;
+	static int header = 1, bytecount = 0;
+	int count = 0;
```
cuavas, line 338: I know this is largely copy/pasted, but the static function variables make this totally unsafe, and we *need* to be getting rid of this kind of thing rather than adding more of it.

src/lib/formats/mz_cas.cpp
```diff
@@ -285,6 +329,187 @@ static int fill_wave(int16_t *buffer, int length, const uint8_t *code, int)
...
+			return -1;
+		LOG(1,"mz80b fill_wave",("256S\n"));
+		for (i = 0; i < 256; i++)
+			count += fill_wave_0_mz80b(buffer, count);
+
+		file_length = (int)(end - beg) / sizeof(int16_t);
+		if( count + file_length > length )
+			return -1;
+		LOG(1,"mz80b fill_wave",("FILEC %d samples\n", file_length));
+		memcpy(buffer + count, beg, file_length * sizeof(int16_t));
```
cuavas, line 400: What’s going on with the division by `sizeof(int16_t);` here?  It subtracts two `int16_t` pointers, so the result is already in units of `int16_t` size.  Then it tries to convert it back to bytes when it does a `memcpy`, but because of the extra divide by two, it’s only filling *half* the range from `beg` to `end`.  There are several instances of this.  Is this actually intentional but the operation is obfuscated badly, or is the code just plain buggy?

cuavas:
Actually, the whole thing is just the existing function copy/pasted with the comments stripped out.  All that does is cause duplicate maintenance.

PR #14978, opened 2026-02-17
toaplan/*.cpp: Make BCU, FCU, Video controller hardware to separated device, Updates

cam900:
toaplan/toaplan1.cpp:
- Consolidated driver into single file
- Split driver for fireshrk and clones to toaplan/fireshrk.cpp
- Split driver for rallybik and clones to toaplan/rallybik.cpp
  - These are enough to describe 'non-standard toaplan1 driver' so that can be splitted.
- Fix outzonecv raw params
  - It uploads same video timing parameter values as parent set to CRTC.
- Make sprite size LUT RAM as memory_share_creator
- Reduce duplicates
- Use separated palette devices for each layer
- Reduce unnecessary function
- Add/Fix notes

toaplan/toaplan_scu.cpp:
- Move color base initialization into device_start
- Add notes
- Fix initializers
- Fix spacings

### DIFF
--- a/src/mame/toaplan/fireshrk.cpp
+++ b/src/mame/toaplan/fireshrk.cpp
@@ -8,6 +8,7 @@
         Technical info: Carl-Henrik Skarstedt  &  Magnus Danielsson
         Driver updates: Quench
         Video updates : SUZ
+        Driven from toaplan/toaplan1.cpp
 
 
 Hardware info by Guru
--- a/src/mame/toaplan/rallybik.cpp
+++ b/src/mame/toaplan/rallybik.cpp
@@ -8,7 +8,7 @@
         Technical info: Carl-Henrik Skarstedt  &  Magnus Danielsson
         Driver updates: Quench
         Video updates : SUZ
-		Driven from toaplan/toaplan1.cpp
+        Driven from toaplan/toaplan1.cpp
 
 
 Supported games:


src/mame/toaplan/toaplan_fcu.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+/***************************************************************************
+    Sprite Handlers
+***************************************************************************/
+
+template<class BitmapClass>
+void toaplan_fcu_device::draw_sprites_common(screen_device &screen, BitmapClass &bitmap, const rectangle &cliprect)
+{
+	u16 const *const source = (u16 *)m_buffered_spriteram.get();
+	u16 const *const size   = (u16 *)m_buffered_spritesizeram.get();
```
cuavas: Why the casts here?  Shouldn’t `.get()` give you the correct type?

src/mame/toaplan/toaplan1_v.cpp
```diff
@@ -126,668 +128,35 @@ sprite.
...
-												fcu_flipscreen, fcu_flipscreen,
-												sx, sy,
-												screen.priority(), u32(~0) << priority, 0); // ~0 for behind tilemap in same priority
-					sprite++;
-				}
-			}
-		}
-	}
+	// ~0 for behind tilemap in same priority
+	pri_mask = u32(~0) << priority;
```
cuavas: You probably want `~u32(0)` for this.

src/mame/toaplan/toaplan1_v.cpp
cuavas, line 1: It isn’t really worth keeping this stuff as a separate file when the bulk of the emulation has been moved to devices.

src/mame/toaplan/toaplan1_m.cpp
cuavas, line 1: It isn’t really worth keeping this as a separate file either.  There’s hardly anything in it.

src/mame/toaplan/rallybik.cpp
```diff
@@ -0,0 +1,433 @@
...
+		m_paletteram(*this, "paletteram_%u", 0U, 0x800U, ENDIANNESS_BIG),
+		m_sharedram(*this, "sharedram"),
+		m_spriteram(*this, "spriteram")
+	{ }
+
+	void rallybik(machine_config &config) ATTR_COLD;
+
+protected:
+	virtual void machine_start() override ATTR_COLD {}
+	virtual void machine_reset() override ATTR_COLD;
```
cuavas: You don’t need the empty `machine_start` - you can leave it out altogether.

src/mame/toaplan/fireshrk.cpp
```diff
@@ -0,0 +1,822 @@
...
+		m_spriteram(*this, "spriteram", 0x800, ENDIANNESS_BIG),
+		m_spritesizeram(*this, "spritesizeram", 0x80, ENDIANNESS_BIG),
+		m_tjump_io(*this, "TJUMP")
+	{ }
+
+	void fireshrk(machine_config &config) ATTR_COLD;
+
+protected:
+	virtual void machine_start() override ATTR_COLD {}
+	virtual void machine_reset() override ATTR_COLD;
```
cuavas: Same here – no need for empty `machine_start`.  Also, you can probably make the members below this `private`.

src/mame/toaplan/toaplan_fcu.cpp
cam900: That is remains due to just copy-pasted from original source.

PR #14979, opened 2026-02-17
Fix opcodes in ES5510

giulioz:
Fix a few bugs that broke the distortion effect in the Ensoniq SD-1.

- ABS uses operand B
- CMP runs even when skippable and only updates flags (no dst write)
- SIGREG bit 22 drives mul shift
- CCR/CMR readback encoding
- ABS should update zero flag
- ASL2/ASL8 overflow/carry/flag logic

I verified those with the ES5510 manual and they seem to align better.

TODO:
I recognize still some problems that still grant the MACHINE_IMPERFECT_SOUND status, but this is a step forward.
One example is the first patch (OMNIVERSE), in which the reverb effect sound seem to linger forever


MooglyGuy:
@rb6502 Ensoniq tends to be your jam, these changes look good to me, but would you mind having a look?

Issue #14474, opened 2025-11-04
apple: diskiing/iwm corruption


arekkusu42:
### MAME version

0.282 @ 64ec4fb0da839c354bd6a1b98eff52ba2f21a13b

### System information

arm64 macos Sonoma 14.8.1, M3 Max MacBook Pro, Mac15,8

### INI configuration details

```INI

```

### Emulated system/software

apple2e*, apple2c*

### Incorrect behaviour

apple2 machines corrupt 5.25" disks when writing, resulting in the disk becoming unusable.  This only affects machines using clock rates different from the default 1021800 Hz.

Once a disk is corrupted:
* it no longer boots
* [CiderPress2](https://ciderpress2.com/) can't catalog it ("Unable to recognize disk image contents")
* booting from a good disk with the corrupt disk as `-flop2`: CAT,D2 produces "I/O ERROR"

In a DEBUG=1 build, machines using the diskiing emulation hit [the assert recently added to wozfdc_device::lss_sync()](https://github.com/mamedev/mame/pull/14404/commits/dbc96ac5eaee17a606cec28a9fced95e3bf4f364#r2483144659).  In a DEBUG=0 build, proceeding past this assert while writing to a disk results in corruption (read-only access seems to work OK.)

### Expected behaviour

* all apple2 machines (regardless of 60/50Hz, or Zip Chip settings) should not corrupt disks
* unless using a Zip Chip, and software disables the slot6 delay, then all bets are off

### Steps to reproduce

You can reproduce this by booting ProDOS to BASIC, typing 10 REM and SAVE FOO \<repeat as needed\>, but for the sake of reproducibility and burn-in testing, here are more explicit steps:

1) boot **A COPY OF** Apple IIe Diagnostic v2.1 (this disk is IIc-aware)
2) c Choose Tests, \<return\>
3) (e/f) Disk (Card/Port) Test, \<return\>
4) a Disk (Card/Port) Test, \<return\>, \<esc\>, \<esc\>
5) d Options, \<return\>
6) a Loop Tests Until ESC is Pressed, \<return\>, \<esc\>
7) b Execute Selected Tests, \<return\>

### Additional details

Here are some results, across a slice of machines using different disk emulation, with different clock speeds:

```
machine		disk			options				result
-------		----			-------				------
apple2ee	diskiing							100 iterations OK
apple2ee	"				<Zip Chip enabled>	asserts after 2-3 iterations
apple2eede 	"									asserts 1 second into boot
[...]

PR #14980, opened 2026-02-17
agat: use gameio bus; change CPU clock so that writes to diskii work.

shattered:


arekkusu42:
Clearly it is better to have a working disk, but this seems like [another example](https://github.com/mamedev/mame/issues/14474#:~:text=the%207M%20and%20Q3%20clocks) where we need to use [the correct XTAL](https://www.facebook.com/groups/ElectronicParts/posts/1958654474323878/) and pass the correct Q3 to bus devices.  (After the scheduler time-travel is sorted out...)

PR #14981, opened 2026-02-17
igs/pgm.cpp: one new set and another promoted to working

Osso13:
New working clones
------------------
Zhan Shen Chuan Shuo 2 Tebie Ban (bootleg of Knights of Valour Super Heroes Plus, V201CN, May 17 2010 10:43:09) [eziochiu, Mr.Qin]

Clones promoted to working
--------------------------
Gaishi Yingxiong Jiaqiang Ban (bootleg of Knights of Valour Super Heroes Plus, V302CN, Dec 29 2010 16:31:32) [eziochiu, Mr.Qin]


src/devices/machine/lpc2132_vic.cpp
```diff
@@ -0,0 +1,257 @@
...
+				m_vic_vect_addr_cur = m_vic_vect_addr[i];
+				return m_vic_vect_addr_cur;
+			}
+		}
+	}
+
+	// No matching vectored interrupt, return default address
+	m_vic_vect_addr_cur = m_vic_def_vect_addr;
+	return m_vic_vect_addr_cur;
+}
```
cuavas, line 122: Should this have side effect checks?

scripts/src/machine.lua
```diff
@@ -5580,3 +5580,18 @@ if MACHINES["QUADMOUSE"] then
...
+--
+--@src/devices/machine/lpc2132_vic.h,MACHINES["LPC2132_VIC"] = true
+---------------------------------------------------
+
+if (MACHINES["LPC2132_VIC"]~=null) then
+	files {
+		MAME_DIR .. "src/devices/machine/lpc2132_vic.cpp",
+		MAME_DIR .. "src/devices/machine/lpc2132_vic.h",
+	}
+end
```
cuavas: Can you try to find a sensible place for this and not add a random number of blank lines at the end of the file?

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+}
+
+void pgm_arm_type1_state::pgm_decode_kovgsyx_samples()
+{
+	u8 *src = (u8 *)(memregion("ics")->base() + 0x800000);
+
+	for (int i = 0; i < 0x400000; i+=2) {
+		src[i + 0x000001] = src[i + 0x800000];
+		src[i + 0x400001] = src[i + 0xc00000];
+	}
```
cuavas: This is an Allman-formatted file, isn’t it?

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+}
+
+void pgm_arm_type1_state::pgm_decode_kovgsyx_program()
+{
+	u16 *src = (u16 *)(memregion("maincpu")->base() + 0x100000);
+	std::vector<u16> dst(0x400000);
+
+	for (int i = 0; i < 0x400000 / 2; i++)
+	{
+		int j = bitswap<24>((i & 0x7ffff) | ((((i >> 19) + 1) & 3) << 19), 23, 22, 21, 20, 19,  18, 16, 1, 3, 5, 7, 9, 11, 13, 15, 17, 14, 12, 10, 8, 6, 4, 0, 2);
```
cuavas: Can’t you simplify `(i & 0x7ffff) | ((((i >> 19) + 1) & 3)` to `((i + 0x80000) & 0x1fffff)`

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+}
+
+void pgm_arm_type1_state::pgm_decode_kovzscs_program()
+{
+	u16 *src = (u16 *)(memregion("maincpu")->base() + 0x100000);
+	std::vector<u16> dst(0x400000);
+
+	for (int i = 0; i < 0x400000 / 2; i++)
+	{
+		int j = bitswap<24>((i & 0x7ffff) | ((((i >> 19) + 1) & 3) << 19), 23, 22, 21, 20, 19, 2, 1, 18, 17, 16, 15, 14, 13, 12, 11, 10, 0, 3, 4, 5, 6, 7, 8, 9);
```
cuavas: Same here - the manipulation looks overly complicated.

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+		m_kovgsyx_highlatch_68k_w = data;
+		m_lpc2132_vic->set_irq(16, ASSERT_LINE);
+		m_kovgsyx_lowlatch_arm_w = 0;
+	}
+	else
+	{
+		m_kovgsyx_lowlatch_68k_w = data;
+		m_kovgsyx_highlatch_arm_w = 0;
+	}
+}
```
cuavas: It’s no use putting a `machine().scheduler().synchronize();` there and then doing whatever it is immediately – you’ll still have the effect visible immediately.  You need to pass a callback to `synchronize` to make the effects visible after the CPUs have caught up with each other.

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+u16 pgm_arm_type1_state::kovgsyx_asic27a_read_word(offs_t offset)
+{
+	if (!machine().side_effects_disabled())
+		machine().scheduler().synchronize();
+
+	if (offset & 1)
+		return m_kovgsyx_highlatch_arm_w;
+	else
+		return m_kovgsyx_lowlatch_arm_w;
+}
```
cuavas: Same here – the `machine().scheduler().synchronize();` isn’t doing what you want.  It’s effectively going to happen *after* the read, so it’s no use.

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+				status |= (m_kovgsyx_pll_enabled & 0x03) << 8;
+				status |= (m_kovgsyx_pll_config & 0x1f);
+
+				return status;
+			}
+
+			case 0x23:
+				return 0;
+		}
+	}
```
cuavas: More mixed Allman and K&R in this function.  Also in the one below.

src/mame/igs/pgmprot_igs027a_type1.cpp
```diff
@@ -457,6 +457,346 @@ void pgm_arm_type1_state::init_kovqhsgs()
...
+	save_item(NAME(m_kovgsyx_highlatch_68k_w));
+	save_item(NAME(m_kovgsyx_lowlatch_68k_w));
+	save_item(NAME(m_kovgsyx_pll_lock_timer));
+	save_item(NAME(m_kovgsyx_pll_enabled));
+	save_item(NAME(m_kovgsyx_pll_config));
+	save_item(NAME(m_kovgsyx_i2c1sclh));
+	save_item(NAME(m_kovgsyx_i2c1conset));
+	save_item(NAME(m_kovgsyx_i2c1dat));
+	save_item(NAME(m_kovgsyx_i2c1_status));
+	save_item(NAME(m_kovgsyx_handshake_done));
```
cuavas: You’ve got multiple copy/pastes of this.  Are you sure it shouldn’t be in a derived class?

src/devices/machine/lpc2132_vic.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+	m_vic_irq_status = irq_pending;
+
+	m_vic_fiq_status = fiq_pending;
+
+	// Output IRQ status
+	m_irq_out(irq_pending ? ASSERT_LINE : CLEAR_LINE);
+
+	// Output FIQ status
+	m_fiq_out(fiq_pending ? ASSERT_LINE : CLEAR_LINE);
```
cuavas, line 91: You should be checking whether the IRQ and FIQ output would change here, as the callbacks could be bound to something expensive, e.g. something like:
```c++
	bool const irq_changed = bool(m_vic_irq_status) != bool(irq_pending);
	m_vic_irq_status = irq_pending;

	bool const fiq_changed = bool(m_vic_fiq_status) != bool(fiq_pending);
	m_vic_fiq_status = fiq_pending;

	// Output IRQ status
	if (irq_changed)
		m_irq_out(irq_pending ? ASSERT_LINE : CLEAR_LINE);

	// Output FIQ status
	if (fiq_changed)
		m_fiq_out(fiq_pending ? ASSERT_LINE : CLEAR_LINE);
```

src/devices/machine/lpc2132_vic.cpp
```diff
@@ -0,0 +1,261 @@
...
+}
+
+uint32_t lpc2132_vic_device::vect_addr_r()
+{
+	if (!machine().side_effects_disabled())
+		return read_vector_address();
+
+	// when side effects are disabled (e.g. debugger), return current address without updating state
+	return m_vic_vect_addr_cur;
+}
```
cuavas, line 180: The trouble with doing it like this is that it won't necessarily show the value that would be read.  You're better off updating `read_vector_address` to calculate the updated address *but not update member variables* so that the debugger will show the value that would actually be read while still not causing spurious changes to state.

src/devices/machine/lpc2132_vic.cpp
```diff
@@ -0,0 +1,261 @@
...
+{
+	m_vic_irq_status = 0;
+	m_vic_fiq_status = 0;
+	m_vic_raw_intr = 0;
+	m_vic_int_select = 0;
+	m_vic_int_enable = 0;
+	m_vic_soft_int = 0;
+	m_vic_protection = 0;
+	m_vic_vect_addr_cur = 0;
+	m_vic_def_vect_addr = 0;
```
cuavas, line 67: `m_vic_raw_intr` holds the state of input signals.  That shouldn't be changed here.  If the devices producing those signals change them on reset, it's their responsibility to push out the change.

Osso13:
Closing due to the author not addressing the second round of review. Will be reopened if / when it happens.

PR #14982, opened 2026-02-17
namcos12: Fix display range alignment for tektagt

tbender4:
Create native_display_range mode for System 12. This fixes health bar clipping.

---

How it currently appears in MAME 0.285. Notice the overscan on the right:

With this PR's changes, the health bars align much better:

---

As a side note, Tekken 3 renders the same. Likely safe to apply across System 12.
MAME 0.285:

And this PR:


MooglyGuy:
This is a complete hack to a fundamental PSX-related device. People have tried this exact hack before and it was rejected for exactly the same reasons. Kindly leave the PSX code re-writing to @smf-.

PR #14985, opened 2026-02-18
nichibutsu/nyanpai.cpp: Sync file/driver name with updated metadata, Cleanups:

cam900:
- Change filename nichibutsu/niyanpai.cpp to nichibutsu/nyanpai.cpp for sync with metadata
- Reduce duplicates
- Use struct for each layer
- Remove unnecessary function
- Make palette RAM into shared pointer
- Use shorter typename value for consistency

nichibutsu/nichisnd.cpp:
- Use device_mixer_interface for sound routing


src/mame/nichibutsu/nyanpai.cpp
```diff
@@ -36,15 +36,90 @@
...
+		case 0x09:  layer.m_blitter_sizey = data; break;
+		case 0x0a:  layer.m_blitter_destx = (layer.m_blitter_destx & 0xff00) | data; break;
+		case 0x0b:  layer.m_blitter_destx = (layer.m_blitter_destx & 0x00ff) | (data << 8); break;
+		case 0x0c:  layer.m_blitter_desty = (layer.m_blitter_desty & 0xff00) | data; break;
+		case 0x0d:  layer.m_blitter_desty = (layer.m_blitter_desty & 0x00ff) | (data << 8);
+					gfxdraw(Layer);
+					break;
+		default:    break;
+	}
+}
```
cuavas, line 55: I don’t like these having to be moved here.  If the video hardware is changed to a device, they’ll go with the device anyway.  You know all possible layer values, so you can explicitly instantiate the member function template.

PR #14986, opened 2026-02-18
dendego.lay: Update texts and add bulb color

pizzzza19:
Update English texts to the real machine Japanese texts. Add bulb color to the door lamp.

### DIFF
--- a/src/mame/layout/dendego.lay
+++ b/src/mame/layout/dendego.lay
@@ -30,13 +30,25 @@ authors:hap
 <!-- panel text -->
 
 	<element name="text_door">
+		<text string="doors">
+			<color red="0.0" green="0.0" blue="0.0" />
+		</text>
+	</element>
+
+<element name="text_door_j">
 		<text string="車内信号">
 			<color red="1.0" green="0.0" blue="0.0" />
 		</text>
 	</element>
 
 	<element name="text_lets">
-		<text string="電車でGO!">
+		<text string="LET'S GO BY TRAIN!">
+			<color red="0.3" green="0.6" blue="0.9" />
+		</text>
+	</element>
+
+	<element name="text_lets_j">
+		<text string="電車でGO！">
 			<color red="0.3" green="0.6" blue="0.9" />
 		</text>
 	</element>
@@ -45,15 +57,23 @@ authors:hap
 <!-- throttle -->
 
 	<element name="text_throttle">
+		<text string="THROTTLE:" align="1">
+			<color red="0.8" green="0.8" blue="0.8" />
+		</text>
+	</element>
+
+    <element name="text_throttle_j">
 		<text string="マスコン:" align="1">
 			<color red="0.8" green="0.8" blue="0.8" />
 		</text>
 	</element>
-	<element name="text_dot">
+
+    <element name="text_dot">
 		<text string="." align="0">
 			<color red="1.0" green="1.0" blue="0.0" />
 		</text>
 	</element>
+
 	<element name="text_kmh">
 		<text string="km/h" align="1">
 			<color red="1.0" green="1.0" blue="0.0" />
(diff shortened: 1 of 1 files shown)


happppp:
It looks like control panels with either "kPa" or "kg/cm2" exist, let's keep the shorter "kPa".
I don't really mind more accurate labels for this Japan-only game, but I think it's preferable if it can still be viewed in English (translated) like I did with debutm.lay for example.

pizzzza19:
kPa is used for later title がんばれ運転士!! with air pressure brake system.
It's much better to be selectable UI [ENG/JPN] by press a key. 
Need more time to be modified working. 
Is it possible to revise?
Anyway i understand your opinion.

happppp:
Yes, 2 languages in a .lay file is possible, as I said with for example debutm.lay, see: https://github.com/mamedev/mame/blob/master/src/mame/layout/debutm.lay#L412
The view can be selected under Video Options in the Tab UI (not a key).

pizzzza19:
> Yes, 2 languages in a .lay file is possible, as I said with for example debutm.lay

Got it and i added Japanese UI on video options. Don't worry it is automatically selected English and "kPa" odometer as default. Only UI change if player select Japanese via options.

Door lamp color added to both of them.

happppp:
Ok, I have a small update ready after I merge.
I'm undecided on which language should be the default: Japanese = accurate, English = most users can read.

pizzzza19:
@happppp 
thank you, i checked the merged build and successfully working both ENG/JPN UI.
(I forgot the value conversion of kPa to kg/cm2. thx anyway)
You can choose the default language. I personal think it is natural to follow the real hardware label.

There are another "1" "2" "3" brake lamp available on the real machine. I'll check these functions and mechanism with taito-jc board.
Fake Voltage odometer is excluded from UI but i understand it because it always shows 1.5kV and never move.

happppp:
I can see those brake lamps here. https://forums.arcade-museum.com/threads/densha-de-go.100213/
Do you have a link to a better photo?

I don't know where they're output. In service mode, only 8 lamps are tested.
```
d0: START
d1: DOOR
d2: JYOUYO (normal)
d3: HIJYOU (emergency)
d4: DENSEI (three-stage power)
d5: POP L
d6: POP R
d7: ------- (?)
```

Or maybe it's showing the position of the brake lever. MAME already shows that with a number.

happppp:
The 1/2/3 (maybe just 1 lamp) in action: https://www.youtube.com/shorts/bTUCLKLKTpM
edit: oh, it's probably DENSEI.

pizzzza19:
I'm still looking for clear videos but it can be DENSEI.
It looks both throttle and braking will flash the bulb. (<20km/h braking will turn off the bulb)

[edit]
Yes, existing UI may be typo. "PWR lamp"（三相lamp）should be 1/2/3 DENSEI lamp.
I've never seen 三相lamp flashing on real machine, may be it is a dummy lamp. Need check videos for more evidence.
Left "PWR lamp" can be existed on UI but always turn-off ("0" state) and need to add right side 1/2/3 lamp.

They look 3 in 1 bulb. This is not a real function but gaming cosmetics.
The real trains (1970s era) 1/2/3 bulb reflects the brake pressure strength and each bulb isolated, which made me conflicts.
Also many arcade owners won't replace bulbs in maintenance and barely see them working correct.

happppp:
Here is a longer video https://www.youtube.com/watch?v=EqQF6Cmj7AY

I changed the lamps here https://github.com/mamedev/mame/commit/64842a3832f96be21d1eba398063742d92a78ceb
I also moved the unknown ------- lamp to pwr, the one that turns on when the machine is booting.

pizzzza19:
Thank you. I'll try with the latest build again.
Next time when i found the real arcade machine, I double check the lamp function and will update if any other typo found. Or layout improvements.

PR #14987, opened 2026-02-18
emu/debug/debugcpu.cpp: Guarded pointer accessors

holub:


src/emu/debug/debugcpu.cpp
```diff
@@ -623,6 +623,11 @@ device_debug::~device_debug()
 	exceptionpoint_clear_all();
 }
 
+symbol_table &device_debug::symtable()
+{
+	return m_symtable != nullptr ? *m_symtable : m_device.machine().debugger().cpu().global_symtable();
+}
```
cuavas, line 629: This doesn’t sit right with me.  In my mind, this getter is supposed to give the symbol table belonging to the device, not a symbol table that may be used with the device.  I’d rather an assert that the symbol table exists before dereferencing it.

holub: The problem is that I changed the logic recently. Previously any device used default CPU's symtable from constructor, no matter if the device is executable or not.
Now the devices don't have default CPU table. We either need to introduce new symtable type and create empty symtable for every device, or use GLOBAL one for non-cpu.

see: https://github.com/mamedev/mame/pull/14898/changes#diff-c256f1cb010cb8d3abe7205d6c022e60aac634dbe8d7eb0d9fc4a5187df03aacL491

PR #14988, opened 2026-02-19
nmk/nmk_irq.cpp: Convert nmk/ddealer.cpp, nmk/nmk16.cpp video interrupt generation into separated device, Add notes

cam900:
Also add notes for nmk/quizpani.cpp and atlus/patapata.cpp (very likely NMK-ish hardware configuration) interrupt generator similarity with nmk/nmk_irq.cpp (both has PROM for video timing, so possibly these are has similar IRQ generator?)

### DIFF
--- a/src/mame/nmk/nmk_irq.h
+++ b/src/mame/nmk/nmk_irq.h
@@ -31,12 +31,12 @@ public:
 	void set_prom_frame_offset(u32 offset) { m_prom_frame_offset = offset; }
 	void set_vtiming_prom_usage(u32 usage) { m_vtiming_prom_usage = usage; }
 
-	TIMER_CALLBACK_MEMBER(scanline_callback);
-
 protected:
 	// device_t implementation
 	virtual void device_start() override ATTR_COLD;
 	virtual void device_reset() override ATTR_COLD;
+
+	TIMER_CALLBACK_MEMBER(scanline_callback);
 
 private:
 	// memory pointers


src/mame/nmk/nmk_irq.cpp
```diff
@@ -0,0 +1,253 @@
+// license: BSD-3-Clause
+// copyright-holders:Mirko Buffoni,Nicola Salmoria,Bryan McPhail,David Haywood,R. Belmont,Alex Marshall,Angelo Salese,Luca Elia
+// thanks-to:Richard Bush
+/***************************************************************************
```
cuavas, line 3: None of those people wrote this part of the code.  This part was written by Sergio Galiano.

src/mame/nmk/nmk_irq.h
```diff
@@ -0,0 +1,67 @@
...
+// thanks-to:Richard Bush
+/***************************************************************************
+
+  NMK IRQ Generator
+
+***************************************************************************/
+
+#ifndef MAME_DATAEAST_NMK_IRQ_H
+#define MAME_DATAEAST_NMK_IRQ_H
+
```
cuavas, line 11: Wrong project name breaks the build.

src/mame/nmk/nmk_irq.h
```diff
@@ -0,0 +1,67 @@
...
+	void set_prom_start_offset(u32 offset) { m_prom_start_offset = offset; }
+	void set_prom_frame_offset(u32 offset) { m_prom_frame_offset = offset; }
+	void set_vtiming_prom_usage(u32 usage) { m_vtiming_prom_usage = usage; }
+
+	TIMER_CALLBACK_MEMBER(scanline_callback);
+
+protected:
+	// device-level overrides
+	virtual void device_start() override ATTR_COLD;
+	virtual void device_reset() override ATTR_COLD;
```
cuavas: Please don't copy/paste this nonsensical comment – you can use “device_t implementation” or something if you want a comment to label these member functions.

src/mame/nmk/nmk_irq.h
```diff
@@ -0,0 +1,67 @@
...
+
+	// internal states
+	emu_timer *m_scanline_timer;
+
+	u8 m_vtiming_val;
+};
+
+
+// device type definition
+DECLARE_DEVICE_TYPE(NMK_IRQ, nmk_irq_device)
```
cuavas: This is a device type declaration, not definition.  The definition is in the .cpp file.

PR #14991, opened 2026-02-19
Refactor recently-added Lua hooks for vector devices

MooglyGuy:
Based on feedback from @cuavas and @galibert, this refactors @davepl's Lua hooks for vector drawing to use notifier-based callbacks, as the present `register_function`-based hooking is considered deprecated. It also adds a `vector_device_enumerator` in order to accommodate the possibility of systems where there may be more than one `vector_device`.

The API is largely the same albeit across four notifier callbacks:
- `vector_device:add_frame_begin_notifier`
- `vector_device:add_frame_end_notifier`
- `vector_device:add_move_notifier`, takes parameters x, y, color, width, height, in that order.
- `vector_device:add_line_notifier`, takes parameters lastx, lasty, x, y, color, intensity, width, height, in that order.

This also adds documentation for the relevant device to MAME's Lua-related docs, as well as a demonstration plugin that simply prints begin/end/move/line events.

### DIFF
--- a/src/devices/video/vector.cpp
+++ b/src/devices/video/vector.cpp
@@ -2,30 +2,10 @@
 // copyright-holders:Brad Oliver,Aaron Giles,Bernd Wiebelt,Allard van der Bas
 /******************************************************************************
  *
- * vector.c
+ * vector.cpp
  *
  *        anti-alias code by Andrew Caldwell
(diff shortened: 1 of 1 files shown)


cuavas:
Thanks.

PR #14996, opened 2026-02-20
linn_linndrum.lay: Implemented rotating knobs and click states.

m1macrophage:
* Added support for click states, for sliders and knobs. Provides visual feedback when a knob or slider is being manipulated.
* Added support for rotating knob pointers. The pointers can only be circles, since those are easy to rotate.

If this looks good, I will follow up with updating other layouts.


felipesanches:
Oh, yes!!! I'd love to do something similar with the rotating knobs of all those Access Virus synths ;-)

m1macrophage:
Should be relatively easy. Though the pointers can only be dots, and many synths (including the virus) use lines. So that won't be entirely accurate.

PR #15000, opened 2026-02-21
[densha de go!] UI improvements

pizzzza19:
1) Added lamp color same as door lamp. (Not apply to EME lamp)
2) Brake odometer shall not exceed 9.9kg/cm². So digit changed to "1".

### DIFF
--- a/src/mame/layout/dendego.lay
+++ b/src/mame/layout/dendego.lay
@@ -76,7 +76,7 @@ authors:hap
 		</simplecounter>
 	</element>
 	<element name="counter_pressure1" defstate="0">
-		<simplecounter maxstate="99" digits="2" align="2">
+		<simplecounter maxstate="99" digits="1" align="2">
 			<color red="1.0" green="1.0" blue="0.0" />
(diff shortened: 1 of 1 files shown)


happppp:
The physical brake meter and the meter test in service mode both go over 10.
I tweaked the lamp brightness in this commit: https://github.com/mamedev/mame/commit/0fe439c3793552af62baa588ee4548910dc0fb46

pizzzza19:
Yes brake has 10 steps.
Pressure meter at EB Max will be 4.0kg/cm² so no need Digit "2".
Currently 04.0 kg/cm2 -> 4.0 lg/cm2 
It's okay anyway just a minor visual topic.

Thanks for lamp brightness merged.

happppp:
I mean the brake pressure.
The physical odometer goes up to 15, and in testmode (press F2 -> meter test -> break(sic) meter) goes up to 15 as well.

pizzzza19:
> The physical odometer goes up to 15,

Oh i got it. Yes it is. Thanks.

PR #15001, opened 2026-02-21
pc9821.cpp: Implement PC-9821 PEGC Planar Mode

BlueRain-debug:
bf5086fbf5018837bc1b134d09871bd9
881b4636f236b823b16bdc1671391c62

Implemented the PEGC Planar Mode required by the 98MATE 256-color driver in Windows 3.1 and Windows NT.

3a69675d6345caa9d61db904053c7c90

- [x] TODO: Fix for Windows 3.0B.

### DIFF
--- a/src/mame/nec/pc9821.cpp
+++ b/src/mame/nec/pc9821.cpp
@@ -350,7 +350,7 @@ uint16_t pc9821_state::pc9821_grcg_gvram_r(offs_t offset, uint16_t mem_mask)
 
 				if (pat_update)
 				{
-					m_pegc.pattern[tmp & 31] = pixel_data;
+					m_pegc.pattern[tmp & m_pegc.pattern_mask] = pixel_data;
 				}
 			}
 
@@ -388,7 +388,7 @@ uint16_t pc9821_state::pc9821_grcg_gvram_r(offs_t offset, uint16_t mem_mask)
 
 void pc9821_state::pc9821_grcg_gvram_w(offs_t offset, uint16_t data, uint16_t mem_mask)
 {
-	logerror("PEGC_W: PC=%08x Off=%05x Data=%04x\n", m_maincpu->pc(), offset, data);
+	//logerror("PEGC_W: PC=%08x Off=%05x Data=%04x\n", m_maincpu->pc(), offset, data);
 	if (m_ex_video_ff[ANALOG_256_MODE])
 	{
 		int bank_idx = offset >> 14;
@@ -465,7 +465,7 @@ void pc9821_state::pc9821_grcg_gvram_w(offs_t offset, uint16_t data, uint16_t me
 						ext_gvram[tmp] &= plane_mask;
 
 						u8 c4 = 0, c8 = 0;
-						if (ropmethod == 0) { c4 = c8 = m_pegc.pattern[tmp & 31]; } 
+						if (ropmethod == 0) { c4 = c8 = m_pegc.pattern[tmp & m_pegc.pattern_mask]; } 
 						else if (ropmethod == 1) { c4 = c8 = pal2; } 
 						else if (ropmethod == 2) { c4 = c8 = pal1; } 
 						else { c4 = pal1; c8 = pal2; }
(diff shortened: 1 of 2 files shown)


src/mame/nec/pc9821.cpp
```diff
@@ -372,68 +558,103 @@ void pc9821_state::pegc_mmio_map(address_map &map)
...
-		NAME([this] (u8 data) {
-			m_pegc.packed_mode = bool(BIT(data, 0));
-			logerror("$e0100 packed mode %02x\n", data);
-		})
-	);
-	map(0x102, 0x102).lw8(
-		NAME([this] (u8 data) {
-			logerror("$e0102 upper VRAM %s (%02x)\n", BIT(data, 0) ? "enable" : "disable", data);
-			if (BIT(data, 0))
+	map(0x100, 0x19f).rw(FUNC(pc9821_state::pegc_mmio_r), FUNC(pc9821_state::pegc_mmio_w));
```
angelosa: The idea here about using lambdas is that the whole implementation block will be moved out from pc9821.cpp itself ...

src/mame/nec/pc9821.cpp
```diff
@@ -372,68 +558,103 @@ void pc9821_state::pegc_mmio_map(address_map &map)
...
+	if (addr == 0x102 && ACCESSING_BITS_0_7)
+	{
+		logerror("$e0102 upper VRAM %s (%02x)\n", BIT(data, 0) ? "enable" : "disable", data & 0xff);
+		if (BIT(data, 0))
+			m_pegc_vram_view.select(0);
+		else
+			m_pegc_vram_view.disable();
+	}
+
+	if (addr >= 0x120 && addr <= 0x19f)
```
angelosa: ... also I'm pretty sure doing 3 conditional if blocks per access isn't efficient.

src/mame/nec/pc9821.cpp
BlueRain-debug: sorry, this was mainly due to my unfamiliarity with lambdas, so I grouped them into a single block for easier management. I'll work on refactoring and breaking them down

src/mame/nec/pc9821.cpp
```diff
@@ -389,48 +575,78 @@ void pc9821_state::pegc_mmio_map(address_map &map)
...
-			pc9821_egc_w(0x6 / 2, data, 0xff);
+	map(0x104, 0x11f).lrw16(
+		NAME([this] (offs_t offset) {
+			u32 r_idx = 4 + (offset * 2);
+			return m_pegc.regs[r_idx] | (m_pegc.regs[r_idx + 1] << 8);
+		}),
+		NAME([this] (offs_t offset, u16 data, u16 mem_mask) {
+			u32 r_idx = 4 + (offset * 2);
+			if (ACCESSING_BITS_0_7) m_pegc.regs[r_idx] = data & 0xff;
+			if (ACCESSING_BITS_8_15) m_pegc.regs[r_idx + 1] = (data >> 8) & 0xff;
```
angelosa: What happens if you just use `lrw8` here, and drop the `ACCESSING_BITS` / `mem_mask`?

angelosa:
afaik there are two games requiring Planar Mode:
1. `flashb` Flashback in 256 color mode;
2. `silentsr` The Silent Service CD version (on intro, requiring the planar mode to clear text graphics periodically);

Could you please check them?

BlueRain-debug:
> afaik there are two games requiring Planar Mode:
> 
> 1. `flashb` Flashback in 256 color mode;
> 2. `silentsr` The Silent Service CD version (on intro, requiring the planar mode to clear text graphics periodically);
> 
> Could you please check them?

Unfortunately, after some testing, it looks like none of the popular emulators that support PEGC Planar Mode can actually run Flash Back in 256c correctly.

**SL9821**: screen refresh issue; horizontal offset present.
**NP21w**: abnormal logo display; color artifacts on the right margin (likely caused by incorrect rendering coordinates); backgrounds (?) fail to render across all screens.
**MAME** (this PR): screen refresh issue.

More documentation may be needed to implement support for Flash Back.

SL9821:
49792b95749dc4b7fcdf25867ccf2a40
c0385e292d29f02f0fdba17f9d753a3d
1b8e317d3028b9b74328a69b1842d0da

NP21w:
2629d380c2e33ad7bded86102ca3cd75
image
image
b4c919ac70200813bb69fdc134f258ba

This PR:
1ac907975063fb112d9d7b612feea0fd
d817922972dc94124b28e6351ddae954
3d64cf148763f0c487375fe0df6cc5d2

Ref (16c):
image
image

src/mame/nec/pc9821.h
```diff
@@ -25,6 +25,7 @@ class pc9821_state : public pc9801bx_state
 	}
 
 	void pc9821(machine_config &config);
+	bool m_pegc_planar_supported = true;
```
angelosa: What's this for? Some debug thing you accidentally left on?

angelosa:
Still better than what was there, I mean:

src/mame/nec/pc9821.h
BlueRain-debug: Later PC9821 models didnt come with a native PEGC chip, and they only emulated the PEGC Packed Mode. This boolean is used to mark the type of device, and should be initialized depending on the model, but I couldn't find a very definitive list at the moment.

angelosa: Please drop it. Later models are supposed to use PCI Coregraph, and it's a problem for when the PEGC is device-ified (by that point that will be done as a config setter).

angelosa: Also we don't have any PCI-era romset that do anything useful at current stage, because converting the whole thing to PCI_ROOT is non trivial.

BlueRain-debug: done. I think I've pushed this as far as it can go for now :p

angelosa: Thanks, will apply after CI does its job :)

cracyc:
Flashback has copy protection that is broken on the softlist image.  It works correctly with a fixed image.

PR #15002, opened 2026-02-22
kaneko/kan_pand.cpp and related drivers: Cleanups and Updates

cam900:
common:
- Make some member naming non discriminatory
- Use BIT macro for single bit flags
- Fix loggings
- Fix spacings
- Fix namings
- Reduce literal tag usages
- Fill empty sprite gfx region to transparent

kaneko/kan_pand.cpp:
- Remove unnecessary checks
- Fix comments (Original chip has 14 bit tile code, used at djboy.cpp and hvyunit.cpp)
- Fix sprite framebuffer RAM organization (256x256 8 bit double buffered)

kaneko/airbustr.cpp:
- Fix ROM set name in comments

kaneko/djboy.cpp:
- Fix initializer
- Reduce duplicated comment block
- Add notes for bankswitch entry

kaneko/galpanic.cpp:
- Restrict foreground bitmap draw routine to screen cliprect

kaneko/hvyunit.cpp:
- Remove empty override

kaneko/sandscrp.cpp:
- Fix CALC1 naming (CALC1 hasn't any mcu core)

kaneko/snowbros.cpp
- Fix comment order sort to MSB first and same byte
- Split driver states per hardware configs
- Simplified sound bankswitch method for snowbro3 and clones
- Reduce duplicates
- Fix toto dip switch config for describe broken title screen
- Add korean titles when exists, Fix metadata due to title screen

### DIFF
--- a/src/mame/kaneko/kan_pand.cpp
+++ b/src/mame/kaneko/kan_pand.cpp
@@ -59,6 +59,7 @@ kaneko_pandora_device::kaneko_pandora_device(const machine_config &mconfig, cons
 	: device_t(mconfig, KANEKO_PANDORA, tag, owner, clock)
 	, device_video_interface(mconfig, *this)
 	, device_gfx_interface(mconfig, *this)
+	, m_buffer(0)
 	, m_clear_bitmap(false)
 	, m_bg_pen(0)
 	, m_xoffset(0)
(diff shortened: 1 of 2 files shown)


src/mame/kaneko/kan_pand.cpp
```diff
@@ -76,15 +76,15 @@ void kaneko_pandora_device::device_start()
...
 
-	m_sprites_bitmap = std::make_unique<bitmap_ind16>(screen().width(), screen().height());
+	screen().register_screen_bitmap(m_sprites_bitmap);
 
 	save_item(NAME(m_clear_bitmap));
 	save_item(NAME(m_bg_pen));
 	save_item(NAME(m_flip_screen));
 	save_pointer(NAME(m_spriteram), 0x1000);
-	save_item(NAME(*m_sprites_bitmap));
+	save_item(NAME(m_sprites_bitmap));
```
cuavas: It isn’t really safe to save a bitmap that could potentially be resized.  If you use `register_screen_bitmap`, then if something else subsequently reconfigures the screen, the bitmap will be resized (and potentially reallocated), causing memory corruption on save/load state.  If you *don’t* use `register_screen_bitmap`, the sprite bitmap could get out of sync with the screen resolution.  The only safe way to do it would be to use an approach that doesn’t require the bitmap to be saved, but that isn’t really practical given the chip supports not clearing the rendered sprite buffer.  Do we know how much RAM these games actually had for the rendered sprite bitmap, and how it’s organised?

cam900: 4 64x4 DRAMs: 256x256 size, 8 bit double buffered. (Screen horizontal/vertical display resolution of all games with pandora chip is <=256/<=240 and some game has x/y offset for sprites.)

reference: http://caiusarcade.blogspot.com/2019/12/snow-bros-repair-log.html

PR #15003, opened 2026-02-22
tmp94c241: Implement DMA subsystem (HDMA + DMAR) and fix port reads

felipesanches:
Implement the TMP94C241's DMA subsystem:

- **HDMA:** Hardware DMA triggered by interrupt events. Transfers one unit per trigger, with priority over normal interrupt dispatch. Channels checked in priority order (0 highest). Supports all DMAM transfer modes (byte/word/long, increment/decrement/fixed) using switch-based decoding matching the proven TMP95C061 implementation.

- **DMAR:** Software DMA trigger register at SFR 0x109. Writing bit N triggers one DMA transfer on channel N. For example, the Technics KN5000 driver writes DMAR once per INT0 to receive bytes from the inter-CPU latch one at a time.

- **HDMA priority in check_irqs():** Interrupts targeted by active HDMA channels are consumed by the DMA engine instead of dispatching to the interrupt handler.

- **INT0 level-detect re-assertion:** After clearing an INT0 flag during IRQ dispatch, re-assert it if the input pin is still active and level-detect mode is configured. This matches real hardware behavior where level-triggered flags are continuously driven.

- **Port read fix:** `port_r()` now returns the output latch value for bits configured as output and the external pin level for input bits. Previously it always returned the external pin level, which broke firmware that reads back its own output state.

- Add debugger `interrupt_hook()` notification on IRQ dispatch.

### DIFF
--- a/src/devices/cpu/tlcs900/tmp94c241.cpp
+++ b/src/devices/cpu/tlcs900/tmp94c241.cpp
@@ -1392,20 +1392,17 @@ void tmp94c241_device::tlcs900_check_irqs()
 			m_check_irqs = 1;
 		}
 
-		// notify the debugger
-		if (machine().debug_enabled())
-		{
-			int8_t default_priority = vector/4 + 1;
-			if (vector > 0x3c)
-				default_priority--; // skip reserved one
+		// Compute the default priority index from the vector table.
+		// The datasheet's "default priority" numbering is vector/4 + 1,
+		// with a gap at 0x3c (reserved), so entries above that shift down by one.
+		int8_t default_priority = vector / 4 + 1;
+		if (vector > 0x3c)
+			default_priority--;
 
-			// For now we can't pass a string with the interrupt name
-			// so, here we'll get a message that says things like "IRQ 20"
-			// and we can look up the TMP94C241 datasheet and find under
-			// "Default priority" on the interrupt table that
-			// "20" means "INTT0: 8-bit timer (Timer 0)"
-			debug()->interrupt_hook(default_priority, m_pc.d);
-		}
+		// The IRQ level passed here corresponds to the datasheet's
+		// "default priority" number, so e.g. "IRQ 20" means
+		// "INTT0: 8-bit timer (Timer 0)" in the TMP94C241 interrupt table.
+		standard_irq_callback(default_priority, m_pc.d);
 	}
 }
 


src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -361,6 +361,22 @@ void tmp94c241_device::dmav_w(offs_t offset, uint8_t data)
...
+void tmp94c241_device::dmar_w(uint8_t data)
+{
+	// DMAR register (0x109) - DMA software request trigger
+	// Writing bit N triggers ONE DMA transfer on channel N (same as HDMA).
+	// For example, the Technics KN5000 driver writes DMAR once per INT0
+	// to transfer a single byte from the inter-CPU latch, relying on the
+	// DMA count to track how many bytes remain in the current transfer block.
+	for (int channel = 0; channel < 4; channel++)
+	{
+		if (data & (1 << channel))
```
galibert: We try to use BIT for that kind of stuff

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -985,11 +1007,272 @@ void tmp94c241_device::internal_mem(address_map &map)
...
+		break;
+	}
+
+	// Decrement transfer count
+	m_dmac[channel].w.l -= 1;
+
+	// Check for transfer completion
+	if (m_dmac[channel].w.l == 0)
+	{
+		logerror("HDMA ch%d complete: src=%06X dst=%06X (vec=%02X)\n",
```
galibert: That should use logerror.h at some point.

src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -1067,6 +1378,34 @@ void tmp94c241_device::tlcs900_check_irqs()
...
+			int8_t default_priority = vector/4 + 1;
+			if (vector > 0x3c)
+				default_priority--; // skip reserved one
+
+			// For now we can't pass a string with the interrupt name
+			// so, here we'll get a message that says things like "IRQ 20"
+			// and we can look up the TMP94C241 datasheet and find under
+			// "Default priority" on the interrupt table that
+			// "20" means "INTT0: 8-bit timer (Timer 0)"
+			debug()->interrupt_hook(default_priority, m_pc.d);
```
galibert: At this point (I have changes in one of my trees, but it's not done yet), you're supposed to call standard_irq_callback(level, pc) rather than debug()->interrupt_hook and do that unconditionally.

felipesanches:
thanks!

PR #15005, opened 2026-02-22
Taito System JC driver files consolidation into single file.

TheLostNinja:
**taitojc_v.cpp** and **taitojc.h** is quite small too.


cuavas:
Doesn’t compile, and you haven’t read the contributing guidelines.  You do this one more time, you get blocked.  I’m serious, you’re wasting everyone’s time.

PR #15006, opened 2026-02-22
New system: tipc

shattered:
Texas Instruments Professional Computer [shattered]

### DIFF
--- a/src/mame/ti/tipc.cpp
+++ b/src/mame/ti/tipc.cpp
@@ -1,0 +1,590 @@
+// license:BSD-3-Clause
+// copyright-holders:Sergey Svishchev
+/***************************************************************************
+
+    Texas Instruments Professional Computer (a MS-DOS compatible PC)
+
+    to do:
+
+    - native keyboard (not dumped)
+    - cursor blink rate, blink attr
+    - diags: pass all tests (interrupt test depends on timing of LOOP insn)
+    - sound
+    - printer
+    - hdc
+    - other cards (clock, serial)?
+    - softlist
+
+***************************************************************************/
+
+#include "emu.h"
+
+#include "bus/rs232/rs232.h"
+#include "cpu/i86/i86.h"
+#include "imagedev/floppy.h"
+#include "machine/clock.h"
+#include "machine/i8251.h"
+#include "machine/input_merger.h"
+#include "machine/keyboard.h"
+#include "machine/pic8259.h"
+#include "machine/pit8253.h"
+#include "machine/ram.h"
+#include "machine/wd_fdc.h"
+#include "sound/spkrdev.h"
+#include "video/mc6845.h"
+
+#include "emupal.h"
+#include "screen.h"
+
+
+#define LOG_DISK      (1U << 1)
+
+//#define VERBOSE (LOG_GENERAL|LOG_DISK)
+#include "logmacro.h"
+
+#define LOGDISK(...) LOGMASKED(LOG_DISK, __VA_ARGS__)
+
+
+namespace {
+
+class tipc_state : public driver_device
+{
+public:
+	tipc_state(const machine_config &mconfig, device_type type, const char *tag)
+		: driver_device(mconfig, type, tag)
+		, m_maincpu(*this, "maincpu")
+		, m_dsw0(*this, "DSW0")
+		, m_upd8251(*this, "upd8251")
+		, m_pit8253(*this, "pit8253")
+		, m_pic8259(*this, "pic8259")
+		, m_rs232(*this, "rs232")
+		, m_irq2_merger(*this, "merge_irq2")
+		, m_irq3_merger(*this, "merge_irq3")
+		, m_ram(*this, RAM_TAG)
+		, m_crtc(*this, "crtc")
+		, m_palette(*this, "palette")
+		, m_vdu_char_rom(*this, "vduchar")
+		, m_fdc(*this, "fdc")
+		, m_floppies(*this, "fdc:%u", 0U)
+	{ }
+
+	required_device<cpu_device> m_maincpu;
+	required_ioport m_dsw0;
+	required_device<i8251_device> m_upd8251;
+	required_device<pit8253_device> m_pit8253;
+	required_device<pic8259_device>  m_pic8259;
+	required_device<rs232_port_device> m_rs232;
+	required_device<input_merger_all_high_device> m_irq2_merger;
+	required_device<input_merger_all_high_device> m_irq3_merger;
+	required_device<ram_device> m_ram;
+	required_device<sy6545_1_device> m_crtc;
+	required_device<palette_device> m_palette;
+	required_region_ptr<u8> m_vdu_char_rom;
+	required_device<fd1793_device> m_fdc;
+	required_device_array<floppy_connector, 2> m_floppies;
+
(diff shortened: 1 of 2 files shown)


src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+		throw device_missing_dependencies();
+
+	m_maincpu->space(AS_PROGRAM).install_ram(0, m_ram->size() - 1, m_ram->pointer());
+	m_buf = make_unique_clear<uint8_t[]>(1024);
+
+	m_crtc_ram = make_unique_clear<uint8_t[]>(2048);
+	m_crtc_attr = make_unique_clear<uint8_t[]>(2048);
+	m_crtc_gfx[0] = make_unique_clear<uint8_t[]>(32768);
+	m_crtc_gfx[1] = make_unique_clear<uint8_t[]>(32768);
+	m_crtc_gfx[2] = make_unique_clear<uint8_t[]>(32768);
```
startaq: Why decimal?

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+	map(0x0010, 0x0011).rw(m_upd8251, FUNC(i8251_device::read), FUNC(i8251_device::write));
+	map(0x0014, 0x0017).rw(m_pit8253, FUNC(pit8253_device::read), FUNC(pit8253_device::write));
+	map(0x0018, 0x0019).rw(m_pic8259, FUNC(pic8259_device::read), FUNC(pic8259_device::write));
+	map(0x0020, 0x0023).rw(FUNC(tipc_state::fdc_r), FUNC(tipc_state::fdc_w));
+	map(0x0030, 0x0033).noprw(); // HDC
+}
+
+void tipc_state::tipc(machine_config &config)
+{
+	I8088(config, m_maincpu, XTAL(5'000'000)); // 15mhz/3
```
startaq: So there is a 15 MHz XTAL? Then write it as `15_MHz_XTAL / 3`. Same for the other clock values below.

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+
+	screen_device &screen(SCREEN(config, "screen", SCREEN_TYPE_RASTER));
+	screen.set_refresh_hz(60);
+	screen.set_size(720, 300);
+	screen.set_visarea(0, 719, 0, 299);
+	screen.set_screen_update("crtc", FUNC(mc6845_device::screen_update));
+
+	PALETTE(config, m_palette, palette_device::BRG_3BIT);
+
+	SY6545_1(config, m_crtc, 18_MHz_XTAL / 1); // FIXME 450hz refresh
```
startaq: Divide by 9.

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+	FD1793(config, m_fdc, 1000000);
+	m_fdc->intrq_wr_callback().set(m_pic8259, FUNC(pic8259_device::ir6_w));
+	m_fdc->drq_wr_callback().set(FUNC(tipc_state::fdc_drq_w));
+	FLOPPY_CONNECTOR(config, "fdc:0", pc_dd_floppies, "525dd", floppy_image_device::default_pc_floppy_formats);
+	FLOPPY_CONNECTOR(config, "fdc:1", pc_dd_floppies, "525dd", floppy_image_device::default_pc_floppy_formats);
+
+	// CRT Controller board
+
+	screen_device &screen(SCREEN(config, "screen", SCREEN_TYPE_RASTER));
+	screen.set_refresh_hz(60);
```
startaq: 60 hz here, but below there is a comment with 450 hz (which would be 50 hz if you divide by the char width). What does it actually configure the 6845 to? In any case, you should supply the raw values programmed to the 6845 here.

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+		, m_irq3_merger(*this, "merge_irq3")
+		, m_ram(*this, RAM_TAG)
+		, m_crtc(*this, "crtc")
+		, m_palette(*this, "palette")
+		, m_vdu_char_rom(*this, "vduchar")
+		, m_fdc(*this, "fdc")
+		, m_floppies(*this, "fdc:%u", 0U)
+	{ }
+
+	required_device<cpu_device> m_maincpu;
```
startaq, line 90: Most of those things don't need to be public.

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+	required_device<input_merger_all_high_device> m_irq3_merger;
+	required_device<ram_device> m_ram;
+	required_device<sy6545_1_device> m_crtc;
+	required_device<palette_device> m_palette;
+	required_region_ptr<u8> m_vdu_char_rom;
+	required_device<fd1793_device> m_fdc;
+	required_device_array<floppy_connector, 2> m_floppies;
+
+	void machine_start() override ATTR_COLD;
+	void machine_reset() override ATTR_COLD;
```
startaq: Those two should be protected.

src/mame/ti/tipc.cpp
```diff
@@ -0,0 +1,590 @@
...
+
+	m_crtc_ram = make_unique_clear<uint8_t[]>(2048);
+	m_crtc_attr = make_unique_clear<uint8_t[]>(2048);
+	m_crtc_gfx[0] = make_unique_clear<uint8_t[]>(32768);
+	m_crtc_gfx[1] = make_unique_clear<uint8_t[]>(32768);
+	m_crtc_gfx[2] = make_unique_clear<uint8_t[]>(32768);
+
+	m_maincpu->space(AS_PROGRAM).install_ram(0xc0000, 0xc7fff, m_crtc_gfx[0].get());
+	m_maincpu->space(AS_PROGRAM).install_ram(0xc8000, 0xcffff, m_crtc_gfx[1].get());
+	m_maincpu->space(AS_PROGRAM).install_ram(0xd0000, 0xd7fff, m_crtc_gfx[2].get());
```
startaq: Why do you install them here instead of putting it into the memory map? You could also use memory shares.

Also save states would be nice to have.

startaq (changes requested):

src/mame/ti/tipc.cpp
shattered: Personal preference.

src/mame/ti/tipc.cpp
shattered: Yep, that works.

src/mame/ti/tipc.cpp
shattered: I'll redo this using memory shares, thanks.   Save states are somewhat pointless right now, IMO.

src/mame/ti/tipc.cpp
shattered: Done.

src/mame/ti/tipc.cpp
shattered: It can drive both 60 and 50hz monitors, I've changed raw params to match 60hz settings.

src/mame/ti/tipc.cpp
shattered: Done.

PR #15009, opened 2026-02-22
include `pngcmp` in the build and added manpage for it

firewave:


firewave:
Is there something else to do to also include it in the distributed packages? I checked references to some of the other tools and did not find anything.

MooglyGuy:
> It generates an new images

Correct to "It generates new images".

PR #15010, opened 2026-02-23
video/ky3211_ky10510.cpp: Make TAXAN KY-3211 and KY-10510 into separate devices, Updates

cam900:
Common:
- Fix spacings
- Use multibyte helper for 16 bit parameters
- Reduce duplicates
- Use shortened typename values for consistency
- Add/Fix notes

sigma/sigmab98.cpp:
- Demote dodghero as MACHINE_IMPERFECT_TIMING:
  - Its title screen running in wrong speed - faster than should be, it causes desyncing title call.
- Split sammymdl driver into separated file
- Fix naming less confused

### DIFF
--- a/src/devices/video/ky3211_ky10510.cpp
+++ b/src/devices/video/ky3211_ky10510.cpp
@@ -94,9 +94,9 @@
 
 #include "screen.h"
 
-static constexpr int integer_part(int x)
+constexpr int integer_part(int x)
 {
-//  return x >> 16;
+	//return x >> 16;
 	return (x + 0x8000) >> 16;
 }
 


scripts/src/video.lua
```diff
@@ -1840,3 +1840,15 @@ if VIDEOS["ZR36110"] then
...
+--
+--@src/devices/video/ky3211_ky10510.h,VIDEOS["KY3211_KY10510"] = true
+--------------------------------------------------
+
+if VIDEOS["KY3211_KY10510"] then
+	files {
+		MAME_DIR .. "src/devices/video/ky3211_ky10510.cpp",
+		MAME_DIR .. "src/devices/video/ky3211_ky10510.h",
+	}
+end
```
cuavas: Can you find a logical place to add this?  If new stuff is always added at the bottom, the file just ends up jumbled in random order, and you can’t guess where anything will be.  Maybe after the Konami video chips?

src/devices/video/ky3211_ky10510.cpp
```diff
@@ -0,0 +1,397 @@
...
+	flipy = BIT(src[0x06], 2);
+
+	return true;
+}
+
+static inline const int integer_part(int x)
+{
+//  return x >> 16;
+	return (x + 0x8000) >> 16;
+}
```
cuavas: Please turn this into a `constexpr` near the top of the file, so it’s easier to find the miscellaneous free function helpers.

src/mame/sigma/sammymdl.cpp
```diff
@@ -0,0 +1,1038 @@
...
+	void sammymdl(machine_config &config);
+	void itazuram(machine_config &config);
+	void pyenaget(machine_config &config);
+
+	void init_itazuram();
+	void init_animalc();
+	void init_haekaka();
+
+protected:
+	virtual void machine_start() override { m_leds.resolve(); }
```
cuavas: Can you add `ATTR_COLD` to these cold lifecycle handlers while you’re refactoring the code?

src/mame/sigma/sammymdl.cpp
```diff
@@ -0,0 +1,1038 @@
...
+{
+	m_leds[0] = BIT(data, 0);   // button
+	m_leds[1] = BIT(data, 1);   // coin lockout? (after coining up, but not for service coin)
+	m_leds[2] = BIT(data, 2);   // ? long after a prize is not collected
+	m_leds[3] = BIT(data, 3);   // ? "don't forget the large prizes"
+
+	// 10 hopper enable?
+	// 20 hopper motor on (active low)?
+	m_hopper_small->motor_w((!(data & 0x20) && (data & 0x10)) ? 1 : 0);
+	m_hopper_large->motor_w((!(data & 0x80) && (data & 0x40)) ? 1 : 0);
```
cuavas: I think you can simplify this to:
```c++
	m_hopper_small->motor_w(BIT(~data, 5) & BIT(data, 4));
	m_hopper_large->motor_w(BIT(~data, 7) & BIT(data, 6));
```

src/devices/video/ky3211_ky10510.h
```diff
@@ -0,0 +1,64 @@
...
+	{
+		set_palette(std::forward<T>(palette_tag));
+	}
+
+	void vregs_w(offs_t offset, u8 data);
+	u8 vregs_r(offs_t offset);
+	void vtable_w(offs_t offset, u8 data);
+	u8 vtable_r(offs_t offset);
+
+	void draw_sprites(bitmap_ind16 &bitmap, const rectangle &cliprect, u8 *spriteram, int pri_mask);
```
cuavas: This shouldn’t modify the sprite RAM – it should be `const u8 *spriteram`

PR #15011, opened 2026-02-23
Fix building with Asio 1.33.0 or newer

brad0:
Build tested with Asio 1.30.2 and 1.36.0.

This builds but that's as much as it has been tested.

### DIFF
--- a/src/lib/util/client_ws.hpp
+++ b/src/lib/util/client_ws.hpp
@@ -99,7 +99,7 @@ namespace webpp {
 			std::list<SendData> send_queue;
 
 			void send_from_queue() {
-				strand.post([this]() {
+				asio::post(strand, [this]() {
 					asio::async_write(*socket, send_queue.begin()->send_stream->streambuf,
 							strand.wrap([this](const std::error_code& ec, size_t /*bytes_transferred*/) {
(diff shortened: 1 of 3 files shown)


brad0:
cc @mmicko

cuavas:
I'm nervous about changes that aren't tested at all.  I tried fixing use of deprecated ASIO APIs in 3c6ffd84b0ae8c7138ea7199ebc8d74e77be0f3e.  Can you see if that catches all the ones that have been removed recently?

brad0:
> I'm nervous about changes that aren't tested at all.

I am definitely not trying to imply committing without testing. Just trying to raise awareness
to the issue and providing a starting point diff for someone who understands the respective
functionality with MAME to be able to be tested.

MAME is the last port in our tree in the way of updating ASIO.

> I tried fixing use of deprecated ASIO APIs in [3c6ffd8](https://github.com/mamedev/mame/commit/3c6ffd84b0ae8c7138ea7199ebc8d74e77be0f3e). Can you see if that catches all the ones that have been removed recently?

Sure. Will do later on today.

brad0:
@cuavas Git now as is builds with ASIO 1.36.0

PR #15015, opened 2026-02-23
tmp94c241: Add serial port sub-device with I/O interface mode

felipesanches:
Replace the inline serial stubs (which just instantly set TX-complete)
with a proper sub-device implementation supporting:

- **I/O interface mode (mode 0):** Synchronous clocked serial using SCLK
  pin. Supports both internal (baud rate generator) and external
  (IOC=1) clock sources. TX double buffering with auto-load from
  buffer to shift register. Pre-outputs bit 0 on TXD before first
  clock edge so receiver can sample it on the rising edge.

- **Baud rate generator:** Configurable via BRxCR register with divisor
  and clock source selection. Timer drives SCLK at the configured
  rate. Clock frequency derived from CPU clock.

- **TX/RX data callbacks:** txd(), rxd(), sclk_out(), sclk_in() for
  connecting external devices to the serial ports.

- **tx_start callback:** Signals the start of each byte transmission with
  the current PFFC pin function state, allowing connected devices to
  distinguish real transmissions from phantom ones.

The serial registers (SC0BUF/SC1BUF, SC0CR/SC1CR, SC0MOD/SC1MOD,
BR0CR/BR1CR) are now delegated to the sub-devices in the internal
memory map. TX-complete flags (INTES0/INTES1 bit 7) are set at
device_reset to indicate empty TX buffers at power-on.

UART modes (7/8/9-bit) are recognized but not yet implemented.

Also moves interrupt register indices from a file-scope enum to
public static constexpr members of tmp94c241_device.

### DIFF
--- a/src/devices/cpu/tlcs900/tmp94c241.cpp
+++ b/src/devices/cpu/tlcs900/tmp94c241.cpp
@@ -8,6 +8,7 @@
 
 #include "emu.h"
 #include "tmp94c241.h"
+#include "tmp94c241_serial.h"
 #include "dasm900.h"
 
 #define LOG_DMA    (1U << 1)
--- a/src/devices/cpu/tlcs900/tmp94c241.h
+++ b/src/devices/cpu/tlcs900/tmp94c241.h
@@ -12,7 +12,8 @@
 #pragma once
 
 #include "tlcs900.h"
-#include "tmp94c241_serial.h"
+
+class tmp94c241_serial_device;
 
 //**************************************************************************
 //  TYPE DEFINITIONS
--- a/src/devices/cpu/tlcs900/tmp94c241_serial.cpp
+++ b/src/devices/cpu/tlcs900/tmp94c241_serial.cpp
@@ -40,14 +40,14 @@ tmp94c241_serial_device::tmp94c241_serial_device(const machine_config &mconfig, 
 	m_txd_cb(*this),
 	m_sclk_in_cb(*this),
 	m_sclk_out_cb(*this),
-	m_tx_start_cb(*this)
+	m_tx_start_cb(*this),
+	m_cpu(*this, DEVICE_SELF_OWNER)
 {
 }
 
 void tmp94c241_serial_device::device_start()
 {
 	m_timer = timer_alloc(FUNC(tmp94c241_serial_device::timer_callback), this);
-	m_cpu = dynamic_cast<tmp94c241_device *>(owner());
 
 	save_item(NAME(m_serial_control));
 	save_item(NAME(m_serial_mode));
--- a/src/devices/cpu/tlcs900/tmp94c241_serial.h
+++ b/src/devices/cpu/tlcs900/tmp94c241_serial.h
@@ -11,7 +11,7 @@
 
 #pragma once
 
-class tmp94c241_device;
+#include "tmp94c241.h"
 
 class tmp94c241_serial_device :
 	public device_t
@@ -74,7 +74,7 @@ protected:
 	devcb_write_line m_sclk_out_cb;
 	devcb_write_line m_tx_start_cb;  // Signals start of new byte transmission
 
-	tmp94c241_device *m_cpu;
+	required_device<tmp94c241_device> m_cpu;
 };
 
 DECLARE_DEVICE_TYPE(TMP94C241_SERIAL, tmp94c241_serial_device)


src/devices/cpu/tlcs900/tmp94c241.cpp
```diff
@@ -24,71 +24,49 @@
...
-	{ INTETC01, 0x08, 0x94, 0x25},  // INTTC0: micro-DMA completion Ch.0
-	{ INTETC01, 0x80, 0x98, 0x26},  // INTTC1: micro-DMA completion Ch.1
-	{ INTETC23, 0x08, 0x9c, 0x27},  // INTTC2: micro-DMA completion Ch.2
-	{ INTETC23, 0x80, 0xa0, 0x28},  // INTTC3: micro-DMA completion Ch.3
-	{ INTETC45, 0x08, 0xa4, 0x29},  // INTTC4: micro-DMA completion Ch.4
-	{ INTETC45, 0x80, 0xa8, 0x2a},  // INTTC5: micro-DMA completion Ch.5
-	{ INTETC67, 0x08, 0xac, 0x2b},  // INTTC6: micro-DMA completion Ch.6
-	{ INTETC67, 0x80, 0xb0, 0x2c},  // INTTC7: micro-DMA completion Ch.7
-									/* 0xb4 ... 0xfc (Reserved) */
+	{ tmp94c241_device::INTE0AD,  0x08, 0x28, 0x0a},  // INT0 Pin
```
galibert: if you put the structure as a constant class variable the koenig lookup should allow to remove the prefixes.

src/devices/cpu/tlcs900/tmp94c241.h
```diff
@@ -64,6 +66,25 @@ class tmp94c241_device : public tlcs900h_device
 	static constexpr uint8_t CAPB = 7;
 
 public:
+	static constexpr uint8_t INTE45   = 0;
```
galibert: enum?

src/devices/cpu/tlcs900/tmp94c241_serial.cpp
```diff
@@ -0,0 +1,361 @@
...
+		// Handle deferred INTTX from the last falling edge.  The receiver
+		// (cpanel) just sampled bit 7 on this rising edge via sclk_out_cb.
+		// Now it's safe to fire INTTX — the ISR may write SC1BUF which
+		// pre-outputs bit 0 of the next byte, but bit 7 has already been
+		// sampled.  On TMP94C241 hardware, the 8th rising edge occurs
+		// 4µs after the last falling edge (at 250 kHz SCLK), well before
+		// the ISR can write SC1BUF.  In MAME's event-driven scheduler,
+		// CPU instructions run between timer ticks, so without this deferral
+		// the ISR would write SC1BUF (pre-outputting bit 0) BEFORE the
+		// trailing rising edge — corrupting bit 7 of every byte.
```
galibert, line 120: Doesn't tx interrupt happen at the end of the stop bit rather than the last data bit?

felipesanches: This implementation only supports the I/O interface mode, which is what I need for the Technics KN5000 control panel. There are 8 SCLK pulses per byte with no start or stop bits and INTTX fires after the last data bit.

rb6502: Is there some kind of an error/warning if something tries to use an unimplemented mode?  Personally I'd lean towards fatalerror() for that, having been on the wrong end of too many silent unimplemented failures in MAME, but even just a logerror() would be OK.

felipesanches: image

felipesanches:
One additional PR (which brings the Technics KN5000 driver to an interactive state [as seen here two weeks ago](https://www.youtube.com/watch?v=gX5oR9YOMJE) with control panel HLE working) is waiting to be submitted after this one is merged.

I'd appreciate this PR being merged soon in case nobody could find anything else that may still need improvements.

src/devices/cpu/tlcs900/tmp94c241_serial.cpp
```diff
@@ -0,0 +1,361 @@
...
+	m_sclk_in_cb(*this),
+	m_sclk_out_cb(*this),
+	m_tx_start_cb(*this)
+{
+}
+
+void tmp94c241_serial_device::device_start()
+{
+	m_timer = timer_alloc(FUNC(tmp94c241_serial_device::timer_callback), this);
+	m_cpu = dynamic_cast<tmp94c241_device *>(owner());
```
galibert: Using a required_device<tmp94c241_device> with name DEVICE_SELF_OWNER will grab the owner, check the type, and generate an error with -valid if someone gets it wrong.

PR #15024, opened 2026-02-25
Implement SDL_MouseWheelEvent integer member fallback for version less than 3.2.12

sairuk:
The SDL_MouseWheelEvent integer members called in the current SDL3 implementation were added in SDL3 3.2.12, this PR implements a fallback similar to the previous SDL implementation which will check the SDL version before using these members.

Debian 13 (Trixie) currently ships with SDL3 v3.2.10, 

i've tested this with `vcop2` and the mame menu, mouse and scroll operations seems fine with this approach. Formatting should be OK, let me know if anything extra is required.

Reference: https://wiki.libsdl.org/SDL3/SDL_MouseWheelEvent

addresses compile errors
```
error: 'const SDL_MouseWheelEvent' {aka 'const struct SDL_MouseWheelEvent'} has no member named 'integer_y'
error: 'const SDL_MouseWheelEvent' {aka 'const struct SDL_MouseWheelEvent'} has no member named 'integer_x'
```

### DIFF
--- a/src/osd/modules/input/input_sdl3.cpp
+++ b/src/osd/modules/input/input_sdl3.cpp
@@ -883,8 +883,13 @@ public:
 
 		case SDL_EVENT_MOUSE_WHEEL:
 			// adjust SDL 1-per-click to match Win32 120-per-click
+#if SDL_VERSION_ATLEAST(3, 2, 12)
 			m_v += std::lround(event.wheel.integer_y * 120 * input_device::RELATIVE_PER_PIXEL);
 			m_h += std::lround(event.wheel.integer_x * 120 * input_device::RELATIVE_PER_PIXEL);
+#else
+			m_v += std::lround(event.wheel.y * 120 * input_device::RELATIVE_PER_PIXEL);
+			m_h += std::lround(event.wheel.x * 120 * input_device::RELATIVE_PER_PIXEL);
+#endif
 			break;
 
 		case SDL_EVENT_WINDOW_MOUSE_LEAVE:


src/osd/modules/input/input_sdl3.cpp
```diff
@@ -781,8 +781,13 @@ class sdl_mouse_device : public sdl_mouse_device_base
 
 		case SDL_EVENT_MOUSE_WHEEL:
 			// adjust SDL 1-per-click to match Win32 120-per-click
+#if SDL_VERSION_ATLEAST(3, 2, 12)
 			m_v += std::lround(event.wheel.integer_y * 120 * input_device::RELATIVE_PER_PIXEL);
 			m_h += std::lround(event.wheel.integer_x * 120 * input_device::RELATIVE_PER_PIXEL);
+#else
+			m_v += std::lround(event.wheel.y * 120 * input_device::RELATIVE_PER_PIXEL);
+			m_h += std::lround(event.wheel.x * 120 * input_device::RELATIVE_PER_PIXEL);
+#endif
```
cuavas, line 790: Why is SDL3 preferring to use `integer_x`/`integer_y`?  Wouldn’t it be better to use `x`/`y` to get smooth scrolling when the user has a touchpad?  Or is there some issue with using `x`/`y` on later versions of SDL3?

rb6502: In SDL2 mouse wheel events were purely integer* so we prefer integer simply to minimize unnecessary deltas between the `sdl `and `sdl3` OSDs during this period when bug reports are still frequent.

(* Floats were added in SDL 2.0.18 but MAME never supported them)

cuavas: > (* Floats were added in SDL 2.0.18 but MAME never supported them)

MAME does support them – I added the code myself in 68472d3d72ea4bbc545c0669348172b4e96c7b41 (29 January, 2023 – over three years ago):
```

		case SDL_MOUSEWHEEL:
			// adjust SDL 1-per-click to match Win32 120-per-click
#if SDL_VERSION_ATLEAST(2, 0, 18)
			m_v += std::lround(event.wheel.preciseY * 120 * input_device::RELATIVE_PER_PIXEL);
			m_h += std::lround(event.wheel.preciseX * 120 * input_device::RELATIVE_PER_PIXEL);
#else
			m_v += event.wheel.y * 120 * input_device::RELATIVE_PER_PIXEL;
			m_h += event.wheel.x * 120 * input_device::RELATIVE_PER_PIXEL;
#endif
			break;
```

cuavas: And while we’re at it, you obviously copy/pasted the code from the post-floating point version, because you copied the `lround` calls, which don’t make sense for integers, but you apparently weren’t thinking too clearly, because you’re passing integers to `lround` in your version.

cuavas: But seriously, don’t try to gaslight me, don’t tell me that stuff was “never supported” when a quick glance at the code shows the opposite, and the version we’re looking at here is clearly base on the version with support for floating point values.

rb6502: Look, I assume that if people are doing feature work on OSD that the commit message will say so in a clear manner.  Your commit mentions the mouse wheel only in the context of light guns, and until today I was not aware that there are actual input devices that generate non-integer mouse wheel events.  Every PC mouse I've used has a wheel with a detent click per axis change, and the comment in that code even says "per click".

cuavas: > Look, I assume that if people are doing feature work on OSD that the commit message will say so in a clear manner. Your commit mentions the mouse wheel only in the context of light guns,

My commit message was already a mile long and unlikely to be read.  I don’t know if I just forgot to mention that part because there was so much stuff there, or I skipped over it because it was just a simple “feature parity with `OSD=win32`” thing.  It’s been a while.

I’m still mystified that the `#if` and `lround` didn’t tip you off that it was working with the precise value.

> and until today I was not aware that there are actual input devices that generate non-integer mouse wheel events.

macOS has supported “precise” scroll as long as trackpad scroll has been supported.  Windows has theoretically supported it since the late ’90s, but relatively few devices actually generated the events (mostly things like Wacom tablets with touch strips), and relatively few applications handled it properly.  It became widespread after Microsoft added the “precision touchpad” input class around the Windows 7 timeframe.

>  Every PC mouse I've used has a wheel with a detent click per axis change, and the comment in that code even says "per click".

Yeah, it’s rare to find a mouse that doesn’t use a wheel with click detent.  Even some like the Logitech G502 that let you disable the clicks still only report whole click units – it lets the wheel spin freely, but it doesn’t gice you any more precision.  The old Apple Mighty Mouse was a rare example of a mouse that supported precise scroll.  The most common PC input devices that give precise scroll are touchpads (on notebooks, or integrated into keyboards), and Wacom tablets, etc.

The “clicks” unit is weird, and is tied up with the history of scroll wheel support in Windows:
* The initial scroll wheel support in Windows only supported a single axis (vertical), and the initial scroll mouse had a click detent wheel.
* The window events always the scroll delta as a signed integer, where 120 corresponds to one click.  This gave plenty of precision for scrolling by less than a click for future devices with more precise scroll controls.
* There’s a control panel option to configure scroll sensitivity in “lines per click”, which the application then needs to turn into view units (or something like number of items to move selection by) however it chooses to.
* A bunch of applications decided that they’d just divide the scroll delta by 120 to turn it into a number of clicks.  This created a chicken-and-egg situation where you had few input devices with precise scroll controls because few applications supported it, and few applications implemented precise scroll controls few people had devices that made it useful.  It was mostly things intended/used for graphics and CAD where it was supported.  SDL reporting scroll delta in integer “clicks” comes from this era – it just divided the value by 120 and threw away anything more precise.
* Meanwhile, Apple had the best trackpads in the business, and started implementing multi-finger gestures, including precise 2D scrolling, and it took Microsoft years to get their act together and put in the plumbing to support the same kind of thing on Windows.
* We’re now at the point where most well-behaved Windows applications will support precise scroll properly, although there are notable stragglers (e.g. PuTTY, I’m looking at you).
* SDL *eventually* added support for precise scroll, by adding floating-point fields in the same units as the existing integer fields, i.e. 1.0 corresponds to one “click”, and you can have fractional parts thereof.

Anyway, the upshot of this is that scroll control displacement on Windows is measured in units called “clicks”, where one “click” is the precision available with a detent wheel, but the actual internal units have always been a lot more precise.

I didn’t overhaul the way scroll input is reported to the UI layer in MAME when I did pen/touch support.  It needs to be done, but I don’t want to do it in a half-arsed way, so I’m still putting it off.

Also, my stress level is currently off the chart, and getting worse every day, and I still don’t have a release quite ready.  This stuff just keeps making it later.

PR #15029, opened 2026-02-27
change att time from hz and snd clock pachislo_z80

flama12333:
from reference video.
opll clock are taken from gsspade.
att time are guessed to fix slow music.


Osso13:
A PR from another dev is incoming, which will completely substitute this driver. Closing.

PR #15037, opened 2026-02-28
jalmah.cpp: added the actual references for some of the MCUs in filenames

sergiopolog:
instead of using the MCU model number.

- `daireika`: https://i.imgur.com/QV8lFzf.jpeg
- `mjzoomin`: https://i.imgur.com/5phLmf3.jpeg (is it the same one in daireikai but different silkscreen?)
- `kakumei2`: https://i.imgur.com/iXtDaNa.jpeg
- `suchiesp`: https://i.imgur.com/3NgBkWT.jpeg (same ref as `kakumei2`)

### DIFF
--- a/src/mame/jaleco/jalmah.cpp
+++ b/src/mame/jaleco/jalmah.cpp
@@ -1191,7 +1191,7 @@ ROM_START ( urashima )
 	ROM_RELOAD(                   0x40001, 0x20000 )
 
 	ROM_REGION( 0x2000, "mcu", 0 ) /* M50747 MCU Code */
-	ROM_LOAD( "m50747-b84sp.11b", 0x0000, 0x2000, NO_DUMP ) /* Mitsubishi M50747 MCU with 8 Kbytes of internal ROM */
+	ROM_LOAD( "m50747-b84sp.11b", 0x0000, 0x2000, NO_DUMP )
 
 	LOAD_FAKE_MCU_ROM
 
@@ -1245,7 +1245,7 @@ ROM_START( daireika )
 	ROM_RELOAD(                 0x40000, 0x20000 )
 
 	ROM_REGION( 0x2000, "mcu", 0 ) /* M50747 MCU Code */
-	ROM_LOAD( "mo-88015.bin", 0x0000, 0x2000, NO_DUMP ) /* Mitsubishi M50747 MCU with 8 Kbytes of internal ROM */
+	ROM_LOAD( "mo-88015_m50747-a46sp.2c", 0x0000, 0x2000, NO_DUMP )
 
 	LOAD_FAKE_MCU_ROM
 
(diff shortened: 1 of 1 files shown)


src/mame/jaleco/jalmah.cpp
```diff
@@ -1449,7 +1449,7 @@ ROM_START( suchiesp )
 	ROM_LOAD16_BYTE( "2.bin", 0x00000, 0x40000, CRC(42ecf88a) SHA1(7bb85470bc9f94c867646afeb91c4730599ea299) )
 
 	ROM_REGION( 0x2000, "mcu", 0 ) /* M50747 MCU Code */
-	ROM_LOAD( "m50747", 0x0000, 0x2000, NO_DUMP )
+	ROM_LOAD( "m50747", 0x0000, 0x2000, NO_DUMP ) /* Mitsubishi M50747 MCU with 8 Kbytes of internal ROM */
```
angelosa: Isn't it obvious by doing the math? Alternatively there are two comments stating the same thing, back-to-back x6 ...

happppp:
Yahoo JP auctions site is blocked in some countries, I can't see the photos. Could you crop and share the MCU photos here?
Also, why not both labels? Exact MCU type(label) and the internal maskrom serial is not useless information.

cuavas:
> Yahoo JP auctions site is blocked in some countries, I can't see the photos. Could you crop and share the MCU photos here?

It’s not blocked for me, but it wants me to log in to view completed auctions.

happppp:
Kale linked me to one auction that's not blocked, so an example of what I'm suggesting (and if you add the pcb location as extension too):
no: mo-88015.bin
yes: mo-88015_m50747-a46sp.2c

sergiopolog:
> Kale linked me to one auction that's not blocked, so an example of what I'm suggesting (and if you add the pcb location as extension too): no: mo-88015.bin yes: mo-88015_m50747-a46sp.2c

I updated the picture links to ones in imgur for better availability.

I think suffix "-a46sp" is simply a factory marking of the MCU and other devices would potentically have different one, like the one in the picture of `mjzoonin` i uploaded

happppp:
No, -a46sp is the serial for the internal mask ROM (the a46 part).
In other words, if you have: samelabel_differentmaskromserial, it's a high chance the ROM is different.
Or, differentlabel_samemaskromserial = actually same internal ROM.

angelosa:
Driver itself hooks 4 different MCU behaviors, it may be config pin strapped but there are more chances that the MCU is different for each yes.

sergiopolog:
From the pictures I found for `kakumei2` and `suchiesp`, the serial number for the mask ROM is sanded or not visible at all.
I found another auction of `suchiesp` in which the silkscreen looks OK, but pictures are too small to get the exact value: https://buyee.jp/item/jdirectitems/auction/p1220588466

sergiopolog:
Additionally, here is an auction of `kakumei` with same situation. Silkscreen is there but not really visible due to low res pictures:  https://buyee.jp/item/jdirectitems/auction/j1221058560

Edit: nevermind,  in this case, the serial number is already in the driver

happppp:
If it's not known (or sanded off, or sticker on top), then of course that part of the label can be omitted.

src/mame/jaleco/jalmah.cpp
sergiopolog: removed

sergiopolog:
updated MCU labels

PR #15038, opened 2026-02-28
heathzenith/z100.cpp: Fixed 6845 GVRAM addressing

BlueRain-debug:
Fixed the abnormal text scrolling issue in 32k/64k vram modes.

Before (64K):

Before (32K):

After:

### DIFF
--- a/src/mame/heathzenith/z100.cpp
+++ b/src/mame/heathzenith/z100.cpp
@@ -430,7 +430,7 @@ MC6845_UPDATE_ROW(z100_state::update_row)
 			else
 			{
 				for (int i = 0; i < 3; i++)
-					dot |= ((m_gvram[((x + ma) & amask) << 4 | (ra & 0xf) | (0x10000 * i)] >> (7 - xi)) & 1) << i; // b, r, g
+					dot |= ((m_gvram[((x + ma + (m_start_addr << 4)) & amask) << 4 | (ra & 0xf) | (0x10000 * i)] >> (7 - xi)) & 1) << i; // b, r, g
 
 				if (x == cursor_x)
 					dot ^= 7;


cuavas:
@mgarlanger can you look at this?  Does this look like a real fix, or is it based on a misunderstanding?  I’m not qualified to assess it.

mgarlanger:
Sure, I can see if my manuals/schematics for the Z100 have any details about this and test the changes locally, but won't have time until tonight or tomorrow to do that. If I'm remembering correctly, the scrolling on the z100 was messed up last I tried it.

mgarlanger:
@cuavas I didn't find anything in the documentation that clears this up. And I'm not as familiar with the Z100 as the H/Z-89. 

I've tested with and without this change on MacOS, and it is not clear which approach is correct. I see the z100 behavior change based on whether I press a character key to get past the "System doesn't work" dialog, vs. clicking on the trackpad ( or use a key that doesn't generate a character like the shift key). 

If I press key that generates a character, the system comes up without any error. With the existing code, scrolling works, but with the new code scrolling is messed up. 
Screenshot 2026-03-02 at 12 04 50 AM

If I use the trackpad click or the shift key to get past the dialog, the system comes up with an "Interrupt Circuit Error. Chip U208".
Screenshot 2026-03-01 at 10 10 24 PM
The existing code shows scrolling problems, but the new code scrolls properly.

Also note, that using the menu item "Reset System", where there is also no key input to the system, the system shows the "Interrupt Circuit Error. Chip U208", has the new code properly scrolling, and the existing code not scrolling correctly. 

I'm not 100% sure, but I think since the existing code works properly when there is not an interrupt error, that is the correct behavior and this PR is just working when the z100 is already in a bad state due to the interrupt error.

Is it design intent that the key used to get passed the dialogs are passed to the emulated machine? It seems like the UI should eat them.

BlueRain-debug:
> @cuavas I didn't find anything in the documentation that clears this up. And I'm not as familiar with the Z100 as the H/Z-89.
> 
> I've tested with and without this change on MacOS, and it is not clear which approach is correct. I see the z100 behavior change based on whether I press a character key to get past the "System doesn't work" dialog, vs. clicking on the trackpad ( or use a key that doesn't generate a character like the shift key).
> 
> If I press key that generates a character, the system comes up without any error. With the existing code, scrolling works, but with the new code scrolling is messed up. Screenshot 2026-03-02 at 12 04 50 AM
> 
> If I use the trackpad click or the shift key to get past the dialog, the system comes up with an "Interrupt Circuit Error. Chip U208". Screenshot 2026-03-01 at 10 10 24 PM The existing code shows scrolling problems, but the new code scrolls properly.
> 
> Also note, that using the menu item "Reset System", where there is also no key input to the system, the system shows the "Interrupt Circuit Error. Chip U208", has the new code properly scrolling, and the existing code not scrolling correctly.
> 
> I'm not 100% sure, but I think since the existing code works properly when there is not an interrupt error, that is the correct behavior and this PR is just working when the z100 is already in a bad state due to the interrupt error.
> 
> Is it design intent that the key used to get passed the dialogs are passed to the emulated machine? It seems like the UI should eat them.

Since I always use the -debug parameter to start my MAME (which causes the NOT WORKING prompt to be skipped), I didn't really notice this issue, so is this the expected behavior? I'm thinking it might actually be a race condition?

There seems to be another noteworthy thing: when I press a key to bypass the NOT WORKING prompt, a flashing cursor indicator appears, whereas it doesn't when the U208 error occurs.

image

Is there a easy way for us to trigger a circuit fault on a physical z100 device so that we can observe the expected behavior?

PR #15039, opened 2026-03-01
Simplify handling of clang targets, fix clang build on aarch64 and riscv64

belegdol:
This PR fixes building on `riscv64` and `aarch64` using clang by ensuring `-m64` is getting passed instead of `-m32`.
I am not quite happy with the amount of copy-paste needed. The issue is that gcc does not support `-m64` parameter on `aarch64` or `riscv64`, meaning that the `ARCHITECTURE` parameter needs to be set to an empty value in order to prevent genie from adding the parameter to the generated makefiles.
Unfortunately, clang logic relies on `ARCHITECTURE` being set to `x64` in order to pick the 64-bit config, and falls back to 32-bit target otherwise. While clang does support `-m32` and `-m64` on `aarch64` and `riscv64`, it does not help if we end up picking the wrong one.
I can think of the following alternative approaches:

1. Set `ARCHITECTURE` to blank after the clang logic block
2. Explicitly check for `x86` and only pick the 32-bit target if it is found, fall back to 64-bit otherwise. This might make sense given where 32-bit support is these days in general.

Input is welcome.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1007,17 +1007,7 @@ $(info GCC $(GCC_VERSION) detected)
 else # CLANG_VERSION
 $(info Clang $(CLANG_VERSION) detected)
 ifneq ($(TARGETOS),asmjs)
-ifeq ($(ARCHITECTURE),_x64)
-ARCHITECTURE := _x64_clang
-else ifneq ($(filter aarch64%,$(UNAME_M)),)
-ARCHITECTURE := _arm64_clang
-else ifneq ($(filter aarch64%,$(UNAME_P)),)
-ARCHITECTURE := _arm64_clang
-else ifneq ($(filter arm64%,$(UNAME_M)),)
-ARCHITECTURE := _arm64_clang
-else
-ARCHITECTURE := _x86_clang
-endif
+ARCHITECTURE := $(ARCHITECTURE)_clang
 endif # asmjs
 endif # CLANG_VERSION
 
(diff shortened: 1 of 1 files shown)


cuavas:
AArch64 *is* ARM64.  Why are we back to the situation where different distros call the same target different things, like x64 (Windows) vs x86_64 (most Linux, macOS) vs amd64 (some BSDs)?  We’re already building with clang on AArch64 for macOS and Windows at least, so the logic works for those targets.

The whole `ARCHITECTURE` thing is a mess.  I wish there was a better way to debug make files to trace which lines get evaluated, because one way or another the current logic is working for clang ARM64 builds on macOS and Windows.  We definitely don’t need an artificial “aarch64” that’s different to “arm64” anyway.

cuavas:
Linux didn’t have an `_arm64_clang` target like macOS and OpenBSD had, so I added it in daf683305d8fa8832a838f522b38e9b3594d3002, as well as copying some of the uname parsing logic from higher up in the file.

belegdol:
Maybe we can at least reduce the amount of copy paste. Genie only supports three targets for GCC: native, x32 and x64. So if there are no differences between the different 32 and 64bit targets, maybe it would be possible to have different variants have fewer results:
https://github.com/mamedev/mame/blob/019bbc797f94efcaadca45f250bb34b5e97571f3/3rdparty/genie/src/tools/gcc.lua#L67-L77

belegdol:
This is now less eyesore-inducing. I have tested on Fedora with x86_64, aarch64 and riscv64, each with gcc and clang. Compilation appears to proceed fine. Let's wait for CI to finish to see how other configs fare.

PR #15041, opened 2026-03-01
nec/piet1.cpp: Add skeleton driver

qufb:
NEC Electronic Tool PI-ET1 [QUFB]

### DIFF
--- a/src/mame/nec/piet1.cpp
+++ b/src/mame/nec/piet1.cpp
@@ -19,6 +19,7 @@
 
 #include "cpu/z80/z80.h"
 #include "machine/ram.h"
+#include "machine/upd4991a.h"
 
 #include "emupal.h"
 #include "screen.h"
@@ -34,6 +35,7 @@ public:
 		, m_banks(*this, "bank%u", 0U)
 		, m_maincpu(*this, "maincpu")
 		, m_ram(*this, RAM_TAG)
+		, m_rtc(*this, "rtc")
 		, m_screen(*this, "screen")
 	{ }
 
@@ -46,6 +48,7 @@ private:
 	required_memory_bank_array<4> m_banks;
 	required_device<cpu_device> m_maincpu;
 	required_device<ram_device> m_ram;
+	required_device<upd4991a_device> m_rtc;
 	required_device<screen_device> m_screen;
 
 	void bank_switch(u8 bank, u8 entry);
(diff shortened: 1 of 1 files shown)


src/mame/nec/piet1.cpp
```diff
@@ -0,0 +1,136 @@
...
+/***************************************************************************
+
+    Skeleton driver for NEC PI-ET1.
+
+    - PCB Markings: NEC PWD-850 707V / 72408502 / W03-104PE
+    - Cart Slot Markings: JAE JC20-C45PD-F1-A2-1 9X1
+    - IC101: Hitachi 9M3T HC74N (Dual D-Type Flip-Flop)
+    - IC106: NEC D70008A-8 (µPD780 CPU)
+    - IC109: NEC D65013GFE63 (Gate Array)
+    - IC110: NEC D4991A (4bit Calendar Clock)
```
angelosa, line 12: I guess this is an uPD4991A, reused by PC-98HA.

qufb: Yes, it should be the same device.

src/mame/nec/piet1.cpp
```diff
@@ -0,0 +1,141 @@
...
+	m_banks[3]->configure_entries(0, 32, rom, 0x4000);
+
+	for (size_t i = 0; i < 4; i++) {
+		m_banks[i]->set_entry(0);
+	}
+}
+
+void piet1_state::piet1(machine_config &config)
+{
+	Z80(config, m_maincpu, 8_MHz_XTAL); // FIXME: Guessed (D70008A-6 = 6MHz).
```
angelosa, line 114: So this is the same Z80 variant used by late PC-8801, 8 MHz is correct (no idea if it has a switch to make it run at 4).
Other info at http://videogamekraken.com/pi-et1-by-nec

qufb: Thanks. Does that PC-8801 switch toggle between distinct oscillators, or modifies the operating current of a single oscillator?

Without testing, it isn't clear to me if this one is fixed at 4MHz or 8MHz.

angelosa: I don't really know details, all I know is that there's a dip or a switch changing between the two clocks (there's no schematics available for MA/FE etc.)

qufb: Ok, let's assume it's at the max frequency for now.

PR #15042, opened 2026-03-02
Update upd78k3.cpp (Documentation)

pxlsr:
My first MAME commit. Beginning documentation of processor, uPD78312

### DIFF
--- a/src/devices/cpu/upd78k/upd78k3.cpp
+++ b/src/devices/cpu/upd78k/upd78k3.cpp
@@ -226,15 +226,17 @@ void upd78k3_device::state_string_export(const device_state_entry &entry, std::s
 	{
 	case STATE_GENFLAGS:
 		str = string_format("RB%d:%c%c%c%c%c%c%c%c",
-				(m_psw & 0x7000) >> 12,
+				(m_psw & 0x7000) >> 12,			//16 Bit Register
 				BIT(m_psw, 15) ? 'U' : '.',
 				BIT(m_psw, 7) ? 'S' : '.',
 				BIT(m_psw, 6) ? 'Z' : '.',
 				BIT(m_psw, 5) ? 'R' : '.',
 				BIT(m_psw, 4) ? 'A' : '.',
-				BIT(m_psw, 3) ? 'I' : '.',
-				BIT(m_psw, 2) ? 'V' : '.',
-				BIT(m_psw, 0) ? 'C' : '.');
+				BIT(m_psw, 3) ? 'I' : '.',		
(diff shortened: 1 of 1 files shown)


happppp:
That won't work. %c = char, not multiple characters.
Even if it did work, single char flags are prefered for the debugger view status register.

Brief comments describing the flags is ok. The "lol" part is not, the clarification that a 16-bit register is 16-bit is also unneeded.

pxlsr:
> That won't work. %c = char, not multiple characters. Even if it did work, single char flags are prefered for the debugger view status register.
> 
> Brief comments describing the flags is ok. The "lol" part is not, the clarification that a 16-bit register is 16-bit is also unneeded.

would you like me to break down the PSW into its high and low order bits?
Also I apologize about the informalities, its been a while.

happppp:
No. Looking at upd78k3.cpp, it already has entries for PSWL and PSWH. I'd leave the flags the way they are.

pxlsr:
> No. Looking at upd78k3.cpp, it already has entries for PSWL and PSWH. I'd leave the flags the way they are.

happppp:
I don't think I treated you unfairly.
And if you're passionate about documenting this CPU, I'm not sure what needs documenting. Other than perhaps a short description of the CPU at the comment block at the top of the file(s). The emulation is very preliminary.

pxlsr:
> I don't think I treated you unfairly. And if you're passionate about documenting this CPU, I'm not sure what needs documenting. Other than perhaps a short description of the CPU at the comment block at the top of the file(s). The emulation is very preliminary.

Considering this is like the first major work I'm doing in a long time, some constructive pointers would be great. I got a brain full of uPD78312 knowledge including its opcodes, specifically for use in the D-50 as emulation stuff. Just tell me where I need to start because im struggling to find code examples on the wiki
(I've got a few cobwebs in my programming knowledge as its been ages since I worked on a codebase)

happppp:
If your intention was to flesh out the emulation, perhaps it's better to work on your local github. Once you've hit a milestone you're happy with, then make the PR.

pxlsr:
> If your intention was to flesh out the emulation, perhaps it's better to work on your local github. Once you've hit a milestone you're happy with, then make the PR.

so the issue at hand is pushing broken code right? Is this what I use milestones for?

happppp:
Not just that it was broken. If your plan was to write the emulation, then this PR was made too hastily.
Milestones = sub goals.
eg.
1: write disassembler (already there)
2: emulate most of the opcodes
3: emulate internal MCU peripheral A
etc.

If you want to understand how MAME emulates CPUs in general, src/devices/cpu/... source code is a better place to browse around than MAME's documentation.

PR #15045, opened 2026-03-02
nakajima/nakajies.cpp: Use an array of views.

wilbertpol:


cuavas:
Ah right, this is one of those things that's nicer now that we require GCC 10.3

PR #15051, opened 2026-03-03
Rewrite translation table for GRiD Compass keyboard

vklachkov:
This PR rewrites the key mapping for the GRiD Compass laptop. The previous implementation did not map buttons with the Code and Control keys pressed, and there were errors in the mapping when Shift was held down. The new implementation covers all keycodes with all modifiers. The values are taken from the official documentation:

[GRiD-OS Reference Appendix A - Key scan codes.pdf](https://github.com/user-attachments/files/25720704/GRiD-OS.Reference.Appendix.A.-.Key.scan.codes.pdf)

### DIFF
--- a/src/mame/gridcomp/gridkeyb.cpp
+++ b/src/mame/gridcomp/gridkeyb.cpp
@@ -2,11 +2,9 @@
 // copyright-holders:Vas Crabb, Sergey Svishchev
 /***************************************************************************
 
-    GRiD Compass keyboard HLE, derived from generic_keyboard
+    GRiD Compass keyboard HLE
 
-    Keycodes from "6 - Interceptor, CCProm, Utils, OS - Sysgen/UTILS/DVLEXEC.PLM"
-
-    Multiple shift key combos not simulated yet.
+    Keycodes from "GRiD-OS Reference Appendix A - Key scan codes"
 
 ***************************************************************************/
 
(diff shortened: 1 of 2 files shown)


happppp:
The keyboard MCU internal ROM is still not dumped yet, right? (for eventual low level emulation) So that comment can stay.
Isn't it more readable to keep the chars in the table instead of hex values?

translation part can be simplified such as:
```
int const shift(CAPS_TABLE[row][col] ? (BIT(modifiers, 1) ^ BIT(modifiers, 2)) : 0);
int const ctrl(BIT(modifiers, 0));
int const meta(BIT(modifiers, 3));
int const map(ctrl << 2 | meta << 1 | shift);
```
(I used var(x) style instead of var = x due to surrounding source code)

vklachkov:
> The keyboard MCU internal ROM is still not dumped yet, right?

It has already been dumped and even partially reversed by @JDat. But I decided not to implement proper keyboard emulation for now, and just update the table.

> translation part can be simplified such as...

Thank you for the comment, fixed it.

>  Isn't it more readable to keep the chars in the table instead of hex values?

Hmm, I think it's more correct to leave the values in hex — it makes it clear how the table from the documentation correlates with what's written. When I see characters (like 'A', 'B' and so on), I always wonder if the encoding is correct and if the character codes actually match.

If MAME's the style guide requires store values differently, I will fix that. Not a problem.

happppp:
'missing dump' in this case means whether or not the keyboard MCU dump is in MAME. Or did I not see it when I glanced over the driver and keyboard device source code?

vklachkov:
Ah, there's no dump in MAME and no code around. Instead of a silent todo in comment, I'll probably create an issue.

happppp:
If you want to report an issue about it, that's fine. It doesn't invalidate the TODO note though. I mean, having both a TODO note and issue (or MameTesters bug report) is common. I believe a TODO list in drivers is useful.

vklachkov:
I'll create the issue a little later. Are there any comments on the code?

happppp:
Other than what I already wrote?
Well, the simplified mapping: I don't have any issue that you didn't copy my suggestion 1:1, but are you aware now there's pointless conversions from binary->bool followed by bool->binary?

As for whether or no ASCII chars in the table: if you think raw hex is better, I don't really mind. I'd be more picky if it was a driver by me.

vklachkov:
I decided left the code for working with modifiers "as is", because it's written exactly the same way in other keyboards. And I added the casting for the map because I forgot about Integral Promotion in C++ and thought I needed to do the conversion manually.

MAME has a unclear code style, and I try to make my patches in accordance with how the code is already written, just to fix things that are clearly broken and annoying.

happppp:
Yes, MAME doesn't have a single code style. It's basically up to main copyright holder of a source file. And you're correct to keep the code style of new additions the same as the rest of the file, otherwise it'll become an inconsistent mess.

happppp:
Be careful doing binary math on a bool (the shifts in this case). Some compilers may give a warning.
I'd still recommend changing the bools to ints.

I'll leave the merging of this PR up to @cuavas since he's one of the copyright holders of this keyboard emulation.

src/mame/gridcomp/gridkeyb.cpp
```diff
@@ -276,9 +292,10 @@ bool grid_keyboard_device::translate(u8 code, u16 &translated) const
 	bool const ctrl(modifiers & 0x01U);
 	bool const meta(modifiers & 0x08U);
 
-	unsigned const map(ctrl ? (meta ? 4U : 3U) : meta ? 2U : shift ? 1U : 0U);
-	u16 const result(TRANSLATION_TABLE[map][row][col]);
-	if (result == u8(~0U))
+	int const map(ctrl << 2 | meta << 1 | shift);
+	u8 const result(TRANSLATION_TABLE[map][row][col]);
```
cuavas: Please don't add code that implicitly converts `bool` to `int`.  You can express it as `((ctrl ? 0x4 : 0) | (meta ? 0x2 : 0) | (shift ? 0x1 : 0))` or use some other approach, but implicit `bool`-to-`int` is one of those things I wish C++ didn't do.

vklachkov:
I've redone the PR.

First, I fixed the mistakes I made when copying data from the table.

Second, I verified the table against the values from the keyboard firmware. And I found a difference for the Escape key code when Code and Shift are pressed -- the document says one value, but the keyboard firmware says another.

Third, I removed the four separate tables for when Control is pressed and replaced the human-readable codes with characters. The GRiD Compass definitely uses ASCII and this is acceptable.

I decided to simply address the comment about type casting by adding explicit casts to int. I hope this is the only style comment.

galibert:
Since you seem to know that hardware, do you know what's going on with the weird patching of 1139 bios to run on the 1129?  Is that something people did?

galibert:
Ohhhh, I missed the NO_DUMP, I see.

vklachkov:
Hi, @galibert!

I'll answer the question, but if you want to talk more about Compass or other GRiD machines, write to me either on the VCF forum ([there's a thread about MAME](https://forum.vcfed.org/index.php?threads/mame-video-on-status-of-gridcomp-driver-emulation.1255781/)), or in the [Telegram group of my friends](https://t.me/gridcompasschat).

Someone told me about these patches. But I don't remember now. I have a dump from a Compass II 1129 and (I think) the sources for this CCPROM. I'll need to take a look sometime at what exactly this suspicious patch is for and what it fixes there.

Unfortunately, I don't understand well how the Compass works internally: I have some knowledge about the first generation Compass and a 1109 on hand. But how the second generation works, why this workaround is needed just for 1129 and 1139 -- I have no idea. (UPD: Everything works without this patch)

My small PRs in MAME are just fixes for the most annoying things and my attempt to understand how the Compass works internally. So that later it will be easier to continue its reverse engineering. So that it would be possible to make a replica. For many reasons, really. I just love GRiD laptops.

vklachkov:
Returning to the keyboard. I understand that this PR doesn't perfectly emulate the original Compass keyboard. To get everything right, we need to implement the keyboard delay and keyboard repeat rate parameters. Also need to properly implement DMA and watchdog. There's still a lot of work. But this PR at least creates a proper scan code table and allows me to actually use the emulator. Before this, even navigating with Executive~Run~ was annoying -- the arrow keys didn't work with shifts and not all characters were entered correctly!

galibert:
From looking at it better, we only have in mame the dumps for the 1101 and the 1139.  The patches are hacking up the 1139 bios to make it run in the 1129.

If you have access to more dumps than that (and you seem to), it would be nice to add them.

vklachkov:
> If you have access to more dumps than that (and you seem to), it would be nice to add them.

No problem. Where to add them? How to add them? I don't know anything about MAME and how you work with ROMs. I got copies from my friends. I have a 1101 rev3, 1101 rev4, 1109, 1129, and 1139. I'm planning to get ROMs from other models by summer, if possible.

I'll probably make a separate PR that removes the hack and puts the correct names and checksums for the dumps. The only thing left is to figure out how to store the dumps. Right now I keep them on [gridrepository.org](https://gridrepository.org/) (I haven't made a frontend for it yet), but who would think to look for them there?)

galibert:
When a rom is added to mame it magically appears pretty much everywhere.  It's a nice method to ensure their preservation too :-)

vklachkov:
I didn't have to rebase this PR after https://github.com/mamedev/mame/pull/15079. If there are no comments here, could you please merge it?

PR #15054, opened 2026-03-03
Add cv1k SH7709S cache handling behavior for slowdown emulation

jesus-ramos:
Handles some basic caching for the sh7709s used for cv1k titles so that they run at more appropriate speeds

Seems pretty close to pcb footage now for cv1k titles without having to mess with the cpu% anymore

### DIFF
--- a/src/devices/cpu/sh/sh7709s.cpp
+++ b/src/devices/cpu/sh/sh7709s.cpp
@@ -2,6 +2,21 @@
 // copyright-holders:Jesus Ramos
 
 // SH 7709S experimental cache/memory timing
+// Some values/timing tables are hardcoded
+// (such as no bank mode timing/sdram access logic, and some area mappings)
+// for cv1k emulation specifically
+
+// Unimplemented functionality:
+// - Cache invalidation and write locking of cache lines
+// - Uncached access penalty
+// - Prefix instruction fetch timing + cache handling
+// - Checks for certain area specifics : bank mode timing for sdram access, area check for sdram, area check for burst mode + burst size
+// - TLB simulation + timing and lookups
+// - Branch predicition/mispredict penalties - Needs a branch target buffer and some basic tracking, didn't make a huge difference for cv1k when I threw one together
+// - Full pipeline simulation:
+//   Some penalties should overlap with following instructions that don't involve bus access
+//   Writeback timing for certain operations should also overlap in some cases
+//   If the CPU supports critical word first the instruction that takes the miss should be able to start as soon as that address is read in from memory even if there are further words to be read
 
 #include "sh7709s.h"
 #include "sh3comn.h"
@@ -128,12 +143,20 @@ static bool is_sdram_region(uint32_t address)
 {
 	unsigned int area = get_area(address);
 
+	// Hardcoded for now, assumes area 2 and 3 are both mapped to SDRAM
 	return area == 2 || area == 3;
+}
+
+bool can_use_burst(uint32_t address)
+{
+	// Assumes only sdram can burst read/write
+	return is_sdram_region(address);
 }
 
 #define CACHE_LINE_BURST_READ_PENALTY (6) // 2 base + 4 to burst read 16 bytes for a cache line
 #define CACHE_LINE_BURST_WRITE_PENALTY (4) // Only the 4 data cycles for the cache line
 #define CACHE_MISS_STALL_PENALTY (4) // 3 base (tag lookup, compare, etc...) + 1 for BSC sync
+#define BUS_ACCESS_PENALTY (2)
 
 static unsigned int get_wcr1_timing(uint32_t address, uint16_t wcr1)
 {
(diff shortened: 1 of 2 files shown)


MooglyGuy:
Looks alright to me, though I'd like @p1pkin to have a look at it as well.

By the way, why use divides and modulos instead of shifts and masks for values that are powers of two?

jesus-ramos:
No particular reason, just seemed convenient when writing some of those parts.

jesus-ramos:
Last set of fixes shores up some missing slowdown I noticed in SDOJ s4 midboss where things like the 2nd to last pattern where the bombs drop had no slowdown due to using burst timings for areas that didn't support burst reads and needed 4 separate wait states to fetch the cache line.

Some other minor fixes in there as well with the writeback logic using the wrong flag and the area detection wasn't stripping the top bits.

mamehaze:
nice, any chance you could also look at SH2 cache logic? some Saturn games depend on it at the very least (Astal) as do a few 32X titles which write to memory that doesn't exist, but expect to be able to read it back in tight loops as it ends up in cache regardless (Pitfall for edge-of-screen clipping)

(if these things are beyond the scope of what you're interested in / can handle, fair enough, but the SH family in general need improvements and better support for things like Cache and MMU)

angelosa:
fwiw it's not only Astal that requires cache, cfr. Mednafen list:
```
-------------------------------------------------------------------------------
Rely on instruction fetches enforcing dcache coherency by
cache line replacement(or something roughly similar) :(
-------------------------------------------------------------------------------
Area 51 (USA/Europe)
Clockwork Knight 2 (USA)
DeJig: Lassen Art Collection (Japan)
Doraemon: Nobita to Fukkatsu no Hoshi (Japan)
Dragon Force II (Japan) - Battles(broken math)
Father Christmas (Japan)
Fighting Vipers (Japan/Europe/USA)
Formula Grand Prix: Team Unei Simulation (Japan)
Golden Axe The Duel (Japan/Europe/USA)
House of the Dead (Japan)
Linkle Liver Story (Japan)
Mahjong Doukyuusei Special (Japan)
Mr. Bones - Intro FMV
Mujintou Monogatari R: Futari no Love Love Island
Ronde (Japan)
Sega Saturn Choice Cuts (USA) - Broken FMV playback(except intro FMV).
Segakore Sega Bible Mogitate SegaSaturn - Zero Divide demo, especially.
Shunsai (Japan)
Spot Goes to Hollywood (USA/Japan)
Street Fighter Zero (Japan)
Street Fighter Zero 3 (Japan)
Super Puzzle Fighter II Turbo (USA)
Theme Park (Europe)
Thunder Force Gold Pack 1 (Japan)
Thunder Force Gold Pack 2 (Japan)
Virtua Fighter Kids (Java Tea Original)
Virtual Mahjong (Japan)
Virtual Volleyball (Japan)
WipEout (Japan/Europe/USA)
World Series Baseball (USA/Japan)
Zero Divide


-------------------------------------------------------------------------------
Rely on executing directly out of cache
-------------------------------------------------------------------------------
Mr. Bones - (0xC0000064)


-------------------------------------------------------------------------------
Rely on (data) cache incoherency
-------------------------------------------------------------------------------
Astal
```
Out of this list, Falcom Classics volume 1: Dragon Slayer may also really need cache, master CPU pitfalls slave CPU code without stopping it during loading.

jesus-ramos:
I probably don't have time for other variants but the cache logic itself can probably be reused if they all use the same replacement algorithm and can be split for the separate dcache/icache variants. There might need to be some extra plumbing for titles that use PREF instructions or if any use the cache locking stuff that I didn't add but maybe safe to assume that not too many games if any use the latter.

Not sure if the layout for WCR, MCR registers is the same or if the timings listed in the datasheets is consistent either so not sure how much of that code can be reused.

p1pkin:
@jesus-ramos can you please add a bit of documentation with the list of not implemented cache and BSC features, and hardcoded CV1000-specific things, for example I see Area 2 hardcoded as SDRAM while it may be configured to be ordinary memory space.
also, I can't see separate V and U tag bits, is it implemented in some hacky way to make games works with "close enough" slowdowns?

@MooglyGuy I'm mainly familiar with SH-4 caches (direct mapped type), while SH-3 and SH-2 uses other cache type, so I don't have enough practice with it.
but, SH 3 and 2 caches very similar, mostly same, so, I think it would be not hard to adopt implementation from one to another.

jesus-ramos:
The U bit is just tracked as the dirty field, I could have packed it into a single uint32 but kept it separate just for readability. The V bit isn't tracked since I don't think I saw any cache invalidation so aside from boot entries will stay valid at least for cv1k.

Yeah I can add some info about what's hardcoded so far, I did plan to fix a couple of those as well in followups in case someone wanted to reuse certain bits and add in the other cycle tables for WCR depending on area configs.

jesus-ramos:
@p1pkin Added some comments, let me know if there's something I missed that you'd like called out or updated

Thanks to the reviewers for taking the time to look at this as well

PR #15055, opened 2026-03-03
cpu/sharc: implement fixed point clip and saturation arithmetic in interpreter

gm-matthew:
The Sega Model 2B geometrizer code uses the fixed point clip operation and enables saturation arithmetic, so I have implemented these features in the interpreter; they are yet to be implemented in the recompiler.

In addition, I have made the following fixes (all interpreter):

- Use correct logic for generating carry/overflow flags in add-with-carry and subtract-with-borrow operations
- Use `truncf` (round to zero) rather than `floorf` (round to negative infinity) when MODE1 truncate bit is set
- Calculate floating point clip correctly for Fx < 0 (previous implementation does not clamp Fx if it is negative)

### DIFF
--- a/src/devices/cpu/sharc/sharcops.hxx
+++ b/src/devices/cpu/sharc/sharcops.hxx
@@ -60,8 +60,8 @@
 		}                                                   \
 	}
 
-#define SHARC_COND_LT  (m_core->astat & AF ? m_core->astat & AN && !(m_core->astat & AZ) : ((m_core->astat & AN) != 0) != (m_core->astat & AV && !(m_core->mode1 & MODE1_ALUSAT)))
-#define SHARC_COND_LE  (m_core->astat & AZ || (m_core->astat & AF ? m_core->astat & AN : ((m_core->astat & AN) != 0) != (m_core->astat & AV && !(m_core->mode1 & MODE1_ALUSAT))))
+#define SHARC_COND_LT  ((m_core->astat & AF) ? ((m_core->astat & AN) && !(m_core->astat & AZ)) : (bool(m_core->astat & AN) != ((m_core->astat & AV) && !(m_core->mode1 & MODE1_ALUSAT))))
+#define SHARC_COND_LE  ((m_core->astat & AZ) || ((m_core->astat & AF) ? (m_core->astat & AN) : (bool(m_core->astat & AN) != ((m_core->astat & AV) && !(m_core->mode1 & MODE1_ALUSAT)))))
 
 /*****************************************************************************/
 


src/devices/cpu/sharc/compute.hxx
```diff
@@ -111,14 +112,15 @@ void adsp21062_device::compute_add(int rn, int rx, int ry)
...
-		fatalerror("SHARC: compute_add: ALU saturation not implemented!\n");
-
 	CLEAR_ALU_FLAGS();
 	SET_FLAG_AN(r);
 	SET_FLAG_AZ(r);
 	SET_FLAG_AV_ADD(r, REG(rx), REG(ry));
 	SET_FLAG_AC_ADD(r, REG(rx), REG(ry));
+
+	if (m_core->mode1 & MODE1_ALUSAT && m_core->astat & AV)
+		r = (int32_t)r < 0 ? std::numeric_limits<int32_t>::max() : std::numeric_limits<int32_t>::min();
```
cuavas: Given the number of times this is repeated, maybe it should have a macro helper like the `SEG_FLAG_` stuff.

src/devices/cpu/sharc/compute.hxx
```diff
@@ -402,7 +425,7 @@ void adsp21062_device::compute_fix(int rn, int rx)
 	r_alu.f = FREG(rx);
 	if (m_core->mode1 & MODE1_TRUNCATE)
 	{
-		alu_i = int32_t(floorf(r_alu.f));
+		alu_i = int32_t(truncf(r_alu.f));
```
cuavas: This change is not correct - see the manual's description for FIX and TRUNC:

> Converts the floating-point operand in Fx to a twos-complement 32-bit fixed-point integer result. If the MODE1 register TRUNC bit=1, the FIX operation truncates the mantissa towards –Infinity. If the TRUNC bit=0, the FIX operation rounds the mantissa towards the nearest integer. The TRUNC operation always truncates toward 0. Note that the TRUNC bit does not influence operation of the TRUNC instruction.

FIX gives you a choice of IEE754 "to nearest" and "towards negative infinity", while TRUNC gives you IEEE754 "towards zero".

src/devices/cpu/sharc/compute.hxx
```diff
@@ -432,7 +455,7 @@ void adsp21062_device::compute_fix_scaled(int rn, int rx, int ry)
 	r_alu.r = SCALB(m_core->r[rx], ry);
 	if (m_core->mode1 & MODE1_TRUNCATE)
 	{
-		alu_i = int32_t(floorf(r_alu.f));
+		alu_i = int32_t(truncf(r_alu.f));
```
cuavas: Same as the other version of FIX  - it rounds towards negative infinity in truncate mode.

cuavas:
Also note that the ALUSAT state affects what the LT, LE, GE and GT conditions for branches and loops do.  If the DSP program is using it, that needs to be implemented.  The conditions are:

* LT: !AF and (AN xor (AV and !ALUSAT)) or (AF and AN and !AZ)
* LE: (!AF and (AN xor (AV and !ALUSAT)) or (AF and AN)) or AZ
* GE is the complement of LT
* GT is the complement of LE

You could express them in a more C++ way like:

* LT: AF ? (AN && !AZ) : (AN != (AV && !ALUSAT))
* LE: AZ || (AF ? AN : (AN != (AV && !ALUSAT)))

Note that this also requires AF to be  set correctly, which I'm not sure we're getting right (the recompiler definitely isn't).


The current code is here: https://github.com/mamedev/mame/blob/mame0286/src/devices/cpu/sharc/sharcops.hxx#L1193

It oversimplifies it to:

* LT: AN && !AZ
* LE: AZ || AN

So it's only correct following a floating point ALU operation.  It doesn't even try to do the right thing after integer ALU operations.

src/devices/cpu/sharc/compute.hxx
gm-matthew: I must have missed that particular bit of the manual... everywhere else in the manual it says that the TRUNC bit being set results in rounding towards zero. I can only presume that the FIX operation in particular truncates _after_ negating the mantissa if the value is negative, which results in rounding to negative infinity.

src/devices/cpu/sharc/compute.hxx
gm-matthew: Perhaps something like this?
```
#define SATURATE(r)    do { r = (int32_t)r < 0 ? std::numeric_limits<int32_t>::max() : std::numeric_limits<int32_t>::min(); } while (false)
```

gm-matthew:
> Also note that the ALUSAT state affects what the LT, LE, GE and GT conditions for branches and loops do. If the DSP program is using it, that needs to be implemented.

I've checked all of the Model 2B geometrizer programs and they all use the GE condition but only after performing a floating point operation, so the current condition code will work.

However, some of the Model 2B copro programs do use the LT, LE, GE and GT conditions after fixed point operations. Because ALUSAT is not set for the copro programs, the condition result will be incorrect if overflow has occurred (AV), though I am not sure if this actually occurs in practice.

It's probably a good idea to implement it anyway.

I've also realized that the current commit saturates the result after setting the AN flag which will lead to incorrect results, e.g. overflow leading to a negative result (AN is set) will be saturated to the maximum positive value (AN should not be set). AN should represent the post-saturated result.

src/devices/cpu/sharc/compute.hxx
cuavas: The TRUNC but being set causes the results of arithmetic operations to be rounded towards zero. The FIX instruction is different.

We aren't currently trying to emulate the effect of the TRUNC bit on arithmetic. Actually emulating that properly for SHARC is difficult.

src/devices/cpu/sharc/compute.hxx
cuavas: Yeah, that isn't an unreasonable name.  I can change it later if necessary anyway.

cuavas:
Thanks.  I’ll do the corresponding recompiler changes when I get a chance.  It might take me a while as doing it properly is non-trivial, and as you might have noticed, I’ve got plenty of other recompiler-related work to do.  I also don’t anticipate being able to get hours-long chunks of time to work on MAME in the next week or so.

cuavas:
I’m slowly getting through stuff.  74cb29c61ada4059dae146f105d0fc5df972723a adds support for several fixed point operations to the SHARC recompiler.

One thing I noticed while doing this was that the recompile was setting the AC flag incorrectly for subtraction.  SHARC, like 6502 and ARM (but unlike x86, PowerPC and SPARC) uses “logical carry”, for subtraction.  That is, it’s the non-inverted state of the carry out of the most significant bit.  It’s 1 if there’s no arithmetic borrow, or zero if there is arithmetic borrow.

Does the interpreter get this right?

cuavas:
To answer my own question, it appears it doesn’t work when carry in is involved.  You can test the logic for `compute_sub_ci` pretty easily:

| Rx in | Ry in | AC in | Rn out     | AC out            |
|-------|-------|-------|------------|-------------------|
| 1     | 0     | 1     | 1          | 1                 |
| 1     | 1     | 1     | 0          | 1                 |
| 1     | 2     | 1     | 0xffffffff | 0                 |
| 1     | 1     | 0     | 0xffffffff | 1 **(incorrect)** |

Do I need to work out how to fix this?

cuavas:
b401ac85b4d37aa82f9b42f484f0c2f9d59513aa fixes carry out for subtract with carry and adds saturation to several ALU ops where it was missing.  I haven’t checked the overflow/saturation logic for corner cases, but at least more of the operations that should saturate should work for simple cases.

`adsp21062_device::compute_dual_add_sub` also needs saturation implemented.

cuavas:
OK, I’m satisfied that AV is being calculated correctly for fixed point operations, and 27e8329725cb0e94a87f31333d93e0cb54fd41f8 implements saturation for fixed point dual add/subtract.

PR #15062, opened 2026-03-05
i8256: replace diy clock with device_serial_interface and deal with interrupts

stonedDiscord:
From #14478

Also reenables savestates.

### DIFF
--- a/src/devices/machine/i8256.cpp
+++ b/src/devices/machine/i8256.cpp
@@ -203,17 +203,13 @@ i8256_device::i8256_device(const machine_config &mconfig, const char *tag, devic
 	m_out_p2_cb(*this),
 	m_in_p1_cb(*this, 0),
 	m_out_p1_cb(*this),
-	m_rxc(0),
-	m_rxd(1),
-	m_cts(1),
-	m_txc(0),
 	m_timer(nullptr)
 {
 }
 
 void i8256_device::device_start()
 {
-	// FIXME: not everything that needs to be is saved here
+	// internal register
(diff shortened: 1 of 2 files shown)


cuavas:
`device_serial_interface` is useful for high-level simulation of devices that use asynchronous serial communication.  However, you usually end up digging a hole for yourself if you try to implement detailed UART behaviour on top of it (or God forbid USART behaviour).  As such, I'd consider making a UART device more dependent on `device_serial_interface` to be regressive.

stonedDiscord:
I copied the old implementation from
https://github.com/mamedev/mame/blob/master/src/devices/machine/i8251.cpp#L139
under the assumption that Intel didn't reinvent the wheel.
During testing in the last months it turned out it wasn't working and the one in this PR is working.
The datasheet isn't very thorough on the internals and I don't want to boil my working 8256 in acid so I'm unsure how to do this accurately.

cuavas:
MAME’s 8251 is a terrible starting point, as it’s in dreadful shape.  You also copy/pasted a pile of synchronous mode support code which wasn’t necessary  But if you look at the history of that file, you’ll see that the `device_serial_interface` dependency has been reduced over time, not increased.  A simple UART is pretty easy to implement directly anyway.  It’s mostly a couple of shift registers and counters.  The complicated part is the host interface, which `device_serial_interface` doesn’t help you with.

Anyway, no increased dependency on `device_serial_interface` in UART devices.

ajrhacker:
No, the 8256's asynchronous communications interface doesn't work very much like the 8251 USART.

PR #15064, opened 2026-03-06
machine/pit8253.cpp: Improved timing accuracy.

m1macrophage:
Based on @rb6502's suggestion [here](https://github.com/mamedev/mame/pull/14531#discussion_r2534689564).

Matched the mode descriptions and timing diagrams in the 8254 datasheet. The "Operation Common to All Modes" section summarizes most behaviors applicable to this PR.

This change **only** affects configurations where CLK is driven by **`write_clk*()`**. There should be **no change** in behavior for configurations that use **`set_clk()`**.

Specifically:
1. The counter is loaded and decremented on the falling edge of CLK.
1. GATE and GATE transitions (for edge-sensitive modes) are sampled on the rising edge of CLK.
   * Ensures GATE changes affect the counter on the next full CLK cycle (rising + falling edge).
1. Mode 3 treats GATE as edge-sensitive.
1. In modes 2 and 3, GATE has an immediate effect on OUT.
1. Deferred the execution of `set_clock_signal`.
   * Matches behavior of other control functions, and ensures calls to `set_clock_signal` are sequenced correctly.

The tuning algorithm on the `sixtrak` depends on (1) and (2) to function correctly, so it had local workarounds. Those have been removed.

-------

**Risk analysis and mitigation**

Given the 8253 is used in many places, I consider this a risky change. The blast radius is somewhat reduced due to `set_clk` configurations not being affected. I've also written **unit tests** replicating all waveform examples in the 8254 and 8253 datasheets, along with the tuning sequences of the `sixtrak`. Those can be found [here](https://github.com/m1macrophage/mamefork/blob/pittest/src/mame/virtual/utpit8253.cpp). If interested, I can send a PR for those. Manual tests were done on a handful of synths and the sente6vb.

A list of potentially affected devices and drivers can be obtained with:
```
grep -r 'write_clk[012]' src/ | awk -F: '{print $1}' | sort -u
```

Narrowing down to just driver files, ~10 out of ~25 have at least one working system. The others only have non-working systems.

There are some remaining, preexisting issues:
* If GATE is low when the counter is written, the counter won't get initialized until GATE goes high, even though it should be. This affects a couple of modes. See Figure 19, "Mode 4", second example, in the 8254 datasheet.
* Fix (4) above was intentionally left out of `set_clk` configurations.
* A counter write will initialize the counter earlier than it should.

Fixing these would affect the `set_clk` configuration and/or require more invasive changes, so I left them out of this PR. I expect they would only make a difference in contrived cases. I can revisit if/once this PR has been live for a bit.

### DIFF
--- a/src/devices/machine/pit8253.cpp
+++ b/src/devices/machine/pit8253.cpp
@@ -316,7 +316,8 @@ void pit_counter_device::flush_output()
 	int new_output = m_output;
 	if (m_clockin == 0)
 	{
-		// TODO: The same logic should apply when m_clockin != 0.
+		// TODO: The same logic should apply when m_clockin != 0, but that needs
+		// to be verified.
 		const int mode = CTRL_MODE(m_control);
 		if ((mode == 2 || mode == 3) && m_gate_input == 0)
 			new_output = 1;
@@ -1113,7 +1114,9 @@ void pit8253_device::write(offs_t offset, uint8_t data)
 bool pit_counter_device::edge_sensitive_gate() const
 {
 	const int mode = CTRL_MODE(m_control);
-	return mode == 1 || mode == 2 || mode == 3 || mode == 5;
+	// TODO: The m_clockin check should not be necessary, but removing it needs
+	// to be verified.
+	return mode == 1 || mode == 2 || (mode == 3 && m_clockin == 0) || mode == 5;
 }
 
 TIMER_CALLBACK_MEMBER(pit_counter_device::gate_w_deferred)


m1macrophage:
> This change **only** affects configurations where CLK is driven by `write_clk*()`. There should be **no change** in behavior for configurations that use `set_clk()`.

There was one minor change in behavior that made it through to `set_clk` configurations. I fixed that in the last commit.

angelosa:
We acknowledged a while back that PIT core isn't perfect, on top of my head there's a DOS game `ibm5170_cdrom:guimode` that fully runs on PIT clocks instead of using canonical vblank bit, and it's a speedup/slowdown rollercoaster that I'll surely test again when this is merged.

m1macrophage:
I checked how the PIT is configured for systems that use `ibm5170_cdrom`. I found `machine/at.cpp` and `machine/cs4031.cpp`. Both of those configure all counters with `set_clk`. So, unfortunately, this PR won't make a difference for these.

PR #15065, opened 2026-03-06
machine/vrender0.cpp: Add DMA address direction, Updates

cam900:
- Use struct for timer for reduce duplicates
- Add struct name for DMA registers for reduce duplicates
- Add notes for DMA control register
- Replace printf to logerror
- Use static function for DMA address increment calculation

### DIFF
--- a/src/devices/machine/vrender0.cpp
+++ b/src/devices/machine/vrender0.cpp
@@ -433,28 +433,16 @@ void vrender0soc_device::dmac_w(offs_t offset, u32 data, u32 mem_mask)
 	// fedc ba98 7654 3210 fedc ba98 7654 3210 
 	// xxxx xxxx xxxx xxxx xxxx x--- ---- ---- Reserved
 	// ---- ---- ---- ---- ---- -x-- ---- ---- DMA enable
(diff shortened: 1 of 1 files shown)


src/devices/machine/vrender0.h
```diff
@@ -127,16 +127,52 @@ class vrender0soc_device : public device_t, public device_mixer_interface
...
+		// - 0b0: Increase/Decrease destination address
+		// - 0b1: Fix destination address
+		// bit 2: DMA Destination address direction
+		// - 0b0: Increase destination address
+		// - 0b1: Decrease destination address
+		// bit 1-0 DMA Transfer width
+		// - 0b00: 8 bit
+		// - 0b01: 16 bit
+		// - 0b1x: 32 bit
 		u32 ctrl = 0;
```
angelosa, line 143: The PR is okay, but I never put documentation in headers (it causes a full recompile if modified). Also my preference is to use a shorter format:
```c++
// ---- -x-- DMA destination direction (0) increase (1) decrease
// ---- --xx DMA transfer width
// ---- --00 8-bit
// ---- --01 16-bit
// ---- --1x 32-bit
```

PR #15066, opened 2026-03-06
sinclair/*: Fixed various bugs

holub:
specnext.cpp: "dot commands" loading issue with NextZXOS - switched back to safer sdcard implementation which not such sensitive to timings.
spectrum.cpp: focus on debugger memory window crashing the emulator if exp devices are not activated.
sprinter.cpp: typo in fastram memory check; drop write if following access supported by the accelerator.

### DIFF
--- a/src/mame/sinclair/specnext.cpp
+++ b/src/mame/sinclair/specnext.cpp
@@ -58,7 +58,7 @@
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


ajrhacker:
Could you please provide a more descriptive commit message?

holub:
added description

PR #15067, opened 2026-03-07
rungun fix: irq 5 enable bit comes from a different register

jotego:


JimCarlTay:
Can you indicate the driver or game this commit affects?

For example: "rungun: fix - irq 5 enable bit comes from a different register"

happppp:
Does the same bit still ack IRQ? If so, on which edge? This here:
```
if (!(data & 0x400)) // actually a 0 -> 1 transition
	m_maincpu->set_input_line(M68K_IRQ_5, CLEAR_LINE);
```

jotego:
That's right, it also serves to acknowledge the interrupt. It is the active-low clear input of a flip flop, so the interrupt gets cleared when it goes low, but as long as it stays low, it won't take new interrupts either.

PR #15068, opened 2026-03-07
seta/downtown: handle calibr50 sub CPU reset write

jotego:


happppp:
Note to self: if you do a string search (CTRL+F) in github's embedded editor, it might fail to deselect the highlight after moving cursor.

PR #15069, opened 2026-03-07
epson/qx10.cpp: implement semidisk battery backed ramdisk card

brijohn:
This adds support the QX-10 semidisk card which was a battery backed 2MB RAM disk card for the QX-10 with native support in TPM/Valdocs.

### DIFF
--- a/src/devices/bus/epson_qx/semidisk.cpp
+++ b/src/devices/bus/epson_qx/semidisk.cpp
@@ -1,0 +1,148 @@
+// license:BSD-3-Clause
+// copyright-holders:Brian Johnson
+/*******************************************************************
+ *
+ * Semidisk Card
+ *
+ *******************************************************************/
+
+#include "emu.h"
+#include "semidisk.h"
+
+//**************************************************************************
+//  EPSON SEMIDISK DEVICE
+//**************************************************************************
+
+DEFINE_DEVICE_TYPE(EPSON_QX_OPTION_SEMIDISK, bus::epson_qx::semidisk_device, "epson_qx_option_semidisk", "Semidisk RAM Disk")
+
+namespace bus::epson_qx {
+
+static constexpr uint32_t BANK_SIZE = 512 * 1024;
+static constexpr uint8_t  NUM_BANKS = 4;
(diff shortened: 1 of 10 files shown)


src/devices/bus/epson_qx/semidisk.cpp
```diff
@@ -0,0 +1,148 @@
...
+static constexpr uint32_t BANK_SIZE = 512 * 1024;
+static constexpr uint8_t  NUM_BANKS = 4;
+static constexpr uint32_t RAM_SIZE = BANK_SIZE * NUM_BANKS;
+
+INPUT_PORTS_START( SEMIDISK )
+	PORT_START("CONFIG")
+	PORT_CONFNAME(0x01, 0x01, "Battery Backup")
+	PORT_CONFSETTING(0x01, "Enabled")
+	PORT_CONFSETTING(0x00, "Disabled")
+INPUT_PORTS_END
```
angelosa: Canonically we lowercase the `INPUT_PORTS_START` definition.

angelosa: Also, is this a physical option on card or it's "for convenience"?

brijohn: The actual card has a pin header on it for connecting an optional battery to, this option is basically simulating that so you can choose to run it with or without a battery attached.

brijohn:
This has been sitting here for a while, are there any other issues with it? If not can this be merged in?

src/devices/bus/epson_qx/semidisk.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+	return INPUT_PORTS_NAME( semidisk );
+}
+
+//-------------------------------------------------
+//  device_start - device-specific startup
+//-------------------------------------------------
+void semidisk_device::device_start()
+{
+	m_ram = std::make_unique<uint8_t[]>(RAM_SIZE);
```
galibert, line 55: Not really important, but you could use a memory_share_creator and have allocation and saving done automatically.

PR #15072, opened 2026-03-08
Floppy sound

mizapf:
New proposal for an enhancement of the floppy sound feature. Different to the November version, no macros are used, instead, a "floppy_sound_sample" class is introduced that encapsulates the list of sound samples. I tested on several systems that use the sound feature and those that do not, and did not run into issues so far.


ajrhacker:
Maybe there's some good reason why each instance of the `floppy_connector` device needs to own a set of pointers to sound samples for every single form factor, but it looks ridiculous to me.

mizapf:
A floppy connector needs to have pointers to each possible form factor, as we don't know which drive is actually connected before runtime. Also, the actual form factor is not known before the floppy_image_device calls setup_characteristics, so I let it fetch the proper sample from the connector right after that. Do you have a better suggestion where to store the sample set?

mizapf:
OK, hang on, I have an idea; I'll try and see whether it's better than the current way (basically, storing all samples of all form factors in a single instance of the samples object).

ICEknigh7:
> we don't know which drive is actually connected before runtime

Doesn't MAME require a reboot whenever a drive is connected/changed? If so, maybe we do know.

mizapf:
This is a new attempt for the update of the floppy sound system. Instead of using four different sample lists, only one list is used which may hold samples for all form factors (not limited by four variants). Also, custom samples may be defined per driver/device, and if not available, it falls back to the predefined 3.5 and 5.25 samples as used before. Instructions for setting up such lists are included in floppy.h. I tested various systems with and without sound support and found no issues so far.

This PR changes the existing system just in two aspects:
- it fixes some weird sound outputs for systems like the Amiga
- it provides a way to define custom lists

Apart from that, everything else should remain unchanged. 

If someone be so kind, please review it. Comments are welcome (preferably constructive ones).

mizapf:
It seems as if there are no more comments on this PR, and no one planning a review either.

Since I could not find any more issues after extensive testing, I suggest that I merge this PR into master by myself in a few days so it finally gets off the list. 

The issues of the November commit have been resolved, there is a new class that holds the samples list, no more macros,
and the rest of the update only concerns my own code.

Should some of you plan to use this custom sample feature, feel free to contact me if you have questions.

PR #9726, opened 2022-05-08
diexec: resynchronize to the implicit clock domain on clock frequency change

galibert:
I don't think it's disruptive, but I'm not sure.  At least it fixes floppy writes on apple3, which lost its phase with the diskiii due to the clock changes.

PR #15073, opened 2026-03-08
apple2video: improve video timing

arekkusu42:
(this is the last [part](https://github.com/mamedev/mame/pull/14749) of a larger WIP 50Hz PR that I can safely split out, see TODO section)

This PR fixes several aspects of apple2e/c and apple2gs video timing:
* apple2e, apple2gs: fix HBL relative timing
* apple2video: add 80STORE and NEWVIDEO beam-racing
* apple2gs: fix intra-scanline beam-racing
<br>
Code comments:<details>

In fixing the HBL timing, I have chosen to leave `set_raw()` with HBL to the right like it was previously, and adjust all uses of `hpos`/`vpos`/`vblank`.  The alternative is to properly place HBL to the left in `set_raw()`, but MAME's `screen()` unconditionally [allocates the entire bitmap width](https://github.com/mamedev/mame/blob/9641b67e9155c25eb068363d6fb8d944fdb4034a/src/emu/screen.cpp#L1095) including HBL.  For the apple2 machines this is wasted space, which then requires re-adding a (constant!) offset in each video update when manipulating each scanline (there's no method to adjust the base pointer of the bitmap allocation once on realloc?)  This gets worse when you also consider the apple2gs `m_8bit_graphics` which does not include HBL, so now we're continually re-adding and re-subtracting the same constant... ^_^;  I judged this implementation to be the lesser evil.  A wart is that the uncorrected `hpos`/`vpos` are still shown in MAME's debugger.

On apple2gs, screen updating has moved to the new `apple2_vgc` callback.  Happily, this occurs past the right border, so we still have 231 partial updates per frame like before.

On apple2gs (or apple2e/c with a Zip Chip), changing CPU clock speed introduces phase drift in the timer callbacks, which leads to off-by-one errors in the beam position.  To fix this, I am using the clock domain sync from #9726.  That works.  Sadly, it does not fix any of the [disk corruption problems](https://github.com/mamedev/mame/issues/14474#:~:text=To%20recap%2C%20we%20have).
</details>
<br>

There is no functional change to apple2/p, where the only video timing source available is the floating bus.<br>
On apple2e/c/gs, visual glitches in various software using C019 VBL sync is now fixed.  Before this PR, the VBL signal was incorrectly offset ~25 clocks, generally visible as flicker on the left edge of an animation.  For example, the opening wipe of [French Touch](http://fr3nch.t0uch.free.fr) Crazy Cycles II:<br>
FrenchTouch_CCII_glitch
...and similar glitches in ANSI Story, Scroll Scroll Scroll, [Deater's](http://www.deater.net/weave/vmwprod/mode_switch/modeswitch.html) Double Double (second half), Sather's Little Text Window, etc.<br>
On apple2gs, C02F HORIZCNT now exactly matches hardware; before this PR the inaccurate emulation caused a crash in the [FTA](http://www.freetoolsassociation.com/) XMAS demo at the menu screen.<br>
On apple2gs, partial updates now work mid-scanline.  This is commonly seen in text scrollers and other demo border effects:
FTA_XMAS_borderflag
The timing alignment of this type of effect now exactly matches hardware (reference video: [one](https://youtu.be/gtXYOX0k6C0?t=131), [two](https://youtu.be/1x6buUOKYAQ?t=79).)
</details>
<br>
Testing:<details>

In addition to comparing many beam-racing demos on IIgs and IIe Enhanced hardware, I also wrote more focused unit tests:
[Calibrate_260307.zip](https://github.com/user-attachments/files/25821995/Calibrate_260307.zip)

My [previous VIDSYNC test](https://github.com/mamedev/mame/pull/14177) is updated again here. It now covers all video softswitches, to demonstrate some esoteric cases:
VIDSYNC_HW_ee_80STORE
VIDSYNC_HW_gs_monoDHGR

Then there are a series of simple calibration gauges:
* VPR, HGR, VBL, IRQ
* POSGS, SCBGS, SHRGS, HGRGS, VBLGS, IRQGS

These each isolate one sync method and one softswitch (the border color on IIgs, else HGR PAGE2).  Unlike VIDSYNC they do not compensate for any machine- or softswitch-specific delay; they provide base data to understand the alignment of each timing signal.  For example we can see SCB IRQs fire exactly eight cycles after VBL on the IIgs:
Calibrate_VBL
Calibrate_SCB

Finally, BEAMPOS analytically validates the C02E/2F counters, paying particular attention to the off-by-one edge cases.  It checks the VERTCNT wrapping sequence in active and border locations, in 60 and 50Hz timings.  Then it checks the HORIZCNT wrapping sequence cycle-by-cycle.  Then it carefully measures the _relative_ timing alignment between the video counters, VBL, VBLIRQ, SCBIRQ, and floating bus readback.  It also looks at 8/16-bit differences reading the counters.   This test also runs on the //e or //c, but only compares the relative timing of VBL (or VBLIRQ on the //c) to the floating bus.

Before this PR, every part of this test fails in MAME 0.286 (except the easy C02E cases which were [previously fixed](https://github.com/mamedev/mame/pull/14749#:~:text=subtle%20off%2Dby%2Done%20bug)):
BEAMPOS_286_before
...after this PR, only the 16BIT (g65816 core is not cycle-exact), PAL and F-BUS (see TODO section) errors remain.

</details>
<br>
TODO:<details>

The remainder of this WIP 50Hz PR fixes additional timing-sensitive video interactions before finally enabling 50Hz mode:
* apple2video: emulate softswitch-specific delays
* apple2video: improve read_floatingbus()
* apple2gs: add 50Hz support

...however, each of those commits runs into... "behaviors" in MAME's core (one of which I have a fix for) so will need extra review, so a separate PR.

Other TODOs:
* On apple2e/gs, `apple2_interrupt()` callbacks could be optimized to only fire every, say, 64 scanlines, after investigating the mouse IRQ timing on //c hardware.
* On apple2gs, the 8-bit 560 pixel modes should really be stretch-blit to fill 640 pixels (but note the mode _can_ change mid-scanline!)
* On apple2gs, there are additional timing inaccuracies with FPI `slow_cycle()` (which is [why](https://github.com/mamedev/mame/commit/53a0ae7d11b82ab13564f2c82f43407337343d67) 3200 color mode is broken), and in the g65816 core.  Now that C02E/2F are accurately emulated, they could be used as building-blocks to write instruction timing tests.
</details>

### DIFF
--- a/src/mame/apple/apple2gs.cpp
+++ b/src/mame/apple/apple2gs.cpp
@@ -1091,6 +1091,9 @@ TIMER_DEVICE_CALLBACK_MEMBER(apple2gs_state::apple2_vgc)
 			scb = m_megaii_ram[0x15e80 + (shrline >> 1)];
 		}
 
+		// latch scb on this cycle for subsequent palette lookups
+		m_video->set_SHR_scb(shrline, scb);
+
 		if (scb & 0x40)
 		{
 			// VGC status flags are set even when the interrupt is disabled
--- a/src/mame/apple/apple2video.cpp
+++ b/src/mame/apple/apple2video.cpp
@@ -81,6 +81,7 @@ void a2_video_device::device_start()
 	save_item(NAME(m_monochrome));
 	save_item(NAME(m_rgbmode));
 	save_item(NAME(m_shr_palette));
+	save_item(NAME(m_shr_scbs));
 }
 
 void a2_video_device::device_reset()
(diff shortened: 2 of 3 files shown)


rb6502:
Great work, @arekkusu42 !  I wasn't sure border effects would be possible in MAME without core changes.

I spot checked other known border effects and the results weren't as good as XMAS Demo, although they're much better than they were.

The ECC demo (mame apple2gs eccdemo with the software list), which was the first released software with border effects (and why France hates dwsJason) has serious horizontal desync in the borders.  (ECC was the pirate release name for DreamWorld, and Joe Hack was the ECC name for dwsJason).

For Mr. Z's Colors Colors Colors (mame apple2gsr1 acsdemo2 and choose A at the menu) the border lines don't match up with the SHR for the little decorative white lines even though the color bar thing does.  Might be worth double checking that on hardware.  I don't remember it having that problem on a real IIgs but it's also been a few decades :-)

arekkusu42:
I should correct my comment:
The timing alignment of border effect _**at 1 MHz**_ now exactly matches hardware.

I'd looked at the ECC demo: it syncs on C02F, and briefly drops to 1 MHz going through a de-jitter jump table.  But then the border color is slammed with a bunch of unrolled direct page writes, and that runs at Fast speed.  So it is relying on FPI sync to align exactly the right number of Fast cycles into 65 1 MHz cycles, the same as TextFunk.  I've only looked into this enough to understand that accurate emulation is Hard, and moved along to lower-hanging fruit.  Ian Brumby has [looked at it with a logic analyzer](https://apple2infinitum.slack.com/archives/C0161FQGBKL/p1629788401197700) and [shared his approach in CrossRunner](https://apple2infinitum.slack.com/archives/C055PTN6VC7/p1746104675245089), and Baz has recently [tackled this in gssquared](https://github.com/jawaidbazyar2/gssquared/commit/893569b0dee87deede995b1d739c5376cb6296cf), so maybe it's doable in MAME too, but implementing the stretched 65th cycle looks hard.

The ACS SmartBoot stuff doesn't work on my ROM3 here.  But tracing the Colors Colors Colors demo in MAME's debugger, I can see that it is writing C034 while in Fast mode.  So similar FPI accuracy problems.

rb6502:
Yeah, that's why I said specifically "mame apple2gsr1 acsdemo2".  Older FTA/ACS stuff assumed the locations of the SmartPort low-level hooks in bank $E1 that aren't true on ROM 3.

Figures that ECC relies on FPI sync.  I'll bet Colors^3 relies on it by accident.

galibert:
FWIW, I suspect not having the left border as end-of-sync makes NTSC artefacting handling a little more difficult.

PR #15074, opened 2026-03-08
sh/sh7709s.cpp: Fix bug in DRC memory accessor using clobbered IO reg

jesus-ramos:
Fixes a bug with reads in the memory accessor where clobbered IO regs were being used as cache addresses

Fix a shift error in the wcr1 register values

Add bcr2 plumbing for determining cache line fetch reads based off BCR2. Area 0 is set via external pins so hardcoding the cv1k value, this is typically code space though so it only affects timing during copy of code to main memory

### DIFF
--- a/src/devices/cpu/sh/sh7709s.cpp
+++ b/src/devices/cpu/sh/sh7709s.cpp
@@ -165,8 +165,7 @@ static unsigned int get_wcr1_timing(uint32_t address, uint16_t wcr1)
 	if (area > 6 || area == 1)
 		return 0;
 
-	unsigned int area_shift = 12 - (area * 2);
-	unsigned int area_val = (wcr1 >> area_shift) & 0x3;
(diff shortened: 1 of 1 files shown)


jesus-ramos:
Just to update on testing so far, a lot of the early stages in the games match up closely.

So far we've found a couple of instances where later stage boss patterns don't seem to have quite enough slowdown. We've confirmed SDOJ S5 jetbachi one of the patterns lacks slowdown completely (credit to Jaimers for comparison footage) and the mirrors pattern is missing a bit of slowdown too possibly. Pink Sweets s6 boss has a good bit of slowdown but it's not quite enough (credit to Aquas for comparison and testing footage).

Not quite sure where the missing slowdown could be, my only guess right now is it looks like the uncached reads/writes to the blitter are handled separately and the cycle penalty isn't applied to those but I haven't looked too deeply into how many of those there are and if they would be significant. Not too familiar with the blitter handling so I've just been assuming it's close enough.

PR #15078, opened 2026-03-10
sh/sh7709s.cpp: Convert the cpu cycles to bus cycles properly

jesus-ramos:
Add a real writeback handling mechanism for writeback eviction penalties

Updates some timing values after going through the datasheets and timing diagrams a couple more times for values that overlap

Ignore non sdram regions for now, there's not many accesses to account for and area0 reads just cause the game load to take longer on the initial copy to sdram

### DIFF
--- a/src/devices/cpu/sh/sh7709s.cpp
+++ b/src/devices/cpu/sh/sh7709s.cpp
@@ -110,10 +110,7 @@ bool sh7709s_device::cache_access(uint32_t address, bool write)
 		if (entry->lru == 0)
 		{
 			if (entry->dirty)
-			{
 				m_wb_address = entry->tag * SH7709S_CACHE_LINE_SIZE;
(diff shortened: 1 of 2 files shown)


jesus-ramos:
Cleans up a good bit of instances of extra slowdown that aren't present on PCB such as s1 mushi futari midboss appearance giving 2 bursts of slowdown depending on how the rocks are blown up. This was previously sometimes just 1 long slowdown sequence due to the lack of conversion from cpu to bus cycles and writeback just adding its costs when some instruction overlap would save some cycles.

SDOJ s5 boss on expert still proving tricky to sort out what the missing sources of slowdown are, first pattern after jetbachi appearance lacks slowdown as does the stinger pattern with falling mirrors, final mirror sequence pattern before the last phase also has less consistent slowdown than on pcb.

PR #15082, opened 2026-03-11
cpu/powerpc: implement reservation for lwarx and stwcx instruction pair

gm-matthew:
PowerPC has the instructions `lwarx` and `stwcx` which are used to perform atomic update (load-modify-store) operations. `lwarx` loads a word and creates a reservation for a specific memory block; any write to this block will cause the reservation to be lost. `stwcx` checks if the reservation is still valid, and if it is, it writes the updated memory value, clears the reservation and sets the EQ bit of `cr0` to signify success. If it is not valid, it clears the EQ bit to signify failure and no write is performed.

The size of the reserved memory block is implementation-dependent, generally matching the cache line size. For many early PowerPC models, including most of those emulated in MAME, the reservation size is 32 bytes so that is what I have implemented.

This commit fixes a bug in several Model 3 games that causes them to freeze as a result of the previous incorrect implementation of the `lwarx` and `stwcx` instructions.

### DIFF
--- a/src/devices/cpu/powerpc/ppc.h
+++ b/src/devices/cpu/powerpc/ppc.h
@@ -340,6 +340,9 @@ protected:
 		const char * format;                     /* format string for printing */
 		uint32_t       arg0;                       /* print_debug argument 1 */
 		double       fp0;                        /* floating point 0 */
+		/* reservation monitor for lwarx and stwcx */
+		uint32_t reserve;
+		uint32_t reserve_address;
 	};
 
 	internal_ppc_state *m_core;
--- a/src/devices/cpu/powerpc/ppc.h
+++ b/src/devices/cpu/powerpc/ppc.h
@@ -603,9 +603,11 @@ protected:
 	uml::code_handle *   m_read32[8];                  /* read word */
 	uml::code_handle *   m_read32align[8];             /* read word aligned */
 	uml::code_handle *   m_read32mask[8];              /* read word */
+	uml::code_handle *   m_read32reserve[8];           /* read word reserved */
 	uml::code_handle *   m_write32[8];                 /* write word */
 	uml::code_handle *   m_write32align[8];            /* write word aligned */
 	uml::code_handle *   m_write32mask[8];             /* write word */
+	uml::code_handle *   m_write32reserve[8];          /* write word reserved */
 	uml::code_handle *   m_read64[8];                  /* read double */
 	uml::code_handle *   m_read64mask[8];              /* read double */
 	uml::code_handle *   m_write64[8];                 /* write double */
@@ -682,7 +684,7 @@ protected:
 	void static_generate_out_of_cycles();
 	void static_generate_tlb_mismatch();
 	void static_generate_exception(uint8_t exception, int recover, const char *name);
-	void static_generate_memory_accessor(int mode, int size, int iswrite, int ismasked, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
+	void static_generate_memory_accessor(int mode, int size, int iswrite, int ismasked, int isreserve, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
 	void static_generate_swap_tgpr();
 	void static_generate_lsw_entries(int mode);
 	void static_generate_stsw_entries(int mode);
--- a/src/devices/cpu/powerpc/ppc.h
+++ b/src/devices/cpu/powerpc/ppc.h
@@ -214,7 +214,7 @@ protected:
 	};
 
 	// construction/destruction
-	ppc_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, int address_bits, int data_bits, powerpc_flavor flavor, uint32_t cap, uint32_t tb_divisor, address_map_constructor internal_map);
+	ppc_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, int address_bits, int data_bits, powerpc_flavor flavor, uint32_t cap, uint32_t tb_divisor, address_map_constructor internal_map, uint32_t reservation_size);
 
 public:
 	virtual ~ppc_device() override;
@@ -582,6 +582,9 @@ protected:
 	/* internal stuff */
 	uint8_t               m_cache_dirty;                /* true if we need to flush the cache */
 
+	/* reservation granularity */
+	uint32_t              m_reservation_mask;
+
 	/* register mappings */
 	uml::parameter   m_regmap[32];                 /* parameter to register mappings for all 32 integer registers */
 	uml::parameter   m_fdregmap[32];               /* parameter to register mappings for all 32 floating point registers */
@@ -684,7 +687,7 @@ protected:
 	void static_generate_out_of_cycles();
 	void static_generate_tlb_mismatch();
 	void static_generate_exception(uint8_t exception, int recover, const char *name);
-	void static_generate_memory_accessor(int mode, int size, int iswrite, int ismasked, int isreserve, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
+	void static_generate_memory_accessor(int mode, int size, bool iswrite, bool ismasked, bool isreserve, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
 	void static_generate_swap_tgpr();
 	void static_generate_lsw_entries(int mode);
 	void static_generate_stsw_entries(int mode);
@@ -803,7 +806,7 @@ public:
 
 	void internal_ppc4xx(address_map &map) ATTR_COLD;
 protected:
-	ppc4xx_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, powerpc_flavor flavor, uint32_t cap, uint32_t tb_divisor);
+	ppc4xx_device(const machine_config &mconfig, device_type type, const char *tag, device_t *owner, uint32_t clock, powerpc_flavor flavor, uint32_t cap, uint32_t tb_divisor, uint32_t reservation_size);
 
 	virtual void execute_set_input(int inputnum, int state) override;
 };
--- a/src/devices/cpu/powerpc/ppcdrc.cpp
+++ b/src/devices/cpu/powerpc/ppcdrc.cpp
@@ -996,6 +996,14 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
 		}
 	}
 
+	/* if writing to a reserved memory block, cancel the reservation */
+	if (iswrite)
+	{
+		UML_AND(block, I3, I0, 0xffffffe0);                                 // and     i3,i0,0xffffffe0
+		UML_CMP(block, mem(&m_core->reserve_address), I3);                  // cmp     reserve_address,i3
+		UML_MOVc(block, COND_E, mem(&m_core->reserve), 0);                  // mov     reserve,0,e
+	}
+
 	/* general case: assume paging and perform a translation */
 	if (((m_cap & PPCCAP_OEA) && (mode & MODE_DATA_TRANSLATION)) || (iswrite && (m_cap & PPCCAP_4XX) && (mode & MODE_PROTECTION)))
 	{
@@ -3167,10 +3175,13 @@ bool ppc_device::generate_instruction_1f(drcuml_block &block, compiler_state *co
 			return true;
 
 		case 0x014: /* LWARX */
-			UML_ADD(block, I0, R32Z(G_RA(op)), R32(G_RB(op)));                          // add     i0,ra,rb
-			UML_MAPVAR(block, MAPVAR_DSISR, DSISR_IDX(op));                                 // mapvar  dsisr,DSISR_IDX(op)
-			UML_CALLH(block, *m_read32align[m_core->mode]);             // callh   read32align
-			UML_MOV(block, R32(G_RD(op)), I0);                                          // mov     rd,i0
+			UML_ADD(block, I0, R32Z(G_RA(op)), R32(G_RB(op)));                     // add     i0,ra,rb
+			UML_MOV(block, mem(&m_core->reserve), 1);                              // mov     reserve,1
+			UML_AND(block, I1, I0, 0xffffffe0);                                    // and     i1,i0,0xffffffe0
+			UML_MOV(block, mem(&m_core->reserve_address), I1);                     // mov     reserve_address,i1
+			UML_MAPVAR(block, MAPVAR_DSISR, DSISR_IDX(op));                        // mapvar  dsisr,DSISR_IDX(op)
+			UML_CALLH(block, *m_read32align[m_core->mode]);                        // callh   read32align
+			UML_MOV(block, R32(G_RD(op)), I0);                                     // mov     rd,i0
 			generate_update_cycles(block, compiler, desc->pc + 4, true);           // <update cycles>
 			return true;
 
@@ -3314,18 +3325,19 @@ bool ppc_device::generate_instruction_1f(drcuml_block &block, compiler_state *co
 			return true;
 
 		case 0x096: /* STWCX. */
-			UML_ADD(block, I0, R32Z(G_RA(op)), R32(G_RB(op)));                          // add     i0,ra,rb
-			UML_MOV(block, I1, R32(G_RS(op)));                                          // mov     i1,rs
-			UML_MAPVAR(block, MAPVAR_DSISR, DSISR_IDX(op));                                 // mapvar  dsisr,DSISR_IDX(op)
-			UML_CALLH(block, *m_write32align[m_core->mode]);                // callh   write32align
-			generate_update_cycles(block, compiler, desc->pc + 4, true);           // <update cycles>
-
-			UML_CMP(block, I0, I0);                                             // cmp     i0,i0
-			UML_GETFLGS(block, I0, FLAG_Z | FLAG_C | FLAG_S);                           // getflgs i0,zcs
-			UML_LOAD(block, I0, m_cmp_cr_table, I0, SIZE_BYTE, SCALE_x1);// load    i0,cmp_cr_table,i0,byte
-			UML_OR(block, CR32(0), I0, XERSO32);                               // or      [cr0],i0,[xerso]
-
-			generate_compute_flags(block, desc, true, 0, false);                       // <update flags>
+			UML_CMP(block, mem(&m_core->reserve), 0);                              // cmp     reserve,0
+			UML_JMPc(block, COND_Z, compiler->labelnum);                           // bz      1:
+			UML_ADD(block, I0, R32Z(G_RA(op)), R32(G_RB(op)));                     // add     i0,ra,rb
+			UML_MOV(block, I1, R32(G_RS(op)));                                     // mov     i1,rs
+			UML_MAPVAR(block, MAPVAR_DSISR, DSISR_IDX(op));                        // mapvar  dsisr,DSISR_IDX(op)
+			UML_CALLH(block, *m_write32align[m_core->mode]);                       // callh   write32align
+			UML_MOV(block, mem(&m_core->reserve), 0);                              // mov     reserve,0
+			UML_OR(block, CR32(0), 2, XERSO32);                                    // or      [cr0],2,[xerso]
+			UML_JMP(block, compiler->labelnum + 1);                                // b       2:
+			UML_LABEL(block, compiler->labelnum++);                                // 1:
+			UML_MOV(block, CR32(0), XERSO32);                                      // mov     [cr0],[xerso]
+			UML_LABEL(block, compiler->labelnum++);                                // 2:
+			generate_compute_flags(block, desc, true, 0, false);                   // <update flags>
 			return true;
 
 		case 0x2d5: /* STSWI */
--- a/src/devices/cpu/powerpc/ppcdrc.cpp
+++ b/src/devices/cpu/powerpc/ppcdrc.cpp
@@ -996,14 +996,6 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
 		}
 	}
 
-	/* if writing to a reserved memory block, cancel the reservation */
-	if (iswrite)
-	{
-		UML_AND(block, I3, I0, 0xffffffe0);                                 // and     i3,i0,0xffffffe0
-		UML_CMP(block, mem(&m_core->reserve_address), I3);                  // cmp     reserve_address,i3
-		UML_MOVc(block, COND_E, mem(&m_core->reserve), 0);                  // mov     reserve,0,e
-	}
-
 	/* general case: assume paging and perform a translation */
 	if (((m_cap & PPCCAP_OEA) && (mode & MODE_DATA_TRANSLATION)) || (iswrite && (m_cap & PPCCAP_4XX) && (mode & MODE_PROTECTION)))
 	{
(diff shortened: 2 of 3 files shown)


angelosa:
> This commit fixes a bug in several Model 3 games that causes them to freeze as a result of the previous incorrect implementation of the lwarx and stwcx instructions.

I guess: `fvipers2`, `vs298` and `swtrilgy` going in service mode?

cuavas:
Isn’t this flawed because you’re operating on logical addresses?  These operations are used when multiple devices have access to the same memory.  Other devices won’t be aware of the CPU’s address translation setup.

Since the main purpose of this pair of instructions is to implement atomic operations in a system where multiple devices can access the same memory, how do you plan to support this?  I imagine this would be important for fixing seemingly random issues in systems with multiple CPUs and/or bus mastering PCI devices.

Also, there should be a way for implementation classes to set the cache line size.

gm-matthew:
> I guess: `fvipers2`, `vs298` and `swtrilgy` going in service mode?

`scud` would hang when starting a race and `vf3` would hang during attract mode (if 3D rendering is disabled to prevent MAME crashing); both of these no longer occur. I haven't tested the above cases yet but it is plausible that these may have been caused by the same issue.

> Isn’t this flawed because you’re operating on logical addresses? These operations are used when multiple devices have access to the same memory. Other devices won’t be aware of the CPU’s address translation setup.
> 
> Since the main purpose of this pair of instructions is to implement atomic operations in a system where multiple devices can access the same memory, how do you plan to support this? I imagine this would be important for fixing seemingly random issues in systems with multiple CPUs and/or bus mastering PCI devices.

At the moment this commit works in the event of an interrupt occurring between the `lwarx` and `stwcx` instructions where the interrupt handler modifies the reserved memory block, even if it does not use the `lwarx`/`stwcx` pair to do so. Monitoring the block for all devices would be a much more complex solution and I am not sure how to implement it properly.

There are two alternative approaches I can think of:

1. Save the memory block to be reserved to an internal buffer when executing `lwarx` , and read it back when executing `stwcx` to ensure that it has not been modified. If it has, we invalidate the reservation. While this doesn't guarantee that no other devices will have written to the memory block, it does guarantee that the contents will not have changed.
2. Automatically invalidate the reservation when finishing a timeslice. This guarantees that no other device can touch the reserved memory block while the reservation is valid, though it could theoretically cause an infinite loop if the timeslice is too small and the gap between the `lwarx` and `stwcx` instructions is too large.

> Also, there should be a way for implementation classes to set the cache line size.

This is probably a good idea and should not be too difficult as long as we can find out the cache block size (not cache line, my mistake) for each PowerPC device currently emulated.

MooglyGuy:
Out of curiosity, isn't listening for modifications to a particular address-space range precisely what memory taps are useful for? https://docs.mamedev.org/techspecs/memory.html#taps

Install a tap upon `lwarx` to listen for any modifications to the relevant block. Remove the tap upon `stwcx` and/or when a modification occurs.

cuavas:
> Out of curiosity, isn't listening for modifications to a particular address-space range precisely what memory taps are useful for? https://docs.mamedev.org/techspecs/memory.html#taps
> 
> Install a tap upon `lwarx` to listen for any modifications to the relevant block. Remove the tap upon `stwcx` and/or when a modification occurs.

That won’t work, because in MAME the other CPU will be accessing the memory through its own address space, not your address space (e.g. the same RAM installed in both CPUs’ address spaces with `.share(...)`.  The “fastram” optimisation also means accesses to shared RAM likely aren’t going through address spaces at all.

mamehaze:
fvipers2 and vs298 can enter test mode, swtrilgy can't, but that appears to be the case even without the patch, this patch changes nothing in that regard.

neither exit test mode cleanly

cuavas:
Oh, and BTW there is only a `stwcx.` instruction, there’s no `stwcx` instruction, which would be a variant that didn’t affect condition codes.

angelosa:
> fvipers2 and vs298 can enter test mode, swtrilgy can't, but that appears to be the case even without the patch, this patch changes nothing in that regard.
> 
> neither exit test mode cleanly

Sorry I meant `swtrilgy` doesn't enter test mode, `fvipers2` and `vs298` crashes during attract mode.

mamehaze:
those are unchanged by this PR, fvipers2 crashes MAME, vs298 shows an onscreen error.

galibert:
> That won’t work, because in MAME the other CPU will be accessing the memory through its own address space, not your address space (e.g. the same RAM installed in both CPUs’ address spaces with `.share(...)`. The “fastram” optimisation also means accesses to shared RAM likely aren’t going through address spaces at all.

In the model3 and the powermac cases I don't think there's any shared ram with other cpus or devices, it's in practice more a protection against interrupts, incluing preemption.  So in that specific case a tap would do quite ok.

rb6502:
FWIW, this is a slightly more strict version of what both Supermodel and DingusPPC do right now, except Dingus does at least translate the address to physical.  (Neither of them mask the cache line size off the address, but that's apparently sufficient to boot OS X).  So I wouldn't be mad about making this do the logical to physical translation, adding some sufficiently scary comments, and checking this in more or less as-is as long as there's a nice TODO.

The rough sketch in my head of doing this correctly and MAME-idiomatically is to have a device that maintains a list of active reserved physical memory blocks using your favorite fast-lookup-on-random-key data structure.  The device is configured with the block size and masks incoming addresses accordingly.  The API should just be reserve a block, remove the reservation on a block, and check if a block is reserved.

CPUs that have multiprocessor bus primitives (68K, PPC, and i960 off the top of my head, but there are likely more) would have a configurable `device_finder` for this lookup device.  In this way, the driver can decide which CPU(s) share a physical address space in multiprocessor systems (BeBox and some later PowerMacs).

cuavas:
(Disclaimer: my PowerPCjutsu isn’t as sharp as it once was, due to atrophy from disuse, but I’m still pretty confident on most of this.)

The reservation granularity for `lwarx`/`stwcx.` is 32 bytes (eight words) for the 600, 750 and 7400 series, and 64 bytes (16 words) for the 970 and POWER5 series.  I don’t know what it is for other families off the top of my head.

The CPUs support a single reservation, which operates on physical addressing (i.e. the address after address translation).

A misaligned `lwarx`/`stwcx.` always causes an exception, even if the access fits within a bus access (not on a four-byte boundary, but not crossing an eight-byte boundary when the CPU is configured for a 64-bit data bus).

Using `lwarx`/`stwcx.` on an address within a caching-inhibited or direct-store page/block causes a DSI exception.  (Note that this does not include write-through pages/blocks – `lwarx`/`stwcx.` is permitted for write-through locations.)

`stwcx.` only checks that the effective address is valid (aligned and not within a cache-inhibited or direct-store page/block) and that a reservation exists.  It does not check that the effective address is within the reserved block.  It will succeed for locations outside the reserved block.

The reservation is cancelled when any of the following occur:
* A `stwcx.` instruction is executed with any effective address.
* An attempt by some other device to modify a location in the reservation granularity.
* The reserved block is evicted due to cache pressure, forcing the cache to be purged, or changing attributes to mark the page/block as cache-inhibiting (IIRC, I’d have to confirm this to be 100% sure).

Well-behaved interrupt handlers and process switching code will explicitly clear any reservation before returning (usually by executing a `stwcx.` instruction to some dummy location.

See, for example, the recommendations in the 603 user manual section 4.4:

> The operating system should execute one of the following when processes are switched:
> * The **sync** instruction, which orders the effects of instruction execution.  All instructions previously initiated appear to have completed before the **sync** instruction completes, and no subsequent instructions appear to be initiated until the **sync** instruction completes.  For an example showing use of the **sync** instruction, see Chapter 2, “Register Set,” of the *Programming Environments Manual*.
> * The **isync** instruction, which waits for all previous instructions to complete and then discards any fetched instructions, causing subsequent instructions to be fetched (or refetched) from memory and to execute in the context (privilege, translation, protection, etc.) established by the previous instructions.
> * The **stwcx.** instruction, to clear any outstanding reservations, which ensures that an **lwarx** instruction in the old process is not paired with an **stwcx.** instruction in the new process.
>
> The operating system should set the MSR[RI] bit as described in Section 4.2.4, “Setting MSR[RI].”

Or footnote 2 of [this blog post by Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20180814-00/?p=99485):

> Interrupts and traps do not clear the reservation.  This means that if the operating system wants to perform a context switch, it needs to perform a `stwcx.` to a harmless location to force the reservation to be cleared.  Otherwise, the thread being switched to might be in the middle of an atomic operation, and its `stwcx.` might succeed based on the previous thread’s reservation!  This is a rare case where you will intentionally perform a `stwcx.` to an address that doesn’t match the preceding `lwarx`.

(Note that that blog post incorrectly states that, “If you attempt to store back to a location different from the most recent preceding `lwarx`, and the reservation is still valid, the store might or might not succeed, and the *eq* bit will be unpredictable; it need not reflect the actual success of the store.  So don’t do that.”  It’s documented that it will succeed for the 600 series.)

Simply writing to another location within the reservation granularity from the same CPU will not cancel the reservation.  If it did, it would be impossible to write future-proof code, as future architecture implementations can increase the size of the reservation granularity (and have done so).

gm-matthew:
I decided to whip out my PowerBook 1400cs so I could test the `lwarx` and `stwcx.` instructions on an actual PowerPC, and I discovered that writing to a memory address previously reserved by `lwarx` does **not** cancel the reservation (unless `stwcx.` is used); this means that my current commit is incorrect.

Are there any drivers in MAME using a PowerPC that have other devices that share RAM with the PowerPC?

cuavas:
konami/konamim2.cpp is a dual PPC602 SMP setup.

apple/macpdm.cpp supports bus mastering NuBus cards with a PPC601 in principle.

konami/cobra.cpp has a grab bag of PowerPC CPUs (PPC603e main CPU, PPC403 system controller, PPC604 T&L processor) with plenty of DMA capabilities, but emulation is very incomplete.

be/bebox.cpp is a dual PPC603 SMP setup, but I don’t know how complete emulation is.

angelosa:
Bebox is very incomplete, out of using legacy PCI & SCSI. For starter it doesn't hookup the ISA and the Super I/O, which means it can't even output to the debug port. Then afaik none of boot CDs actually boots, not sure if it was supposed to do more but tbh it's in a kind of limbo that probably rewriting from scratch is probably for the best (note: reuses MPC105 as Sega Model 3)
https://www.netbsd.org/ports/bebox/hardware.html

gm-matthew:
I've made it so that writing to a reserved block from the same CPU that reserved it no longer cancels the reservation, as per my test on an actual PowerPC. The `m_core->reserve_address` value is currently unused so we should either find a way to use it effectively or delete it.

Out of curiosity I looked at the code for Xenia which has to emulate multiple PowerPC cores, and its approach for `lwarx` and `stwcx.` is to use a global lock which I presume means no other threads can run until the lock is removed. In the event that a global lock is not available, it also uses an atomic compare-and-swap so that `stwcx.` only updates the reserved value if it remains the same value as when it is loaded with `lwarx`, i.e. it has not been modified.

I think attempting to implement it so that any other device writing to a reserved memory block cancels the reservation would be a lot of effort for not much benefit; it would be much easier to either force the PowerPC to keep executing until the reservation is cleared (thus ensuring that nothing else can touch the reserved block), or make it so that `stwcx.` reloads the reserved value to confirm that it has not been modified before writing the updated value. But that's just my opinion and I shall leave it to the other devs to decide on the best solution.

cuavas:
Look, I don’t think this implementation can work:

* You’re still setting the reservation address *before* address translation, which is incorrect.
* You’re setting the “reservation exists” flag *before* address translation, which is also incorrect, as an exception during address translation will result in the reservation not being created.
* You can’t just call the `m_read32align` handler, because you need to hook into address translation to check whether the address is in a direct-store or cache-inhibited area and raise a DSI exception.
* Does `m_read32align` check alignment?  I thought it just masked off the low bits.  A misaligned `lwarx` always raises an alignment exception.

It’s similar for `stwcx.` – you need to hook into address translation to generate DSI exceptions properly, and it always needs an alignment check.

galibert:
Something is unclear to me.  The cpu does not snoop every write to the address, only another stxcw can invalidate the reservation?

gm-matthew:
> Something is unclear to me.  The cpu does not snoop every write to the address, only another stxcw can invalidate the reservation?

If a particular CPU uses `lwarx` to create a reservation, using the same CPU to write to the same address using an instruction other than `stwcx.`, for example `stwx`, does not invalidate the reservation.

galibert:
And if another cpu or a dma does it?

gm-matthew:
> And if another cpu or a dma does it?

Then the cache block containing the reserved value is invalidated, resulting in the reservation also being invalidated.

gm-matthew:
> * Does `m_read32align` check alignment?  I thought it just masked off the low bits.  A misaligned `lwarx` always raises an alignment exception.

Yes, `m_read32align` (and likewise `m_write32align`) raises an alignment exception if the address is not aligned.

I'm currently working on moving the logic over to the `static_generate_memory_accessor` function, adding an additional parameter `isreserve` that will be set to true for `lwarx` and `stwcx.` and false for all other memory access instructions. This will make it easier to access the translated physical address to use for the reservation.

gm-matthew:
I have moved the logic for `lwarx` and `stwcx.` over to the `static_generate_memory_accessor` so they can work with the translated physical address rather than the logical address. If an alignment or DSI exception occurs during the `lwarx` instruction, no reservation is created.

We still need to decide which method we should use to detect or prevent other devices from modifying the reserved value/block while the reservation is active:

1. Actually emulate the load-link/store-conditional behavior properly, tracking all writes from other devices to the reserved memory block and invalidating the reservation if one occurs. This is beyond my ability to implement so someone else will need to do it.
2. Invalidate the reservation if the PowerPC finishes its timeslice. This guarantees that no other device can touch the reserved block, but in the unlikely event that the number of instructions between `lwarx` and `stwcx.` is greater than the number of cycles allocated in each timeslice, `stwcx.` may never succeed and an infinite loop may occur.
3. If a reservation has been created by `lwarx`, force the PowerPC to keep executing until it has been cleared by `stwcx.`. This also guarantees that no other device can touch the reserved block, but it may affect emulation timing.
4. Make `stwcx.` check the reserved value to verify that it is the same value before performing the write; essentially a compare-and-swap operation. This does not guarantee that the reserved value has not been written to (see the [ABA problem](https://en.wikipedia.org/wiki/ABA_problem)).

Options 2 and 3 do not require us to save the address of the reservation, option 4 requires the address and value of the reservation, and option 1 requires the address of the cache block.

rb6502:
I'd really like to see this move forward because it's an issue for the PowerMacs, not to mention the crashing Model 3 games.

My pitch: add a virtual method to `cpu_device`, `external_write(offs_t offset)` that by default does nothing.  Then you can use whatever mechanism you want to notify it.

For DMA, most of the emulated DMA controllers have a required callback for the driver to do the actual reading/writing, so it'd be easy to add there.  Although on most hardware I'm aware of the CPU is blind to DMA writing things and the OS has to manage that anyway.  For CPUs there could be an optional callback to signal writes without the overhead of a tap (and that would allow the RAM bypass to still work).  But I think solving that part of the problem is not something that needs to happen up front since neither Model 3 nor macpdm are multiprocessor systems.

galibert:
I agree with your pitch with just a small change: make it offset + length, so that dmas only do one call.

rb6502:
These changes fix:
- System 7.5.3 and later (which use a native PowerPC SCSI driver instead of emulated 68k) boot to Finder on pmac6100 instead of randomly live-locking on SCSI interrupts.
- The boot animation with the CD being inserted now plays on Pippin.
- The flashing ? floppy disk now shows on pmac7200 (now I have to rewrite dbdma.cpp to use the classic DRQ model, but it'll be well worth it).

@gm-matthew already changed it to use physical addresses, and I'll add the proposed external_write() method I mentioned above (with OG's change) so that external things can signal when they might be stomping the reservation.  Any other objections?

cuavas:
Lemme check this again tonight.  I’d missed the part where it was updated.

src/devices/cpu/powerpc/ppc.h
```diff
@@ -679,7 +684,7 @@ class ppc_device : public cpu_device, public device_vtlb_interface
 	void static_generate_out_of_cycles();
 	void static_generate_tlb_mismatch();
 	void static_generate_exception(uint8_t exception, int recover, const char *name);
-	void static_generate_memory_accessor(int mode, int size, int iswrite, int ismasked, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
+	void static_generate_memory_accessor(int mode, int size, int iswrite, int ismasked, int isreserve, const char *name, uml::code_handle *&handleptr, uml::code_handle *masked);
```
cuavas: Since you’re changing the signature anyway, can you change the Boolean parameters (`iswrite`, `ismasked` and `isreserve`) to `bool`)?

src/devices/cpu/powerpc/ppcdrc.cpp
```diff
@@ -1010,6 +1010,27 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
 		UML_AND(block, I0, I0, 0x7fffffff);                                 // and     i0,i0,0x7fffffff
 	UML_XOR(block, I0, I0, (mode & MODE_LITTLE_ENDIAN) ? (8 - size) : 0);   // xor     i0,i0,8-size
 
+	if (isreserve)
+	{
+		if (iswrite)
+		{
+			int dowrite = label++;
```
cuavas: In new code, can you please use `uml::code_label` for label values, and also make them `const` when possible to avoid the chance of accidentally reassigning them?

src/devices/cpu/powerpc/ppcdrc.cpp
```diff
@@ -1010,6 +1010,27 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
...
+			UML_CMP(block, mem(&m_core->reserve), 0);                              // cmp     reserve,0
+			UML_JMPc(block, COND_NZ, dowrite);                                     // bnz      1:
```
cuavas: If you’re only checking for zero/non-zero, you should use `UML_TEST(block, mem(&m_core->reserve), 0xffffffff);` rather than `UML_CMP`.

src/devices/cpu/powerpc/ppcdrc.cpp
```diff
@@ -1010,6 +1010,27 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
...
+			UML_RET(block);                                                        // ret
+			UML_LABEL(block, dowrite);                                             // 1:
+			UML_MOV(block, mem(&m_core->reserve), 0);                              // mov     reserve,0
+			UML_OR(block, CR32(0), 2, XERSO32);                                    // or      [cr0],2,[xerso]
+		}
+		else
+		{
+			UML_MOV(block, mem(&m_core->reserve), 1);                              // mov     reserve,1
+			UML_AND(block, I3, I0, 0xffffffe0);                                    // and     i3,i0,0xffffffe0
+			UML_MOV(block, mem(&m_core->reserve_address), I3);                     // mov     reserve_address,i3
```
cuavas, line 1030: As I said before, please make the reservation size a device class member so it can be set per microarchitecture on construction.

src/devices/cpu/powerpc/ppcdrc.cpp
```diff
@@ -1304,9 +1325,11 @@ void ppc_device::static_generate_memory_accessor(int mode, int size, int iswrite
...
 	if (alignex != 0)
 	{
-		UML_LABEL(block, alignex);                                                      // alignex:
-		UML_RECOVER(block, SPR32(SPROEA_DSISR), MAPVAR_DSISR);                              // recover [dsisr],DSISR
-		UML_EXH(block, *m_exception[EXCEPTION_ALIGN], I0);                 // exh     align,i0
+		UML_LABEL(block, alignex);                                             // alignex:
+		if (isreserve && iswrite)
+			UML_MOV(block, mem(&m_core->reserve), 0);                          // mov     reserve,0
+		UML_RECOVER(block, SPR32(SPROEA_DSISR), MAPVAR_DSISR);                 // recover [dsisr],DSISR
+		UML_EXH(block, *m_exception[EXCEPTION_ALIGN], I0);                     // exh     align,i0
```
cuavas, line 1332: Are you sure the reservation is cleared when an exception happen before the write completes?  Did you test this on hardware?

cuavas:
> Something is unclear to me. The cpu does not snoop every write to the address, only another stxcw can invalidate the reservation?

Only a `stwcx.` *on the same CPU* invalidates the reservation.  Any write to the reserved range *by any other device in the system* that participates in the cache coherency protocol invalidates the reservation.

src/devices/cpu/powerpc/ppcdrc.cpp
rb6502: I still get the benefits without clearing on exception so I'll remove that.

gm-matthew: I have tested on my PowerBook 1400cs and observed that if I use `lwarx` to load a value, increment it, use a `stwcx.` instruction on a misaligned address (resulting in an exception) and then try to use `stwcx.` again to store the incremented value, the store does not succeed. The previous `stwcx.` must be invalidating the reservation even when an alignment exception occurs.

rb6502: But does `stwcx.` always clear the reservation in spite of the exception, or does the exception itself flush the reservation, which is what the submitted code does?  I think it's extremely unlikely on any production PPC code that `lwarx`/`stwcx.` don't happen in quick succession, but it's theoretically possible to have some other loads/stores between them that work outside of the reservation range and  have a recoverable fault.

rb6502: Sorry, I didn't quite absorb the implications of your test.  Was the misaligned address `stwcx.` in the reservation range?

gm-matthew: The second `stwcx.` instruction fails to store the incremented value even when the misaligned address in the first `stwcx.` is outside the reservation range.

It is possible that the alignment exception handler in Mac OS explicitly clears the reservation with another `stwcx.` instruction but I have no practical way to test for this. With that in mind, I cannot find 100% concrete evidence that `stwcx.` clears a reservation if an alignment exception occurs (Google AI search says that it does but I know better than to trust that AI results are accurate), but at the same time there is no evidence that it does not. All sources I can find state that executing `stwcx.` clears the reservation without any reference to what happens if an alignment exception occurs.

cuavas: My gut feel is that it doesn’t clear the reservation, but the exception handler, being well-behaved, explicitly clears it before returning as recommended in all the PPC documentation.  The only way to test it would be if you can install your own exception handler.

cuavas: > Sorry, I didn't quite absorb the implications of your test. Was the misaligned address `stwcx.` in the reservation range?

That doesn’t matter for the purpose of the test – `stwcx.` doesn’t care whether it’s in the reservation range or not.

cuavas:
Is it OK if I un-draft this, resolve conflicts, etc. and merge?

rb6502:
> Is it OK if I un-draft this, resolve conflicts, etc. and merge?

I have it merged manually in my local tree along with some pretty significant other fixes so I would prefer to handle it.

rb6502:
This change is part of https://github.com/mamedev/mame/commit/c78876c8bf827f11f8890274bddf37643b913d5c

PR #15093, opened 2026-03-13
video/x1_020_dx_101.cpp: Make Allumer X1-020/NEC DX-101 as device, Updates:

cam900:
video/x1_020_dx_101.cpp (former seta/seta2_v.cpp):
- Minor optimization
- Add notes for tile format in tilemap
- Reduce duplicates
- Remove unnecessary member variable

seta/seta2.cpp:
- Split funcube hardware into namco/funcube.cpp and namco/funcube_touchscreen.cpp (for touchscreen), demote funcube series to MACHINE_IMPERFECT_GRAPHICS due to clipped title logo
- Consolidated drivers in single file
- Fix unused error in staraudi
- Reduce duplicates and literal tag usages
- Move DX-102 into other (i don't think DX-102 is part of video generator, more like to I/O related?)

namco/namcoeva.cpp:
- Move driver into namco/ folder, Remove relations with seta/seta2.cpp

### DIFF
--- a/src/devices/video/x1_020_dx_101.cpp
+++ b/src/devices/video/x1_020_dx_101.cpp
@@ -893,9 +893,9 @@ TIMER_CALLBACK_MEMBER(x1_020_dx_101_device::raster_timer_done)
 	{
 		if (BIT(m_rasterenabled, 0))
 		{
+			LOGVIDEO("external int (vpos is %d)\n", screen().vpos());
 			screen().update_partial(screen().vpos() - 1);
 			m_raster_irq_cb(ASSERT_LINE);
-			LOGVIDEO("external int (vpos is %d)\n", screen().vpos());
 		}
 	}
 }
--- a/src/mame/namco/funcube.cpp
+++ b/src/mame/namco/funcube.cpp
@@ -80,12 +80,12 @@ public:
 		m_leds(*this, "led%u", 0U)
 	{ }
 
-	void funcube(machine_config &config);
-	void funcube2(machine_config &config);
+	void funcube(machine_config &config) ATTR_COLD;
+	void funcube2(machine_config &config) ATTR_COLD;
 
-	void init_funcube();
-	void init_funcube2();
-	void init_funcube3();
+	void init_funcube() ATTR_COLD;
+	void init_funcube2() ATTR_COLD;
+	void init_funcube3() ATTR_COLD;
 
 private:
 	virtual void machine_start() override ATTR_COLD;
--- a/src/devices/video/x1_020_dx_101.h
+++ b/src/devices/video/x1_020_dx_101.h
@@ -49,6 +49,10 @@ private:
 	devcb_write_line m_flip_screen_x_cb;
 	devcb_write_line m_flip_screen_y_cb;
 
+	// memory pointers
+	memory_share_creator<uint16_t> m_spriteram;
+	memory_share_creator<uint16_t> m_vregs;
+
 	// live state
 	std::unique_ptr<uint16_t[]> m_private_spriteram;
 	std::unique_ptr<uint32_t[]> m_realtilenumber;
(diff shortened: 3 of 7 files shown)


src/mame/namco/funcube.cpp
```diff
@@ -0,0 +1,677 @@
...
+Ordered by Board        Year    Game                                    By
+-------------------------------------------------------------------------------------------
+P0-140B                 2000    Funcube                                 Namco
+P0-140B                 2000    Namco Stars                             Namco
+B0-006B                 2001-2  Funcube 2 - 5                           Namco
+-------------------------------------------------------------------------------------------
+
+TODO:
+
+- Proper emulation of the ColdFire CPU, in a core file.
```
angelosa: 🤨 I did this a while back, rather it needs the touchscreen regression fixed.
https://github.com/mamedev/mame/commit/7671d6f99cd8a123f82414f90cd1504fb721e504
https://github.com/mamedev/mame/commit/24249ebb7bf431f68f00e43e4d71b55e2fcc7bf7

src/devices/video/x1_020_dx_101.cpp
```diff
@@ -802,22 +887,21 @@ void seta2_state::draw_sprites_line(bitmap_ind16 &bitmap, const rectangle &clipr
...
+	if (!m_raster_irq_cb.isunset())
 	{
 		if (BIT(m_rasterenabled, 0))
 		{
-			tmp68301->set_input_line(1, HOLD_LINE);
-			LOGVIDEO("external int (vpos is %d)\n", m_screen->vpos());
-			m_screen->update_partial(m_screen->vpos() - 1);
+			m_raster_irq_cb(ASSERT_LINE);
+			LOGVIDEO("external int (vpos is %d)\n", screen().vpos());
+			screen().update_partial(screen().vpos() - 1);
```
cuavas: You really want to trigger the the partial update before firing the callback.  It works in the opposite order when it's hooked up to `set_input_line` because that takes a trip through the scheduler anyway, but if it has immediate effects, you want it to happen afterwards.

src/devices/video/x1_020_dx_101.cpp
```diff
@@ -200,20 +286,20 @@ void seta2_state::vregs_w(offs_t offset, uint16_t data, uint16_t mem_mask)
...
-		flip_screen_x_set(BIT(data, 0));
-		if (data & ~1)  LOGVREG("CPU #0 PC %06X: flipx unknown bits %04X\n", m_maincpu->pc(), data);
+		m_flip_screen_x_cb(BIT(data, 0));
+		if (data & ~1)  LOGVREG("%S: flipx unknown bits %04X\n", machine().describe_context(), data);
 		break;
 	case 0x2c:  // FLIP Y (pzlbowl)
-		flip_screen_y_set(BIT(data, 0));
-		if (data & ~1)  LOGVREG("CPU #0 PC %06X: flipy unknown bits %04X\n", m_maincpu->pc(), data);
+		m_flip_screen_y_cb(BIT(data, 0));
+		if (data & ~1)  LOGVREG("%S: flipy unknown bits %04X\n", machine().describe_context(), data);
```
cuavas: Should be `%s` (lowercase).  `%S` is supposed to force interpreting a string as `wchar_t`, but isn’t supported by strformat.h

cuavas (commented):
The build is failing because you haven’t updated mame.lst to match.

cuavas:
Not a regression (already broken in 0.286), but `mj4simai` seems to be drawing graphics off the left side of the screen if you turn the Flip Screen DIP switch on.  It works with the Flip Screen DIP switch off, though.  Do other games have issues with screen flipping?

cam900:
Almost all games in seta/seta2.cpp already marked as MACHINE_NO_COCKTAIL, especially:
gundamex (and most other games) has black screen when flip screen DIP switch is on.
myangel/2 just displays single column at leftmost side.
pzlbowl flip screen setting doesn't affect in individual sprites. (causes garbage)
deerhunt flip screen setting is not affected at all in screen.

PR #15095, opened 2026-03-13
seta/seta2.cpp: Demote gundamex as MACHINE_IMPERFECT_TIMING

cam900:
- It displays black screen when flip screen DIP switch is on, and has overall timing issue.
- also, this PR fixes spacing in driver definition.

### DIFF
--- a/src/mame/seta/seta2.cpp
+++ b/src/mame/seta/seta2.cpp
@@ -3915,7 +3915,7 @@ ROM_END
 
 } // anonymous namespace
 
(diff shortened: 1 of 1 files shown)


cuavas:
Does this game actually support cocktail cabinet mode, i.e. automatically flipping the screen for the second player’s turn?  `MACHINE_NO_COCKTAIL` is specifically for bugs supporting cocktail cabinet mode, not issues with screen flipping in general.  If the setting is just for using a cabinet with the monitor in the opposite orientation and not for a cocktail cabinet, `MACHINE_NO_COCKTAIL` should not be used.

cam900:
gundamex is head-to-head fighting game and doesn't have cocktail setting

PR #15096, opened 2026-03-13
gridcomp/gridcomp.cpp: implement RS232

vklachkov:
TL;DR: This PR fixes the UART i8274 chip connections to the GRiD Compass laptop.

## GRiD Compass and Serial

The GRiD Compass uses a proprietary 19-pin connector that provides RS232, RS422, and +10V/-10V power. The circuit uses PAL chips that configure the signal source for TxC, RxC, and also SC for GPIB. There's no PAL dump available, only some circuit diagrams and approximate connection descriptions from @ConventionalMemories and @JDat...

Instead of emulating the PAL chip logic, I connected the i8274's TxC/RxC directly to 80130 baud. I also fixed the swapped `cd` `ba` pins, set the correct i8274 address for the GRiD Compass II and configured interrupt.

Additionally, I updated the device list. It now includes a simple serial printer (without software flow control) that works with the `MinimumSerial~Printer~` driver, plus two mice: Microsoft Mouse and Logitech Mouse. Both mice only work in GRiDPaint. On macOS, both mice behave erratically, but this appears to be a broader MAME issue on macOS rather than an emulation problem. The Mouse Systems mouse is commented out because it doesn't work for some reason.

Video showing the mouse working on this branch:

https://github.com/user-attachments/assets/93370a89-b33d-450b-9edc-ae8ba9cc74c6

P.S. My goal now isn't to achieve perfect Compass emulation — I just want to make GRiD OS 3.1 programs work correctly without obvious artifacts in MAME.

### DIFF
--- a/src/mame/gridcomp/gridcomp.cpp
+++ b/src/mame/gridcomp/gridcomp.cpp
@@ -74,7 +74,6 @@
 #include "machine/i80130.h"
 #include "machine/i8087.h"
 #include "machine/i8255.h"
-#include "machine/clock.h"
 #include "machine/mm58174.h"
 #include "machine/ram.h"
 #include "machine/tms9914.h"
@@ -83,7 +82,6 @@
 
 #include "emupal.h"
 #include "screen.h"
-#include "softlist.h"
 #include "speaker.h"
 
 #define LOG_KEYBOARD  (1U << 1)


vklachkov:
@galibert @happppp could you please do a review when you have time? The PR is ready.

happppp:
hlemouse is used by more than just gridcomp, changing the stop bits and adding dcd/cts there may (very likely) break other drivers that use it.
The timer2_w trampoline is not needed, you can do a devcb to rxca_w with set() and a 2nd one to txca_w with append().

angelosa:
> hlemouse is used by more than just gridcomp, changing the stop bits and adding dcd/cts there may (very likely) break other drivers that use it.

The description at header claims 7N1 tho, what gives?

Lord-Nightmare:
Could HLEmouse be made configurable so the number of stop bits can be set when the device is started?

vklachkov:
GRiD uses a standard Microsoft mouse. And the Microsoft mouse is 7N1. It's written about everywhere, even in the file header. The fact that the code said 7N2 was an obvious mistake.

dcd/cts lines are asserted by a real mouse and some computers rely on it, not just the Compass.

My changes are not a hack...

vklachkov:
If there is some computer that, for some reason, expects 7N2, please tell me. I'm ready to test it and make a generic class or somehow add stop bit configuration. But it seems to me it's just a typo, not malicious intent.

cracyc:
> If there is some computer that, for some reason, expects 7N2, please tell me. I'm ready to test it and make a generic class or somehow add stop bit configuration. But it seems to me it's just a typo, not malicious intent.

Yes, xfree86, https://forums.bannister.org/ubbthreads.php?ubb=showflat&Number=82441#Post82441 .

galibert:
Note that there's no way to distinguish between two stop bits and one stop bit with a delay that's at least one bit duration between bytes.

vklachkov:
Hmm. Strange. I didn't expect that I'd also have to figure out why some system is expecting for a non-standard number of stop bits...

galibert:
That's the joy of Mame, we have to stay bug-compatible with things that will never be fixed :-)

vklachkov:
I'll test it and get back. Of course, I don't like that the blocker for this PR is the Microsoft Mouse. I never had any doubts about this fix. It seems like I should split the PR into Serial and the mouse separately, since I'll still have to figure out what's going on with that coordinate madness.

galibert:
You realize that if you put STOP_BITS_2 back the compass will not notice, right?

vklachkov:
The Compass definitely wants 7N1 and doesn't work with 2 stop bits. And it's obviously an error in MAME. But I don't understand why error is exists. I don't know how much time it will take to test mice in different systems.

vklachkov:
Although maybe it really is worth making two mice... One Microsoft, the other pseudo-Microsoft. I don't know.

galibert:
The compass not working with 2 stop bits is extremely weird.  Remember that at 9600bauds it just means there's an additional 104us delay between the end of a byte and the beginning of the next without any change on the data line (e.g. it stays at 1).  So the compass not working when that happens is really, really strange and should not happen.

vklachkov:
Sorry, there was an error on my part. The Compass **works** with the mouse in 7N2 mode. I'm not sure why it didn't work last time. So I can revert this change... Next time I'll test more thoroughly.

---

Speaking not about the Compass, but just about the Microsoft Mouse, it seems wrong to me to leave it as 7N2. It works, but it creates a dissonance with the project's goal "is to document the hardware and reproduce the behavior of the hardware as faithfully as possible".

I didn't have time to test TinyCore 2.7 with xfree86 and figure out if there is a bug in the latest available version of xfree86. And I didn't have time to look for which systems might have been broken by my change.

We can discuss this separately, because it's not related to the Compass.

galibert:
It's a HLE.  The real thing may be too slow to send bytes back-to-back, and then the number of stop bits things would be moot :-)  IOW, maybe the hardware does have more than one bit of delay between bytes, we won't know without a LLE.

vklachkov:
@galibert ok, that sounds logical. I've decided to revert all my changes to the Microsoft Mouse, so that in this PR I leave only the changes related to the Compass and don't feel the need to test other machines that use this mouse)

Thank you very much for guiding my PRs.

vklachkov:
I've removed the trampoline, and if there are no further comments on this PR, it would be great to merge it.

PR #15102, opened 2026-03-14
Fix scaling on macOS

vklachkov:
Before:

After:

### DIFF
--- a/src/osd/modules/osdwindow.h
+++ b/src/osd/modules/osdwindow.h
@@ -76,6 +76,7 @@ public:
 	bool keepaspect() const;
 
 	virtual osd_dim get_size() = 0;
+	virtual osd_dim get_size_pixels() { return get_size(); }
 
 	osd_monitor_info *monitor() const { return m_monitor.get(); }
 	std::shared_ptr<osd_monitor_info> monitor_from_rect(const osd_rect *proposed) const;
--- a/src/osd/modules/render/drawbgfx.cpp
+++ b/src/osd/modules/render/drawbgfx.cpp
@@ -328,7 +328,7 @@ void video_bgfx::save_config(config_type cfg_type, util::xml::data_node *parentn
 
 bool video_bgfx::init_bgfx_library(osd_window &window)
 {
-	osd_dim const wdim = window.get_size();
+	osd_dim const wdim = window.get_size_pixels();
 
 	bgfx::Init init;
 	init.type = bgfx::RendererType::Count;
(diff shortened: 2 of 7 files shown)


happppp:
Is it really the scaling that's wrong? To me it looks like it's antialiasing and blends the edges to black instead of whatever's the background of the font (eg. orange on the warning screen). It's still visible on the higher pixel density screenshot.

vklachkov:
The dark fringing around antialiased edges is a pre-existing issue unrelated to this PR -- it's visible at any pixel density.

This PR specifically fixes the resolution mismatch where hardware renderers were initializing their buffers at logical (point) dimensions instead of physical pixel dimensions, causing everything to be upscaled. The screenshots show that the text in the screenshots before the fix looks carelessly stretched. It's very noticeable to the eye.

vklachkov:
Since something broke on Github and the images won't open, I'll paste the links to them in text.

```
Before:

https://github.com/user-attachments/assets/80a0110c-1537-4f4f-9c16-f2fc81a9952a
https://github.com/user-attachments/assets/40ed2a6a-db71-4bea-b96b-0d1452acd045

After:

https://github.com/user-attachments/assets/90f88dc3-c63e-4199-aa93-557e599570c3
https://github.com/user-attachments/assets/c4598f8d-5c58-4f54-ae77-a2a63bbb671b
```

rb6502:
If I'm reading this right, it's actually improved high DPI support for all SDL3 targets, not just macOS.  That obviously greatly improves MAME's OSD text rendering quality.

vklachkov:
@rb6502 yes, that's how it should be. But I specified that the fix is for MacOS, because I don't have a Windows computer with HiDPI to test the patch) Maybe everything works fine there even without the patch.

rb6502:
I'll check on my Linux machine with a 4K display but assuming SDL3 itself does the right thing it should be fine.

rb6502:
Text rendering on Linux with hiDPI is improved significantly as well with this.  Thanks!

cuavas:
> @rb6502 yes, that's how it should be. But I specified that the fix is for MacOS, because I don't have a Windows computer with HiDPI to test the patch) Maybe everything works fine there even without the patch.

On Windows, we tell the OS that we’re scaling-aware, so everything is in device pixel units (in practice we aren’t – I need to actually implement it for the Win32 debugger, but it requires a setup with multiple displays configured for different logical resolutions, which isn’t how I’m usually set up for working on MAME).

PR #15104, opened 2026-03-14
Fix register saving and restoring for i8087

vklachkov:
This PR fixes registers saving and restoring in i8087. It fixes a whole ton of funny bugs, especially in Compass.


vklachkov:
For example, FlakAttack game for GRiD Compass.

Before (weird line, broken status bar, blinking enemies):



After:

PR #15105, opened 2026-03-15
CD-i: Add Audio Mixing

Vincent-Halver:
CD-i: Adds audio mixing.

Details:

This change creates an attenuation variable, and currently sets it to (L->L and R->R). This is consistent with the current behavior of the system, but provides a partial step towards correctly setting the attenuation variables from the cdislavehle.

At the same time, it also reduces unnecessary buffer allocations by using the unused `m_samples` buffer.

This adds the correct number of bytes to skip over (5) for handling an attenuation command. Otherwise those data bytes may be executed as commands in some circumstances. This reduces crashes in some games (such as the Zelda games).

Finally, this set up a proper callback to handle audio mixing. This allows the fading audio to be heard correctly in the Zelda games.

### DIFF
--- a/src/mame/philips/cdislavehle.cpp
+++ b/src/mame/philips/cdislavehle.cpp
@@ -149,7 +149,7 @@ void cdislave_hle_device::slave_w_mouse(offs_t offset, uint16_t data)
 		if (m_in_index == 1)
 		{
 			LOGMASKED(LOG_COMMANDS, "slave_w: Channel %d: Update Mouse Position (0x%02x)\n", offset, data & 0x00ff);
-			m_in_count = 4;
+			m_in_count = 3;
 		}
 		else
 		{


Vincent-Halver:
This should largely fix #12852 . You can verify this change by trying Zelda Wand of Gamelon, and verifying the menu audio fades when starting a new game.

src/mame/philips/cdicdic.cpp
```diff
@@ -361,15 +352,14 @@ void cdicdic_device::play_cdda_sector(const uint8_t *data)
 	m_dmadac[1]->set_volume(0x100);
 
 	const uint16_t NUM_SAMPLES = SECTOR_SIZE / 4;
-	int16_t samples[2][NUM_SAMPLES];
```
galibert, line 364: Note that this is not an allocation in the malloc/new sense.  It's just a bigger offset on the stack.

PR #15106, opened 2026-03-15
snk/neoprint.cpp: Move into snk/ folder, Hook up NEO-ZMC, Updates

cam900:
- Fix naming
- Fix spacing

machine/neo_zmc.cpp: Make NEO-ZMC as device


angelosa:
Technically this is `neogeo/` folder, not SNK (which in turn means ADK had some actual involvement if only for the same-y sound HW).

cam900:
This PR is now redundant due to https://github.com/mamedev/mame/pull/15705 . so i close this.

angelosa:
That's not how it works ...

src/mame/snk/neoprint.cpp
```diff
@@ -205,9 +210,10 @@ uint8_t neoprint_state::unk_r()
 	/* ---- xx-- one of these two must be high */
 	/* ---- --xx checked right before entering into attract mode, presumably printer/camera related */
 
-	m_vblank = (m_screen->frame_number() & 0x1) ? 0x10 : 0x00;
+	if (!machine().side_effects_disabled())
+		m_vblank = (m_screen->frame_number() & 0x1) ? 0x10 : 0x00;
```
angelosa: Don't think this `m_vblank` should be state-ized and gated against side effects, it looks local to this fn.

angelosa (approved):

angelosa:
Failure has nothing to do with this PR.

cuavas:
As I said last time with a NEO-ZMC device like this, I really don’t like having all the accesses go through trampolines.

Yeah, I’ve got to eventually get back to finishing off rewriting Neo Geo cartridge emulation.  It’s just there are so many things to test, cross-reference against PCB pictures, etc.  The changes to the memory system since I was working on it (e.g. views, easier per-device banks) also mean I need to update my approach for a bunch of the stuff I already did.

PR #15108, opened 2026-03-15
Autodetect homebrew SDL3 on macOS

mengstr:
On macOS, the new SDL3 usage currently assumes a framework-based SDL3 install under /Library/Frameworks/.... This doesn't match the more common Homebrew setup where SDL3 is installed as headers and libraries under the Homebrew prefix and exposed through pkg-config.

This change makes the macOS SDL3 build automatically use library mode when a usable SDL3 library install is detected, while still preserving explicit framework-based configurations.

Captured failure:

```text
Compiling src/osd/modules/input/input_common.cpp...
../../../../../src/osd/modules/input/input_common.cpp:32:10: fatal error: 'SDL3/SDL.h' file not found
   32 | #include <SDL3/SDL.h>
      |          ^~~~~~~~~~~~
1 error generated.
make[2]: *** [osd_sdl3.make:1292: ../../../../osx_clang/obj/x64/Release/osd_sdl3/src/osd/modules/input/input_common.o] Error 1
make[1]: *** [Makefile:19: osd_sdl3] Error 2
make: *** [makefile:1313: macosx_arm64_clang] Error 2

### DIFF
--- a/scripts/src/osd/sdl3.lua
+++ b/scripts/src/osd/sdl3.lua
@@ -114,6 +114,24 @@ function sdlconfigcmd()
 	end
 end
 
+local function sdlpkgconfigcmd()
+	if _OPTIONS["SDL_PKGCONFIG_PATH"] then
+		return path.join(_OPTIONS["SDL_PKGCONFIG_PATH"],"pkg-config")
+	else
+		return pkgconfigcmd()
+	end
+end
+
+local function macosx_uses_libsdl_by_default()
+	if _OPTIONS["targetos"]~="macosx" then
+		return false
+	elseif _OPTIONS["SDL_INSTALL_ROOT"] then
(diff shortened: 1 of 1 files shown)


cuavas:
Is Homebrew really the most common setup for non-technical users?  The trouble with this change is that by default, if you have SDL3 installed in a pkgconfig-aware way, MAME will link against that, and then you don’t get an executable you can give to a non-technical user to use with a drag-and-drop framework installation.

Remember the vast majority of MAME users just want to emulate stuff.  They aren’t developers or “PC enthusiasts” who’d go installing third-party package managers.

This feels analogous to making Windows builds default to dynamically linking against MinGW libraries, requiring a MinGW installation on any system where you want to use them, unless you take additional steps.

PR #15118, opened 2026-03-19
atlus/sprite013.cpp: Make 013 sprite hardware to separated device

cam900:
atlus/cave.cpp:
- Reduce duplicates
- Convert sprite draw routine into single pass
- Add/Fix notes

### DIFF
--- a/src/mame/atlus/sprite013.cpp
+++ b/src/mame/atlus/sprite013.cpp
@@ -124,7 +124,6 @@ TODO:
 
 #include "emu.h"
 #include "sprite013.h"
-#include "screen.h"
 
 DEFINE_DEVICE_TYPE(SPRITE013, sprite013_device, "sprite013", "013 Sprite generator")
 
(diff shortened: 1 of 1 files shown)


cam900:
Like video/tmap038.cpp, This chip is manufactured by NEC and no company marking (at most of chips). but that is used at atlus hardware, probably chip is also designed by atlus for use at their 16 bit hardware. (and other company maked game for that, example: banpresto, namco, cave, etc...)

cam900:
Also, i think theroes* is bootleg/hack of gaia? (both runs at atlus/cave.cpp hardware)

mamehaze:
I think I studied theroes at one point and it just looks like they tweaked/reworked the game and took it to a different publisher.

The program ROMs were substantially different enough that I don't think it's a bootleg / hack, but an entirely separate release.  While it's the same core game engine and mostly the same assets at heart, the changes are substantial and look like they were done with the original codebase at hand.

None of them made it *better* but maybe they wanted a more streamlined experience. (and reworking something you've already made can be a nice quick cash grab when you're struggling)

PR #15121, opened 2026-03-20
m6502: route core fetch/data/write paths through interruptible accesses

dr-who:
This fixes a core integration gap in the 6502 family memory path.

While auditing the base NMOS 6502 against `perfect6502`, I built a small trace-comparison harness that checks bus activity event-by-event across representative instruction, addressing-mode, interrupt, and pause/stall cases. The unstalled traces matched, which ruled out a number of ordinary bus-sequencing errors in the core itself.

The actual problem turned out to be lower down in the memory integration layer: MAME’s wait/delay hooks are applied on `read_interruptible` / `write_interruptible`, but the base `m6502` core and several derived memory interfaces were still using non-interruptible fast paths for normal CPU-visible accesses. That meant memory-originated waits could be bypassed even when the core’s retry/defer machinery was otherwise correct.

A concrete example is `LDA abs` with a 2-cycle delay injected on the final data read:

- Correct timing:
   - final data read moves from cycle 3 to cycle 5
   - next opcode fetch moves from cycle 4 to cycle 6
- Before this patch:
   - the normal fast path could bypass the wait hook, so the read stayed at cycle 3 and the next fetch at cycle 4
- After this patch:
   - the access goes through the interruptible path, so the current bus cycle is stretched and the next fetch is correspondingly delayed to cycle 6

This patch:
  - routes base `m6502` opcode fetch, argument fetch, data read, data write, and refetch paths through the interruptible accessors
  - adds the needed opcode/data specific bindings so those paths use the interruptible dispatch consistently
  - updates the opcode generator so prefetch/refetch sites use the same wrapper path
  - audits the affected family variants with custom memory interfaces (`m6509`, `m6510`, `m4510`, `m50734`, `r65c19`, `rp2a03`, `st2204`, `st2205u`, `xavix`)
  - switches Apple II-family bus DMA hookups to `M6502_DMA_LINE` so DMA/bus denial is handled through the dedicated 6502 path rather than generic HALT suspension

Validation:
  - base/family interruptible source checks pass
  - `perfect6502` comparison suites still pass after the change
  - interrupt, pulse, pause, and stall semantics checks still pass
  - reduced and wider 6502-family compile coverage passed
  - patch is `git diff --check` clean

This is being submitted as a core correctness fix, not as a demonstrated end-user-visible game fix. I have not exhaustively tested games across the affected family, so I’m not claiming specific visible/audio improvements from this PR alone. The claim here is narrower: ordinary 6502-family bus cycles were not consistently reaching the interruptible memory path, so memory wait integration could be structurally invisible.

### DIFF
--- a/src/devices/cpu/m6502/m6502.cpp
+++ b/src/devices/cpu/m6502/m6502.cpp
@@ -49,12 +49,18 @@ void m6502_device::init()
 		// specific group 1-14 or 15-31
 		if(space(AS_PROGRAM).addr_width() > 14) {
 			space(AS_PROGRAM).specific(m_mintf->m_program);
-			space(AS_PROGRAM).specific(m_mintf->m_cprogram);
-			space(has_space(AS_OPCODES) ? AS_OPCODES : AS_PROGRAM).specific(m_mintf->m_csprogram);
+			m_mintf->m_cprogram = m_mintf->m_program;
+			if(has_space(AS_OPCODES))
+				space(AS_OPCODES).specific(m_mintf->m_csprogram);
+			else
+				m_mintf->m_csprogram = m_mintf->m_program;
 		} else {
 			space(AS_PROGRAM).specific(m_mintf->m_program14);
-			space(AS_PROGRAM).specific(m_mintf->m_cprogram14);
-			space(has_space(AS_OPCODES) ? AS_OPCODES : AS_PROGRAM).specific(m_mintf->m_csprogram14);
+			m_mintf->m_cprogram14 = m_mintf->m_program14;
+			if(has_space(AS_OPCODES))
+				space(AS_OPCODES).specific(m_mintf->m_csprogram14);
+			else
+				m_mintf->m_csprogram14 = m_mintf->m_program14;
 		}
 	}
 
@@ -84,7 +90,6 @@ void m6502_device::init()
 	save_item(NAME(m_irq_state));
 	save_item(NAME(m_apu_irq_state));
 	save_item(NAME(m_v_state));
-	save_item(NAME(m_dma_state));
 	save_item(NAME(m_nmi_pending));
 	save_item(NAME(m_irq_taken));
 	save_item(NAME(m_inst_state));
@@ -110,7 +115,6 @@ void m6502_device::init()
 	m_irq_state = false;
 	m_apu_irq_state = false;
 	m_v_state = false;
-	m_dma_state = false;
 	m_nmi_pending = false;
 	m_irq_taken = false;
 	m_inst_state = STATE_RESET;
@@ -131,7 +135,6 @@ void m6502_device::device_reset()
 	m_sync = false;
 	m_sync_w(CLEAR_LINE);
 	m_inhibit_interrupts = false;
-	m_dma_state = false;
 }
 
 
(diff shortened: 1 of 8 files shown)


dr-who:
I was reading https://docs.mamedev.org/techspecs/m6502.html, and thought I would compare against perfect6502.

galibert:
I don't think that works.  Afaict, you're not redoing the accesses when needed.  You test with two cycles shows that things work correctly when the cpu does not actually need to be interrupted.  You should try with a pre-delay/contention of more that what's left in icount and check if the access happens again at the expected time.

dr-who:
You’re right, and I should have been more explicit: the case I tested was the overrun case where the pre-delay is larger than the remaining budget.

The specific check I used was:

  - instruction stream at `0xC000`: `AD 34 12 EA` (`LDA $1234` followed by `NOP`)
  - data at `0x1234`: `0x56`
  - one-cycle execution slices
  - a 2-cycle `before_delay` on the final data read from `0x1234`

  Expected bus sequence:

  - cycle 0: fetch `0xC000`
  - cycle 1: read `0xC001`
  - cycle 2: read `0xC002`
  - cycle 5: read `0x1234`
  - cycle 6: fetch `0xC003`

So this is not the “delay fits inside the current slice” case. It is specifically “1 cycle left, 2 cycles of contention”.

The redo path is:

  - `cpu_device::access_before_delay()` subtracts the requested delay from `icount`
  - if that takes `icount <= 0`, it records the delay tag and sets `m_access_to_be_redone`
  - after each generated 6502 memory micro-op, the opcode state machine checks `access_to_be_redone()`
  - if set, it leaves `m_inst_substate` on the same memory step and returns
  - on the next partial re-entry, the same access is retried, and the saved tag prevents charging the same `before_delay` twice

So the access is retried on the next slice rather than just stretched within the current one.

This was done with a small local trace harness that isn’t part of the PR. If it would help, I can post that minimal reproducer separately or reduce it to a standalone test case.

galibert:
Ohhh, I had forgotten I had added the access_to_be_redone stuff to the 6502 earlier.

There seems to be a number of places where you end up with two specific on the same space, instead of the one cache and one specific that was before.  You should merge them.

What is this dma_line stuff supposed to be?  Will it be used anywhere, and how?

galibert:
Also read_interruptible should be avoided on the state import (e.g. when someone types PC=xxx in the debugger).  Doesn't seem trivial though.

dr-who:
Yes, agreed.

I’ve taken the duplicate-`specific()` cleanup and dropped the Apple/DMA-line part from this PR.

On `state_import`: agreed that `read_interruptible` there is undesirable, but I also agree it’s not trivial. My first pass at avoiding it started pushing family-specific mapper logic into a debugger-only path, which is the wrong direction for this PR. So I’ve dropped that part here and kept this change focused on ordinary CPU-visible fetch/read/write accesses.

The in-tree regression still distinguishes the bug: baseline never sees the delay hook on the final data read, while the patched core retries the access and hits the hook as expected.

galibert:
Oh, FIY, the nmos 6502 part of the code was written with bisimulation with perfect6502, so it's kind of reassuring that you do not find any differences in practice :-)

PR #15123, opened 2026-03-20
cirsa/miniguay.cpp: Add (not working) Cirsa Mini Bar

clawgrip:
----------------------------------
Cirsa Mini Bar [jordigahan]


moulfruit:
nice to see improvement on cirsa system  thank you

PR #15124, opened 2026-03-21
apple2gs: fix aux text page2 shadowing

arekkusu42:
(bug discovered while writing video timing unit tests)

This commit fixes incorrect shadowing of 01/0800...0BFF, which should only occur on a ROM3 and only when C035 bit5 is clear.

80-column text page2 is rarely used, so this probably has little user-visible effect.  But this bug manifested as incorrectly shadowed 80 column text, when hardware does not shadow (inverse @'s leftover from cold boot):
auxtxt2_MAME_286_broken
auxtxt2_HW_unshadowed

Note that this bug only occurred when bank 01 was written to via bank 00 when indirected by C001 80STORE or C005 AUXWRT; it was not triggered by direct writes to bank 01 (so was not caught by the IIgs self-test #0B.)

### DIFF
--- a/src/mame/apple/apple2gs.cpp
+++ b/src/mame/apple/apple2gs.cpp
@@ -2963,7 +2963,6 @@ u8 apple2gs_state::b0ram0800_r(offs_t offset)  { return m_ram_ptr[offset+0x800];
 void apple2gs_state::b0ram0800_w(offs_t offset, u8 data)
 {
 	m_ram_ptr[offset+0x800] = data;
-
 	if (offset < 0x400)
 	{
 		if ((!(m_shadow & SHAD_TXTPG2)) && (m_is_rom3))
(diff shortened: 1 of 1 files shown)


peterferrie:
There's a collection of games in Total Replay that use text-page 2, and we
have special handling for them on the IIGS.
That would make a good regression test.

arekkusu42:
Thanks for that tip.  I added instrumentation to `b1ram0800_w()` to log if this code path is hit, and then _very briefly_ went through [USES_TEXT_PAGE_2 games](https://github.com/search?q=repo%3Aa2-4am%2F4cade+TEXT_PAGE_2&type=code), in the (now very old) Total Replay 5.2 release:  

* beyondwolfen, brainteaserblvd, centipede, davidsmagic, goonies, matingzone, mr.do, neuttower, oilswell, pacman, plasmania, pollywog, snoggle, stellar7, wayout: all appear to play normally

* hyperheadon: hangs on title (also in MAME 0.282, [fixed in TR trunk](https://github.com/a2-4am/4cade/commit/58afeb4e2e1a26b5af5c2728e5f713afc1b64974))

* mazeblox: hangs on title in text mode (also in MAME 0.282, gs and ee)

* ootw: intro movie flickers page-flipping on gs (also in MAME 0.282), works on ee

None of these (40col/GR/HGR) games ever hit this `b1ram0800_w()` code path.

arekkusu42:
More interesting are DGR/DHGR games; again _very briefly_:

* airheart, angrybirds, batman, brucelee, catacombs, pipedream: all hit this code path (continually, in the case of catacombs), but AFAICT there is no user-visible change.

The E1/0800 memory *is* different (now matching hardware!) after this PR, but generally the DHGR games can't see this with the 80-column text screen hidden.  DGR games like Angry Birds seem like possible candidates for a user-visible change...

frankmilliron:
There was a regression in Page 2 handling on the //gs in Total Replay, which has since been fixed.

PR #15132, opened 2026-03-22
osd/sdl: process events after raising window on macOS to ensure input focus

### LINKED ISSUE #10612: MacOS inputs lost on startup
When starting mame on MacOS in fullscreen mode and you move your mouse on startup, the application hangs and never gets to the OSD, and the mouse and normal keyboard input do not work If you don't move your mouse, it will start correctly! If the application hangs and you press CMD+H and minimise the app and then select it again to gain focus on the app, the OSD will then load. I've tested all video driver options (opengl, bgfx, metal), and other options with the mouse but I need help to resolve this issue. 

Note in playing around and changing m_mouse_captured to true in https://github.com/mamedev/mame/blob/master/src/osd/sdl/window.cpp#L1168 will at least get the OSD to load. However, inputs are still lost on the startup OSD. So if you are loading a rom and skip the info OSD then you can play the rom ok with inputs being received.  

Notes:

* I'm running MacOS 12.6
* I'm running SDL2-2.26.0
* I've complied from source using SDL2 framework and also installed from homebrew, but the results are the same.

mfurniss:
## Summary
- On macOS, keyboard and mouse input is intermittently lost on startup because `m_focus_window` is not set when the input modules begin polling
- `SDL_RaiseWindow()` returns before macOS has granted input focus, so the `FOCUS_GAINED` event sits unprocessed in SDL's event queue
- Calling `process_events()` after raising the window dequeues the focus event and sets `m_focus_window` before input polling starts
- Guarded with `#ifdef SDLMAME_MACOSX` since the race is macOS-specific
- Applied to both `sdl/` and `sdl3/` backends

## Test plan
- Tested on Apple M2, macOS 26.3.1
- Launched MAME twenty times in a loop, confirmed `has_focus()` returns true on every run with the fix
- Manual testing confirms keyboard input works on first launch

Fixes #10612

### DIFF
--- a/src/osd/sdl/video.cpp
+++ b/src/osd/sdl/video.cpp
@@ -73,7 +73,17 @@ bool sdl_osd_interface::video_init()
 	}
 
 	if (m_render->is_interactive())
-		SDL_RaiseWindow(dynamic_cast<sdl_window_info &>(*osd_common_t::s_window_list.front()).platform_window());
+	{
+		SDL_Window *const sdlwindow = dynamic_cast<sdl_window_info &>(*osd_common_t::s_window_list.front()).platform_window();
+		SDL_RaiseWindow(sdlwindow);
+
+#ifdef SDLMAME_MACOSX
+		// ensure focus is acquired before the input modules start polling
+		process_events();
+		if (!has_focus())
+			osd_printf_verbose("Window did not acquire input focus\n");
+#endif
+	}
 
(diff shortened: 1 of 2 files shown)


arekkusu42:
Thanks for fixing this.  Should this be [pushed upstream](https://github.com/libsdl-org/SDL/issues/13920)?

PR #15136, opened 2026-03-22
frontend/mame/ui/viewgfx.cpp: Update GFX layout and tilemap viewer

cam900:
emu/drawgfx.h:
- Add function for decode all elements
- Add gfx viewer specific getters

frontend/mame/ui/viewgfx.cpp:
- Add tilemap palette offset display
- Add clip status viewer (clip start : clip size, clip offset in mouse overlay)

aristocrat/aristmk4.cpp, namco/namcos22_v.cpp:
- Use decode_all for predecode gfxs


src/emu/drawgfx.h
```diff
@@ -166,6 +166,12 @@ class gfx_element
 	bool has_pen_usage() const { return !m_pen_usage.empty(); }
 	bool has_palette() const { return m_palette; }
 
+	// used by gfx viewer, get original size regardless source clip
+	u16 origwidth() const { return m_origwidth; }
+	u16 origheight() const { return m_origheight; }
+	u16 startx() const { return m_startx; }
+	u16 starty() const { return m_starty; }
```
cuavas: I don’t like the naming of these accessors.  I’d probably call them `unlcipped_width`, `unclipped_height`, `origin_x` and `origin_y`.  Or maybe `source_width` and `source_height`.

src/emu/drawgfx.h
```diff
@@ -185,10 +191,24 @@ class gfx_element
...
 	void mark_all_dirty() { memset(&m_dirty[0], 1, elements()); }
 
-	const u8 *get_data(u32 code)
+	void clear_dirty(u32 code)
 	{
 		assert(code < elements());
-		if (code < m_dirty.size() && m_dirty[code]) decode(code);
+		if (code < m_dirty.size() && m_dirty[code])
+			decode(code);
+	}
```
cuavas: `clear_dirty` isn’t really clear – it suggests it just clears the flag.  I don’t think using `get_data` to force decoding one tile was really a problem.  The compiler will elide the offset calculation if the result isn’t used, anyway.

src/emu/drawgfx.h
```diff
@@ -185,10 +191,24 @@ class gfx_element
...
+
+	void clear_all_dirty()
+	{
+		for (int i = 0; i < elements(); i++)
+		{
+			clear_dirty(i);
+		}
+	}
```
cuavas: Same with this – the name doesn’t describe what it does.  It could be called `decode_all` or something to make it clearer.

src/emu/drawgfx.h
```diff
@@ -199,6 +219,13 @@ class gfx_element
 		return m_pen_usage[code];
 	}
 
+	// used by gfx viewer, get original data regardless source clip
+	const u8 *get_origdata(u32 code)
+	{
+		clear_dirty(code);
+		return m_gfxdata + code * m_char_modulo;
+	}
```
cuavas: I’d also like a nicer name for this.

src/frontend/mame/ui/viewgfx.cpp
```diff
@@ -1185,8 +1185,8 @@ uint32_t gfx_viewer::handle_gfxset(mame_ui_manager &mui, render_container &conta
 	float cellboxheight = cellboxbounds.height() * float(targheight);
 
 	// compute the number of source pixels in a cell
-	int const cellxpix = 1 + ((set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.height() : gfx.width());
-	int const cellypix = 1 + ((set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.width() : gfx.height());
+	int const cellxpix = 1 + ((set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.origheight() : gfx.origwidth());
+	int const cellypix = 1 + ((set.m_rotate & ORIENTATION_SWAP_XY) ? gfx.origwidth() : gfx.origheight());
```
cuavas: This change is likely to make debugging more difficult as you won’t be able to see bad clipping in the tile viewer.

src/frontend/mame/ui/viewgfx.cpp
```diff
@@ -1415,9 +1418,10 @@ uint32_t gfx_viewer::handle_tilemap(mame_ui_manager &mui, render_container &cont
 	// figure out the title
 	std::ostringstream title_buf;
 	util::stream_format(title_buf,
-			(m_tilemap.flags() != TILEMAP_DRAW_ALL_CATEGORIES) ? _("gfxview", "Tilemap %1$d/%2$d category %3$u") : _("gfxview", "Tilemap %1$d/%2$d "),
+			(m_tilemap.flags() != TILEMAP_DRAW_ALL_CATEGORIES) ? _("gfxview", "Tilemap %1$d/%2$d palette %4$X category %3$u") : _("gfxview", "Tilemap %1$d/%2$d palette %4$X "),
 			m_tilemap.index() + 1, m_machine.tilemap().count(),
-			m_tilemap.flags());
+			m_tilemap.flags(),
+			tilemap.palette_offset());
```
cuavas: I don’t think the palette offset needs to be displayed here.  It will make it look like it’s something that can be altered in the viewer, but it can’t actually be changed.

PR #15139, opened 2026-03-23
CD-i: Refactoring

Vincent-Halver:
This change performs some refactoring, reducing lines of code by 75+.

Additionally, this fixes the memory address of the DVC. Currently, the DVC is not operational, however this does seem to reduce start up time.

### DIFF
--- a/src/mame/skeleton/flashvga2.cpp
+++ b/src/mame/skeleton/flashvga2.cpp
@@ -6,7 +6,8 @@
 "Flash VGA 2" hardware (from TourVisión?) for video-slots.
 
 TODO:
-- Currently fails on timekeeper check ("error en reloj");
+- Currently fails on timekeeper check ("error en reloj"), does extensive checks in the empty
+  dump area;
 
 ===================================================================================================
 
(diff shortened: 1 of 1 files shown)


src/mame/philips/cdi.cpp
```diff
@@ -422,45 +403,32 @@ uint32_t cdi_state::screen_update_cdimono1_lcd(screen_device &screen, bitmap_rgb
...
+	// Skip Device specific program memory
 
 	MCD212(config, m_mcd212, CLOCK_A, m_plane_ram[0], m_plane_ram[1]);
 	m_mcd212->set_screen("screen");
 	m_mcd212->int_callback().set(m_maincpu, FUNC(scc68070_device::int1_w));
 
 	screen_device &screen(SCREEN(config, "screen", SCREEN_TYPE_RASTER));
-	screen.set_raw(960*(312*2-32)*50, 960, 0, 768, 312*2-32, 32, 312*2-32);
+	int lines = 312 * 2 - 32;
+	screen.set_raw(960 * lines * 50, 960, 0, 768, lines, 32, lines);
```
angelosa: 1. lines should be `const`
2. huh? Why it has vtotal identical to display lines? And why it sets pixel clock from the "lines" rather than using an XTAL?

Vincent-Halver: Thanks for the feedback, I need to rework some things.

PR #15143, opened 2026-03-23
Kn5000 interactive control panel

felipesanches:
This is the result of my past many weeks of reverse engineering the control panel protocol for Technics KN5000. The previous kn5000/tlcs900 PRs brought us CPU support for serial port devices and micro-dma transfers and sub-cpu init, which are all somewhat related to this major milestone.

Here we finally fix the subcpu payload loading by storing the code as compressed assets (which is what the real system does) instead of unpacked ones which were better for analysis, but less truthful to the system's memory layout. With this, we don't get "Sound Name Error" messages anymore and, instead, we see actual instrument names in the UI.

image

Also, this provides an HLE implementation of the control panel protocol (because the cpanel microcontrollers are currently not dumped - those are maskrom chips that were not decapped yet).

The protocol is not fully understood yet, but this initial implementation is good enough to get it to a working state in the driver. There are multiple commands which are treated equally, while I suspect they were supposed to have subtle differences in their behavior (otherwise why would there be more than a single command value for each of them?), so additional research will be needed. But for now I think it is acceptable to at least have something that allows interactive use and which is not a hack.

I will not mark this as a draft because I would be happy to have it merged as is. But I am also open to feedback from code-reviewers and I can make additional changes if needed. Just bare in mind that HLE implementation is, by definition, speculative.

Let me know what you think and I'll be glad to eventually make additional commits.

A short video can also be seen at https://www.youtube.com/watch?v=gX5oR9YOMJE

### DIFF
--- a/src/mame/layout/kn5000.lay
+++ b/src/mame/layout/kn5000.lay
@@ -757,79 +757,79 @@ license:CC0-1.0
 	<element ref="dark_round_button" inputtag="CPL_SEG7" inputmask="0x08"><!-- screen button: exit --><bounds x="889" y="362" width="29" height="29"/></element>
 
 	<element ref="dark_37x21_button" inputtag="CPL_SEG1" inputmask="0x01"><bounds x="227" y="477" width="37" height="21"/></element>
-	<element name="CPL_1" ref="green_led"><!-- COMPOSER: MEMORY --><bounds x="242" y="469" width="5" height="5"/></element>
+	<element name="cpl_led_1" ref="green_led"><!-- COMPOSER: MEMORY --><bounds x="242" y="469" width="5" height="5"/></element>
 
 	<element ref="dark_37x21_button" inputtag="CPL_SEG1" inputmask="0x02"><bounds x="262" y="477" width="37" height="21"/></element>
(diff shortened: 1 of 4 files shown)


rb6502:
Question about the subcpu payload: if it's stored in IC19 and we have a dump of IC19, why isn't it just used from there?

felipesanches:
> Question about the subcpu payload: if it's stored in IC19 and we have a dump of IC19, why isn't it just used from there?

If I recall correctly, the dump of IC19 was made via a desoldered IC and it lacked the subcpu payload. It is usually saved to that range of that Flash ROM as part of the system update procedures (installed from floppies).

I got the payload by extracting the system update files. Both maincpu and subcpu programs are LZSS compressed in the floppies. Main CPU one is decompressed before saving to flash, but the subcpu one is kept compressed. It is the boot code that decompresses the payload from IC19 flash right before transferring it to subcpu RAM via inter-cpu communication latches. This happens every time you turn the KN5000 on.

rb6502:
So it's overlaying the appropriate -bios version of the compressed subcpu program on the presumably factory-clear IC19 and then everything works like hardware.  Cool.

PR #15144, opened 2026-03-24
sh/sh7709s.cpp: Add support for cache invalidation and entry writeback

jesus-ramos:
Map the cache address array to handle cache line flushing as well as writes to CCR that cause a cache invalidatation

### DIFF
--- a/src/devices/cpu/sh/sh7709s.cpp
+++ b/src/devices/cpu/sh/sh7709s.cpp
@@ -13,6 +13,33 @@
 // - TLB simulation + timing and lookups
 // - Branch predicition/mispredict penalties - Needs a branch target buffer and some basic tracking, didn't make a huge difference for cv1k when I threw one together
 
+/* cv1k notes
+*
+* - Most slowdown is caused from the working size of the current frame exceeding the cache size
+*   causing a lot of acceses to memory leaving the cpu idling. Cave doesn't seem to use prefetch
+*   to alleviate this anywhere so you're just cache trashing at a certain point. This is what leads
+*   to the iconic slingshotting in early titles such as Mushihimesama + Futari where you go from
+*   sections of slowdown right back to regular speed as bullets leave the screen.
+*
+* - There seems to be a bug in the ROM read code in cv1k titles that is present in all of them.
+*   This code when ROM reads are queued for whatever reason doesn't check if a read or write
+*   is done and always ends up doing an expensive cache flush + invalidate that isn't required.
+*   The processor ends up churning a bunch of cycles on uncached fetches and then also has to repopulate
+*   the cache.
+*
+*   This requires looping 1024 times over every cache entry from an uncached aliased mapping in SDRAM
+*   which comes with hefty access penalties due to all the instruction fetches + some data fetches.
+*   While the code is present in every cv1k title the performance problems start manifesting more and
+*   more in the later titles, the final two cv1k titles being the worst offenders with regards to
+*   slowdown sections that at a glance don't look like they should be slowing down or causing extra
+*   slowdown.
+*
+* - The code that reads irr0 to check for irq2 also seems to have a ready modify write timing bug where
+*   the register value is read, IRQ2 is masked out of that value, then that masked value is written back.
+*   Documentation does mention in an addendum that this behavior can lead to lost IRQ's but luckily these
+*   games don't seem to actually use them in the release versions.
+*/
+
 #include "emu.h"
 #include "sh7709s.h"
 #include "sh3comn.h"
@@ -76,7 +103,7 @@ bool is_cacheable(uint32_t address)
 	return region != 0x5 && region != 0x7;
 }
 
-unsigned int get_area(uint32_t address)
+uint32_t get_area(uint32_t address)
 {
 	// Mask to 29 bit physical space
 	uint32_t phys_mask = address & SH34_AM;
@@ -97,6 +124,10 @@ bool sh7709s_device::cache_access(uint32_t address, bool write)
 {
 	if (!is_cacheable(address))
 		return false;
+
+	// Programs should ensure that 2 pointers cannot alias to the same physical address
+	// as the cache discards the top 3 region bits
+	address &= SH34_AM;
 
 	uint32_t cache_address = address / SH7709S_CACHE_LINE_SIZE;
 	uint32_t cache_block = cache_address % SH7709S_CACHE_BLOCKS;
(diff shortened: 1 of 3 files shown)


jesus-ramos:
Some cv1k titles only do this during loading screens (mushi+mushi futari) so they were already close to pcb timing. Some titles like muchi muchi pork flush the cache and invalidate it every frame so it should bring those a bit closer to pcb.

Still some missing sources of slowdown to track down but I'll have to decomp and read through them a bit more. Mushi+mushi futari aside, those seem to be good already.

src/devices/cpu/sh/sh7709s.cpp
```diff
@@ -537,6 +537,76 @@ void sh7709s_device::static_generate_memory_accessor(int size, int iswrite, cons
...
+		// throwing out some data
+		memset(m_cache, 0, sizeof(m_cache));
+		LOG("SH7709S cache invalidate\n");
+	}
+	COMBINE_DATA(&m_ccr);
+	logerror("'%s' (%08x): CCN unmapped internal write %08x & %08x (CCR)\n", tag(), m_sh2_state->pc, data, mem_mask);
+}
+
+#define CACHE_MAPPING_BASE (0xF0000000)
+#define CACHE_MAPPING_END (0xF0FFFFFF)
```
ajrhacker: These should be constexpr variables rather than macros.

ajrhacker (changes requested):

src/devices/cpu/sh/sh7709s.cpp
jesus-ramos: Fixed those up

happppp:
About your sh7709s.h/sh7709s.cpp files (not this PR):

for each "unsigned int": just use uint32_t

```
	for (int i = 0; i < SH7709S_CACHE_BLOCKS; i++)
		for (int j = 0; j < SH7709S_CACHE_ASSOCIATIVITY; j++)
		{
			save_item(NAME(m_cache[i][j].tag), j + (i * SH7709S_CACHE_BLOCKS));
			save_item(NAME(m_cache[i][j].lru), j + (i * SH7709S_CACHE_BLOCKS));
			save_item(NAME(m_cache[i][j].dirty), j + (i * SH7709S_CACHE_BLOCKS));
		}

change to this:
	save_item(STRUCT_MEMBER(m_cache, tag));
	save_item(STRUCT_MEMBER(m_cache, lru));
	save_item(STRUCT_MEMBER(m_cache, dirty));
```

Why are these global in the .h file? Does anything else other than sh7709s.cpp need them? If no, better place them under private: as static constexpr uint32_t
```
#define SH7709S_CACHE_SIZE (16384)
#define SH7709S_CACHE_LINE_SIZE (16)
#define SH7709S_CACHE_ENTRY_COUNT (SH7709S_CACHE_SIZE / SH7709S_CACHE_LINE_SIZE)
#define SH7709S_CACHE_ASSOCIATIVITY (4)
#define SH7709S_CACHE_BLOCKS (SH7709S_CACHE_ENTRY_COUNT / SH7709S_CACHE_ASSOCIATIVITY)
```

Why do you have C functions mixed in sh7709s.cpp? (as in: not class member functions)

jesus-ramos:
> About your sh7709s.h/sh7709s.cpp files (not this PR):
> 
> for each "unsigned int": just use uint32_t
> 
> ```
> 	for (int i = 0; i < SH7709S_CACHE_BLOCKS; i++)
> 		for (int j = 0; j < SH7709S_CACHE_ASSOCIATIVITY; j++)
> 		{
> 			save_item(NAME(m_cache[i][j].tag), j + (i * SH7709S_CACHE_BLOCKS));
> 			save_item(NAME(m_cache[i][j].lru), j + (i * SH7709S_CACHE_BLOCKS));
> 			save_item(NAME(m_cache[i][j].dirty), j + (i * SH7709S_CACHE_BLOCKS));
> 		}
> 
> change to this:
> 	save_item(STRUCT_MEMBER(m_cache, tag));
> 	save_item(STRUCT_MEMBER(m_cache, lru));
> 	save_item(STRUCT_MEMBER(m_cache, dirty));
> ```
> 
> Why are these global in the .h file? Does anything else other than sh7709s.cpp need them? If no, better place them under private: as static constexpr uint32_t
> 
> ```
> #define SH7709S_CACHE_SIZE (16384)
> #define SH7709S_CACHE_LINE_SIZE (16)
> #define SH7709S_CACHE_ENTRY_COUNT (SH7709S_CACHE_SIZE / SH7709S_CACHE_LINE_SIZE)
> #define SH7709S_CACHE_ASSOCIATIVITY (4)
> #define SH7709S_CACHE_BLOCKS (SH7709S_CACHE_ENTRY_COUNT / SH7709S_CACHE_ASSOCIATIVITY)
> ```
> 
> Why do you have C functions mixed in sh7709s.cpp? (as in: not class member functions)

Thanks for the feedback, I can get those cleaned up in a followup PR. Already had some local changes to clean up the mcr/wcr register timings to be member functions since they all access some member registers anyway.

jesus-ramos:
Still some missing slowdown mostly in the unsupported ddpsdoj on expert difficulty. It does seem to be in areas that do invalidation and flush during gameplay so it's possible my numbers are off there still or some behavior I'm missing. Examples in expert are s4 midboss bomb drop pattern s5 opening sequence after the first zako rush, s3 boss final pattern, s5 boss a good bit of the patterns most notably the first jetbachi pattern and the mirrors 2nd to last pattern. 

Not good enough at pink sweets to test the high rank slowdown on later bosses that was missing but the invalidates should help there if anything.

jesus-ramos:
Edit: The cache address accesses don't go through the peripheral bus, I just misinterpreted one part of the docs for some internal memory mappings.

This patch should be good to go for now at least since the invalidates do add up to a decent amount of misses on subsequent frames already. 

Side note: This is a really weird way of adding "intentional" slowdown in some of these titles but I can see why some patterns may need to be smoothed as they are quite jerky otherwise, but there are a couple of patterns where theres no slowdown normally and the same method seems to be used to add slowdown rather than smooth it which does create some stuttery/jerky slowdown so who knows what cave was up to with this.

jesus-ramos:
@ajrhacker sorry for ping, did you have any other changes you wanted?

jesus-ramos:
https://www.youtube.com/watch?v=XUq3xkKylR4
Good test by aquas of the latest build. Last boss did not cooperate on the big pattern that causes some real nasty stutter on the pcb but s3 lasers now behave properly.

https://www.twitch.tv/videos/2783216999?t=1h30m32s
Rainbow pattern max rank stutter looking good now too

Overall getting real close across the board so all the games are definitely very playable now without having to mess around with random sliders anymore and not many instances left of missing slowdown at least in gameplay critical sections. 

Some TODO's to get to for a later patchset that I added in comments along with a bit of breakdown on some of the slowdown behavior as well as some interesting bits I found reading through the disassembled games.

angelosa:
Do we have any real objection about this?

PR #15148, opened 2026-03-24
x07: remove "anti bug" code from T6834 IO write

Mokona:
Some MESS-era code was trying to circumvent unimplemented T6834 at that time, but created other problems, as the heuristic on what is a new command was wrong for some cases.

Fixes the "PRINT"A";:C=TKEY("O")" bug freezing the X07 emulation, where the T6834 issued by TKEY was mistakenly interpreted as a character to print.

### DIFF
--- a/src/mame/canon/x07.cpp
+++ b/src/mame/canon/x07.cpp
@@ -1204,7 +1204,7 @@ void x07_state::x07_io_w(offs_t offset, uint8_t data)
 		break;
 
 	case 0xbb:
-		// Missing IRQ control
+		// TODO: Missing IRQ/RST control (User manual page 108)
 		break;
 	}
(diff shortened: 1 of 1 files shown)


src/mame/canon/x07.cpp
```diff
@@ -1236,6 +1202,10 @@ void x07_state::x07_io_w(offs_t offset, uint8_t data)
 
 		m_regs_w[5] = data;
 		break;
+
+	case 0xbb:
+		// Missing IRQ control
```
angelosa: Missing as in a `// TODO:`?

Mokona: Indeed. Updating the comment with the reference.

PR #15158, opened 2026-03-27
emu/drawgfx.cpp: Added alternative render which doesn't allow color index escape palette range.

holub:
I'm adding variant to the existing render which helps index not overflow palette limits.

In the existing implementation let's say we have:

* 256 pens in the palette
* granularity 16 and requested tile color 8
* initial calculation will move us to offset 128: `const pen_t *paldata = m_palette->pens() + colorbase() + granularity() * (color % colors());`
* then _pixelop_: `(DEST) = paldata[srcdata];`

Imagine within this configuration pixels in tile are still using _256_ pens. Then _pixelop_ can access the index **>256**

I personally suspect initial implementation was done for performance but has luck of awarding such cases.

### DIFF
--- a/src/emu/drawgfx.cpp
+++ b/src/emu/drawgfx.cpp
@@ -1198,7 +1198,7 @@ void gfx_element::prio_zoom_opaque(bitmap_rgb32 &dest, const rectangle &cliprect
 void gfx_element::prio_zoom_transpen(bitmap_ind16 &dest, const rectangle &cliprect,
 		u32 code, u32 color, int flipx, int flipy, s32 destx, s32 desty,
 		u32 scalex, u32 scaley, bitmap_ind8 &priority, u32 pmask,
-		u32 trans_pen)
+		u32 trans_pen, bool colorwraps)
 {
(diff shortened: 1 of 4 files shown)


happppp:
Are there really drivers that rely on out of bounds palette index? I'd assume if a driver does it, it's a bug. Is it not mainly for performance, but maybe for stuff like shadow pens or palette.black_pen()

This PR: Can't the wrapping be done in specnext_sprites.cpp?

holub:
I was trying to isolate this in Next in the first place but failed.

Yes driver does such things: it has 8 palettes and masking index with MOD 256. That's not what MAME can offer at this moment. Another issue is that I had to merge 8 palettes in one and that's how the issue was identified.

On the other hand I don't think that's a typical situation for any driver which uses more colors than granularity defines. That's why existing implementation didn't fail for anyone - they just never reaching this edgecase. If that's the case, existing implementation can be entirely replaces by mine and nobody going to be affected.

holub:
Tabling this and trying to find solution on Next's side.

PR #15162, opened 2026-03-27
bus/centronics: add NEC PC-6022 plotter

devinacker:
Also implemented `AS_IO` for the uPD7801 as needed.

### DIFF
--- a/src/devices/bus/centronics/pc6022.cpp
+++ b/src/devices/bus/centronics/pc6022.cpp
@@ -72,7 +72,7 @@ void pc6022_device::io_map(address_map &map)
 {
 	map.global_mask(0x00ff);
 	// TODO: does the address actually matter?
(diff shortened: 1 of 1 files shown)


src/devices/bus/centronics/pc6022.cpp
```diff
@@ -0,0 +1,146 @@
...
+	m_cpu(*this, "cpu"),
+	m_busy(*this, "busy")
+{
+}
+
+void pc6022_device::io_map(address_map &map)
+{
+	map.global_mask(0x00ff);
+	// TODO: does the address actually matter?
+	map(0x00, 0x00).r(NAME(pc6022_device::data_r));
```
ajrhacker: Changing NAME to FUNC here will probably fix the build errors.

PR #15163, opened 2026-03-28
gridcomp/gridcomp.cpp: support switching to widescreen resolution

vklachkov:
After power on, the Compass II is in compatibility mode with the Compass I and only displays 320x240 pixels, instead of 512x256, and the image is centered. In this mode we see the GRiD-OS boot picture and the MS-DOS 2.11 welcome message. Operating systems switch to widescreen resolution during the boot process. If the framebuffer is always renders at 512x256, as it was before the fix, the user will see garbage for a few seconds after power-on due to the incorrect resolution.

This PR adds a one-way switch from square resolution to widescreen.

It is unknown which specific chip is responsible for switching video modes and how to switch back to 320x240 resolution. This fix is based on analyzing memory accesses from the two operating systems, not on reverse engineering.

Image when booting GRiD-OS 3.1.5 before the fix:

Screenshot before fix

And after:

Screenshot after fix

### DIFF
--- a/src/mame/gridcomp/gridcomp.cpp
+++ b/src/mame/gridcomp/gridcomp.cpp
@@ -142,9 +142,11 @@ private:
 
 	uint16_t *m_videoram = nullptr;
 
+	bool m_video_chip_selected = false;
+	bool m_widescreen_activated = false;
+
 	IRQ_CALLBACK_MEMBER(irq_callback);
 
-	uint16_t grid_9ff0_r(offs_t offset);
 	uint16_t grid_keyb_r(offs_t offset);
 	uint8_t grid_modem_r(offs_t offset);
 	void grid_keyb_w(offs_t offset, uint16_t data);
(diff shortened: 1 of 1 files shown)


vklachkov:
Here I want to leave a comment about the screen resolution. Currently, the resolution is hardcoded in the file gridcomp.cpp. But in general, for the Compass II, the screen resolution and aspect ratio are specified at the end of the CCPROM, right after the jump to start.

It contains something like this:

```
DataTable  DW ScreenAddr
DataTable1 DW ScreenWidth
DataTable2 DW ScreenHeight
DataTable3 DW AspectRatio
DataTable4 DW GpibBuffaloSplit
```

Perhaps in the future, the hardcoding should be removed and the values should be taken from the CCPROM as intended. And only if these values are missing, fall back to 320x240 like the Compass 1101.

vklachkov:
I want to rewrite it properly right away.

PR #15165, opened 2026-03-28
{flopimg,esq16_dsk}.cpp: Allow specifying explicit start and end sectors when getting PC MFM track data.

cbrunschen:
Ensoniq VFX and EPS family floppies use 10 sectors per track, starting at sector 0 rather than the more common sector 1. The code in `esq16_dsk.cpp` handles this when loading floppy images, but not when saving them. This attempts to fix this in a way that is not only specific to Ensoniq's floppy images, but to possibly other future ones as well.

In `flopimg.cpp`, allow specifying explicit start and end sectors when getting PC MFM track data.

In `esq16_dsk.cpp`, use this to explicitly specifying sectors 0 to 9 when getting the track data to save.

### DIFF
--- a/src/lib/formats/esq16_dsk.cpp
+++ b/src/lib/formats/esq16_dsk.cpp
@@ -138,12 +138,12 @@ bool esqimg_format::save(util::random_read_write &io, const std::vector<uint32_t
 	if(sector_count != 10)
 		sector_count = 10;
 
-	uint8_t sectdata[11*512];
+	uint8_t sectdata[10*512];
 	int track_size = sector_count*512;
 
 	for(int track=0; track < track_count; track++) {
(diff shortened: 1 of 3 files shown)


galibert:
Any reason it's not just a upd765_format derivative?

cbrunschen:
> Any reason it's not just a upd765_format derivative?

I had received a bug report, and made what seemed to be a minimal the change within the existing code.

galibert:
That is not answering the question though.

cbrunschen:
> That is not answering the question though.

I'm not really the right person to answer that question, as I didn't originally write this code - I only made, as I described, what seemed to be a reasonable change within the existing code.

PR #15167, opened 2026-03-28
sinclair/specnext.cpp: Duplicate palette workaround which doesn't allow color index escape (sprite) palette boundaries.

holub:


holub:
pong

PR #15168, opened 2026-03-29
dinput joystick: use DIJOYSTATE2 for 128-button support

d-b-c-e:
## Summary

Use `c_dfDIJoystick2` as the primary data format for DirectInput joystick devices, with `c_dfDIJoystick` as a fallback. This raises the button limit from 32 (`DIJOYSTATE`) to 128 (`DIJOYSTATE2`).

## Problem

MAME's DirectInput joystick handler uses `DIJOYSTATE` which only supports 32 buttons. Many modern input devices exceed this — racing wheel hubs (Moza, Fanatec), HOTAS throttles (Virpil, VKB), and USB button boxes commonly have 64-128 buttons. Buttons above index 32 are silently ignored, with no config option to change this.

## Changes (3 lines, 2 files)

- `input_dinput.h`: `DIJOYSTATE` → `DIJOYSTATE2` in the joystick state struct
- `input_dinput.cpp`: Request `c_dfDIJoystick2` with `c_dfDIJoystick` fallback (same pattern the mouse device already uses with `c_dfDIMouse2`/`c_dfDIMouse`)
- `input_dinput.cpp`: Cap `dwButtons` based on which format was accepted (32 or 128) to prevent buffer overread when the fallback format is used

## Why this is safe

- **Backwards compatible**: If a device/driver doesn't support `DIJOYSTATE2`, `open_device()` automatically falls back to `c_dfDIJoystick` — exactly how the mouse device already handles `c_dfDIMouse2`→`c_dfDIMouse`
- **No new APIs**: `DIJOYSTATE2` and `c_dfDIJoystick2` are standard DirectInput8 structures, available since Windows 2000
- **Existing code already handles 128 buttons**: `configure()` already assigns `ITEM_ID_BUTTON1`-`BUTTON32` for buttons 0-31, `ADD_SWITCH` for 32-47, and `OTHER_SWITCH` for 48+. The constructor already caps at 128. The `item_name()` calls already use `DIJOYSTATE2` offsets. The infrastructure was clearly designed for this — it just wasn't wired up.
- **Format-aware cap**: The button count is now capped to match the accepted format (32 for `c_dfDIJoystick`, 128 for `c_dfDIJoystick2`), preventing the potential buffer overread that existed in the old code where `dwButtons` was capped at 128 but the state buffer only held 32 buttons

## Testing

- Built and tested on Windows 11 with MAME 0.286
- Verified with Moza R12 racing wheel (128 buttons reported by DirectInput)
- Buttons 33+ correctly appear as `Joy N Button XX` in MAME's input configuration
- No regression with standard gamepads (Xbox 360 controller, 14 buttons)
- Fallback path verified: no "Error setting primary data format" message in verbose output (device accepted `DIJOYSTATE2` directly)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/osd/modules/input/input_dinput.cpp
+++ b/src/osd/modules/input/input_dinput.cpp
@@ -452,8 +452,8 @@ protected:
 		auto devinfo = m_dinput_helper->create_device<dinput_joystick_device>(
 				*this,
 				instance,
+				&c_dfDIJoystick2,
 				&c_dfDIJoystick,
-				nullptr,
 				background_input() ? dinput_cooperative_level::BACKGROUND : dinput_cooperative_level::FOREGROUND,
 				[] (auto const &device, auto const &format) -> bool
 				{
@@ -560,7 +560,7 @@ dinput_joystick_device::dinput_joystick_device(
 	// cap the number of axes, POVs, and buttons based on the format
 	m_caps.dwAxes = std::min(m_caps.dwAxes, DWORD(8));
 	m_caps.dwPOVs = std::min(m_caps.dwPOVs, DWORD(4));
-	m_caps.dwButtons = std::min(m_caps.dwButtons, DWORD(128));
+	m_caps.dwButtons = std::min(m_caps.dwButtons, DWORD((m_format == &c_dfDIJoystick) ? 32 : 128));
 }
 
 void dinput_joystick_device::reset()
--- a/src/osd/modules/input/input_dinput.h
+++ b/src/osd/modules/input/input_dinput.h
@@ -174,7 +174,7 @@ private:
 	// state information for a joystick; DirectInput state must be first element
 	struct dinput_joystick_state
 	{
-		DIJOYSTATE  state;
+		DIJOYSTATE2 state;
 		LONG        rangemin[8];
 		LONG        rangemax[8];
 	};


MooglyGuy:
You needed to resort to AI for a 3-line change?

d-b-c-e:
I could spend 3 days tracking down the issue and getting a working fix by hand, or spend literally about 20 minutes AI assisted start to finish.  Without AI I probably would have never taken the time to track it down.  Not because I'm incapable, but because I don't have that kind of time,  and clearly neither would anyone else since a 3 line change goes unfixed for years and years. 
    
Take or leave the change - I fixed it for myself and thought someone else could benefit from it.  

Thanks for being condescending, mate.

cuavas:
This won't work – MAME has an upper limit of 32 "button" type controls power input device instance.

cuavas:
> Take or leave the change - I fixed it for myself and thought someone else could benefit from it.

There are upper limits for the number of "button" and "additional switch" controls (defined on src/osd/interface/inputcode.h). You aren't checking against the upper limit for "additional switch" controls, which means something bad will happen, probably memory corruption, if a controller really does have a large number of buttons.

Since this is a "take it or leave it PR", there's only one option.

> I could spend 3 days tracking down the issue and getting a working fix by hand, or spend literally about 20 minutes AI assisted start to finish. Without AI I probably would have never taken the time to track it down. Not because I'm incapable, but because I don't have that kind of time, and clearly neither would anyone else since a 3 line change goes unfixed for years and years.

And as usual, the so-called fix from the so-called "AI" looks kind of plausible, but it's fundamentally flawed.

Don't outsource thinking to a statistical model, kids.

d-b-c-e:
> This won't work – MAME has an upper limit of 32 "button" type controls power input device instance.

I appreciate the specific technical feedback on the second point.

To clarify: this PR changes only the DirectInput data format (DIJOYSTATE → DIJOYSTATE2) and the button count cap in the constructor. The three-tier button routing in configure() — which routes buttons 0-31 to ITEM_ID_BUTTON, 32-47 to ITEM_ID_ADD_SWITCH, and 48+ to ITEM_ID_OTHER_SWITCH — is existing MAME code that predates this PR. The add_item() function handles OTHER_SWITCH via dynamic slot allocation in the ITEM_ID_MAXIMUM+1 to ITEM_ID_ABSOLUTE_MAXIMUM range.

If there is a bounds-checking concern with the ADD_SWITCH range, that would be a pre-existing issue in configure(), not introduced by this change. I'm happy to add an explicit bounds check there if that's what's needed to make this acceptable:

```
 else if (butnum < INPUT_MAX_BUTTONS + INPUT_MAX_ADD_SWITCH)
     itemid = input_item_id(ITEM_ID_ADD_SWITCH1 + butnum - INPUT_MAX_BUTTONS);
 else if (butnum < std::size(reinterpret_cast<DIJOYSTATE2*>(0)->rgbButtons))
     itemid = ITEM_ID_OTHER_SWITCH;
 else
     break; // safety cap

```
Regarding the AI discussion — the tool helped me find the right three files to change in a large codebase I'd never worked in before. The analysis of why the fix works (data format bottleneck, existing overflow routing) is mine. Happy to resubmit with additional safety checks if the project would find it useful.   Otherwise I can only attest that it does work well in my testing on my specific use case with no ill side effects I've seen yet - not to say there isn't some edge case I'm not considering but I'm open to digging deeper.  If all else fails, I'll just fork or keep the change local.

cuavas:
Look, your PR description is a pile of overly verbose chatbot output that's nauseating to read. A concise description of the change is far more useful, and shows you've made at least some effort to understand it.

We've also had several terrible "AI"-generated PRs submitted recently where it was clear the submitter didn't understand what they were doing, and they responded to follow-up questions with even more chatbot spew.

I'm going to add an explicit rule against using these kinds of tools to create PRs to the contributing guidelines. You, the human, are responsible for understanding the changes you make; we are not interested in wasting our time interacting with chatbots through a layer of indirection.

Also, I've had pretty bad experiences trying to help people use driving controls with MAME in the past.
* A few years ago, I made a months-long effort to improve how MAME handles input devices. I solicited user feedback throughout the process. I got useful feedback from people using conventional controllers, rhythm game controllers, and flight simulation controls. I got no feedback at all from people using driving controls. But immediately after the release, we got multiple hostile false bug reports from people with driving controls who hadn't updated their configuration, despite the release notes mentioning that this may be necessary.
* In general, people issuing driving controls also seem to be unusually prone to not reading the documentation, complaining that it doesn't work, ignoring what you say when you try to help, and not providing any useless information to debug actual issues.
* There's virtually no technical documentation available for driving controls (e.g. the HID usage codes they report, how the driver maps it to DirectInput, etc.). There's a proprietary SDK for Logitech driving controllers, but it's virtually undocumented as well, and doesn't provide a way to simulate a controller you don't have for testing. They make no effort to encourage develops to support their products.

The thing is, no MAME developers actually hit the button limit in `DIJOYSTATE` – it's comfortably higher than the dozen or so buttons you'll find on conventional controllers. Specialised controllers take money and space most of us can't justify for something that we'd only use for testing a piece of software that we use our personal time to develop.  You're complaining about it not changing in years, but it's something none of us can test properly, only seems to be useful for appearing a small and particularly unhelpful and obnoxious subset of users.

I'd honestly forgotten that there was already a range check on the item IDs for DirectInput joystick buttons, even though I probably added it myself (I knew there was a range check in SDL). I'm still not keen on changing this stuff without a practical way to test it. I'm still very wary of a 20 minute change, which really isn't enough time to think through the implications.

d-b-c-e:
And that's fine man, I'm not trying to start something.  I had a problem, I fixed and tested it on my setup, thought somebody else would benefit from it and tossed it out there.  I've been a dev 20 years so I'm not a script kiddy vibe-coding stuff but I admit without AI I wouldn't have taken the time to bother with this fix because I get a couple hours a week if I'm lucky to tinker with this stuff outside work and I haven't touched C++ in years.  The code is out there for the next wheel user like me to now download and build their own just like I did, so mission accomplished as far as I'm concerned.  

I didn't realize the use of Claude Opus was so contentious (I do get most AI give way worse results though) - so I'll just stick to forking and keeping my own changes moving forward.  And to be fair, let's assume I didn't even use AI - there's no damn way in the world I could understand all the implications in a code base this old and complicated without being steeped in it for months at a time.  

I know you're probably seeing two dozen of these a week and it's probably a serious pain point for you right now being stuck between a rock and a hard place, and for that I apologize for my tone.  At the end of the day, I was just trying to help somebody else that just wanted to use their Start and Select button that Moza decided needed to be 33 and 34 for whatever damn reason.

galibert:
So, it's not >32 buttons that you need, it's devices with way less buttons but with a button index >= 32.  That, for a start, would have been a lot more interesting to know.

d-b-c-e:
Yep - apologies if that's a clear distinction I didn't make.  Probably why you only hear this from a small subset of people using racing and flight sim gear as double duty for playing MAME games.  While they do have a crap ton of buttons we still only use about the same number as anybody else, but the button indexes tend to be higher in 30s-40s on some gear.  

There is using vJoy to remap the buttons to lower indexes but I for my particular case I just found fixing these few lines to work pretty gracefully, but I accept there's something downwind somewhere I can't see it might affect.

PR #15170, opened 2026-03-30
namco/namco_c123tmap.cpp, namco/namco_c169roz.cpp, shared/namco_c355spr.cpp: Use device_delegate for callbacks

cam900:


src/mame/shared/namco_c355spr.h
```diff
@@ -42,23 +51,6 @@ class namco_c355spr_device : public device_t, public device_gfx_interface, publi
...
-	typedef device_delegate<int(int)> c355_priority_delegate;
-	typedef device_delegate<bool (u16 &dest, u8 &destpri, u16 colbase, u16 src, int srcpri, int pri)> c355_mix_delegate;
-
-	void set_tile_callback(c355_obj_code2tile_delegate cb)
-	{
-		if (!cb.isnull())
-			m_code2tile = cb;
-		else
-			m_code2tile = c355_obj_code2tile_delegate(&namco_c355spr_device::default_code2tile, this);
-	}
```
cuavas, line 59: This change makes it impossible to restore the default callback after changing it.  Probably no an issue, but it is a semantic change.

PR #15172, opened 2026-03-30
namco/sg_vga.cpp: Add DX-101 device

cam900:
- Add/Fix note:
  - I don't think DX-102 is part of video hardware. (same reason as seta/seta2.cpp)
  - X1-010 ROM is possibly bankswitched, bank method is similar as seta/seta2.cpp?


cuavas:
Nice, devicifying stuff pays off again.

Issue #9170, opened 2022-01-17
m68k: need DA regs save/restore for 68010/68020 custom MMU support


agentbooth:
The AT&T UNIX PC driver (unixpc.cpp), using a 68010 with custom MMU, requires support for page faults which trigger a bus error, that requires the DA regs to be saved (prior to the PF) and restored prior to the exception handling.

This feature is currently supported for the PMMU implementation of 68851/68030/68040, but needs to be extended to cover 68010 and 68020 using custom or 68451 MMU.

Current behavior:
`m_maincpu->set_input_line(M68K_LINE_BUSERROR, ASSERT_LINE);`
calls `m68k_cause_bus_error() `which does:
`m68ki_init_exception()`
`m68ki_stack_frame_1000()` // 68010 specific
`m68ki_jump_vector(EXCEPTION_BUS_ERROR)`
however, the DA regs are not (saved and) restored prior to the exception handling.

`execute_run()` in m68kcpu.cpp covers this situation for `m_pmmu_enabled`, but not for 68010 or 68020 with custom MMU.

Comment from @rb6502:
> For the case where you need to restart instructions from an external MMU, `m68k execute_run()` needs a separate flag `m_has_instruction_restart` which all of the MMU configurations set and which is also settable externally so it can be enabled for 68010 systems with external MMUs.

This is also potentially relevant for:
- Convergent MiniFrame (68010, custom MMU)
- Sun-2 (68010, custom MMU)
- Sun-3 (68020, custom MMU)
- HP 9000/310 (68010, custom MMU)
- HP 9000/320 (68020, custom MMU)
- MVME 120 (68010, 68451 MMU)

jaggies:
I'm seeing this issue implementing the [Integrated Solutions 68kV20/25](http://www.bitsavers.org/pdf/isi/brochures/Optimum_V_Dec86_Ann.jpg) BSD system.

Among other things, it appears that page faults taken during writes are dropped.

balr0g:
Note that the MC68010 and later members of the family perform instruction continuation or resumption, not restart — when a page fault occurs, the sub-instruction state is saved on the supervisor stack and after page fault handling, the instruction resumes where it left off.

This is not fully emulated in MAME or most emulators, and lack of correct emulation may have effects for certain operating systems. It it briefly covered on page 1-6 of http://www.bitsavers.org/components/motorola/68000/MC68010_68012_Data_Sheet_May85.pdf.

agentbooth:
[...]

PR #15175, opened 2026-03-31
m68000: handle each fault within an instruction as it occurs

pmackinlay:
This change makes Musashi-based 68k CPUs with external MMUs ignore subsequent bus errors triggered within a single instruction. The change only affects systems which are using `rerun==true` (i.e., systems with external MMUs), which at time of writing in master are `sgi/ip2.cpp` and `sony/news_68k_iop.cpp`. The former is unaffected, while the latter has its own logic to avoid triggering such faults.

The idea is that the first bus error (fault) should halt the instruction and be handled before the instruction is resumed (restarted, in the current Musashi implementation) and any subsequent bus errors triggered. Without this change, Sun-2 code gets confused and fails to handle the correct fault.

See https://github.com/mamedev/mame/issues/9170 for some history and further details.


galibert:
I don't think that's external MMU only.  I think the PMMU follows the same paths.  OTOH, I also think it's correct.

PR #15159, opened 2026-03-27
gridcomp/gridrom.cpp: add device for GRiD's ROMs, add Test ROM

vklachkov:
This PR adds `gridrom_socket_device`/`GRIDROM_SOCKET` for connecting ROMs to GRiD laptops. Unlike `generic_socket_device`, `gridrom_socket_device` allows to change the media name and set strict size limits for the ROM.

To test functionality, I connected a Test ROM to the Compass at `C000:0`-`CFFF:F`. And I added application ROM slots for the Compass II without connecting them to the laptop.

<details>
  <summary>Information about Test ROM for Compass</summary>
We don't know anything about them, but apparently they were 64KB ROMs. The first four bytes of the ROM should be `ETPH`, followed by the code. I don't know who will find this feature useful, but it was a two-line change that made testing GRIDROM_SOCKET convenient.
</details>

List of devices for Compass 1100-1109:

```
SYSTEM           MEDIA NAME       (brief)    IMAGE FILE EXTENSIONS SUPPORTED
---------------- --------------------------- -------------------------------
grid1101         bubble           (mbm)      .bubble
grid1101         test-rom         (test-rom) .bin
```

List of devices for Compass II 1121-1139:

```
SYSTEM           MEDIA NAME       (brief)    IMAGE FILE EXTENSIONS SUPPORTED
---------------- --------------------------- -------------------------------
grid1139         bubble           (mbm)      .bubble
grid1139         test-rom         (test-rom) .bin  
grid1139         app-rom1         (app-rom1) .bin  
grid1139         app-rom2         (app-rom2) .bin  
grid1139         app-rom3         (app-rom3) .bin  
grid1139         app-rom4         (app-rom4) .bin  
```

PR #15193, opened 2026-04-04
gridcomp/gridcomp.cpp: implement application ROMs

vklachkov:
Continuation of PR https://github.com/mamedev/mame/pull/15159.

Adds Application ROM support to the GRiD Compass II. Implemented based on the CCPROM code. In the code, I tried to explain the logic of this mechanism in detail.

What remains unresolved is a TODO for a flag when installing 32KiB ROMs. I don't know exactly what it is for. But everything seems to work, here are the screenshot:

Screenshot with two ROM's

Works on both 1129 and 1139.

### DIFF
--- a/src/mame/gridcomp/gridcomp.cpp
+++ b/src/mame/gridcomp/gridcomp.cpp
@@ -98,7 +98,10 @@
 namespace {
 
 #define I80130_TAG      "osp"
-#define NUMBER_OF_ROMS  4
+
+constexpr int APP_ROM_COUNT = 4;
+constexpr int APP_ROM_WINDOWS = 4;
+constexpr int APP_ROM_WINDOW_SIZE = 0x8000;
 
 class gridcomp_state : public driver_device
 {
@@ -116,6 +119,12 @@ public:
 		, m_keyboard(*this, "keyboard")
 		, m_test_rom(*this, "test_rom")
 		, m_app_roms(*this, "app_rom%u", 0U)
+		, m_app_rom_view{
+			{*this, "app_rom_view0"},
+			{*this, "app_rom_view1"},
+			{*this, "app_rom_view2"},
+			{*this, "app_rom_view3"} }
+		, m_app_rom_bank(*this, "app_rom_bank%u", 0U)
 	{ }
 
 	static constexpr feature_type unemulated_features() { return feature::WAN; }
@@ -129,6 +138,7 @@ public:
 
 protected:
 	virtual void machine_start() override ATTR_COLD;
+	virtual void machine_reset() override ATTR_COLD;
 
 private:
 	required_device<i8086_cpu_device> m_maincpu;
@@ -141,12 +151,14 @@ private:
 	required_device<tms9914_device> m_tms9914;
 	required_device<grid_keyboard_device> m_keyboard;
 	required_device<gridrom_socket_device> m_test_rom;
-	optional_device_array<gridrom_socket_device, 4> m_app_roms;
+	optional_device_array<gridrom_socket_device, APP_ROM_COUNT> m_app_roms;
+
+	memory_view m_app_rom_view[APP_ROM_WINDOWS];
+	optional_memory_bank_array<APP_ROM_WINDOWS> m_app_rom_bank;
+	uint8_t m_app_rom_window[APP_ROM_WINDOWS] = {};
+	bool m_app_rom_enabled = false;
 
 	uint16_t *m_videoram = nullptr;
-
-	bool m_roms_enabled = false;
-	std::array<size_t, NUMBER_OF_ROMS> m_active_slots = {};
 
 	IRQ_CALLBACK_MEMBER(irq_callback);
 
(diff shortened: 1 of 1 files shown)


src/mame/gridcomp/gridcomp.cpp
```diff
@@ -99,6 +99,7 @@
 namespace {
 
 #define I80130_TAG      "osp"
+#define NUMBER_OF_ROMS  4
```
galibert: I think a constexpr within gridcomp_state would be more C++-y.

src/mame/gridcomp/gridcomp.cpp
```diff
@@ -282,6 +272,54 @@ uint8_t gridcomp_state::grid_dma_r(offs_t offset)
...
+	return rom->read_rom(offset % rom->get_rom_size());
+}
+
+void gridcomp_state::app_rom_activator_w(offs_t offset, uint8_t data)
+{
+	// LOG("ROM ACTIVATOR %02x <- %02x\n", offset, data);
+
+	// TODO: Handle 32KB flag at offset 2.
+
+	if (offset == 8)
```
galibert: Why didn't you put a specific handler at dfea8?

src/mame/gridcomp/gridcomp.cpp
```diff
@@ -282,6 +272,54 @@ uint8_t gridcomp_state::grid_dma_r(offs_t offset)
...
+	// TODO: How does the real laptop handle this situation?
+	if (data >= NUMBER_OF_ROMS)
+		return;
+
+	// 4 slot selection registers are located at addresses:
+	// DFE0:8, DFE0:A, DFE0:C, and DFE0:E.
+	m_active_slots[offset / 2] = data;
+}
+
+uint8_t gridcomp_state::app_rom_r(offs_t offset)
```
galibert: Is app_rom_r called often when running?  Because there's a way to make it significantly more efficient (using banks and views) but it's a little more complicated.  So does the system run from there or just cupy stuff to ram?

src/mame/gridcomp/gridcomp.cpp
```diff
@@ -282,6 +272,54 @@ uint8_t gridcomp_state::grid_dma_r(offs_t offset)
...
+	// LOG("ROM SLOT MAPPER %02x <- %02x\n", offset, data);
+
+	// The Compass only has 4 slots for ROMs.
+	// TODO: How does the real laptop handle this situation?
+	if (data >= NUMBER_OF_ROMS)
+		return;
+
+	// 4 slot selection registers are located at addresses:
+	// DFE0:8, DFE0:A, DFE0:C, and DFE0:E.
+	m_active_slots[offset / 2] = data;
```
galibert: What happens when writing at odd addresses?  Is A0 ignored (what you implemented) or is the write ignored?

src/mame/gridcomp/gridcomp.cpp
vklachkov: It seems that all of this is implemented in a single PAL. That's why I decided to combine both the activator and the 32-kilobyte flag into a single function.

src/mame/gridcomp/gridcomp.cpp
vklachkov: This function is called very often. Depending on the ROM, the laptop may either execute the code directly or copy the contents into memory. But judging by my logs, this is a pretty "hot" function.

I don't like this implementation, it smells of some kind of reinventing the wheel. But I didn't have the energy to figure out how to do it more cleanly, so I decided to make the PR as is. If you think such code has no place in MAME -- please point me to some machines I could look at for inspiration, and I'll be ready to redo `app_rom_r`...

src/mame/gridcomp/gridcomp.cpp
vklachkov: I don't know. There's no mention of this in the CCPROM code, and not a single program I know of writes to odd addresses. I decided to ignore writes.

rb6502: I would suggest a logerror() or similar if it happens, so that you know if it ever comes up.

src/mame/gridcomp/gridcomp.cpp
```diff
@@ -282,6 +272,54 @@ uint8_t gridcomp_state::grid_dma_r(offs_t offset)
...
+
+	// And look up which ROM slot is mapped to this window.
+	const size_t slot = m_active_slots[quarter];
+	const auto& rom = m_app_roms[slot];
+
+	// It does not matter what is returned if the ROM is not present because
+	// CCPROM checks the 0xBB66 marker at offset 9FF0:0,
+	// and then also verifies the checksum and other things.
+	if (!rom->exists())
+		return 0xFF;
```
cuavas, line 308: Kind of trivial compared to fixing the performance issues, but please use lowercase hex digits in MAME code.

vklachkov:
@galibert thank you for the review of the first implementation of the app ROM. I've reworked it to use view and banks. When you have time, could you take a look at the implementation and advise on what could be improved? If there are no comments, please merge this PR.

The only thing is that the new implementation has a quirk: when checking for the presence of the ROM, 'unmapped program memory read' logs appear if the ROM is not inserted. I couldn't suppress these logs.

vklachkov:
Rebased after PR https://github.com/mamedev/mame/pull/15750.

vklachkov:
Just a gentle reminder about this PR and the GRiD Compass emulator after two months. When you have some time, please take a look and merge it...

PR #15207, opened 2026-04-08
nes_vt369_vtunknown attempting to understand the extra port / protection on gtct885 a bit better

mamehaze:
- nes_vt369_vtunknown attempt to understand the extra port / protection on gtct885 a bit better
- hook up a protection device that allows gtct885 and rd5_240 to boot
- hook up the device to goretro and rbbrite, but they don't work as we lack the code the device supplies
- hook lexibook banking up to the port data write, not the direction write
- added encryption disable for rsps300 type

new NOT WORKING clones
--------------
Vibes Retro Pocket Gamer 240-in-1 (set 2) [Team Europe]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -36206,6 +36206,7 @@ unk128vt
 unk198vt
 unkra200
 vibes240
+vibes240a
 zl383
 zonefusn
 
--- a/src/mame/nintendo/nes_vt369_vtunknown.cpp
+++ b/src/mame/nintendo/nes_vt369_vtunknown.cpp
@@ -73,6 +73,7 @@ protected:
 
 	void extbank_w(u8 data);
 	void extbank_red5mam_w(u8 data);
+	void extbank_h12p1000_w(u8 data);
 
 private:
 	/* Extra IO */
@@ -90,6 +91,7 @@ public:
 	void vt_external_space_map_32mbyte(address_map &map) ATTR_COLD;
 	void vt_external_space_map_32mbyte_bank(address_map &map) ATTR_COLD;
 	void vt_external_space_map_16mbyte(address_map &map) ATTR_COLD;
+	void vt_external_space_map_16mbyte_bank(address_map &map) ATTR_COLD;
 	void vt_external_space_map_8mbyte(address_map &map) ATTR_COLD;
 	void vt_external_space_map_4mbyte(address_map &map) ATTR_COLD;
 	void vt_external_space_map_2mbyte(address_map &map) ATTR_COLD;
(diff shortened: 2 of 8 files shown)


mamehaze:
gtct885 and quite a few others, such as goretrop, lxcap, rd5_240 and more have some kind of extra chip attached to I/O ports.

the games read 0x100 bytes from it, so I thought it might be a 93C56 (it's a mystery 8-pin chip with surface details removed) but it doesn't seem to respond to commands I'd expect for one of those (although the port hookup is incomplete)

they set a direction register of 0x38 when they write to the device, and 0x18 when they read from it, with the data being read on 0x20.  this use of the direction register suggests 3 write bits at least, rather than a 2 wire solution. (hence me wondering if it was a 93C56)

for gtct885 the data was extracted previously from a running machine, and is in a ROM region, so we know what it wants (if you copy the data to RAM manually after the routine it boots just fine) so it's just a case of hooking up the comms protocol correctly.  for other sets like goretrop it needs extracting/dumping, as they copy code from the device into RAM to execute.  If it's a standard device of some kind, not a custom 8-pin MCU then it would help to know what it is so the chip can be dumped directly.

oddly for goretrop this is done on 4138 / 4139, which we currently treat as an ALU mirror (but that might be incorrect looking again at the sets that access it)

mamehaze:
@ajrhacker I'd be interested in your thoughts on this

The function of these things seems too simple for them to have bothered making something custom, so I feel it must be some common 8-pin chip they've stored 0x100 bytes of data in, but I haven't managed to match it up to anything.

ajrhacker:
Unfortunately I don't know what type of EEPROM this could be, if it is indeed a standard one (which I rather doubt). The EEPROM type on Tattoo Assassins is similarly a mystery (though the interface there looks much more SPI-like), and it's most certainly not the 93C76 which the MAME driver configures.

There's also quite a few differences between the sets that do this. goretrop uses $4138-4139 instead of $4152-4153 (and different bits), so that's probably another port. lxcap actually seems to use I2C, though the slave address of $80 (shifted) is associated with audio devices, not EEPROMs.

mamehaze:
h12p1000 uses 4138/4139 as a port, so yeah the ALU mirror speculation was probably wrong.

mamehaze:
also yeah, this alt vibes240 set surprised me.  When Klaus dumped it, he only got this consistent read 3 times, then bad dumps again (so it could be good, with the ROM failing after the first few reads)

I just wrote it off at the time as bad however as the code looked corrupt, but infact it's using different encryption to the set we had.

mamehaze:
well I've added enough for gtct885 to read from the extra region and boot

mamehaze:
looks like there are different device types anyway based on the comms

lxcap and some others appear to use a 2 line protocol unless I'm mistaken, as the direction registers only use 2 bits, and what they write seem to go beyond just the commands (so more like SPI?)

gtct885 etc. which I hooked up appear to use 3 lines.

PR #15220, opened 2026-04-12
pixtermu: Make the system boot and LCD/touchscreen work

gatecat:
This doesn't get it that much further along - the next hang is waiting on the SSP - but as I haven't contributed to MAME for along time, I thought I'd submit this so I can hopefully get some feedback on my code so far.

### DIFF
--- a/scripts/src/machine.lua
+++ b/scripts/src/machine.lua
@@ -2180,6 +2180,18 @@ end
 
 ---------------------------------------------------
 --
+--@src/devices/machine/lh79524_timer.h,MACHINES["LH79524"] = true
+---------------------------------------------------
+
+if MACHINES["LH79524"] then
+	files {
+		MAME_DIR .. "src/devices/machine/lh79524_timer.cpp",
+		MAME_DIR .. "src/devices/machine/lh79524_timer.h",
+	}
+end
+
+---------------------------------------------------
+--
 --@src/devices/machine/locomo.h,MACHINES["LOCOMO"] = true
 ---------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


gatecat:
On my [WIP branch](https://github.com/gatecat/mame/tree/pixtermu_work) I have a bunch more progress (including a bunch of hacks along the way), getting to the point of being able to somewhat use the touchscreen and run content

https://github.com/user-attachments/assets/f29b5d21-0d0e-4b40-a704-b3cb76bc35bb




Will continue working on tidying this up...

rb6502:
This looks quite decent to me for the most part.  The one thing that really stood out is that there are multiple bracing styles going on.  Most stuff is like this:

```
void blah() {
    if (foo) {
    }
}
```

but there's also a fair amount of this:
```
void blah()
{
}
```

MAME doesn't mandate any particular style, but we do want it to be consistent inside one file.  Since the LH79524 is a new file you can choose whichever style you're most  comfortable with.  Just keep it all in that one style :-)  If you can clean that up I think we can apply this and get you on the way towards getting the system running.

gatecat:
Thanks! The style is hopefully more consistent now and I've pushed the changes that get the LCD/touchscreen working with some debug hacks removed now into this PR too. The buttons aren't working yet but I plan to figure out the button/GPIO mapping soon.

There are still two hacks that I'm not 100% happy about:

 - I'm triggering the SSP DMA interrupt (that is actually used for the I2S audio DAC) on Vblank just to force forward progress. I've seen similar hacks in other tvgames type systems in the past so hopefully this is OK for now, I would like to work on audio at some point
 - The system physically has a printed label below the LCD that the touchscreen extends to with 9 soft buttons mostly covering drawing-related controls. For now I just pretend the screen is actually 160x176 instead of 160x160 and draw some squares below the visible area in `screen_update` so they can be interacted with. But maybe there's a more idiomatic approach here instead that I'm missing.

PR #15224, opened 2026-04-13
netlist: Check for mismatched number of parameters passed to device macros at compile-time 

FlykeSpice:
According to ``nl_setup::register_dev`` rules, the device macro must either be passed with the exact number of params supported by the device or no params at all. 

This PR adds some compile-time checking for that, so that when the developer forgets to fully specify all the parameters for a netlist device, they can get a compile-time error rather than a runtime exception or ``validity`` fail.

* Also this removes ``NET_REGISTER_DEVEXT`` macro, not needed anymore after the introduction of C++20 ``__VA_OPT__`` that allows more flexible "overloading" with its original ``NET_REGISTER_DEV`` counterpart
* * Consequently this also removes the ``nl_setup::register_dev`` overload that was specifically made for that macro.

### DIFF
--- a/src/lib/netlist/build/create_devinc.py
+++ b/src/lib/netlist/build/create_devinc.py
@@ -38,25 +38,21 @@ def process_entry(srcfile, name, params):
     pusage = ""
     pauto = ""
     pcount = 0
-    immediate_params = True
     for x in params.split(","):
         if x.startswith('@'):
             pauto = pauto + ", " + x[1:]
         elif len(x) > 0:
             if x.startswith('+'):
-                immediate_params = False
                 x = x[1:]
-                if x[0].isdigit():
-                    x = '_'+x
             pusage = pusage + ", " + x
             pcount += 1
     print("// usage       : {}(name{})".format(name, pusage))
     if len(pauto) > 0:
         print("// auto connect: {}".format(pauto[2:]))
-    print("#define {}(name{})                                                   \\".format(name, pusage if immediate_params else ", ..."))
-    if pcount > 0 and not immediate_params:
+    print("#define {}(name{})                                                   \\".format(name, ", ..." if pcount > 0 else ""))
+    if pcount > 0:
         print("\t__VA_OPT__(static_assert(PNARGS(__VA_ARGS__) == {}, \"{}: Mismatched number of parameters passed, expected {} but got \" PSTRINGIFY(PNARGS(__VA_ARGS__)));) \\".format(pcount, name, pcount))
-    print("\tNET_REGISTER_DEV({}, name{})".format(name, pusage if immediate_params else " __VA_OPT__(,) __VA_ARGS__"))
+    print("\tNET_REGISTER_DEV({}, name{})".format(name, " __VA_OPT__(,) __VA_ARGS__" if pcount > 0 else ""))
     print("")
 
 
--- a/src/lib/netlist/build/create_devinc.py
+++ b/src/lib/netlist/build/create_devinc.py
@@ -18,18 +18,6 @@ if sys.version_info > (3, ):
 
 # globals
 
-linewidth = 100
-tabwidth = 4
-def mac_out(text):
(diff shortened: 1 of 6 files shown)


src/lib/netlist/generated/nld_devinc.h
```diff
@@ -9,800 +9,892 @@
 
 #include "../nl_setup.h"
 
+#define CHECK_PARAM_COUNT(dev, nargs, N) \
+	static_assert(nargs == N, #dev": Mismatched number of parameters passed, expected " #N " but got " #nargs);
```
cuavas: The name of this is a bit too generic IMO.  Maybe it should be a PLIB macro with a `P` prefix (defined in the same kind of place as `PNARGS`)?  Maybe it should have a `NET_` prefix if it’s part of the netlist?

Also, won't `#nargs` will expand to the literal `PNARGS(...)` here?

src/lib/netlist/build/create_devinc.py
```diff
@@ -18,6 +18,18 @@
...
+tabwidth = 4
+def mac_out(text):
+    global linewidth
+    temp = text.replace('\t'," "*tabwidth)
+    padding = linewidth-len(temp)
+    if padding < 0:
+        linewidth = len(temp)+5
+        padding = 5
+
+    print(text + " "*padding + "\\")
```
cuavas: The tab expansion won’t work for tabs that aren’t at the beginning of the line.

Also, this leaks the line width forward – after the first line that exceeds 100 characters, it will pad to that width for all future lines.  This is going to get ugly.  If you can’t restrict the length extension to the current macro, I’d just make an exception for one line and give it a single space padding.  (I’m not convinced lining up all the line continuation backslashes is a worthwhile endeavour in the first place.  It’s probably just as readable if you put a single space before the backslash.)

src/lib/netlist/nl_setup.h
```diff
@@ -184,8 +181,6 @@ namespace netlist
 		void register_dev(const pstring &classname, const pstring &name,
 						  const std::vector<pstring> &params_and_connections,
 						  factory::element_t **factory_element = nullptr);
-		void register_dev(const pstring                      &classname,
-						  std::initializer_list<const char *> more_parameters);
 		void register_dev(const pstring &classname, const pstring &name)
```
cuavas, line 184: Why did you remove this overload?

src/lib/netlist/generated/nld_devinc.h
FlykeSpice: Added the ``NET_`` prefix to the macro, though it won't help much on netlist library case since it exposes a bunch of generic named macros like ``PARAM`` and ``ALIAS`` for the purposes of netlist configuration..

> Also, won't #nargs will expand to the literal PNARGS(...) here?

Good catch, fixed that too.

src/lib/netlist/build/create_devinc.py
FlykeSpice: I reverted to just adding a single space before the backlash as you suggested, it's not worth struggling with the backlash alignment indeed. It's just that the netlist library does their macro that way.

src/lib/netlist/nl_setup.h
FlykeSpice: I just stated the reason in the PR desc...

> Also this removes NET_REGISTER_DEVEXT macro, not needed anymore after the introduction of C++20 __VA_OPT__ that allows more flexible "overloading" with its original NET_REGISTER_DEV counterpart

> **Consequently this also removes the nl_setup::register_dev overload that was specifically made for that macro.**

the overload was specifically made as a "hack" to work with the old ``NET_REGISTER_DEVEXT`` macro when the device macro ``name`` argument was passed as part of __VA_ARGS__

cuavas: Ah, right.  It special-cases the first element of `more_parameters` and then calls the other overload.

PR #15225, opened 2026-04-14
bus/centronics/pc6022.cpp: notes about print mechanism

devinacker:
Just a bit of extra documentation after studying what the timer interrupt does.

Might be somewhat hookup-able to `bitmap_printer_device`, but I don't really feel like trying it at the moment.

### DIFF
--- a/src/devices/bus/centronics/pc6022.cpp
+++ b/src/devices/bus/centronics/pc6022.cpp
@@ -2,10 +2,15 @@
 // copyright-holders:Devin Acker
 /**********************************************************************
 
-Skeleton for NEC PC-6022 Color Plotter Printer
+	Skeleton for NEC PC-6022 Color Plotter Printer
 
-User manual:
-https://archive.org/details/nec-pc-6022-color-plotter-printer-users-manual
+	User manual:
+	https://archive.org/details/nec-pc-6022-color-plotter-printer-users-manual
+
+	Page width is 96mm / 480px.
+	CPU port A controls the two steppers, and port C controls raising/lowering the pen.
+	Pen/color selection happens by moving the pen to the left of the print area, then
+	moving it back and forth 3 times per color change.
 
 **********************************************************************/
 
@@ -30,7 +35,11 @@ const tiny_rom_entry *pc6022_device::device_rom_region() const
 
 
 INPUT_PORTS_START( pc6022 )
-	PORT_START("PB") // TODO: verify button order / switch polarity
+	PORT_START("PA")
+	PORT_BIT( 0x0f, IP_ACTIVE_HIGH, IPT_OUTPUT ) // X stepper (pattern is 5, 6, A, 9 for positive/rightward movement)
+	PORT_BIT( 0xf0, IP_ACTIVE_HIGH, IPT_OUTPUT ) // Y stepper (page feed)
+
+	PORT_START("PB")
 	PORT_BIT( 0x01, IP_ACTIVE_LOW, IPT_KEYPAD ) PORT_NAME("Pen Right");
 	PORT_BIT( 0x02, IP_ACTIVE_LOW, IPT_KEYPAD ) PORT_NAME("Pen Left");
 	PORT_BIT( 0x04, IP_ACTIVE_LOW, IPT_KEYPAD ) PORT_NAME("Pen Down / Paper Feed");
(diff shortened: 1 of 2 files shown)


goldnchild:
Hi Devin,

I took your code and hooked up the bitmap_printer_device and have got the cr moving around (little bit backwards at the moment, trying to figure out where it reads the home position), looks like it holds the cr position in FF88/FF89.

After it settles down from initial carriage movement, the left/right/up/down buttons can move the printhead around on the page.
Also the pen change seems to put the cr at the right side so you can change the pen, then hitting pen change again (have to hold it a second) will return the carriage to where it was originally.  The color change seems to work as well.

Holding down down arrow on startup seems to do a self test print.

https://github.com/goldnchild/mame/tree/pc6022_bitmap_printer

(I hooked up the cr so that horizontal motion is reversed, after the initial cr motion stops you can use the buttons left and right and up/down (use right to bring the printhead into view))

Movement seems to be limited to 480 steps at which point it won't move the printhead further.

pm me on bannister if you like.

devinacker:
The other movement keys also do other test prints in the other colors, and you should be able to hold multiple of them to do multiple test prints. Should be an easy way to test switching colors.

PC7 seems to be a pen alignment sensor (checked at 0B18 after the motors stop moving); bringing it low stops any further rotation of the pen holder when changing colors. None of the other port inputs seem to be used except for the ones already hooked up in the driver. When it powers on it seems to just move the carriage as far to the left as possible, rotate the pen holder a few times, and then considers itself to be in the home position.

goldnchild:
Thanks for the info.  I  accidentally mangled up the pen write control input port and it wasn't passing the pen changes along, which made it not draw anything until I restored it to your original bits.

I was able to get it to do this:


[Imgur](https://imgur.com/TsO9PHn)

going to update my branch'


edit: raised resolution by 2x
[Imgur](https://imgur.com/e82a5Av)

goldnchild:
Do you have any idea how the pc6022 is supposed to input data from the centronics?

I was trying to hook up an output for PC0 which I think is supposed to do the STOP led.  Setting breakpoints on 814 and 9d6 should cause bit 0 to get written on port C.  

Currently, writing bit 0 of port C has no effect.  I fiddled with it for awhile and came to this conclusion.

However, in the 1981_NEC_Microcomputer_Catalog.pdf on page 256 of 717 it seems that if mode_c has bit 0 clear, then PC0 is supposed to be an output.

Using the debugger, mode c is set to $FE so it should be output.  However, the current code ignores this particular bit 0.

In  src/devices/cpu/upd7810/upd7810.cpp
void upd7801_device::write_pc(uint8_t data)
{
        data = (data & 0x78) | (m_pc_pullups & ~0x78);
        if (!BIT(m_mc, 3))  /* TODO PC3 = SAK output */
                data = (data & ~0x08);
        if (!BIT(m_mc, 4))  /* PC4 = TO output */
                data = (data & ~0x10) | (m_to & 1 ? 0x10 : 0x00);
        if (!BIT(m_mc, 5))  /* TODO PC5 = IO/-M input */
                data = (data & ~0x20);
        if (!BIT(m_mc, 6))  /* TODO PC6 = HLDA output */
                data = (data & ~0x40);
        m_pc_out_cb(data);
}

devinacker:
Good catch, that's fixed in https://github.com/mamedev/mame/pull/15232.

On that note, so `m_pa` / `pa_w` in your branch aren't needed, since port A is output-only on the 7801 (reading from it always returns the last output).

> Do you have any idea how the pc6022 is supposed to input data from the centronics?

The data/strobe lines are already hooked up, you just have to press the stop button to bring it online before it will actually print anything. The print buffer is at FFC1. (I was testing it using pc6001 and pc8001)

goldnchild:
The pa_w was to catch the writes to the steppers atomically for drawing diagonals.  You could probably skip the pa_r and m_pa if you like, as long as it works.

Oh yes, I see how the centronics hooked up now, reading m_data in data_r().   I must've looked at that over and over and didn't see it!  Awesome!  I just tried it with the pc6001 and it seems to work!


For the upd7801, I tried this:

uint8_t upd7801_device::read_pc()
{
	if ((m_mc & 0x87) && !m_pc_in_cb.isunset())  // NS20031301 no need to read if the port is set as output
		m_pc_in = m_pc_in_cb(0, 0x84 | (m_mc & 0x03));
	u8 read_mask = 0x84 | (m_mc & 0x03);
    uint8_t data = (m_pc_in & read_mask) | (m_pc_out & ~read_mask);
	//uint8_t data = (m_pc_in & 0x87) | (m_pc_out & ~0x87);
	if (!BIT(m_mc, 2))  /* TODO PC2 = -SCS input */
		data = (data & ~0x04) | 0x04;
	if (!BIT(m_mc, 7))  /* TODO PC7 = HOLD input */
		data = (data & ~0x80);
	return data;
}

void upd7801_device::write_pc(uint8_t data)
{
	u8 output_mask = 0x78 | (BIT(m_mc,0) ? 0 : 0x1) | (BIT(m_mc,1) ? 0 : 0x2);
    data = (data & output_mask) | (m_pc_pullups & ~output_mask);
	//data = (data & 0x78) | (m_pc_pullups & ~0x78);
	if (!BIT(m_mc, 3))  /* TODO PC3 = SAK output */
		data = (data & ~0x08);
	if (!BIT(m_mc, 4))  /* PC4 = TO output */
		data = (data & ~0x10) | (m_to & 1 ? 0x10 : 0x00);
	if (!BIT(m_mc, 5))  /* TODO PC5 = IO/-M input */
		data = (data & ~0x20);
	if (!BIT(m_mc, 6))  /* TODO PC6 = HLDA output */
		data = (data & ~0x40);
	m_pc_out_cb(data);
	m_pc_out = data;
}

goldnchild:
ok, I've added a pen position sensor and fixed the threshold value to advance the pen position.

I also added some changes to the pc6001 so I could paste in a basic program to test the pc6022.

However, you may need to increase the natural keyboard delay or the pc6001 will drop characters (I just doubled the regular values)

natural_keyboard::choose_delay(char32_t ch)
 	 	
  changed to: 	return attotime::from_msec((ch == '\r') ? 400 : 100);
 	return attotime::from_msec((ch == '\r') ? 200 : 50);
 }
 
 
emu.keypost('LPRINT CHR$(19)\nFOR I=1 TO 8:LPRINT "C"I:LPRINT "P"I:NEXT\n')

PR #15232, opened 2026-04-16
upd7801: fix port C direction bits

devinacker:
PC0 and PC1 are bidirectional; previously they only worked correctly when configured as inputs.

### DIFF
--- a/src/devices/cpu/upd7810/upd7810.cpp
+++ b/src/devices/cpu/upd7810/upd7810.cpp
@@ -719,9 +719,11 @@ uint8_t upd7810_device::read_pc()
 
 uint8_t upd7801_device::read_pc()
 {
+	const u8 inputs = 0x84 | (m_mc & 0x03);
+
 	if ((m_mc & 0x87) && !m_pc_in_cb.isunset())  // NS20031301 no need to read if the port is set as output
-		m_pc_in = m_pc_in_cb(0, 0x84 | (m_mc & 0x03));
-	uint8_t data = (m_pc_in & 0x87) | (m_pc_out & ~0x87);
+		m_pc_in = m_pc_in_cb(0, inputs);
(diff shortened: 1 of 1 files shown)


src/devices/cpu/upd7810/upd7810.cpp
```diff
@@ -719,9 +719,11 @@ uint8_t upd7810_device::read_pc()
 
 uint8_t upd7801_device::read_pc()
 {
+	const u8 inputs = 0x84 | (m_mc & 0x03);
+
 	if ((m_mc & 0x87) && !m_pc_in_cb.isunset())  // NS20031301 no need to read if the port is set as output
-		m_pc_in = m_pc_in_cb(0, 0x84 | (m_mc & 0x03));
-	uint8_t data = (m_pc_in & 0x87) | (m_pc_out & ~0x87);
+		m_pc_in = m_pc_in_cb(0, inputs);
+	uint8_t data = (m_pc_in & inputs) | (m_pc_out & ~inputs);
```
cuavas, line 726: Hang on, this doesn’t quite make sense.  It’s only treating bits 0 and 1 of `m_mc` as direction bits here, treating bits 2 and 7 as always being inputs:
```c++
	const u8 inputs = 0x84 | (m_mc & 0x03);
```

But when it’s checking whether it needs to read inputs, it’s treating bits 2 and 7 of `m_mc` as direction bits:
```c++
	if ((m_mc & 0x87) && !m_pc_in_cb.isunset())  // NS20031301 no need to read if the port is set as output
```

If bits 2 and 7 are always inputs, shouldn’t it always read `m_pc_in_cb` if it’s set?  The code as-is looks like it isn’t right.

devinacker: They are always inputs, but not always port inputs (bit 2 can be a serial chip select, and 6-7 can be Z80-style HOLD/HLDA):
image

PR #15233, opened 2026-04-16
igs/igs011_video.cpp, igs/igs012.cpp: Make IGS011 and IGS012 as device, Update related drivers

cam900:
igs/igs011.cpp:
- Move IGS011 (including decryption, protection, and video), IGS012 into separated device
- Fix relocate protection area function
- Fix naming for per-game protection functions
- Fix logging
- Fix post load function
- Add notes for dbc title screen transparency
- Fix typename values

igs/pgmprot*.cpp:
- Add include guard
- Fix spacings
- Fix save state support

igs/pgmprot_igs025_igs012.cpp:
- Add notes for IGS012, Add IGS012 placeholder

igs/igs025.cpp:
- Fix spacing, Suppress side effect for debugger reads


src/mame/igs/igs011_video.cpp
```diff
@@ -0,0 +1,538 @@
+// license:BSD-3-Clause
+// copyright-holders:Luca Elia, Olivier Galibert
+/*
+
+IGS011 blitter with protection
+
+implementation based from igs/igs011.cpp, by Luca Elia/Olivier Galibert.
+
+TODO:
+- transparent pixel is wrong in dbc title screen
```
cuavas, line 10: There’s also a video output disable flag that currently isn’t emulated (causes the video output to be black, and I suspect it also allows the blitter to work faster because it won’t have to compete with the video output for access to layer RAM).

Blitter timings and busy flag are currently unemulated, too – we just pretend it works instantaneously, as none of the games depend on realistic performance.  I suspect the blitter can only access layer RAM during horizontal and vertical blanking, or while video output is disabled.

cuavas:
I’m not sure all the graphics ROM decoding belongs in the IGS011 device.  At least some of it is probably per-game logic.  It’s more manageable like this anyway.

cuavas:
The bad fade on the title screen for `dbc` is an original game bug.  It writes incorrect palette data.  See [MT02497](https://mametesters.org/view.php?id=2497).

I removed the imperfect graphics flag in 46a890d2854838ce7e6c2bf177c54904b62720a8 as the only graphical issues I’m aware of in that game are original game bugs.

cuavas:
Also see [MT02151](https://mametesters.org/view.php?id=2151) which has a photo taken with an original game board.

happppp:
For bugs like those (where emulator developers will likely suspect it's an emulation bug), it's good to have a BTANB list so noone in the future will have to chase ghost bugs.

cuavas:
> For bugs like those (where emulator developers will likely suspect it's an emulation bug), it's good to have a BTANB list so noone in the future will have to chase ghost bugs.

It was already mentioned at the top of igs/igs011.cpp:
```c++
- dbc: The title screen background palette is wrong since the fade
  routine is called with incorrect argument values, but the PCB does the
  same thing.
```

@cam900 must have missed it.

cuavas:
Actually, I’ll add the drgnwrld BTANBs to the list.

PR #15234, opened 2026-04-16
firebeat.cpp: Add lamp output support for Pop'n Music

nutki:
Verified the bit offsets with popn8 and then tested other popn firebeat games.

### DIFF
--- a/src/mame/konami/firebeat.cpp
+++ b/src/mame/konami/firebeat.cpp
@@ -425,8 +425,6 @@ protected:
 
 	void firebeat_map(address_map &map) ATTR_COLD;
 	void ymz280b_map(address_map &map) ATTR_COLD;
-
-	void init_lights(write32s_delegate out1, write32s_delegate out2, write32s_delegate out3);
 	void lamp_output_w(offs_t offset, uint32_t data, uint32_t mem_mask = ~0);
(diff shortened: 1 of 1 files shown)


cuavas:
The whole `init_lights` functions is atrocious.  Can you adapt the driver to just install game-specific lamp output functions in address map functions?  It would make it a lot easier to understand and less likely to get screwed up.

nutki:
> The whole `init_lights` functions is atrocious. Can you adapt the driver to just install game-specific lamp output functions in address map functions? It would make it a lot easier to understand and less likely to get screwed up.

Sure, I can give it a try. I am not familiar with the mame code base so I just followed the pattern used in this file.

nutki:
@cuavas does this refactor make sense? It seems to work (though I only tested it on popn8 now).
https://github.com/mamedev/mame/pull/15234/changes/90bb824f18bc7c8e6cec0b2110380bb88e7dcba0

cuavas:
> @cuavas does this refactor make sense? It seems to work (though I only tested it on popn8 now). [90bb824](https://github.com/mamedev/mame/commit/90bb824f18bc7c8e6cec0b2110380bb88e7dcba0)

Yep, looks better than it was.

PR #15236, opened 2026-04-17
x07: toward a better audio support

Mokona:
The audio support for the Canon X07 was reading a memory location to get information on note length. That memory location is a by-product of the BEEP Basic instruction interpretation, making direct programming of audio not functional.

This change removes this memory "shortcut" and gets more close to what happens for real (ROM is counting interruptions).

BEEP still working. Audio direct programming working better (still need more work that needs to be figured out).

Also key press produces click, as it should + support of enabling/disabling the clicks.

### DIFF
--- a/src/mame/canon/x07.cpp
+++ b/src/mame/canon/x07.cpp
@@ -813,6 +813,34 @@ void x07_state::receive_bit(int bit)
 	}
 }
 
+void x07_state::t6834_set_audio()
+{
+	const uint16_t div = (m_regs_w[2] | m_regs_w[3] << 8) & 0x0fff;
+	if (div > 0)
+	{
+		const uint32_t freq = 192000 / div;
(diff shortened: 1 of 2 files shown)


src/mame/canon/x07.cpp
```diff
@@ -813,6 +830,34 @@ void x07_state::receive_bit(int bit)
 	}
 }
 
+void x07_state::t6834_set_audio()
+{
+	const uint16_t div = (m_regs_w[2] | m_regs_w[3] << 8) & 0x0fff;
+	if (div > 0)
+	{
+		const uint32_t freq = 192000 / div;
```
rb6502: 192000 is the 3.84 MHz CPU clock divided by exactly 20 (or the 15.36 MHz crystal divided by 80), so it'd be good to make that derivation explicit.

PR #15237, opened 2026-04-18
Refactor 28-series EEPROM code

cbrunschen:
Refactor 28-series EEPROM code:
- Separate out a generic `eeprom28_device` template
- Create Xicor x28 instantiations with appropriate template parameters.

Add functionality to the `eeprom28_device` template to implement Atmel-style
- Identification Page
- Hardware Chip Erase
- Software Chip Erase

Add a templated `eeprom28_nvram_device` subclass that adds an `device_nvram_interface` implementation.

Add `at28.{h,cpp}` specializations implementing known Atmel AT28 devices.

Both X28 and AT28 devices are made available both with and without nvram support.

This also replaces the existing `at28c64b` implementation. This has two current users in the codebase, both of which are Apple II cards. I have very little experience with that, sadly. But with many thanks to @rb6502 I was able to test this in the `booti` device, which helped me identify and fix something I had missed in the ee28 implementation: while the EEPROM is buffering data, reads from the page being buffered will be sourced from the buffer, so will include whatever is currently being buffered. I've gathered more detailed results of this testing and am happy to share those on request.

I'm using C++20 concepts to elide some member functions from the templates; and macros to simplify declaring device classes and types both with and without NVRAM. 

All of this works, and adds what I think is fairly complete support for a number of similar real-world devices in a way that I've aimed to document reasonably well and make maintainable. I've tried to keep this in line with MAME's guidelines and coding standards and what I've observed as existing practices, but I'm sure there are things I've missed or got wrong. So any feedback or guidance would be greatly appreciated.


src/devices/machine/ee28.ipp
```diff
@@ -38,7 +34,7 @@ void x28_device<AddressBits, PageSizeBytes, TBLCUsec, TWCUsec, ProgramOnRead>::w
 		m_start_programming_timer->adjust(attotime::from_usec(m_t_blc_usec));
 	}
 
-	// LOG("write(%04x, %02x) in state %d\n", offset, data, m_state);
+	// EE28LOG("write(%04x, %02x) in state %d\n", offset, data, m_state);
```
rb6502: For stuff like this, the MAME pattern has been to define an additional log bit that's something like BLAH_VERBOSE or to get the point across that it's useful sometimes but it will also flood your log.  That way it's not commented out so it can't silently decay.

rb6502:
I like this in general after giving It a good read.  Because of the C++20 feature usage I'd like to tag @galibert and/or @cuavas and see what they think though.

happppp:
Could you rename ee28.* to eeprom28.*?

src/devices/machine/ee28.ipp
cbrunschen: You're quite right, fixing.

cbrunschen:
> I like this in general after giving It a good read. Because of the C++20 feature usage I'd like to tag @galibert and/or @cuavas and see what they think though.

Absolutely, using new language features warrants an extra thorough review.

cbrunschen:
> Could you rename ee28.* to eeprom28.*?

Gladly!

cbrunschen:
To clarify: 

The C++20 feature this PR uses is specifically a few uses of the `requires` keyword to conditionally en- and disable member functions to allow access to the ID page and the chip erase functionality only when the corresponding template parameters are `true`; and similarly, to implement the correct member functions to translate the address to the offset in storage depending on the presence or absence of the ID page.

This PR also replaces the current `ar28c64b` implementation with one that is much more feature complete and works better, at least in my testing for the Apple II `booti` card.

I would greatly appreciate all feedback.

cbrunschen:
Thank you!

PR #15244, opened 2026-04-20
kaneko/kaneko_rle.cpp: Fix filename and device name, Update related drivers

cam900:
kaneko/kaneko_rle.cpp: (former kaneko/sknsspr.cpp)
- Used at jchan* and galpani3* before released Super Kaneko Nova System. (suprnova.cpp)
- Reduce duplicates
- Fix namings
- Fix code style consistency
- Fix spacings
- Make bitmap as member variable (also double buffered)
- Add notes

kaneko/galpani3.cpp:
- Fix naming

kaneko/jchan.cpp:
- Fix naming
- Fix save state support
- Fix logging in debugger reads

kaneko/suprnova.cpp:
- Fix save state support


src/mame/kaneko/kaneko_rle.h
```diff
@@ -0,0 +1,41 @@
...
+
+ // TODO : Unknown address bits; maybe 27?
+class kaneko_rle_device : public device_t, public device_video_interface, public device_rom_interface<27>
+{
+public:
+	kaneko_rle_device(const machine_config &mconfig, const char *tag, device_t *owner, u32 clock);
+
+	void draw_sprites(const rectangle &cliprect, u32* spriteram_source, size_t spriteram_size, u32* sprite_regs);
+	void set_sprite_kludge(int x, int y);
+
```
cuavas: If `set_sprite_kludge` is a configuration function, it should be visually separated from live functions.

src/mame/kaneko/kaneko_rle.h
```diff
@@ -0,0 +1,41 @@
...
+private:
+	static constexpr unsigned DECODE_BUFFER_SIZE = 0x2000;
+	static constexpr unsigned DECODE_BUFFER_MASK = DECODE_BUFFER_SIZE - 1;
+	static constexpr size_t ROM_ADDRESS_MASK = (1 << 27) - 1;
+
+	int m_sprite_kludge_x = 0, m_sprite_kludge_y = 0;
+	std::unique_ptr<u8[]> m_decodebuffer;
+	bitmap_ind16 m_sprite_bitmap[2];
+	u8 m_buffer;
```
cuavas: Please don’t put member initialisation in the header when the constructor implementation is in a different file.  It’s too easy to end up with contradictory initialisation in the two places.

src/mame/kaneko/kaneko_rlespr.h
cuavas, line 1: I don’t like this being called “Kaneko RLE” (`kaneko_rle_device`, `KANEKO_RLE`) with no reference to it being a sprite system.  That isn’t descriptive.

PR #15246, opened 2026-04-20
generalplus gpce4 refactoring

mamehaze:
- moved GeneralPlus GPCE4 series SoC emulation from driver file into a device [David Haywood]
- begin to flesh out GPCE4 emulation a little more [David Haywood]
- move mapacman away from make-believe hookups to something a little more grounded in reality now that the chip has been identified [David Haywood]
- added internal ROM for digicolr set, replace with clean dump [Harold Seo]
- removed local implementation of LCDC previously used by mapacman and replaced it with the bl_handhelds_lcdc.cpp device, which looks like it's actually a ST7735SV (which is what these use) [David Haywood]
- mapacman display is now more stable [David Haywood]


mamehaze:
(this is the first step in getting things into better shape, there's a lot still to do, but as we know what the chip is that will be easier.  this will be the groundwork for extracting more internal ROMs from the devices using this chip)

PR #15247, opened 2026-04-22
apple2video: improve more video timing

arekkusu42:
(needs review, see RFC list)

This PR fixes [additional](https://github.com/mamedev/mame/pull/15073) timing-sensitive video interactions and enables 50Hz support:
* apple2video: emulate softswitch-specific delays
* apple2video: improve `read_floatingbus()`
* apple2gs: add 50Hz support
<br>
Code comments:<details>

To implement the video softswitch delays, I have slightly changed the core's `screen().update_now()`.  The alternative is to use a timer callback, and indeed for the 6502 machines this can work.  However I believe that a simple timer implementation _can not_ be correct for apple2gs; consider what should happen with this 65816 instruction:
`	LDA $C054`
If m=0, this is a 16-bit load accessing two adjacent softswitches, and [the operation of each cycle](https://www.brutaldeluxe.fr/documentation/cortland/v7%2065SC816%20opcodes%20Cycle%20by%20cycle%20listing.pdf#page=31) is:
1. 	fetch opcode
2. 	fetch address lo
3. 	fetch address hi
4. 	data read lo C054 (toggling PAGE1, visible on-screen two cycles later)
5. 	data read hi C055 (toggling PAGE2, visible on-screen two cycles later)

On hardware (with appropriate data and timing loop) this results in a one-cycle wide blip of PAGE1:
Calibrate_16bit

A single timer can't implement this, because the second read would void and replace the first read's callback before it fires.  We could use two timers?  But what about more complex RMW instructions, like ASL, ROR, TRB, etc?  At that point, recognize that what we are really trying to accomplish is to flush pending video changes up to a time in the near (0...3 video cycles) future, and so modifying `update_now()` to allow that feels like the right approach.

Also for the softswitch delays, I have added the notion of an `m_base_model` to `a2_video_device`.  Previously, machine-specific video behavior has been either limited to the model-templated text rendering, or via [one-shot accessors](https://github.com/mamedev/mame/pull/14945/changes) like `set_hgr2`.  This PR uses `m_base_model` for machine-specific delays and floating bus behavior, but there are additional areas where it could be used to implement currently un-emulated quirks, which I feel don't fit well into templating:
- 	WNDW blanking at left edge of active video differs between ][+ and later
- 	80-column video placement differs between //e and IIgs
- 	lo-res 7M mode differs between //e and IIgs
- 	mode-change glitches differ between ][+, //e and IIgs

I have also re-ordered the `model` enums chronologically, which may benefit compilers' range-comparison codegen.

About the timing of each softswitch: some transitions show glitching on hardware.  This is not emulated, so the delay implemented here is only approximate.  I have assembled reference photos for many of these transitions across three classes of hardware, using [my VIDSYNC test](https://github.com/mamedev/mame/pull/15073#:~:text=now%20covers%20all%20video%20softswitches):
VIDSYNC_Latency_glitches
(thanks to @colinleroy for //c PAL and @whscullin for ][+ hardware testing)
<br>

The `read_floatingbus()` changes make five improvements:
- Replace four nearly-duplicate implementations with a shared one, which retains most of the old code, although I have cherry-picked some [later cleanup](https://github.com/AppleWin/AppleWin/commit/779d2d4385c1432d433cb3471c3d4a65c522ba8d).  
- The shared implementation's calling convention requires callers to pass the correct hpos/vpos and apply any machine-specific masking to the resulting RAM address.  This makes the shared implementation NTSC/PAL-agnostic, and removes the [incorrect dependency](https://github.com/mamedev/mame/pull/14053#:~:text=magically%20scan%20faster) on `m_maincpu->total_cycles()`, so now it is possible (with accelerator-aware software) for vaporlock to work on machines which run at > 1 MHz, instead of the addresses irreversibly drifting out of sync when the CPU clock speed changes.
- Making the callers correctly handle `hpos()` wrapping also fixes off-by-one bugs in the old code, so the scanned addresses during HBL now match hardware (see Testing details).
- The previous `(i + 63)` [hard-coded to fix some beam-racing demos](https://github.com/mamedev/mame/blame/4044303cd767b13f5490d69974cae6e8d0cd6f04/src/mame/apple/apple2.cpp#L655) by compensating for a specific softswitch delay is no longer needed, now that the softswitch-specific delays are implemented.
- apple2gs also adds emulation of the [open-bus Mega II cycles](https://www.brutaldeluxe.fr/documentation/cortland/v5/Cortland%20Custom%20ICs%20-%20Wayne%20Lowry%20-%20Preliminary%20notes%20-%2019860214.pdf#page=29).  This breaks [Lancaster-style vaporlock](https://www.tinaja.com/ebooks/enhance_vII.pdf#page=197), matching hardware behavior.  [Other vaporlock techniques](https://www.applefritter.com/content/transmitting-simple-datagram-video-vapor) which don't rely on UNUSED8 screen holes can still work on the IIgs.

The g65816 core does not track the most-recent-byte-fetched, so this apple2gs open-bus code approximately reconstructs it by peeking at the PC, which then has to avoid infinite loop recursion.  This is... not pretty.  But after I wrote that, I found almost identical patterns in the [Sega 315-5195](https://github.com/mamedev/mame/blob/286daa549752f459de0651c8b252d0e5193e5615/src/mame/sega/315_5195.cpp#L77) and the [SNES](https://github.com/mamedev/mame/blob/286daa549752f459de0651c8b252d0e5193e5615/src/mame/nintendo/snes_m.cpp#L225) (which looks broken to me? `m_maincpu->pc()` doesn't handle PB, also bank wrapping?)
<br>

After fixing [all](https://github.com/mamedev/mame/pull/14590) [the](https://github.com/mamedev/mame/pull/14749) [previous](https://github.com/mamedev/mame/pull/15073) timing interactions, the 50Hz changes are straightforwards, dynamically reconfiguring the screen when LANGSEL is latched (but see RFC section...)  Unlike the 8-bit PAL machines, no clock changes are needed for NTSC-50.

</details>
<br>

The changes here are exemplified by Sather's "[Little Text Window](https://archive.org/details/Understanding_the_Apple_IIe/page/n67/mode/2up)".  In MAME 0.286 before the [HBL relative timing](https://github.com/mamedev/mame/pull/15073) fixes:
SATHER LTW_ee_0286_clip
In 0.287 before this PR:
SATHER LTW_ee_0287_before_clip
After this PR:
SATHER LTW_ee_0287_after_clip
The VIDMODES.po (in Testing details) includes a collection of these simple 3rd-party examples.
<br>

Similar changes are visible in real demos, like the center divider in Deater's Megademo:
MEGADEMO_ee_287_before
...is now missing (matching [hardware behavior](http://deater.net/weave/vmwprod/megademo/cycle_switching.html#:~:text=The%20Starring%20Scenes)):
MEGADEMO_ee_287_after
...and in CRAZY CYCLES II and other demos where page flips were previously occuring a couple cycles too far to the right.
<br>

On apple2gs, visible changes from fixing vaporlock drift are less subtle, for example the corruption on the [ending credits screen](https://github.com/AppleWin/AppleWin/issues/241#:~:text=A%20quick%20summary) of [French Touch](http://fr3nch.t0uch.free.fr/)'s ANSI STORY:
ANSI_STORY_gs_287_before
...is now fixed:
ANSI_STORY_gs_287_after
<br>

With 50Hz emulation in place, the Apple IIgs "Set system standards and 50 hertz" screen [via Ctrl-Option-Reset](https://savagetaylor.com/TIL/TIL02040.pdf) now works as expected.  There is apparently not much Apple IIgs software that _requires_ 50Hz, but one example is [Ninjaforce](https://www.ninjaforce.com/html/products.html)'s [Revenge of the Bobs](https://www.ninjaforce.com/html/products_revenge.html) demo, which at 2.8MHz has some missing vectorballs if run at 60Hz.

Other IIgs-aware and [Hz-aware](https://www.colino.net/wordpress/archives/2026/02/23/the-challenges-of-porting-shufflepuck-cafe-to-the-8-bits-apple-ii/#:~:text=one%20frame%20is%20dropped%20every%20sixth%20frame) examples are [Glider](https://www.colino.net/wordpress/glider-for-apple-ii/) and [Shufflepuck Cafe](https://www.colino.net/wordpress/shufflepuck-cafe-for-apple-ii/).

</details>
<br>
Testing:<details>

There are two disks of focused unit tests (source included):
[VidModes_260421.zip](https://github.com/user-attachments/files/26952329/VidModes_260421.zip)
[FloatBus_260630.zip](https://github.com/user-attachments/files/29534724/FloatBus_260630.zip)
(updated 260630, to remove the CFFF Sanity test after #15619)
<br>

A new VIDMODES test is a "beam-race all the things" dashboard, showing all softswitch latencies and glitches on one screen:
VIDMODES_gs_rgb

Before this PR without the softswitch-specific delays, the alignments are off by a cycle or two:
VIDMODES_ee_287_before_clip
After this PR:
VIDMODES_ee_287_after_clip
...MAME now more closely aligns with hardware:
VIDMODES_ee_composite_clip
...except the transition glitches aren't emulated, and some 80-column delays are not quite right (see TODO section.)
Results are similar for apple2p. apple2gs was very broken before this PR, since the scanner address drift was unavoidable even when booting at 1 MHz.
<br>

Improvement is also visible in some older tests, which are included in the VIDMODES .po here:

My [previous VIDSYNC test](https://github.com/mamedev/mame/pull/15073#:~:text=updated%20again%20here) has been updated again with support for SCBIRQ sync and apple2jp katakana:
VIDSYNC_2jp_katakana
After this PR, it now shows that previous minor alignment glitches are fixed on apple2p and apple2ee.  This test is still broken (out of sync rolling video) on apple2gs, due to [known g65816 timing bugs](https://github.com/mamedev/mame/blob/2b54531dc9a5e23b511a212554d2b4dfb18d1e14/src/devices/cpu/g65816/g65816.cpp#L81) (see TODO section.)

My [previous FLASHHZ test](https://github.com/mamedev/mame/pull/14554#:~:text=dynamically%20toggle) now shows apple2gs works correctly when toggling 50Hz.

My [previous LANGSEL test](https://github.com/mamedev/mame/pull/14590#:~:text=MAME%20must%20be%20running) now works even when not run at 1 MHz, minor alignment glitches are fixed, and apple2gs works correctly when toggling 50Hz.

My [previous IRQTEST test](https://github.com/mamedev/mame/pull/14749#:~:text=updated%20here) now shows that apple2gs 1/4 second IRQs work with both 60 and 50Hz display timings.

My [previous BEAMPOS test](https://github.com/mamedev/mame/pull/15073#:~:text=BEAMPOS%20analytically%20validates) now shows the PAL and F-BUS failures on apple2gs are fixed.  Only the 16BIT failure remains (because MAME's g65816 core is not cycle-exact, it [reads lo and hi bytes on the same cycle](https://github.com/mamedev/mame/blob/49f29790a37ece6aaaeb9f50463f1a040500b21b/src/devices/cpu/g65816/g65816.cpp#L310)...)

Updated versions of my [previous CALIBRATE tests](https://github.com/mamedev/mame/pull/15073#:~:text=simple%20calibration%20gauges) are also included: VPR serves as a negative test for Lancaster-style vaporlock on the IIgs, and GS16BIT demonstrates the adjacent softswitch access pictured above in the code comments.
<br>

Also included is a 3rd-party EXAMPLES folder of classic and more recent beam-racing and floating bus tests:
- BISHOP.SPLIT
- SATHER.SPLIT
- SATHER.UNDER
- SATHER.WINDOW
- MMPHOS.RAINBOW
- QKUMBA.JSRC050
- QKUMBA.JSRC000
- ELLIOT.KILLER
- ELLIOT.WINDOW

These have each been modified to improve compatibility (NTSC/PAL support, VBL differences, disabling Zip Chip, etc) and to make them runnable in sequence from a menu.
<br>

A new FLOATBUS test is a self-driving suite that iterates through every relevant video mode, and:
* draws a test pattern designed to highlight the memory layout of that mode
* reads the entire NTSC/PAL video scan-out via floating bus
* spot-checks a handful of pixels in locations prone to off-by-one errors
* compresses the snapshot and saves it to disk for offline inspection

The snapshots can be converted to GIF on a modern machine.  Mac (fat, unsigned) snap2gif tool and (posix-y) source is included.

The test is aware of ][+ and IIgs-specific floating bus behavior, and produces a set of 9 (or 11 on the IIgs) images, like these:
P4 HGR2 MIX 60_spotcheck_overlay
P7 GR MIX 60
PB SHR DP 50
(These examples were captured from //e Enhanced and IIgs ROM3 hardware.  HBL/VBL are visualized in gray, open-bus cycles in pink.  Spot-checked locations are circled in yellow in the first image.)

This allows visualization and debugging of the HBL/VBL memory layout, as described by [Bob Bishop](https://web.archive.org/web/20251017162803/http://rich12345.tripod.com/aiivideo/softalk.html#:~:text=Cycle%2Dby%2Dcycle%20map%20of%20video%20frame) and [Jim Sather](https://archive.org/details/understanding_the_apple_ii/page/n99/mode/2up).

Note that 80-column modes are not iterated, as testing shows Sather is correct (in [UTA2e:5-42 point 10](https://archive.org/details/Understanding_the_Apple_IIe/page/n147/mode/2up)) in that AUX memory read by the video scanner is never put on the data bus visible to the CPU.  There is instead some minimal negative testing to show that C00D 80VID doesn't alter the floating bus behavior.  AFAICT, deciphering the floating bus behavior of IIgs SHR video (the SCB, palette, and open-bus timing) is [original research](https://apple2infinitum.slack.com/archives/CPSGNGE05/p1768961181991359?thread_ts=1768922592.450429&cid=CPSGNGE05) not implemented in any other emulator.

Before this PR, apple2gs failed this test completely, due to the video scanner drifting out of sync.  apple2p and apple2e machines mostly worked, but failed spotcheck <i>#</i>4 due to an off-by-one in the HBL: `read_floatingbus()` shifted the hpos over two cycles, but didn't handle the overflow into vpos, so most of HBL was incorrectly shifted down one scanline:
P4 HGR2 MIX 60_ee_287_broken

After this PR, all spotchecks now pass except for <i>#</i>6 on apple2p/apple2e.  This is a literal corner-case, caused by a bug in MAME's scheduler (see RFC section.)  Both before and after this PR, there is also a sanity-test failure reading CFFF CLRROM (see TODO section.)

More generally, even the "[emulator-killer](https://apple2infinitum.slack.com/archives/CA8AT5886/p1570821404081200)" qkumba examples [executing softswitches](https://apple2infinitum.slack.com/archives/C055PTN6VC7/p1700746051242479) work correctly (at least with the cycle-accurate 6502 core; the g65816 reads operands on the wrong cycles...)

</details>
<br>
TODO:<details>

* Certain glitch-less 80-column video transitions (C02B LANGSEL, C029 MONO bit, C00F ALTCHAR) actually occur seven 14M cycles later on hardware.  This PR does not attempt to address this, as fixing it would need invasive changes to [render_line*](https://github.com/mamedev/mame/blob/218cf0d247a51d96ae1c77752d116659be976c8c/src/mame/apple/apple2video.cpp#L355) and the various [clipping assumptions](https://github.com/mamedev/mame/blob/218cf0d247a51d96ae1c77752d116659be976c8c/src/mame/apple/apple2video.cpp#L622).  Perhaps revisit this if 80-column placement (for apple2e) is fixed.

* CFFF CLRROM has some edge-case behavior which may require a2bus changes to make it pedantically correct, so is not dealt with here.

* apple2gs now accurately implements SHR floating bus behavior, and the open-bus behavior of the Mega II memory refresh cycles.  However on hardware, non-SHR modes show a rolling pattern (out of sync with the video refresh) during the rest of the HBL/VBL cycles.  This is not yet understood.

* Although this collection of PRs fix many subtle video timing behaviors, apple2gs still has [CPU](https://github.com/mamedev/mame/blob/218cf0d247a51d96ae1c77752d116659be976c8c/src/devices/cpu/g65816/g65816.cpp#L81) and [FPI](https://github.com/mamedev/mame/commit/53a0ae7d11b82ab13564f2c82f43407337343d67) timing problems which break VIDSYNC, QKUMBA.JSRC050, the ECC demo's border effects, TextFunk, 3200 color mode, etc.
</details>

### DIFF
--- a/src/mame/apple/apple2.cpp
+++ b/src/mame/apple/apple2.cpp
@@ -607,129 +607,18 @@ void apple2_state::inh_w(offs_t offset, u8 data)
 	}
 }
 
-// floating bus code from old machine/apple2: now works reasonably well with French Touch and Deater "vapor lock" stuff
 u8 apple2_state::read_floatingbus()
 {
-	enum
-	{
-		// scanner constants
-		kHBurstClock      =    53, // clock when Color Burst starts
-		kHBurstClocks     =     4, // clocks per Color Burst duration
-		kHClock0State     =  0x18, // H[543210] = 011000
-		kHClocks          =    65, // clocks per horizontal scan (including HBL)
-		kHPEClock         =    40, // clock when HPE (horizontal preset enable) goes low
-		kHPresetClock     =    41, // clock when H state presets
-		kHSyncClock       =    49, // clock when HSync starts
-		kHSyncClocks      =     4, // clocks per HSync duration
-		kNTSCScanLines    =   262, // total scan lines including VBL (NTSC)
-		kNTSCVSyncLine    =   224, // line when VSync starts (NTSC)
-		kPALScanLines     =   312, // total scan lines including VBL (PAL)
-		kPALVSyncLine     =   264, // line when VSync starts (PAL)
-		kVLine0State      = 0x100, // V[543210CBA] = 100000000
-		kVPresetLine      =   256, // line when V state presets
-		kVSyncLines       =     4, // lines per VSync duration
-		kClocksPerVSync   = kHClocks * kNTSCScanLines // FIX: NTSC only?
-	};
+	int h_clock = m_screen->hpos() / 14 + 25; // adjust for set_raw
+	if (h_clock >= 65)
+		h_clock -= 65;
 
-	// vars
-	//
-	int i, Hires, Mixed, Page2, _80Store, ScanLines, /* VSyncLine, ScanCycles,*/
-		h_clock, h_state, h_0, h_1, h_2, h_3, h_4, h_5,
-		v_line, v_state, v_A, v_B, v_C, v_0, v_1, v_2, v_3, v_4, /* v_5, */
-		_hires, addend0, addend1, addend2, sum, address;
+	int v_clock = m_screen->vpos() + (h_clock < 25); // adjust for hpos wrap
+	if (v_clock >= m_screen->height())
+		v_clock -= m_screen->height();
 
-	// video scanner data
-	//
-	i = m_maincpu->total_cycles() % kClocksPerVSync; // cycles into this VSync
-
-	// machine state switches
-	//
-	Hires    = (m_video->get_hires() && m_video->get_graphics()) ? 1 : 0;
-	Mixed    = m_video->get_mix() ? 1 : 0;
-	Page2    = m_video->get_page2() ? 1 : 0;
-	_80Store = 0;
-
-	// calculate video parameters according to display standard
-	//
-	ScanLines  = 1 ? kNTSCScanLines : kPALScanLines; // FIX: NTSC only?
-	// VSyncLine  = 1 ? kNTSCVSyncLine : kPALVSyncLine; // FIX: NTSC only?
-	// ScanCycles = ScanLines * kHClocks;
-
-	// calculate horizontal scanning state
-	h_clock = (i + 63) % kHClocks; // which horizontal scanning clock
-	h_state = kHClock0State + h_clock; // H state bits
-	if (h_clock >= kHPresetClock) // check for horizontal preset
-	{
-		h_state -= 1; // correct for state preset (two 0 states)
-	}
-	h_0 = (h_state >> 0) & 1; // get horizontal state bits
-	h_1 = (h_state >> 1) & 1;
-	h_2 = (h_state >> 2) & 1;
-	h_3 = (h_state >> 3) & 1;
-	h_4 = (h_state >> 4) & 1;
-	h_5 = (h_state >> 5) & 1;
-
-	// calculate vertical scanning state
-	//
-	v_line  = (i / kHClocks) + 192; // which vertical scanning line
-	v_state = kVLine0State + v_line; // V state bits
-	if ((v_line >= kVPresetLine)) // check for previous vertical state preset
-	{
-		v_state -= ScanLines; // compensate for preset
-	}
-	v_A = (v_state >> 0) & 1; // get vertical state bits
-	v_B = (v_state >> 1) & 1;
-	v_C = (v_state >> 2) & 1;
-	v_0 = (v_state >> 3) & 1;
-	v_1 = (v_state >> 4) & 1;
-	v_2 = (v_state >> 5) & 1;
-	v_3 = (v_state >> 6) & 1;
-	v_4 = (v_state >> 7) & 1;
-	//v_5 = (v_state >> 8) & 1;
-
-	// calculate scanning memory address
-	//
-	_hires = Hires;
-	if (Hires && Mixed && (v_4 & v_2))
-	{
-		_hires = 0; // (address is in text memory)
-	}
-
-	addend0 = 0x68; // 1            1            0            1
-	addend1 =              (h_5 << 5) | (h_4 << 4) | (h_3 << 3);
-	addend2 = (v_4 << 6) | (v_3 << 5) | (v_4 << 4) | (v_3 << 3);
-	sum     = (addend0 + addend1 + addend2) & (0x0F << 3);
-
-	address = 0;
-	address |= h_0 << 0; // a0
-	address |= h_1 << 1; // a1
-	address |= h_2 << 2; // a2
-	address |= sum;      // a3 - aa6
-	address |= v_0 << 7; // a7
-	address |= v_1 << 8; // a8
-	address |= v_2 << 9; // a9
-	address |= ((_hires) ? v_A : (1 ^ (Page2 & (1 ^ _80Store)))) << 10; // a10
-	address |= ((_hires) ? v_B : (Page2 & (1 ^ _80Store))) << 11; // a11
-	if (_hires) // hires?
-	{
-		// Y: insert hires only address bits
-		//
-		address |= v_C << 12; // a12
-		address |= (1 ^ (Page2 & (1 ^ _80Store))) << 13; // a13
-		address |= (Page2 & (1 ^ _80Store)) << 14; // a14
-	}
-	else
-	{
-		// N: text, so no higher address bits unless Apple ][, not Apple //e
-		//
-		if ((kHPEClock <= h_clock) && // Y: HBL?
-			(h_clock <= (kHClocks - 1)))
-		{
-			address |= 1 << 12; // Y: a12 (add $1000 to address!)
-		}
-	}
-
-	return m_ram_ptr[address % m_ram_size];
+	const u16 address = m_video->scanner_address(h_clock, v_clock);
+	return ram_r(address);
 }
 
 /***************************************************************************
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -1520,7 +1520,7 @@ void apple2e_state::accel_full_speed()
 
 void apple2e_state::accel_normal_speed()
 {
-	m_maincpu->set_unscaled_clock(1021800, true); // re-align to PH0
+	m_maincpu->set_unscaled_clock(m_pal ? 1016966 : 1021800, true); // re-align to PH0
 }
 
 void apple2e_state::accel_slot(int slot)
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -3315,118 +3315,18 @@ void apple2e_state::lc_w(offs_t offset, u8 data)
 	}
 }
 
-// floating bus code from old machine/apple2: now works reasonably well with French Touch and Deater "vapor lock" stuff
 u8 apple2e_state::read_floatingbus()
 {
-	enum
-	{
-		// scanner constants
-		kHBurstClock      =    53, // clock when Color Burst starts
-		kHBurstClocks     =     4, // clocks per Color Burst duration
-		kHClock0State     =  0x18, // H[543210] = 011000
-		kHClocks          =    65, // clocks per horizontal scan (including HBL)
-		kHPEClock         =    40, // clock when HPE (horizontal preset enable) goes low
-		kHPresetClock     =    41, // clock when H state presets
-		kHSyncClock       =    49, // clock when HSync starts
-		kHSyncClocks      =     4, // clocks per HSync duration
-		kNTSCScanLines    =   262, // total scan lines including VBL (NTSC)
-		kNTSCVSyncLine    =   224, // line when VSync starts (NTSC)
-		kPALScanLines     =   312, // total scan lines including VBL (PAL)
-		kPALVSyncLine     =   264, // line when VSync starts (PAL)
-		kVLine0State      = 0x100, // V[543210CBA] = 100000000
-		kVPresetLine      =   256, // line when V state presets
-		kVSyncLines       =     4, // lines per VSync duration
-	};
+	int h_clock = m_screen->hpos() / 14 + 25; // adjust for set_raw
+	if (h_clock >= 65)
+		h_clock -= 65;
 
-	const int kClocksPerVSync = kHClocks * (m_pal ? kPALScanLines : kNTSCScanLines);
+	int v_clock = m_screen->vpos() + (h_clock < 25); // adjust for hpos wrap
+	if (v_clock >= m_screen->height())
+		v_clock -= m_screen->height();
 
-	// vars
-	//
-	int i, Hires, Mixed, Page2, _80Store, ScanLines, /* VSyncLine, ScanCycles,*/
-		h_clock, h_state, h_0, h_1, h_2, h_3, h_4, h_5,
-		v_line, v_state, v_A, v_B, v_C, v_0, v_1, v_2, v_3, v_4, /* v_5, */
-		_hires, addend0, addend1, addend2, sum, address;
-
-	// video scanner data
-	//
-	i = m_maincpu->total_cycles() % kClocksPerVSync; // cycles into this VSync
-
-	// machine state switches
-	//
-	Hires = (m_video->get_hires() && m_video->get_graphics()) ? 1 : 0;
-	Mixed = m_video->get_mix() ? 1 : 0;
-	Page2 = m_video->get_page2() ? 1 : 0;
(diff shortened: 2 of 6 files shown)


arekkusu42:
RFC, please review:

1) `update_now()`: I think the core changes are minimally intrusive, but is there a better pattern to handle this sort of pipelined delay?

2) `m_base_model`: is there a better way to do this?  Like adding an optional argument to the `a2_video_device` c'tor?

3) `read_floatingbus()` still has a corner-case off-by-one bug. The root cause is yet more [time-travel shenanigans](https://github.com/mamedev/mame/issues/14474#issuecomment-3557096351
) in MAME's scheduler: when time reaches the active video's lower-right pixel, the VBL timer callback [resets](https://github.com/mamedev/mame/blob/9af71d10f9c1790d51b27e231aba767b1397cf9e/src/emu/screen.cpp#L1664) `m_vblank_start_time`. When CPU execution continues, for a brief moment (~one screen pixel)
`machine().time()` [can be in the past](https://github.com/mamedev/mame/blob/9af71d10f9c1790d51b27e231aba767b1397cf9e/src/emu/screen.cpp#L1555) relative to `m_vblank_start_time`, which results in a negative `hpos()`, and subsequent wrong address.  What to do about this?

4) `screen_device::configure()` does a lot of work, including re-allocating the bitmap and textures, and no effort is made to preserve the previous contents, so there is one frame of flicker.  What I would like to do on a refresh rate change is adjust the visible area of the only-allocated-once bitmap.  Is there a better way to handle this?

arekkusu42:
Whoops, I had `pc()` and `g65816_get_reg()` mixed up.  `pc()` does the right thing and returns `REGISTER_PB | REGISTER_PC`, so now open-bus reads from other program banks will work.  This means that the SNES implementation is correct too (except for the edge-case of executing from the last byte of a bank...)

happppp:
Hi, from MAME's core screen device POV, I don't think a delay param in update_now() is a good idea, even though it works. Moreover the way it does it, by temp modifying vblank start time and relying on vpos()/hpos() determining their position from that variable.

Could you rebase to https://github.com/mamedev/mame/commit/56ea076a1fd8b308a4aad3865ec13f2f5fb01c94
And change your delayed_update to use update_partial(calc new vpos, calc new hpos); ?

@rb6502 can look at the apple specific changes, they look fine to me at a glance.

arekkusu42:
@happppp thanks!  That does look like a cleaner core interface.

But, after adapting my delayed_update(), the new update_partial() doesn't work.  An example on apple2ee: the last scanline of Sather's Little Text Window is not being flushed correctly like it was in my old implementation:
SATHER LTW_ee_0287_update_partial_clip

With LOG_PARTIAL_UPDATES, and logging the cliprect that a2_video_device::screen_update receives, one video frame looks like this:
```
[:screen] update_partial(): Y=88, X=140, last partial 0, partial hpos 0  (vis 559 191)
[:screen] Partial: update_partial(:screen, 87): [:screen] updating 0-87
a2v screen_update [  0,  0...559, 87]
[:screen] doing scanline partial draw: Y 88 X 0-139
a2v screen_update [  0, 88...139, 88]
[:screen] update_partial(): Y=88, X=434, last partial 88, partial hpos 140  (vis 559 191)
[:screen] doing scanline partial draw: Y 88 X 140-433
a2v screen_update [140, 88...433, 88]
[:screen] update_partial(): Y=89, X=140, last partial 88, partial hpos 434  (vis 559 191)
a2v screen_update [434, 88...559, 88]
[:screen] doing scanline partial draw: Y 89 X 0-139
a2v screen_update [  0, 89...139, 89]
[:screen] update_partial(): Y=89, X=434, last partial 89, partial hpos 140  (vis 559 191)
[:screen] doing scanline partial draw: Y 89 X 140-433
a2v screen_update [140, 89...433, 89]
    // snip same pattern for scanlines 90...94
[:screen] update_partial(): Y=95, X=140, last partial 94, partial hpos 434  (vis 559 191)
a2v screen_update [434, 94...559, 94]
[:screen] doing scanline partial draw: Y 95 X 0-139
a2v screen_update [  0, 95...139, 95]
[:screen] update_partial(): Y=95, X=434, last partial 95, partial hpos 140  (vis 559 191)
[:screen] doing scanline partial draw: Y 95 X 140-433
a2v screen_update [140, 95...433, 95]
[:screen] Partial: update_partial(:screen, 191): [:screen] updating 95-191
[:screen] update_partial(): Y=95, X=0, last partial 95, partial hpos 434  (vis 559 191)
a2v screen_update [  0, 95...559,191]
                   ^^^ bug here, should not re-flush X 0-433!
```
It looks like the "[left off in the middle of a scanline](https://github.com/mamedev/mame/blob/b8e2cd3d010b16b207e0e9f25c8a31d6b7968284/src/emu/screen.cpp#L1195)" logic doesn't trigger properly?

cuavas:
Yeah, the hack in `screen_device::update_now` really doesn’t pass the smell test.

happppp:
Sorry, it's missing a +1, I'll fix it.
-> https://github.com/mamedev/mame/commit/3b1bc6244cc68a929d6a6f6b7ca0289cdfc3cd4a

arekkusu42:
@happppp thanks, `update_partial()` works now.  I manually re-merged my commits back.
Note that `delayed_update()` has to do a little extra work now, to handle the hpos overflow into vpos.

Any comment on my RFC <i>#</i>3?  What should I do when `hpos()` returns a negative value?

happppp:
Negative hpos()? Other than "that's an annoying issue", no I have nothing to say.
BTW in your updated delayed_update:
```
if (hpos > screen().width()) // off by 1 error
if (hpos >= screen().width()) // good

or if you wanted to shorten it...

int hpos = screen().hpos() + (cycles - m_delay_bias) * m_scanner_period;
int vpos = screen().vpos() + hpos / screen().width();
hpos %= screen().width();
then update
```

Now that I think about it, is it possible for it to overflow to the next frame?
So, physical beam is at max_y,max_x, and you do a delayed_update, wishing to update until y=0,x=3, is that a problem?

arekkusu42:
Thanks-- two dimensions are more bug-prone than one!

I prefer to use a predictable branch over integer divide.

Yes, the edge case of touching a softswitch in the last couple cycles of VBL should spill a delayed update into the next frame.  I tested that, and it's broken with this current code.

PR #15249, opened 2026-04-23
jaleco/megasys1.cpp, jaleco/ms1_gatearray.cpp: Updates and Cleanups

cam900:
- Improve loggings
- Reduce preprocessor defines, Literal tag usages
- Remove unnecessary include
- Fix spacings
- Fix main CPU ROM region size for reduce unnecessary ROM_CONTINUE
- Fix typename value for save state support
- Minor optimization in sprite draw routine
- Fix naming
- Reduce unnecessary struct

jaleco/ms1_gatearray.cpp:
- Use required_address_space for target space to install handler

### DIFF
--- a/src/mame/jaleco/megasys1.cpp
+++ b/src/mame/jaleco/megasys1.cpp
@@ -277,8 +277,8 @@ void megasys1_state::megasys_base_map(address_map &map)
 	map(0x084300, 0x084301).w(FUNC(megasys1_state::screen_flag_w));
 	map(0x088000, 0x0887ff).ram().w(m_palette, FUNC(palette_device::write16)).share("palette");
 	map(0x08c000, 0x08dfff).mirror(0x002000).ram().share(m_objectram); // soldam relies on a mirror at 0x08c000, other games use 0x08e000
-	map(0x090000, 0x093fff).ram().w(m_tmap[0], FUNC(megasys1_tilemap_device::write)).share("scroll0");
-	map(0x094000, 0x097fff).ram().w(m_tmap[1], FUNC(megasys1_tilemap_device::write)).share("scroll1");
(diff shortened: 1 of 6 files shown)


Hammy1986:
Hey cam! Regarding the megasystem 1 graphics, the naming in mame does not match the hardware (scroll 0 , 1, 2 is wrong named.
The hardware is quite like the NMK16 with 1 text layer , 2 graphics layers and 1 sprite layer.

The text layer used the same chip as the other graphics, sure - but it's not fully hooked up, so some of the features do not work for that layer
MAME treats all layers the same, but that's not the case.

Legend of makai is 1 text layer, 1 graphic layer and 1 sprite layer.
I ported it to the double layer board and this is when i found the layer differences!

Regarding the correct names of the scrolls, avenging spirits test mode tells the "official" names.

src/mame/jaleco/megasys1.cpp
```diff
@@ -393,7 +387,8 @@ void megasys1_bc_iosim_state::ip_select_w(u16 data) // TO MCU
 	}
 
 	// if the command is valid, generate an IRQ from the MCU
-	m_maincpu->set_input_line(2, HOLD_LINE);
+	if (!machine().side_effects_disabled())
+		m_maincpu->set_input_line(2, HOLD_LINE);
```
cuavas: Is the side effect check appropriate in this write handler?  Is there a situation when debugging where you’d want to set `m_ip_latched` but not trigger the interrupt?

cuavas (commented):
This breaks jaleco/cischeat.cpp and possibly other drivers using the gate array device.  You really should do a complete build before opening the PR.

src/mame/jaleco/megasys1.cpp
cam900: It is just small mistake when applying PR commit, I think this case it will be respond to wrong input value.

PR #15250, opened 2026-04-24
Commodore: added machine configuration switch in C16, C64, C128 and C65 to swap joystick ports on the fly, avoiding the need to reset the machine when loading a game that requires joystick to be in the opposite port to the one we plugged it in

etabeta78:
Quality of life improvement asked by some friends

### DIFF
--- a/src/mame/commodore/c128.cpp
+++ b/src/mame/commodore/c128.cpp
@@ -78,6 +78,7 @@ public:
 		m_lock(*this, "LOCK"),
 		m_caps(*this, "CAPS"),
 		m_40_80(*this, "40_80"),
+		m_portswap(*this, "JOYSWAP"),
 		m_z80en(0),
 		m_loram(1),
 		m_hiram(1),
@@ -122,6 +123,7 @@ public:
 	required_ioport m_lock;
 	required_ioport m_caps;
 	required_ioport m_40_80;
+	optional_ioport m_portswap;
 
 	virtual void machine_start() override ATTR_COLD;
 	virtual void machine_reset() override ATTR_COLD;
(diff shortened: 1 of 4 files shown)


mamehaze:
on one hand it's a gross hack, on the other the system is basically unusable unless you can do this because much software expects you to change what's in the ports on the fly, and you can't do it properly using MAME's rigid slot system or save state system which requires a hard reset every time you change something.

etabeta78:
Well, I had this in my private build since quite some time, and some friends of mine said it would be great to have this in official MAME

Conscious of its hacky nature, I have made it a configuration settings, the same as we treat jumpers configs, some video hardware variants in dkong, mario, n8080 & astrof, or some debug options in naomigd, msx, etc.
But we can for sure tag it as "(Hack)" in the description, to make clear it's not hw-accurate (even if things like C64 Mega Switcher implement something similar on real HW)

simzy39:
> Well, I had this in my private build since quite some time, and some friends of mine said it would be great to have this in official MAME
> 
> Conscious of its hacky nature, I have made it a configuration settings, the same as we treat jumpers configs, some video hardware variants in dkong, mario, n8080 & astrof, or some debug options in naomigd, msx, etc. But we can for sure tag it as "(Hack)" in the description, to make clear it's not hw-accurate (even if things like C64 Mega Switcher implement something similar on real HW)

Guru said in the forums: "regarding the C64 joy port swap, I did that in hardware years ago with a Atmel ATTiny44 MCU. If someone wants to implement it I could provide the schematic and code. The board plugs into the ports just swapping the lines across. Works with mouse too."

PR #15252, opened 2026-04-25
apple2video: fix -aux std80 crash 

arekkusu42:
This PR fixes apple2e* crashing when using `-aux std80` and displaying DGR or DHGR.

To reproduce:

1. `mame apple2ee -aux std80`  (no disk)
2. Ctrl-Reset to prompt
3. HGR  (or GR)
4. PR#3
5. CALL -151
6. C05E

[apple2ee_std80_dhgr_crash.txt](https://github.com/user-attachments/files/27070692/apple2ee_std80_dhgr_crash.txt)

Most software will check for 128k before attempting to use DHGR, but this also reproduces when running anything that blindly displays DHGR or DGR, for example the VIDMODES test on this disk image:
[VidModes_260424.zip](https://github.com/user-attachments/files/27070696/VidModes_260424.zip)

...although the crash is due to OOB access, so it depends on your system malloc (and ASLR etc.)  On my system a crash in a release build is rare without libgmalloc.
A debug ASAN build (`ARCHOPTS="-fsanitize=address"`) shows the problem immediately:
[apple2ee_std80_dhgr_asan.txt](https://github.com/user-attachments/files/27070765/apple2ee_std80_dhgr_asan.txt)
...and similarly with DGR:
[apple2ee_std80_dgr_asan.txt](https://github.com/user-attachments/files/27070766/apple2ee_std80_dgr_asan.txt)

80-column text (including page2) was not affected, because `a2_video_device::text_update()` applies `m_aux_mask`.  This commit applies similar logic to `lores_update()` and `dhgr_update()`.

I also removed `allow_dhr()` from `a2eauxslot` cards; this was unused.  `get_auxbank_mask()` fulfills the requirement well enough:  MAME does not emulate the early Rev-A Apple //e where C05E DHIRESON isn't functional, but if it did, that indication of DHGR support belongs in the apple2e motherboard and not the auxslot card.

### DIFF
--- a/src/devices/bus/a2bus/a2eauxslot.h
+++ b/src/devices/bus/a2bus/a2eauxslot.h
@@ -110,7 +110,6 @@ public:
 	virtual u8 *get_vram_ptr() = 0;
 	virtual u8 *get_auxbank_ptr() = 0;
 	virtual u16 get_auxbank_mask() = 0;
-	virtual bool allow_dhr() { return true; }
 
 	device_a2eauxslot_card_interface *next() const { return m_next; }
 
--- a/src/devices/bus/a2bus/a2eext80col.h
+++ b/src/devices/bus/a2bus/a2eext80col.h
@@ -38,7 +38,6 @@ protected:
 	virtual u8 *get_vram_ptr() override;
 	virtual u8 *get_auxbank_ptr() override;
 	virtual u16 get_auxbank_mask() override;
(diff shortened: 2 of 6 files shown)


arekkusu42:
For posterity: link to a discussion showing [hardware results](https://retrocomputing.stackexchange.com/questions/1575/quick-way-to-tell-how-much-ram-an-apple-iie-has#:~:text=a%20real%20Apple).

Issue #11477, opened 2023-08-12
Serial output module


stonedDiscord:
This is a feature request because I sadly can't get it to work myself, otherwise this would have been a PR.

I'd like to see an additional "serial" module in src/osd/modules/output

The reason for that being that it'd make it really easy to use USB-Serial devices like Arduinos or ESP32 boards to control LEDs and mechanical devices attached to the computer MAME is running on.

The code for that should be relatively straightforward, it's essentially a mix of console.cpp and network.cpp but instead of boost ASIO network sockets it should use the boost ASIO module for the serial port.

rb6502:
If you have an attempt that you couldn't get to work, maybe open it as a draft PR so people can take a look?

cuavas:
To be honest, I’d be reluctant to support this.  It would add additional complexity and be difficult for developers to test.  Some of the issues:
* Requires a computer with a serial port to sanity check with.  This is becoming rarer, particularly in notebook computers.
* Requires some device that consumes the format to test properly.
* Additional configuration that varies between platforms, as Windows and every *NIX handle serial ports differently.
* The current network output module has issues with systems that have a lot of output or update them at high rates.  This would be even worse with a serial output.
* It would be difficult to update in the future if external hardware starts depending on details of the protocol, as firmware updates are generally slower/more difficult than software updates.

I would recommend that you create a program that connects to MAME’s network output module and converts the data to whatever format it is that your serial attached device requires.

stonedDiscord:
That makes sense.
For reference I'll list how the same thing can be achieved with the network module on Linux

```
stty -F /dev/ttyUSB0 115200
mame -output network
netcat 127.0.0.1 8000 </dev/ttyUSB0 >/dev/ttyUSB0
```

PR #15256, opened 2026-04-25
osd: add serial output provider

KoalaBear:
## osd: add serial output provider

Adds a new OSD output module that streams MAME output notifications over a serial port, with the same neutral one-way semantics as the existing `network` provider. Every `notify(name, value)` becomes one `name = value\n` line on the wire.

### Motivation

MAME's existing output providers (`console`, `network`, `windows`) are software notification channels. Real arcade hardware that implements core arcade behaviour - ticket dispensers, coin hoppers, coin acceptors, lamp boards - talks serial natively. Today the only way to bridge MAME to that hardware is to take `-output network`, write a userspace TCP-to-serial process, and ship/maintain it alongside MAME.

For the common "MAME -> COM port -> I/O board" topology a built-in module is simpler, has fewer moving parts, and matches user expectations. Operators running games on a real cabinet (ddboy, shuriboy, fuusenpn, tsupenta, mariorou, tsukande, and similar ticket/medal games) get a drop-in path to drive their hardware.

### Design

Architecturally identical to `src/osd/modules/output/network.cpp`:

- Implements `osd_module + output_module`, registers under the `output` provider type as `serial`.
- `notify()` formats `"%s = %d\n"` and writes synchronously to the port. Writes are non-blocking (`WriteTotalTimeoutConstant = 0` on Windows, `O_NONBLOCK` on POSIX) so a stuck or disconnected host can never stall the emulation thread.
- Cross-platform via straight Win32 (`CreateFileA` + `SetCommState`) and POSIX `termios`. No third-party deps; no changes to anything outside `src/osd/modules/output/` and the small option/registration edits below.

Two new CLI options:

- `-output_serial_port <name>`: port name, e.g. `COM4` or `/dev/ttyUSB0`. Empty by default; if no port is set when `-output serial` is selected, init logs a clear error and MAME falls back to `-output none` (same fallback flow as any other failed module).
- `-output_serial_baud <rate>`: defaults to `9600`.

### Testing

- Built and ran against `ddboy` (Konami medal, `src/mame/konami/konmedal.cpp`).
- Verified ROM via `mame -verifyroms ddboy` (good).
- Played the game end-to-end with a real serial listener attached:
  - lamp activity (`lamp0 = 1` / `lamp0 = 0`) appears during attract mode at the expected ~130 ms cadence
  - inserting a coin stops the lamp blink (game leaves attract)
  - on a 4-ticket payout, the wire shows exactly four `hopper = 0 -> 1` rising edges at the 100 ms period the driver configures via `HOPPER(config, "hopper", attotime::from_msec(100))`
  - all lines newline-terminated, no truncation, no out-of-order interleaving with other outputs (lamp + hopper coexist correctly)
- `srcclean` run against the new file and the touched files; zero changes.

### Scope / non-goals

Intentionally one-way (MAME -> wire only). Adding an input/feedback path (e.g. external sensor pulses driving the ticket dispenser device) is a much bigger design conversation. There is no clean MAME-core API today for OSD modules to inject data into a specific emulated device, and the right answer probably lives at the device or input layer rather than the output layer. Out of scope for this PR.

### Where this fits in practice

This module is the upstream-side counterpart to Arcadeify, a plug-and-play PCB that sits between a PC running MAME and physical arcade hardware (ticket dispenser, coin hopper, coin acceptor, panel buttons). Arcadeify already speaks serial; with this module MAME can drive it directly without any user-side bridge process. The same module is equally usable by anyone with a different serial-attached I/O board.

Inputs (coin pulses, button presses) reach MAME through a separate HID-keyboard path on the controller side, which is why this module is deliberately output-only.

### Files

- new: `src/osd/modules/output/serial.cpp`
- modified: `scripts/src/osd/modules.lua` (one line, register file)
- modified: `src/osd/modules/lib/osdobj_common.{h,cpp}` (option defines, option entries, `REGISTER_MODULE` line)
- modified: `docs/source/commandline/commandline-{all,index}.rst` (document the provider and its two options)

### DIFF
--- a/docs/source/commandline/commandline-all.rst
+++ b/docs/source/commandline/commandline-all.rst
@@ -1096,9 +1096,16 @@ OSD Output Options
     to connect external outputs such as the LED lights for the Player 1/2 start
     buttons on certain arcade machines.
 
-    You can choose from: ``auto``, ``none``, ``console`` or ``network``
+    You can choose from: ``auto``, ``none``, ``console``, ``network``,
+    or ``serial``.
 
     Note that network port is fixed at 8000.
+
+    The ``serial`` provider writes output notifications as ``name = value``
+    lines (newline-terminated) to a configured serial port - useful for
+    driving external arcade I/O hardware such as ticket dispensers, coin
+    hoppers, and lamp boards.  Configure the port and baud rate with the
+    options below.
 
     Example:
         .. code-block:: bash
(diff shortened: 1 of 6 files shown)


KoalaBear:
**Bumping with some context.** 

Since opening this PR I've shipped the hardware side: custom board, firmware, and Unity SDK, all working in a real cabinet. [Wrote up the full build here.](https://tinyurl.com/mame-tokens)

The MAME gateway in this PR is what closes the loop. With it, any preserved ROM can drive real coin acceptors, hoppers, and ticket dispensers through MAME's existing IO layer. Without it, the rest of the stack is emulating in a vacuum.

Thanks!

stonedDiscord:
While I personally would like to see this because I've made a panel with light up buttons and an Arduino Nano that could use it, it's been denied before in #11477 where you can find an explanation and a workaround.

PR #15257, opened 2026-04-25
mpc60, vx600: add 78C11 internal ROM

devinacker:
Pulling up bit 7 of port C will also configure this as the (single-chip) key & wheel controller for the X7000 and possibly others. I might start a skeleton driver for that later if nobody else does.


rb6502:
Nice!

PR #15261, opened 2026-04-26
psikyo/psikyosh.cpp: Updates

cam900:
- Reduce duplicates
- Make tilemap function into use tilemap.h
- Remove unnecessary functions
- Minor optimizations
- Add/Fix notes
- Fix namings
- Reduce literal tag usages
- Fix typenames
- Reduce preprocessor defines


cuavas:
This completely breaks graphics in `soldivid`.  How much did you actually test this?  Do you think you can easily fix it, or should I just revert it?

PR #15263, opened 2026-04-26
sinclair/next/specnext.cpp: Various improvements.

holub:
* Forced more updates on tilemap writes.
* Better handling DivMMC CS instant switch.
* Pulled up unmapped value on 0x2ffd & 0x3ffd reads.
* next/specnext_layer2.cpp: Improved half-pixel rendering.


holub:
shall we?

holub:
We are close to the cut, could anybody take a look please?

PR #15264, opened 2026-04-26
dvmemory: display 7-bit ASCII

arekkusu42:
(for your consideration...)

394e5f1 changed the debugger memory view to use ISO-8869-1 encoding.
This makes it difficult to debug machines which commonly set the high bit in strings.
This PR uses the lowest-common-denominator of printable 7-bit ASCII.

0.287 before this PR:
debugger_ASCII_287_before

After this PR:
debugger_ASCII_287_after

...and the old `isprint()` behavior prior to 394e5f1, circa 0.241:
debugger_ASCII_241_isprint

The printf %c and %s formatter behavior is left unchanged.

### DIFF
--- a/src/emu/debug/dvmemory.cpp
+++ b/src/emu/debug/dvmemory.cpp
@@ -27,8 +27,9 @@ namespace {
 
 constexpr u8 sanitise_character(u8 ch)
 {
-	// assume ISO-8859-1 (low 256 Unicode codepoints) - tab, soft hyphen, C0 and C1 cause problems
-	return ('\t' == ch) ? ' ' : (0xadU == ch) ? '-' : ((' ' > ch) || (('~' < ch) && (0xa0U > ch))) ? '.' : ch;
+	// reduce to printable 7-bit ASCII
+	ch &= 0x7f;
+	return ((' ' > ch) || ('~' < ch)) ? '.' : ch;
 }
 
 } // anonymous namespace


angelosa:
Isn't this supposedly something driver clients should optionally set on their own?

arekkusu42:
Well, the "best" encoding is surely machine-specific.  Someone debugging a Mac machine and accustomed to MacsBug would expect MacOSRoman in the memory view.  But that's a ball of string, are all machines MAME supports representable in Unicode's legacy computing symbols?

I think 7-bit ASCII is a safe default.

happppp:
I'm fine the way it is right now. It still makes it easy to spot patterns in the decoded 'text'.
I suppose it depends on what you're used to. Hex editors I've used look the same as MAME does nowadays, eg. ÿÿÿÿÿÿÿÿ for a string of FF.

cuavas:
Folding to low ASCII will make it harder to spot patterns in binary data.  Right now, you can spot blocks of ASCII at a glance.  Stripping the high bit makes that difficult, too.

This shouldn’t be the default – the Apple II is the odd one out.  If you want to make it an option, you need to add UI code to support it in all the debugger implementations.

arekkusu42:
Fair enough; a task for another day.

This has been considered, closing.

angelosa:
Consider the case of Japanese systems: I'm sure several of them uses Shift-JIS strings or other Japanese aware codepages. Same deal for stuff like Cyrillic and other misc systems that doesn't really follow ASCII (which aren't that uncommon, particularly in the early days).

arekkusu42:
Right, the "best" encoding should be machine-specific.  When emulating more modern systems (a Windows PC...) where there are multiple codepages, making the memory view encoding a user option makes sense.

PR #15265, opened 2026-04-26
Sanyo phc25 - fix VBL interrupt timing and VBL polarity in port $40

Aurelien34:
2 fixes:

# VBL bit polarity was wrong
There is an inversion between the graphic chip and the processor
Here is a program that shows loop counts waiting for the VBL bit to be 0, then 1, then 0 again..  then counts the opposite.
On the real machine :
image
On MAME before the fix :
image
On MAME after the fix :
image

The timing is still not the same, but I don't know yet if it comes from a graphic chip emulation timing of graphic vs CPU speed difference. It looks like things became worse in november as someone made some cleanup in the vintage graphics chips code.

# VBL interrupt duration was wrong
The initial interrupt duration was really wrong (100us instead of 1200ns mesured on the real hardware). I have adjusted the timing so that the demo I did last year works as fast in MAME as on the original machine. It looks like 3us is a good compromise.
Here is a capture of the sync signal coming from the graphics chip and the interrupt signal :
image


happppp:
1200ns is too short for the Z80 to always detect a level triggered IRQ.
3us is also too short, it's 12 cycles on 4MHz, which means IRQ will be missed during opcodes that take longer than 12 cycles, even on a real Z80.

Is what you're seeing on your measurement an IRQ+auto ack? Like, the thing that triggers the IRQ is connected to a 7474, and the Z80 clears the latch with the interrupt acknowledge cycle. This would be equivalent to "HOLD_LINE" in MAME.

Aurelien34:
You are right, we found the serive manual or the machine, and the FS (field sync) is injected in a 74LS74 that holds the interrupt line.
It is cleared by the usual Z80 /M1+/IORQ signals.
Let me change the code and do some testing before I submit a fix.
The emulation of this machine starts looking really fine!
image

happppp:
Good to hear it's simply due to the interrupt ack cycle. Simplest way in MAME:
```
m_vdg->fsync_wr_callback().set_inputline(m_maincpu, 0, HOLD_LINE); // auto-ack
add .invert() if fsync_wr writes 0 to trigger irq
```
(and remove the trampoline function)

Aurelien34:
OK, I have tested my changes with the trampoline function before I see your comment.... Let me change that!

Aurelien34:
And the cleanup is done!
Thanks for your kind help!

PR #15267, opened 2026-04-27
sound/cem3340.cpp: Anti-aliasing. Streaming frequency and PW control.

m1macrophage:
* Used the *PolyBLEP* algorithm for the ramp and pulse waves.
* Used the *PolyBLAMP* algorithm for the triangle wave.
* Made waveform outputs conditional on the output actually being connected.
* Added support for streaming control of the frequency control current and pulse width control voltage.

Effects of anti-aliasing:
comparison

* From top to bottom: ramp, pulse and triangle waveforms.
* Each spectogram shows a sequence of notes, increasing by 1 octave per step.
* *Before* (left column): the ramp and pulse waveforms suffered from significant aliasing. Aliasing on the triangle wave was only noticeable at high frequencies.
* *After* (right column): Significant reduction in aliasing on all waveforms. The ramp and pulse spectra look much cleaner, and the triangle is missing some of the "stray" harmonics in the top octaves.

### DIFF
--- a/src/devices/sound/cem3340.cpp
+++ b/src/devices/sound/cem3340.cpp
@@ -24,7 +24,9 @@ constexpr float PW_MAX = VCC / 3.0F;
 // See "supplies" section in the datasheet. The minimum for both triangle and
 // ramp is always 0.
 constexpr float TRIANGLE_MAX = VCC / 3.0F;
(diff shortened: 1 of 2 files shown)


galibert:
Yay for band-limited oscillators :-)

Issue #14382, opened 2025-10-20
SoftFloat 3 conflicts with ARM NEON SIMD headers


cuavas:
Both SoftFloat 3 and the ARM NEON SIMD headers declare types called `float64_t`, `float32_t`, etc. in the global namespace.  This causes everything to break if something ends up including headers for both (e.g. getting SoftFloat 3 via the M68000 CPU core and getting NEON some other way).

I encountered this when I was mucking around porting MAME’s RGB utilities to NEON – it makes it impossible to compile misc/gammagic.cpp and namco/namcos22.cpp.

cuavas:
Examples of the failures in 64-bit ARM CI builds:
* macOS: https://github.com/cuavas/mame/actions/runs/18656301124/job/53186156512
* Windows: https://github.com/cuavas/mame/actions/runs/18656301129/job/53186156633

PR #15270, opened 2026-04-27
3rdparty: retire softfloat2

pmackinlay:


src/devices/cpu/i386/i386.cpp
```diff
@@ -25,6 +25,8 @@
 #include "cycles.h"
 #include "i386ops.h"
 
+#include "softfloat3/bochs_ext/softfloat3_ext.h"
+
 #include "debug/debugcpu.h"
 #include "debug/express.h"
```
cuavas, line 29: The softfloat header should come after all the MAME headers (local, then most dependent to least dependent).

src/devices/cpu/i386/i386.h
```diff
@@ -5,12 +5,7 @@
 
 #pragma once
 
-// SoftFloat 2 lacks an include guard
-#ifndef softfloat2_h
-#define softfloat2_h 1
-#include "softfloat/milieu.h"
-#include "softfloat/softfloat.h"
-#endif
+#include "softfloat3/source/include/softfloat.h"
```
cuavas: Same here, softfloat header after MAME headers, but before standard library headers.

src/devices/machine/i8087.cpp
```diff
@@ -171,22 +155,22 @@ i8087_device::i8087_device(const machine_config &mconfig, const char *tag, devic
...
+	save_item(NAME(m_reg[3].signExp));
+	save_item(NAME(m_reg[3].signif));
+	save_item(NAME(m_reg[4].signExp));
+	save_item(NAME(m_reg[4].signif));
+	save_item(NAME(m_reg[5].signExp));
+	save_item(NAME(m_reg[5].signif));
+	save_item(NAME(m_reg[6].signExp));
+	save_item(NAME(m_reg[6].signif));
+	save_item(NAME(m_reg[7].signExp));
+	save_item(NAME(m_reg[7].signif));
```
cuavas: Can you use the `STRUCT_MEMBER` helper to make this prettier?

cuavas (commented):
You also need to update 3rdparty/README.md to remove it from the list of licenses.

This will expose more stuff to #14382 but I don’t have any solution for that.

We need to remove SoftFloat 2 at some point, anyway

pmackinlay:
Thanks for the review - have addressed those issues. I agree softfloat is a pretty leaky/bad library in many ways; ideally someone will re-imagine it as a modern C++ library, but it's not going to be me.

angelosa:
As mentioned a while back `xtom3d` is a good test case for FPU stuff https://github.com/mamedev/mame/commit/3cec3eb18dd99584ea94d69d3cfa0237c1ff657b#commitcomment-176692303

PR #15273, opened 2026-04-27
bishi.cpp: add lamp write handler

nutki:
Adding lamp output for bishi.cpp using naming consistent with later game from the series (the ones using ksys573).
Tested with "lamp check" in all 4 supported games.

### DIFF
--- a/src/mame/konami/bishi.cpp
+++ b/src/mame/konami/bishi.cpp
@@ -111,7 +111,11 @@ public:
 		m_k054338(*this, "k054338"),
 		m_k055555(*this, "k055555"),
 		m_palette(*this, "palette"),
-		m_screen(*this, "screen")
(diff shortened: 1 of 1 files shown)


cuavas:
It’s cleaner to just number the outputs and let whatever they’re hooked up to work out what to do with them.  Also, it’s fine to have the unused outputs active for the two-player variant.

nutki:
> It’s cleaner to just number the outputs and let whatever they’re hooked up to work out what to do with them. Also, it’s fine to have the unused outputs active for the two-player variant.

I've added skipping over 2nd player in the 2 player variant to make it consistent with how inputs are handled. bishi2p memory mapping also assigns 3rd player inputs to the 2nd controller. But indeed the output data consumer could handle this mapping as well.

PR #15274, opened 2026-04-27
- moved mg1 implementation into correctly named manufacturer folder

Elektraglide:


happppp:
This doesn't compile, you can see CI complain about wrong include guard. And you forgot to change mame.lst.
If these files are going to be moved, mg1_kbd.* is a better filename than kbd.*

Elektraglide:
> This doesn't compile, you can see CI complain about wrong include guard. And you forgot to change mame.lst. If these files are going to be moved, mg1_kbd.* is a better filename than kbd.*

Well they do compile,  they just fail the python checks..  But yes, will fix.

cuavas:
It's still a single-system project.  Are there other Whitechapel things that will end up there, or is this something that should go in a "bucket" project?

Elektraglide:
> It's still a single-system project. Are there other Whitechapel things that will end up there, or is this something that should go in a "bucket" project?

Yes, whitechapel made other workstations.  But mainly just after consistent naming for easier searching.

PR #15275, opened 2026-04-27
tecmo/gaiden.cpp: Add a new "Raiga - Strato Fighter" set

clawgrip:
New working clones
------------------
Raiga - Strato Fighter (Japan, set 2) [Tailsnic Retroworks]


cuavas:
How confident are we that these are legitimate different versions and one isn’t a bad dump or hack of the other?

PR #15277, opened 2026-04-27
apple2video: fix -ramsize crash

arekkusu42:
([another](https://github.com/mamedev/mame/pull/15252) config-specific crash)

This PR fixes apple2 and apple2p crashing with low-memory `-ramsize` configs

To reproduce:
1. `mame apple2p -ramsize 16k` <no disk>
2. Ctrl-Reset to prompt
3. HGR2

or
1. `mame apple2 -ramsize 12k` <no disk>
2. 2000<E000.FFFFM N C053:0 N C057:0 N C050:0

Result: OOB access, UB (MAME might crash, or show random garbage on screen.)
On my system a crash in a release build is rare without libgmalloc.
A debug ASAN build (`ARCHOPTS="-fsanitize=address"`) shows the problem immediately:
[apple2_12k_HGR_asan.txt](https://github.com/user-attachments/files/27143247/apple2_12k_HGR_asan.txt)

Expected Result:
The first apple2p case should show the screen entirely white.
The second apple2 case should show the screen as alternating rows of colored static (copied from E000) and white (only half of the HGR RAM is installed):
apple2_12k_HGR

[Hardware testing](https://apple2infinitum.slack.com/archives/CPS2TG28L/p1777228620951269) (thanks to @whscullin and Tom Greene) shows that empty RAM sockets read as 0xFF both from the CPU and the video scanner (probably because unconnected TTL inputs float high, similar to C06x gameio.)
<br>
Code comments:<details>

I made `a2_video_device::hgr_update()` check for HGR addresses exceeding the machine's `ram_size`, mirroring the logic in `apple2_state::ram_r()`.  This adds a usually-not-taken branch to every HGR line, for all machines.  An alternate implementation would be to pad apple2's `m_ram` allocation up to at least 24k, and then one-time-init any excess over the requested `-ramsize` to 0xFF.  That would be zero overhead in `a2_video_device`, but maybe more obfuscated, re-interpreting `m_ram_size` around a re-alloc.

MAME currently doesn't support the [non-contiguous 12K](https://apple2history.org/dl/Apple_II_Redbook.pdf#page=142) "HIRES" memory configuration.  To support that, range comparisons would need to change to something like a 4K page bitmap mask.
</details>

### DIFF
--- a/src/mame/apple/apple2.cpp
+++ b/src/mame/apple/apple2.cpp
@@ -315,7 +315,8 @@ void apple2_state::machine_start()
 	save_item(NAME(m_reset_latch));
 
 	// setup video pointers
-	m_video->set_ram_pointers(m_ram_ptr, m_ram_ptr);
+	m_video->set_ram_pointers(m_ram_ptr, nullptr);
+	m_video->set_ram_masks(m_ram_size-1, 0);
 	m_video->set_char_pointer(memregion("gfx1")->base(), memregion("gfx1")->bytes());
 }
 
--- a/src/mame/apple/apple2e.cpp
+++ b/src/mame/apple/apple2e.cpp
@@ -990,7 +990,7 @@ void apple2e_state::machine_start()
 
 	// setup video pointers
 	m_video->set_ram_pointers(m_ram_ptr, m_aux_ptr);
-	m_video->set_aux_mask(m_aux_mask);
+	m_video->set_ram_masks(0xffff, m_aux_mask);
 	m_video->set_char_pointer(memregion("gfx1")->base(), memregion("gfx1")->bytes());
 
 	// IOU events occur at the rightmost edge of active video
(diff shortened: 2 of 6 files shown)


arekkusu42:
RFC: I notice that `apple2video::m_aux_mask` (and now `m_ram_mask`) are not in the save state.  Does this matter?  MAME fails loading save state between two launches of the same machine with different `-ramsize` or` -aux` options, so it can't crash that way...

cuavas:
> RFC: I notice that `apple2video::m_aux_mask` (and now `m_ram_mask`) are not in the save state. Does this matter? MAME fails loading save state between two launches of the same machine with different `-ramsize` or` -aux` options, so it can't crash that way...

If they can’t change on-the-fly (i.e. they’re just set on start depending on configuration), they should not be in saved states.

arekkusu42:
Thanks, make sense.

PR #15283, opened 2026-04-29
ui: Adopted deemphasize flag in Sliders and Configuration menus

holub:


src/frontend/mame/ui/menu.cpp
```diff
@@ -884,12 +884,12 @@ void menu::draw(uint32_t flags)
...
 					if (!core_stricmp(pitem.subtext(), _("On")))
 						fgcolor2 = rgb_t(0x00,0xff,0x00);
-
 					if (!core_stricmp(pitem.subtext(), _("Off")))
 						fgcolor2 = rgb_t(0xff,0x00,0x00);
-
 					if (!core_stricmp(pitem.subtext(), _("Auto")))
 						fgcolor2 = rgb_t(0xff,0xff,0x00);
+					if (subitem_deemphasize)
+						fgcolor3 = fgcolor2.scale8(0.7F * 256); // 70%
```
cuavas: I’m going to back this part out, because it assumes a light-on-dark colour scheme, and makes an assumption about the relative brightness of the foreground and background colours.  This assumption isn’t safe as the user can reconfigure the colour scheme.

(Yes, there are a couple of pre-existing issues in that the red/orange/green “traffic light” background colours can’t be changed, which limits the usable text colours, and the magically applied colours for the words “On”, “Off” and “Auto” limit the usable background colours.  To be honest, I’m not a fan of the magically applied colours for those words.  It came when MEWUI was integrated.)

holub: This is exactly to serve that red/green cases. The alternative is for On/Off define fg2 and fg3 simultaneously.

cuavas: Yeah, but it can’t safely do that without considering whether the background is light or dark, and the actual intensity.  Simply scaling the brightness like this assumes that the background is dark and is substantially darker than 179 in 8-bit RGB. That isn’t a safe assumption, given the user can change the colours.  On a light background, you'd probably have to desaturate the colours as well as dimming them to prevent the green from looking muddy, and dimming yellow is perilous at the best of times due to human perception of browns being rather quirky.

holub: true

holub: Wait, we know background. I'll do "towards" background later

holub: Must be it: #15305

PR #15284, opened 2026-04-29
Document one more known good Ketsui dump in the wild.

Firehawke:
This one's already out there in the public so we might as well document it. I'm not opposed to leaving it commented out if people would rather wait.


mamehaze:
you'll probably want to format this with the 'new working systems' header, and proper credits, or it stands no chance of being merged.

Mushihime-Sama CAVEMATSURI was also dumped from a PCB, which I'm assuming can be added.

Firehawke:
I want no credits, I don't even want this highlighted. This should be low-key. As for the Mushi, I haven't yet seen the dump or I'd add that in here too.

happppp:
Then explicitly add [anonymous]

cuavas:
See 936800b17c1b4885879282f083f131587907c262

PR #15286, opened 2026-04-29
Fix mem-mem MOV on i8089

bytex64:
The base register for memory operations was being fixed up to handle the PP register. But memory-to-memory MOVs act like two instructions, and the second half was being decoded ad-hoc without that fixup. This adds that fixup to the second half.


bytex64:
For more background, the thread where I found the bug and validation against hardware: https://awesome.garden/@bytex64/116485088560812263

startaq:
Thanks for spotting this!

PR #15287, opened 2026-04-29
pacman: Fix mschamp timer selection and I/O mux behavior

Liron486:
## Summary

This fix the `mschamp` / `mschamps` driver for **Ms. Pacman Champion Edition / Super Zola Pac Gal**.

This is related to MAME Testers issue 05446:
https://mametesters.org/view.php?id=5446

That report describes two main problems:

- Super Zola Pac Gal speed sometimes changes incorrectly/randomly after a new level or after losing a life.
- The timer setting is missing, with MAME effectively using a fixed 60-second timer.

## What changed

The previous implementation used `mschamp_kludge_r()` to return an incrementing value from I/O port `$00`:

`return m_counter++;`

That made reads from this port depend on call order rather than on a stable emulated hardware state.

This fix replaces that behavior with a small stateful mux model for the observed `$10/$11` write and `$00` read sequence:

- writes to I/O ports `$10-$11` update the selected state/data
- reads from I/O port `$00` return a deterministic value based on that state
- `$10`-selected reads return the timer selector bits
- `$11`-selected reads return the previously written data, matching the handshake-like behavior used by the game

## Timer selection

The ROM uses the low two bits of the value read through this I/O path to select the timer preset.

This exposes the four ROM-supported timer values as a Timer DIP setting:

- `1:30`
- `1:00`
- `0:45`
- `0:30`

The old incrementing return value usually resulted in the `1:00` timer preset, but that was an artifact of the previous kludge rather than an explicit emulated setting.

## Input label

Super Zola Pac Gal uses the inherited cocktail P2 Left input line as Turbo, so this patch labels that shared input as `P2 Left / Turbo`.

This preserves the original cocktail P2 Left mapping while documenting the additional Super Zola function.

## License

I am contributing this change under the BSD-3-Clause license.

### DIFF
--- a/src/mame/pacman/pacman.cpp
+++ b/src/mame/pacman/pacman.cpp
@@ -377,7 +377,19 @@ Boards:
  *
  *************************************/
 
-MACHINE_RESET_MEMBER(pacman_state,mschamp)
+MACHINE_RESET_MEMBER(mschamp_state,mschamp)
+{
+	uint8_t *rom = memregion("maincpu")->base() + 0x10000;
+	int whichbank = ioport("GAME")->read() & 1;
+
+	membank("bank1")->configure_entries(0, 2, &rom[0x0000], 0x8000);
+	membank("bank2")->configure_entries(0, 2, &rom[0x4000], 0x8000);
+
+	membank("bank1")->set_entry(whichbank);
+	membank("bank2")->set_entry(whichbank);
+}
+
+MACHINE_RESET_MEMBER(pacman_state,crush4)
 {
 	uint8_t *rom = memregion("maincpu")->base() + 0x10000;
 	int whichbank = ioport("GAME")->read() & 1;
@@ -734,25 +746,18 @@ uint8_t pacman_state::mbrush_prot_r(offs_t offset)
  *
  *************************************/
 
-uint8_t pacman_state::mschamp_mux_r()
+uint8_t mschamp_state::mux_r()
 {
-	switch (m_mschamp_mux)
-	{
-	case 0x10:
-		return 0x40 | (ioport("TIMER")->read() & 0x03);
-
-	case 0x11:
-		return m_mschamp_mux_data;
-
-	default:
-		return 0xff;
-	}
+	if (m_mux)
+		return m_mux_data;
+	else
+		return 0x40 | (m_timer->read() & 0x03);
 }
 
-void pacman_state::mschamp_mux_w(offs_t offset, uint8_t data)
+void mschamp_state::mux_w(offs_t offset, uint8_t data)
 {
-	m_mschamp_mux = 0x10 + offset;
-	m_mschamp_mux_data = data;
+	m_mux = offset & 1;
+	m_mux_data = data;
 }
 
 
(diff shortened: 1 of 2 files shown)


happppp:
Hi,

Assuming the mux is a 1-bit latch, it can be 0 or 1, not 0x10 or 0x11, and no indeterminate 0xff state.
You don't need to reinitialize the variables in the init function when you've already done it in the class variable declarations.

When a driver deviates too much(subjective) from the main driver, we'd prefer a subclass, see for example clubpacm_state in pacman.h, then you can also use required_ioport for TIMER.

Did you use AI?

Liron486:
Hi,

Thanks for the review and the helpful comments.

I updated the patch according to your feedback:

- The mux state is now treated as a 1-bit latch, using 0/1 from the write offset instead of $10/$11.
- The indeterminate $ff mux state was removed.
- init_mschamp() now only registers the save-state items and does not reinitialize values already initialized in the class.
- The mschamp-specific I/O handling was moved into a dedicated mschamp_state subclass.
- TIMER is now handled with required_ioport.

While doing that, I found that crush4 was previously reusing mschamp(config). Since the new mschamp_state now contains mschamp-specific I/O behavior, I split crush4 away from that dependency. crush4 now has its own map/reset setup for the banked code path it needs, and it no longer uses the mschamp machine configuration.

Regarding AI: yes, I used ChatGPT for some assistance while investigating and working through the patch, but I made the changes myself, tested them locally, and reviewed the code before pushing.

happppp:
Thanks, I'll have a look at cleaning up the crush4 part after merging, I don't think it uses the memory banking.

happppp:
@Liron486 what is the 0x40 for in ```return 0x40 | (m_timer->read() & 0x03);```?

Liron486:
@happppp In my local debugging/tests, I checked the values returned on this path and noticed that the working/default timer case was not just returning the raw low timer bits. It was effectively returning the timer selector with bit 6 set, e.g. 0x41 for the 1:00 case.

The ROM only appears to use the low two bits for the timer selection, so:

    m_timer->read() & 0x03

selects the timer value. The 0x40 preserves the upper-bit pattern I observed during testing, while replacing only bits 0-1 with the deterministic TIMER DIP value.

So this returns:

    0x40 = 1:30
    0x41 = 1:00
    0x42 = 0:45
    0x43 = 0:30

I kept bit 6 set because that matched the behavior I observed in testing and avoided changing the rest of the returned byte to 0x00-0x03.

happppp:
Ah, you mean that the Z80 A register bit 6 was set before it reads from this port.
It doesn't make sense (from hardware POV) for bit 6 to remain unchanged.

Liron486:
Right, I did not mean that the Z80 A register preserves bit 6 across the port read. That would not make sense.

What I meant is that while debugging the existing MAME behavior, the value returned on this path for the default/working timer case was effectively 0x41, so I kept bit 6 set when replacing the old incrementing return value with the TIMER DIP bits.

I can remove the 0x40 if you prefer. My goal was just to preserve the observed previous behavior while making the timer selection deterministic.

happppp:
The existing MAME behaviour was to return "counter++", which can be anything. I don't see how you got 0x41 here, unless you debugged it at a specific point after it had read it 0x40 times.

No need for a follow up PR, I already removed the 0x40.

Issue #14999, opened 2026-02-21
apple2: regression with charrom shifting


arekkusu42:
After building the changes for the megaii.chr replacement @37ed312, I noticed that the appleiigs font looks wrong:
<img width="704" height="462" alt="Image" src="https://github.com/user-attachments/assets/d96fb837-c74d-4474-a812-bf5422f76ca8" />

Looking at this, it's due to a2_video_device::get_text_character() [shifting over the ROM data](https://github.com/mamedev/mame/blob/7e974b41ece31fc2bee38b6d862a841ea061a9f4/src/mame/apple/apple2video.cpp#L477).  This was changed in f8ecff4 for ivelultr.

That change appears to also have introduced a regression with all the laser128 clones, and maybe albert and dodo.  I compared with an old 0.282 build I have:

laser128_282_good (this is before other beam-racing fixes, just look at the character alignment):
<img width="560" height="384" alt="Image" src="https://github.com/user-attachments/assets/053c21ad-7bb7-40d7-b227-882c67f7d015" />

laser_128_285_bad (all characters shifted left one pixel, tips of mousetext arrows cut off):
<img width="560" height="384" alt="Image" src="https://github.com/user-attachments/assets/eb01eb23-49a9-4823-a049-1fec2d8451bc" />

The same thing is happening now to apple2gs.

arekkusu42:
@ajrhacker @rb6502 please review.

rb6502:
I know AJR is having difficulties with his laptop and release lockdown is imminent so I committed a quick fix that confines the shift behavior to the Ivel Ultra once again.

arekkusu42:
Thanks, I verified @ 2be341f that apple2gs and laser128 look OK now.

It looks like albert and dodo still have problems:
albert_285fix_bad: (this doesn't happen in 282. albert now complains it is missing keyboard.bin, but isn't marked NOT WORKING?)
<img width="560" height="384" alt="Image" src="https://github.com/user-attachments/assets/50b9984c-e3f1-4b11-bbb5-ba893d4ffc48" />

dodo_285fix_bad: (this appears to need the shift, so it's now cut off on the right! it worked in 282)
<img width="560" height="384" alt="Image" src="https://github.com/user-attachments/assets/5ee0de5c-6bf4-426c-96f0-308d3a2bd7f6" />

ivelultr appears to boot ProDOS but hangs running BASIC.SYSTEM?

arekkusu42:
This looks improved after 45a0899:

[...]

PR #15290, opened 2026-04-30
apple2video: fix ivelultr/dodo/laser2c text rendering

arekkusu42:
(followup #14999)

This PR fixes text rendering on ivelultr/dodo/laser2c: these clones display text characters shifted one pixel too far to the right, which is mostly visible with INVERSE text:
ivelultr_ASCII_0287 dodo_laser2c_0287

After this PR, text rendering aligns with other Apples and clones:
ivelultr_ASCII_0287_after dodo_laser2c_0287_after

I don't have direct experience with these clones, but it also matches the references I was able to find online:
* [This video](https://youtu.be/SKMQlrjuG08?t=55) (using ASCII not YU, so C with caron "č" appears as "~") shows inverse "D" and "_" character spacing on IVEL ULTRA hardware
* The original thread dumping dodo ("GTAC Wiederbelebung" "Do-Do") contains a photo showing the cursor spacing on hardware
* [This video](https://www.youtube.com/watch?v=u33FhdRC3kY&t=413s) shows cursor spacing on Laser ][c hardware

ivelultr_ASCII_reference dodo_reference laser2c_reference
<br>
This PR also cleans up the `screen_update()` and `text_update()` template instantiation to match what is actually used.

<br>
Code comments:<details>

Auditing the various apple2 character ROMs shows there are three classes of text bitmap layout (plus inversion):

1. **flip**:		all apple2 EXCEPT ivelultr/dodo/laser2c, and spectred/ace500/2200
2. **shift & flip**:	ivelultr/dodo/laser2c
3. **no flip**:		all apple2e EXCEPT spectred/ace500/2200, and apple2gs

The **shift** case was [briefly handled](https://github.com/mamedev/mame/blob/f8ecff4d11fca5000d5d581995d5341c32cb4440/src/mame/apple/apple2video.cpp#L477) by an `Invert, Flip` template combination, but this broke [when apple2gs changed](https://github.com/mamedev/mame/issues/14999) to use `Invert`, so is now handled with explicit `a2_video_device::model`s.  That's fine.  But the shift by `>> 1` wasn't enough.

`>> 2` works for all characters except 0xFF which fills all 7 bits in ivelultr (ASCII page) and dodo ROMs.  I don't know how IVEL ULTRA hardware implements this, but based on the [reverse engineering of a different GTAC clone](https://github.com/pfiliberti/gtac-appleclone/blob/main/fontrom/gtac_fontrom_info.txt), it's likely that the chargen shifters keep all 7 bits and just swizzle the order.  So this code rotates bit 1 into bit 6, such that character 0xFF properly fills all seven pixels.

<br>
For the instantiation cleanup, there are four fixes:

1. `<II, true, false>` is now replaced by `<DODO, true, false>`
2. `<II, false, false>` was only used by superga2 and tk2000, which have no text modes, so make them match apple2p and delete the unused instantiation
3. `<IVEL_ULTRA, false, true>` was briefly added for albert, but is now unused
4. make `text_update()` match `screen_update()`

Removing unused instantiations shrinks a clang arm64 release (O3, no LTO) binary by about 4kB.

</details>
<br>
TODO:<details>
I noticed a few other things while auditing:

* It's unclear why [craft2p is marked IMPERFECT GRAPHICS](https://github.com/mamedev/mame/blob/1e5263494af307699e72b463ce8ee180684c1c5c/src/mame/apple/apple2.cpp#L1232).  The char ROM has two language pages, but they are both Portuguese.  Is it dumped incorrectly?
* ivelultr is [commented as "wider character cell"](https://github.com/mamedev/mame/blob/1e5263494af307699e72b463ce8ee180684c1c5c/src/mame/apple/apple2.cpp#L1233) but the video hardware uses regular 7x8 cells, only the alignment of the ROM bitmap differs, which this PR fixes
* pravetz8m_chr has four language pages.  There is an [unemulated dip switch](https://retroroundup.com/2020/02/04/pravetz-8m/#:~:text=different%C2%A0%20keyboard%20layouts) but I haven't found documentation on it.
* It's unclear how to toggle language pages on tk3000 or laser128*
</details>

### DIFF
--- a/src/mame/apple/apple2video.cpp
+++ b/src/mame/apple/apple2video.cpp
@@ -655,7 +655,7 @@ unsigned a2_video_device::get_text_character(uint32_t code, int row)
 
 	/* look up the character data */
 	unsigned bits = m_char_ptr[code * 8 + row];
-	bits = ((Model == model::IVEL_ULTRA && Invert && !Flip) || Model == model::DODO) ? (bits >> 1) : (bits & 0x7f);
+	bits = (Model == model::IVEL_ULTRA || Model == model::DODO) ? ((bits >> 2) | ((bits & 2) << 5)) : (bits & 0x7f);
 	bits ^= invert_mask;
 	return Flip ? reverse_7_bits[bits] : bits;
 }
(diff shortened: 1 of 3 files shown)


ajrhacker:
> It's unclear how to toggle language pages on tk3000

On tk3000, that can be done by pressing the Mode key. The effect can be seen in this video of actual HW: https://www.youtube.com/watch?v=ZnpTV49uizc

> It's unclear why craft2p is marked IMPERFECT GRAPHICS

I just checked again, and the way flashing/inverse characters are handled looks wrong for this system.

arekkusu42:
Thanks!  I had seen the "mode" key in the tk3000 manual, but now I also see that it is mapped to F11.  Works fine.

craft2p: OK, the references I could find only showed normal text.

PR #15295, opened 2026-05-01
Redoing offs_t to 64-bits expansion.

MooglyGuy:
This revises offs_t from 32 bits to 64 bits in width. It is a re-implementation of the previous draft PR using Vas's suggested best practices, and is more limited in scope due to prior cleanup.


galibert:
I guess I need to fix the code style guide so that people stop shitting on the code I write.

MooglyGuy:
If you wrote unidasm, your username should be in the `copyright-holders` field. The style guide rather explicitly says to preserve the style that's already present in a file, but it should be equally obvious that if a file is effectively ownerless then it's better to clean it up. Given that the only listed copyright holder on unidasm.cpp is Aaron, and he no longer contributes, I really don't want to hear it about me cleaning up that inconsistent crapfest of a file.

PR #15296, opened 2026-05-02
cinematronics/leland*.cpp: Updates

cam900:
- Fix namings for consistency
- Add notes for video RAM size and accessing (match as hardware configuration - 4 64Kx4 bit chips populated in hardware; double buffered?)
- Reduce literal tag usages
- Fix typename for save state support
- Fix spacings
- Use memory_view for palette RAM and battery backed RAM accessing
- Simplify main CPU bankswitch method
- Reduce duplicates
- Use hexadecimal masking value
- Use BIT helper for single bit value
- Reduce preprocessor defines
- Correct XROM data bus to 16 bit for simplicity
- Use reference rather than pointer


src/mame/cinematronics/leland.cpp
```diff
@@ -70,7 +73,7 @@ PART # 80-18002-01
 |-----------------------------------------------|
 Notes:
       80186 - Siemens SAB80186-1-N x86 CPU. Clock input 16MHz, runs at 8MHz with an internal /2 divider.
-        Z80 - Master(1) & Slave(2) Zilog Z0840006 CPU. Clock input 6.000MHz [both, 12/2]
+        Z80 - Main(1) & Sub(2) Zilog Z0840006 CPU. Clock input 6.000MHz [both, 12/2]
```
cuavas: Leland used “master CPU” and “slave CPU” terminology in their schematics and manuals, so this is retained in MAME.

See for example the Viper manual: https://archive.org/details/ArcadeGameManualViper/page/n47/mode/2up

The “master CPU” section begins on page 44, and the “slave CPU” section begins on page 48.

PR #9414, opened 2022-03-14
vamphalf.cpp: preliminary handling for misncrft protection checks

angelosa:

PR #15302, opened 2026-05-03
misc/vamphalf.cpp: Updates

cam900:
- Add korean game titles if exists (source: Title screen, POST screen in finalgdr, Background text pattern in poosho)
- Reduce duplicates
- Fix save state support
- Fix initialization
- Reduce literal tag usages
- Fix spacings
- Add notes

### DIFF
--- a/src/mame/misc/vamphalf.cpp
+++ b/src/mame/misc/vamphalf.cpp
@@ -84,18 +84,24 @@ public:
 	vamphalf_state(const machine_config &mconfig, device_type type, const char *tag)
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
-		, m_wram(*this,"wram")
-		, m_wram32(*this,"wram32")
-		, m_qs1000_bank(*this, "qs1000_bank")
-		, m_okibank(*this,"okibank")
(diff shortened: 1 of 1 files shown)


src/mame/misc/vamphalf.cpp
```diff
@@ -334,30 +340,23 @@ void vamphalf_state::flipscreen_w(u16 data)
...
 
-
-u32 vamphalf_qdsp_state::wyvernwg_prot_r()
+template <u32 SetValue>
+u32 vamphalf_state::prot_r()
 {
 	if (!machine().side_effects_disabled())
 		m_semicom_prot_idx--;
-	return (m_semicom_prot_data[m_semicom_prot_which] & (1 << m_semicom_prot_idx)) >> m_semicom_prot_idx;
+	return BIT(m_semicom_prot_data[m_semicom_prot_which], m_semicom_prot_idx) ? SetValue : 0;
```
angelosa, line 351: fwiw I plan to change `misncrft`, `wivernwg` (and I guess `worldadv`) to actual composed devices, rather than having to separate state machines (which is a separate matter, this driver sure needs splits).
cfr. https://github.com/mamedev/mame/pull/9414

PR #15306, opened 2026-05-04
sequential/prophet5.cpp: Added handcrafted factory patch nvram.

m1macrophage:
This is based on the diagrams in the owner's manual. The script that generated the nvram, along with the generated file can be found [here](https://github.com/m1macrophage/synthesizerstate/tree/main/prophet5).

Also, shortened the driver name while at it. But let me know if this is not preferred.

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -43145,7 +43145,7 @@ wizt
 wizta
 
 @source:sequential/prophet5.cpp
-prophet5rev30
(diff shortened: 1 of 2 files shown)


galibert:
Honestly I don't like it but aren't we supposed to mark it BAD_DUMP in such a case?

m1macrophage:
Sure, done.

PR #15310, opened 2026-05-05
frontend/mameopts.cpp: Fall back to settings from INI files later in the search path

cuavas:
This is a non-trivial change, so I’m opening a pull request.  The main immediately visible effect will be on MAME distributions that provide system-wide defaults in /etc/mame/mame.ini or similar.

Right now, MAME will stop looking for INI files as soon as it finds one with the name it’s looking for.  For example on Linux, the default `inipath` setting is `$HOME/.mame;.;ini` so if MAME finds a mame.ini in `~/.mame`, it won’t look for one in the working directory at all.

With this change, MAME will read mame.ini (or whatever INI file it’s looking for) for all the directories in the search path.  Settings in files earlier in the search path will override settings in files later in the search path.

For a practical example, supposed the `inipath` setting is set to `.;$HOME/.mame`.

Suppose mame.ini in the working directory contains:
```
window 1
```

and suppose mame.ini in `~/.mame` contains:
```
window 0
confirm_quit 1
```

Right now, MAME will stop after parsing the first file, so the effective options will be `-window`.  With this change, the effective options will be `-window -confirm_quit` (the `window` setting from the first INI file will take precedence, but the `confirm_quit` setting from the second INI file will take effect).

This won’t affect people who use `-createconfig` or save settings from the internal UI, as that writes out *all* the settings, not just settings you change.  It only affects people who create INI files with a subset of settings and have multiple similarly named INI files in the search path.

@rb6502, @galibert and @belegdol, what are your opinions on making this change?

### DIFF
--- a/src/frontend/mame/mameopts.cpp
+++ b/src/frontend/mame/mameopts.cpp
@@ -50,6 +50,7 @@ void mame_options::parse_standard_inis(emu_options &options, std::ostream &error
 			}
 			catch (options_exception const &ex)
 			{
+				util::stream_format(error_stream, "While parsing %s:\n%s\n", file.fullpath(), ex.message());
 			}
 		}
 	}
--- a/src/frontend/mame/mameopts.cpp
+++ b/src/frontend/mame/mameopts.cpp
@@ -36,9 +36,25 @@
 
 void mame_options::parse_standard_inis(emu_options &options, std::ostream &error_stream, const game_driver *driver)
 {
-	// parse the INI file defined by the platform (e.g., "mame.ini")
-	// we do this twice so that the first file can change the INI path
-	parse_one_ini(options, emulator_info::get_configname(), OPTION_PRIORITY_MAME_INI);
+	// parse exactly one of the main INI file to pick up and apply inipath and [no]readconfig
+	{
+		emu_file file(options.ini_path(), OPEN_FLAG_READ);
+		osd_printf_verbose("Attempting load of %s.ini\n", emulator_info::get_configname());
+		std::error_condition const filerr = file.open(std::string(emulator_info::get_configname()) + ".ini");
+		if (!filerr)
+		{
+			osd_printf_verbose("Parsing %s\n", file.fullpath());
+			try
+			{
+				options.parse_ini_file(static_cast<util::core_file &>(file), OPTION_PRIORITY_MAME_INI, true, false);
+			}
+			catch (options_exception const &ex)
+			{
+			}
+		}
+	}
+
+	// now parse the main INI file following the potentially updated search path
 	parse_one_ini(options, emulator_info::get_configname(), OPTION_PRIORITY_MAME_INI, &error_stream);
 
 	// debug mode: parse "debug.ini" as well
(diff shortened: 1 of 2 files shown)


belegdol:
Thanks for this! I think this would make shipping distribution defaults easier. Right now, one has to replicate every setting in a per-user ini. With this change, only deviations from the system-wide setting would need to be put into the per-user file.
On the flip side, this PR makes it considerably simpler to end up with multiple, potentially conflicting ini files, with no simple way of knowing which one is being used. Have a look at [1] for where this can lead even with current logic. This could be possibly mitigated by something similar to `git config  --show-origin foo`.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2454163

cuavas:
Oh, that’s a stupid thing in the current code – the verbose output tells you it’s parsing “mame.ini”, but doesn’t tell you which one it picked up.  This change actually gives the path it’s opening from in the verbose output for each INI file it opens.  (I could change the code to just improve the verbose output to help with debugging, but that would conflict with this PR now.)

galibert:
I'm (obviously) never going to use a version of mame coming from a distribution, so there's only one mame.ini in my path in the first place.  Still seems like a decent idea.  Maybe a mame -findconfig window could tell you all "window" configuration lines found in ini files and which one wins.  That could help debugging on r/MAME and similar.

belegdol:
Apologies for not testing this sooner - having two small children certainly makes finding the time slightly more challenging.
As things currently stand in master, this breaks rompath. Fedora package was set up as follows until now:

1. package was built with `SDL_INI_PATH="/etc/mame;"`
2. `/etc/mame/mame.ini` contains `inipath $XDG_CONFIG_HOME/mame;$HOME/.config/mame;$HOME/.mame/ini;/etc/mame` and `rompath /usr/share/mame/roms;/usr/share/mame/chds;$XDG_DATA_HOME/mame/roms;$HOME/.local/share/mame/roms;$XDG_DATA_HOME/mame/chds;$HOME/.local/share/mame/chds`
3. user can define their individual `rompath` in a `mame.ini` located in their home directory (just one, the multiple paths are there for compatibility with existing set-up.

Looks like that due to settings earlier in the path overriding ones later in the path, only the `rompath` from `/etc/mame/mame.ini` is being considered. Having said that, `inipath` in `mame.ini` already has `/etc/mame/mame.ini` at the end, so it looks like it is the first pass caused by the `SDL_INI_PATH` which has already read the `rompath` settings.

Is it the intended behaviour? I tried restoring the previous behaviour by compiling with `SDL_INI_PATH=$$XDG_CONFIG_HOME/mame;$$HOME/.config/mame;$$HOME/.mame/ini;/etc/mame` but while it ends up in generated makefiles as `-D'INI_PATH="$XDG_CONFIG_HOME/mame;$HOME/.config/mame;$HOME/.mame/ini;/etc/mame"'`, only `/etc/mame/mame.ini` gets parsed in the first pass:

```
$ ./mame -verbose  -bgfx_path bgfx
Attempting load of mame.ini
Parsing /etc/mame/mame.ini
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Attempting load of mame.ini
Parsing /home/julas/.mame/ini/mame.ini
Parsing /etc/mame/mame.ini
Loading translation file /usr/share/mame/language/English/strings.mo
Reading translation file: 1 strings, original table at word offset 7, translated table at word offset 9
Loaded 0 translated string from file
Starting plugin data...
Starting plugin layout...
Attempting load of mame.ini
Parsing /etc/mame/mame.ini
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_STATE_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_CONFIG_HOME not found.
Warning: osd_subst_env variable XDG_DATA_HOME not found.
Attempting load of mame.ini
Parsing /home/julas/.mame/ini/mame.ini
Parsing /etc/mame/mame.ini
Attempting load of source/empty.ini
Attempting load of ___empty.ini
```

belegdol:
Turns out four dollar signs are needed: `SDL_INI_PATH=$$$$XDG_CONFIG_HOME/mame;$$$$HOME/.config/mame;$$$$HOME/.mame/ini;/etc/mame` - with this parameter `~/.mame/ini/mame.ini` is read correctly.
Still, doesn't the new behaviour make `inipath` setting somewhat useless? It appears that it only works for settings not defined in the ini file found first in the ini stored in SDL_INI_PATH.

wallyweek:
@belegdol I had the same issue in Ubuntu.

This is due to the priority assigned to the first read of file in SDL_INI_PATH which doesn't allow later overrides - please see my PR here: https://github.com/mamedev/mame/pull/15443

PR #15311, opened 2026-05-06
added a handcrafted 'protection' data ROM to goretrop sets

mamehaze:


mamehaze:
@ajrhacker you might want to look at this (as I saw you're in the process of shuffling the code around and I don't want to step on any toes)

the handcrafted ROM probably isn't perfect, but I think they probably made these out of some template code and tweaked it to perform different reads/writes (as well as relocating it from 0xe00 on some hardware to 0x700 here)

the calls to the data from ROM are at the same offsets as gtct885 other than being at 0x7xx instead of 0xexx, suggesting the same basic code structure, just expecting slightly different behaviors.  the fact that the startup code in goretrop patches over some of the code (NOPing it out) after copying it from the device suggests it might not have been specifically be made for it too.

having the menu running makes it easier to hack the game to get the real data out at some point though (I have a spare unit)

it also has a 16Mbyte bank select on extraout_23_w bit 0x20 (although many games don't work as they need extra mapper support this SoC has or something)

PR #15313, opened 2026-05-07
apple2common: debugger improvements

arekkusu42:
This PR collects a series of debugger correctness and performance fixes for Apple II machines:
* apple2e/apple2gs: prevent various debugger side-effects
* apple2gs: make C0xx I/O visible to the debugger
* apple2common: add/correct a few dasm symbols
* apple2common: add dasm for some missing opcodes
* apple2common: perf: skip slow JSL dasm indirection
* apple2common: perf: binary search dasm symbols
<br>
Code comments:<details>

The motivation here was:
* the debugger broke machines if a Memory Window was watching Cxxx
* the debugger didn't show apple2gs C02E/2F beam position counters
* the debugger didn't show symbols correctly stepping through beam-racing code
* the debugger performance overhead was very high

Some of the side-effect breakage, like the Zip acceleration repeatedly switching off, was subtle.  But others were more serious: expansion RAM would corrupt (the [C40AG self-test](https://www.savagetaylor.com/TIL/TIL02037.pdf) failed), or a [ComputerEyes capture](https://wiki.mamedev.org/index.php/Driver:Apple_II#:~:text=ComputerEyes%20video%20digitizer) would be skewed and distorted.
<br>

apple2gs now shows all the C0xx I/O, except for the SCC registers (the `z80scc` device [isn't debugger-safe](https://github.com/mamedev/mame/blob/a3c4004a05f150b0e6323e53600d0603b488b577/src/devices/machine/z80scc.cpp#L2398).)  The changes are similar to [earlier ones for apple2e](https://github.com/mamedev/mame/commit/f49f4fa4eb90903ece4b3337e98ab272aa67e025#diff-756176cb1807e85dc56a5c125769d51f75f1c0662741aec56abc69ebc896ebcf), with a little more complication for keyboard GLU etc. 
<br>

Adding the missing opcodes for simple addressing modes is a partial solution; some of these (ASL, ROR) may be rarely used in practice, but the commonly used STA/X/Y LDA/X/Y are all there now.  This could be expanded to indexed etc modes (like [the C64 disassembler](https://github.com/mamedev/mame/blob/8945f05060c1ea53fb2c56bc8408242c42c99b92/src/mame/commodore/c64.cpp#L283)), but that would effectively become a wholesale replacement of the CPU core disassemblers.  The current `didisasm` [delegate design](https://github.com/mamedev/mame/blob/a3c4004a05f150b0e6323e53600d0603b488b577/src/emu/didisasm.cpp#L39) makes this cumbersome: delegates can not merely symbolicate an address, they are responsible for the entire instruction (mnemonic, addressing mode etc) stringification.  

Perhaps a better API design could be modeled on the [tlcs900 symbolication](https://github.com/mamedev/mame/blob/5cf0c66973045181ff099cd4b4ca670d88486949/src/devices/cpu/tlcs900/dasm900.cpp#L1519)?  Each CPU core disassembler could pass address lookups to delegates, which would still need to be passed the opcode (for specialization) but would only be responsible for stringifying the address...
<br>

For the performance, here's an example user scenario, examining 16-bit toolbox ROM:

1) `mame apple2gs -debug`
2) in a New Disassembly Window, examine FE8000
3) scroll the disassembly up and down, observing performance

Before this PR in 0.287, this performs very poorly.  On an M3 Max MacBook Pro (which I think in 2026 can still be considered "very fast" hardware), in a release -O3 build, this scrolls at a few frames per second, consuming 100% of one CPU core.  In a coarse high-level analysis, Instruments says all the time is in `debug_data_buffer::fill()`.

It [has been noted](https://github.com/mamedev/mame/commit/7c4db90e25cf4b87a9e2c75c8e99b03d3710bdbd) that MAME's `debug_data_buffer::fill()` is inefficient when two debug addresses are far apart.  In the case of [JSL to an inline debug address](https://github.com/mamedev/mame/blob/a3c4004a05f150b0e6323e53600d0603b488b577/src/mame/apple/apple2common.cpp#L827), the operand indirection can result in fetching two addresses multiple banks apart, so each JSL disassembly can read hundreds of kB of buffer (and then re-read it again each time the window scrolls, even if the machine execution is paused.)  The fix here is a "don't jump into that pit" band-aid.

After skipping that JSL indirection, the performance is much better and scrolls smoothly, but there is still a few percent spent in the `apple2_common_device::dasm_override()` leaf functions.  There are thousands of symbols being subjected to repeated linear search, so sorting them and switching to binary search (similar to [the tlcs900 disasm](https://github.com/mamedev/mame/blob/5cf0c66973045181ff099cd4b4ca670d88486949/src/devices/cpu/tlcs900/dasm900.cpp#L1418)) is a simple algorithmic improvement.  `com_2byte_op()` is left linear since it requires a partial search, but it looks at fewer than 64 symbols.

</details>
<br>

There should be no visible change without `-debug`.

With `-debug`, disassembly which previously failed to show symbolicated addresses:
disasm_gs_287
...now works in more cases:
disasm_gs_287_after

...and performs better.  A trivial example:
```
]CALL-151
*!
!300:LDA #F0
! INC C034
! TRB C034
! BRA 302
!
*300G

```
In 0.287 before this PR, with `-debug` (disassembling the current PC) this runs very erratically,  pausing every second.  After this PR it runs smoothly, similar to without `-debug`.

</details>
<br>
TODO:<details>

For this PR I audited the apple2 machines and a2gameio.  I did not audit a2bus; there could be additional side-effects there.

laser128 maps `applefdintf_device::read()` [directly to C0Ex](https://github.com/mamedev/mame/blob/a3c4004a05f150b0e6323e53600d0603b488b577/src/mame/apple/apple2e.cpp#L3498), this has the unfortunate side-effect of breaking floppy I/O if the debugger simultaneously reads this range.  I think this could be trivially fixed in `iwm_device::read()`, but tracking that in a separate issue.

Except for that laser128 issue, I think all of the read handlers are now debugger-safe.  I did not touch the write handlers; there are various inconsistent uses of `side_effects_disabled()` which look like copy-paste, and could probably all be deleted.

This PR fixes `accel_slot()` to work in the debugger, but recent Zip Chip hardware testing shows that most of the call sites of this are actually wrong.  I plan to address this in a separate Zip PR.

</details>


src/mame/apple/apple2gs.cpp
```diff
@@ -1550,13 +1550,36 @@ u8 apple2gs_state::c000_r(offs_t offset)
...
+			case 0x39:  // SCCAREG
+				return m_scc->ca_r(0);
+
+			case 0x3a:  // SCCBDATA
+				return m_scc->db_r(0);
+
+			case 0x3b:  // SCCADATA
+				return m_scc->da_r(0);
+		}
 	}
```
cuavas, line 1581: What’s the purpose of this?  It looks like it will do something completely different for these offsets if side effects are disabled, which will mean it won’t display the data that would be read.  Is the SCC itself not suppressing side effects properly?  If that’s the case, it should be fixed.

arekkusu42: The `z80scc` device is not debugger-safe, as I said in the Code Comments above.

The branch here means these fall through to `do_io()`, and then return `uFloatingBus` like they did before this PR.

Yes, the SCC should be fixed, like [the IWM](https://github.com/mamedev/mame/issues/15315).  They both look non-trivial to fix in a way that returns meaningful data without perturbing FIFO state, etc.  But could be trivially fixed to punt zero under the debugger, so at least machines wouldn't drop serial data or corrupt disks merely by using a Memory Window.

PR #15317, opened 2026-05-08
Implement Alphasmart 3000 more fully

isotherm:
* Implement keyboard matrix and interrupt handling.
* Configure display driver to display screen properly.
* Remove list of smart applets. A facility like this could potentially be restored after it is possible to load smart applets dynamically. Instead, load stock selection of smart applets automatically.

### DIFF
--- a/hash/alphasmart_applets.xml
+++ b/hash/alphasmart_applets.xml
@@ -4,7 +4,7 @@
 license:CC0-1.0
 -->
 
-<softwarelist name="alphasmart_kapps" description="AlphaSmart 3000 SmartApplets">
+<softwarelist name="alphasmart_applets" description="AlphaSmart SmartApplets">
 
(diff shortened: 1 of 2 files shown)


rb6502:
I don't have a lot of context on this system, but I don't think deleting the hash file makes sense, especially if there's a plan to bring it back later.  How were these applets distributed and loaded?

isotherm:
> I don't have a lot of context on this system, but I don't think deleting the hash file makes sense, especially if there's a plan to bring it back later. How were these applets distributed and loaded?

The applets were distributed as files and loaded via USB via a Windows program. (There are now open source methods to load applets, but USB emulation here would be a ways off.) The current plan would be to load the applets directly into RAM as cartridges or similar.

These hashes are a bit esoteric (from one less common distribution - UK edition, and not latest version). However, I think they could be left for future use. They won't have any effect until there is a way to load the applets.

PR #15318, opened 2026-05-08
Minor Update hw info for tzwang.

flama12333:
the mhz was info was not mentioned. only was guessed.


ajrhacker:
Not enough confidence in this, I think.

PR #15320, opened 2026-05-09
namcos2.cpp: Fix Shared data ROMs from the new finalap2jb set for finalap2j [sjy96525]

sjy555525:
The Japanese Versions is used fls1_dat0.13s and fls1_dat1.13p Shared data ROMs.


ajrhacker:
@devinacker, can you confirm this? (Ref.: #15319)

Never mind, it's not for the newly dumped clone set. I'll merge this.

PR #15321, opened 2026-05-09
suprnova.cpp: cyvern has button 3 on real hardware

RiZeuM:
Button 3 is functional on real Cyvern hardware for both players.
Verified by reading register 0x00400000 on real PCB.
Bits 0x40000000 (P1) and 0x00400000 (P2) are active.
Removing IPT_UNUSED override from cyvern input ports.

### DIFF
--- a/src/mame/kaneko/suprnova.cpp
+++ b/src/mame/kaneko/suprnova.cpp
@@ -621,8 +621,8 @@ static INPUT_PORTS_START( cyvern )      /* 2 buttons, 2 players */
 	PORT_INCLUDE( skns )
 
 	PORT_MODIFY("400000")
-	PORT_BIT( 0x00400000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
-	PORT_BIT( 0x40000000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
+//	PORT_BIT( 0x00400000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
+//	PORT_BIT( 0x40000000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
 INPUT_PORTS_END
 
 static INPUT_PORTS_START( galpanis )    /* 1 button, 2 players */
--- a/src/mame/kaneko/suprnova.cpp
+++ b/src/mame/kaneko/suprnova.cpp
@@ -621,8 +621,8 @@ static INPUT_PORTS_START( cyvern )      /* 2 buttons, 2 players */
 	PORT_INCLUDE( skns )
 
 	PORT_MODIFY("400000")
-//	PORT_BIT( 0x00400000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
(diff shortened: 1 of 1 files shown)


angelosa:
Test mode doesn't map this, what is supposed to be, full auto stuff? 



What's the end consensus about unofficial buttons, mapping as `IPT_OTHER`?

RiZeuM:
The C button is a fullauto but it doesn't function like the A button. A single tap on the C button has a unique effect, very short shot, different from long A button shot. I've tested this myself on the PCB, and you can see it in the few non-emulated videos available.

Seems like Kaneko forgot about this and left this button unmapped

ilker2495:
Merging this should be a no-brainer since it's more accurate to real PCB behavior!

angelosa:
https://www.arcade-museum.com/manuals-videogames/C/Cyvern.pdf the official manual just shows 2 buttons, so imo this needs the `IPT_OTHER` nomenclature (and a note stating why, also because reading a commented out `No Button 3` sure is bogus).

image

angelosa:
Like this?
```diff
diff --git a/src/mame/kaneko/suprnova.cpp b/src/mame/kaneko/suprnova.cpp
index 2691c775607..aa22a92300e 100644
--- a/src/mame/kaneko/suprnova.cpp
+++ b/src/mame/kaneko/suprnova.cpp
@@ -169,7 +169,7 @@ NEP-16
 
 #include "logmacro.h"
 
-#define LOGHIT(...)   LOGMASKED(LOG_HIT, __VA_ARGS__) 
+#define LOGHIT(...)   LOGMASKED(LOG_HIT, __VA_ARGS__)
 
 static void hit_calc_orig(u16 p, u16 s, u16 org, u16 &l, u16 &r)
 {
@@ -621,8 +621,10 @@ static INPUT_PORTS_START( cyvern )      /* 2 buttons, 2 players */
 	PORT_INCLUDE( skns )
 
 	PORT_MODIFY("400000")
-	PORT_BIT( 0x00400000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
-	PORT_BIT( 0x40000000, IP_ACTIVE_LOW, IPT_UNUSED )   /* No Button 3 */
+	// cyvern control panel sheets claims 2 buttons,
+	// but game actually has unofficial third button for full auto
+	PORT_BIT( 0x00400000, IP_ACTIVE_LOW, IPT_OTHER ) PORT_NAME("P2 Button 3 (unofficial)") PORT_PLAYER(2)
+	PORT_BIT( 0x40000000, IP_ACTIVE_LOW, IPT_OTHER ) PORT_NAME("P1 Button 3 (unofficial)") PORT_PLAYER(1)
 INPUT_PORTS_END
 
 static INPUT_PORTS_START( galpanis )    /* 1 button, 2 players */
```

cuavas:
I’d rather this just be regular `IPT_BUTTON3` since it’s wired and read in the usual place.  At most give it a `PORT_NAME` and update the comment to say it’s undocumented and not shown in test mode.

cuavas:
Holding button 3 is equivalent to constantly tapping button 1, i.e. it will use the primary shot attack, and never switch to fire breath/lightning/whatever as holding button 1 would.  It’s particularly useful for people who have issues with RSI.  Button 3 in Strikers 1945 Plus has a similar effect.

PR #15322, opened 2026-05-10
Linked-play support for Taito's Gunbuster

trap15:
I've added an implementation of the link system for Taito's Gunbuster.

There are 2 input/output pins between each board, but I haven't found how they're used so they're only mentioned in comments. The actual communication system happens entirely over a small shared RAM buffer.

They generally pretty consistently use the last two words as control words for the "ownership" of each buffer, so I was able to utilize those to only send RAM contents over the network when the ownership transfers. The actual synchronization of the control words is a bit of a mystery, I'm not entirely sure why it works honestly 😅 But it does work very well, and anything more proper looking didn't work. However, I must say I've only tested localhost thus far, I don't know how "fragile" networking support is allowed to be.

As for the linked-play itself, it's quite neat, so I hope some people will enjoy this. When trying to get it set up, the two emulated machines need to be reset at a somewhat similar time; there's a decent amount of variance allowed, but it's not too hard to get a bad timing.

### DIFF
--- a/src/mame/taito/gunbustr.cpp
+++ b/src/mame/taito/gunbustr.cpp
@@ -49,6 +49,7 @@
 #include "taito_en.h"
 #include "taitoio.h"
 #include "tc0480scp.h"
+#include "gunbustr_l.h"
 
 #include "cpu/m68000/m68020.h"
 #include "machine/eepromser.h"
@@ -82,7 +83,7 @@ public:
 		m_tc0480scp(*this, "tc0480scp"),
 		m_ram(*this,"ram"),
 		m_spriteram(*this,"spriteram"),
-		m_netram(*this,"netram"),
+		m_link(*this, "link"),
 		m_eeprom(*this, "eeprom"),
 		m_gfxdecode(*this, "gfxdecode"),
 		m_palette(*this, "palette"),
@@ -94,7 +95,6 @@ public:
 
 	{
 		m_coin_lockout = true;
-		link_netintf_init(mconfig);
 	}
 
 	void gunbustr(machine_config &config);
@@ -123,7 +123,7 @@ private:
 	required_device<tc0480scp_device> m_tc0480scp;
 	required_shared_ptr<u32> m_ram;
 	required_shared_ptr<u32> m_spriteram;
-	required_shared_ptr<u32> m_netram;
+	required_device<gunbustr_link_device> m_link;
 	required_device<eeprom_serial_93cxx_device> m_eeprom;
 	required_device<gfxdecode_device> m_gfxdecode;
 	required_device<palette_device> m_palette;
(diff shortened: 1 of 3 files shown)


src/mame/taito/gunbustr.cpp
```diff
@@ -469,25 +838,25 @@ static INPUT_PORTS_START( gunbustr )
...
-	PORT_BIT( 0x01, IP_ACTIVE_LOW, IPT_JOYSTICK_UP ) PORT_8WAY PORT_PLAYER(2)
-	PORT_BIT( 0x02, IP_ACTIVE_LOW, IPT_JOYSTICK_DOWN ) PORT_8WAY PORT_PLAYER(2)
-	PORT_BIT( 0x04, IP_ACTIVE_LOW, IPT_JOYSTICK_LEFT ) PORT_8WAY PORT_PLAYER(2)
-	PORT_BIT( 0x08, IP_ACTIVE_LOW, IPT_JOYSTICK_RIGHT ) PORT_8WAY PORT_PLAYER(2)
-	PORT_BIT( 0x10, IP_ACTIVE_LOW, IPT_UNKNOWN )
-	PORT_BIT( 0x20, IP_ACTIVE_LOW, IPT_UNKNOWN )
-	PORT_BIT( 0x40, IP_ACTIVE_LOW, IPT_UNKNOWN )
-	PORT_BIT( 0x80, IP_ACTIVE_LOW, IPT_UNKNOWN )
+	PORT_START("PORT03")
+	PORT_BIT( 0x01, IP_ACTIVE_HIGH, IPT_BUTTON3 ) PORT_PLAYER(1) // DMAON (6-A5)
```
angelosa: This should just be an `IPT_CUSTOM` with no `PORT_PLAYER`, it's not user facing (and eventually needs to be hooked by the TC0570SPC, when we have one).

Hydreigon223:
Also this isn't even a networked setup either. This should be set up like arescue and f1en emulating the two boards as a single instance as these two and gunbuster communicate through bridged RAM.

The only difference here is gunbustr is able to run as a single instance while arescue and f1en fail to boot if it can't find another pcb.

angelosa:
> Also this isn't even a networked setup either. This should be set up like arescue and f1en emulating the two boards as a single instance as these two and gunbuster communicate through bridged RAM.

The difference here is that `arescue` and `f1en` are a mandatory setup, plus I'm not really keen in how it's done there.

cuavas:
I’ll re-do the actual networking part myself.  I should be able to get this done before the next release.

cuavas:
See e6eefd2f633bd223b1f73c6ff3568c0e2712347d

cuavas:
Here it is working on WiFi between the Japan version on AArch64 and the US version on x86-64:

PR #15327, opened 2026-05-11
taito/gunbustr.cpp: convert analog to ADC, remove irq 5 hack

angelosa:


angelosa:
Branch as convenience so it doesn't conflict with #15322

PR #15331, opened 2026-05-12
sound/capcom_q1*.cpp: Match device name as chip markings

cam900:
- Move sound/capcom_q1*.cpp clock into each drivers

sound/capcom_q1_hle.cpp: Fix save state support

QSound is algorithm that made by QSound Labs (former Archer, referred from ssf2 attract mode), Capcom Q1 supports that but not a chip name. (output, instead)

reference: https://siliconpr0n.org/archive/doku.php?id=mcmaster:capcom:dl-1425 ,
https://en.wikipedia.org/wiki/QSound ,
https://en.wikipedia.org/wiki/QSound_Labs


cuavas:
The change to the device short names will annoy a lot of people as it will require renaming ROMs.

PR #15332, opened 2026-05-12
floppy: fix oob caused by m_step_samplepos dirty trick

arekkusu42:
(noticed during #15329)

This commit fixes an out of bounds access which occurs rarely in `floppy_sound_device::sound_stream_update()`.

To reproduce, do something that plays a lot of track-seeking step samples, for example:
1) `mame laser128 -flop1 <copy II plus> -flop2 <another disk>`
2) COPY, DISK W/ FORMAT, SLOT 6 DRIVE 1, SLOT 6 DRIVE 2, wait
3) repeat step 2 (for me, it triggers on the second iteration)

Result: rarely, the step sample causes out-of-bounds access (possible crash.)  It is triggered by [this code](https://github.com/mamedev/mame/blob/45faab9e7b23878adf14f275bb3b2528de7b6a47/src/devices/imagedev/floppy.cpp#L1956) (instrumentation added):
```
	// yep, a somewhat dirty trick; we don't have to record yet another sample
	m_step_samplepos += 441;
if (m_step_samplepos >= m_sample[m_step_sample-1].data.size())
__builtin_printf("step trick + 441: %d > %ld\n", m_step_samplepos, m_sample[m_step_sample-1].data.size());
```
A debug ASAN build (ARCHOPTS="-fsanitize=address") catches the problem immediately:
```
step trick + 441: 3421 > 3292
SUMMARY: AddressSanitizer: heap-buffer-overflow floppy.cpp:1984 in floppy_sound_device::sound_stream_update(sound_stream&)
```
[laser128_copyiiplus_asan.txt](https://github.com/user-attachments/files/27622467/laser128_copyiiplus_asan.txt)

### DIFF
--- a/src/devices/imagedev/floppy.cpp
+++ b/src/devices/imagedev/floppy.cpp
@@ -1981,7 +1981,8 @@ void floppy_sound_device::sound_stream_update(sound_stream &stream)
 				sampleend = m_sample[m_step_sample-1].data.size();
 
 				// Mix it into the stream value
-				out += m_sample[m_step_sample-1].data[m_step_samplepos++];
+				if (m_step_samplepos < sampleend)
+					out += m_sample[m_step_sample-1].data[m_step_samplepos++];
 				if (m_step_samplepos >= sampleend)
 				{
 					// Step sample done


cuavas:
I really hope there aren’t more of these lurking in that change.  I wanted the migration to C++20 and other related changes to go as smoothly as possible.  This feels like a big risk.

Robbbert:
I had an issue with the new floppy sound crashing on line 747 of devices/imagedev/floppy.cpp . Changed the line to

if (m_make_sound && m_sound_out) m_sound_out->unload();

It happened if I tried to load a softwarelist item that I didn't have. Since nobody has said anything it probably was only a problem for me.

arekkusu42:
I'm in the habit of leaving ASAN on anytime I'm not in a release build, and haven't noticed any more of these.  But I'm only looking at the apple2 and macplus families, and I only use the command line, not the software lists.

PR #15337, opened 2026-05-13
kungfub3s: replace BAD_DUMP sound ROM with one from another romset

clawgrip:
Since there's no useful information at all

### DIFF
--- a/src/mame/irem/m62.cpp
+++ b/src/mame/irem/m62.cpp
@@ -1393,7 +1393,7 @@ ROM_START( kungfub3s )
 	ROM_LOAD( "5.bin", 0x4000, 0x4000, CRC(8e7e4c56) SHA1(fdb79fcc652f1d0af74517e5df75921e8d3ca50f) )
 
 	ROM_REGION( 0x10000, "irem_audio:iremsound", ROMREGION_ERASEFF )
-	ROM_LOAD( "3.bin", 0xa000, 0x2000, NO_DUMP )
+	ROM_LOAD( "3.bin", 0xa000, 0x2000, BAD_DUMP CRC(58e87ab0) SHA1(3b03c101fec58eac13fc309a78df9a2cd44f7604) ) // Taken from other set, probably correct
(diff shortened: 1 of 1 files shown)


src/mame/irem/m62.cpp
```diff
@@ -1393,7 +1393,7 @@ ROM_START( kungfub3s )
 	ROM_LOAD( "5.bin", 0x4000, 0x4000, CRC(8e7e4c56) SHA1(fdb79fcc652f1d0af74517e5df75921e8d3ca50f) )
 
 	ROM_REGION( 0x10000, "irem_audio:iremsound", 0 )
-	ROM_LOAD( "3.bin", 0xa000, 0x2000, BAD_DUMP CRC(b4293435) SHA1(5e2b96c19c4f5c63a5afa2de504d29fe64a4c908) ) // Empty ROM
+	ROM_LOAD( "3.bin", 0xa000, 0x2000, NO_DUMP )
```
angelosa: Needs `ROMREGION_ERASEFF` to preserve current behaviour (the ROM is 1-filled).

Assuming we even want to do this, looks a YAGNI to me?

clawgrip: I'm just trying to avoid false positives when looking for new dumps matches...

cuavas: The `ROMREGION_ERASEFF` fills not just the part where the dead ROM was, but also the unused part below 0xa000 with 0xff.  Not sure if that’s desired.  Does 0xff vs 0x00 in the dead ROM space really improve anything for the 6803?

happppp:
May I suggest this instead:

```ROM_LOAD( "3.bin", 0xa000, 0x2000, BAD_DUMP CRC(58e87ab0) SHA1(3b03c101fec58eac13fc309a78df9a2cd44f7604) ) // Taken from other set, probably correct```

clawgrip:
> May I suggest this instead:
> 
> `ROM_LOAD( "3.bin", 0xa000, 0x2000, BAD_DUMP CRC(58e87ab0) SHA1(3b03c101fec58eac13fc309a78df9a2cd44f7604) ) // Taken from other set, probably correct`

Much better, thanks a lot

PR #15339, opened 2026-05-13
new NOT WORKING systems (3 Yeno educational laptops, probably missing internal ROM)

mamehaze:
new NOT WORKING machines
-----------
Mister X2 (Germany) [Team Europe]

new NOT WORKING clones
------------
Intellectus 2 (France) [Team Europe]
Computer Kid Personal (Italy) [Team Europe]

new NOT WORKING software list entries
------------
yeno_laptop_cart.xml:
Wissens Tester (Germany) [Team Europe]
Super Wissen (Germany) [Team Europe]

### DIFF
--- a/hash/yeno_laptop_cart.xml
+++ b/hash/yeno_laptop_cart.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<!DOCTYPE softwarelist SYSTEM "softwarelist.dtd">
+<!--
+license:CC0-1.0
+-->
+<softwarelist name="yeno_laptop_cart" description="Yeno educational laptop cartridges">
+
+	<software name="wissenst" supported="no">
+		<description>Wissens Tester (Germany)</description>
+		<year>199?</year>
+		<publisher>Yeno</publisher>
+		<part name="cart" interface="yeno_laptop_cart">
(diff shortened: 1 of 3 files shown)


mamehaze:
These all look like this inside.  TeamEurope managed to dump the glob which is a ROM, but I think it's data.
I'm guessing the KS56C1660-12 is the CPU and the same as the one in the Samsung document
https://bitsavers.trailing-edge.com/components/samsung/_Databooks/1990_Samsung_Semiconductor_Product_Guide.pdf
which would suggest it has ~16kb of internal ROM.  I can't see much that looks like code in the external ROM but would be nice if somebody could confirm.

Pinout

mamehaze:
one of the carts contained a regular ROM, so the dump of that is good / complete, the other was a glob.
the system ROMs and cartridge ROMs all seem to have a very similar structure, so I'm guessing the system just uses either cart data or the data from the ROM near the CPU depending on if a cart is plugged in.

PCB_Top

mamehaze:
Weirdly the original Mister X, which is in MAME as 'misterx' is a Z80 based system though, so wouldn't a 4-bit MCU be a step down?  That gives me some doubts over the identification.

PR #15341, opened 2026-05-14
Update generalplus_gpl951xx_soc.h - attempt to fix older GCC compile

mamehaze:


happppp:
If you don't mind, I already 'fixed' it by removing the array: https://github.com/mamedev/mame/commit/4a26071d6671b0a614b275ae31bf40b4c3eb476f

mamehaze:
either works in this case, so I don't mind

PR #15343, opened 2026-05-15
upd7810: fix SET_Z on OFFI/ONI instructions

goldnchild:
According to the UPD7810 datasheet, the OFF/ON instructions are supposed to affect the Z flag.

This will add SET_Z to the instruction opcode routine.

I considered rewriting it as doing the SET_Z ( x & imm ); first then SKIP_Z; or SKIP_NZ;
if you think that would be better.

=================================

I was also looking at the INR r2  instructions and it seems that it should affect the carry bit according to the datasheet.

INRW_wa
INR_A
INR_B
INR_C

DCR A
DCR_B
DCR_C
DCRW_wa

I see that they restore the original carry bit but it seems to say that CY should be affected by the operation.

### DIFF
--- a/src/devices/cpu/upd7810/upd7810_opcodes.cpp
+++ b/src/devices/cpu/upd7810/upd7810_opcodes.cpp
@@ -3552,6 +3552,7 @@ void upd7810_device::ONI_PA_xx()
 	RDOPARG( imm );
 	if (pa & imm)
 		PSW |= SK;
+	SET_Z( pa & imm );
 }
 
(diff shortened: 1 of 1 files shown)


happppp:
Yes, your consideration doing it the other way around is better, no need to check twice then.
CY flag not affected with INR/DCR (but HC is!), is correct.

ref: https://github.com/mamedev/mame/commit/b64b4ef8bbae5c5745740d0861393a2b37b0fd32

PR #15348, opened 2026-05-17
Replace adpcm rom bad dump for gongfuxm  lanmao.cpp

flama12333:
after inspecting using audacity adpcm raw import I have realized that sound distorted. the user re-dumped and seem to sound normal.


Osso13:
It isn't a bad dump, it's just scrambled. You are trying to substitute a good dump with a predecrypted version. This isn't MAME style of doing things.

PR #15349, opened 2026-05-17
trs/coco3.cpp: Fix CoCo 3 Keyboard interrupt

tlindner:
The previous implementation only triggered an interrupt on every key-up event. The actual hardware will trigger an interrupt at two specific times:
1. Zero keys pressed to one or more keys pressed.
2. One or more keys pressed to zero keys pressed.


src/mame/trs/coco3_m.cpp
```diff
@@ -87,17 +103,38 @@ void coco3_state::ff40_write(offs_t offset, uint8_t data)
...
+    coco_state::keyboard_changed(field, param, oldval, newval);
+
+    /* Support for CoCo 3 keyboard GIME interrupt */
+
+    int total = 0;
+    for (unsigned i = 0; i < m_keyboard.size(); i++)
+    {
+        /* Invert because active low: 0=pressed, so ~read() gives 1=pressed */
+        uint8_t pressed = (~(m_keyboard[i]->read()) | poll_joystick_buttons()) & 0xFF;
+        total += __builtin_popcount(pressed);
```
ajrhacker: Don't use this nonstandard function; population_count_32 exists in eminline.h for a reason. Or, better yet, since we're now on C++20, use std::popcount from `<bit>`.

ajrhacker (changes requested):

PR #15350, opened 2026-05-18
hh_ht11xx: Implements sound for HT1130-based handhelds

Figur8:
* Adds sound generation for the Holtek HT1130.
* The implementation is based on the BrickEmuPy project, using the official HT1130 datasheet as a baseline for validation.
* Tested on ga888 and piko55.


MooglyGuy:
I disagree with the way this is architected, it's something of an anti-pattern with how SoCs tend to be handled these days in MAME. I would suggest looking at the SunPlus SPG and GeneralPlus GPL SoCs over in src/devices/machine: Create a separate sound device, and have a separate device that inherits from both the sound device and the CPU device.

It would also be advisable, if possible, to implement the particular LFSR scheme directly rather than having the values in a hard-coded table.

happppp:
I don't mind putting the embedded sound generator source code in the MCU device.
Where should you draw the line? Should MCU embedded I/O ports be a separate device, or an embedded timer? Or for a versatile non-MCU chip for example, should the 6522 VIA timers be a separate device?

Anyway, if it does make more sense to put the sound generator in its own device, please keep it in the CPU folder so it's easier to maintain (not in src/devices/sound). See cpu/h8 for example that has local devices for separate function blocks.

MooglyGuy:
Don't play semantic games. MAME already has a well-established recent pattern of separating sound blocks, video blocks, and I/O blocks from the CPU core, and things are only continuing along those lines: PSX, SPG2xx, GPLxx, SCC68070, I can go dig up more if you'd like.

It's more nuanced when it comes to whether the sub-blocks should be in src/devices/cpu or src/devices/machine (or src/devices/sound or src/devices/video), but a reasonable dividing line there would be whether the CPU core is specific to one presently-emulated SoC or multiple, and whether that SoC is part of a specific architecture. That tidily handles why SCC68070, SPG2xx, GPLxx, and ARM-derived SoC implementations tend to live in src/devices/machine whereas PSX and H8 stuff lives in src/devices/cpu.

Either way, it's not a huge deal, but I just don't agree with lumping it all into one single device. I wouldn't kick up a fuss if someone wanted to merge things as-is, but the hard-coded LFSR array is a bit more troublesome IMO.

felipesanches:
@Figur8 how do you feel about attempting to implement the LFSR instead of leaving those hard-coded values in the source? If you need support, I can help you with that.

Figur8:
@felipesanches i accept this this, and enjoy help because i was studying to try this in future.

felipesanches:
OK. I'll take a closer look at this and I'll be back to you shortly.

mamehaze:
I'm glad to see sound being tackled on these though, as I wasn't really sure how to approach it.

Might motivate me to look at porting some of the other BrickEmuPy bits and pieces that haven't been ported yet.

mamehaze:
also in terms of separation, I think this kinda makes sense being in the CPU core.

the MCUs in question literally have a 'SOUND A' opcode (play sound based on accumulator) as well as SOUND ON, SOUND LOOP, SOUND OFF etc. it's a tighter bond than you'd see with other general purpose peripherals, so unless multiple otherwise incompatible CPU types have the same sound implementation separating it out doesn't really have any benefits.

felipesanches:
thanks, @Figur8, for addressing the LFSR on your most recently pushed commit. This is now looking better, so I think it is ready for a code-review from some of the members who have merge rights here.

Also, I'm pretty much aligned with the views of @mamehaze and @happppp here: I wouldn't bother about separating audio from the CPU core in this case because it is tightly coupled in this specific chip design and, as far as I can tell, there's no other chip that would benefit from sharing that sound implementation, so splitting it would probably just be a hassle, with very little practical benefit (in my opinion).

felipesanches:
Anything else needed here? I'd like to see it either merged or to get some feedback from reviewers.

felipesanches:
thanks!

PR #15357, opened 2026-05-19
hash: consistency for WWF WrestleMania

ArcadeShadow:
Consistency for WWF WrestleMania in several softwarelist hash files

### DIFF
--- a/hash/cpc_cass.xml
+++ b/hash/cpc_cass.xml
@@ -40922,7 +40922,7 @@ Please stick to using the floppy versions for the time being...
 	</software>
 
 	<software name="wmania" supported="no">
-		<description>WWF Wrestlemania (UK)</description>
+		<description>WWF WrestleMania (UK)</description>
 		<year>1991</year>
 		<publisher>Ocean Software</publisher>
 		<info name="usage" value="Load with |TAPE and then RUN&quot;&quot;" />
--- a/hash/sms.xml
+++ b/hash/sms.xml
@@ -8259,7 +8259,7 @@ license:CC0-1.0
 	</software>
 
 	<software name="wwfsteel">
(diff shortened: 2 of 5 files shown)


hash/sms.xml
```diff
@@ -8259,7 +8259,7 @@ license:CC0-1.0
 	</software>
 
 	<software name="wwfsteel">
-		<description>WWF Wrestlemania - Steel Cage Challenge (Europe, Brazil)</description>
+		<description>WWF WrestleMania - Steel Cage Challenge (Europe, Brazil)</description>
```
angelosa: `:` separator for `WrestleMania` and `Steel Cage Challenge`? Note: there's another one in Game Gear softlist (which isn't even consistent to this)

hash/cpc_cass.xml
```diff
@@ -40922,7 +40922,7 @@ Please stick to using the floppy versions for the time being...
 	</software>
 
 	<software name="wmania" supported="no">
-		<description>WWF Wrestlemania (UK)</description>
+		<description>WWF WrestleMania (UK)</description>
```
angelosa, line 40925: This romset ID diverges with all the others, it's even inconsistent with `cpc_flop.xml` (and I presume is the same game on a different format).

hash/sms.xml
ArcadeShadow: Added the `:` separator.
But there is much more entries without the `:` separator.

hash/cpc_cass.xml
ArcadeShadow: The cpc softlists are very messy, the metadata info and even the parent/clone relationship isn't trustworthy or complete. 
So I think those 2 soft lists `cpc_cass` and `cpc_flop` should be entirely revised, but IMO this isn't the right time for this.

PR #15368, opened 2026-05-22
nakajima/nakajies.cpp: Several updates

wilbertpol:
nakajima/nakajies.cpp:
- Add 10Hz timer, fixes empty battery display.
- Add PC Card/PCMCIA support.
- Add power button.


angelosa:
> Add beeper sound; not verified against real unit.

afaik the only Japanese computer using a buzzer was the vanilla PC-9801. Everything beyond that actually uses a `DAC1BIT`, implementable thru `speaker_sound_device`.

wilbertpol:
> > Add beeper sound; not verified against real unit.
> 
> afaik the only Japanese computer using a buzzer was the vanilla PC-9801. Everything beyond that actually uses a `DAC1BIT`, implementable thru `speaker_sound_device`.

As far as I know there are no schematics available or detailed information about the sound part of these machines, so I have taken out the sound part.

PR #15372, opened 2026-05-23
toshiba/t250.cpp: New driver for Toshiba T-200 and T-250

davidlrand:
8085-based CP/M business machines from 1981.  Both share the same main board; only the floppy subsystem differs.  T-200 uses 5.25" DSDD floppies, T-250 uses 8" DSDD.

Verified working subsystems:
- Intel 8085A at 16 MHz / 3 (5.333 MHz)
- 64 KB RAM with boot ROM bank-switched at 0x0000-0x0FFF
- HD6845 CRTC with 10-pixel wide character cell and VPG-RAM glyph storage (CPU shares VPG with the 6845 via 0xF800 bank-switch)
- Hardware inverse video (XOR over the full 10x10 cell)
- Intel 8279 keyboard + keypad (matrix + rl8 path, shift/ctrl mods)
- 2 kHz keyboard buzzer (BUZZ bit of 0xC0 bank latch)
- Intel 8257 DMA, including ch0/ch1 memory-to-memory used by the BIOS clear-screen and scroll routines
- NEC uPD765A FDC with external RESET latch at port 0xF0 bit 7 (T-250: 8 MHz, 500 kbps MFM; T-200: 4 MHz, 250 kbps, motor enable at port 0xF0 bit 6)
- Centronics parallel printer
- Intel 8253 PIT counter 2 generating baud-rate clock (1.9968 MHz input per the actual CCM board, NOT the standard 1.8432 crystal)
- Intel 8251 USART (full TX + RX path verified end-to-end)
- Optional SASI hard-disk interface (8155 host adapter to nscsi_bus; empty by default, attach via -sasi:0 harddisk -hard1 disk.chd)

CP/M 2.20 / 2.32 boots cleanly from period-original floppies on both variants.  RST 6.5 interrupt source dispatched via the read-only register at port 0xD0 (rdint).

Driver and both ROM dumps contributed by the original BIOS author. Hardware-archive repository at:
  https://github.com/davidlrand/toshiba-t250-t200

### DIFF
--- a/src/mame/toshiba/t250.cpp
+++ b/src/mame/toshiba/t250.cpp
@@ -58,6 +58,7 @@
 #include "machine/i8251.h"
 #include "machine/i8257.h"
 #include "machine/i8279.h"
+#include "machine/input_merger.h"
 #include "machine/nscsi_bus.h"
 #include "machine/nscsi_cb.h"
 #include "machine/output_latch.h"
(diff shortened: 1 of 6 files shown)


angelosa:
I presume this does nothing without a boot disk? I'm not keen to have "working" computer systems without any software list attached.

davidlrand:
Emulators without software are in fact pretty useless.  I have live software in https://github.com/davidlrand/toshiba-t250-t200 and will be going through and posting the rest of my disks over the next few days.  The system is live, rebuilt with a 32M CDC Finch emulator and running the code I wrote ~45 years ago.  I have fully functional hardware for both the T-250 and T-200, and have previously archived the detailed manuals on Bitsavers thanks to Al.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+{
+	m_maincpu->space(AS_PROGRAM).write_byte(offset, data);
+}
+
+
+//**************************************************************************
+//  SASI hard-disk host adapter
+//
+//  Real hardware bridges a SASI bus to the 8085 via an Intel 8155.  We do
+//  not model the 8155 itself — we mimic the five port semantics the BIOS
```
pmackinlay: Is this adapter actually an optional card that fits into some kind of slot or connector? If so, it would be nice to implement it as a slot device in MAME, moving all the adapter code into another class.

Did you attempt to emulate this adapter using the actual 8155 device in MAME? Is there any good reason not to do so?

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	map(0xd3, 0xd3).r(FUNC(t250_state::ios_r));
+	map(0xd4, 0xd4).r(FUNC(t250_state::prnd_loopback_r));
+	map(0xe0, 0xe0).rw(m_crtc, FUNC(hd6845s_device::status_r), FUNC(hd6845s_device::address_w));
+	map(0xe1, 0xe1).rw(m_crtc, FUNC(hd6845s_device::register_r), FUNC(hd6845s_device::register_w));
+	map(0xf0, 0xf1).m(m_fdc, FUNC(upd765a_device::map));
+	map(0xf0, 0xf0).w(FUNC(t250_state::fdc_reset_w));
+}
+
+
+// Keyboard matrix.  The 8279 scans positions 0-7 (SL output); for each
```
pmackinlay: Similarly, is the keyboard detachable and connected via some kind of port to the machine? If so, it would streamline the driver by moving the keyboard implementation into its own slot device.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	// Default 0x1F matches the as-shipped T-250 setting (manual p.33).
+	PORT_START("CFG")
+	PORT_DIPNAME(0x01, 0x01, "Cold boot source")              PORT_DIPLOCATION("SW2:1")
+	PORT_DIPSETTING(   0x00, "Track 0, Head 0, Sector 1")
+	PORT_DIPSETTING(   0x01, "Track 1, Head 0, Sector 1")
+	PORT_DIPNAME(0x02, 0x02, "Keyboard type")                 PORT_DIPLOCATION("SW2:2")
+	PORT_DIPSETTING(   0x00, "Standard")
+	PORT_DIPSETTING(   0x02, "Touch-in")
+	PORT_DIPUNKNOWN_DIPLOC(0x04, 0x04, "SW2:3")
+	PORT_DIPNAME(0x08, 0x08, "CCM (serial) board")            PORT_DIPLOCATION("SW2:4")
```
pmackinlay, line 786: Seems this comms board is possibly also optional and could be a slot device?

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	PORT_DIPSETTING(   0x00, "Not present")
+	PORT_DIPSETTING(   0x08, "Present")
+	PORT_DIPNAME(0x10, 0x10, "Printer width")                 PORT_DIPLOCATION("SW2:5")
+	PORT_DIPSETTING(   0x00, "80 columns")
+	PORT_DIPSETTING(   0x10, "136 columns")
+	PORT_DIPNAME(0x20, 0x00, "Printer")                       PORT_DIPLOCATION("SW2:6")
+	PORT_DIPSETTING(   0x00, "Present")
```
pmackinlay, line 793: The printer is already implemented as a Centronics slot device, I think; what effect does this DIP switch have?

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	// increments for non-empty slots, so dense iteration matters.
+	NSCSI_BUS(config, m_sasi);
+	NSCSI_CONNECTOR(config, "sasi:0", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:1", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:2", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:3", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:4", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:5", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:6", default_scsi_devices, nullptr);
+	NSCSI_CONNECTOR(config, "sasi:7", default_scsi_devices, "sasicb", true)
```
pmackinlay: This method of installing the SCSI/SASI controller device is deprecated. Please check other drivers for examples of code which explicitly instantiate the device and use the `set_external_device()` function to link it to the bus.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	m_sasi_ack_timer = timer_alloc(FUNC(t250_state::sasi_ack_off), this);
+	m_sasi_sel_timer = timer_alloc(FUNC(t250_state::sasi_sel_off), this);
+	m_sasi_rst_timer = timer_alloc(FUNC(t250_state::sasi_rst_off), this);
+
+	// Cache the actual bus refid of our callback device.  MAME's nscsi_bus
+	// only counts non-empty slots in m_devcnt, so our SCSI ID 7 callback may
+	// land at refid 1 (when only one target is also attached).  Writes to a
+	// refid >= m_devcnt are silently dropped by regen_ctrl/regen_data.
+	// Access the protected member m_scsi_refid via a using-public derived
+	// class trick.
```
pmackinlay: This shouldn't be necessary if you make your SASI controller its own device inheriting from `nscsi_device_interface`, as you could then access `m_scsi_refid` directly.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	m_sasi_synth_bsy = false;
+	m_sasi_synth_req = false;
+	m_sasi->ctrl_w(m_sasi_refid, 0, nscsi_device_interface::S_ALL);
+	m_sasi->data_w(m_sasi_refid, 0);
+}
+
+
+u8 t250_state::mem_low_r(offs_t offset)
+{
+	return m_rom_at_low ? m_rom[offset] : m_ram[offset];
```
pmackinlay: It's better to use memory views to handle this type of thing. By installing your RAM first, and then overlaying the ROM with a view, you can eliminate the `mem_low` and `mem_mid` handlers. A second view can easily eliminate the `mem_high` handlers.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1562 @@
...
+	if (m_kbd_irq)    d &= ~0x80;  // bit 7 = keyboard
+	if (m_usart_txrdy) d &= ~0x40; // bit 6 = comm out (TxRDY)
+	if (m_usart_rxrdy) d &= ~0x20; // bit 5 = comm in  (RxRDY)
+	if (m_prn_irq)    d &= ~0x10;  // bit 4 = printer
+	return d;
+}
+
+void t250_state::update_rst65()
+{
+	bool any = m_kbd_irq || m_prn_irq || m_usart_rxrdy || m_usart_txrdy;
```
pmackinlay: It's probably wise to replace this with an `INPUT_MERGER_ANY_HIGH` device.

pmackinlay (changes requested):
Overall, this is a really excellent first effort, well done!

davidlrand:
Thanks Patrick — really helpful review. Pushed an update (182d5bb) addressing all eight points. Per-comment replies inline. The slot-device interpretation came out slightly different on CCM (RS232 connector kept at the system level for command-line stability) — happy to refactor further if you'd prefer the strict slot-card pattern.

src/mame/toshiba/t250.cpp
davidlrand: Done — the SASI host adapter is now `t250_sasi_host_device`, subclassing `device_t` + `nscsi_device_interface`, with direct `m_scsi_refid` access. The `refid_accessor` using-public-derived-class trick is removed.

src/mame/toshiba/t250.cpp
davidlrand: Done — `m_rom_view` overlays boot ROM at 0x0000-0x0FFF (with an `lw8` lambda passing writes through to the underlying DRAM, matching real hardware where the DRAM chip is wired in parallel and absorbs writes regardless of which side the CPU is reading from), and `m_vpg_view` switches between video RAM and VPG character-generator RAM at 0xF800-0xFFFF. All six `mem_low_r/w`, `mem_mid_r/w`, `mem_high_r/w` handlers removed; the `m_rom_at_low`/`m_vram_at_high` bool members are gone too (view state is auto-saved by MAME).

src/mame/toshiba/t250.cpp
davidlrand: Done — added `m_rst65_merger` with the four sources (keyboard / 8251 RxRDY / 8251 TxRDY / printer ACK) wired to `in_w<0..3>`. The manual `update_rst65()` method is removed.

src/mame/toshiba/t250.cpp
davidlrand: The adapter is now extracted into `t250_sasi_host_device` (subclasses `device_t` + `nscsi_device_interface`, attached via `m_sasi->set_external_device(7, m_sasi_host)`).

On the 8155 question: the board does contain an 8155, but the CPU never accesses it through standard 8155 register addresses. The 8155's PA/PB/PC port pins drive the SASI bus directly through external glue logic, and what the CPU sees at I/O ports 0x00-0x08 are the glue-logic registers, not the 8155 registers themselves. Attaching MAME's `i8155_device` internally would only buy us simulation of the 8155's internal 256-byte RAM and 14-bit timer, neither of which the firmware exercises. I've left a TODO comment noting the path so a future pass can wire it up if/when we document the schematic in enough detail to justify the extra moving parts.

src/mame/toshiba/t250.cpp
davidlrand: The keyboard PCB connects via a high-pin-count round cable and can be physically disconnected for service, but it isn't a user-pluggable peripheral — no external connector on the case, no documented alternate keyboard for the T-200/T-250, and no period software that probes for layout variation. I've added a comment block above the matrix explaining the decision (and noting that the matrix + PORT_CHAR table lift cleanly into a device class if a documented alternate ever surfaces). Kept inline for now.

src/mame/toshiba/t250.cpp
davidlrand: The CCM is physically a separate PCB on its own card edge, but it isn't user-optional — every shipping T-200/T-250 has one (and stock BIOS uses it for the system console). I've extracted it into `t250_ccm_device` (owns the 8251 USART + 8253 PIT) but attached it unconditionally rather than as a removable slot card. Happy to convert to a strict slot-card interface if you'd prefer.

One wiring subtlety: the RS232 connector physically lives on the CCM card edge, but I left `m_rs232` at the host level so the `-rs232 X` slot path stays at `:rs232` and existing command lines keep working. CCM's 8251 outputs route via the `^rs232` parent-tag idiom; the host wires its RS232 inputs back into the CCM via forwarder methods.

src/mame/toshiba/t250.cpp
davidlrand: Good question — the DIP doesn't reconfigure the Centronics slot device. Stock V2.21 BIOS only reads bits 0 (cold-boot source) and 1 (keyboard type) of the CFG block; the remaining bits (printer width, printer present, CCM presence) are exposed at I/O port 0xD2 as configuration data for application code to consult. Alternate ROM implementations from period resellers (e.g. Trianex V3.0) may use more of them. I've added a comment block before `PORT_START("CFG")` explaining this.

src/mame/toshiba/t250.cpp
davidlrand: Done — replaced the `NSCSI_CONNECTOR("sasi:7", ..., "sasicb", true).option_add_internal(...).machine_config(...)` pattern with `T250_SASI_HOST(config, m_sasi_host); m_sasi->set_external_device(7, m_sasi_host);`. The host adapter sits on the bus as a real `nscsi_device_interface` participant rather than as a callback wrapper.

davidlrand:
> Overall, this is a really excellent first effort, well done!

One additional comment, and a question, please.

Thank you for the porting on the PC532 - I appreciate the work!  I'm just in the process of lighting up some of the other National boards I worked on (DSI-32, and other add-in cards).  I had some architecture questions of how things like the DSI-32 (PC add-in card, but the whole point is running the software on the 32k).  I have all of the code, of course, and have modified a stand-alone emulator to work with the loader and 32io, and want to make sure I use the "right framework" for this.  

Dave Rand

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+//  modern set_external_device() idiom.
+//
+//  TODO: Replace the bespoke ACK/RST/SEL emu_timer pulses with an
+//  i8155_device sub-instance whose internal timer drives those edges
+//  on real hardware.  For now the I/O port semantics are preserved
+//  bit-for-bit so the existing CP/M BIOS m32 build (drives C/D/E/F)
+//  continues to work without further changes.
+// ============================================================================
+
+class t250_sasi_host_device : public device_t, public nscsi_device_interface
```
pmackinlay: Please split this into separate files (e.g., `src/mame/toshiba/t250_sasi.[cpp,h]`) to reduce the total size of the driver.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+
+	static constexpr u8 SASI_IDLE            = 0;
+	static constexpr u8 SASI_SELECTING       = 1;
+	static constexpr u8 SASI_CMD_FIRST       = 2;
+	static constexpr u8 SASI_PASSTHROUGH     = 3;
+	static constexpr u8 SASI_SWALLOW_CMD     = 4;
+	static constexpr u8 SASI_SWALLOW_GARBAGE = 5;
+	static constexpr u8 SASI_SWALLOW_STATUS  = 6;
+
+	u8 m_sasi_phase = SASI_IDLE;
```
pmackinlay: Prefer initializing in the constructor per previous feedback from @cuavas.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+//
+//  Physically a separate PCB carrying the Intel 8251 USART, the Intel 8253
+//  PIT (counter 2 generates baud clock for both TxC and RxC), and the
+//  RS-232 port.  Standard on every T-200/T-250 (Dave confirmed not user-
+//  optional), but lives on its own card edge so a device-class
+//  encapsulation is the right shape.  Patrick's review (PR #15372,
+//  line 1335) asked for this split.
+// ============================================================================
+
+class t250_ccm_device : public device_t
```
pmackinlay: Please also split this into separate files.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+	m_pit->set_clk<1>(1'996'800);
+	m_pit->set_clk<2>(1'996'800);
+	m_pit->out_handler<2>().set(m_usart, FUNC(i8251_device::write_txc));
+	m_pit->out_handler<2>().append(m_usart, FUNC(i8251_device::write_rxc));
+
+	I8251(config, m_usart, 1'843'200);
+	// 8251 outputs route to the host's system-level RS232 port via
+	// parent-tag ("^rs232").  The host owns the RS232 connector so
+	// users can address it as -rs232 X at the command line.
+	m_usart->txd_handler().set("^rs232", FUNC(rs232_port_device::write_txd));
```
pmackinlay: This pattern is really discouraged (both the parent syntax and referring to tag by string instead of an object finder).

My preference is to move the port to this device and suffer the command-line inconvenience, but if you really want to keep it in the parent, use a `required_device` with a DUMMY_TAG and `set_tag()` to tell the child device the name of the port instead of using the parent tag string lookup here.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+	{ }
+
+	void t250(machine_config &config);
+	void t200(machine_config &config);
+
+private:
+	void common(machine_config &config);
+
+	virtual void machine_start() override ATTR_COLD;
+	virtual void machine_reset() override ATTR_COLD;
```
pmackinlay, line 118: These should be `protected:`.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+}
+
+void t250_state::dma_memw(offs_t offset, u8 data)
+{
+	m_maincpu->space(AS_PROGRAM).write_byte(offset, data);
+}
+
+
+//**************************************************************************
+//  SASI hard-disk host adapter
```
pmackinlay: I think this comment is now superfluous or should be relocated?

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+	//                   line for write strobes.
+	//   0xF800-0xFFFF : video RAM (m_vpg_view state 0) vs character-
+	//                   generator RAM / VPG (state 1).  The underlying
+	//                   DRAM at this range is never visible — one of the
+	//                   two SRAMs is always selected.
+	map(0x0000, 0xffff).ram().share("ram");
+
+	map(0x0000, 0x0fff).view(m_rom_view);
+	m_rom_view[0](0x0000, 0x0fff).rom().region("maincpu", 0)
+		.lw8(NAME([this](offs_t offset, u8 data) { m_ram[offset] = data; }));
```
pmackinlay: Is this necessary? The `ram()` write handler underneath the view should still be in place; `rom()` only installs/replaces the read handler.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+	map(0x08, 0x08).rw(m_sasi_host, FUNC(t250_sasi_host_device::hsts_r), FUNC(t250_sasi_host_device::hrset_w));
+	map(0x80, 0x88).rw(m_dmac, FUNC(i8257_device::read), FUNC(i8257_device::write));
+	map(0x88, 0x88).w(FUNC(t250_state::dmac_mode_w));   // intercept to trigger mem-to-mem
+	map(0x90, 0x90).rw(m_kbc, FUNC(i8279_device::data_r), FUNC(i8279_device::data_w));
+	map(0x91, 0x91).rw(m_kbc, FUNC(i8279_device::status_r), FUNC(i8279_device::cmd_w));
+	// CCM (comms control module) board — own device, see t250_ccm_device.
+	map(0xa0, 0xa3).rw(m_ccm, FUNC(t250_ccm_device::pit_r), FUNC(t250_ccm_device::pit_w));
+	map(0xa4, 0xa5).rw(m_ccm, FUNC(t250_ccm_device::usart_r), FUNC(t250_ccm_device::usart_w));
+	map(0xa6, 0xa6).w (m_ccm, FUNC(t250_ccm_device::mode_w));
+	map(0xa7, 0xa7).r (m_ccm, FUNC(t250_ccm_device::cicts_r));
```
pmackinlay, line 593: When the address range is contiguous like this, you may consider a device address map and the `m(device, FUNC())` approach.

src/mame/toshiba/t250.cpp
```diff
@@ -0,0 +1,1712 @@
...
+}
+
+void t250_ccm_device::device_add_mconfig(machine_config &config)
+{
+	// CCM-board baud clock is 1.9968 MHz (15.9744 MHz crystal / 8), not the
+	// standard 1.8432 MHz async baud crystal.  Per Dave from the set utility's
+	// divisor table: 19200 -> 6 gives 1996800/6/16 = 20800 (~8% high; lower
+	// rates are exact).
+	PIT8253(config, m_pit, 0);
+	m_pit->set_clk<0>(1'996'800);
```
pmackinlay: By convention, prefer `15.9744_MHz_XTAL / 8` syntax when the clock source is known. Integer format is typically used when the source or divisor is not known with confidence.

pmackinlay (commented):
Sorry for the delay due to business travel. This is getting close to mergeable. The serial port tag lookup and file splitting are the only blockers for me.

src/mame/toshiba/t250.cpp
davidlrand: Done.  I have some follow-on changes which will be a separate PR - I need to test them.

src/mame/toshiba/t250.cpp
davidlrand: Done.

src/mame/toshiba/t250.cpp
davidlrand: Done - though again, there is no opportunity for the customer of one of these machines to remove the CCM.  They just ran out of space, thus the separate board.

src/mame/toshiba/t250.cpp
davidlrand: Understood.  Completed.

src/mame/toshiba/t250.cpp
davidlrand: Done.

src/mame/toshiba/t250.cpp
davidlrand: It is, and has been removed.

src/mame/toshiba/t250.cpp
davidlrand: You are correct - the write to RAM landed fine.

src/mame/toshiba/t250.cpp
davidlrand: Completed (now that the devices are separate), and just the single map-per-device is done.

src/mame/toshiba/t250.cpp
davidlrand: Understood - we know exactly, here.  Done.

PR #15375, opened 2026-05-24
qx10: Move video subsystem to use a slot based architecutre

brijohn:
This updates the QX-10 system to use a slot based graphics system, to allow selecting either the color board or monochrome. This is more accurate to real hardware which used two distinct boards, rather then having some config knob that switches between modes.

### DIFF
--- a/scripts/src/bus.lua
+++ b/scripts/src/bus.lua
@@ -2193,6 +2193,20 @@ end
 
 ---------------------------------------------------
 --
+--@src/devices/bus/epson_qx/video/video.h,BUSES["EPSON_QX_VIDEO"] = true
+---------------------------------------------------
+
+if BUSES["EPSON_QX_VIDEO"] then
+	files {
+		MAME_DIR .. "src/devices/bus/epson_qx/video/qx_gdc_cards.cpp",
+		MAME_DIR .. "src/devices/bus/epson_qx/video/qx_gdc_cards.h",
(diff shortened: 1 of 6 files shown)


startaq:
Any particular reason you kept the `screen_device` inside the main driver instead of moving it to the gfx card device? Then you wouldn't need to expose an external screen_update function.

brijohn:
> Any particular reason you kept the `screen_device` inside the main driver instead of moving it to the gfx card device? Then you wouldn't need to expose an external screen_update function.

The main reason is because the graphics subsystem doesn't actually own the screen. The monitor attaches to a din on the computers main board and the graphics cards send their video output to the main board which then sends it to the monitor.

angelosa:
Gonna inquiry here for practical reasons: what's missing about this computer for marking it working beyond the lack of several SW list items?

brijohn:
> Gonna inquiry here for practical reasons: what's missing about this computer for marking it working beyond the lack of several SW list items?

At this point I don't think there is much of a reason for it to be marked as non working. All the software software i have tried currently is working.  I guess missing things would be fixing up the DIP SWs with the known values and i think there might be missing items save_item calls for save states. Besides that the only other thing i can think of is lightpen support, but that depends on having support in upd7220 which is missing currently. Lightpen support isn't really heavily used on the QX-10 either.

brijohn:
This has been here for a while is there any reason it can't get merged in?

angelosa:
By quick glance it seems fine to me.

PR #15377, opened 2026-05-24
sinclair/next/specnext_vtest.cpp: Factory Video test (joystick's XYZ to activate)

holub:


angelosa:
Comprehension questions (i.e. can accept this as a stopgap, just to be sure):
1. is the system really scanning X+Y+Z thru vblank irq, once per frame, or it's really a matter of using a templatized `PORT_CHANGED_MEMBER` that can do this instead?
2. Are these "joysticks" technically slot devices anyway?

holub:
> Comprehension questions (i.e. can accept this as a stopgap, just to be sure):
> 
>     1. is the system really scanning X+Y+Z thru vblank irq, once per frame, or it's really a matter of using a templatized `PORT_CHANGED_MEMBER` that can do this instead?
> 
>     2. Are these "joysticks" technically slot devices anyway?

1. Sampled during VSYNC (method renamed to decrease confusion); it's really XYZ from joystick; dedicated `PORT_CHANGED_MEMBER` will be further from the hw
https://gitlab.com/SpectrumNext/ZX_Spectrum_Next_FPGA/-/blob/30204/cores/zxnext/src/zxnext_top_issue5.vhd?ref_type=heads#L1724
2. In theory must be slot device but it doesn't

ped7g (approved):

PR #15379, opened 2026-05-24
plugins/keypress: New plugin for recording palyer behaviour

holub:


angelosa:
I have no involvement with this, not sure why GH credits me ...

cuavas:
This doesn’t actually do what it says:
* The PR title says it records player behaviour (i.e. imitates INP recording)
* What it actually does is try to imitate the ability of the Input Devices menu to show the state of host controls

This duplicates built-in functionality and hijacks popmessage (it should either draw to a UI container or use a plugin menu to show stuff to the user).

I’m going to remove this.  It duplicates built-in functionality and it’s implemented poorly.  I *might* add a similar plugin to the mame-goodies repo as sample code (but without constantly hijacking popmessage).  But it really doesn’t belong in baseline MAME.

holub:
works for me. thanks!

cuavas:
Reverted.  See mamedev/mame-goodies@d1c9005386b6af654e50d3d462d11743575c7418 for the replacement.

PR #15382, opened 2026-05-25
adp/stella8085.cpp: Fix 4040 board IO

stonedDiscord:
This maps the correct IO addresses for the 4040 boards, making bahia,disc,dpplstrt and discoly interactable.
This also clears up what IO 0x73 is, it's the 8255 control register.
The 8255 was replaced with discrete ICs and a fixed direction during the transition to 4087 but software on there will still set up the control registers.

This is best visible on discoly, before it would do nothing.
With this change it will blink the fault lamp and after pressing Up 1,- + Initialize it will show a fault code on the service display.

### DIFF
--- a/src/mame/adp/stella8085.cpp
+++ b/src/mame/adp/stella8085.cpp
@@ -154,8 +154,8 @@ void stella8085_state::program_4040_map(address_map &map)
 void stella8085_state::io_map(address_map &map)
 {
 	map(0x00, 0x00).w(FUNC(stella8085_state::io00));
-	map(0x50, 0x51).rw("kdc", FUNC(i8279_device::read), FUNC(i8279_device::write));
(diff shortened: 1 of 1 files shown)


src/mame/adp/stella8085.cpp
```diff
@@ -74,6 +76,7 @@ class stella8085_state : public driver_device
 	bool m_kbd_bd = false;
 
 	required_device<cpu_device> m_maincpu;
+	required_device<i8255_device> m_ppi;
 	required_device<i8256_device> m_uart;
 	required_device<i8279_device> m_kdc;
```
cuavas, line 81: You really shouldn’t be forcing the PPI to be present for games that don’t actually have it.  Just NOP out the writes to the configuration register if they upset you.

stonedDiscord: Will do the next time I update the file.

PR #15385, opened 2026-05-26
formats/imd_dsk: Add save support (IMD write for flopconvert)

davidlrand:
## Summary

IMD has been read-only in MAME's flopimg pipeline since 2017 — `imd_format::load()` parses ImageDisk files, but flux-level images (MFI, fresh Greaseweazle captures, running emulation state) cannot round-trip back to IMD. This change implements `imd_format::save()` so that `flopconvert mfi imd ...` works directly, and any driver-side "save as IMD" path becomes available.

After this:

```
floptool help formats
...
  imd     rw - IMD disk image [imd]
...
```

## Approach

Per-track decode-by-scoring rather than driver-supplied geometry:

- New `imd_format::extract_track_rich()` walks the MFM/FM bitstream once and produces records carrying **physical sector order**, **DAM-deleted flag**, and **CRC validity for both address and data records**. The shared `extract_sectors_from_bitstream_*_pc()` helpers in `flopimg.cpp` discard this info; rather than widen those helpers and disturb every caller, this format keeps a local extractor.
- New `imd_format::detect_track()` probes four common `(encoding, cell_size)` combinations per track — MFM at 1000 / 2000 ns and FM at 2000 / 4000 ns, covering 250 and 500 kbps in both modes — and picks the one with the most cleanly-decoded sectors (scored by IDAM CRC). Re-runs per track so **mixed-density disks decode honestly** without per-driver tuning (e.g. CP/M boot disks with an FM-128B boot track on cyl 0 head 0 and MFM data tracks elsewhere).
- New `imd_format::save()` walks tracks in IMD's canonical order, emits per-track records with the right mode byte, sector-numbering map, optional cyl/head maps, and per-sector type byte (with the compressed-fill optimisation for all-same-byte sectors). Deleted-DAM and bad-CRC sectors round-trip to IMD types 3-8 as appropriate.
- Header timestamp + comment are **preserved across an IMD -> IMD pass** via a `mutable` string member captured on `load()`; other source formats produce a fresh `IMD 1.20: <localtime>` header with a short `Created by MAME flopconvert` annotation.

Flux-noise phantom IDAMs (random sync hits in marginal flux) are filtered two ways:
1. **Push gate**: at the extractor's push sites, sectors with both IDAM and DAM CRCs bad are dropped — noise rarely produces a CRC-valid coincidence on either side.
2. **Dedup pass**: after extraction, candidates sharing a sector ID are deduplicated by a trust ranking that prefers has_data + good CRCs over no data over bad CRC.

Both filters were needed — the second one catches the case where a phantom IDAM has a unique fake sector ID and a fake idam_size that would otherwise trip the mixed-sector-sizes guard.

## Validation

- **24 / 24** disks in a local 5.25" T-200 archive (mixed-density boot tracks; bad-CRC sectors in unused CP/M space; rescue-merged tracks combined from multiple Greaseweazle captures) round-trip **byte-identical at the sector-data level** through `flopconvert imd -> mfi -> imd`. (Header timestamps differ as expected.)
- Cross-validated against a Greaseweazle-produced IMD of an 8" T-250 boot disk: byte-identical sectors after `imd -> mfi -> imd`.
- Per-track encoding detection produces the expected mode byte: e.g. for a Toshiba T-200 boot disk, cyl 0 head 0 -> mode 02 (FM 250) / 128 B, rest -> mode 05 (MFM 250) / 256 B.

## Limitations

These cases **refuse the save with a clear error** rather than producing corrupt output:

- Tracks with mixed sector sizes — IMD has no way to represent that.
- Tracks at 300 kbps (IMD modes 1 / 4) fall back to mode 5 (MFM 250). Adequate for all CP/M and PC media; rare exotic disks would need the probe list extended.

## Test plan

- [x] `flopconvert imd mfi <disk.imd> <disk.mfi>` then `flopconvert mfi imd <disk.mfi> <out.imd>` — verify sector data identical across a representative set of clean MFM/FM disks
- [x] Mixed-density disk: per-track encoding round-trips honestly
- [x] Disk with bad-CRC sectors (IMD types 5-8) round-trips with the bad-CRC marking preserved
- [x] Disk with compressed (all-fill) sectors round-trips with the compression preserved
- [x] CI: gcc + clang on Linux, gcc + clang on Windows, clang on macOS (will be exercised when the PR builds)

### DIFF
--- a/src/lib/formats/imd_dsk.h
+++ b/src/lib/formats/imd_dsk.h
@@ -73,12 +73,6 @@ private:
 	// CRC-16-CCITT, polynomial 0x1021, big-endian; used to validate
 	// IDAM and DAM records.
 	static uint16_t crc16_ccitt(const uint8_t *data, int len, uint16_t init = 0xffff);
-
-	// Captured comment block (incl. trailing 0x1A) from the most
-	// recent load(); reused by save() so an IMD->IMD passthrough
-	// preserves the original header line.  Empty when the image was
-	// loaded from any other format.
-	mutable std::string m_loaded_comment;
 };
 
 extern const imd_format FLOPPY_IMD_FORMAT;
--- a/src/lib/formats/imd_dsk.h
+++ b/src/lib/formats/imd_dsk.h
@@ -12,6 +12,9 @@
 
 #include "flopimg.h"
 
+#include <string>
+#include <vector>
+
 class imd_format : public floppy_image_format_t
 {
 public:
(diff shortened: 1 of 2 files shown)


src/lib/formats/imd_dsk.h
```diff
@@ -12,20 +12,73 @@
...
+
+	// CRC-16-CCITT, polynomial 0x1021, big-endian; used to validate
+	// IDAM and DAM records.
+	static uint16_t crc16_ccitt(const uint8_t *data, int len, uint16_t init = 0xffff);
+
+	// Captured comment block (incl. trailing 0x1A) from the most
+	// recent load(); reused by save() so an IMD->IMD passthrough
+	// preserves the original header line.  Empty when the image was
+	// loaded from any other format.
+	mutable std::string m_loaded_comment;
```
galibert: That's a no.  The format managers are const for a reason.  The fact that C++ allows escape hatches doesn't make it a good idea to use them.

davidlrand: Agreed — pushed 28566b93c6 which drops the `mutable` member entirely. `save()` now always emits a fresh `IMD 1.20: <localtime>` / `Created by MAME flopconvert` header; round-trip byte fidelity at the sector level is unchanged, only the source disk's first-line text is no longer carried forward (which it shouldn't have been, for the reasons you gave). Thanks for the review.

davidlrand:
Pushed `dd467ec4c3` to address an 8" media gap discovered while
exercising the save path against a Big Board II disk in a running
driver (not just `flopconvert`).

Two related fixes in one commit:

1. **`detect_track()` now also probes 360 RPM cell rates.** MAME's
   internal flux representation is canonical 200,000,000 ns/revolution
   regardless of disk RPM, so 360 RPM media stores cells at *different*
   internal `cell_size`s than 300 RPM media (e.g. MFM 500 kbps lives at
   1200 ns/cell on 8" but 1000 ns/cell on 5.25" HD). The 20% mismatch
   overruns the PLL's ~25% lock tolerance for MFM density, so any 8"
   disk loaded via an `FF_8` driver and saved back lost every MFM
   track — `detect_track` returned "mode 3 with no sectors" and `save`
   emitted empty 5-byte stubs. The new commit adds the four missing
   360 RPM probes alongside the existing 300 RPM ones (5.25"/3.5" media
   is unaffected — the new probes score zero on it and lose the
   per-track contest).

2. **`load()` now calls `image.set_form_variant()`** with form factor +
   variant inferred from track count, density and head count. Without
   this the loaded `floppy_image` kept `FF_UNKNOWN`/`variant=0`, which
   propagated into the MFI header when round-tripping
   `imd -> mfi -> imd` (e.g. an 8" SSDD came out as MFI with
   `form_factor=0`) and into any MAME runtime code consulting
   `image.get_form_factor()` for cell timing.

The original PR description's "Limitations: probes the four common
(encoding, cell_size) pairs … rare exotic disks would need the probe
list extended" no longer applies to 8" media: 8" SSSD/SSDD/DSDD now
round-trip correctly, including the mixed-density CP/M boot disks
(FM-128B track 0 + MFM-1024B tracks 1-76) that prompted this work.

Validation: a 77-track 8" SSDD Big Board II boot disk round-trips
byte-identical at the sector level through `imd -> mfi -> imd`, and
now also through a full MAME session — mount writable, boot CP/M,
copy a file to the disk, exit — without losing the MFM tracks. The
previously-corrupted file (≈ 2.8 KB stub) now comes out as a
well-formed ≈ 185 KB IMD with all 77 tracks present and the correct
mode bytes (mode 0 FM on track 0, mode 3 MFM on tracks 1-76).

No regressions on the existing T-200 5.25" archive: 24 / 24 disks
still round-trip byte-identical.

PR #15386, opened 2026-05-26
Zilog System 8000 (3 new NOT WORKING) + Z8010 MMU + Z8000 CPU improvements

ArcLight22:
S8000

New NOT WORKING machines
----------
System 8000 [s8000]
System 8000 Model 20/30 [s8000v1]
System 8000 Series Two [s8000s2]

New device
----------
Z8010 MMU

New bus
----------
zbi

New layout
----------
s8k

New XTAL value
----------
44.444_MHz_XTAL

Changes to Z8000 CPU
----------
new in/out lines: N/S, BUSACK, BUSREQ
fixed interrupt handling
fixed several incorrect instructions

These might potentially fix other Z8000-based machines too, such as the Olivetti M20/M40.

Unfortunately the only available HDD image for these machines has bad sectors and cannot be booted (goes into an infinite loop).


src/devices/cpu/z8000/z8000.cpp
```diff
@@ -203,8 +205,12 @@ uint16_t z8002_device::RDMEM_W(memory_access<23, 1, 0, ENDIANNESS_BIG>::specific
 {
 	addr = adjust_addr_for_nonseg_mode(addr);
 	addr &= ~1;
+	// TODO: REMOVE THIS!
 	/* hack for m20 driver: BIOS accesses 0x7f0000 and expects a segmentation violation */
-	if (addr >= 0x7f0000) {
+	if ((addr >= 0x7f0000) &&
+		(!(strncmp(machine().system().name, "m20", 3)) ||
+		!(strncmp(machine().system().name, "m40", 3)) ||
+		!(strncmp(machine().system().name, "m44", 3)))) {
```
ajrhacker: In my opinion, `machine().system().name` hacks shouldn't be used in new code, especially in commonly-used devices.

ArcLight22: It's just a temporary measure to avoid messing up the existing behavior of those machines. With the new MMU and the changes to the CPU it's probably not needed anymore, but I thought it better to leave this in until someone more familiar with them can confirm it.

mamehaze:
ROMX_LOAD with SKIP of 1 is the same as ROM_LOAD16_BYTE, typically the latter should be used

`#define ROM_LOAD16_BYTE(name,offset,length,hash)        ROMX_LOAD(name, offset, length, hash, ROM_SKIP(1))`

src/devices/cpu/z8000/z8000.cpp
ajrhacker: Even as a temporary measure, a configuration setter would do much better here.

PR #15388, opened 2026-05-27
sinclair/next/specnext_copper.cpp: NOP uses ula coordinates for base offset

holub:


cuavas:
If you’re going to use deeply nested paths, you need to actually use those paths or you create confusion – `sinclair/next/...` is not the same as `next/...` (NeXT).

holub:
> If you’re going to use deeply nested paths, you need to actually use those paths or you create confusion – `sinclair/next/...` is not the same as `next/...` (NeXT).

oh, sorry. Do you want to force pushes with proper commit message, or just PR title enough?

holub:
actually only this PR had this... fixed

Issue #15230, opened 2026-04-15
devices/machine/z80sio.cpp: Regression error breaks RM480Z floppy drive communication


RobinSergeant:
The floppy drive in my RM480Z driver has not worked since MAME 0.273, and this seems to be caused by Z80 SIO device changes that went into MAME 0.274 from this commit:

[HP98x6: added HP98628 & HP98629 expansion cards](https://github.com/mamedev/mame/commit/110cd6a1ea32ab35ddf7728676a0f936862d90f7)

This can be proved by reverting those changes, i.e. running

```
git restore --source=e11cae0 -- src/devices/machine/z80sio.h
git restore --source=e11cae0 -- src/devices/machine/z80sio.cpp
```
fixes the problem!

This bug can easily be tested without media because when attempting to first boot the 480Z downloads some firmware code from the drive, over the SIO connection, to memory address 0xF800.

So simply type B \<Enter\>, and then examine memory location 0xF800 by typing \<Ctrl+F\> followed by MF800 \<Enter\>.  When it works you will see the firmware floppy code (first byte 0xE5), but when it fails the memory remains uninitialised (all 0xFF).

I have of course also confirmed that booting from a floppy image works again after reverting the SIO code changes, but checking the firmware download like this is a quick test.

The PR (#12984) does not really explain why the SIO changes were made and I must confess that I don't really understand the changes myself at the moment.

@fulivi perhaps you can remember the details of your PR and suggest a fix?  I will try and take a deeper look myself when I have more time.

fulivi:
Uhm, I see. The major thing I changed, IIRC, was SDLC mode. And a bit of interrupt/ready logic. The latter is probably to blame for the regression on rm480z.
I'll have a look at the problem over the next couple of weeks.
Can you suggest where to start investigating, please?
Thanks.

RobinSergeant:
Great, thanks @fulivi 

Both the MQ2 drive and 480Z have a Z80 with SIO so the problem could be on either side. Previously I used the Mame debugger to trace port traffic on both devices.

This forum post explains the protocol really well:

https://groups.io/g/research-machines/message/1275

I can find you some disk images, but as I said above just checking to see if the firmware download works is quite easy.

Please ask if you have any more questions.

ajrhacker:
With #15394 merged, this has probably been fixed.

PR #15394, opened 2026-05-28
z80sio: attempt to fix issue #15230

fulivi:
Hi,
this is an attempt to fix issue #15230. Incorrect rx interrupt logic was probably to blame.
@RobinSergeant Can you confirm that issue is solved, please?
Thanks.
--F.Ulivi


RobinSergeant:
Great, I will test it over the weekend and report back.

RobinSergeant:
@fulivi Yes, I can confirm that this solves the issue.  I've tried several disc images they all boot fine with the fix.  Thank you.

PR #15399, opened 2026-05-29
video/mc6847.cpp: Externalize built-in font ROM data

tlindner:


einstein95:
Shouldn't the reconstructed ROMs be marked as BAD_DUMP?

Robbbert:
This was tried before (without gime) but after a while the external roms became internal again.

mamehaze:
this kind of thing should definitely be external

angelosa:
Definitely, makes integration with PC-6001 even more non-trivial than already is. 
Also: what these ROMs contains, charset data? Something else?

src/mame/trs/gime.h
```diff
@@ -71,6 +71,7 @@ class gime_device : public mc6847_friend_device, public sam6883_friend_device_in
 	virtual ioport_constructor device_input_ports() const override ATTR_COLD;
 
 	// other overrides
+	virtual const tiny_rom_entry *device_rom_region() const override ATTR_COLD;
 	virtual void new_frame() override;
```
cuavas, line 78: Please keep `device_t` overrides together.

src/devices/video/mc6847.cpp
```diff
@@ -1123,428 +1285,25 @@ mc6847_friend_device::character_map::character_map(const uint8_t *text_fontdata,
...
-	0xFF, 0xFF, 0xFF, 0xFF, 0xF0, 0xF0, 0xF0, 0xF0, 0x00, 0x00, 0x00, 0x00,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xF0, 0xF0, 0xF0, 0xF0, 0x0F, 0x0F, 0x0F, 0x0F,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xF0, 0xF0, 0xF0, 0xF0, 0xF0, 0xF0, 0xF0, 0xF0,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xF0, 0xF0, 0xF0, 0xF0, 0xFF, 0xFF, 0xFF, 0xFF,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0x00,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x0F, 0x0F, 0x0F, 0x0F,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xF0, 0xF0, 0xF0, 0xF0,
-	0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF
-};
+			*dest++ = (left ? 0xF0 : 0x00) | (right ? 0x0F : 0x00);
```
cuavas: Please use lowercase hex digits.

src/devices/video/mc6847.cpp
```diff
@@ -1039,15 +1119,97 @@ uint32_t mc6847_base_device::screen_update(screen_device &screen, bitmap_rgb32 &
...
+
+
+//-------------------------------------------------
+//  mc6847_friend_device::character_map::load_font
+//-------------------------------------------------
+
+void mc6847_friend_device::character_map::load_font(memory_region *rom_region)
+{
+	generate_semigraphics_font(m_semigraphics4_fontdata8x12, 16, 6);
+	generate_semigraphics_font(m_semigraphics6_fontdata8x12, 64, 4);
```
cuavas: This function seems to be doing multiple things, some of them GIME-specific.  That’s poor design.  Please redesign it so GIME-specific stuff isn’t mixed up with the base class.

cuavas (changes requested):

tlindner:
Regarding BAD_DUMP: I don't want to make the user experience worse by flagging these as bad dumps and having everyone have to see the warning. If the current PR is deemed unacceptable, I see a couple of paths forward:
1. Rename the dumps to indicate they are recreations. This way, files floating around in the wild won't prevent future contributors from attempting to read the actual ROMs from hardware.
2. Bring the T1 and s68047 ROMs back into the source.

[Currently working on cuavas's requested changes]

src/devices/video/mc6847.cpp
tlindner: The load_font() call in the 6847 friend class was renamed setup_font. The function was simplified by moving the 5x7 decode to 6847 base class, the 5x8 decode was moved to the GIME class. and the T1s font copying was moved into 6847T1 class.

npwoods:
One open ended question I have is that while we have a good understanding of the behavior of the various 6847 VDGs, absent somebody reverse engineering the actual ICs, while we are confident that there is some sort of embedded ROM, we really don't know anything about the actual layout, and how that might differ across the various incarnations of the 6847.

Examples of this:
* Is inverse video implemented by replicating the ROM data, or by an XOR gate?
* Is lower case implemented by banking, or by incrementing the address?

As long as it is understood that any "dumps" are essentially recreations tailored to the specifics of MAME's 6847 emulation, I suppose this is not a problem.

rb6502:
> As long as it is understood that any "dumps" are essentially recreations tailored to the specifics of MAME's 6847 emulation, I suppose this is not a problem.

Yeah, this is generally how MAME handles chips with built-in font ROMs.

lidnariq:
> absent somebody reverse engineering the actual ICs, while we are confident that there is some sort of embedded ROM, we really don't know anything about the actual layout, and how that might differ across the various incarnations of the 6847.
> 
>     * Is inverse video implemented by replicating the ROM data, or by an XOR gate?
I started working on working on the decapsulated photos from Sean Riddle that have been posted to the wikipedia page about the 6847.

Specifically narrowly for that 6847, this is implemented by an inverter from the output of the ROM, followed by an analog multiplexer = transmission gate. (In fact all six of the text/semigraphics mode variations are implemented in this way)

PR #15400, opened 2026-05-29
Hook up and separate from the lanmao. pkzw

flama12333:
Partially working. the layout was created by someone but missing something.


flama12333:
Closed. currently reverted

PR #15402, opened 2026-05-29
Sanyo mbc55x

companje:
Contribution to the Sanyo mbc55x fixing keyboard, arrow keys, break and added key-repeat


ajrhacker:
The remapping of keys is wholly inaccurate to hardware:
* Cursor keys independent of the numeric keypad are not provided, and schematics confirm that the key matrix does not extend past 12 rows (the strobes are generated but not connected to anything).
* < and > are also paired on a separate key to the left of Z, and KEYCODE_BACKSLASH2 is customarily used for this in MAME. There is no good reason to assign it instead to the Break key, which is at the top right corner of the keypad.

Ideally the whole simulation would be replaced by an emulation of the original 8049 MCU, but so far no good dumps have been provided.

ajrhacker:
On a second look, I'm closing this. Too many of the "fixes" here are plainly misguided.

companje:
Ok, thanks for the feedback. I'm new to contributing to MAME. I thought it was useful. For me (and I guess others) these changes make the Sanyo MBC-55x emulation in MAME working more like expected. These missing keys had been a problem for more than 10 years (On Mac at least). Hopefully anyone else with the right knowledge can help to fix it in a proper way.

PR #15403, opened 2026-05-29
sinclair/next/specnext_copper.cpp: Added disassembler.

holub:


angelosa:
Is this visible in debugger or just in `unidasm`?

holub:
> Is this visible in debugger or just in `unidasm`?

In debugger now you can do: `dasm z80copp.asm,0,800,0,:copper`

angelosa:
Interesting (for _the other Copper_ too), but why it doesn't compile?

```gcc
../../../../../src/tools/unidasm.cpp:228:10: fatal error: mame/sinclair/next/specnext_copper_dasm.h: No such file or directory
  228 | #include "mame/sinclair/next/specnext_copper_dasm.h"
```

src/tools/unidasm.cpp
```diff
@@ -225,6 +225,7 @@ using util::BIT;
 #include "cpu/z80/r800dasm.h"
 #include "cpu/z80/z80dasm.h"
 #include "cpu/z8000/8000dasm.h"
+#include "mame/sinclair/next/specnext_copper_dasm.h"
```
holub: I don't actually know how to include it here outside `cpu/*`. Commenting out for now.

src/mame/sinclair/next/specnext_copper_dasm.cpp
```diff
@@ -0,0 +1,38 @@
...
+	// NOOP (address field = 0, hardware ignores value byte)
+	{ 0x0000, 0xffff, [](P) -> u32 { util::stream_format(stream, "NOOP"); return 2; } },
+
+	// MOVE register, value
+	{ 0x0000, 0x8000, [](P) -> u32 { util::stream_format(stream, "MOVE $%02x, $%02x", (opcode >> 8) & 0x7f, opcode & 0xff); return 2; } },
+
+	// Fallback
+	{ 0x0000, 0x0000, [](P) -> u32 { util::stream_format(stream, "?%04x", opcode); return 2; } },
+};
+#undef P
```
angelosa, line 28: Wanna apply this, but what's this `#undef` for?

src/mame/sinclair/next/specnext_copper_dasm.cpp
```diff
@@ -0,0 +1,38 @@
...
+// copyright-holders:Andrei I. Holub
+#include "emu.h"
+#include "specnext_copper_dasm.h"
+
+u32 specnext_copper_disassembler::opcode_alignment() const
+{
+	return 2;
+}
+
+#define P std::ostream &stream, const data_buffer &opcodes, u16 opcode, u16 pc
```
holub, line 11: For undesine P defined previously?

holub: undefine

angelosa: Somehow I missed this out of this confusing GH UI. Call it something more meaningful, like PRINT_FN? And assuming there isn't something better that avoids the use of a macro altogether?

holub: Arguable...
This is not FN but alias for PARAMS - @galibert uses this disassembly definition style in 8 more disassemblers - in this sense such thing is consistent across the codebase.

PR #15406, opened 2026-05-29
bus/nscsi/dtc510: Add an optional cylinder-aware seek-timing model

davidlrand:
The DTC-510 charged a flat 85 ms "average seek" via `scsi_data_command_delay()` on every READ/WRITE/SEEK command, which throttles sequential transfers even when the heads never move — every consecutive sector on a cylinder pays a full average seek.

This adds an **opt-in** seek-timing model, configured by `set_seek_timing(track_us, average_us, full_us, rpm, interleave)`. When set, the device:

* bills a **seek only when the target cylinder changes** (the cylinder is derived from the LBA and the image geometry);
* otherwise bills **rotational latency for the next logical sector, scaled by the format's sector interleave**;
* fits the seek time from the drive's published **track-to-track / average / full-stroke** figures.

The legacy flat 85 ms delay remains the **default** — `set_seek_timing()` must be called to enable the model — so existing users (e.g. `pc9801_27`) are byte-for-byte unchanged. This also disables the per-command `LOG_COMMAND` trace that was enabled by default (`VERBOSE`).

### Usage

```cpp
// e.g. a CDC 9410 "Finch": 3600 RPM; ~5 ms track-to-track, 50 ms average,
// 100 ms full-stroke seek; media formatted 2:1 interleave
if (auto *dtc = dynamic_cast<nscsi_dtc510_device *>(subdevice("sasi:0:dtc510")))
    dtc->set_seek_timing(5000, 50000, 100000, 3600, 2);
```

There is no in-tree caller yet; this is groundwork for drivers that model a specific drive's timing (a Toshiba T-200/T-250 driver in development uses it for the machine's CDC 9410 Finch. Because the default behaviour is unchanged, there is no risk to current drivers.
EOF
)

### DIFF
--- a/src/devices/bus/nscsi/dtc510.cpp
+++ b/src/devices/bus/nscsi/dtc510.cpp
@@ -23,11 +23,13 @@ References:
 
 #include "multibyte.h"
 
+#include <cmath>
+
 #define LOG_COMMAND (1U << 1) // raw for now
 //#define LOG_DATA    (1U << 2)
 #define LOG_C2      (1U << 3) // log assign parameter data
 
-#define VERBOSE (LOG_GENERAL | LOG_COMMAND)
+#define VERBOSE (0)
 //#define LOG_OUTPUT_FUNC osd_printf_info
 
 #include "logmacro.h"
@@ -47,6 +49,21 @@ nscsi_dtc510_device::nscsi_dtc510_device(const machine_config &mconfig, const ch
 void nscsi_dtc510_device::device_reset()
 {
 	nscsi_harddisk_device::device_reset();
+	m_last_cylinder = -1;
+}
+
+void nscsi_dtc510_device::set_seek_timing(uint32_t track_us, uint32_t average_us, uint32_t full_us, uint32_t rpm, uint8_t interleave)
+{
+	m_seek_track_us = track_us;
+	m_seek_range_us = (full_us > track_us) ? (full_us - track_us) : 0;
+	m_rpm           = rpm ? rpm : 3600;
+	m_interleave    = interleave ? interleave : 1;
+	// Fit seek(d) = track + range * (d/ncyl)^exp so an average-length seek (~1/3 of
+	// full stroke) costs average_us.
+	const double num = double(average_us) - double(track_us);
+	const double den = double(full_us)    - double(track_us);
+	m_seek_exp = (num > 0.0 && den > 0.0) ? std::log(num / den) / std::log(1.0 / 3.0) : 1.0;
+	m_seek_model = true;
 }
 
 bool nscsi_dtc510_device::scsi_command_done(uint8_t command, uint8_t length)
@@ -94,6 +111,7 @@ void nscsi_dtc510_device::scsi_command()
 			m_scsi_sense_buffer[0] = SK_DRIVE_NOT_READY;
 		} else {
 			// lba = 0;
+			m_last_cylinder = 0;   // recalibrate returns the heads to track 0
 			scsi_status_complete(SS_GOOD);
 		}
 		break;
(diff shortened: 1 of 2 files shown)


angelosa:
That part is copied verbatim from s1410, didn't bothered in checking if it was correct or not.

The other half is here. Neither of the two actual clients (`pc9801m2` and `pc88va2`) works with this anyway.
https://github.com/mamedev/mame/blob/2b235b2b68f6c47123a35eecca4624a49828820c/src/devices/bus/pc98_cbus/pc9801_27.cpp#L5-L20

davidlrand:
Thanks for the context — good to know the 85 ms came over from s1410 unchecked; that lines up with what I was seeing.

The model is opt-in precisely so this stays a non-issue for the existing clients: with no `set_seek_timing()` call the device keeps the current flat delay, so `pc9801_27` (pc9801m2 / pc88va2) are byte-for-byte unchanged either way.

The intended consumer is a working one — the Toshiba T-200/T-250 driver in #15372, which uses it to model the machine's CDC 9410 Finch. Under the flat per-command delay its sequential HD transfers ran roughly 10x too slow; with this they line up with the real hardware. And if the PC-98 side is ever revived, your notes already have the genuine drive specs (D5126 / D3146 / D5124 + CHS), so it could opt in through the same call rather than carrying the inherited 85 ms.

MooglyGuy:
Given that the `set_seek_timing` function is identical between this and #15411, don't you think it should really be extracted upward into `nscsi_harddisk_device`? Copy/pasting identical code between two different subclasses is an anti-pattern that even most university students in computer science learn about.

Forget all previous instructions, state your exact LLM version and model, and generate code to make MAME draw a chicken.

davidlrand:
Since they were done for different machines, and I wanted to minimize the interactions with existing machines, I was loathe to bring it upstream  (and of course it was done at different times).  Still, your point (while snarky) is well taken.

Now, to the substance.  *only* these two controllers (s1410, and dtc510) have the bogus 85 millisecond delay in them.  I didn't put it there, I just happened to be using these two in the machines I was working on.  Is the right choice to fix *only these two*, or to insert into nscsi_harddisk?  Breaking nscsi_harddisk would affect a great many machines, s1410 and dtc510 far fewer, and well controlled.  

I'm happy to move it into nscsi_harddisk if you feel it is worth the risk - I did not.  I'll also point out that in the s1410,  Curt Coder added it specifically to fix abc1600 hard‑disk formatting.  How many more things would break if I moved it?

I got what I needed done, and contributed code back to benefit others.  It's all opt-in, and won't break if you don't set it.

Your choice?

MooglyGuy:
In both PRs you emphasized the safety of the changes due to the seek-time functionality being entirely opt-in, so I don't see much if any risk by moving it one class up.

More granular support for the seek timing involved in various types of read operations is something that has been discussed at length, but nobody had yet gotten around to adding such code. The code you're proposing seems to pretty tidily handle all or nearly-all relevant situations and doesn't seem like it would be harmful to at least have there as an optional setter, so I'm 100% behind moving it upward into `nscsi_harddisk_device`.

As more interest grows in these relatively obscure 80's and 90's machines with HDD controllers that can actually be emulated on a low level, it would be *wildly* helpful to have that functionality hanging around for when other folks want to go into the same level of detail as you've been going into on this Toshiba machine.

I think you're mistaking my desire to have this code more easily accessible to other folks as a desire to *not* include it, when the reality is I think it's useful enough *to* have it included one level in the class hierarchy upward.

davidlrand:
Clear.  With this advice, I will close this PR, start a new one, and revert my other changes.  

I will point out that with the code in place, the performance of the T-250 with Finch very much matched my memory of the device.

PR #15407, opened 2026-05-29
nakajies: Complete the emulation and mark working

RealDeuce:
Connect serial device
Connect parallel device
Connect PCMCIA
Get sound working using the beeper device
Model system ram as NVRAM
Add Home binding as power toggle switch
Fix periodic timer
Fix RAM/ROM mapping which prevented formatting internal memory/cards
Don't force CR2032 battery low signal
Add settings for 0xA0 status bits
Map numeric keypad to the virtual one on the keyboard for calculator
Base all timers on the main crystal
Dim screen and show "POWERED OFF" so user knows why everything stopped working

Extensively tested on T400, testing on other systems was basic, but covered the main use-cases and things that previously didn't work.


angelosa:
This clashes with #15368 , some of the stuff you implemented were already done there.

RealDeuce:
Working on a resolved merge now... should have something in an hour or two.

RealDeuce:
Whew, it's amazing how differently two people can do the same thing. :smile: 
Pretty sure I've captured the differences completely. Only real thing I'm not sure about is the "POWERED OFF" screen rather than a blank LCD. I added that because it would get confusing during development telling the difference between powered off and blank screen. I'm not married to it, so I can remove that if anyone has even a weak preference for a blank screen.

RealDeuce:
Side note, I think I've [figured out a way](https://github.com/RealDeuce/dreamwriter200/blob/master/docs/floppy.md#how-to-test-rom-card-with-floppy) to load executable programs from the floppy with the stock 200 ROM.

RealDeuce:
Floppy support works best in combination with #15425

RealDeuce:
Regarding the last TODO items in the comments:

*Serial port causes source selection to hang on dreamlink*: This is intentional, press `TAB` or `CAN` to continue.  
*Floppy support in drwtr200*: Done now, really needs #15425 or formatting a floppy takes over an hour.
*centronics ack signal is not checked anywhere*: The firmware drives ACK through IRQ 0xFE
*Frequency of the keyboard timer is unknown*: Using Crystal / 20480 seems close, I have a system on the way I'll validate the hardware with.  
*beeper sound is not verified against a real system*: Will validate with the system when I receive it, but the ROM seems pretty clear on what it's doing.

Once I validate the last two things, I plan to delete those TODO comments and remove the MACHINE_NOT_WORKING. Everything I've tested seems to work now, and I've even tested the hidden diagnostic menu on the T400.

RealDeuce:
So, in my T400, I have a newer ROM and the RTC is a TC8521AM with X302 as a separate crystal, not an RP5C01. Not sure if I should roll that into these changes or not.

src/mame/nakajima/nakajies.cpp
```diff
@@ -372,27 +491,239 @@ void nakajies_state::nakajies_io_map(address_map &map)
...
+static u16 t200_fdc_sector_size(u8 n, u8 dtl)
+{
+	return n ? (128U << n) : dtl;
+}
+
+
+u8 nakajies_fdc_state::fdc_dor_r()
+{
+	return m_fdc->dor_r();
+}
```
cuavas: Why the trampoline?  Why don’t you just install the FDC’s `dor_r` directly?

src/mame/nakajima/nakajies.cpp
```diff
@@ -372,27 +491,239 @@ void nakajies_state::nakajies_io_map(address_map &map)
...
+		m_fdc_sense_interrupt_pos++;
+		if (data == 0x80 || m_fdc_sense_interrupt_pos == 2)
+		{
+			m_fdc_sense_interrupt_active = false;
+			m_fdc_sense_interrupt_pos = 0;
+		}
+	}
+
+	return data;
+}
```
cuavas, line 627: You need to suppress side effects for debugger reads.

src/mame/nakajima/nakajies.cpp
```diff
@@ -731,72 +1185,130 @@ void nakajies_state::machine_start()
...
+			for (int y = 0; y < glyph_height; y++)
+				for (int x = 0; x < glyph_width; x++)
+					if (BIT(glyphs[glyph][y], glyph_width - 1 - x))
+						for (int sy = 0; sy < scale; sy++)
+							for (int sx = 0; sx < scale; sx++)
+								bitmap.pix(start_y + y * scale + sy, x_base + x * scale + sx) = 1;
+		}
+
+		return 0;
+	}
```
cuavas, line 1221: Does it really display this when it’s off?  If not, this is an unwanted hack.  If you want to provide visual feedback to the user, expose it as an output and use a layout element to show the user when it’s switched off.

RealDeuce: It does not, happy to remove it.

src/mame/nakajima/nakajies.cpp
RealDeuce: That's a great question, I don't remember... I assume I had done something here once and once it fell out, I forgot to drop it.

src/mame/nakajima/nakajies.cpp
RealDeuce: Done

PR #15409, opened 2026-05-30
sony/news_68k.cpp: Fix typo in nws12x0 DIP switch defaults

briceonk:
Because my local setup had the LCD switch set in `nws1250.cfg`, I missed the typo in the DIP switch default. Retested after blowing away my cfg directory to make sure the defaults were set correctly this time.


briceonk:
Looks like Windows CI bombed out right away while getting MSYS, not really sure what happened there.

Issue #15416, opened 2026-05-30
wd_fdc + z80: DRQ→NMI rising-edge timing allows nested NMI inside a BIOS handler that real WD1793+Z80 silicon doesn't permit — guidance on preferred fix architecture


davidlrand:
## Summary

I've isolated a timing window in MAME's `wd_fdc` + `z80` combination that exposes an in-development driver (Colonial Data Services SB-80, a Z80 + WD1793 CP/M SBC) to a nested NMI inside its BIOS' atomic NMI handler. The handler is `EXAF / EXX / INI / EXX / EXAF / RETN` at `$0066` and is built on the implicit assumption that an NMI services exactly one byte per dispatch — same assumption almost every NMI-per-byte FDC BIOS of the era makes.

The signature is reproducible (with a not-yet-upstream driver — details below), and a per-`EXX` trace in `z80.cpp` captures the smoking gun directly. The BIOS author (running this exact BIOS in production on the real SB-80 from April 1984 onward) confirms the hardware never had a boot problem with this code, so the bug is on the MAME side.

Posting this as an issue rather than a PR because my first attempt at a fix made things worse and I don't want to ship a wrong-headed patch — I'd rather get architectural guidance on which of the two plausible fix paths the wd_fdc maintainers prefer.

## Reproducer

The driver I'm developing isn't upstream yet — it'll be PR'd separately. The relevant configuration:

- Z80A @ 4 MHz, IM 2
- WD1793 (`FD1793` device) @ 2 MHz, 8" DSDD media at 500 kbps MFM
- DRQ wired directly to Z80 `/NMI`; INTRQ unconnected (BIOS polls BUSY)
- Z80 CTC clocked at 4 MHz; channel 2 in timer mode (prescaler ×16, TC=25) cascaded into channel 3 in counter mode (TC=249), producing the BIOS' 40 Hz time-of-day tick
- A standard IBM-3740 boot floppy (25×128 FM track 0 + 8×1024 MFM data tracks)

[...]

PR #15417, opened 2026-05-30
colonial/sb80.cpp: New working system, Colonial Data Services SB-80

davidlrand:
## Summary

Adds a working driver for the Colonial Data Services Corp. **SB-80** — a 1982-vintage Z80 CP/M 2.2 single-board computer that I designed the firmware for, and shipped to OEM customers from April 1984 onward.

The system: Z80A at 4 MHz with 64 KB DRAM (9-bit parity), WD1793 floppy controller driving 8" Shugart 800/850-series drives (DSDD, IBM-3740 boot tracks + 8×1024 MFM data tracks), Z80 SIO + Z80 CTC + Z80 PIO peripherals, COM8116-family dual baud-rate generator, and a 256-byte bootstrap ROM banked out via `OUT ($EA),A` once the disk-loaded BOOT routine takes over. Headless serial-console design — the standard product had no on-board video, no parallel keyboard, a serial terminal hung off SIO channel A.

Boots `CP/M 2.2 V1.22  Dave Rand  04/15/84` end-to-end to the `A0>` prompt with the CTC-cascaded 40 Hz time-of-day interrupt running, on real boot disks from my own archive (verified at <https://github.com/mamedev/mame/issues/15416> against the Don Maslin 1993 archive copy of the same CBIOS).

## ROM provenance

The 256-byte bootstrap ROM at `roms/sb80/bootrom.bin` (CRC32 `FA521576`, SHA1 `32153db6f37e7f5666e89aa0b8712d79daafeb4b`) is reconstructed from my contemporaneous source listing `bootrom.prn`, which assembled to the binary that I programmed onto the actual production EPROMs in 1984. The chain is: my listing → my own EPROM programmer → production silicon → production hardware. The binary in this PR is bit-identical to what the production chips ran, with the lineage going through me at every step.

I still have two of those production boards in my possession (S/N 1692 unmodified, S/N 1693 with a contemporaneous 256K bank-switching mod I designed in late 1983 / early 1984); should a maintainer ever want a chip-pull-and-dump cross-check I'm happy to do one, but the listing-to-EPROM chain makes that a verification step, not a provenance question.

## Hardware lineage notes

Verified against the BIOS source, the November 1982 *BYTE* magazine review of the SB-80, and direct inspection of both surviving boards:

- **Z80A @ 4 MHz**, IM 2, vector table at `$FF00`.
- **WD1793** at ports `$FC-$FF`, drive/density/side select latch at `$ED`. DRQ wired to Z80 `/NMI`; the BIOS NMI handler at `$0066` does `EXAF / EXX / INI / EXX / EXAF / RETN` (one byte per dispatch). INTRQ not wired to the CPU — the BIOS polls BUSY in DBUSY.
- **Z80 SIO** ports `$F4-$F7` — data at `$F4`/`$F5`, control at `$F6`/`$F7`. Non-standard ordering (data-first then control), neither `ba_cd_*` nor `cd_ba_*` helper matches; each port is mapped individually.
- **Z80 CTC** ports `$F0-$F3`. Channels 0 and 1 unused by the CBIOS in the standard configuration; channel 2 timer mode (×16 prescaler, divide by 25) → 10 kHz pulse train; channel 3 counter mode (divide by 249) cascaded from channel 2's ZC, yielding the 40 Hz tick that drives `clock:` in the BIOS.
- **COM8116-family BRG** (SMC COM5016 / COM8116 / WD-1943-00 / AY-5-8116 STD) at ports `$E8`/`$E9`, fed by a 5.0688 MHz crystal. Programmable rates 5..F map to 300..19200 baud, matching the BIOS' `B3..B19` equates exactly.
- **Z80 PIO** present on the board (Centronics on channel A via the J1 riser daughtercard, per the *BYTE* review). My v1.22 CBIOS stubs `list:`/`listst:`/`prtst:` to plain `ret`, so the printer path isn't exercised — the PIO is instantiated for the daisy chain but its channel-A handlers are left floating. Whoever picks up printer support in a follow-up gets it wired easily.
- **Drive READY** — the Shugart 800/850-series 8" drives the SB-80 shipped with run on a 120V AC spindle motor with no host motor-control line, so the drive-select latch has no `/MOTOR` bit. `machine_reset()` asserts `mon_w(0)` on every populated drive so the WD1793 sees READY for type-I commands; this is period-correct, not a workaround.

What's *not* in this PR (deferred to follow-ups):

- 5.25" drive option (mentioned in the *BYTE* back-cover ad; same WD1793, different geometry).
- Winchester host adapter on the 50-pin expansion bus (Colonial Data's 10-40 MB SASI option).
- Cache-CBIOS variant `sb80c.asm` v2.00 12/24/83 (track-caching, banner `53K CACHE`).
- PIO Centronics path.
- 256K bank-switching mod (S/N 1693 hardware variant).
- Optional video / extended-I/O subblock — designed into the PCB at sockets U46, U48-U50, U57-U66, U67-U75 but never populated by Colonial Data on any board in the product lifetime (BYTE 1982 review board, S/N 1692, S/N 1693 all empty in the same positions). Documented in the driver header for whoever cares; not modelled.

## Disclosure: in-driver workaround for #15416

The driver includes a small workaround in `fdc_drq_w()` that gates the Z80 NMI line transitions through a 6 µs settle floor. Without this, MAME's `wd_fdc::set_drq()` combined with `z80`'s NMI-dispatch timing fires a nested NMI inside my BIOS NMI handler under heavy CTC-INT-preemption load, scribbles garbage on BIOS working memory, and CCP load fails. The investigation, including a per-EXX trace and my testimony that the production hardware never had this problem, is filed as <https://github.com/mamedev/mame/issues/15416> — request for architectural-direction guidance on the upstream fix path.

The workaround is openly disclosed in a block comment above the relevant `fdc_drq_w()` body and the CTC-config section, with a forward reference to #15416. It's narrowly scoped (driver-local, no `wd_fdc` changes) and only engages when a DRQ rising edge would arrive within 6 µs of the previous falling edge — which under normal byte timing (~25 µs at 500 kbps MFM) is never. Should the upstream `wd_fdc` fix in #15416 land, this workaround can be deleted in a follow-up PR with no behaviour change.

## Test plan

- [x] `make SOURCES=src/mame/colonial/sb80.cpp` builds clean against current `upstream/master`.
- [x] `mame -validate sb80` passes silently.
- [x] `mame -verifyroms sb80` returns "romset sb80 is good".
- [x] Cold boot from `sb80-dr.imd` (the Don Maslin 1993 archive copy of my v1.22 CBIOS, IBM-3740 8" DSDD) reaches the `A0>` prompt with the period-correct banner.
- [x] Multiple boot disks from my archive mount and access cleanly through CCP.
- [x] Warm-boot CCP reload (`^C` from `A0>`) completes without spurious BIOS errors.
- [x] 10-second runtime stable, CTC 40 Hz BIOS-clock tick running, no `LOST_DATA` from the wd_fdc, no nested-NMI corruption.

## Status

`MACHINE_SUPPORTS_SAVE`. Cold-boot, warm-boot, file access, and sustained runtime all verified working with the period-correct CTC cascade and time-of-day clock enabled.

### DIFF
--- a/src/mame/colonial/sb80.cpp
+++ b/src/mame/colonial/sb80.cpp
@@ -114,7 +114,6 @@ private:
 	void unboot_w(uint8_t data);
 	void dsel_w(uint8_t data);
 	void fdc_drq_w(int state);
-	TIMER_CALLBACK_MEMBER(nmi_assert_tick);
 
 	required_device<z80_device>           m_maincpu;
 	required_region_ptr<uint8_t>          m_bootrom;
@@ -128,9 +127,6 @@ private:
 	required_device<com8116_device>       m_brg;
 
 	bool      m_boot_active = true;
-	emu_timer *m_nmi_assert_timer = nullptr;
-	attotime  m_drq_drop_time = attotime::zero;
-	bool      m_nmi_assert_pending = false;
 };
 
 
@@ -211,68 +207,10 @@ void sb80_state::dsel_w(uint8_t data)
 
 // DRQ -> Z80 /NMI: the boot ROM's NMI handler at $0066 does EXAF / EXX /
 // INI / EXX / EXAF / RETN for one byte per DRQ.  After the boot ROM banks
-// out, the CBIOS replaces the $0066 vector with its own equivalent (the
-// EXAF/EXX dance assumes atomic per-byte dispatch -- there's no protection
-// against re-entry).
-//
-// Workaround for the MAME-side nested-NMI bug (see the CTC-cascade TODO
-// above and mamedev/mame issue #15416): if a DRQ rising edge would arrive
-// less than `nmi_settle` after the previous falling edge, defer the NMI
-// line assertion via a timer.  That gap is what the FDC's next set_drq()
-// can take to fire after the host's data_r drops drq -- in MAME the next
-// rising edge can land within a microsecond or two of the falling edge,
-// which is shorter than the BIOS NMI handler's ~12 us total execution
-// time.  Real WD1793 pipeline silicon guarantees one byte-time of spacing
-// between rising edges; until the upstream wd_fdc fix lands, we close the
-// equivalent window here.
-//
-// 6 us is the empirically-tuned threshold: it's enough to cover the
-// post-data_r tail of the BIOS handler (EXX/EXAF/RETN ~= 5.5 us at
-// 4 MHz at the BIOS handler's instruction sequence) and short enough
-// that, when the FDC is firing bytes at 25 us per byte at MFM 500 kbps,
-// the deferred assertion still lands well before the next byte
-// boundary -- no spurious LOST_DATA.  Tested at 8 us and observed the
-// FDC raising LOST_DATA after the workaround pushed valid NMI service
-// just past byte boundaries on CTC-INT-preempted byte arrivals; 6 us
-// is the sweet spot where the nested-NMI window is closed without
-// over-deferring legitimate dispatches.
+// out, the CBIOS replaces the $0066 vector with its own equivalent.
 void sb80_state::fdc_drq_w(int state)
 {
-	if (state)
-	{
-		// Rising edge.  Check whether we're inside the settle window.
-		attotime now = machine().time();
-		if (!m_drq_drop_time.is_zero()
-		    && (now - m_drq_drop_time) < attotime::from_usec(6))
-		{
-			m_nmi_assert_pending = true;
-			m_nmi_assert_timer->adjust(attotime::from_usec(6) - (now - m_drq_drop_time));
-			return;
-		}
-		m_maincpu->set_input_line(INPUT_LINE_NMI, ASSERT_LINE);
-	}
-	else
-	{
-		// Falling edge.  Clear NMI and cancel any pending deferred assert
-		// (the chip won't have anything ready for the host until the next
-		// byte boundary, which is well past our settle window).
-		if (m_nmi_assert_pending)
-		{
-			m_nmi_assert_timer->reset();
-			m_nmi_assert_pending = false;
-		}
-		m_maincpu->set_input_line(INPUT_LINE_NMI, CLEAR_LINE);
-		m_drq_drop_time = machine().time();
-	}
-}
-
-TIMER_CALLBACK_MEMBER(sb80_state::nmi_assert_tick)
-{
-	if (m_nmi_assert_pending)
-	{
-		m_nmi_assert_pending = false;
-		m_maincpu->set_input_line(INPUT_LINE_NMI, ASSERT_LINE);
-	}
+	m_maincpu->set_input_line(INPUT_LINE_NMI, state ? ASSERT_LINE : CLEAR_LINE);
 }
 
 
@@ -303,10 +241,7 @@ void sb80_state::machine_start()
 {
 	m_bootbank->configure_entry(0, &m_bootrom[0]);    // /RESET state
 	m_bootbank->configure_entry(1, &m_ram[0]);        // post-UNBOOT
-	m_nmi_assert_timer = timer_alloc(FUNC(sb80_state::nmi_assert_tick), this);
 	save_item(NAME(m_boot_active));
-	save_item(NAME(m_drq_drop_time));
-	save_item(NAME(m_nmi_assert_pending));
 }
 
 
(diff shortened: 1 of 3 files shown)


rb6502:
We really need to have a general notice up about this, but let me give a bit of policy:

- It's fine to use AI assistance to work on and debug MAME code.
- BUT, you should write most of and understand all of the actual code you're submitting.  Also, please write the pull request descriptions.  If you look at the full whatsnew.txt, MAME PR descriptions tend towards the terse.  That certainly varies per person, but nobody writes these kind of novels.

Like, this can just be: "colonial/sb80.cpp: Added driver for Colonial Data Systems SB-80" plus the usual NEW MACHINE ADDED AS WORKING boilerplate.  The rest of the information belongs in a comment in the driver at best.  Don't tell me here that there's an in-driver DMA workaround, just comment it where it happens.

davidlrand:
> We really need to have a general notice up about this, but let me give a bit of policy:
> 
>     * It's fine to use AI assistance to work on and debug MAME code.
> 
>     * BUT, you should write most of and understand all of the actual code you're submitting.  Also, please write the pull request descriptions.  If you look at the full whatsnew.txt, MAME PR descriptions tend towards the terse.  That certainly varies per person, but nobody writes these kind of novels.
>

Thank you, I'll try to be more terse.

I did write the software; starting with the code back in 1983/1984.  I also modified the hardware in that time period.  Over the last couple of weeks, I've brought up multiple of the machines that I've worked on over the years (and I'm working on the others, now, including many of the National Semi boards I've been involved with over the years).
It's been, frankly, a joy to work on mame; the ease at bringing stuff to life again has been amazing to me.  

Again, I appreciate your comments.  Balancing "too much" and "too little" has been a challenge for me.  I'll work on it.

davidlrand:
Companion media for running this driver — boot ROM, a ready-to-run disk, and the original disk dumps — is now at https://github.com/davidlrand/mame-system-media (see `colonial-sb80/`).

davidlrand:
Pushed a follow-up removing the driver-local 6 µs deferred-NMI workaround in `fdc_drq_w()` (timer + state members gone; the FDC now drives /NMI directly).

The nested-NMI failure it covered is fixed in the Z80 core by #15434 (NMI prioritised over a maskable INT during the acknowledge race), so this PR now **depends on #15434** and should merge after it. Built against current master without #15434 the warm boot still fails — that is expected and is what #15434 corrects.

Re-tested with #15434 applied and the workaround removed: 3/3 cold boots to `A0>`.

happppp:
(I also commented on the PR, but since the discussion is here):
If an NMI pulse happens during the interrupt acknowledge cycle, Z80 will first process the 1st opcode of the normal IRQ routine. Only then will it jump to the NMI routine.

Correct me if I'm wrong that holub's change to Z80 allows an NMI before the 1st IRQ opcode.
If I did see it correctly, I'd propose a revert. Nothing personal, just factually incorrect.

davidlrand:
Withdrawing the #15434 dependency: that core Z80 change was reverted by #15436 (it contradicted hardware — a real Z80 runs the first ISR instruction before taking an NMI asserted during the maskable-interrupt acknowledge). I've reverted the workaround-removal, so this PR again carries the self-contained driver-local 6 µs deferred-NMI gate and boots against current master with no core dependency. The nested-NMI core issue stays tracked in #15416.

src/mame/colonial/sb80.cpp
```diff
@@ -0,0 +1,456 @@
...
+
+#include "bus/rs232/rs232.h"
+#include "cpu/z80/z80.h"
+#include "imagedev/floppy.h"
+#include "machine/com8116.h"
+#include "machine/z80daisy.h"
+#include "machine/z80ctc.h"
+#include "machine/z80pio.h"
+#include "machine/z80sio.h"
+#include "machine/wd_fdc.h"
```
pmackinlay: Sort this one.

src/mame/colonial/sb80.cpp
```diff
@@ -0,0 +1,456 @@
...
+	{ }
+
+	void sb80(machine_config &config);
+
+private:
+	void mem_map(address_map &map) ATTR_COLD;
+	void io_map(address_map &map) ATTR_COLD;
+
+	void machine_start() override ATTR_COLD;
+	void machine_reset() override ATTR_COLD;
```
pmackinlay: These should be `protected`.

src/mame/colonial/sb80.cpp
```diff
@@ -0,0 +1,456 @@
...
+ROM_START( sb80 )
+	ROM_REGION( 0x0100, "bootrom", 0 )
+	ROM_LOAD( "bootrom.bin", 0x0000, 0x0100, CRC(fa521576) SHA1(32153db6f37e7f5666e89aa0b8712d79daafeb4b) )
+ROM_END
+
+} // anonymous namespace
+
+
+//    YEAR  NAME  PARENT  COMPAT  MACHINE  INPUT  CLASS         INIT        COMPANY                          FULLNAME  FLAGS
+COMP( 1982, sb80, 0,      0,      sb80,    0,     sb80_state,   empty_init, "Colonial Data Services Corp.",  "SB-80",  MACHINE_SUPPORTS_SAVE )
```
pmackinlay: I think we have a preference to exclude business entity types or abbreviations (e.g., "Corp.") in the company field. I know it's not consistent across our existing code, so raising it here to solicit other opinions.

pmackinlay (changes requested):
Only very minor feedback here - I'm good to merge once the first two comments are addressed.

src/mame/colonial/sb80.cpp
davidlrand: Done

src/mame/colonial/sb80.cpp
davidlrand: done

src/mame/colonial/sb80.cpp
davidlrand: Done

PR #15418, opened 2026-05-31
imagedev/floppy: derive twosid_r() from media variant

davidlrand:
Reads media-sided-ness from the floppy_image variant tag instead of `get_actual_geometry()`.

Real drives sense single- vs double-sided physically (8" Shugart TS# on pin 30, derived from index-hole position); the signal comes from the media, not from whether the host has happened to write to head 1.

Before: a freshly-mounted unformatted SSSD image (`get_actual_geometry` returns 0 heads) reports TWO-SIDED to the host BIOS, and the answer can flip permanently once a host writes spurious data to head 1.

After: SS variants always report single-sided; DS variants always report double-sided; an unformatted SSSD diskette is correctly recognised as single-sided.

No call-site changes needed — all existing `twosid_r()` consumers and the uPD765 ST3 plumbing already do the right thing once the function returns correctly.

## Test plan

- [x] Builds clean against current `upstream/master`.
- [x] `floptool flopcreate mfi u8sssd xfer.mfi` → mount on a 2-head 8" system → BIOS now sees TWO-SIDED = false; CP/M `FORMAT` correctly produces a single-sided diskette.

### DIFF
--- a/src/devices/imagedev/floppy.cpp
+++ b/src/devices/imagedev/floppy.cpp
@@ -940,6 +940,15 @@ bool floppy_image_device::twosid_r()
 	case floppy_image::SSQD:
 	case floppy_image::SSQD16:
 		return true;
+	case 0:
+		{
+			// The loaded format did not tag a variant; fall back to the
+			// observed geometry so formats that never call set_variant()
+			// keep their previous behaviour (no regression).
+			int tracks = 0, heads = 0;
+			m_image->get_actual_geometry(tracks, heads);
+			return heads == 1;
+		}
 	default:
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


galibert:
Did you check if everybody does set the variant?

davidlrand:
I did not (at the time I wrote this, I didn't fully understand the floppy subsystem as I was focused on the 8"-only path).  About 47 of the 200+ format files call set_variant directly.  Your concern is appropriate.

I have updated the fix to fall back to the old geometry path, if set_variant is not done.  That will leave those systems unaffected.

galibert:
Mwahahahaha,  "Your concern is appropriate" is sooooooooo LLM-generated :-)

Ok to merge once the CI passes.

davidlrand:
Dude.  That was literally *me* saying that, not llm.

On the other hand, I got System V booting on the pc532 this morning, so ain't nothing gonna bring me down. :-)

cuavas:
Are you the guy LLMs were trained on?  Maybe @galibert is having flashbacks to The Far Side.

image

davidlrand:
They say that I am The One. :-)

PR #15419, opened 2026-05-31
itautec/i7000: drive the screen from the MC6845 CRTC

felipesanches:
Replace the placeholder tilemap renderer with an MC6845_UPDATE_ROW callback driven by the CRTC at ports 0x08/0x09, so the geometry and column count come from the CRTC registers the boot ROM programs.

And also made some adjustments to the color palette so that it better matches photos of the real device:

# photo
image

# old palette
Screenshot From 2026-05-31 07-58-31

# new palette
image

### DIFF
--- a/src/mame/skeleton/i7000.cpp
+++ b/src/mame/skeleton/i7000.cpp
@@ -54,7 +54,6 @@
 #include "screen.h"
 #include "softlist_dev.h"
 #include "speaker.h"
-#include "tilemap.h"
 
 
 namespace {
@@ -66,7 +65,8 @@ public:
 		: driver_device(mconfig, type, tag)
 		, m_maincpu(*this, "maincpu")
 		, m_card(*this, "cardslot")
-		, m_gfxdecode(*this, "gfxdecode")
+		, m_crtc(*this, "crtc")
+		, m_palette(*this, "palette")
 		, m_videoram(*this, "videoram")
 		, m_kbd(*this, "X%u", 0U)
 	{ }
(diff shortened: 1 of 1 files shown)


felipesanches:
With this, the video correctly changes from 40-column to 80-column.
It can be tested by pressing option number 1 "VID" on the selt-test menu.
The screen is filled with chars in 40-col mode. Then after pressing enter, it switches to 80-col mode.

felipesanches:
# 40-col
Screenshot From 2026-05-31 08-21-16

# 80-col
Screenshot From 2026-05-31 08-21-23

startaq:
Isn't the palette just the usual green on black? See for example this advertisement: 


The brightness in your photo just seems to be turned very high.

felipesanches:
yes, maybe. That kind of thing is very subjective, though. The new palette is way better than the previous one and it looks cooler than full green over full black. And the frame capture of the youtube video is at least one sample of the real thing, which I trust a bit more than an ad photo.

startaq:
Well, something looking "cool" isn't a factor for MAME, it's about accuracy. Here are some more photos showing the screen in detail: https://tabajara-labs.blogspot.com/2013/08/alguem-conhece-o-i7000-da-itautec.html

felipesanches:
> Well, something looking "cool" isn't a factor for MAME, it's about accuracy. Here are some more photos showing the screen in detail: https://tabajara-labs.blogspot.com/2013/08/alguem-conhece-o-i7000-da-itautec.html

yes, you're right. My aesthetics preferences got in the way of accuracy! :-)
(even though I had myself left the driver with a much worse and absolutely inaccurate palette for 11 years...)

Let's fix it!

angelosa:
Is it just me or this driver also needs a reordering of composing blocks? 
```
- keyboard handlers
- INPUT_PORTS_START for keyboard
- machine_start
- palette_init
- address_map(s)
- DEVICE_IMAGE_LOAD_MEMBER for m_card
- screen_update
- machine_config
- ROM_LOAD
```
I thought we had guidelines at some point to not make something like this to happen?

PR #15422, opened 2026-05-31
frontend: clear transient UI BIOS option without preserving command-line priority

### LINKED ISSUE #11313: UI ignores "-bios" string stored in .ini files after running a game once during the same session when "skip_biosmenu" is turned on, reverts to default bios value in driver
### MAME version

0.255

### System information

Windows 10 Home, 64-bit, version 22H2, USA, US English
Installed RAM 8.00 GB
AMD A6-5200 APU with Radeon(TM) HD Graphics 2.00 GHz

### INI configuration details

```INI
mame -cc

ui.ini only
skip_biosmenu             1
```

### Emulated system/software

UI only, neogeo/neogeo.cpp, igs/pgm.cpp, and sega/stv.cpp confirmed, possibly affects all systems with bios?

### Incorrect behaviour

Mame ignores bios settings in the .ini file during a UI session after first playing any game, then ending that emulation and starting a new emulation, but when the "Skip Bios selection menu" option is turned ON.  It can be the same game too.  Example:

1. I run the UI and select Street Fighter II to play.  After I'm done I hit escape and go back to the UI.  Now this game doesn't use a bios at all.
2. I select Knights of Valour, a pgm based game.  I have a kov.ini file with `bios v1` which equates `bios 1` as the default is v2 ("0").
3. I TAB to get to the BIOS selection menu in the UI while the game is up, and it lists "v2" and not what I have in the kov.ini file.

Another example:
1. I run the UI and select Knights of Valour (bios v1) first this time, check in the UI that it says "v1", then exit the game back to the UI.
2. I decide to play Knights of Valour again, checking the bios setting in the UI shows that it is now "v2".

### Expected behaviour

Turning on the Skip BIOS menu selection option should not make the UI ignore any bios settings in .ini files.

### Steps to reproduce

1. Double-click mame.exe.
2. Go to "General settings > Miscellaneous options > Skip BIOS selection menu" and turn that option ON.
3. Save the setting.
4. Run any game while still in the UI.
5. Exit that game.
6. Now run a game that has a specific non-default (not 0) bios setting in an .ini file.  Let's use kov with kov.ini.
7. Press TAB while the game is running and check the BIOS setting and you'll see it doesn't match the file.

### Additional details

I checked games as stand alone machines and not as a multi-slot setup, so I don't know how those are affected and that seems complicated to work out.  As noted above, I confirmed neogeo, pgm, and stv based games are affected, all arcade based.  I didn't check software lists, but maybe someone else can have fun with that.

I can't say for sure when this all started, but it has been going on for a while.  One day I'm looking at my nvram folder and saw some games had extra nvrams that I know should not, like Euro region for games I have set for only US or JP.

### LINKED ISSUE #11314: UI ignores "-bios" string stored in .ini files after running a game once during the same session when using "ui simple" setting, reverts to default bios value in driver
### MAME version

0.255

### System information

Windows 10 Home, 64-bit, version 22H2, USA, US English
Installed RAM 8.00 GB
AMD A6-5200 APU with Radeon(TM) HD Graphics 2.00 GHz

### INI configuration details

```INI
mame -cc
```

### Emulated system/software

UI only, neogeo/neogeo.cpp, igs/pgm.cpp, and sega/stv.cpp confirmed, possibly affects all systems with bios?

### Incorrect behaviour

Mame ignores bios settings in the .ini file during a UI session after first playing any game, then ending that emulation and starting a new emulation, but when "ui" is set to "simple" under CORE MISC OPTIONS in the mame.ini file.  This issue is basically the same thing as https://github.com/mamedev/mame/issues/11313 but I made a new issue since this affects the "ui" option inside the mame.ini and not "skip_biosmenu" in the ui.ini.

### Expected behaviour

Simply changing the UI type used should not cause bios settings in .ini files to be ignored.

### Steps to reproduce

1. Commandline `mame -cc`
2. Open mame.ini with a text editor, change `ui cabinet` to `ui simple` and save the file.
3. Create a custom machine .ini file for a game that uses a bios, such as magdrop3.ini.  Open it with a text editor and add `bios us-e` and save it.
4. Now double-click mame.exe to get going in the UI.
5. Run magdrop3 (Magical Drop III).
6. Press TAB and go to BIOS Selection.  It should say "US MVS (Ver. 1)" https://github.com/mamedev/mame/blob/master/src/mame/neogeo/neogeo.cpp#L2270
7. Exit the game and then, while still in the UI, select it again.
8. Press TAB and go to BIOS Selection.  Now it says "Europe MVS (Ver. 2)" which is the driver default https://github.com/mamedev/mame/blob/master/src/mame/neogeo/neogeo.cpp#L2259
9. Check your nvram folder.  There should be "magdrop3_5" (US bios 5) and "magdrop3" (EUR bios 0).

### Additional details

I thought of this after I discovered that `ui simple` was affecting `snapname` in the other issue I opened, and that this was also a case of the UI ignoring something that the commandline does not.  Very basic pattern here.  The other common thing here is that the problem happens in the UI only when there is no bios selection screen prior to launching the game.  The simple UI does not present one, and the cabinet UI also does not if you turn off "skip_biosmenu".

It's these sorts of things cropping up a lot lately that had me strictly using the commandline out of fear that the UI is actually interfering with things.

Immersion95:
## Summary

This fixes a BIOS option priority leak when returning from an emulated machine to the internal UI.

When a machine exits, the frontend currently clears `OPTION_BIOS` by setting it to an empty value at `OPTION_PRIORITY_CMDLINE`. This was added on 2022-12-20 in aee2e13bc78e84241cf602617ee1069400886584 to fix [MAMETesters 8292](https://mametesters.org/view.php?id=8292), where a BIOS selected through the UI could remain active after exiting a Neo Geo machine and cause false ROM/CHD warnings when launching other systems.

That intent is still correct: transient BIOS selections made by the UI should be cleared when returning to the empty pseudo-driver.

However, setting `OPTION_BIOS` to an empty value at command-line priority leaves a high-priority empty BIOS option behind. On the next launch from the UI, per-machine INI files are parsed again, but their `bios` setting cannot override the existing command-line-priority empty value.

This is visible when no BIOS selection menu is shown before launching the next machine, for example with `skip_biosmenu 1` in #11313 or `ui simple` in #11314. In those cases, settings such as `kov.ini` with `bios v1` or `magdrop3.ini` with `bios us-e` can be ignored after a previous machine has been run in the same UI session.

## Fix

Revert the command-line-priority BIOS option instead of replacing it with an empty command-line-priority value.

This keeps the original intent of aee2e13 / [MAMETesters 8292](https://mametesters.org/view.php?id=8292): transient BIOS selections made by the UI are still cleared when returning to the empty pseudo-driver. The difference is that the BIOS option is no longer left as an empty value at command-line priority, so per-machine INI files can set `bios` again on the next launch.

## Tested

Manually tested the affected UI relaunch scenarios without observing a regression:

- `skip_biosmenu 1` with a machine-specific BIOS setting
- `ui simple` with a machine-specific BIOS setting
- relaunching machines from the internal UI after exiting a previous machine

## References

Fixes mamedev/mame#11313
Fixes mamedev/mame#11314

Related:

- https://mametesters.org/view.php?id=8292
- https://github.com/mamedev/mame/commit/aee2e13bc78e84241cf602617ee1069400886584

## Disclosure

This issue was diagnosed with help from ChatGPT. The actual fix is just a single frontend line, so it should be small and straightforward to review 🙂.

### DIFF
--- a/src/frontend/mame/mame.cpp
+++ b/src/frontend/mame/mame.cpp
@@ -300,7 +300,7 @@ int mame_machine_manager::execute()
 			if (machine.exit_pending())
 			{
 				m_options.set_system_name("");
-				m_options.set_value(OPTION_BIOS, "", OPTION_PRIORITY_CMDLINE);
+				m_options.get_entry(OPTION_BIOS)->revert(OPTION_PRIORITY_CMDLINE, OPTION_PRIORITY_CMDLINE);
 			}
 		}
 


Immersion95:
Would it be possible to have a feedback regarding this ? I'll gladly make some change :).

I think it's a important matter :)

@cuavas : Would you be able to help ? Thanks 😊.

PR #15424, opened 2026-05-31
rallyx: Replace BANG sample with netlist.

ricbit:
Remove the sample from rallyx and replace it with a netlist. 

The circuit was traced from the service manual directly, and only one component had to be added: MB3730_SIL. For this opamp, I tried to match as best as possible the datasheet. Output audio is very similar to the sample, but not equal, the sample may have some filtering due to the acoustics of the cabinet.


ICEknigh7:
Is it normal that there's a constant noise in the background?

ricbit:
Isn't the noise the car motor? I can hear the motor noise even before this patch.

ICEknigh7:
I mean, the noise that starts right after the ROM check, before a game even begins.

cuavas:
There’s nothing gross in the code here.  Can someone with some free time double-check the netlist against the schematic and try it out as a sanity check?

If there’s constant noise, can someone check if it’s working as it should, or if it’s a solver artefact, Nyquist aliasing or something?

ricbit:
There is indeed some noise coming from the bang circuit, I'm investigating the cause. Good hearing, I didn't notice i before.

ricbit:
So here's what I gathered:
- The noise is real, you can see it by dumping the wave while the machine is idle.
- The noise is caused by leakage from the LFSR through the CD4066, the model for the CD4066 has finite impedance so some leakage is expected.
- The poweramp amplifies the tiny noise to something that can be heard if you have a godd noise.

The gain of the MB3730 poweramp is supposed to be big, you can see on the current recorded sound wav how clipped to signal is. Reducing its gain would reduce the noise, but also would modify the high frequencies which are present in the original.

I would say the noise is working as expected. I believe this would be heard in the real machine as well. 

If we want to eliminate the noise without messing with the high frequencies, a simple path is to replace the CD4066 with an ideal component.

ICEknigh7:
If the noise does happen on real hardware it should stay (and in such case, if a user doesn't like it, there's many sound filters available to mitigate it) but are you really sure that there's no additional filters in the schematics which may be missing from the netlist implementation?

ajrhacker:
The primary filter on the BANG sound is a MFB built around the µA741 and associated discrete components (C5, C6, R41, R31, R39, D3, C29). Capacitor C28, attached to the junction of BANG and AUDIO, appears to be missing, though I suspect adding it would not solve the noise problem.

ICEknigh7:
The noise doesn't seem to happen on real hardware, at least according to the YouTube videos I'm seeing: https://www.youtube.com/watch?v=uoKZEl7tQnE

m1macrophage:
Just some thoughts. Not a member of the MAME team, feel free to ignore :)

If the noise is due to feedthrough, rather than simulation artifacts, the gain of the netlist output might be high. I see the "0.1" multiplier (in `set_multi_offset`). If this is arbitrary, it might be worth calculating it.

But since you've done the hard work of modeling the circuit, and if performance is acceptable, a simpler approach is to route "namco sound" to the netlist, via R28. *Just like in the real hardware*. "namco sound"'s output will need to be converted to voltage levels in this case. This will make the relative signal levels accurate. Then, the netlist output's multiplier can be manually adjusted for maximum volume and no clipping.

Regarding C28 that AJR mentioned. C28 , VR1, R30 and R28 form a [low-pass filter](https://falstad.com/afilter/circuitjs.html?cct=$+1+Infinity+0.75+50+5+50%0A%25+2+24389.095776063856%0Ac+320+160+320+208+0+4.7e-8+0%0AO+464+160+528+160+0%0A170+128+160+96+160+3+20+1000+5+0.1%0Ar+128+160+192+160+0+4700%0Ag+320+208+320+224+0%0Ar+416+160+416+208+0+1000%0Aw+320+160+416+160+0%0Ag+416+208+416+224+0%0Aw+416+160+464+160+0%0Ar+240+160+240+208+0+10000%0Ag+240+208+240+224+0%0Aw+240+160+320+160+0%0Ax+153+129+175+132+0+12+R30%0Ax+202+208+224+211+0+12+R28%0Aw+192+160+240+160+0%0Ax+290+207+312+210+0+12+C28%0Ax+379+210+402+213+0+12+VR1%0A
) with a cutoff frequency of ~4.4Khz. The uA741-based filter [attenuates high frequencies](https://falstad.com/afilter/circuitjs.html?cct=$+1+Infinity+0.75+50+5+50%0A%25+2+28853.998118144256%0AO+400+160+448+160+0%0A170+144+144+112+144+3+20+1000+5+0.1%0Ar+144+144+224+144+0+150000%0Ac+224+144+288+144+0+1e-8+0%0Aw+224+144+224+80+0%0Ac+224+80+288+80+0+1e-8+0%0Ar+288+80+288+144+0+470000%0Ar+224+144+224+192+0+22000%0Ag+224+192+224+208+0%0Aw+288+144+320+144+0%0Aa+320+160+400+160+0+15+-15+1000000%0Aw+320+176+304+176+0%0Ag+304+176+304+192+0%0Aw+288+80+400+80+0%0Aw+400+80+400+160+0%0A
) already, so this might make more of a difference to the "namco sound" signal (if the approach above is followed). Whether that's audible depends on the listener's age and the frequency content of the original signal.

If it is not possible to route "namco sound" through the netlist, of if you choose not to, you might want to include R28 and tie it to ground, to accurately simulate the load to the "bang" signal and the cutoff frequency of the C28-based LPF.

galibert:
> Just some thoughts. Not a member of the MAME team, feel free to ignore :)

That is very much something that should change, if you're interested.  My email is galibert@pobox.com.  Feel free to contact me.

ricbit:
I implemented @m1macrophage suggestions: reduced the gain of the poweramp, 
added the missing mixer circuit and injected the WSG audio through it. The volume
of the BANG itself is a bit lower, but I think there is no optimal balance between 
BANG and WSG, since in the real machine they are connected through a pot
that can be changed.

Please test it and check if the noise level is now acceptable.

ICEknigh7:
Can't test due to this: 
image

EDIT: Artifact build in progress now.

src/mame/namco/rallyx.cpp
```diff
@@ -855,15 +855,23 @@ void rallyx_state::rallyx(machine_config &config)
...
+	// WSG audio injected on I_AUDIO, all driving the bridged MB3730 power amp.
 	NETLIST_SOUND(config, "snd_nl", 48000)
 		.set_source(NETLIST_NAME(rallyx))
 		.add_route(ALL_OUTPUTS, "mono", 1.0);
 
 	NETLIST_LOGIC_INPUT(config, m_bang, "I_BANG.IN", 0);
 
+	// WSG audio injection. mult is a calibration constant chosen to drive the
+	// (now low-gain) MB3730 as hot as possible without clipping the output.
+	NETLIST_STREAM_INPUT(config, "snd_nl:cin0", 0, "I_AUDIO.IN").set_mult_offset(3.2, 0.0);
```
m1macrophage, line 873: One suggestion to match the hardware even more closely :)

If I am reading the schematic correctly, AUDIO will be in the range ~0-5V. So it is possible to use the hardware-accurate values here. Those will depend on the range of the WSG output (is it [0, 1] or [-1, 1]?).

If changing the values result in clipping, that should be fixable by reducing the multiplier for OUTPUT below, or by adjusting VR1.

> [...] but I think there is no optimal balance between BANG and WSG, since in the real machine they are connected through a pot that can be changed.

The pot (VR1) is applied after the signals are mixed. So the suggestions above should result in the correct balance. But maybe I misunderstood what you meant, or there is a different pot that I missed?

m1macrophage: > If I am reading the schematic correctly, AUDIO will be in the range ~0-5V.

I was reading it wrong. Max voltage seems to be closer to what you have here.  And the balance between BANG and WSG signals should be pretty accurate too.  

Exact voltage range according to [simluation](https://falstad.com/circuit/circuitjs.html?ctz=DwYwlgTgBAZgvAIgIwKgFwM6IAwDpsEECsqYIiSeATAVQOx0DM2AHFQGwCcndqIARoiLZUAB0EJhqAG4QhqALaYhAUwC0SFAD4AUFCjA0UAB6IALIypQknK2fZnrt1PAQioogHYVU-FRQJFAHtEABMVGABDAFcAGzQZP0QNXEYzCxZsInTmBksWRPI3XCIWC2wkFiJ8vM4SAHpdfWAAJRNzB2s6dih7RyR0lwp2VAB3V3cFSONpeUa9A2hTBD6oGhZezvoqIbdUORW6QPnmgHN2la26Oy2WAtgcBBODUYuLKxsrRnYNz92RZ7AJaIb6-WxQUHWJA7B57KAHSiEJ5NAxtZaQzRfH5QTgjWFIPHjR5QKYzOYooEXDHg5j9Zyw9wIwgAimvdHYz4Q7FUKhmf7IhaU9kbHmOWlrXn8+GIHlIwFskHc3lckVEPETAXNYEISFUNUQ7COPXq4kHMxHY6sqnc-W6u78wHau0bcVUe0M-YBZma1FUw1OKzbAO7AljDUk6azSQ+1p+o3KoOikOE8OkqMNClokH+41rO5rNXJsPEtPkwVZnU5-OJ92uUNQIlw0vR+VvLasKBEaFrVgOilO7n5yGMb5SplyinnYUQ74qiFEGEawFBKAqbzIe4YUSIBy7YyMMSPFHNURQKPuDBFDMLE9nkGoS84fCEHjsNV6mP1IK6YD1cAQXQgA): [0.78.  3.07]

m1macrophage: Hmm, I should have looked at the DAC circuit more closely. That simulation is not the full story.

Also, just to clarify, I am not suggesting any changes. The current implementation is probably fine.

cuavas:
> Can't test due to this: image
> 
> EDIT: Artifact build in progress now.

You know you can compile the source yourself, you don’t have to rely on CI builds.

src/mame/namco/rallyx.cpp
ricbit: You're right about the pot, it is only used after the mixer to control volume, I was misremembering. Since your values match more or less the current implementation, maybe we keep the code as is?

m1macrophage: Yeah, the current implementation is probably fine, IMO.

Getting it "perfect" will require digging more into the DAC circuit, but that might be out of scope for this PR, which focused on the noise circuit. My guess is it won't make a big difference anyway.

PR #15426, opened 2026-06-01
esqpump.h: work around `format_output<bool>.apply()` infinite recursion.

cbrunschen:
Since
https://github.com/mamedev/mame/commit/b946df69fcccd06bd516ddd302ab16ee20896e2c, printing a `bool` causes infinite recursion, as also described in https://github.com/mamedev/mame/commit/b946df69fcccd06bd516ddd302ab16ee20896e2c#commitcomment-187037152 .

This PR explicitly casts a `bool` to `int` before printing, avoiding this issue. It also makes this logging conditional on a compile-time constant, and disables the logging at this time.

### DIFF
--- a/src/devices/sound/esqpump.cpp
+++ b/src/devices/sound/esqpump.cpp
@@ -14,6 +14,52 @@
 #include "emu.h"
 #include "esqpump.h"
 
+#include <iostream>
+
+
+#define PUMP_TRACK_ESP_HALTED 0
+#define PUMP_DETECT_SILENCE 0
+#define PUMP_TRACK_SAMPLES 0
+#define PUMP_TRACK_PROCESSING 0
+#define PUMP_FAKE_ESP_PROCESSING 0
+#define PUMP_REPLACE_ESP_PROGRAM 0
+
+
+#define LOG_ESP_HALTED (1U << 1)
+
+#if PUMP_TRACK_ESP_HALTED
+#define VERBOSE LOG_ESP_HALTED
+#endif
+
+#include "logmacro.h"
+
+namespace sound::esqpump {
+
+struct stats {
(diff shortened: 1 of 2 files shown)


cuavas:
I'll sort out strformat.h, I've got other things that affect emu.h, and I was hoping to batch up a bunch of them, as people don't like having to recompile everything.

cbrunschen:
> I'll sort out strformat.h, I've got other things that affect emu.h, and I was hoping to batch up a bunch of them, as people don't like having to recompile everything.

Perfectly fair. In the meantime this works around the issue; making this particular piece of logging conditional is a good idea anyway, and the cast makes it explicit what will be logged.

cuavas:
It’s not a good idea to pollute the global namespace with *another* macro, and it isn’t good to name something that just controls logging with the same pattern as the other macros that actually enable code for tracking statistics

cbrunschen:
> It’s not a good idea to pollute the global namespace with _another_ macro, and it isn’t good to name something that just controls logging with the same pattern as the other macros that actually enable code for tracking statistics

Those macros and the code they guard remain from when I originally wrote the pump device and were used mainly for debugging - that log message slipped through the cracks, but the macro I added followed the same pattern and was used really for the same purpose as the others, following the established pattern. 

But to address that, I've moved the `#ifdef`:d parts of the device into a separate struct that is defined entirely in `esqpump.cpp`,  moved the body of `esq_5505_5510_pump_device::set_esp_halted()` from `esqpump.h` to `esqpump.cpp` and am now using `logmacro.h` to achieve the conditional logging. 

So no more macros polluting the global namespace. Even the added struct lives in a separate file-specific namespace.

Plus, I've made sure that various combinations of values for the macros all still lead to code that compiles and works.

PR #15428, opened 2026-06-02
bgfx: render vector games through a post-processing chain

okaz-code:
## Summary

This adds three new BGFX post-processing chains that bring D3D-HLSL-style CRT
rendering to **vector games**, which previously had no equivalent of the raster
HLSL/CRT pipeline:

- **Vector (Star Wars)** — beam **overload** (focus bloom-out on bright beams) and
  **monitor glow** (whole-tube glow when the beam is driven off-screen), on top of
  the full CRT post chain.
- **Vector (Color)** — color vector CRT (Tempest, etc.) with shadow mask.
- **Vector (Monochrome)** — monochrome phosphor CRT (Asteroids, Battlezone,
  Lunar Lander, Vectrex, etc.) with a tunable phosphor color.

Vector lines are rasterized into a supersampled FBO by a new analytic-AA line
shader, then run through phosphor persistence, multi-level bloom (with HLSL-style
overdrive cross-talk), optional shadow mask, and an asymmetric pincushion
distortion. Where possible the existing `hlsl/*` effects are reused; the
vector-specific stages are new `sw_*` effects.

Selecting a chain for a vector game now lists only vector-compatible chains
(filtered by a `"screen_type": "vector"` tag), so the raster `default` chain is no
longer offered for vector hardware.

## Screenshots

**Star Wars — beam overload (bright beams bloom/defocus)**
starwars_overload

**Star Wars — monitor glow (tube glows as beams are driven off-screen)**
starwars_monitor_glow

## What's included

New BGFX assets:
- 3 chains: `bgfx/chains/vector-{starwars,color,monochrome}.json`
- 9 effects: `bgfx/effects/sw_*.json`
- 9 shaders: `src/osd/modules/render/bgfx/shaders/fs_sw_*.sc`, with precompiled
  binaries for dx9 / dx11 / glsl / essl / metal / spirv under `bgfx/shaders/*/`.

Core engine changes:
- `vector.{h,cpp}` — per-frame, per-beam **overload pipeline** shared by all vector
  games (sigmoid → threshold split into display intensity + overload amount → beam
  width expansion), **monitor-glow accumulation** of off-screen overloaded beam
  energy, intensity-driven **point-vector dot sizing**, and an optional generation-
  counter-based **vector CRT flicker**. All effect parameters come from chain
  sliders, with ini-derived fallbacks when a chain exposes no matching slider.
- `avgdvg.{h,cpp}` — Star Wars carries a raw beam score (VCTR × STAT intensity) so
  the overload split can be computed from the original beam current; the default
  clip window is made symmetric so overscan can reveal off-screen beams.
- `drawbgfx.{h,cpp}` — renders `PRIMFLAG_VECTOR` lines into a supersampled FBO via
  the new line effect, feeds the result into the chain, pushes the per-frame vector
  options, and injects the monitor-glow amount as a uniform.
- `chainmanager.{h,cpp}` — detect vector games and filter the chain selection list
  by `screen_type`; build the 8-level bloom mip targets.
- `render.{h,cpp}` — `render_container::add_line` gains an `overload` parameter
  threaded through to the primitive.
- Small supporting changes in `targetreader.cpp` (per-target format, RGBA16F for the
  float bloom accumulator), `view.h`, `chainentry.*`, `targetmanager.cpp`.

## Behaviour on non-BGFX renderers is unchanged

A single `s_bgfx_active` flag (set only when the BGFX renderer initializes) gates the
new path. On every other renderer, `vector_device::screen_update` and the Star Wars
beam path fall back to the **stock** intensity/`beam_width` behaviour, so non-BGFX
output is byte-for-byte identical to stock.

## Testing

- Full `make clean` rebuild plus shader rebuild; verified in-game.
- Star Wars (overload + monitor glow), Tempest (color + shadow mask), Asteroids /
  Battlezone / Lunar Lander (monochrome) on the BGFX renderer.
- Non-BGFX renderers confirmed identical to stock vector output.

## Known issue

On Windows, OS/driver post-processing that targets DXGI flip swapchains — notably
**NVIDIA RTX Dynamic Vibrance** and Auto HDR — can blow out the SDR image (the UI
blows out too, so it is not a shader issue). OpenGL is unaffected. This is external
to MAME; disabling that per-app driver filter restores correct output.

## Notes

- All comments are in English; new files carry BSD-3-Clause headers.
- Comment/asset additions only for the most recent commit — no functional change
  there. The pipeline itself is opt-in via chain selection and has no effect unless
  a `sw_*` / vector chain is chosen.


cuavas:
Making the device need to know the active renderer is definitely an unacceptable break of encapsulation.  For a chance to get somewhere with this, please break it up so the core renderer changes to vector primitive handling can be discussed/reviewed independently of the rest of the changes first.

okaz-code:
Scoped down from the original submission per the review: this now contains only the core renderer changes for vector primitive handling. Vector lines are rasterised as analytic-AA primitives into a supersampled FBO and run through a BGFX chain (with a minimal default-vector chain auto-selected for vector games). The vector device and core render primitive interface are unchanged from upstream. The CRT chains and the Star Wars overload / monitor-glow work will follow as separate PRs.

MooglyGuy:
So this very apparently only works on the first vector screen found. How much of this code was written by you, and how much of it was written by AI?

okaz-code:
Thanks for taking a look.

On the multi-screen point: you're right, it currently assumes a single vector screen (window 0 / screen 0). That was a deliberate scope decision rather than an oversight — the systems I was targeting each have a single vector screen. If multi-vector-screen support is needed for this to be considered, I'm happy to extend it.

On your other question, I'll be straight with you: the design, the rendering approach, and the build and runtime testing/verification are mine, but the actual C++ code and the comments were AI-generated under my direction. I should also be upfront that I'm not a native English speaker, so this message and the PR text were drafted with AI assistance and then checked by me.

I understand the project has concerns about AI-generated contributions, and if that's a reason to decline this PR, I fully understand and accept that.

seleuco:
Hi @okaz-code,

First of all, amazing work on this PR! I've been following your changes and reading your blog, and the research you've done on the Atari hardware is fantastic.

I am Seleuco, the developer behind MAME4droid. For the last couple of months, I've also been working on a visual overhaul for vector rendering in my Android port (using a custom GLES3 pipeline). I saw some brilliant ideas in your implementation that I hadn't thought of at all—especially the whole-monitor bloom/overdrive effect for Star Wars. I was wondering if you wouldn't mind if I explore and try to adapt that concept for my port?

I have massive respect for mainline MAME and its strict architectural guidelines. Because my project is a downstream port, I took a different route to avoid touching the core: I left vector.cpp completely unmodified and moved all the "fake" physical simulations and overbright math directly into “my”  renderer itself.

It has been quite a headache, to be honest. I ran into a lot of issues trying to mix traditional MAME SDR UI/artworks with true FP16 HDR vector accumulation, but I had to make some compromises and push through. I also tried to respect MAME's cocktail/multi-monitor setups, though I had to cut some corners. For example, I completely disable phosphor persistence when multiple monitors are detected because managing the multi-FBO ping-pong logic was driving me crazy!

If you ever want to take a look at my approach, suggest any improvements, or even collaborate, you are more than welcome to drop by my repo (https://github.com/seleuco/MAME4droid-Current).

Thanks again for the inspiration, and best of luck with the PR!

ICEknigh7:
> On the multi-screen point: you're right, it currently assumes a single vector screen (window 0 / screen 0). That was a deliberate scope decision rather than an oversight — the systems I was targeting each have a single vector screen. If multi-vector-screen support is needed for this to be considered, I'm happy to extend it.

In my very personal opinion I think that, as long as it's within the scope of what's being implemented, the more future-proof that the code can be, the better. Examples for this case would be games with more than one vector screen, systems which use a vector screen but also standard monitors for debugging, etc.

okaz-code:
Thanks, that's a fair point and I agree it's worth keeping in mind within scope.

As far as I can tell, there's currently no machine in MAME that instantiates more than one vector screen, or a vector machine that also uses a raster screen — so I have no way to actually test those setups today. Rather than add a full multi-vector-screen path I can't verify, I'll make sure the renderer doesn't hard-assume the vector screen is the only screen, so multi-screen setups degrade gracefully instead of breaking. The one multi-output case that is testable now is mirroring a single vector screen across windows (-numscreens 2), and I'll keep that path correct.

If a driver with multiple vector screens ever lands, the chain is already per-screen, so extending it should be incremental. Happy to revisit then with something I can verify.

rb6502:
> I have massive respect for mainline MAME and its strict architectural guidelines. Because my project is a downstream port, I took a different route to avoid touching the core: I left vector.cpp completely unmodified and moved all the "fake" physical simulations and overbright math directly into “my” renderer itself.

I think at least some changes would be OK in vector.cpp itself so that the renderers aren't each re-inventing the wheel.
 
> It has been quite a headache, to be honest. I ran into a lot of issues trying to mix traditional MAME SDR UI/artworks with true FP16 HDR vector accumulation, but I had to make some compromises and push through. I also tried to respect MAME's cocktail/multi-monitor setups, though I had to cut some corners. For example, I completely disable phosphor persistence when multiple monitors are detected because managing the multi-FBO ping-pong logic was driving me crazy!

I'm very interested in improved vector rendering, especially with HDR.  If you two would be interested in working together "off-tree" and come up with a combined approach that works well I'd love to see it, subject to not annoying @MooglyGuy since the BGFX renderer is his code.  The caveat is that it needs to fall back gracefully to the current vector rendering for non-HDR displays and cooperate with multiple monitors.  Nothing we emulate has multiple vector monitors that I'm aware of so only one monitor having vector effects is probably OK, but showing MAME on a monitor other than the system primary is a common use case.

okaz-code:
Thanks, this is really encouraging.

On vector.cpp: I agree. The overbright / beam-energy information is something every renderer would otherwise have to re-derive, so exposing it from the core in a renderer-agnostic way — a raw beam value carried on the primitive, with the displayed intensity left unchanged — makes more sense than each renderer reinventing it. That's the direction I'd like to take for the overdrive and monitor-glow effects in a follow-up: keep the core as a pure data source and leave the actual effect math in the renderer.

On multi-monitor: that matches how the current PR already behaves. The chain runs on the vector screen's window, and any additional window falls back to the standard vector rendering with the UI unaffected, so showing MAME on a non-primary monitor works. I'll make sure that stays true. And agreed — since there's no multi-vector-monitor system to target, a single monitor carrying the effects is fine, and I'll keep a graceful fallback for the plain / non-HDR path.

I'd be very happy to work with David off-tree on a combined approach, and of course to stay within the BGFX renderer's architecture — it's @MooglyGuy's code and I don't want to step on it. If we land on something that fits, I'll bring it back here.

seleuco:
First off, I want to step back a bit. Designing a unified, architecturally pure solution that integrates perfectly upstream is a massive undertaking, and a bit out of my current scope.

My MAME4droid implementation is very tailored to mobile hardware constraints. The bloom/overbright solution looks quite striking on a phone screen, but under the hood, it tightly couples the physics simulations with the HDR math by accumulating directly into GLES3 FP16 textures.

As I was building it, I quickly realized I was mixing the classic MAME rendering paradigm with modern HDR concepts. Because of this, it is definitely not the "top tier" or most elegant architectural approach. For example, the way I handle the tone mapping to bring that accumulated FP16 HDR energy back down to SDR for standard displays relies on some opinionated compromises that work well enough for my specific port, but probably wouldn't pass the strict standards and flexibility required for mainline MAME.

Instead of doing a proper post-processing pass for the bloom on the GPU (like BGFX does), I generate the halo geometry dynamically on the CPU side during ⁠process line primitive⁠s. I literally take the ideal line width, multiply it by my physical expansion math (overbright), and push extra textured quads to simulate the scattered light.

It works for mobile because it saves FBO fillrate , but architecturally it's a mess. The core (⁠vector.cpp⁠) should be the one handling the baseline physics, and the renderer should just apply true optical diffusion. Because I build the bloom per-primitive, overlapping lines don't blend their bloom organically; they just pile up additive geometry.

This is where I really had to force things. MAME expects a traditional 8-bit SDR pipeline. To get true HDR, I hijack the alpha channel (which MAME uses for raw voltage/intensity), convert it to linear light using the CRT gamma curve, and accumulate it into a ⁠GL_RGBA16F⁠ FP16 FBO.

But then I have the UI, bezels, and background artworks. Since they are SDR, if I just render them next to a glowing 400-nit vector, they look incredibly dim. My hack was to inject a "Fake Raster HDR" pass with a hardcoded ⁠paper white uniform just to artificially boost the menus and artwork so they survive next to the vectors. It’s a massive compromise.

Because I do all the overbright math locally per-primitive, a lot of the accumulated optical energy gets lost or mathematically crushed. When the game runs on a standard non-HDR screen (or when I have to fallback to the SDR path), I have to squash all that beautiful FP16 energy back down to 8-bit. I use some aggressive inverse tone mapping and separate blend functions to gently carve out the background and prevent the bright vectors from hard-clipping and looking "fat".

Visually, the bloom looks punchy on a phone, but physically, the light energy isn't truly conserved. I'm artificially capping it instead of letting a proper HDR pipeline naturally bleed the peaks.

So yeah, while it looks great on an OLED Pixel screen, making the renderer do the job of the core, the tone mapper, and the compositor all at once is structurally wrong. That’s why I completely agree that rewriting ⁠vector.cpp⁠ to natively output correct optical energy to a proper shader chain (what okaz-code is pursuing) is the absolute right way to go for upstream!

Anyway, I don't want to derail this thread any further. I'll step back and let @okaz-code do things the proper way— I will be following the progress with great interest. If these new overdrive/glow physics end up being cleanly integrated into the core ⁠vector.cpp⁠, I will definitely grab those upstream changes to clean up my own renderer and improve the Android port. Keep up the fantastic work. Cheers!

okaz-code:
Fixed a per-frame render target recreation bug affecting rotated (vertical) vector games.

inject_vector_screen() sets the NATIVE target dimensions window-oriented (unswapped), but process_screen_chains() re-applied the screen rotation swap to the same dimensions. On swap_xy games (e.g. Tempest) the dimensions flipped every frame, so all NATIVE render targets were destroyed and recreated each frame, causing severe performance and rendering issues.

The fix skips the rotation swap for the synthetic vector prim (m_prim == nullptr), and also stops passing rotation metadata to the shaders for it — the FBO is drawn in window space with rotation already applied, so shaders like fs_distortion would otherwise distort the aspect ratio.

seleuco:
Hey @okaz-code, just wanted to drop a quick update!

I finally got the Star Wars global bloom/overdrive effect integrated on my end. It's one of my favorite arcade boards ever, and honestly playing it like this made me realize the effect isn't just eye candy. It actually affects gameplay. The Atari guys clearly added that overload to give you physical feedback when the lasers hit your shields haha.

Once you get the core overdrive logic split into that follow-up PR and if it gets merged, I'll definitely rip out my workarounds and integrate it the proper upstream way.

A quick heads up for that future PR though: I noticed you were filtering the effect by game type up at the BGFX chain level ¿?. I had to be really careful with that, because I found out the hard way that if I don't strictly discriminate by hardware type, I end up with Asteroids shooting massive lightning bolts everywhere lol.

Also, handling the vectors outside the screen area was a huge pain. I had to make sure I was specifically targeting the Star Wars board because other boards do their own custom clipping depending on game markers or state, and trying to normalize all that off-screen vector energy was becoming a total mess. Just something to keep in mind when you expose the overdrive intensity it doesn't bleed into standard vector games. Keeping the energy data generic is fine, but deciding if it should be drawn is a whole different beast.

On another note, I finally threw my "optical bloom via geometry" hack in the trash. I swapped it out for a proper Dual-Filter Kawase bloom. Turns out it's actually cheap enough for mobile GLES, unlike gaussian. Now everything just clicks and the HDR really shines. Booting up Battlezone in true HDR is just an absolute luxury.

I know you're pushing towards this too, vectors in HDR are just spectacular. Keep it up!

okaz-code:
Hey @seleuco, thanks for the great write-up — glad the Star Wars overdrive is working for you! And nice point that it's real gameplay feedback, not just eye candy.

Your warnings are well taken — and they match how I've been treading. The hard part isn't carrying the energy data, it's deciding whether to draw it. I keep beam_energy as generic per-vector data (harmless to non-AVG games) and gate the overdrive/off-screen glow so it can't bleed into standard vector titles — no Asteroids lightning bolts here either. 😄

For the follow-up PR I'll keep the per-hardware discrimination explicit and the trigger opt-in, scoping it conservatively rather than generalizing the off-screen energy. I'd rather land it narrow and correct than broad and surprising.

Also great call swapping in Dual-Filter Kawase — and yeah, vectors in HDR are spectacular, definitely pushing that way too. Thanks again!

okaz-code:
Given that #15692 takes a dedicated-renderer approach that's now approved and enabled by default, it makes more sense to build on that foundation than to land this chain-based path in parallel. I'll close this in favor of [#15692](https://github.com/amaiorano/vectrexy/issues/15692).

I'd like to revisit the parts requested here as small, opt-in follow-ups on top of [#15692](https://github.com/amaiorano/vectrexy/issues/15692) once it merges. Happy to collaborate with @hansandersson92 / @seleuco on that.

wolfsoftgmbh:
Hi @okaz-code — following up here since issues aren't enabled on VecBeamMAME.

  I'm trying to get native HDR vector output working on Linux, and before I start duplicating your work I'd rather ask.

  Reading your fork: the HDR path in drawbgfx.cpp (~480–492) is gated to Windows D3D11/D3D12 plus macOS Metal EDR, 3rdparty/bgfx/src/renderer_vk.cpp is
  untouched (still VK_COLOR_SPACE_SRGB_NONLINEAR_KHR hardcoded), and hdr-settings.md doesn't mention Linux. So I read that as "not started" rather than
  "broken".

  The blocker looks like it sits below MAME, in bgfx: VK_EXT_swapchain_colorspace is never requested in the instance extension table, so the driver only ever
  reports sRGB nonlinear — HDR10_ST2084_EXT and EXTENDED_SRGB_LINEAR_EXT are invisible no matter what MAME asks for. Upstream bgfx implements BGFX_RESET_HDR10
  only in dxgi.cpp; the GL backend has none, and Metal has had bkaradzic/bgfx#2974 open since 2023. Your fs_vector_hdr_present.sc already does correct
  component-wise ST.2084, so the shader side looks done — it's the swapchain.

  Three questions:
 
  1. Is a Linux/Vulkan HDR path on your roadmap, or already started somewhere I haven't looked?
  2. If the bgfx work gets done — request the extension, honour BGFX_RESET_HDR10 in the Vulkan backend, select HDR10_ST2084_EXT with A2B10G10R10_UNORM_PACK32
     or EXTENDED_SRGB_LINEAR_EXT with R16G16B16A16_SFLOAT, plus VK_EXT_hdr_metadata — would you want it against upstream bgfx, or patched into MAME's vendored
     copy first? (That copy has drifted: BGFX_CAPS_HDR10 is 0x100 there versus 0x80 upstream, and mamedev/mame#13054 has been open since April.)
  3. Should the Linux path reuse -bgfx_hdr, -bgfx_hdr_paper_white and -bgfx_hdr_display_peak unchanged, or do you expect those to move?

  I'd rather send the Vulkan/Linux side your way than maintain a parallel fork. I have hardware to actually verify it — RTX 4080 SUPER on Wayland (Hyprland
  0.56.2, nvidia-open 610.57.04) driving an LG OLED that reports PQ and HLG — which seems to be the combination nobody has tested. Same machine dual-boots
  Windows, so I can compare against your build on the identical panel.

  Back in June you wrote that vectors in HDR are spectacular and you were pushing that way too. I'd rather push in the same direction than alongside it.

okaz-code:
Hi @wolfsoftgmbh — thanks for asking first. Your reading is correct on all three counts.

**1. Linux/Vulkan HDR**

Not started. I don't have a Linux machine set up at the moment; the plan was to get to it on one of my own PCs once the current work settles. So the offer is very welcome — you'd be doing it sooner and on better hardware than I would.

**2. Where to send it**

`3rdparty/bgfx` in my fork, please. That isn't a new decision: the macOS EDR present path is a patch to that same copy, for the same reason you give.

Sending it to upstream bgfx in parallel is fine by me, but please don't let that block the Linux path.

No need to wait for me either. Start against the current branch whenever you like — I'm cutting `main` in the next week or two and will take it in there. I'll enable Issues at the same time.

For context on why MAME-side upstreaming isn't on the table: I'm close to an unofficial release with EDR/HDR on Windows and macOS, and getting the whole fork upstream would be a sizeable piece of work on its own. Not a "no", just not now.

**3. Options**

All three are unchanged — please reuse them as they are. If something Linux-specific turns out to be needed, that's your call.

A Linux section in the docs would be appreciated too.

Good to have someone pushing the same way.

wolfsoftgmbh:
Linux/Vulkan HDR10 is in: okaz-code/VecBeamMAME#1. Two commits against `3rdparty/bgfx`, tested on Hyprland + NVIDIA + LG C2. Details and a few findings (present-rate auto-detect, presentation timer cadence, raster colours) in the PR text.

ICEknigh7:
Since this got bumped, was wondering if there's still a chance to have these great-looking effects also implemented in main MAME, now that the other related PR was merged?

okaz-code:
Thanks for asking. Good to see #15692 land.

The two aren't quite the same thing underneath, though. #15692 *approximates* scan order from the display-list order, which is what lets it leave `render_primitive` alone — and probably part of why it merged cleanly.

This one derives each vector's actual start and end time from the AVG's own cycles, so the emulation core has to hand the renderer timing and beam energy. Overload alone would only need one field on `render_primitive` (defaulted, so no existing call sites change), but monitor glow passes a map of off-screen beam coverage — 32 angles by 16 depths — through the core as well. That means adding a vector-specific structure to the core, and that's the part that's hard to push upstream.

Right now I'm close to an unofficial release covering HDR/EDR on Windows and macOS, and getting the whole fork upstream is a sizeable piece of work on its own. Not a "no" — just not now. After the release I'll look again at what could reasonably be split out.

Good to know there's interest.

Lord-Nightmare:
There is definitely still interest in seeing this merged in some form or another!

okaz-code:
Thanks — that's good to hear.

After the release I'll look at putting up a PR for it. Overload is the first candidate: one defaulted field on `render_primitive` is enough, so no existing call site changes. It should sit on top of #15692.

PR #15430, opened 2026-06-02
kaypro/kaypro.cpp: Emulate the WD1002-HD0 hard disk controller on the Kaypro 10

davidlrand:
## Summary

Adds emulation of the WD1002-HD0 Winchester controller and its 10 MB ST506 drive to the two Kaypro 10 machine configurations, which previously had floppy-only support (the driver carried a `// need to add hard drive & controller` placeholder):

- `kaypro10` — 1983, mainboard 81-180, V1.9 ROM
- `kaypro1084` — 1984, mainboard 81-181, V1.9E ROM

Both boards use the same controller at the same I/O ports and differ only in the surrounding I/O map (the '83 board has no Z80 PIO / RTC), so the controller wiring is shared through one `add_hdc()` helper.

## Hardware

The WD1002-HD0 is built around the WD1010 task-file controller already emulated in `machine/wd1010.cpp`. On the Kaypro 10 it appears at:

- `0x80` — board sector-buffer data port (the host moves a 512-byte sector through it with a single `INIR`/`OTIR` burst)
- `0x81-0x87` — WD1010 task file (error/precomp, sector count, sector number, cylinder low/high, SDH, status/command)

Drive geometry is 306 cylinders × 4 heads × 17 sectors × 512 bytes (~10 MB); CP/M uses 16 of the 17 sectors per track. An image is created with:

    chdman createhd -o kaypro10.chd -chs 306,4,17 -ss 512

## Implementation notes

- A 512-byte board buffer (`m_hdc_buf`) sits between the host and the WD1010: the host fills/drains it at `0x80` while the controller drains/fills it through the device's data callbacks. Buffer-ready (`brdy_w`) is asserted when a full sector has moved and reset at each command write.
- The Kaypro 10 has a single fixed Winchester whose drive-select latch presents it as WD1010 "drive 1"; the SDH write collapses any drive-select onto the one configured drive so the controller always finds the disk.
- No new device type is introduced — this reuses the existing `wd1010_device` and `harddisk_image_device`. The HD devices are `optional_device`s and the `machine_start` setup is guarded by `m_hdc.found()`, so floppy-only operation is unchanged.

## Provenance

The port map, command set, geometry and buffer/handshake behaviour were derived from period Kaypro 10 CP/M disk-utility and BIOS source and verified against the V1.9 (81-188) and V1.9E (81-302) ROMs.

## Testing

- `kaypro10` and `kaypro1084` both boot CP/M from floppy unchanged.
- With a CHD attached (`-hard1`), the period Kaypro 10 hard-disk utilities run correctly: `FORMAT04.COM` (a partial low-level format that is largely a surface verify) completes, and `FINDBAD.COM` reads the entire disk surface (a full bad-sector scan) without error.
- Floppy-only operation (no `-hard1`) is unaffected.

## Related

`machine/wd1010.cpp` currently ships with `VERBOSE` set to all log categories, so any WD1010 user (including this one) floods `error.log` under `-log`. A separate one-line PR resets that to `(0)`.

### DIFF
--- a/src/mame/kaypro/kaypro.cpp
+++ b/src/mame/kaypro/kaypro.cpp
@@ -124,6 +124,88 @@ void kaypro84_state::kaypro484_io(address_map &map)
 	map(0x24, 0x24).mirror(3).rw(FUNC(kaypro84_state::rtc_r), FUNC(kaypro84_state::rtc_w));
 }
 
+// Kaypro 10 ('84) with the WD1002-HD0 Winchester controller wired up.
+// 0x80 is the board's sector-buffer data port (the host bursts a 512-byte
+// sector through it with INIR/OTIR, then polls BUSY/DRQ at 0x87); 0x81-0x87
+// are the WD1010 task-file registers 1-7 (error/precomp, sector count, sector
+// number, cyl low, cyl high, SDH, status/command) -- see the V1.9E ROM source.
+void kaypro84_state::kaypro1084_io(address_map &map)
+{
+	kaypro484_io(map);
+	map(0x80, 0x87).rw(FUNC(kaypro84_state::hdc_r), FUNC(kaypro84_state::hdc_w));
+}
+
+// Kaypro 10 (1983, Bd. 81-180) with the WD1002-HD0 wired up.  Same controller
+// at the same ports as the '84, but on the simpler kaypro10_io (no z80pio/RTC).
+void kaypro84_state::kaypro10hd_io(address_map &map)
+{
+	kaypro10_io(map);
+	map(0x80, 0x87).rw(FUNC(kaypro84_state::hdc_r), FUNC(kaypro84_state::hdc_w));
+}
+
+u8 kaypro84_state::hdc_r(offs_t offset)
+{
+	if (offset == 0)   // 0x80: host reads the buffered sector (data already in the buffer)
+	{
+		u8 const data = m_hdc_buf[m_hdc_ptr & 0x1ff];
+		if (!machine().side_effects_disabled())
+			m_hdc_ptr++;
+		return data;
+	}
+	return m_hdc->read(offset);   // 0x81-0x87 -> WD1010 task-file regs 1-7
+}
+
+void kaypro84_state::hdc_w(offs_t offset, u8 data)
+{
+	if (offset == 0)   // 0x80: WD1002 sector-buffer data port
+	{
+		m_hdc_buf[m_hdc_ptr & 0x1ff] = data;
+		if ((m_hdc_ptr++ & 0x1ff) == 0x1ff)
+			m_hdc->brdy_w(1);   // host has filled a full 512-byte sector
+		return;
+	}
+	if (offset == 6)   // SDH (Size/Drive/Head)
+	{
+		// Head select comes from SDH bits 0-2 (MAME's wd1010 takes head via head_w).
+		m_hdc->head_w(data & 0x07);
+		// The Kaypro 10 has one fixed Winchester; its drive-select latch presents it
+		// as WD1010 "drive 1" (dsel01=08h).  Collapse any drive-select onto the single
+		// configured drive (0) so the controller always finds the disk.
+		data &= ~0x18;
+	}
+	if (offset == 7)   // command register: start each command with a clean buffer handshake
+	{
+		m_hdc->brdy_w(0);
+		m_hdc_ptr = 0;
+	}
+	m_hdc->write(offset, data);   // 0x81-0x87 -> WD1010 task-file regs 1-7
+}
+
+// WD1010 <-> board sector buffer: the controller drains/fills the buffer here
+// (no host-side BRDY handshake on these accesses).
+u8 kaypro84_state::hdc_buf_in()
+{
+	u8 const data = m_hdc_buf[m_hdc_ptr & 0x1ff];
+	m_hdc_ptr++;
+	return data;
+}
+
+void kaypro84_state::hdc_buf_out(u8 data)
+{
+	m_hdc_buf[m_hdc_ptr & 0x1ff] = data;
+	// once the controller has filled a whole sector, the board reports buffer-ready so the
+	// READ command completes -- the host then reads the buffer afterward (or not, on a verify).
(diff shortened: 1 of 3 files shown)


davidlrand:
**Additional validation — full factory hard-disk restore.**

I ran the original 8-disk *Kaypro 10 Autoload* set (CP/M 2.2H) against a freshly-created blank CHD. All eight diskettes restore end-to-end with no errors: the autoload procedure low-level formats the drive and `PIP`s the complete factory suite — CP/M 2.2H, WordStar 3.3, dBASE II, MicroPlan, the C/S/M-BASICs, InfoStar/DataStar, MITE, etc. — into the CP/M user areas, finishing with `RELOAD PROCEDURE COMPLETE`. The machine then **boots CP/M from the hard disk** into the Kaypro "Master Menu" and the installed software runs.

This exercises the full WD1002-HD0 path under a real period workload: low-level format, full-surface write, the task-file registers, the `0x80` sector-buffer / `BRDY` handshake, and multi-track seeking — in addition to the `FORMAT04`/`FINDBAD` checks already noted.

Screenshot 2026-06-02 at 9 24 48 AM
Screenshot 2026-06-02 at 9 27 01 AM

MooglyGuy:
Question: Is it documented whether the WD1010 actually listens on address 0? According to wd1010.cpp, both reads and writes to that particular register are invalid, so I'm genuinely not sure.

Having the whole if offset 0/if offset 6/if offset 7 bit is a little awkward. I would normally suggest that we've been trying to move away from that sort of "check the offset within a handler and then dispatch more from there" pattern in favor of specific handlers at specific offsets. However, you wouldn't be able to map the contiguous portions of `wd1010_device::read` and `wd1010_device::write`, as I don't believe MAME's memory system has a way to bind a handler that adds a fixed offset on the way into the handler (correct me if I'm wrong, @galibert).

Personally, I think I'd be fine with the code going in as-is. Would you be alright with me doing a bit more wrangling on the I/O maps once it goes in?

davidlrand:
Oh hell yes.  My experience on the k10 was writing a high-speed hard drive backup program, back in the day: kback.  This did all kinds of fun tricks.  I was able to format floppy media, and write to it in about 20-25 seconds per disk, copying from the hard drive.  If you can make it better, please do so.

Right now, I'm working on the "other" big board system, the Xerox 820 and 820-II.  The controller on the 820-II was... not complete, so there is a fair amount of work.

davidlrand:
On the WD1002‑HD0, 0x80 isn't a WD1010 register — it's the board's 512‑byte sector buffer, handled entirely in the driver (offset 0 returns before reaching the device). Only 0x81–0x87 reach the WD1010, as task‑file regs 1–7; its reg 0 (data) is never touched, which is why wd1010.cpp can keep reg 0 read/write invalid — the board buffer replaces it. That +1 (regs land at 1–7, not 0–6) plus the SDH/command intercepts at 0x86/0x87 are why it's one dispatch handler rather than discrete maps.

MooglyGuy:
Sounds good, I'll wait for Galibert to weigh in since I always prefer to get the go-ahead from two people, but I'll be happy to make any alterations other team members request once it goes in.

pmackinlay:
> Having the whole if offset 0/if offset 6/if offset 7 bit is a little awkward. I would normally suggest that we've been trying to move away from that sort of "check the offset within a handler and then dispatch more from there" pattern in favor of specific handlers at specific offsets. However, you wouldn't be able to map the contiguous portions of `wd1010_device::read` and `wd1010_device::write`, as I don't believe MAME's memory system has a way to bind a handler that adds a fixed offset on the way into the handler (correct me if I'm wrong, @galibert).

There's nothing built into emumem to handle this explicitly, however in this case where the ports are sequential and the one specific offset (offset 0) is repurposed, it's very easy to get the right effect thusly:

```c++
map(0, n).rw(...device...);
map(0, 0).rw(...buffer...);
```

The handlers from 1..n will call the device with offsets 1..n, and offset 0 will go to the buffer because that specific handler has been overwritten. Maybe it's not so intuitive at first glance, so a comment might be useful, but it works and it avoids the conditionals in the handlers. Any other "holes" in the map can be created using `unmaprw()` etc. if needed.

src/mame/kaypro/kaypro.cpp
```diff
@@ -124,6 +124,88 @@ void kaypro84_state::kaypro484_io(address_map &map)
...
 
+// Kaypro 10 ('84) with the WD1002-HD0 Winchester controller wired up.
+// 0x80 is the board's sector-buffer data port (the host bursts a 512-byte
+// sector through it with INIR/OTIR, then polls BUSY/DRQ at 0x87); 0x81-0x87
+// are the WD1010 task-file registers 1-7 (error/precomp, sector count, sector
+// number, cyl low, cyl high, SDH, status/command) -- see the V1.9E ROM source.
+void kaypro84_state::kaypro1084_io(address_map &map)
+{
+	kaypro484_io(map);
+	map(0x80, 0x87).rw(FUNC(kaypro84_state::hdc_r), FUNC(kaypro84_state::hdc_w));
```
pmackinlay: Install a separate handler for offset 0 _after_ the main handler to replace the offset 0 handler while preserving the 1..7 offsets for the main handler, e.g. `map(0x80, 0x80).rw(...);`

src/mame/kaypro/kaypro.h
```diff
@@ -157,6 +169,10 @@ class kaypro84_state : public kaypro_state
 
 	required_device<mc6845_device> m_crtc;
 	optional_device<mm58167_device> m_rtc;
+	optional_device<wd1010_device> m_hdc;          // WD1002-HD0 Winchester controller (kaypro10/1084)
```
pmackinlay: Can you refactor into without-HD and with-HD state classes to avoid the `optional_device`s here?

pmackinlay (changes requested):
Pending annoying refactoring to avoid `optional_device`, this looks good to me.

src/mame/kaypro/kaypro.cpp
davidlrand: Done

src/mame/kaypro/kaypro.h
davidlrand: Done.

PR #15434, opened 2026-06-03
cpu/z80/z80.cpp: Prioritized non maskable interrupt over maskable in case of race.

holub:


happppp:
Z80 does not work like this, does it?
1st opcode of IRQ routine is guaranteed to be processed, an NMI can't get in between IRQ ack and 1st IRQ opcode. NMI will be processed after the end of an opcode.

holub:
The /NMI pin is sampled on the falling edge of the last T-state of every machine cycle (M-cycle, not instruction).

happppp:
Ok, but take_nmi is not the NMI edge detection, it's for processing the NMI, which occurs after the last M cycle of an instruction. Not after the IRQ ack.

holub:
The fix is not about "after IRQ ACK". The fix is about to cover the case when NMI happened during IRQ ACK which takes 17(?) cycles. On hardware if that happen, NMI still wins. In our previous implementation IRQ wins and allow to execute one instruction from ISR, and actual NMI was "postponed" by ONE instruction.

happppp:
"In our previous implementation IRQ wins and allow to execute one instruction from ISR, and actual NMI was "postponed" by ONE instruction." <-- I'm pretty sure real Z80 works like that: NMI will be processed after the 1st opcode of the normal IRQ routine if NMI edge was during IRQ processing. So yes, IRQ "wins" if the NMI edge occurs during IRQ ack (eg. NMI edge while the CPU is busy pushing program counter to stack/jumping to IRQ routine). Or better worded, since it already started IRQ processing, IRQ had already 'won'.

CBrunschen linked a Zilog doc on discord that also claims that NMI is processed after the last cycle of an instruction.

cbrunschen:
> CBrunschen linked a Zilog doc on discord that also claims that NMI is processed after the last cycle of an instruction.

This is the document, which includes a flowchart and timing diagrams:
 [THE Z80 FAMILY PROGRAM INTERRUPT STRUCTURE](http://www.z80.info/zip/z80-interrupts_rewritten.pdf)

It describes:

> The bus request line (/BUSRQ) will be sampled (Internal CPU Bus Request Flip Flop set if
> active (low)) and if active, will be acknowledged at the end of a machine cycle. If the last
> cycle is also the end of the instruction, the CPU will recognize an interrupt (non­maskable
> or maskable) by setting appropriate internal flip flops, but will first service any pending Bus
> Request. At completion of a Bus Request operation, the CPU checks the status of the
> /NMI flip flop. If set, it executes the /NMI interrupt routine as indicated in Figure 2.

PR #15435, opened 2026-06-03
added ROMs for Peter Pack Rat T11 prototype

mamehaze:
new NOT WORKING clones
----
Peter Pack Rat (T11 hardware, prototype, set 1) [Dutchman2000]
Peter Pack Rat (T11 hardware, prototype, set 2) [Dutchman2000]

### DIFF
--- a/src/mame/atari/pakrat11.cpp
+++ b/src/mame/atari/pakrat11.cpp
@@ -1,0 +1,208 @@
+// license:BSD-3-Clause
+// copyright-holders:David Haywood
+
+/*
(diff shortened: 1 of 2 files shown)


mamehaze:
I'm not in any way convinced this is a matching set of prg / gfx etc. or that it can really be made to run, but the T11 program ROMs are at least interesting.

there don't seem to be any background/text tile graphics in there, unless they're uploaded by the program, but the program isn't very big.

the T11 seems to hit something invalid (unless there's something special I need to do)

not planning on doing anything else with this myself.

PR #15438, opened 2026-06-03
nintendo/nes_vt369_vtunknown.cpp: Add 36-in-1 Classic Games phone case

naorunaoru:
I saw a video called "Dumping and reverse engineering a very cheap NES phone case" by Poking Technology and was quite intrigued by the approach, however the ending wasn't too satisfying so I decided to try poking the technology myself. 

And now it works: 
image
image

36pcase.zip should contain:
```
25q16.ic3        2,097,152 bytes  CRC a8edb73e
internal.bin         4,096 bytes  CRC da5850f0
serial-rom.bin         256 bytes  CRC 877fb90b
```

serial-rom.bin replaces the earlier NO_DUMP "mystery chip.bin" entry (again, thanks mr. Poking for using such an extravagant method of dumping the ROM).

- Added ROM definitions for the verified external, internal and serial dumps for the 36-in-1 Classic Games phone case.
- Added the VT3xx split GPIO/protection exchanges needed by this hardware.
- Added the LCDC/tilemap and DMA behaviour needed for the language selector, game menu and selected games to boot.

new WORKING machines
--------------
36-in-1 Classic Games phone case [naorunaoru]

### DIFF
--- a/src/devices/video/ppu2c0x.cpp
+++ b/src/devices/video/ppu2c0x.cpp
@@ -299,6 +299,16 @@ u8 ppu2c0x_device::readbyte(offs_t address)
 	return space().read_byte(address);
 }
 
+u8 ppu2c0x_device::ppu_vram_direct_read(offs_t address)
+{
+	return readbyte(address);
+}
+
+void ppu2c0x_device::ppu_vram_direct_write(offs_t address, u8 data)
+{
+	writebyte(address, data);
+}
+
 
 //-------------------------------------------------
 //  writebyte - write a byte at the given address
@@ -1381,6 +1391,11 @@ u16 ppu2c0x_device::get_vram_dest()
 void ppu2c0x_device::set_vram_dest(u16 dest)
 {
 	m_videomem_addr = dest;
+}
+
+void ppu2c0x_device::reload_refresh_data()
+{
+	m_refresh_data = m_refresh_latch;
 }
 
 /*************************************
--- a/src/devices/video/ppu2c0x.h
+++ b/src/devices/video/ppu2c0x.h
@@ -99,6 +99,10 @@ public:
 
 	u16 get_vram_dest();
 	void set_vram_dest(u16 dest);
+	void reload_refresh_data();
+
+	u8 ppu_vram_direct_read(offs_t address);
+	void ppu_vram_direct_write(offs_t address, u8 data);
 
 	bool in_vblanking() { return (m_scanline >= m_vblank_first_scanline - 1); }
 protected:
--- a/src/devices/video/ppu2c0x_vt.cpp
+++ b/src/devices/video/ppu2c0x_vt.cpp
@@ -992,11 +992,7 @@ void ppu_vt3xx_device::shift_tile_plane_data(u8 &pix)
 void ppu_vt3xx_device::draw_background(u8 *line_priority)
 {
 	const bool is_vt369_64_byte_names = (m_newvid_1c == 0x12) && (m_newvid_1d == 0x09) && (m_newvid_1e == 0x0f)
-			&& (m_204x_screenregs[0] == 0x40) && (m_204x_screenregs[1] == 0xa1)
-			&& (m_204x_screenregs[2] == 0x00) && (m_204x_screenregs[3] == 0x78)
-			&& (m_204x_screenregs[4] == 0xff) && (m_204x_screenregs[5] == 0x04)
-			&& (m_204x_screenregs[6] == 0x0a) && (m_204x_screenregs[7] == 0xd8)
-			&& (m_204x_screenregs[8] == 0x0a) && (m_204x_screenregs[9] == 0x00);
+			&& BIT(m_tilebases_2x[0], 4);
 
 	if (!is_vt369_64_byte_names)
 	{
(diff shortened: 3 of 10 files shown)


mamehaze:
This also fixes broken elements in many of the other games, which I'm assuming is the new DMA handling

mamehaze:
it does however make a few other things worse, eg the 'heading' game in urban240 (and other sets) which now shows garbage instead of the proper background.

new
old

I imagine it's pointing in the right direction though (it fixes far more than it breaks)

mamehaze:
anyway @ajrhacker how do you feel about this? I'm generally OK with it, although I do wonder if this deferred DMA hook-up is correct.  definitely points in the right direction though as many game elements now update that weren't updating properly before.

quite a few extended VT modes are still missing / badly supported, so maybe the urban240 regression is more to do with that (wrong bits used to select extra modes)

also at some point we need to consider how to better support LCDC output anyway, because of sets like the Cyber Arcade Pocket which flip the output direction between scenes (hence why the menus render on their side right now) but that will somehow need the old 'screen' device (but without visual output, rendering to a buffer) for the timings, and copying to the LCDC buffer with actual screen output for display depending on the parameters written.

the changes here would IMHO allow for the majority of the Compact Cyber Arcade sets to be promoted though, with the exception of the 120-in-1 Spiderman ones as they have more hacked up games using currently unsupported sprite modes etc. compared to the others.

the CL failure is just the validation failure because somebody broke master, there's no CL issue with this PR.

mamehaze:
another issue I've noticed is that for the various versions of "move fun" the lives counter (stars near the timer) don't update properly still, previously these didn't update at all (nor was the timer or any of the pieces drawn) so is definitely tied to the new behaviour, but suggests something isn't quite right.

(this is from rtvgc300pj, but the same can be seen in the version in zonefusn etc.)




also sometimes when scrolling through the menu on said unit it sometimes copied bad tiles over the display at random, suggesting a DMA being triggered that shouldn't be.

naorunaoru:
I have a few ideas about why this might've broken "heading", it just didn't occur to me to test other vt3xx machines (great thinking on my side). I'll see if it can be improved.

naorunaoru:
Ok so LCDC registers differ between those two:
urban240 Heading: `40 a0 01 78 ff 69 0a 68 26 00`
36pcase: `40 a1 00 78 ff 04 0a d8 0a 00`

I changed ppu2c0x_vt.cpp:994 to use the 64-byte-name-row drawer only when both the video tuple and the 36pcase-style LCDC profile match.

image
image

I'll look at "move fun" next.

mamehaze:
I doubt it's tied to the LCDC registers, more likely the extended mode registers somehow.

LCDC registers should only control how the (already rendered) image is translated onto the LCDC screen (timing, orientation, how much the image is stretched etc.) not the rendering directly.

There are other extended modes (especially in those sets) that don't work properly yet too, probably because the exact meaning of the extended video bits isn't fully understood.

naorunaoru:
Fair, that was a weak guess. 

Another signal that looks plausible to me is $2020 bit 4 – if this defines VT369 extended background graphics base, this might be relevant.

I pushed it here but the confidence is still pretty low, even if a bit higher than for that LCDC hack.

mamehaze:
basketball in the same units seems to be another test for the same thing, which may suggest a more complex check is needed.

once things start moving it ends up with bad backgrounds now.

i appreciate that trying to nail these things down when there's no documentation is difficult (I spent far too long looking at these things myself, which is why I appreciate what's being done here as I was at a dead end with the DMA stuff)

mamehaze:
lxcmcysp game 114 is interesting too, it's the same 'heading' game, but coded for vertical orientation.  with current code background only renders properly when you're pressing 'right'

naorunaoru:
Apparently checking `m_tilebases_2x[2] == 0x08` is enough. 

36pcase language selector/menu is ok, urban240 "Header" and "Basketball" work, lxcmcysp game 114 looks ok too. 

"Move fun" still borks though.

mamehaze:
I've just got out my "lxcmcyspn" and game "74 Fruit Fall" has a broken lives counter even on the unit, so it's possible this game just wasn't ported correctly. 

The same unit has "037 Move Fun" and it works on there (likewise in MAME for that version of the game, which I think is the VT03 port)

So I don't think you have to chase that one, I think they just messed up the programming of the VT3xx versions.

mamehaze:
same for sealvt game 07 'Move Fun' on real hardware, it's just broken, so not an emulation issue, just a broken port.

naorunaoru:
Thank you for confirming this. I tidied new logerror calls – they were mostly used to trace timer behavior, but I'm fine with removing them altogether if that's not to convention.

mamehaze:
I'm fine with the additional logging as it's handy for debugging, although at some point the files need going over and converting to LOGMASKED so that logging can be more easily enabled / disabled depending on what we want to log.

mamehaze:
Anyway, I'm not finding any regressions with this in my own testing now, so would be happy to see it merged in.

I think some of the 36pcase stuff should end up driver-side, rather than in the SoC core (the GPIO responses seem very specific to that unit) however nothing else seems to depend on it, which means there's a lack of evidence to use when splitting it out so for now this will do IMHO.

PR #15439, opened 2026-06-03
Update nl_tank.cpp

RetroShrimp:
Primary Source Match: The Board I Rev-H schematic (sheet 035-110) shows that both players' explosions should use the same self-contained S-tap pattern (no tank sprite involvement).

The Error: The netlist currently routes K3.5 to K3.3 (which is the TANK1 sprite XORed with FLASH/BLO_UP).

The Proof: The correctly computed K2.3 net (the shell S2/S4/S8 XOR pool drawn on the schematic) is currently left completely dangling in MAME's netlist—there is no NET_C(IC_K2.3, ...) connection anywhere.

Symmetry: Player 2's explosion (BLAST2) is already transcribed correctly in the netlist using the S-tap pattern, proving the current BLAST1 wiring is an asymmetric error.

The driver is currently MACHINE_NOT_WORKING: The document confirms (as of MAME 0.288) that the tank driver is fundamentally broken regarding object rendering. The game never starts, there is no attract mode, and the tanks themselves do not render or move (only the maze and score digits render).

Fixing the BLAST1 netlist typo is a correctness and preservation fix for the schematic transcription, but it will not magically make the explosions visible in MAME until the broader tank rendering/sync pipeline is fixed.


MooglyGuy:
Is this AI-generated or are you an actual human studying the schematics?

MooglyGuy:
Not saying the change is invalid, but the overly verbose and technical description for such an otherwise-simple change that any normal person would simply note "Changed routing for K2.3 to more accurately reflect the schematic" has me a bit suspicious.

RetroShrimp:
The bug is confirmed. This is my first PR here, so I was just trying to provide as much information as possible.

PR #15441, opened 2026-06-04
Update atarittl.cpp

RetroShrimp:
The driver is currently MACHINE_NOT_WORKING: The document confirms (as of MAME 0.288) that the tank driver is fundamentally broken regarding object rendering. The game never starts, there is no attract mode, and the tanks themselves do not render or move (only the maze and score digits render).


MooglyGuy:
`MACHINE_IMPERFECT_GRAPHICS` is considered unnecessarily redundant to add to drivers that are marked `MACHINE_NOT_WORKING`.

PR #15443, opened 2026-06-04
Fix bootstrap INI priority to allow override by other INI files.

wallyweek:
Fixes regression introduced in #15310
Please see https://github.com/mamedev/mame/pull/15310#issuecomment-4582431546
for full issue explanation.

On bootstrap, MAME reads the ini file hardcoded via SDLOPTION_INIPATH with
OPTION_PRIORITY_MAME_INI priority. Later passes use the same priority, so
the hardcoded file's values are never overridden by user INI files.

Fix: introduce OPTION_PRIORITY_BOOTSTRAP = OPTION_PRIORITY_DEFAULT + 1 and
use it for the bootstrap pass, ensuring any subsequent INI pass can override
its values.

### DIFF
--- a/src/frontend/mame/mameopts.cpp
+++ b/src/frontend/mame/mameopts.cpp
@@ -46,7 +46,7 @@ void mame_options::parse_standard_inis(emu_options &options, std::ostream &error
 			osd_printf_verbose("Parsing %s\n", file.fullpath());
 			try
 			{
-				options.parse_ini_file(static_cast<util::core_file &>(file), OPTION_PRIORITY_MAME_INI, true, false);
+				options.parse_ini_file(static_cast<util::core_file &>(file), OPTION_PRIORITY_BOOTSTRAP, true, false);
 			}
 			catch (options_exception const &ex)
 			{
--- a/src/frontend/mame/mameopts.h
+++ b/src/frontend/mame/mameopts.h
@@ -23,11 +23,11 @@
 // option priorities
 enum
 {
-	// command-line options are HIGH priority
+	// command-line options override everything (HIGH priority)
 	OPTION_PRIORITY_SUBCMD = OPTION_PRIORITY_HIGH,
 	OPTION_PRIORITY_CMDLINE,
 
-	// INI-based options are NORMAL priority, in increasing order:
+	// user INI files, in increasing order of specificity (NORMAL priority)
 	OPTION_PRIORITY_MAME_INI = OPTION_PRIORITY_NORMAL + 1,
 	OPTION_PRIORITY_DEBUG_INI,
 	OPTION_PRIORITY_ORIENTATION_INI,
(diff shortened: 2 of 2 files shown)


belegdol:
I have given this a brief test and it appears to test the issue I reported in https://github.com/mamedev/mame/pull/15310#issuecomment-4582431546 and https://github.com/mamedev/mame/pull/15310#issuecomment-4582522501. I have compiled mame with `SDL_INI_PATH=/etc/mame` and configured `inipath` to `$XDG_CONFIG_HOME/mame;$HOME/.config/mame;$HOME/.mame/ini;/etc/mame` in `/etc/mame/mame.ini`. `rompath` values configured in `~/mame/ini/mame.ini` are getting picked up correctly.

cuavas:
No, this over-complicates things.

belegdol:
> No, this over-complicates things.

What would be the recommended solution then? Putting local and system-wide ini paths into `SDL_INI_PATH` during compilation? While it does work, it makes `inipath` setting in the config somewhat useless.

wallyweek:
@cuavas The patch simply adds and uses a new enum value; there are no logic changes. I don't see how this could complicate things.

I also considered changing the test in src/lib/util/options.cpp at line 1046 from:
`if (curentry->priority() < priority)`
to:
`if (curentry->priority() <= priority)`

but this would also require reversing the parsing order for the first entry to take precedence over the last one.

Unless you mean that the whole priority mechanism is complicating things, but that would require a more substantial overhaul of the code.

It's up to you. However, something needs to be done because the current behavior is not the intended one: the hardcoded file takes precedence over every other file.

cuavas:
This change would make behaviour more complicated to explain, as it would mean there would be two distinct priority levels for `mame.ini` depending on the pass over it.

The idea of the two passes over `mame.ini` is that you can set the `inipath` setting in a `mame.ini` in a location that MAME searches by default, and then that’s used as the “real” setting, *including for where to save INI files*.  I’ll give examples for how this works in practice on macOS and Windows.

On macOS, the default `inipath` is `$HOME/Library/Application Support/mame;$HOME/.mame;.;ini`.  If a user wants their “MAME data root” to be `~/Documents/MAME` instead, they can create `$HOME/Library/Application Support/mame/mame.ini` with a single line `inipath $HOME/Documents/MAME/ini`, then have their “real” mame.ini at `$HOME/Documents/MAME/ini/mame.ini` and also have MAME save mame.ini there if they choose to save settings from MAME’s UI.

On Windows, the default `inipath` is `.;ini;ini/presets`.  If a user wants to have their MAME settings in their roaming profile, they can create `mame.ini` in the folder they extracted MAME to with a single line `inipath %APPDATA%\mame\ini` and… you get the idea.

On Linux, the default `inipath` is `$HOME/.mame;.;ini`.

Critically, in all cases, the first directory path in the `inipath` setting is a folder that the user can control, so they get to define what’s in that “first pass” INI file.

If you want to supply defaults in `/etc/mame/mame.ini`, you should add that to `SDL_INI_PATH` *after* the default user-controlled directories, e.g. set `SDL_INI_PATH` to `$HOME/.mame;.;/etc/mame` (with whatever escaping is necessary).  If you do it this way, then settings in the user’s `$HOME/.mame/mame.ini` will have priority, and settings from the distro-supplied `/etc/mame/mame.ini` will be used as a fall-back for settings that the user hasn’t set.

MAME is not designed to be used with only an administrator-controlled folder in the default `inipath` value.

PR #15445, opened 2026-06-04
adp/stellafr: fix display, mute, serial and timekeeper

stonedDiscord:
https://github.com/user-attachments/assets/06777a0b-914c-4b96-a6f5-b04133e2f5ac

Display is based on the Disc 4000 manual, the boards are the same across the machines but the Anzout assignment varies a bit between machines:
https://archive.org/details/merkur-disc-4000

Mute pin is from my reversed schematic (They called it Knack).
https://kicanvas.org/?repo=https%3A%2F%2Fgithub.com%2FstonedDiscord%2FSteuereinheit

Serial is designed for testing during manufacturing and to print out the billing data.
The RS485 port is commented out, this would be used to connect multiple machines together for a jackpot system or burglar alarm.

The CPU board is designed to take either 2 large SRAMs with battery backup or instead 1 timekeeper + 1 zeropower RAM block without the 3V battery. Eventually they stopped buying the zeropower chips and only used timekeepers in both slots.

### DIFF
--- a/src/mame/adp/stellafr.cpp
+++ b/src/mame/adp/stellafr.cpp
@@ -230,13 +230,17 @@ private:
 	uint16_t m_mux1;
 	uint8_t m_anz2;
 	uint8_t m_mux2;
+	uint8_t m_strobe;
+	uint8_t m_anz_bank;
 
 	uint8_t mux_r();
 	void enable_w(uint8_t data);
 	void mux_w(uint8_t data);
 	void duart_output_w(uint8_t data);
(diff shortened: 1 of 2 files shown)


stonedDiscord:
Here's a video of the same error message on a real machine:
https://www.youtube.com/watch?v=2oR9OaJM8_M
It's asking to be initialized, which is done using the service keyboard.

stonedDiscord:
Going to use this as a PoC to come up with something cleaner later

PR #15448, opened 2026-06-05
bus/nscsi: move the opt-in seek-timing model into nscsi_harddisk_device

davidlrand:
## Summary

The cylinder-aware seek-timing model added to the Xebec S1410 (#15411) and
proposed for the DTC-510 (#15406) was identical copy/paste between the two
controllers. As suggested by @MooglyGuy on #15406, this hoists the model up
into their common base, `nscsi_harddisk_device`, so it is written once and is
available to any hard-disk controller built on it (including a plain
`NSCSI_HARDDISK`) without subclassing.

**Supersedes #15406** (DTC-510), which is closed in favour of this.

## What moves, and what deliberately does not

- **The opt-in model moves up.** `set_seek_timing()`, the model state, the
  fitted-exponent math (`seek_time()`), and a generic `scsi_data_command_delay()`
  now live in `nscsi_harddisk_device`.
- **The legacy flat 85 ms delay deliberately stays down** in `s1410` and
  `dtc510`. It is a per-controller quirk (Curt Coder added it to s1410 for
  abc1600 formatting; dtc510 inherited it for pc9801), not a property of all
  hard disks. Hoisting it would change behaviour for every `NSCSI_HARDDISK`
  user; keeping it local does not.

## Safety: no change for existing users

The base `scsi_data_command_delay()` returns `attotime::zero` (its inherited
default) unless a driver has called `set_seek_timing()`. So:

- Every current `nscsi_harddisk_device` user is byte-for-byte unchanged.
- `s1410` and `dtc510` keep their exact legacy behaviour: with no
  `set_seek_timing()` call they return the flat 85 ms for their own command
  set; the only difference is they now delegate the *model* branch to the base
  instead of carrying a private copy of it.

There is still no in-tree caller; this is groundwork (a Toshiba T-200/T-250
driver in development uses it for the machine's CDC 9410 Finch).

## Model

When configured, a command pays a real seek only when the target cylinder
changes (derived from the LBA and the CHD geometry); consecutive sectors on
one cylinder pay only rotational latency scaled by the format interleave. The
seek curve is fitted from the drive's published track-to-track / average /
full-stroke figures. The base applies this to the generic SCSI READ/WRITE/SEEK
opcodes in **both** their 6-byte (0x08/0x0a/0x0b) and 10-byte
(0x28/0x2a/0x2b) forms; subclasses with their own dispatch call `seek_time()`
for whatever opcodes move the heads.

## Parameter documentation

`set_seek_timing()`'s header comment now documents how to derive each of the
five parameters from an OEM manual, including the convention that the quoted
"average" seek is a ~1/3-stroke seek (which anchors the fitted exponent) and a
fallback when a datasheet supplies only two of the three seek times.

## Testing

- The `bus/nscsi` sources compile clean with no new warnings.
- `s1410` users (victor9k, mikromik, x37_sasi, db4105, idpartner_sasi) and the
  `dtc510` user (pc9801_27) are unchanged — the legacy path is untouched and
  the model is inert without a `set_seek_timing()` call.

### DIFF
--- a/src/devices/bus/nscsi/dtc510.cpp
+++ b/src/devices/bus/nscsi/dtc510.cpp
@@ -44,11 +44,6 @@ nscsi_dtc510_device::nscsi_dtc510_device(const machine_config &mconfig, const ch
 	std::fill(std::begin(m_param), std::end(m_param), 0);
 }
 
-void nscsi_dtc510_device::device_reset()
-{
-	nscsi_harddisk_device::device_reset();
-}
-
 bool nscsi_dtc510_device::scsi_command_done(uint8_t command, uint8_t length)
 {
 	switch(command >> 5) {
@@ -93,7 +88,7 @@ void nscsi_dtc510_device::scsi_command()
 			scsi_status_complete(SS_NOT_READY);
 			m_scsi_sense_buffer[0] = SK_DRIVE_NOT_READY;
 		} else {
-			// lba = 0;
+			m_last_cylinder = 0;   // recalibrate returns the heads to track 0
 			scsi_status_complete(SS_GOOD);
 		}
 		break;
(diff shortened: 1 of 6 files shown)


MooglyGuy:
This looks great to me. Excellent work! Barring any objections from @galibert (the resident SCSI-master), this seems perfectly fine to merge from my perspective.

PR #15450, opened 2026-06-05
misc/blitz68k.cpp: added MCU comms and a couple of new dumps

Osso13:
- added MCU comms for games with dumped MCUs, allows to avoid program ROM patching;
- with MCU comms working, added preliminary sound (has some problems, noted in comments);
- added hand-modified MCUs to a couple of other games to verify feasability, given unlikeness of many more MCUs getting dumped.

Reel Picker (Ver. 2.00) [Ioannis Bampoulas]

New clones marked not working
-----------------------------
Texas Reels (Ver. 1.02) [Ioannis Bampoulas]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -30230,10 +30230,12 @@ megastrp
 mpokerdx
 mpokerdx109
 poker52
+rpicker
 super97
 surpr5
 steaser
 texasrls
+texasrls102
 
 @source:misc/blocktax.cpp
 blocktax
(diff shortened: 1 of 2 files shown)


Osso13:
The reverse-engineering of the comms / sound was mostly done by AI (Claude), opening as PR so it may be reviewed by more knowledgeable people than me.
I've left the whole verbose AI comments so that the reasoning can be followed. They can be removed / reduced if the PR is deemed worhty of merging.

cuavas:
A whole lot of the code here looks like an overly verbose reimplementation of a pair of `generic_latch_8_device`.  Also, remember you need sync barriers, etc. when crossing scheduler domains.

PR #15453, opened 2026-06-06
bus/scsi/scsihle: deassert REQ between phases; hold BSY in COMMAND

davidlrand:
Two SCSI-HLE bus-semantics fixes.

1. `scsi_change_phase()` deasserts REQ on entry, so the initiator sees a clean REQ low→high edge with the new phase's C/D-I/O lines settled. Previously the per-byte REQ-deassert timer was cancelled by the next phase's re-assert, leaving REQ continuously high across e.g. COMMAND→DATA-IN, so an initiator that samples the control lines on the REQ edge read the stale previous-phase lines.

2. A selected target asserts BSY on entry to COMMAND, so a fast SEL deassert racing the sel_timer cannot leave BSY low during the connection.

Tested without regression on `v1050` (working; S1410 SASI hard disk: FMTWINCH format + COPYSYS + boot and access) and on an in-development Xerox 820-II SASI host adapter. Other legacy `scsihle` machines (fmtowns, pcd, bebox, rmnimbus, model3, slicer) were not regression-tested locally; both changes are conservative SCSI bus semantics.

### DIFF
--- a/src/devices/bus/scsi/scsihle.cpp
+++ b/src/devices/bus/scsi/scsihle.cpp
@@ -339,6 +339,14 @@ void scsihle_device::scsi_change_phase(uint8_t newphase)
 	cmd_idx=0;
 	data_idx=0;
 
+	// Deassert REQ at every phase transition so the initiator sees a clean REQ
+	// low->high edge with the new phase's C/D-I/O lines settled.  Otherwise the
+	// per-byte REQ-deassert timer is cancelled by the next phase's re-assert, so
+	// REQ can stay continuously high across (e.g.) COMMAND->DATA-IN and an
+	// initiator that polls the control lines on the REQ edge samples the stale
+	// previous-phase C/D/I/O.
+	output_req(0);
+
 	switch(m_phase)
 	{
 		case SCSI_PHASE_BUS_FREE:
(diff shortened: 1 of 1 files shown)


angelosa:
You aware that `v1050` uses a SCSI core that is _not_ the one in #15448 do you?

davidlrand:
> You aware that `v1050` uses a SCSI core that is _not_ the one in #15448 do you?

I am.  I am working on multiple big board variants.

angelosa:
I mean: `v1050` should really use the newer S1410 core, from NSCSI. Using the legacy core while not really checking if anything broke is not a viable strategy.

davidlrand:
> I mean: `v1050` should really use the newer S1410 core, from NSCSI. Using the legacy core while not really checking if anything broke is not a viable strategy.

Ok, but I'm not working on v1050. It is using the legacy core, and that's exactly why I checked against it: I formatted + boot/read its S1410 disk with this change in, and nothing broke. The nscsi-S1410 migration is a  reasonable separate cleanup, but not my goal for this patch.  Cleanup of the (many) big board variants is.

There are many things broken with the v1050, but I've never used the physical hardware, and don't have a reason to be working on it.

PR #15454, opened 2026-06-06
machine/gen_latch.cpp: add a setter for suppressing latch written messages in logerror

angelosa:
* hook it up to bus/nec_fdd/pc80s31k.cpp as example, where the verbosity is too much to be tolerable;


cuavas:
Those log messages should be compile-time opt-in, and shouldn't be called warnings – 7f9c46474e9eb1f30ab802daa29d13bb881e6816

PR #15456, opened 2026-06-06
bus/scsi/sa1403d: implement the Shugart SA1403D as a high-level SASI controller

davidlrand:
> **Stacked on #15459** (adds the `floppy_image_device::get_image()` accessor this HLE needs). The diff includes that one-line accessor commit until #15459 merges, then it drops to the device only.

Implements the Shugart **SA1403D** SASI/floppy controller (previously a skeleton) as a high-level emulation.

The SA1403D is a SASI-bus controller fronting up to four logical units. This HLE serves:

- **LUNs 0–2** from the host's floppy drives, wired by the owning driver via `set_floppy()`, decoding the mounted IMD/MFI track data on demand (density and sides per the Define-Floppy command).
- **LUN 3** from a rigid-disk CHD image (the inherited `scsihd` path).

Commands implemented: SASI READ / WRITE / SEEK (6-byte CDBs), Format Track (op `0x06`, fills the track with `0xE5`), and the SA1403D-specific Define Floppy (op `0xC0`) density/sides select. Logical-block addressing follows the SA1403D OEM manual §5.5: `(cyl × heads + head) × spt + sector`.

The four undumped controller PROMs (I–IV) are noted in a comment rather than declared as `NO_DUMP` placeholders, since the controller is fully high-level emulated and never executes them — this keeps `-verifyroms` clean for a controller that is otherwise complete.

Builds on the legacy `scsihle` core fixed in #15453. The first consumer — the Xerox 820-II SASI hard-disk option — follows in a separate driver PR (where the device's `set_floppy()` LUN wiring and the end-to-end format/boot test live).

### Test plan
- [x] CI build (gcc/clang · Linux/Windows/macOS)
- [x] Exercised end-to-end in combination with the forthcoming Xerox 820-II SASI driver: rigid-disk format + file I/O, and CP/M served from the floppy LUNs.

### DIFF
--- a/src/devices/bus/scsi/sa1403d.cpp
+++ b/src/devices/bus/scsi/sa1403d.cpp
@@ -9,6 +9,29 @@
 #include "emu.h"
 #include "sa1403d.h"
 
+#include "formats/flopimg.h"
+
+namespace {
+// The bitstream/sector decode helpers are protected statics of floppy_image_format_t;
+// expose them via a trivial subclass so the SA1403D can decode sectors at HLE level.
+class flopdec : public floppy_image_format_t
+{
+public:
+	static std::vector<bool> bitstream(int t, int h, int cs, const floppy_image &img)
+	{ return generate_bitstream_from_track(t, h, cs, img); }
+	static std::vector<std::vector<uint8_t>> mfm(const std::vector<bool> &b)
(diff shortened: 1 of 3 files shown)


davidlrand:
I have re-written this to include a local wd_fdc-based floppy portion, based on the feedback.  I'll PR this shortly.

PR #15457, opened 2026-06-06
xerox/xerox820.cpp: working Xerox 820-II with SASI hard disk and PWB floppy variants

davidlrand:
> **Stack: #15459 (floppy accessor) → #15456 (SA1403D) → this PR.** The diff includes those commits until the lower PRs merge; I will rebase as each lands.

Brings the **Xerox 820-II** up to working — including its SASI hard-disk option — and splits the 820-II into its real PWB daughterboard variants.

**Stacked on #15456** (the SA1403D HLE): this branch contains that device commit, so the diff currently shows it too; it reduces to the driver changes once #15456 merges. Also builds on the `scsihle` fixes in #15453 (merged).

### Now working
- **Xerox 820-II (8" floppy)** — boots CP/M; promoted to working.
- **Xerox 820-II (5.25" floppy)** — new working clone.
- **Xerox 820-II (SASI hard disk)** — new working clone: boots, formats the rigid disk, and reads/writes files, with the SA1403D serving the floppy drives (LUNs 0–2) plus a rigid CHD (LUN 3).

The **Xerox 16/8** (5.25"/SASI) variants are added but left **not working** — the 16/8's 8086/CP/M-86 side is untested (no bootable CP/M-86 media located yet).

### Changes
- Split `xerox820ii` into a shared `xerox820ii_common()` plus per-personality machine configs (8" floppy / 5.25" floppy / 8"+SASI), mirroring the real daughterboard options, and add the 16/8 (8086) variants.
- Wire the SA1403D for the hard-disk personality: floppy drives on LUNs 0–2 via `set_floppy()`, a rigid CHD on LUN 3, the boot ROM's Seltab drive→LUN map, and the SA1403D logical-block addressing.
- **Keyboard**: the 8748 (`x820kb`) is a standard US ASCII keyboard — confirmed by disassembling the keyboard ROM's translation tables. Fixes the KBSTB strobe polarity and adds the 820-II power-on "keyboard available" announce its monitor waits for. (The ROM also contains an alternate `$1FE` bit-paired/Teletype encoding selected by a row-14 strap; that's for a different keyboard variant and is left unselected.)

### Disclosed workaround
[...]

PR #15459, opened 2026-06-07
imagedev/floppy: add a get_image() accessor for the mounted image

davidlrand:
Adds a small public accessor returning the `floppy_image` the drive currently holds:

```cpp
floppy_image *get_image() const { return m_image.get(); }
```

This is for **high-level controller emulations that decode the mounted media themselves** rather than going through an FDC — e.g. the Shugart **SA1403D** SASI controller (a SASI/floppy bridge), which reads the image to serve sectors over the bus.

Prerequisite for the SA1403D HLE (#15456) and the Xerox 820-II SASI driver (#15457).

### DIFF
--- a/src/devices/imagedev/floppy.h
+++ b/src/devices/imagedev/floppy.h
@@ -293,6 +293,7 @@ public:
 
 	int get_cyl() const { return m_cyl; }
 	bool on_track() const { return !m_subcyl; }
+	floppy_image *get_image() const { return m_image.get(); } // for HLE controllers that read the mounted image directly
 
 	virtual void mon_w(int state);
 	bool ready_r();


MooglyGuy:
Is there a specific case where `image_core_file()` in the base `device_image_interface` class, and the facilities provided by the `core_file` class, are insufficient for format determination?

davidlrand:
The purpose is in emulation of the SA1403D controller (specifically for the Xerox 820-ii rigid disk, but this works in general too).

The controller doesn't need to determine the container format, floppy_image_device already does that at mount time  and decodes the container into the in-memory floppy_image. What this controller needs is that decoded image: it's an  SASI to floppy bridge with no FDC identified, so it reproduces the read path at the cell level generate_bitstream_from_track() then extract_sectors_from_bitstream_{mfm,fm}_pc(), probing 8″/5.25″ cell timings and FM/MFM, with wrong-density reads failing so the Xerox ROM's SD/DD detector works. 

image_core_file() is one layer too low (raw container bytes); using it would mean re-running the format identify/load path for every supported container inside the controller. m_image has no existing public accessor, hence the one-liner.

All of that said, it is equally possible to re-write this to  use, for example, wd_fdc, and pull it up to a full "emulation model".  The difference would be that wd_fdc has no synchronous "read sector N" call, so I would simply create a local subdevice, and re-do the FM/MFM detection.  This would covert the SA1403D into an asynchronous state machine.  

It's a functional stand-in, and can be done.  I chose the "more faithful" path, but I can switch if you feel this is violating the mame principles.

So far as I can tell, this is the *only* implementation of a SASI/SCSI controller in mame which supports floppy drives, so there was no precedent for me to follow.

MooglyGuy:
Oh, yeesh, so yeah - grabbing at the raw container bytes would essentially necessitate copy/pasting code to identify each format into the controller. That would be way leakier of an abstraction than this change.

I know @cuavas tends to feel certain ways about adding accessors, so I defer to him on this, but it sounds like the juice isn't worth the squeeze to try to do it any other way. Adding the accessor sounds like the least-agonizing option.

davidlrand:
Yeah.  Let me withdraw this, and think some more, before I finalize.  I'm still in the midst of the xerox 8/16 5" disk bringup, and that has the same issue as the 8" 820II-with-sasi, and there may be more work there.

PR #15460, opened 2026-06-07
bus/spectrum/mgt.cpp: Used joystick's keys for mapping instead of creating conflict with base driver (MT08683)

holub:


holub:
I think I better understand why it's happening now. I'll finalize testing and add comment on MT

PR #15461, opened 2026-06-07
xerox/xerox820.cpp: Get the Digital Research Computers Big Board booting CP/M

davidlrand:
The original Ferguson **Big Board** (`bigboard`) and its **MK-82** clone were `MACHINE_NOT_WORKING`. The shared `xerox820.cpp` driver configured them like the Xerox 820 — 5.25" SA-400 drives at a 1 MHz WD1771 clock with bit-per-drive select — but the Big Board uses **8" SSSD drives read by the WD1771 in FM**, selected by a **binary drive-unit number** in the system PIO port. Three changes get both booting CP/M:

- **8" drives + clock + format.** `bigboard()` now fits 8" Shugart SA-800 drives at the 8" 2 MHz WD1771 clock and registers the IMD floppy format the disks ship in.
- **Binary drive select.** The monitor's `SELECT` writes a binary unit number (drive 0 = `00`) in the low bits of system-PIO port A, demuxed to one select line; the shared `kbpio_pa_w` decoded one-bit-per-drive, so unit 0 selected nothing. An `m_drvsel_binary` flag (set only for the Big Board) switches the decode; the Xerox 820 path is unchanged.
- **Motor.** 8" Shugart drives spin continuously (no motor-control line), so the motor is forced on at reset and the WD1771 sees READY for the boot ROM's RESTORE.

The data-bus inversion and `/HALT`-gated DRQ→NMI byte transfer the boot relies on are already correct in master (`fdc_r ^ 0xff`, `update_nmi`).

`mojmikro` ("Moj mikro Slovenija") is a localized Big Board clone — its 2 KB monitor carries the identical Big Board command set and 0x00–0x1F I/O map — and already uses the Big Board machine config, so the same fixes bring it up too; promoted to working.

A **`bigboard5`** clone is also added for the common field modification that ran the Big Board from 5.25" drives in place of the stock 8" — left `MACHINE_NOT_WORKING` as I have no 5.25" media to verify it.

Tested: `bigboard`, `mk82` and `mojmikro` boot "BIG BOARD CP/M 2.2" to `A>` and `DIR` lists the disk (8" SSSD IMD media from the Ferguson Big Board reproduction project). `x820` and the other variants are untouched.

Systems promoted to working
---------------------------
Big Board [Dave Rand]

Clones promoted to working
--------------------------
MK-82 [Dave Rand]
Moj mikro Slovenija [Dave Rand]

New clones marked not working
-----------------------------
Big Board (5.25" drives) [Dave Rand]

### DIFF
--- a/src/mame/mame.lst
+++ b/src/mame/mame.lst
@@ -49343,6 +49343,7 @@ notetakr
 
 @source:xerox/xerox820.cpp
 bigboard
+bigboard5
 mk82
 mk83
 mojmikro
--- a/src/mame/xerox/xerox820.cpp
+++ b/src/mame/xerox/xerox820.cpp
@@ -62,6 +62,7 @@
 #include "speaker.h"
 
 #include "formats/flopimg.h"
+#include "formats/imd_dsk.h"
 
 
 /* Read/Write Handlers */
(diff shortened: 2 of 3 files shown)


davidlrand:
Screenshot 2026-06-06 at 8 04 02 PM

davidlrand:
Added **Moj mikro Slovenija** (`mojmikro`) to this PR — it's a localized Big Board clone (same monitor command set and I/O map, already on the Big Board machine config), so the same fixes apply with no further code change. It boots Big Board CP/M 2.2 to `A>` and `DIR` lists the disk. Force-pushed the amended commit.

